PR #2114, opened 2020-08-11
Do not accept new tx via API if node is not in sync

hoenirvili:
## Motivation
When we submit a transaction we should verify if the node is in sync first.

Closes #2103 

## Changes
I've added two checks one that checks if the transaction byte payload is even filled and one for testing the equality between the number of layers processed and the last layer in state.

## Test Plan

Looks like the TransactionService by default is not turn on and I did first turned on the transaction service in code (config) and then started the node(that is not in sync) and executed a grpcurl comand with the **xdr** serialized payload.

## TODO

- [x] Explain motivation or link existing issue(s)
- [x] Test changes and document test plan
- [x] Update documentation as needed

PR #2139, opened 2020-09-10
Do not accept new tx via API if node is not in sync (take two)

### LINKED ISSUE #2103: API: don't allow submission of new transactions until node is synced
## Description
Right now, the TransactionService GRPC server accepts new transactions via `SubmitTransaction`, even while the node is syncing/not participating in gossip, without an error. However, incoming tx data is transmitted as a gossip message, which means that it's dropped silently if the node is still syncing. This should be an error in the GRPC service.

Note that this is also an issue in the old GRPC API, but we don't need to fix it since that will be deprecated soon in favor of the new API.

## Affected code
api/grpcserver/transaction_service.go

This issue appears in commit hash: a1f0055

This is what the logs look like now:

```
2020-08-06T11:51:33.442-0400    INFO    00000.defaultLogger     GRPC TransactionService.SubmitTransaction
2020-08-06T11:51:33.442-0400    INFO    00000.defaultLogger     GRPC TransactionService.SubmitTransaction to address: 0xf9bc4f997d235bA5a9A36C1A92D07882C2B2FdB6 (len: 20), amount: 1, gaslimit: 2, fee: 1
2020-08-06T11:51:33.443-0400    INFO    00000.defaultLogger     GRPC TransactionService.SubmitTransaction BROADCAST tx address f9bc4f997d235ba5a9a36c1a92d07882c2b2fdb6 (len 20), gas limit 2, fee 1 id 56abc07a23 nonce 0
```

And if you subsequently submit a second tx with a higher nonce:

```
2020-08-06T11:51:33.448-0400    INFO    00000.defaultLogger     GRPC TransactionService.SubmitTransaction
2020-08-06T11:51:33.448-0400    INFO    00000.defaultLogger     GRPC TransactionService.SubmitTransaction to address: 0xECAca87B528E8f87602D21Aa0150148902e47508 (len: 20), amount: 1, gaslimit: 2, fee: 1
2020-08-06T11:51:33.448-0400    ERROR   00000.defaultLogger     tx failed nonce and balance check: incorrect account nonce! Expected: 0, Actual: 1
```

lrettig:
(This is a rebase of #2114 against the latest develop branch authored by @hoenirvili)

## Motivation
When we submit a transaction we should verify if the node is in sync first.
Closes #2103 

## Changes
I've added two checks one that checks if the transaction byte payload is even filled and one for testing the equality between the number of layers processed and the last layer in state.

## Test Plan

Looks like the TransactionService by default is not turn on and I did first turned on the transaction service in code (config) and then started the node(that is not in sync) and executed a grpcurl comand with the **xdr** serialized payload.

## TODO


lrettig:
@hoenirvili I cherry-picked the commits from #2139 here onto the latest develop branch and there were no conflicts

api/grpcserver/transaction_service.go
```diff
@@ -20,6 +21,14 @@ type TransactionService struct {
 	Network api.NetworkAPI // P2P Swarm
 	Mesh    api.TxAPI      // Mesh
 	Mempool *state.TxMempool
+	syncer  Syncer
+}
+
+// Syncer defines ways of interacting with the node syncer
+// business logic
+type Syncer interface {
```
lrettig, line 29: Following up on the thread from the previous PR (https://github.com/spacemeshos/go-spacemesh/pull/2114#discussion_r482355573), in principle I agree with your point about the "interface segregation principle," but I don't think it's worth defining a new interface here for a single method - especially since, as I pointed out, the object being passed in already implements both methods.

In any case we do this many times in our code already - if we think that should be changed, we should probably do that as a separate issue.

hoenirvili:
So if the pipeline succeeds this can be merged?

lrettig (commented):
See one attached comment. Let's see if the tests pass. I can't approve this PR since I opened it, so either you'll need to open another new PR (that's annoying) or else we need someone else to approve this one.

api/grpcserver/transaction_service.go
hoenirvili: Ok, I'll then reuse the interface.

hoenirvili:
I think you can close the pr.

lrettig:
Closing in favor of #2146

PR #2141, opened 2020-09-15
[Merged by Bors] - only close the socket from the higher level. if encountered read/write…

y0sher:
… error it's already closed.

## Motivation

Fix the bug that occurred in testnet. 
bug flow: once a connection had to be removed because our cache was full. we would call Close on this connection. which will close the closeChan channel. this will result in an error returned inside `MsgConnection`'s reading flow which will call the same `Close` again in result in close of a closed channel.

## Changes

only close the udp socket from the higher level. if encountered read/write error it's already closed.

## Test Plan
added testing.

### DIFF
--- a/p2p/net/udp.go
+++ b/p2p/net/udp.go
@@ -280,7 +280,7 @@ func (n *UDPNet) addConn(addr net.Addr, ucw udpConn) {
 			if time.Since(c.Created()) > maxUDPLife {
 				delete(n.incomingConn, k)
 				if err := c.Close(); err != nil {
-					n.logger.Warning("Error closing udp connection: %s", err)
+					n.logger.With().Warning("Error closing udp socket", log.Err(err))
 				}
 				evicted = true
 				break
@@ -288,7 +288,9 @@ func (n *UDPNet) addConn(addr net.Addr, ucw udpConn) {
 		}
 
 		if !evicted {
-			n.incomingConn[lastk].Close()
+			if err := n.incomingConn[lastk].Close(); err != nil {
+				n.logger.With().Warning("Error closing udp socket", log.Err(err))
+			}
 			delete(n.incomingConn, lastk)
 		}
 	}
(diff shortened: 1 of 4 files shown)


p2p/net/udp_test.go
```diff
@@ -294,3 +312,63 @@ func TestUDPNet_Cache(t *testing.T) {
 	require.True(t, closed)
 
 }
+
+func TestUDPNet_Cache2(t *testing.T) {
+	// This test fails by panic before latest fix to udp.go. it crashed the testnet.
+	// instead
```
lrettig: something missing from this comment?

p2p/net/udp.go
```diff
@@ -274,6 +274,7 @@ func (n *UDPNet) addConn(addr net.Addr, ucw udpConn) {
 	evicted := false
 	lastk := ""
 	if len(n.incomingConn) >= maxUDPConn {
+		n.logger.Debug("Removing some udp session")
 		for k, c := range n.incomingConn {
 			lastk = k
 			if time.Since(c.Created()) > maxUDPLife {
```
lrettig, line 280: This code path isn't tested in the test, since `c.Created()` will never be `> maxUDPlife`, could we test this code path too?

lrettig (dismissed):
Make a couple of small tweaks to log/test, overall looks good, see one comment about untested code path. Also confirmed locally that the test fails on develop and passes with these changes 👍

p2p/net/udp_test.go
y0sher: lol you are right :) adding

p2p/net/udp.go
y0sher: sure, you are right.

noamnelke (approved):

y0sher:
bors merge

y0sher:
dismissed @lrettig to make bors work, I addressed the comments in last commit. thanks for the help :)!
bors merge

PR #2142, opened 2020-09-15
Add NTP servers accessible in China

### LINKED ISSUE #2063: NTP sync failing in China
One of our users running a node in mainland China reports that he keeps losing his connection, and has to restart the app several times per day - although he is able to smesh and is receiving rewards. Here's a snipped from a recent logfile. Could anyone help me figure out what's going on here?

Some stuff I noticed:
- the node is unable to synchronize with several of the NTP servers, but it is able to reach the Apple Asia server
- it keeps bootstrapping, even though the address book is full
- there appear to be many P2P errors

Here's the log: https://gist.github.com/lrettig/968ce46ab928c29b87b5aceb5a9210b0

lrettig:
## Motivation
Fixes #2063 

## Changes
Adds a bunch of NTP servers that should be accessible in China. Also adds `pool.ntp.org` which should work anywhere. Sources:

- https://www.ntppool.org/en/
- https://www.pool.ntp.org/zone/cn
- https://intl.cloud.tencent.com/document/product/213/32379
- https://partners-intl.aliyun.com/help/doc-detail/92704.htm

## Test Plan
No tests required, no change in functionality.
Will ask community members in China to test.


lrettig:
@noamnelke I added servers as discussed in #2063. I tried to keep this as simple as possible for now but note that this is not an ideal solution for two reasons:

- probabilistically nodes in China will still fail occasionally. We currently make five requests and the node exits if < 3 of these succeed. If it unluckily happens to choose, e.g., the Google servers, it will still fail.
- nodes outside China will end up syncing time against the China NTP servers - I tested them locally and they all work for me here, but this isn't ideal.

Ways we might decide to improve this:

1. don't exit after a single round of failures - add a retry, or several retries.
1. try to only list servers that work everywhere - that would mean, e.g., removing the google servers. `pool.ntp.org` should theoretically work anywhere.
1. do something smarter/more complicated, e.g., detecting the node's region (or letting the user specify it in the config) and choosing a pool of servers based on this.

noamnelke:
You're right about the possibility of it failing even after this change. I think there's an easy solution, though, that doesn't require any code changes. The number of NTP queries, where 3 or more must succeed, which you quoted as 5 is configurable.

For a user in China, there's a 55% chance of success at randomly picking a working server. I asked Sofi to [run the numbers for me](https://docs.google.com/spreadsheets/d/13ZWFC4S3kssYl52yOpbMHU2WTSwd03d7R4TwQfQhvUk/edit?usp=sharing) and we need 18 attempts to get to a total of 99.99% success rate. It really depends on how often we run this check. The current default check interval is 30 minutes, so it's expected to fail once every ~20 days.

If we make `ntp-queries=15` and increase the number of tries here to 3, the probability of failure becomes negligible (0.000000001 = 1/1,000,000,000 = failure every ~57 years):

https://github.com/spacemeshos/go-spacemesh/blob/ce9d5da1a17aee64458de361cbe4f85a03d1b49f/timesync/ntp.go#L193

@y0sher @lrettig you see any problems with what I suggested?

y0sher:
Sounds good to me. 
note about `pool.ntp.org`. we removed it since we had weird problem, it might be related to the servers themselves or maybe something with our cloud provider (gcp), requests just came back empty even though there was no error. we couldn't reproduce as it was really rare and happened only on the gcp machine. but we couldn't write code to properly handle this since we didn't know what exactly happened and we couldn't reproduce.

noamnelke:
@lrettig given what @y0sher wrote about pool.ntp.org, it's probably best to remove it.

lrettig:
> you see any problems with what I suggested?

It totally makes sense and I trust your math @noamnelke.

lrettig:
> If we make `ntp-queries=15` and increase the number of tries here to 3

@noamnelke I made the first change, not sure what you meant about "increase the number of tries" as this is already what it's set to - please double check but I think we're good here.

noamnelke:
@lrettig 
> already what it's set to

You're right, I misread the code (the line I referenced is the initialization, not the number of attempts) - we can just leave it at 3, as it is now.

y0sher (approved):
seems like you got this. hope we won't have issues with `poo.ntp.org` servers. also adding so much china server won't impact the non-china miners ?

noamnelke:
let's not merge this for now.

@y0sher 
> adding so much china server won't impact the non-china miners ?

It will. I agree with @lrettig that we should make the list configurable, so users in China can use Chinese servers and Americans can use American servers.

There's probably more we can do to improve the experience here, like randomize the order of servers and when one fails put it at the bottom, so it's tried last next time. Try a few servers concurrently and stop waiting once we have enough successes, then put the successful servers at the top of the list so the fastest ones are tried first next time. If we had that, we could include both Chinese and western servers and it would just sort itself out.

lrettig:
Okay, once the new testnet is up let's start with a separate app/node build
for our Chinese smeshers and see if this list of NTP servers works better
for them.

>

lrettig:
FYI I created a new tag for the current testnet that includes these changes, and I've asked some of our Chinese smeshers to test it: https://github.com/spacemeshos/go-spacemesh/releases/tag/v0.1.14-ntpfix

lrettig:
One of our community members in China has been testing the pool of candidate NTP servers we identified. They report that on one system, they're all producing errors except nist.gov, while on another system on a different network, they all work. Not sure what to make of these results yet, I think we need to test more. (Most of this testing has been at the OS level, outside of go-spacemesh, but the same community member is also testing running nodes.)

I'll post more here when I learn more.

lrettig:
This was handled in #2221

Issue #1371, opened 2019-08-25
Rename "Nipst" as "NIPoST" throughout the codebase and docs


moshababo:
# Overview / Motivation
As discussed and agreed with Tal Moran from here on "Nipst" will be forever changed to "NIPoST".

# The Task
Rename "Nipst" as "NIPoST" throughout the codebase and docs.

moshababo:
Closed by https://github.com/spacemeshos/go-spacemesh/pull/2149.

PR #2131, opened 2020-08-29
[Merged by Bors] - API phase 11: cleanup old API and protoc code

lrettig:
## Motivation
See #1764 
Closes #2077, Closes #2078, Closes #2153, Closes #411, Closes #2163, Closes #2145, Closes #2168

## Changes
- Removes old GRPC API and HTTP-JSON gateway code
- Removes protobuf code
- Updates tests to use new API
- Lots of cleanup of the system tests

## Test Plan
No new tests required, this does not add any new functionality. Affected tests are updated/removed.

## TODO
- [x] Remove old API code
- [x] Update tests to use new API instead
- [x] Remove all protobuf-related code and references
- [x] Update README
- [x] Update to API v1.0.0, adding GatewayService to new API for poet (#2133)
- [x] Implement GatewayService (#2156)
- [x] Update poet (https://github.com/spacemeshos/poet/issues/98)
- [x] Fix p2p tests broadcast issue
- [x] Update hardcoded go-spacemesh docker images in configs
- [x] Open issue to replace all `curl`/json http gateway server API interactions in tests with python grpc library (#2185)
- [x] Update system tests (including tx_generator WalletAPI, test_p2p) to use new API
- [x] Make sure spacecraft has been updated to factor in new ports and commandline flags (https://github.com/spacemeshos/spacecraft/issues/1)
- [x] Update hardcoded poet docker image spacemeshos/poet:develop, fix in configs (needs https://github.com/spacemeshos/poet/pull/105)
- [x] Merge #2157 
- [x] Merge #2159
- [x] Merge #2169

PR #2149, opened 2020-09-24
[Merged by Bors] - Smesher API + GPU/Tree-free PoST integration

moshababo:
Implemented the new [Smesher API](https://github.com/spacemeshos/api/blob/master/proto/spacemesh/v1/smesher.proto) (instead of the previous Mining API), while integrating GPU/Tree-free PoST.
This PR is based on https://github.com/spacemeshos/post/pull/41, which need to be reviewed first, hence this description won't repeat what was already been specified there.

Main remarks:

* Created `PostSetupManager` for implementing the new Post setup API. The previous `PostClient` was removed, and setup functionalities previously implemented by the ATX builder were removed as well. 
* Refactored the ATX builder to comply with the new Smeshing API. Its dependency with `PostSetupManager` is only regarding the setup completion status and generation of proofs. 
* In addition to the raw Post, the NIPost now contains the Post metadata as well, that the proof is expected to be verified upon during the syntactic validation. Some of the metadata values are expected to be derived from consensus, hence should be validated in the contextual validation, but it is currently still being derived from the node configuration, so it’s still taking place in the syntactic validation (which will need to remain anyway for validating min/max boundaries).
* Added `PostSetupOpts` config. It is used for initiating a new data creation session - either via the Smesher API, or on node start. 
* Renamed the ATX’s `commitment` as `initialPost`. As discussed with @tal-m, in the tree-free PoST the proof is *not* a commitment to the data.
* Changed NIPoST’S PoET challenge to contain the initial Post indices, instead of its Merkle root, which is no longer applicable. The reason is the same - it is expected to be known at the time the challenge was constructed. 
* Renamed “nipst” as “NIPost” throughout the codebase (closing https://github.com/spacemeshos/go-spacemesh/issues/1371).
* Some of the old API code and test-code was commented out or disabled due to compile-time errors caused to the new defined interfaces. It is expected to be fully removed soon.


api/grpcserver/smesher_service.go
```diff
@@ -25,40 +27,34 @@ func (s SmesherService) RegisterService(server *Server) {
...
-	isSmeshing := stat == activation.InitDone
-	return &pb.IsSmeshingResponse{IsSmeshing: isSmeshing}, nil
+	return &pb.IsSmeshingResponse{IsSmeshing: s.smeshing.Smeshing()}, nil
 }
 
 // StartSmeshing requests that the node begin smeshing
 func (s SmesherService) StartSmeshing(ctx context.Context, in *pb.StartSmeshingRequest) (*pb.StartSmeshingResponse, error) {
 	log.Info("GRPC SmesherService.StartSmeshing")
 
+	// TODO(moshababo): remove DataDir and CommitmentSize from the request proto definition.
```
lrettig: Please open an issue for this here: https://github.com/spacemeshos/api/issues

api/grpcserver/smesher_service.go
```diff
@@ -25,40 +27,34 @@ func (s SmesherService) RegisterService(server *Server) {
...
+	// TODO(moshababo): check why JSON request via HTTP gateway doesn't decode `coinbase` properly
```
lrettig: This is a bit tricky since it's expecting bytes, not a string. Possibly useful: https://github.com/spacemeshos/api/issues/109

api/grpcserver/smesher_service.go
```diff
@@ -78,21 +76,16 @@ func (s SmesherService) StopSmeshing(ctx context.Context, in *pb.StopSmeshingReq
...
 
-	_, _, coinbase, _ := s.Mining.MiningStats()
-	addr, err := types.StringToAddress(coinbase)
-	if err != nil {
-		log.Error("error converting coinbase: %s", err)
-		return nil, status.Errorf(codes.Internal, "error reading coinbase data")
-	}
-	return &pb.CoinbaseResponse{AccountId: &pb.AccountId{Address: addr.Bytes()}}, nil
+	addr := s.smeshing.Coinbase()
+	return &pb.CoinbaseResponse{AccountId: &pb.AccountId{Address: addr[:]}}, nil
```
lrettig: For my own edification, what does `addr[:]` do? Is it the same as `addr.Bytes()`?

api/grpcserver/smesher_service.go
```diff
@@ -125,31 +119,112 @@ func (s SmesherService) SetMinGas(ctx context.Context, in *pb.SetMinGasRequest)
...
-	return nil, status.Errorf(codes.Unimplemented, "this endpoint is not implemented")
+
+	providers := s.post.PostComputeProviders()
+
+	res := &pb.PostComputeProvidersResponse{}
+	res.PostComputeProvider = make([]*pb.PostComputeProvider, len(providers))
+	for i, p := range providers {
+		hs, err := p.Benchmark()
+		if err != nil {
+			return nil, err
```
lrettig: Please don't return `err` directly here. The pattern we've been using is to pass the raw error to `log.Error` and then return `status.Error(codes.Internal, ...)` with a short description of the error fit for external consumption.

api/grpcserver/smesher_service.go
```diff
@@ -125,31 +119,112 @@ func (s SmesherService) SetMinGas(ctx context.Context, in *pb.SetMinGasRequest)
...
+	return res, nil
 }
 
 // CreatePostData requests that the node begin post init
 func (s SmesherService) CreatePostData(ctx context.Context, in *pb.CreatePostDataRequest) (*pb.CreatePostDataResponse, error) {
 	log.Info("GRPC SmesherService.CreatePostData")
-	return nil, status.Errorf(codes.Unimplemented, "this endpoint is not implemented")
+
+	if in.Data == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "`Data` must be provided")
```
lrettig:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "`Data` must be provided")
```

api/grpcserver/smesher_service.go
```diff
@@ -125,31 +119,112 @@ func (s SmesherService) SetMinGas(ctx context.Context, in *pb.SetMinGasRequest)
...
+	}
+
+	if _, err := s.post.CreatePostData(&activation.PostOptions{
+		DataDir:           in.Data.Path,
+		DataSize:          in.Data.DataSize,
+		Append:            in.Data.Append,
+		Throttle:          in.Data.Throttle,
+		ComputeProviderID: int(in.Data.ProviderId),
+	}); err != nil {
+		return nil, err
```
lrettig: Same comment as above, please wrap the error

api/grpcserver/smesher_service.go
```diff
@@ -125,31 +119,112 @@ func (s SmesherService) SetMinGas(ctx context.Context, in *pb.SetMinGasRequest)
...
+	}, nil
 }
 
 // StopPostDataCreationSession requests that the node stop ongoing post data creation
 func (s SmesherService) StopPostDataCreationSession(ctx context.Context, in *pb.StopPostDataCreationSessionRequest) (*pb.StopPostDataCreationSessionResponse, error) {
 	log.Info("GRPC SmesherService.StopPostDataCreationSession")
-	return nil, status.Errorf(codes.Unimplemented, "this endpoint is not implemented")
+
+	if err := s.post.StopPostDataCreationSession(in.DeleteFiles); err != nil {
+		return nil, err
```
lrettig: Same comment as above, please wrap the error

lrettig (dismissed):
Only reviewed the API changes. See inline comments. Also noticed that you added a few TODO items, it would be nice if they were linked to open issues if they remain when the code is merged.

lrettig:
Some of this is going to cause conflicts with #2131, which totally removes the old API code. Consider leaving these changes out of this PR, as that one will blow away this code entirely.

api/grpcserver/smesher_service.go
moshababo: https://github.com/spacemeshos/api/issues/114

api/grpcserver/smesher_service.go
moshababo: I've used a similar encoding (and inspected the encoding used in responses), but it doesn't work well for `coinbase` field for some reason (unlike `account_id`, both of the same type). 
We also have examples of this in https://github.com/spacemeshos/cliwallet.

My format:
```
{
    "coinbase": {
        "address": "QWxhZGRpbjpvcGVuIHNlc2FtZQ=="
    }
}
```
I've tried it a while ago. Will give it another shot before merging, just for clients' well-being.

api/grpcserver/smesher_service.go
moshababo: Yes, it's the same. I guess I changed it after I rewrote this section, and didn't notice the `Bytes()` alias, which is probably preferable. Fixed.

moshababo:
@lrettig thanks for reviewing. 

---

> Only reviewed the API changes. See inline comments. Also noticed that you added a few TODO items, it would be nice if they were linked to open issues if they remain when the code is merged.

They will be solved prior to merging. 

> Some of this is going to cause conflicts with #2131, which totally removes the old API code. Consider leaving these changes out of this PR, as that one will blow away this code entirely.

I've made changes only to avoid compile-time and test errors. If you expect #2131 to be merged soon, i'll wait and then solve conflicts.

api/grpcserver/smesher_service.go
lrettig: That's really strange. `coinbase` in the SmesherService is of type `account_id` which is just an address passed as an array of bytes, as elsewhere in the API. Do you have sample curl commands, one that works somewhere else and one that fails here?

lrettig:
> If you expect #2131 to be merged soon, i'll wait and then solve conflicts.

I'm still working on it, but I expect it to be ready to be merged by next week.

api/grpcserver/smesher_service.go
avive: It should work ok. CLIWallet uses binary data for all addresses in GRPC calls and it works as expected. There's no need to base64 binary data in grpc - only when working higher-level via the json-proxy as there's no binary data in json.

avive:
The implementation of the pos progress status stream throws an exception if a pos init process is not currently in progress. The desirable pattern for clients is to subscribe to events on this stream _before_ setting up pos so they can get all errors and status on it while it is ongoing. The stream should be closed by client at any time but never by node (except maybe when running out of resources) . Expecting users to subscribe to stream only _after_ starting a pos init job is not a good pattern because they can miss error events on the stream this way and this is not the normal usage pattern of grpc streaming apis. In addition, I noticed that the stream by the node when a pos setup is complete - this also shouldn't be done and any future pos operation should send data over the stream before the client closed it. I think this is the most straightforward way to implement this kind of functionality.

moshababo:
> The implementation of the pos progress status stream throws an exception if a pos init process is not currently in progress. The desirable pattern for clients is to subscribe to events on this stream before setting up pos so they can get all errors and status on it while it is ongoing. 

Fixed. (also see https://github.com/spacemeshos/cli-wallet/pull/23#issuecomment-855256532)

> The stream should be closed by client at any time but never by node (except maybe when running out of resources)

> In addition, I noticed that the stream [IS CLOSED] by the node when a pos setup is complete - this also shouldn't be done and any future pos operation should send data over the stream before the client closed it. I think this is the most straightforward way to implement this kind of functionality.

I’d rather finalize the stream functionality in a follow-up PR, and focus now on the finishing the general API design and implementation.

avive:
ok

avive:
Building this branch fails on Windows 10 pro with a gcc linker error where building developer branch builds just fine. Likely a GCC issue due to differences between behavior on unix and windows systems. Env: PowerShell / Mingw64.

common/types/activation.go
```diff
@@ -82,13 +82,13 @@ func CalcAtxHash32Presorted(sortedView []ATXID, additionalBytes []byte) Hash32 {
...
-// ActivationTxHeader is the header of an activation transaction. It includes all fields from the NIPSTChallenge, as
+// ActivationTxHeader is the header of an activation transaction. It includes all fields from the NIPoSTChallenge, as
 // well as the coinbase address and total weight.
 type ActivationTxHeader struct {
-	NIPSTChallenge
+	NIPoSTChallenge
 	id       *ATXID // non-exported cache of the ATXID
-	Space    uint64
 	Coinbase Address
+	NumUnits uint
```
noamnelke, line 91: I think it's better to use `uint64` as the header ends up on the wire and this makes the encoding more explicit (this is assuming that `uint`s are encoded as 64-bit unsigned integers anyway and not in an architecture-dependent way, which would be even worse).

moshababo:
bors try

sudachen:
bors try

sudachen:
bors try

noamnelke:
bors try

noamnelke:
bors try

noamnelke:
bors try

noamnelke:
bors try

moshababo:
bors try

moshababo:
bors merge

noamnelke (approved):

common/types/activation.go
moshababo: Yes, already wrote it down for a follow-up PR. Thanks for reminding.

moshababo:
bors merge

moshababo:
bors merge

PR #2157, opened 2020-10-07
[Merged by Bors] - Upgrade to API v1.0.0

### LINKED ISSUE #2104: API account counter issue
## Description
Calling `GlobalStateService.Account` returns the actual account counter (nonce) (along with balance) from global state. However, this is _not_ the counter value that should be included in a newly-submitted tx. For a new tx, the system expects the counter to match the _projected_ counter value returned by `TransactionProcessor.GetProjection`, which factors in pending transactions:

https://github.com/spacemeshos/go-spacemesh/blob/1c818789facf0e7219feb7d8d5737ce6b6776835/mesh/meshdb.go#L576-L584

If a client naively reads account data from the API, then tries to submit a tx, it will fail as a result.

CC @noamnelke @avive, would appreciate your input on this.

## Steps To Reproduce
1. Call `GlobalStateService.Account`
2. Submit a new, valid tx to `TransactionService.SubmitTransaction` using the account counter you just got. If there are any pending txs from this sender, it will fail.

## Expected Behavior
I see a few options:

- return projected counter and balance instead of current values from global state - probably a bad idea
- add a new API endpoint to get _projected_ account data
- add projected data to the existing account API endpoints

### LINKED ISSUE #2133: Upgrade to latest API version
The API protobuf in https://github.com/spacemeshos/api has gotten somewhat ahead of what's implemented here. Pull in the latest version of the API go module and update the implementation as necessary.

lrettig:
## Motivation
Closes #2133 
Closes #2104 

## Changes
- Upgrade to spacemesh API v1.0.0. Major API changes include:
  - split account state into current/projected (spacemeshos/api#111)
  - align node error types with zapcore error types (spacemeshos/api#102)
  - harmonize layernum type as uint32 (spacemeshos/api#120)
  - add an unimplemented SmesherService endpoint (EstimatedRewards) (spacemeshos/api#101)
- Upgrade google modules (required by API)

## Test Plan
No new tests are needed. No new functionality has been added to the API. Existing tests have been updated.

## Documentation
There are no docs to update. Master API docs live in the API repository and have already been updated. Implementation docs are still a work in progress in #2083.

## TODO
- [x] Finish upgrading API code

Note: Google modules were upgraded because cmd/sync imports google.golang.org/api/option, and the version it was importing was incompatible with google.golang.org/grpc >= v1.30.0. More info here:

- https://github.com/google/trillian/issues/2195
- https://github.com/etcd-io/etcd/issues/12124

Here's the error I was getting:

```
> go mod tidy
go: finding module for package google.golang.org/grpc/naming
github.com/spacemeshos/go-spacemesh/cmd/sync imports
        google.golang.org/api/option imports
        google.golang.org/api/internal imports
        google.golang.org/grpc/naming: module google.golang.org/grpc@latest found (v1.32.0), but does not contain package google.golang.org/grpc/naming
```

For reference here are the API changes: https://github.com/spacemeshos/api/compare/49b94a501b3e79f2975499dc35c2286d5a5ead17...v1.0.0


api/grpcserver/mesh_service.go
```diff
@@ -157,10 +156,15 @@ func (s MeshService) getFilteredActivations(startLayer types.LayerID, addr types
...
 
 // AccountMeshDataQuery returns account data
-func (s MeshService) AccountMeshDataQuery(ctx context.Context, in *pb.AccountMeshDataQueryRequest) (*pb.AccountMeshDataQueryResponse, error) {
+func (s MeshService) AccountMeshDataQuery(_ context.Context, in *pb.AccountMeshDataQueryRequest) (*pb.AccountMeshDataQueryResponse, error) {
 	log.Info("GRPC MeshService.AccountMeshDataQuery")
 
-	startLayer := types.LayerID(in.MinLayer)
+	var startLayer types.LayerID
+	if in.MinLayer == nil {
+		startLayer = 0
```
y0sher: it's 0 by default so you could just do 
```go
if in.MinLayer != nil {
   	startLayer = types.LayerID(in.MinLayer.Number) 
}
``` 
instead of if else

api/grpcserver/mesh_service.go
```diff
@@ -357,15 +361,27 @@ func (s MeshService) readLayer(layer *types.Layer, layerStatus pb.Layer_LayerSta
 }
 
 // LayersQuery returns all mesh data, layer by layer
-func (s MeshService) LayersQuery(ctx context.Context, in *pb.LayersQueryRequest) (*pb.LayersQueryResponse, error) {
+func (s MeshService) LayersQuery(_ context.Context, in *pb.LayersQueryRequest) (*pb.LayersQueryResponse, error) {
 	log.Info("GRPC MeshService.LayersQuery")
 
+	var startLayer, endLayer types.LayerID
+	if in.StartLayer == nil {
+		startLayer = 0
```
y0sher: same comment as above, you can make this a bit simpler

api/grpcserver/mesh_service.go
```diff
@@ -357,15 +361,27 @@ func (s MeshService) readLayer(layer *types.Layer, layerStatus pb.Layer_LayerSta
 }
 
 // LayersQuery returns all mesh data, layer by layer
-func (s MeshService) LayersQuery(ctx context.Context, in *pb.LayersQueryRequest) (*pb.LayersQueryResponse, error) {
+func (s MeshService) LayersQuery(_ context.Context, in *pb.LayersQueryRequest) (*pb.LayersQueryResponse, error) {
```
y0sher, line 362: in others places you just didn't include _

lrettig: if `in` is used, i.e. we need one named argument, then I think all of the arguments need names, right? that's why some include `_` and some don't

y0sher: you are right, I did a quick check on go playground but it yielded false result because I wrote it like this 
``` 
func F(int, x int) {}
   F(1,1)
```
now this worked because the compiler treated the first `int` as the name of the first parameter.

lrettig: wow that's funky! I didn't even know that was possible. good to know.

y0sher (approved):
LGTM :+1:

lrettig:
bors merge

PR #2060, opened 2020-07-02
[Merged by Bors] - Migrate CI to github actions

lrettig:
## Motivation

I've been struggling to reproduce errors I've seen in travis due to its ultra-wacky run environment (for background see https://gist.github.com/fulldecent/d84cd1abdcc6930865d1b862c4aed917 and https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally). I've had more luck with github actions (this tool lets you run them locally: https://github.com/nektos/act). Github actions also makes it very easy to run across an array of platforms.

## Changes
- Mimics travis config, runs all tests using github actions for PRs and push against develop, staging, and trying branches

## To do
- [x] Set up app tests
- [x] Set up system tests
- [x] Set up dockerbuild/push
- [x] Add env vars as repository secrets (done in repo settings, not in this PR)
- [x] Finalize workflow (re-enable unit tests, disable system tests on PR, re-enable system tests on push to staging/trying, etc., see https://github.com/spacemeshos/go-spacemesh/pull/2060#issuecomment-690746582)
- [x] Document workflow and rationale (see https://github.com/spacemeshos/go-spacemesh/pull/2060#issuecomment-690746582)
- [x] Add link from main README
- [x] Disable tests on some paths
- [x] Add slack notification

PR #2161, opened 2020-10-16
Tell bors to check GH actions job statuses

lrettig:
## Motivation
Missed this in #2060

## Changes
- adds GH actions job statuses to the list of statuses that bors checks for success before merge

For reference see https://github.com/bors-ng/bors-ng/issues/730

## Test Plan
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/spacemeshos/go-spacemesh)](https://goreportcard.com/report/github.com/spacemeshos/go-spacemesh)
 [![Bors enabled](https://bors.tech/images/badge_small.svg)](https://app.bors.tech/repositories/22421)
 <a href="https://godoc.org/github.com/spacemeshos/go-spacemesh"><img src="https://img.shields.io/badge/godoc-LGTM-blue.svg"/></a>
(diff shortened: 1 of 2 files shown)


lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try-

lrettig:
bors try

noamnelke (approved):
I'm wondering if there's no way to "bundle" these together. This will make it less likely to forget to add a new test here, for example.

lrettig:
> I'm wondering if there's no way to "bundle" these together. This will make it less likely to forget to add a new test here, for example.

Good idea. The only way I can think of doing that is to have a master list of these in the Makefile, then have `make` auto-generate `bors.toml` dynamically (rather than storing the file statically). We might have to play around to get the timing/configuration right: just generating it in make/CI might not be enough, we might need a git hook or a github hook. Let me know if you want to try that.

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

noamnelke:
@lrettig 
> Let me know if you want to try that.

Sounds too complicated. Wondering if we can't reference the workflow name here?

lrettig:
I'm closing this in favor of #2165. Let's consolidate there as that PR already includes these changes and it may be easier to merge everything together.

PR #2167, opened 2020-10-22
[Merged by Bors] - Add timeouts to GH actions workflow

lrettig:
## Motivation

I've noticed that these tests hang sometimes, this should make our lives a little easier by timing out tests if they're obviously hanging. I reviewed the times of the last few successful runs of each job and added a pretty large buffer in these timeouts. (see, e.g., https://github.com/spacemeshos/go-spacemesh/runs/1294892834?check_suite_focus=true)

## Changes
- Adds `timeout-minutes` to each workflow job (per https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes)
- Removes initial slack notification, as it's redundant since this test is so quick anyway

## Test Plan
N/A


noamnelke (approved):
I've noticed this, too. I'm wondering why they hang... Don't think we've seen this behaviour on Travis 🤔

I think that timing out is probably better than just hanging, but still pretty frustrating if we don't know why.

lrettig:
> I've noticed this, too. I'm wondering why they hang... Don't think we've seen this behaviour on Travis 🤔
> 
> I think that timing out is probably better than just hanging, but still pretty frustrating if we don't know why.

Agree. Took another quick look just now and I'm not sure. Here are some recent timeouts:

- https://github.com/spacemeshos/go-spacemesh/runs/1294942848?check_suite_focus=true
- https://github.com/spacemeshos/go-spacemesh/runs/1305072616?check_suite_focus=true
- https://github.com/spacemeshos/go-spacemesh/runs/1294892834?check_suite_focus=true
- https://github.com/spacemeshos/go-spacemesh/runs/1305304499?check_suite_focus=true
- https://github.com/spacemeshos/go-spacemesh/runs/1305072616?check_suite_focus=true

All of these failed right after `github.com/spacemeshos/go-spacemesh/events` (likely `github.com/spacemeshos/go-spacemesh/filesystem` or `github.com/spacemeshos/go-spacemesh/hare`), or else right after `github.com/spacemeshos/go-spacemesh/p2p/metrics` (`github.com/spacemeshos/go-spacemesh/miner`?). Can't tell for sure what's going on without enabling verbose log output.

Let's merge this and investigate further if the timeouts continue.

lrettig:
bors merge

lrettig:
bors merge

Issue #2168, opened 2020-10-23
P2P version comparison logic is broken


lrettig:
## Description

There are several issues with the both this logic and the way it's tested:

- Timeout looking for bootstrap message in test wasn't long enough so it was always passing
- The actual logic that compares versions during bootstrap is broken, leading to nonsensical errors like `Error handing network message err=wrong client version want atleast: 0.0.1, got: go-spacemesh/0.0.2, err=<nil>`
- `test_diff_client_ver` test in test_p2p.py seems to be assuming that a _newer_ client ("v2") trying to bootstrap into a network running older clients will fail, which doesn't match the comparison logic
- `test_sync_gradually_add_nodes` in test_sync.py seems to be assuming the exact opposite, that a "v2" client will bootstrap and sync successfully
- When running the test locally, it overrides both the main docker image and the "v2" image and makes them the same, so the test will always fail (see #2163)
- The test also passes the same set of command line args in both docker images, which causes issues if these have changed (as they have with, e.g., the new API)

## Steps To Reproduce
1. Run the p2p system test, `make dockertest-p2p`

## Expected Behavior
I think the expected behavior is that a bootstrap node allows a node with a newer client version to connect to it, but not vice-versa. We should discuss the expected behavior. Also, what about non-bootstrap nodes? What about connections in the opposite direction? This asymmetry seems a bit odd to me.

## To do
- [x] Discuss and decide what to do about `test_diff_client_ver`: we probably want to try connecting an _older_ node rather than a _newer_ one here
- [x] Fix broken logic (#2169)
- [ ] Make it possible to successfully run this test locally, don't override `clientv2` docker image name (#2131)

## Environment
Please complete the following information:
 - OS: all
 - Node Version: bc91d5bf2bb95b742eff53be6281dd63022bd347, 44e4fc3a884a071e4de8b6565f91c35a8e1ed7a2

## Additional Resources 
The faulty comparison logic is here:

https://github.com/spacemeshos/go-spacemesh/blob/bc91d5bf2bb95b742eff53be6281dd63022bd347/p2p/version/version.go#L12-L58

The faulty test logic is here:

https://github.com/spacemeshos/go-spacemesh/blob/bc91d5bf2bb95b742eff53be6281dd63022bd347/tests/p2p/test_p2p.py#L314-L330

PR #2169, opened 2020-10-23
[Merged by Bors] - Fix p2p version comparison test and logic

lrettig:
## Motivation
See #2168 

## Changes
- Fix version comparison logic

## Test Plan
- Added regression test for logic

## TODO
- [x] Discuss and decide what to do about `test_diff_client_ver`: we probably want to try connecting an _older_ node rather than a _newer_ one here

### DIFF
--- a/p2p/version/version.go
+++ b/p2p/version/version.go
@@ -4,6 +4,7 @@ package version
 import (
 	"errors"
 	"fmt"
+	"github.com/spacemeshos/go-spacemesh/log"
 	"strconv"
 	"strings"
 )
@@ -44,7 +45,7 @@ func CheckNodeVersion(reqVersion string, minVersion string) (bool, error) {
 		}
 	}
 
-	for i := 0; i < 2; i++ {
+	for i := 0; i < 3; i++ {
 		if numa[i] > numb[i] {
 			return true, nil
 		}
@@ -54,5 +55,8 @@ func CheckNodeVersion(reqVersion string, minVersion string) (bool, error) {
 		}
 	}
 
-	return false, nil
+	// this should not happen, it means the versions perfectly match but we already tested that above
+	log.With().Error("error comparing node version strings",
+		log.String("reqVersion", reqVersion), log.String("minVersion", minVersion))
+	return false, errors.New("error comparing node version strings")
 }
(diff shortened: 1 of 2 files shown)


lrettig:
> Make it possible to successfully run this test locally, don't override clientv2 docker image name

Addressed in #2131 in 5374ea0ca0f71391c9f6cf10789c1193df93bb0e

> decide what to do about the case where clientv2 image needs a different set of command line flags

Most of the places these are used in the tests, they should be overridden. The only exception I've found is the p2p test. Need to discuss with everyone else how we want this test to work, but I suspect the answer here will be, manually create and push a docker image, set it in the config file, and set `noreplace`.

y0sher:
I don't see a problem with the latter solution.. this runs only a thin p2p node so it shouldn't require a lot of changes are rebuilding, though if we don't do an automatic process it's worth to somehow mention this in the tests or code so if someone changes something she will notice that the dockerimage should be updated.

lrettig:
> I don't see a problem with the latter solution.. this runs only a thin p2p node so it shouldn't require a lot of changes are rebuilding, though if we don't do an automatic process it's worth to somehow mention this in the tests or code so if someone changes something she will notice that the dockerimage should be updated.

Agree on your point about documenting this, I can do that as part of this PR. Regarding the core logic of the [test in question](https://github.com/spacemeshos/go-spacemesh/blob/edb8baedd9fe84b10c114e6b033b9a843044c295/tests/p2p/test_p2p.py#L314), do you agree that it should be testing that a bootstrap server running a _newer_ client version (say, `v0.2.0`) should refuse to allow a client running an _older_ version (say, `v0.1.0`) to connect to it? If so, I think we're on the same page here, and I can sort out the code. Thanks!

lrettig:
Confirmed on R&D call today that this is the goal - in the long term we want a more robust design here (e.g., versioning individual protocols) but in the short term we want nodes running newer versions of the protocol to refuse to bootstrap nodes running older versions. I'll make the necessary changes here and then mark this PR as ready to review and merge.

lrettig:
I reduced the scope of this PR to one simple fix of the logic. I'll handle updating the P2P test in #2131. @y0sher could you please review?

y0sher (approved):
Sorry it took so long, I was sure it is still a draft. Looks perfectly good to me.

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge-

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

Issue #2102, opened 2020-08-06
Some more minor log cleanup


lrettig:
## Description
There are a couple of log statements that still need to be cleaned up, see below

## Affected code
p2p, sync?

This issue appears in commit hash: a1f005549ddf99ceaf24724d93a306930db1460d, 38903f680e0a9597930c3d3071ec8174763176ae

```
2020-08-06T09:01:38.984-0400    INFO    32678.app               Genesis account created: 0x1, Balance: %!s(uint64=100000000000000000)   {"node_id": "326786413a66aa696ce5e84b7a29cfedb3b85acb182db0739c2c6e51423afaa7"}
2020-08-06T08:49:54.814-0400    DEBUG   f6d5b.p2p               removed request %!(EXTRA uint64=349)    {"node_id": "f6d5bc4a29b690508922503d9eab7d9d5ec3a393f183188604a62d25c8ea8946", "P2PID": "86w
2020-08-06T08:49:45.673-0400    WARN    f6d5b.sync              unfamiliar block was requested (id: %s) {"node_id": "f6d5bc4a29b690508922503d9eab7d9d5ec3a393f183188604a62d25c8ea8946", "block_id": "c322b3bea6", "errmsg": "leveldb: not found"}
2020-08-06T09:13:12.678-0400	DEBUG	c15cc.mesh         	add block [169 248 147 93 25 59 143 79 175 6 9 201 3 156 215 185 109 75 130 128]	{"node_id": "c15ccb9782fb6a6e636363072b4cbfabcca779a77b968d6225938dadda2e9af5"}
2020-08-12T16:47:41.192-0400    ERROR   4f33f.blockListener     received invalid block [0 0 0 0 0 0 0 5 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2 0 0 0 97 3 24 234 205 227 237 97 43 12 15 65 225 153 111 33 30 59 47 35 245 113 235 214 161 168 253 134 224 244 181 37 22 118 108 60 74 38 249 ...]%!(EXTRA *xdr.UnmarshalError=xdr:DecodeBool: bool not 0 or 1 - read: '64')  {"node_id": "4f33fc92af8f4039daf1de03220e97323ef21a28b188bc230b70f339e0fd222e"}
[...]

PR #2176, opened 2020-10-28
Add DB index for coinbase -> ATX

### LINKED ISSUE #2064: Add db indices for ATX
See e.g.:

https://github.com/spacemeshos/go-spacemesh/blob/bb63ec70d3d9f7ace9a8c08f755af6d7e0fdafac/api/grpcserver/mesh_service.go#L160-L172

This is required by the API. We don't currently have any way to look activations up by layer, or by coinbase. We should add indices for one or both of these.

shibe-dev:
## Motivation
Partially addresses #2064 (adds index by coinbase, still need to add index by layer)

## Changes
Added keys for atxs, they're indexed by coinbase now

## Test Plan
Added new test


noamnelke (changes requested):
Thanks for your contribution, @howydev!

There's a problem with serializing unbounded lists for storing in the database. As the list grows (e.g. more and more ATXs are added to the same coinbase address) appending ATXs to the list will become increasingly expensive. For each append you'll now need to deserialize a long list of previous values, append the new value in-memory and finally serialize the updated list and store it. This becomes more expensive in terms of processing and memory as the lists grow.

The solution to this is to use prefixed database keys (e.g. the ATX coinbase index key could be `cb_<address>_<layer>`). This allows you to instantly append new values without reading from the database.

When fetching the list, an iterator can be instantiated using `db.Find(prefix)`. This allows you to read the list incrementally instead of having to deserialize it all at once.

I also assume the consumer of these methods will want to have some sort of pagination - which can also be done using the iterator approach.

shibe-dev:
Hey noam, appreciate the review! I'll spend more time thinking about implementation details/design for my contributions in the future

activation/atxdb.go
```diff
@@ -42,6 +42,14 @@ func getAtxBodyKey(atxID types.ATXID) []byte {
 	return []byte(fmt.Sprintf("b_%v", atxID.Bytes()))
 }
 
+func getAtxCoinbaseKey(atx *types.ActivationTx) []byte {
+	return []byte(fmt.Sprintf("cb_%v_%v", atx.Coinbase.Bytes(), atx.PubLayerID.Bytes()))
```
noamnelke: I prefer reusing the code here because it makes the intention clearer.
```suggestion
	return []byte(fmt.Sprintf("%v_%v", getCoinbaseAtxPrefix(atx.Coinbase), atx.PubLayerID.Bytes()))
```

noamnelke (commented):
This change is good, but I don't like to introduce unused methods (essentially zombie code) to the codebase.

I think going another step and actually adjusting the implementation here to use `GetAtxIterByCoinbase` would be worthwhile:
https://github.com/spacemeshos/go-spacemesh/blob/bb63ec70d3d9f7ace9a8c08f755af6d7e0fdafac/api/grpcserver/mesh_service.go#L115-L198

activation/atxdb.go
shibe-dev: Hey Noam, i appreciate the suggestion, it makes a lot of sense. I'll make the changes as suggested and update usage in the API service. Also, i'd love to learn how to better spot ideas like this - is there any resource/do you have any resource to recommend that will help me develop this skillset?

shibe-dev:
hey @noamnelke - just made the suggested changes. I couldn't find a way to use the new GetAtxIterByCoinbase method directly but I assume this is what you meant. Also, I changed the method name for 1 of of the getters, it fits the getter naming convention in that file better

api/grpcserver/mesh_service.go
```diff
@@ -123,22 +123,16 @@ func (s MeshService) getFilteredTransactions(startLayer types.LayerID, addr type
...
+	var atxids []types.ATXID
+	atxIter := s.Mesh.GetAtxIterByCoinbase(addr)
+	for atxIter.Next() {
+		var a types.ATXID
+		err = types.BytesToInterface(atxIter.Value(), &a)
+		if err != nil {
+			return nil, err
 		}
+		atxids = append(atxids, a)
 	}
```
noamnelke: Two things this code is missing are releasing the iterator and checking for errors.

Example code from [here](https://github.com/syndtr/goleveldb):
> Iterate over database content:
> ```golang
>iter := db.NewIterator(nil, nil)
>for iter.Next() {
>	// Remember that the contents of the returned slice should not be modified, and
>	// only valid until the next call to Next.
>	key := iter.Key()
>	value := iter.Value()
>	...
>}
>iter.Release()
>err = iter.Error()
>...
>```

Also see documentation [here](https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.NewIterator) and [here](https://godoc.org/github.com/syndtr/goleveldb/leveldb/iterator#Iterator).

noamnelke (changes requested):
Hey Howey, first - a comment about using git: when you put everything in one commit and then replace it after my comments it's much harder for me to review the changes since my last review. It's best to add additional commits with changes so I can focus on what's new faster. In general, force-pushing should be reserved for fixing the commit structure or fixing a wrong commit right after pushing it for the first time. Whenever someone reviewed a commit, the polite thing to do is keep it and add additional commits, as needed.

Regarding your question about how to learn these things: you're already doing the right thing! The skill you're referring to is called "experience" and you learn it from your own mistakes or from other's past mistakes, through mentoring like we're doing here now. Don't be discouraged by my comments - just keep it up!

shibe-dev:
hey Noam, I apologize, it was not my intention to be impolite, I'll definitely follow that convention from now on. My intention with the rebase/squashes and force pushes was to clean up commit history since it doesnt look nice for an issue fix to have multiple commits associated with it. I assume this process should be done at the end instead after the PR is approved?

noamnelke:
No need to apologize 🙂

>I assume this process should be done at the end instead after the PR is approved?

Yes, we've actually automated this process. Our testing and merging mechanism, bors, is set up to squash and then merge PRs, so each PR only adds a single commit to `develop`.

shibe-dev:
@noamnelke Wow ok, bors auto squashing commits is really handy. I've 2 questions if you don't mind:
1. I'm slightly confused by the phrasing/wording on the documentation on leveldb iterators. To me it sounds like db iterators should be run in separate goroutines if there's a danger of having them run concurrently (which potentially would be possible as they're being used in other methods other than API methods). I've decided against taken the initiative to move all iterators into goroutines because of my confusion on the matter - please advice 
2. This question is about design choices, maybe @lrettig would be better suited for this question - I noticed in this issue that a lot of structs referenced interfaces modeling already-implemented structs instead of the struct itself. For example, txAPI is an interface which is basically the Mesh struct. In this example, is loose coupling specifically because API functionality is considered a "separate-enough module" or does the Spacemesh team foresee lots of changes/potentially overhauling the base Mesh struct?

lrettig:
Re: releasing iterators, I don't see any `Release` calls in our existing code, e.g.:

https://github.com/spacemeshos/go-spacemesh/blob/9763dca6f20b2138f6722c7b40a6dcc909672fa7/activation/atxdb.go#L607-L621

It strikes me that we need to fix this, unless @noamnelke feels otherwise.

> To me it sounds like db iterators should be run in separate goroutines if there's a danger of having them run concurrently (which potentially would be possible as they're being used in other methods other than API methods). I've decided against taken the initiative to move all iterators into goroutines because of my confusion on the matter

I'm pretty sure that none of the API code you're touching writes to the DB, it's all read-only, so I'm not sure this is an issue. Again, I don't see our existing code (same block I linked above) using read locks or separate goroutines for iterators. Will let @noamnelke weigh in on this as well.

> is loose coupling specifically because API functionality is considered a "separate-enough module" or does the Spacemesh team foresee lots of changes/potentially overhauling the base Mesh struct?

Actually I think @noamnelke could give a better answer than me since he's much more familiar with golang but I'll try. I think it's best to think of the struct and the interface as two different (but related) things. This is best practice in golang: interfaces should be as small as possible for a given job, and they are composable. (I come from an OOP world so I still struggle sometimes to separate the idea of data structure and methods.)

Here are a couple of resources I've found helpful on this topic:

- https://medium.com/@meeusdylan/go-interfaces-keep-them-small-f148ae200d6b
- https://dev.to/wagslane/best-practices-for-writing-clean-interfaces-in-go-5c2j

api/grpcserver/mesh_service.go
```diff
@@ -123,22 +123,21 @@ func (s MeshService) getFilteredTransactions(startLayer types.LayerID, addr type
...
-		if layer == nil || err != nil {
-			return nil, status.Errorf(codes.Internal, "error retrieving layer data")
-		}
 
-		for _, b := range layer.Blocks() {
-			if b.ActiveSet != nil {
-				atxids = append(atxids, *b.ActiveSet...)
-			}
+	var atxids []types.ATXID
+	atxIter := s.Mesh.GetAtxIterByCoinbase(addr)
```
lrettig: Iterators are a pretty low-level abstraction and I'd prefer to keep them out of the API code if possible. I think it's better if this code lives inside the Mesh, and you just call a simple function here that returns all of the ATX data (like the existing `Mesh.GetATXs()`) - something like `Mesh.GetATXsByCoinbase()`.

activation/atxdb.go
```diff
@@ -635,6 +648,11 @@ func (db *DB) GetEpochAtxs(epochID types.EpochID) (atxs []types.ATXID) {
 		}
 		atxs = append(atxs, a)
 	}
+	atxIterator.Release()
+	err := atxIterator.Error()
```
lrettig: You can collapse this line and the next line into one line, e.g., `if err := atxIterator.Error(); err != nil {`

api/api_test.go
```diff
@@ -203,10 +204,15 @@ func (t *TxAPIMock) GetLayerStateRoot(layer types.LayerID) (types.Hash32, error)
 func (t *TxAPIMock) GetBalance(addr types.Address) uint64 {
 	panic("implement me")
 }
+
 func (t *TxAPIMock) GetNonce(addr types.Address) uint64 {
 	panic("implement me")
 }
 
+func (t *TxAPIMock) GetAtxIterByCoinbase(coinbase types.Address) database.Iterator {
```
lrettig: As I mentioned elsewhere, let's keep Iter stuff out of the API if possible

api/grpcserver/grpcserver_test.go
```diff
@@ -295,6 +296,10 @@ func (t *TxAPIMock) ProcessedLayer() types.LayerID {
 	return types.LayerID(layerVerified)
 }
 
+func (t *TxAPIMock) GetAtxIterByCoinbase(coinbase types.Address) database.Iterator {
```
lrettig: Let's keep this out of the API code

api/node.go
```diff
@@ -70,6 +71,7 @@ type Syncer interface {
 type TxAPI interface {
 	AddressExists(types.Address) bool
 	ValidateNonceAndBalance(*types.Transaction) error
+	GetAtxIterByCoinbase(coinbase types.Address) database.Iterator
```
lrettig: I don't think this belongs in the TxAPI interface

lrettig: As mentioned elsewhere I think it can be replaced with something like `GetATXsByCoinbase`

database/interface.go
```diff
@@ -65,6 +65,8 @@ type Iterator interface {
 	iterator.IteratorSeeker
 	Key() []byte
 	Value() []byte
+	Release()
```
lrettig: I don't think you need to change the interface at this level, see the existing https://godoc.org/github.com/syndtr/goleveldb/leveldb/iterator#CommonIterator

lrettig: i.e., maybe replace `iterator.IteratorSeeker` with `iterator.CommonIterator`?

lrettig (changes requested):

shibe-dev:
hey @noamnelke and @lrettig - I feel pretty good about the latest revision, looking forward to hearing your comments!

mesh/mesh.go
```diff
@@ -825,3 +826,25 @@ func (msh *Mesh) GetATXs(atxIds []types.ATXID) (map[types.ATXID]*types.Activatio
...
 }
+
+// GetATXsByCoinbase returns a list of all atx ids corresponding to a coinbase address
+func (msh *Mesh) GetATXsByCoinbase(coinbase types.Address) ([]types.ATXID, error) {
+
+	var atxIds []types.ATXID
+	atxIter := msh.GetAtxIterByCoinbase(coinbase)
+	for atxIter.Next() {
+		var a types.ATXID
+		err := types.BytesToInterface(atxIter.Value(), &a)
```
lrettig: You can collapse this line and the next line as well, e.g.,

`if err := ...; err != nil {`

mesh/mesh.go
```diff
@@ -825,3 +826,25 @@ func (msh *Mesh) GetATXs(atxIds []types.ATXID) (map[types.ATXID]*types.Activatio
...
+		if err != nil {
+			return nil, err
+		}
+		atxIds = append(atxIds, a)
+	}
+
+	atxIter.Release()
```
lrettig: Should probably move this up immediately after you get the iterator, using [defer](https://gobyexample.com/defer), e.g.,

`defer atxIter.Release()`

This ensures that it gets run even if there's an error

If you need to do multiple things here, e.g., calling `Release()` and then checking for an error (I'm not totally sure you do need to do that, please double check, but if you do), you can wrap it all in a function, e.g.

```
defer func() {
  atxIter.Release()
  ... do other things ...
}()
```

mesh/mesh.go
```diff
@@ -825,3 +826,25 @@ func (msh *Mesh) GetATXs(atxIds []types.ATXID) (map[types.ATXID]*types.Activatio
...
+	if err := atxIter.Error(); err != nil {
```
lrettig: Out of curiosity what would cause an error here? Just an error releasing?

database/memory_database_iterator.go
```diff
@@ -88,3 +89,9 @@ func (iter *MemDatabaseIterator) Release() { return }
 
 // Error is a stub to comply with DB interface
 func (iter *MemDatabaseIterator) Error() error { return nil }
+
+// SetReleaser is a stub to comply with DB iterator interface
```
lrettig: Why don't we need to actually implement this? Maybe add a comment to this effect

mesh/mesh.go
```diff
@@ -825,3 +826,25 @@ func (msh *Mesh) GetATXs(atxIds []types.ATXID) (map[types.ATXID]*types.Activatio
 	}
 	return atxs, mIds
 }
+
+// GetATXsByCoinbase returns a list of all atx ids corresponding to a coinbase address
+func (msh *Mesh) GetATXsByCoinbase(coinbase types.Address) ([]types.ATXID, error) {
```
lrettig: This returns atxIDs, not ATXs, could you make the method name clearer?

api/grpcserver/mesh_service.go
```diff
@@ -147,11 +135,9 @@ func (s MeshService) getFilteredActivations(startLayer types.LayerID, addr types
 		return nil, status.Errorf(codes.Internal, "error retrieving activations data")
 	}
 	for _, atx := range atxs {
-		// Filter here, now that we have full data
-		if atx.Coinbase == addr {
-			activations = append(activations, atx)
-		}
+		activations = append(activations, atx)
```
lrettig, line 137: This loop and this new array are now redundant, it's just copying the existing `atxs` array

activation/atxdb.go
```diff
@@ -518,6 +526,11 @@ func (db *DB) storeAtxUnlocked(atx *types.ActivationTx) error {
 		return err
 	}
 
+	err = db.atxs.Put(getAtxCoinbaseKey(atx), atx.ID().Bytes())
```
lrettig: Can collapse this and the next line (see other comment for example)

activation/atxdb.go
```diff
@@ -617,6 +630,10 @@ func (db *DB) GetNodeLastAtxID(nodeID types.NodeID) (types.ATXID, error) {
 	if exists := nodeAtxsIterator.Last(); !exists {
 		return *types.EmptyATXID, ErrAtxNotFound(fmt.Errorf("atx for node %v does not exist", nodeID.ShortString()))
 	}
+	nodeAtxsIterator.Release()
```
lrettig: As commented elsewhere, move this into a `defer` statement immediately after getting the iterator

activation/atxdb.go
```diff
@@ -710,6 +731,11 @@ func (db *DB) GetFullAtx(id types.ATXID) (*types.ActivationTx, error) {
 	return atx, nil
 }
 
+// GetAtxIterByCoinbase returns an iterator over all ATXs associated with that coinbase address
+func (db *DB) GetAtxIterByCoinbase(coinbase types.Address) database.Iterator {
+	return db.atxs.Find(getAtxCoinbasePrefix(coinbase))
```
lrettig: Out of curiosity what happens here if the coinbase that's passed in is empty, or invalid, or there's just no matching data? It would be good to add tests for each of these cases

lrettig (changes requested):

mesh/mesh.go
shibe-dev: I had that question too! I couldn't find it in the documentation but my educated guess would be any other iterator errors (inc running multiple iterators concurrently) would cause iter.Next() to be false and then the error surfaces when you do iter.Error. Not sure how I can confirm this theory though

lrettig:
@howydev one quick comment, please hit "Resolve conversation" on the code review comments once you've handled the issue in question - it's easier as a reviewer to quickly scroll through this PR and be able to see what's already been resolved and what's still outstanding. Thanks!

mesh/mesh.go
lrettig: I think the answer is [here](https://godoc.org/github.com/syndtr/goleveldb/leveldb/iterator#CommonIterator):

>     // Error returns any accumulated error. Exhausting all the key/value pairs
>     // is not considered to be an error.
> ...
> When encounter an error any 'seeks method' will return false and will yield no key/value pairs. The error can be queried by calling the Error method. Calling Release is still necessary.

api/grpcserver/mesh_service.go
shibe-dev: atxs variable is a map. Just did a quick google search, I wasn't able to find a more efficient way of extracting all values from a map into a slice/array

lrettig: good point!

shibe-dev:
@lrettig: Lol i only realized there were "resolve conversation" buttons earlier today. Will use those from now on 
1. Regarding errors causing release to not be called - very good catch, ive updated all iter releases in the codebase to use defer logic instead
2. Regarding tests/coinbase addresses - I think there are some overlaps in the cases you mentioned. Since address type is a fixed length array all "invalid address" cases would either be caught at compile or run time or fall into the "no matching keys" case. I've added a test for no-matching-keys

lrettig:
I deleted `api/api_test.go` since it was deleted upstream

activation/atxdb.go
```diff
@@ -513,8 +521,10 @@ func (db *DB) storeAtxUnlocked(atx *types.ActivationTx) error {
 	if err != nil {
 		return err
 	}
-	err = db.atxs.Put(getAtxBodyKey(atx.ID()), atxBodyBytes)
-	if err != nil {
+	if err = db.atxs.Put(getAtxBodyKey(atx.ID()), atxBodyBytes); err != nil {
+		return err
+	}
+	if err = db.atxs.Put(getAtxCoinbaseKey(atx), atx.ID().Bytes()); err != nil {
```
lrettig: There's a series of three `db.atxs.Put` calls here. Are these atomic? If so, where is this guaranteed? What happens if one of these succeeds and one fails?

activation/atxdb.go
```diff
@@ -617,12 +633,20 @@ func (db *DB) GetNodeLastAtxID(nodeID types.NodeID) (types.ATXID, error) {
 	if exists := nodeAtxsIterator.Last(); !exists {
 		return *types.EmptyATXID, ErrAtxNotFound(fmt.Errorf("atx for node %v does not exist", nodeID.ShortString()))
 	}
-	return types.ATXID(types.BytesToHash(nodeAtxsIterator.Value())), nil
+
+	return types.ATXID(types.BytesToHash(nodeAtxsIterator.Value())), err
```
lrettig: This, and the defer func as written, won't work. This line should still `return ..., nil`. You need to turn the `error` return value into a named var (`err error`), then set it inside the defer func as you're doing on line 624. Here's more:

- https://yourbasic.org/golang/defer/#use-func-to-return-a-value
- See #3 in https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1
- https://yourbasic.org/golang/named-return-values-parameters/

activation/atxdb.go
```diff
@@ -617,12 +633,20 @@ func (db *DB) GetNodeLastAtxID(nodeID types.NodeID) (types.ATXID, error) {
...
 }
 
 // GetEpochAtxs returns all valid ATXs received in the epoch epochID
 func (db *DB) GetEpochAtxs(epochID types.EpochID) (atxs []types.ATXID) {
 	atxIterator := db.atxs.Find(getEpochPrefix(epochID))
+	defer func() {
+		atxIterator.Release()
+		if err := atxIterator.Error(); err != nil {
+			db.log.Panic("atxdb iterator error")
```
lrettig: Probably not great to add a panic. Why not do the same as above, and have this method just return an error in this case?

api/grpcserver/grpcserver_test.go
lrettig: I still feel like this doesn't belong here

api/grpcserver/mesh_service.go
```diff
@@ -200,7 +186,7 @@ func (s MeshService) AccountMeshDataQuery(_ context.Context, in *pb.AccountMeshD
 
 	// Gather activation data
 	if filterActivations {
-		atxs, err := s.getFilteredActivations(startLayer, addr)
+		atxs, err := s.getFilteredActivations(addr)
```
lrettig: I think you're ignoring `startLayer` here, which isn't good

mesh/mesh.go
```diff
@@ -825,3 +826,23 @@ func (msh *Mesh) GetATXs(atxIds []types.ATXID) (map[types.ATXID]*types.Activatio
...
 	return atxs, mIds
 }
+
+// GetAtxIDsByCoinbase returns a list of all atx ids corresponding to a coinbase address
+func (msh *Mesh) GetAtxIDsByCoinbase(coinbase types.Address) (atxIds []types.ATXID, err error) {
+
+	atxIter := msh.GetAtxIterByCoinbase(coinbase)
+	defer func() {
+		atxIter.Release()
+		err = atxIter.Error()
```
lrettig: yes! this method looks great.

activation/activationdb_test.go
```diff
@@ -331,6 +331,55 @@ func Test_DBSanity(t *testing.T) {
...
+	err := atxdb.storeAtxUnlocked(atx1)
+	assert.NoError(t, err)
+	err = atxdb.storeAtxUnlocked(atx2)
+	assert.NoError(t, err)
+	err = atxdb.storeAtxUnlocked(atx3)
+	assert.NoError(t, err)
+
+	iter := atxdb.GetAtxIterByCoinbase(coinbase1)
+	assert.True(t, iter.Next())
+	assert.Equal(t, iter.Value(), atxid3.Bytes())
```
lrettig, line 362: Why do they come back in this particular order? Could you add a comment to this effect?

lrettig (changes requested):

activation/atxdb.go
lrettig: you got this perfect [below](https://github.com/spacemeshos/go-spacemesh/pull/2176#discussion_r528021525) :)

activation/atxdb.go
shibe-dev: hey @lrettig - this is the last task I'm working on and changing the return types causes a lot of changes around the codebase. i've to update interfaces/change values in blockoracle.go, node.go and others. Just to check this is planned/intentional?

lrettig: Thanks for checking. I just noticed that there's already a panic statement below in this method, so I think it's not so terrible to add another one. If we feel that this method signature should be changed to return an error, and that these panics should be removed, we can open a new issue to follow up on that.

activation/atxdb.go
shibe-dev: good catch - I've changed them to use a batch write instead. Also included more precise error messages to narrow down which put throws an error when one fails

activation/atxdb.go
```diff
@@ -504,18 +512,27 @@ func (db *DB) storeAtxUnlocked(atx *types.ActivationTx) error {
...
-	err = db.atxs.Put(getAtxHeaderKey(atx.ID()), atxHeaderBytes)
+	atxBodyBytes, err := types.InterfaceToBytes(getAtxBody(atx))
 	if err != nil {
 		return err
 	}
 
-	atxBodyBytes, err := types.InterfaceToBytes(getAtxBody(atx))
-	if err != nil {
-		return err
+	batch := db.atxs.NewBatch()
```
lrettig, line 527: Is there no need to "close" a batch (e.g., using `defer`) if an error occurs? What happens if you return here without calling `batch.Write()`? (I have no idea, I'm just curious.)

api/grpcserver/mesh_service.go
```diff
@@ -147,11 +135,11 @@ func (s MeshService) getFilteredActivations(startLayer types.LayerID, addr types
 		return nil, status.Errorf(codes.Internal, "error retrieving activations data")
 	}
 	for _, atx := range atxs {
-		// Filter here, now that we have full data
-		if atx.Coinbase == addr {
+		if atx.PubLayerID.Uint64() >= startLayer.Uint64() {
```
lrettig: Why is the `.Uint64()` type conversion necessary here?

lrettig (commented):
Added a couple of questions. Looking pretty good. Also:

- please merge develop here and resolve the merge conflicts - you'll want to totally blow away all of the protobuf-related code (`api/pb/*`) as this has been removed upstream.
- this index by coinbase will speed things up a bunch, but of course there is still no index by layer number, so things could still be slow as the number of layers gets large. Could you please make sure there's an open issue to indicate this and track future work on this?

Thanks!

activation/atxdb.go
shibe-dev: The usage mirrors the [documentation specification](https://godoc.org/github.com/syndtr/goleveldb/leveldb#Batch)("batch writes" example) as well as how it's used in the rest of the codebase. There doesn't seem to be a "close" operation implemented for batches from what I can see. From the [implementation details](https://github.com/syndtr/goleveldb/blob/5c35d600f0caac04c20d52438103f1a7aa612598/leveldb/batch.go#L67), batch is just a struct that holds all the operations to be performed with the associated key/value data so if we don't call batch.Write(), garbage collection should clean it up once its out of scope

shibe-dev: also maybe relevant is: batch.Put() modifies data in batch object without touching the DB

api/grpcserver/mesh_service.go
shibe-dev: good catch, it is not necessary

shibe-dev:
hey @lrettig - once again, thanks for reviewing my contribution, I'm really learning a lot through this process. Regarding adding index by layer number, I'll open a separate issue to track it. Once it is implemented, how should it be used in the API/should it just be zombie code for now? 

Regarding the inefficiency, is there a checkable condition to select the more efficient getter depending on the network/state? Something like "if (currentLayer - startLayer) * layerSize > len(GetAtxIDsByCoinbase(coinbase), use GetAtxIDsByCoinbase()". Else it would make sense to scrap GetAtxIDsByCoinbase entirely and just implement GetAtxIDsByLayer

lrettig:
> Regarding adding index by layer number, I'll open a separate issue to track it. Once it is implemented, how should it be used in the API/should it just be zombie code for now?

It shouldn't change the API at all! The API already allows the caller to specify a start/end layer, but we have no way to perform an efficient database query using these, hence the inefficiency. These changes should be purely about efficiency and they should be "behind the scenes".

> Regarding the inefficiency, is there a checkable condition to select the more efficient getter depending on the network/state? Something like "if (currentLayer - startLayer) * layerSize > len(GetAtxIDsByCoinbase(coinbase), use GetAtxIDsByCoinbase()". Else it would make sense to scrap GetAtxIDsByCoinbase entirely and just implement GetAtxIDsByLayer

That's a good question. I can't be sure without better understanding the database layout, and things like how many ATXs there are per layer, per coinbase, and per (layer x coinbase). We might eventually need to index on (layer x coinbase), too. I think we can kick the can down the road and for now just open an issue to point this out, that it needs more R&D in future.

shibe-dev:
I think I worded it badly - by "changes in API" I meant changes in implementation in getFilteredActivations() and mesh_service.go. Thanks for the answer! I feel like its a good first issue for the community, wanted to have a clearer picture/include more details to help the next guy out. That said I'll be happy to jump on it if nobody else wants to take a stab

lrettig:
Hey @howydev, sorry for the delay here. I just took another look at this PR and it looks good. I'd like to get this merged this week. I will give it one more look before pulling the trigger, as it's been a while since I last reviewed it.

Let me know if you'd like to work on the other part of the parent request, adding the layer -> ATX index as well.

activation/atxdb.go
```diff
@@ -504,18 +512,27 @@ func (db *DB) storeAtxUnlocked(atx *types.ActivationTx) error {
...
+			atx.ID(), err)
 	}
-	err = db.atxs.Put(getAtxBodyKey(atx.ID()), atxBodyBytes)
-	if err != nil {
-		return err
+	if err = batch.Put(getAtxBodyKey(atx.ID()), atxBodyBytes); err != nil {
+		return fmt.Errorf("failed to store atx body for atxid %x with error %v",
+			atx.ID(), err)
+	}
+	if err = batch.Put(getAtxCoinbaseKey(atx), atx.ID().Bytes()); err != nil {
```
lrettig: Why do we include layerID in this index as well? `getAtxCoinbaseKey` doesn't appear to be used anywhere; when we read the data, it appears that we're reading only the coinbase prefix but not the full key including layer.

I'm trying to work out whether this solves the issue discussed in https://github.com/spacemeshos/go-spacemesh/pull/2176#issuecomment-736081089 about whether or not we need a separate index by layer. Can we use string prefix search with `db.Find` to find ATXs for a _range_ of layerIDs as well? Hmm.

activation/activationdb_test.go
```diff
@@ -331,6 +331,56 @@ func Test_DBSanity(t *testing.T) {
...
+	atxid3 := atx3.ID()
+
+	err := atxdb.storeAtxUnlocked(atx1)
+	assert.NoError(t, err)
+	err = atxdb.storeAtxUnlocked(atx2)
+	assert.NoError(t, err)
+	err = atxdb.storeAtxUnlocked(atx3)
+	assert.NoError(t, err)
+
+	// DB stores atxs sorted using byte by byte comparison of atxids. Thus 3 > 2001 since 3 > 2
```
lrettig: This does not appear to be the case right now (this test is failing). Why would the database sort things, i.e., not continue to store them in the order they were originally stored?

lrettig:
The main outstanding question/challenge here is how to index on (coinbase, layer) where layer >= startLayer. The simplest way to do this is to pick some number of bytes/chars for the layer number portion of the key, and add a key such as "c_<coinbase_addr>_l_<layerid>", then use `iterator.Seek("c_<coinbase_addr>_l_<startlayerid>")` and rely on the ordinary, lexicographic key sorting in layerdb. We could be slightly more efficient if we store the layerid as a number instead of as a string, but that would require a custom [comparator](https://github.com/google/leveldb/blob/master/doc/index.md#comparators).

activation/activationdb_test.go
noamnelke: LevelDB stores values sorted by key: https://github.com/google/leveldb#features (see 2nd bullet).

noamnelke:
>We could be slightly more efficient if we store the layerid as a number instead of as a string, but that would require a custom comparator.

Actually it's easier than you think, the string doesn't have to be printable or human readable, so you can encode the bytes of the integer as a string, then do the same when searching.

lrettig:
> > We could be slightly more efficient if we store the layerid as a number instead of as a string, but that would require a custom comparator.
> 
> Actually it's easier than you think, the string doesn't have to be printable or human readable, so you can encode the bytes of the integer as a string, then do the same when searching.

I think you and I are saying exactly the same thing here :) Right now I'm storing the layerid as a string, without any special encoding. It would be more efficient, in terms of space if not code, to encode it - but then the default lexicographic sort would break.

activation/atxdb.go
shibe-dev: hey lane! I'm looking at the code again. I included layerID in the key to create unique keys for each stored atx. The getter returns an iterator over all keys matching the prefix and here doesn't seem to be a method to get data matching the suffix or other parts of the key so I can't see how to do it without adding another key

shibe-dev:
hey @lrettig i see what you're trying to do. Getting lexicographical ordering = numerical ordering will allow delegation of the range search to the db.Find which is more elegant. I'll spend today looking into how byte encoding works under the hood

noamnelke:
Getting the binary representation of the variable `number` as the string `str` in a way that it would be correctly sorted is as easy as:
```go
var byteArray []byte
binary.BigEndian.PutUint32(byteArray, number)
str := string(byteArray)
```

lrettig:
> Getting the binary representation of the variable `number` as the string `str` in a way that it would be correctly sorted is as easy as:
> 
> ```go
> var byteArray []byte
> binary.BigEndian.PutUint32(byteArray, number)
> str := string(byteArray)
> ```

The database takes bytearrays (not strings) as keys, so you'd leave out the last line here.

What we're doing right now - both in production, and in the changes in this PR (see https://github.com/spacemeshos/go-spacemesh/pull/2176/files#diff-fff248a50b1152bd06a08461c24bd1d5d03b3f94798e89323e940f94223d65c8) - is:

```
	return []byte(fmt.Sprintf("c_%v_", somevalue.Bytes()))
```

which, as far as I can tell (based on [this](https://golang.org/pkg/fmt/)), prints the bytes as a string in base10 format (similar to what we see output [in our logs](https://github.com/spacemeshos/go-spacemesh/issues/2102)). This is crazy inefficient in terms of storage usage since a one-byte number may occupy three bytes in a key, the length is variable, etc.

@noamnelke am I missing something obvious here? It seems like this is an issue quite separate from this PR that we should discuss.

noamnelke:
@lrettig I don't know if it's obvious, but you ARE missing something... The reason I suggested adding that last line to convert the byte array to a string is to avoid the issue you mention later, namely the automatic conversion to decimal characters, which is both inefficient and causes the wrong sort order. We should construct the key, as a string, as we do now, but use the string that was generated with my method, then the last step, as we do now, is to convert the string to a byte array.

Am _I_ missing something?

lrettig:
@noamnelke my point is that, quite apart from this PR, we're already using this antipattern in production. Here's one such example, there may be others:

https://github.com/spacemeshos/go-spacemesh/blob/35432834b1bcc8aa4110e033c9d4e2e9701bc7d7/activation/atxdb.go#L25-L43

I opened #2268 to follow up on this. We don't need to fix existing, broken keys in this PR but let's not introduce any new ones here.

antonlerner:
closed due to inactivity, too bad, a lot of work was out in here 
Please re open if you continue your work on this

PR #2177, opened 2020-10-28
[Merged by Bors] - Improve CI config

### LINKED ISSUE #2172: Remove hardcoded list of "required statuses" from bors.toml
## Description
See this thread: https://github.com/spacemeshos/go-spacemesh/pull/2165#discussion_r511658541

Take one of the suggested courses of action, or come up with a better way so that we don't have to keep this list of "required statuses" up to date manually in `bors.toml`

## Affected code

This issue appears in commit hash: 4d27c598a63f28b478dd803abb6002f8a93452cb

- bors.toml

CC @noamnelke

lrettig:
## Motivation
Closes #2172 

We want to maintain a single, master list of tests that need to pass before we merge things, rather than having it spread across multiple files

## Changes
- Replaces the list of required status checks in bors.toml with a single, new consolidated test defined in master CI workflow
- Updates the GH actions CI badge in README to always display the status of the `develop` branch (rather than whatever happened to run last)
- Clearly delineate where in the CI workflow we can count on secrets being available and where we cannot; don't attempt to run jobs that rely on secrets (e.g., slack notification) if they're not available
- Only send summary status updates to slack at the end of each stage (rather than one per job)

## Test Plan
N/A

Inspiration:
- https://github.com/pyvisa/pyvisa-py/pull/236/files
- https://github.com/taiki-e/pin-project/blob/3795d2db7abc4e04b3c1ec96e1542c55705ea3fc/.github/workflows/ci.yml

Note: Another way we might do this is to use a prefix when naming the tests, and use a wildcard in the `status` config in bors.toml: see https://bors.tech/documentation/#configuration-borstoml (on second thought, this may not work: https://github.com/bors-ng/bors-ng/issues/946)


noamnelke (approved):
We're still explicitly listing the tests, but I guess it's better to at least explicitly list them closer to where they're defined (rather that having to remember to update bors and figuring out where to do so).

I assume duplicating the job is unavoidable? We can't put the `if` inside the `run` or at least put an `if` on the `step`? (that would allow us to not list the dependencies twice)

lrettig:
> We're still explicitly listing the tests, but I guess it's better to at least explicitly list them closer to where they're defined (rather that having to remember to update bors and figuring out where to do so).

Right. I think this is the only way to achieve the desired workflow, where a bunch of things run in parallel, and we wait for them all to succeed, or for one to fail, before continuing.

> I assume duplicating the job is unavoidable? We can't put the `if` inside the `run` or at least put an `if` on the `step`? (that would allow us to not list the dependencies twice)

Not sure, I'm testing

lrettig:
@noamnelke by the way, regarding the secrets, as discussed they're only accessible when the person who triggered the workflow has write permission to the relevant workflow file - in other words, in PRs opened by external contributors, the secrets are not accessible by design. This should never be an issue in the "push to staging/trying/develop" workflows since these can only be triggered by bors. I've made this explicit in comments and caused the jobs that rely on them (right now, just slack notification) to be skipped if they're not available.

It would be nice to cleanly delineate the workflows that do and do not rely on secrets, i.e., the `pull_request` and `push` workflows. However, in its current format, GH Actions workflows are sort of designed so that all related jobs live in a single workflow, in a single file (and a workflow cannot currently be split across files). It may be possible to have one workflow depend on another (see https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run), this is [not entirely clear](https://github.community/t/how-do-i-specify-job-dependency-running-in-another-workflow/16482/12) and it would be messy. So it would probably require a bunch of code duplication, which I'm trying to avoid here.

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try-

lrettig:
bors try-

lrettig:
bors retry

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

noamnelke (approved):
LGTM!

lrettig:
bors merge

PR #2181, opened 2020-10-29
[Merged by Bors] - Sync causes blocks to be considered invalid fix

antonlerner:
## Motivation
Blocks are invalid because sync ATXs does not end before node is marked synced- causing blocks to be invalid because ATXs are not yet found

## Changes
Re ordered syncing of ATXs do be done before layers are validated. 

## Test Plan
Standart tests

## TODO

### DIFF
--- a/sync/syncer.go
+++ b/sync/syncer.go
@@ -336,11 +336,13 @@ func (s *Syncer) synchronise() {
 	// we have all the data of the prev layers so we can simply validate
 	if s.weaklySynced(curr) {
 		s.handleWeaklySynced()
+		err := s.syncEpochActivations(curr.GetEpoch())
+		if err != nil {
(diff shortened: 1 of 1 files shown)


antonlerner:
bors try

noamnelke:
If a referenced ATX is missing - isn't the node supposed to try to fetch it from neighbors before deeming the block invalid?

I'm not saying the change in this PR shouldn't happen, but it seems to me like this would be an optimization (since what I wrote above may be inefficient). It seems to me that if we fetched missing ATXs from neighbors on-demand we wouldn't experience the sync issues we're seeing and so doing the change in this PR may just cover up the underlying issue.

Am I missing something?

y0sher (approved):

noamnelke:
bors try

noamnelke:
bors merge

Issue #2180, opened 2020-10-29
Compute account balances and rewards


avive:
# Motivation
We'd like to be able to get a cvs file of all accounts' addresses, transactions, balance and rewards on a spacemesh network at a given layer. We currently have a script that can generate this file (see https://github.com/spacemeshos/logs-parser). The issue with this script is that currently go-sm log files do not output a full account addresses, and only 5 hex chars, so we don't know the full account address for accounts in the generated reports. This blocks us from transferring balances for testnet users between networks without users providing the full address they want to be credited which is a manual process.

There are two ways to address this issue and automate this operation:
1. Write a script that extracts all accounts addresses and balances from a full node db. Note that this will enable balance transfer but not insight on number of rewards and rewards amounts as currently the rewards events are not stored in the full node db and just go to the log file when rewards computation runs.
2. Change go-sm logging to output standard address string format (20 hex chars instead of 5). e.g. `0x24f7be4558fbab3883a3f5f5cb95f7629ed61975`

Once one of these solutions is implemented, we can write a script that credits accounts on a new network based on the old network ledger.

PR #2183, opened 2020-10-29
Output canonical address string (40 hex chars) instead of 7 in logs

avive:
## Motivation
Output canonical address string (last 40 bytes of public key, hex formatted) instead of 7 for easier extraction of user addresses from log files and address related debugging. See #2180.

## Changes
Changed log related formatting calls to use a.String() instead of a.Short(). 

## TODO

### DIFF
--- a/api/grpc_server.go
+++ b/api/grpc_server.go
@@ -174,7 +174,7 @@ func (s SpacemeshGrpcService) SubmitTransaction(ctx context.Context, in *pb.Sign
 		return nil, err
 	}
 	log.Info("GRPC SubmitTransaction to address: %s (len: %v), amount: %v gaslimit: %v, fee: %v",
-		tx.Recipient.String(), len(tx.Recipient), tx.Amount, tx.GasLimit, tx.Fee)
+		tx.Recipient.Short(), len(tx.Recipient), tx.Amount, tx.GasLimit, tx.Fee)
 	if err := tx.CalcAndSetOrigin(); err != nil {
 		log.With().Error("failed to calc origin", log.Err(err))
 		return nil, err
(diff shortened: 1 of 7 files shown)


api/grpc_server.go
```diff
@@ -174,15 +174,15 @@ func (s SpacemeshGrpcService) SubmitTransaction(ctx context.Context, in *pb.Sign
 		return nil, err
 	}
 	log.Info("GRPC SubmitTransaction to address: %s (len: %v), amount: %v gaslimit: %v, fee: %v",
-		tx.Recipient.Short(), len(tx.Recipient), tx.Amount, tx.GasLimit, tx.Fee)
+		tx.Recipient.String(), len(tx.Recipient), tx.Amount, tx.GasLimit, tx.Fee)
```
lrettig: You can leave out the changes to this file since it'll go away soon in #2131 (and this would cause a merge conflict)

avive: done

noamnelke:
I don't have a strong objection to the change itself, other than increasing the log size.

However, I strongly disagree with this as a reasonable solution to what's described in the motivation. It should be almost trivial to write a gRPC method that returns a csv- or json-formatted list of balances (full address + balance). It should be just as easy to create a similar list of rewards received, if needed (e.g. coinbase account + total rewards).

Unless you have another reason to merge this change, I think we should implement what I suggested.

avive:
I think we should do both. Log size issue should be dealt on with setting proper log levels on various nodes - not by making log data obscure...

lrettig:
Agree. While I don't think we should print e.g. the full node id everywhere in the logs, in the case of account address, there is a standard format here and I think that should be used in the logs.

noamnelke (approved):
Fair enough, I concede.

avive:
bors merge

avive:
Too many changes to develop including rename of grpc server file name. It is easier to re-do this change from current develop then to try to merge this oldish pr.

PR #2187, opened 2020-11-02
[Merged by Bors] - Remove CI status badge from README

lrettig:
## Motivation
"CI status" for the develop branch is not really well-defined since we don't actually run any CI tasks on the develop branch. Bors runs them on staging, then merges to develop. If we set the status badge to develop, as it is now, it shows "no status." If we set it to staging, it'll show the status of the last bors run, which may have failed (and not been merged), which is misleading.

@noamnelke you were right, I concede :)


noamnelke (approved):
Haha, while it's fun to be right, I must admit that you also convinced me that there's value to the badge.

How about we host our own badge that says "CI: passing" that links to `ci_and_bors.md` where we explain our flow and why the badge always says "passing"?

noamnelke:
Something like this:

[CI: passing](https://github.com/spacemeshos/go-spacemesh/blob/develop/ci.md)

The code for this is:
```markdown
[![CI: passing](https://img.shields.io/badge/CI-passing-success?logo=github&style=flat)](https://github.com/spacemeshos/go-spacemesh/blob/develop/ci.md)
```

noamnelke:
We can maybe link to a section in that doc like this:

## You're lying! Your "CI: passing" badge is static and doesn't reflect the true status!

We make use of bors to ensure that commits that break any of our tests cannot get merged to `develop`. This is why our status is always passing - it's static, but it's not a lie!

antonlerner:
closed due to inactivity

lrettig:
I think we still want this. Will finish now.

lrettig:
bors merge

PR #2188, opened 2020-11-02
[Merged by Bors] - Fix flaky app test

antonlerner:
Fixed closing sync and block listener


sync/syncer.go
```diff
@@ -215,12 +215,13 @@ func (s *Syncer) Close() {
 	close(s.exit)
 	close(s.forceSync)
 	s.peers.Close()
+	s.syncLock.Lock()
+	s.syncLock.Unlock()
```
y0sher, line 220: why not close inside the lock? this could be still be a bit racy

y0sher (approved):

sync/syncer.go
antonlerner: the lock is taken to make sure close has been processed and the loop has exited. after this happens there can be not more requests to sync

antonlerner:
bors merge

antonlerner:
bors merge

PR #2128, opened 2020-08-26
[Merged by Bors] - basic impl of node fetcher

antonlerner:
## Motivation
Added basic fetch mechanism to handle multiple calls for same and different Hashes

- refactored block, ATX, and Tx gossip and sync functions so that the same function will handle both gossip and sync data
- added new layers logic which is to replace the existing sync mechanisms
- refactored fetcher to support batches of requests
- did some code cleanup and better separation of concerns as for sync and block handling

Closes #2108

Issue #1621, opened 2019-12-24
Use the PoET version mentioned in `go.mod` for automation runs


noamnelke:
The current implementation uses the latest `develop`, instead we should extract the PoET version from `go.mod` and use that.

lrettig:
We ran into issues with poet version again today after https://github.com/spacemeshos/poet/pull/105 was merged yesterday (my fault). We put into place a temporary fix in #2128 where we pinned a specific poet build: https://github.com/spacemeshos/go-spacemesh/commit/e353ccc393eda807411ef001015b2d3e8dabf4fb#diff-7002836ab7b8ee5645f5352721e0a15f9dc5eeac359a716460926725f2be4d72

However we need a more permanent solution. Note that this will also rely on changes on the poet side: https://github.com/spacemeshos/poet/issues/106

moshababo:
@dshulyak still relevant?

lrettig:
It's still relevant: there's still a poet version pinned in tests context:

https://github.com/spacemeshos/go-spacemesh/blob/ffc0c927d0d3e96dce55597cba2da8a54bcfc4e6/systest/testcontext/context.go#L32

Which doesn't necessarily match the version pinned in `go.mod`

dshulyak:
it was updated recently. but it may lag behind periodically as those changes are usually independent. maybe we should enforce during review to keep version in go.mod and version in tests in sync. maybe it is not important though as long as they are compatible

PR #2189, opened 2020-11-03
Pin poet version

lrettig:
## Motivation
Poet was recently upgraded to use the new API in https://github.com/spacemeshos/poet/pull/105. This is causing system tests on this side to break since we don't pin the poet server version, but the new API code hasn't been merged here yet. This change pins to an older build of the poet, before the change.

## Changes
- pins poet version throughout system tests

## TODO
- [ ] Fix poet versioning so we don't have to rely on commit hashes
- [ ] Make version dynamic, not hardcoded, perhaps through Makefile/env
- [ ] Specify poet version in one place only, not three

### DIFF
--- a/tests/k8s/bootstrapoet-w-conf-ss.yml
+++ b/tests/k8s/bootstrapoet-w-conf-ss.yml
@@ -38,7 +38,7 @@ spec:
         - name: bootstrap-vol
           mountPath: /usr/share/spacemesh
       - name: poet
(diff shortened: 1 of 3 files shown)


antonlerner (approved):

lrettig:
These changes were merged as part of #2128. Will follow up on outstanding improvements here in #1621.

PR #2196, opened 2020-11-08
update MultiAccountsState to store root hash and not string

### LINKED ISSUE #2194: Store account state root internally as byte array
It's currently stored as a string. It should be stored as a bytearray, and only converted to a string when printed:

https://github.com/spacemeshos/go-spacemesh/blob/8a072181bfe2633d2a90461deb29627c1283706e/common/types/account_state.go#L15

avive:
## Motivation
Fix #2194


avive:
bors try

noamnelke (approved):
Oh, how I hate that superfluous `0x` prefix! But that's outside the scope of this PR...

avive:
> Oh, how I hate that superfluous `0x` prefix! But that's outside the scope of this PR...

It is good in formatted strings because it gives end users some assurance that they don't look at random bytes garbage. In smapp, explore and CLIWallet we have canonical string formats for txs, addresses, etc... which uses them which I think is better than not.. isn't this an emergent standard from both eth and btc projects to have 1 canonical string representation like that with the prefix? 

I double-checked in in few btc explorers there is no 0x but it is the norm in eth explorers and dashes to display them. I think that the 2 chars is a small price to pay in the front-end for better usability. These numbers are less scary for `normal` users when they are prefixed. I see this as the main reason to keep including these in the sm canonical string display format for binary data.

noamnelke:
I think there's a big difference between what we display in the front-end (SMApp) vs. logs.

I don't think the `0x` makes it less scary for users in the front-end. Normies don't know the difference between hexadecimal and random bytes. It's either a number, or it's a number with letters sparkled in. If they even notice the consistent prefix, they have no idea what it means or why it's there. For this reason I see no advantage in adding the prefix in the front end (SMApp) and making it the canonical presentation. It just adds clutter. But I don't care so much - if you have reason to believe that users find it useful for reasons that elude me, then do add it.

In the logs this compromises my ability as a developer to quickly parse addresses. Instead of being able to compare the first few characters of addresses, I now need to look at characters 2-7, which is significantly harder to do quickly. Looking at the last few characters is an option, but sometimes addresses are cut off and in general, it's less convenient. Short 5 character addresses are the best, as the eye can differentiate/match them significantly more easily, but adding the prefix really ruins it. Additionally, in the logs we know what format addresses are, and we don't need visual clues. Even if we didn't - even a novice developer should be able to recognize a hex-string with zero cognitive load.

`</rant>`

avive:
Noam, it does on the ui-level because it is a visual clue that is shared among all data. We agree that internally in go-sm all addresses, key, etc should be binary values and not stored as hex screens and that presentation of hex data in log files is different than presentation on the ui level. Smapp has a canonical format to display 0x before every hex string and we can discuss changing it separately if you feel strongly about it. Unifying to a canonical string address in logs is probably a good idea but also a separate task.

avive:
@noamnelke waiting your reply on my last comment.

noamnelke:
I don't have much to say that I haven't said already.

The presentation in logs is not the same as how it's stored in the DB. We store raw bytes, but print the hex representation. For the reasons I outlined above, prefixing the hex strings (short/long) with anything makes the logs harder to parse at a glance. It's easy to scan the first few chars of each address in a list, it's harder to do that with a prefix.

If you want the UI to appeal to non-developers, prefixing addresses with `0x` is useless. Add a little icon, or an `Address` label. Add a `(?)` with a little explanation + link for further reading. To non-developers `0x` is meaningless. Especially when it's not only applied to addresses, but also to node IDs, network IDs (soon), miner IDs, etc - all of which are hexadecimal.

If you're unconvinced we can have a short call about this.

lrettig:
> If you want the UI to appeal to non-developers, prefixing addresses with `0x` is useless. Add a little icon, or an `Address` label. Add a `(?)` with a little explanation + link for further reading. To non-developers `0x` is meaningless.

For the record, I disagree with this part. Prefixing addresses (of accounts, smart contracts, etc.) expressed as hexadecimal stings with `0x` has been standard practice for a long time and many non-developers will immediately recognize this. It's even become a meme, hence the existence of the "0x" project and dozens of other projects with names that start with "0x". As to whether the prefix belongs in the logs or not, I think it's contextual - if it's clear from the context in a given log line that what's being printed is an address then the prefix is probably not necessary, but we should always use it in user-facing applications like smapp.

noamnelke:
Ethereum's target audience is developers. This is not our target audience. I don't think that what's standard practice on Ethereum is necessarily appropriate for us. You think your average, non-techie, 19 year old Arts major finds `0x` recognizable? I would guess not. I'm also not sure myself what it's supposed to signify, that's it's hex vs. decimal vs. base64 (why is it important to distinguish between them?)? That it's an address (will we only put it on addresses and not miner IDs or transaction IDs?)?

lrettig:
I think it's about standardization. The average, non-technical user doesn't know anything about the underlying protocol, but they know that a string that starts "http://" is a web URL that can be typed or pasted into the address bar of their web browser. The same goes for Ethereum-style crypto wallet addresses and "0x". It could and maybe should be something more specific, but this is sort of the standard we've landed on for now.

noamnelke:
In that case, let's prefix only addresses with `sm:`. This will be more familiar (kind of like `http:\\`) and clear (it's clearly an acronym of Spacemesh) to non-technical users and, more importantly, won't be confused with other (Spacemesh or not) hexadecimal numbers.

Prefixing any other (non-address) hexadecimal number, like miner IDs, transaction IDs, etc. with the same prefix as addresses (be it `0x` or `sm:`) will have an adverse effect on the potential benefit that users would recognize these numbers as something that can be pasted into the "to" field of transactions. Are we at least in agreement on this last point?

avive:
> In that case, let's prefix only addresses with `sm:`. This will be more familiar (kind of like `http:\\`) and clear (it's clearly an acronym of Spacemesh) to non-technical users and, more importantly, won't be confused with other (Spacemesh or not) hexadecimal numbers.
> 
> Prefixing any other (non-address) hexadecimal number, like miner IDs, transaction IDs, etc. with the same prefix as addresses (be it `0x` or `sm:`) will have an adverse effect on the potential benefit that users would recognize these numbers as something that can be pasted into the "to" field of transactions. Are we at least in agreement on this last point?

We should keep 0x for now, it gives user confidents that at least there's no copy and paste error in addresses for the prefix which is an issue IMO for bitcoin addresses. Smapp already displays 0x for all hashes and I don't think we should invest time in changing it now. Also the ledger app already implemented with 0x for display. Regarding logs - this is the dev team preference, however, there's a short-term issue that we rely on log to figure out rewards because the rewards in the api are not fully implemented and there are known issues in the api that we decided it doesn't make sense to invest in fixing because the whole reward scheme is about to change. Considering other tasks I think this is a low priority issue for now. Also - the API data item type for displaying hashes (layer, tx id, account, pub keys, etc...) is already bytes so this is just a front-end display issue...

lrettig:
For the record, other chains use different types of prefixes for different types of address. E.g. this is how cosmos does it using bech32 prefixes: https://docs.cosmos.network/master/basics/accounts.html#addresses-and-pubkeys. Personally I think this improves UX.

Some more reading:

- https://en.bitcoin.it/wiki/List_of_address_prefixes
- https://en.bitcoin.it/wiki/Bech32

noamnelke:
What Cosmos did is great, it's exactly what I meant (perhaps something shorter). I knew about the bitcoin prefixes, but I think they're less useful for non-technical users (you need to be told about them to even notice, imo).

nkryuchkov:
@avive @noamnelke @lrettig 
Should this be merged after conflicts are resolved?

PR #2197, opened 2020-11-09
new CI infra with private ELK per test

AmitShaul:
## Motivation
Make the CI great again
merged -->

## Changes
added a new infra for running CI tests with private elk cluster, after test is finished all tests data is being sent to a central elk cluster.
in this repo I added another folder parallel to "tests", this is only for testing the new code.
in case of success "tests" will be overwritten.
makefile and ci.yml were changed to run both of our CIs , the new and the old, to run in parallel to test the new code without touching the old one.

THIS PR IS FOR TESTING the new code and not the final code!


lrettig:
Hi @AmitShaul, could you briefly summarize here the difference between the current testing infrastructure and this new ELK-based infrastructure? What's the difference/benefit for switching over? What problem(s) is it intended to address?

Regarding the PR, I'm a little confused about why you needed to duplicate the tests files, dockerfiles, etc. If these changes are intended to eventually replace the existing infrastructure, then we should update the existing files in place. If the intention is to run in parallel for a while, then we can just do that using a separate git branch, can't we?

BTW none of the new tests are currently being triggered by the CI here since it only runs the stage1 tests inside a PR. There are a few options if you want to try running the system tests. You could run the system tests locally, you could try [using `act`](https://github.com/spacemeshos/go-spacemesh/blob/develop/ci.md#testing-locally), or you could change the ci.yml workflow file here to trigger all of the tests in this branch. Let me know if you need a hand with this!

noamnelke:
@lrettig the idea was to run the new env in parallel for a while, but make it not interfere with normal tests. The way it's set up now, it should run in parallel, but bors won't depend on it (please comment if you don't think it will do this).

noamnelke:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

noamnelke:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

lrettig:
> @lrettig the idea was to run the new env in parallel for a while, but make it not interfere with normal tests. The way it's set up now, it should run in parallel, but bors won't depend on it (please comment if you don't think it will do this).

My concern here is around code duplication. I can't compare the old and new tests without a diff. Rather than duplicating all of the code, can't we parameterize the difference and move it into env vars or Makefile or a config file or something?

If that's not possible for some reason then my Plan B would be to keep this in a separate branch rather than duplicating the code.

AmitShaul:
bors try

AmitShaul:
@lrettig this is a temporary setup to make both code run in parallel. 
I agree that code duplication is a horrible thing (it was really hard for me to get that going).
once we know that my code work I will open a new PR without the code duplication and reverse this PR.

lrettig:
> I agree that code duplication is a horrible thing (it was really hard for me to get that going).
> once we know that my code work I will open a new PR without the code duplication and reverse this PR.

Okay, gotcha, thanks for clarifying. I know that it's sometimes easier to run these tests using CI, but as a reminder, you can definitely run the tests locally, either directly (e.g., `make dockertest-late-nodes`) or [using act](https://github.com/spacemeshos/go-spacemesh/blob/develop/ci.md#testing-locally).

I'll convert this PR into a draft now to make it clear that it isn't intended to be merged.

lrettig:
Per separate conversation with @AmitShaul I moved the new elk tests into their own "stage" in the CI which runs only after the existing system tests - hopefully this won't overload the system with too many tests running in parallel. Or maybe instead we want to _replace_ the existing tests? If these do not depend on them and this PR is only for testing, that would make sense - it would save resources and time.

AmitShaul:
bors try

noamnelke:
@lrettig eventually, we'll replace the existing tests with the new ones. For now, we want to run them in parallel - to test the new tests.

Since they run on completely different infrastructure, there's no harm in running them in parallel, but there's also no strong incentive to do so - whatever is more convenient.

lrettig:
> Since they run on completely different infrastructure, there's no harm in running them in parallel, but there's also no strong incentive to do so - whatever is more convenient.

Hmm, @AmitShaul suggested that he _doesn't_ want them to all run in parallel since, when they do, tests seem to fail. So now they're running serially, as part of a new "phase" of tests, after the existing tests.

In any case, if the new tests don't actually depend upon the old tests in any way, for purposes of testing it may be faster/easier/cheaper to just _replace_ the old tests with the new ones.

noamnelke:
>In any case, if the new tests don't actually depend upon the old tests in any way, for purposes of testing it may be faster/easier/cheaper to just replace the old tests with the new ones.

Amit already tested the tests and they worked. We wanted to merge this PR so that the tests run in addition to the old tests, so that we gain confidence that they pass when they should and fail when they should in the real world before replacing the old tests for good.

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

noamnelke (approved):

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

.github/workflows/ci.yml
```diff
@@ -209,3 +210,95 @@ jobs:
       - name: Mark the job as failed
         if: env.status != 'success'
         run: exit 1
+
+  #################### STAGE 4 ######################
+  ## stage4: this is for testing the new CI env
+  systemtest-dummy:
+    # only run on push, not on pull_request
+    if: always() && github.event_name == 'push'
```
lrettig, line 218: Remind me why the `always()` is necessary here. Do you want this to run even if stage1 or stage2 fails? If so, the `needs: ci-stage2` would be an issue.

.github/workflows/ci.yml
```diff
@@ -209,3 +210,95 @@ jobs:
...
+    timeout-minutes: 80
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: p2p tests
+        run: make dockertest-p2p-elk
+  systemtest-hare-elk:
+    # only run on push, not on pull_request
+    if: always() && github.event_name == 'push'
+    needs: systemtest-blocks-add-node-elk
```
lrettig, line 285: Is there a reason it depends on this specific prior test? Or is this arbitrary?

.github/workflows/ci.yml
```diff
@@ -209,3 +210,95 @@ jobs:
...
+    timeout-minutes: 80
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: hare system tests
+        run: make dockertest-hare-elk
+  systemtest-sync-elk:
+    # only run on push, not on pull_request
+    if: always() && github.event_name == 'push'
+    needs: systemtest-blocks-add-node-elk
```
lrettig, line 296: Same question here as previous question.

lrettig (commented):
See questions!

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

PR #2198, opened 2020-11-10
[Merged by Bors] - improve logging

noamnelke:
## Motivation
Richer and more consistent logs would make the current testnet investigation easier.

## Changes
Improve logging.

## Test Plan
No need for new tests.

### DIFF
--- a/common/types/block.go
+++ b/common/types/block.go
@@ -37,7 +37,7 @@ func (id BlockID) AsHash32() Hash32 {
 	return Hash20(id).ToHash32()
 }
 
-var layersPerEpoch int32 = 1
+var layersPerEpoch int32
 
 // EffectiveGenesis marks when actual blocks would start being crated in the network, this will take account the first
 // genesis epoch and the following epoch in which ATXs are published
@@ -50,7 +50,7 @@ func getLayersPerEpoch() int32 {
 // SetLayersPerEpoch sets global parameter of layers per epoch, all conversion from layer to epoch use this param
 func SetLayersPerEpoch(l int32) {
 	atomic.StoreInt32(&layersPerEpoch, l)
-	atomic.StoreInt32(&EffectiveGenesis, getLayersPerEpoch()*2-1)
+	atomic.StoreInt32(&EffectiveGenesis, l*2-1)
 }
 
 // LayerID is a uint64 representing a layer number. It is zero-based.
--- a/common/types/block.go
+++ b/common/types/block.go
@@ -48,9 +48,9 @@ func getLayersPerEpoch() int32 {
 }
 
 // SetLayersPerEpoch sets global parameter of layers per epoch, all conversion from layer to epoch use this param
-func SetLayersPerEpoch(l int32) {
-	atomic.StoreInt32(&layersPerEpoch, l)
-	atomic.StoreInt32(&EffectiveGenesis, l*2-1)
+func SetLayersPerEpoch(layers int32) {
+	atomic.StoreInt32(&layersPerEpoch, layers)
+	atomic.StoreInt32(&EffectiveGenesis, layers*2-1)
 }
 
 // LayerID is a uint64 representing a layer number. It is zero-based.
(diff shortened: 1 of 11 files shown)


noamnelke:
bors try

common/types/block.go
```diff
@@ -37,7 +37,7 @@ func (id BlockID) AsHash32() Hash32 {
 	return Hash20(id).ToHash32()
 }
 
-var layersPerEpoch int32
+var layersPerEpoch int32 = 1
```
antonlerner: why do you set a value here? it is better have a test panic than initialising to an arbitrary number and causing unexpected results in tests

antonlerner (approved):

common/types/block.go
noamnelke: Some tests don't care about this, but since I print the epoch number in the log in more places they now panic.

I first tried to just set this to something in the test, but this interfered with other tests and I was too lazy to figure out what's going on. If you strongly believe this is a bad idea I can invest more time in it - I just didn't feel it's worth it.

lrettig: My gut feeling here is, if we have a test that asserts that this is 1 by default, and that it gets set to something else where we expect it to change, then it's fine, but I defer to you two

sync/validation_queue.go
```diff
@@ -86,18 +86,18 @@ func (vq *blockQueue) handleBlocks(bjb fetchJob) {
 }
 
 func (vq *blockQueue) handleBlock(id types.Hash32, block *types.Block) {
-	vq.With().Info("start handling", block.ID(), block.MinerID())
+	vq.With().Info("start handling block", block.Fields()...)
 	if err := vq.fetchBlockDataForValidation(block); err != nil {
-		vq.Error("block fetching data failed", block.ID(), log.Err(err))
+		vq.With().Error("block data fetching failed", append(block.Fields(), log.Err(err))...)
```
lrettig:
```suggestion
		vq.With().Error("fetching block data failed", append(block.Fields(), log.Err(err))...)
```

lrettig (approved):
LGTM, and I'm curious to see a diff of the new, improved logs compared to the old ones :)

common/types/block.go
noamnelke: Figured it out: `TestForEachInView_InMem_WithLimitedLayer` was passing only because `EffectiveGenesis` was zero. Once I called `SetLayersPerEpoch` it got set to some other value, failing the test.

I updated the expected value of the test so it now passes with `LayersPerEpoch=3`.

sync/validation_queue.go
noamnelke: funny, we've now gone through almost all possible positions for the word `fetching`... All accept `block data failed fetching`

common/types/block.go
```diff
@@ -50,7 +50,7 @@ func getLayersPerEpoch() int32 {
 // SetLayersPerEpoch sets global parameter of layers per epoch, all conversion from layer to epoch use this param
 func SetLayersPerEpoch(l int32) {
 	atomic.StoreInt32(&layersPerEpoch, l)
-	atomic.StoreInt32(&EffectiveGenesis, getLayersPerEpoch()*2-1)
+	atomic.StoreInt32(&EffectiveGenesis, l*2-1)
```
lrettig: I can't visually distinguish between "el times two minus one" or "one times two minus one" here - can we rename the `l` variable?

sync/validation_queue.go
lrettig: how about "fetching the fetching fetchers"? 😬

lrettig: totally unrelated but funny: https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo

noamnelke:
bors merge

noamnelke:
bors merge-

noamnelke:
bors merge

sync/validation_queue.go
noamnelke: My English teacher once challenged us to punctuate [this one](https://en.wikipedia.org/wiki/James_while_John_had_had_had_had_had_had_had_had_had_had_had_a_better_effect_on_the_teacher) (I remember it with Tom and Bob, though).

PR #2212, opened 2020-11-25
[Merged by Bors] - Implement Golden ATX

### LINKED ISSUE #2174: Implement Golden ATX
## Description
All ATXs generated in first epoch must point to the golden ATX as the ref and pos ATX. 

A golden ATX should be a configurable hash that will be inserted into the DB when launching a network. this will help identify all ATXs that are part of the same network since all of them eventually must point to same golden ATX. 

The golden ATX will be changes between networks to avoid data leaks from one network to another

nkryuchkov:
## Motivation

Closes #2174 

## Changes

- Implement Golden ATX
- Adjust unit tests for Golden ATX

## Test Plan

- [x] Unit tests

- [x] System tests

## TODO

- [X] Implement Golden ATX for PrevATX field
- [X] Implement Golden ATX for PositioningATX field
- [x] Adjust unit tests for Golden ATX
- [x] Ensure that system tests pass. If not, adjust them for Golden ATX


activation/activation.go
```diff
@@ -113,11 +116,12 @@ type syncer interface {
 }
 
 // NewBuilder returns an atx builder that will start a routine that will attempt to create an atx upon each new layer.
-func NewBuilder(nodeID types.NodeID, coinbaseAccount types.Address, signer signer, db atxDBProvider, net broadcaster, mesh meshProvider, layersPerEpoch uint16, nipstBuilder nipstBuilder, postProver PostProverClient, layerClock layerClock, syncer syncer, store bytesStore, log log.Log) *Builder {
+func NewBuilder(nodeID types.NodeID, coinbaseAccount types.Address, goldenATXID types.ATXID, signer signer, db atxDBProvider, net broadcaster, mesh meshProvider, layersPerEpoch uint16, nipstBuilder nipstBuilder, postProver PostProverClient, layerClock layerClock, syncer syncer, store bytesStore, log log.Log) *Builder {
```
antonlerner: perhaps this is a good time to consolidate `coinbaseAccount` `goldenAtx` and `layersPerEpoch` into a Config struct

activation/atxdb.go
```diff
@@ -435,27 +438,34 @@ func (db *DB) SyntacticallyValidateAtx(atx *types.ActivationTx) error {
...
+
+		// we found no ATXs for this node ID, and it reported golden ATX as prevATX -- this is valid
+		return nil
+	} else if err != nil {
+		db.log.With().Error("could not fetch node last ATX", atx.ID(),
+			log.FieldNamed("atx_node_id", atx.NodeID), log.Err(err))
+		return fmt.Errorf("could not fetch node last ATX: %v", err)
+	}
+
+	if atx.PrevATXID == db.goldenATXID {
```
antonlerner: this check is redundant, since the next check covers this case, if we've found an ATX in our DB this is the ATX that this node should point to

activation/atxdb.go
```diff
@@ -435,27 +438,34 @@ func (db *DB) SyntacticallyValidateAtx(atx *types.ActivationTx) error {
...
-		if err == nil { // we found an ATX for this node ID, although it reported no prevATX -- this is invalid
-			return fmt.Errorf("no prevATX reported, but other ATX with same nodeID (%v) found: %v",
-				atx.NodeID.ShortString(), lastAtx.ShortString())
+	if atx.PrevATXID == *types.EmptyATXID {
+		return fmt.Errorf("no prevATX reported")
+	}
+
+	lastAtx, err := db.GetNodeLastAtxID(atx.NodeID)
+	if _, ok := err.(ErrAtxNotFound); ok {
+		if atx.PrevATXID != db.goldenATXID {
```
antonlerner: add test for this case

activation/activation.go
nkryuchkov: @antonlerner Good idea. What do you think about passing all these args in just one `Config` struct?

activation/atxdb.go
nkryuchkov: Yes, it's redundant if we don't need a separate error for this case. Removed.

activation/atxdb.go
nkryuchkov: Added

activation/atxdb.go
```diff
@@ -346,7 +349,16 @@ func (db *DB) SyntacticallyValidateAtx(atx *types.ActivationTx) error {
 	if atx.NodeID.Key != pub.String() {
 		return fmt.Errorf("node ids don't match")
 	}
-	if atx.PrevATXID != *types.EmptyATXID {
+
```
antonlerner, line 353: update the comment

activation/atxdb.go
```diff
@@ -785,15 +797,15 @@ func (db *DB) HandleAtxData(data []byte, syncer service.Fetcher) error {
 
 // FetchAtxReferences fetches positioning and prev atxs from peers if they are not found in db
 func (db *DB) FetchAtxReferences(atx *types.ActivationTx, f service.Fetcher) error {
-	if atx.PositioningATX != *types.EmptyATXID {
+	if atx.PositioningATX != *types.EmptyATXID && atx.PositioningATX != db.goldenATXID {
```
antonlerner, line 797: do you need the first condition now? should there be a case where an ATX is empty?

activation/atxdb.go
```diff
@@ -785,15 +797,15 @@ func (db *DB) HandleAtxData(data []byte, syncer service.Fetcher) error {
...
 		db.log.Info("going to fetch pos atx %v of atx %v", atx.PositioningATX.ShortString(), atx.ID().ShortString())
 		err := f.FetchAtx(atx.PositioningATX)
 		if err != nil {
 			return err
 		}
 	}
 
-	if atx.PrevATXID != *types.EmptyATXID {
+	if atx.PrevATXID != *types.EmptyATXID && atx.PrevATXID != db.goldenATXID {
```
antonlerner: same here

activation/atxdb.go
```diff
@@ -435,7 +447,7 @@ func (db *DB) SyntacticallyValidateAtx(atx *types.ActivationTx) error {
 // ContextuallyValidateAtx ensures that the previous ATX referenced is the last known ATX for the referenced miner ID.
 // If a previous ATX is not referenced, it validates that indeed there's no previous known ATX for that miner ID.
 func (db *DB) ContextuallyValidateAtx(atx *types.ActivationTxHeader) error {
-	if atx.PrevATXID != *types.EmptyATXID {
+	if atx.PrevATXID != db.goldenATXID {
```
antonlerner: a suggestion - if you change `GetNodeLastAtxID` to return GoldenATX if no ATX wAs found this flow can be shorter... it might be not an easy switch since this is used elsewhere... so just take a look and think wether it's worth it or not

activation/activation.go
```diff
@@ -320,13 +329,13 @@ func (b *Builder) MiningStats() (int, uint64, string, string) {
 // the rewards for blocks made by this miner will go to this address
 func (b *Builder) SetCoinbaseAccount(rewardAddress types.Address) {
 	b.accountLock.Lock()
-	b.coinbaseAccount = rewardAddress
+	b.conf.CoinbaseAccount = rewardAddress
```
antonlerner: why not have a variable inside Builder for this? it looks a bit strange to change config from whithin, unless there is some logic I'm missing here

antonlerner (approved):

activation/activation.go
nkryuchkov: I agree. It would be better to unwrap config to variables inside `Builder`.

activation/atxdb.go
nkryuchkov: Just an extra check. However, it's actually redundant. Removed.

activation/atxdb.go
nkryuchkov: I think `ContextuallyValidateAtx` is a bit complex to understand, so I think it's worth it. I made the suggested changes in the last commit. If we don't need them now, I can create a new branch and remove the last commit from this PR.

nkryuchkov:
bors merge

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

activation/activation.go
```diff
@@ -219,6 +231,7 @@ func (b *Builder) buildNipstChallenge(currentLayer types.LayerID) error {
 	}
 	if prevAtx, err := b.GetPrevAtx(b.nodeID); err != nil {
 		challenge.CommitmentMerkleRoot = b.commitment.MerkleRoot
+		challenge.PrevATXID = b.goldenATXID
```
noamnelke: I think this is wrong. The previous ATX ID shouldn't be set for a node's first ATX.

activation/activation.go
```diff
@@ -411,7 +424,7 @@ func (b *Builder) PublishActivationTx() error {
 
 	var activeSetSize uint32
 	var commitment *types.PostProof
-	if b.challenge.PrevATXID == *types.EmptyATXID {
+	if b.challenge.PrevATXID == b.goldenATXID {
```
noamnelke: This should be undone, along with the change that sets the golden ATX as the previous ATX of a miner's first ATX.

activation/activation_test.go
```diff
@@ -792,9 +838,9 @@ func TestActivationDb_CalcActiveSetFromViewHighConcurrency(t *testing.T) {
 	coinbase2 := types.HexToAddress("bbbb")
 	coinbase3 := types.HexToAddress("cccc")
 	atxs := []*types.ActivationTx{
-		newActivationTx(id1, 0, *types.EmptyATXID, 12, 0, *types.EmptyATXID, coinbase1, 0, []types.BlockID{}, &types.NIPST{}),
-		newActivationTx(id2, 0, *types.EmptyATXID, 300, 0, *types.EmptyATXID, coinbase2, 0, []types.BlockID{}, &types.NIPST{}),
-		newActivationTx(id3, 0, *types.EmptyATXID, 435, 0, *types.EmptyATXID, coinbase3, 0, []types.BlockID{}, &types.NIPST{}),
+		newActivationTx(id1, 0, goldenATXID, 12, 0, goldenATXID, coinbase1, 0, []types.BlockID{}, &types.NIPST{}),
+		newActivationTx(id2, 0, goldenATXID, 300, 0, goldenATXID, coinbase2, 0, []types.BlockID{}, &types.NIPST{}),
+		newActivationTx(id3, 0, goldenATXID, 435, 0, goldenATXID, coinbase3, 0, []types.BlockID{}, &types.NIPST{}),
```
noamnelke: The change to the previous ATX here should also be undone.

activation/atxdb.go
```diff
@@ -328,10 +331,11 @@ func (db *DB) deleteLock(viewHash types.Hash12) {
 
 // SyntacticallyValidateAtx ensures the following conditions apply, otherwise it returns an error.
 //
+// - If the PrevATX and PositioningATX are not empty ATX.
 // - If the sequence number is non-zero: PrevATX points to a syntactically valid ATX whose sequence number is one less
 //   than the current ATX's sequence number.
-// - If the sequence number is zero: PrevATX is empty.
-// - Positioning ATX points to a syntactically valid ATX.
+// - If the sequence number is zero: PrevATX equals to the Golden ATX.
```
noamnelke: This should be undone.

activation/atxdb.go
```diff
@@ -328,10 +331,11 @@ func (db *DB) deleteLock(viewHash types.Hash12) {
...
+// - Positioning ATX points to a syntactically valid ATX or the Golden ATX.
```
noamnelke:
```suggestion
// - Positioning ATX points to a syntactically valid ATX or the Golden ATX for ATXs in epoch 1.
```

activation/atxdb.go
```diff
@@ -346,7 +350,16 @@ func (db *DB) SyntacticallyValidateAtx(atx *types.ActivationTx) error {
 	if atx.NodeID.Key != pub.String() {
 		return fmt.Errorf("node ids don't match")
 	}
-	if atx.PrevATXID != *types.EmptyATXID {
+
+	if atx.PrevATXID == *types.EmptyATXID {
+		return fmt.Errorf("empty previous ATX")
+	}
+
```
noamnelke: This should be undone.

noamnelke: ...along with most changes in this method.

blocks/blockeligibilityvalidator.go
```diff
@@ -102,10 +94,6 @@ func (v BlockEligibilityValidator) BlockSignedAndEligible(block *types.Block) (b
 	message := serializeVRFMessage(epochBeacon, epochNumber, counter)
 	vrfSig := block.EligibilityProof.Sig
 
-	if genesisNoAtx {
-		v.log.Info("skipping VRF validation, genesis block with no ATX") // TODO
-		return true, nil
-	}
```
noamnelke, line 108: 👏🏻 Good job on cleaning up the tech-debt in this file!!

blocks/blockoracle.go
```diff
@@ -65,14 +67,11 @@ func (bo *Oracle) BlockEligible(layerID types.LayerID) (types.ATXID, []types.Blo
 	if !bo.isSynced() {
 		return types.ATXID{}, nil, nil, fmt.Errorf("cannot calc eligibility, not synced yet")
 	}
+
 	epochNumber := layerID.GetEpoch()
 	bo.log.Info("asked for eligibility for epoch %d (cached: %d)", epochNumber, bo.proofsEpoch)
-	if epochNumber.IsGenesis() {
-		bo.log.Warning("asked for eligibility for epoch 0, cannot create blocks here")
-		return *types.EmptyATXID, nil, nil, nil
-	}
```
noamnelke: Why remove this check?

blocks/blockoracle_test.go
```diff
@@ -116,26 +118,6 @@ func testBlockOracleAndValidator(r *require.Assertions, activeSetSize uint32, co
...
-	blockOracle := NewMinerBlockOracle(committeeSize, activeSetSize, layersPerEpoch, activationDB, beaconProvider, vrfsgn, nodeID, func() bool { return true }, lg.WithName("blockOracle"))
-	for layer := uint16(0); layer < layersPerEpoch; layer++ {
-		activationDB.atxPublicationLayer = types.LayerID((layer/layersPerEpoch)*layersPerEpoch - 1)
-		layerID := types.LayerID(layer)
-		atxID, _, _, err := blockOracle.BlockEligible(layerID)
-		r.NoError(err)
-		r.Equal(*types.EmptyATXID, atxID)
-	}
-}
-
```
noamnelke: Why remove this functionality?

cmd/root.go
```diff
@@ -32,7 +33,7 @@ func AddCommands(cmd *cobra.Command) {
 	cmd.PersistentFlags().IntVar(&config.OracleServerWorldID, "oracle_server_worldid",
 		config.OracleServerWorldID, "The worldid to use with the oracle server (temporary) ")
 	cmd.PersistentFlags().StringVar(&config.PoETServer, "poet-server",
-		config.OracleServer, "The poet server url. (temporary) ")
+		config.PoETServer, "The poet server url. (temporary) ")
```
noamnelke, line 36: 👏🏻

cmd/sync/sync.go
```diff
@@ -109,13 +113,14 @@ func (app *syncApp) start(cmd *cobra.Command, args []string) {
...
 		LayersPerEpoch:  uint16(app.Config.LayersPerEpoch),
+		GoldenATXID:     goldenATXID,
 	}
 	types.SetLayersPerEpoch(int32(app.Config.LayersPerEpoch))
 	lg.Info("local db path: %v layers per epoch: %v", path, app.Config.LayersPerEpoch)
 
 	if remote {
 		if err := getData(app.Config.DataDir(), version, lg); err != nil {
-			lg.Error("could not download data for test", err)
+			lg.Error("could not download data for test: %v", err)
```
noamnelke: As long as we're touching it, we might as well do it like it should be:
```suggestion
			lg.With().Error("could not download data for test", log.Err(err))
```

sync/syncer.go
```diff
@@ -837,15 +837,15 @@ func (s *Syncer) FetchAtx(ID types.ATXID) error {
...
+	if atx.PositioningATX != s.GoldenATXID {
 		s.Log.Info("going to fetch pos atx %v of atx %v", atx.PositioningATX.ShortString(), atx.ID().ShortString())
 		_, err := s.fetchAtx(atx.PositioningATX)
 		if err != nil {
 			return err
 		}
 	}
 
-	if atx.PrevATXID != *types.EmptyATXID {
+	if atx.PrevATXID != s.GoldenATXID {
```
noamnelke: should be undone

noamnelke (changes requested):
There should be a check that the Golden ATX is only used for ATXs published in epoch 1. It's invalid to use it in any other epoch.

In general, I'm missing new tests for this new functionality.

nkryuchkov:
bors try

activation/atxdb.go
```diff
@@ -398,7 +407,7 @@ func (db *DB) SyntacticallyValidateAtx(atx *types.ActivationTx) error {
 		}
 	}
 
-	if atx.PositioningATX != *types.EmptyATXID {
+	if atx.PositioningATX != db.goldenATXID {
```
noamnelke, line 410: This `if` should be updated.

An ATX with the golden ATX as the positioning ATX must have publication epoch 1 (not 0 or any other epoch - it currently uses `IsGenesis()` which is wrong).

The error message is also wrong: `no positioning atx found` should be something like `Golden ATX used for ATX in epoch X, but is only valid in epoch 0`.

noamnelke (changes requested):

blocks/blockoracle.go
nkryuchkov: As eligibility proofs cannot be calculated in the genesis epochs, this check seemed redundant to me. However, having it seems more consistent now, so I'm going to revert this.

blocks/blockoracle_test.go
nkryuchkov: Because of https://github.com/spacemeshos/go-spacemesh/pull/2212#discussion_r554573621, empty ATX is not returned in genesis anymore, but as I decided to revert the check removal, I'm going to revert this change as well

nkryuchkov:
bors try

nkryuchkov:
bors try-

nkryuchkov:
bors try

nkryuchkov:
bors try-

nkryuchkov:
bors try

nkryuchkov:
bors try

sync/syncer.go
```diff
@@ -736,7 +736,7 @@ func (s *Syncer) fetchRefBlock(block *types.Block) error {
 func (s *Syncer) fetchAllReferencedAtxs(blk *types.Block) error {
 	var atxs []types.ATXID
 
-	if blk.ATXID != s.GoldenATXID {
+	if blk.ATXID != *types.EmptyATXID {
```
noamnelke: This made me think: A block whose associated ATX is the `GoldenATXID` or the `EmptyATXID` - either of these - is syntactically invalid. We should make sure this check is performed for block syntactic validity and if it is - we can cancel this check and assume that it's never either of these values at this point.

blocks/blocks_test.go
```diff
@@ -236,6 +236,11 @@ func TestBlockHandler_BlockSyntacticValidation_syncRefBlock(t *testing.T) {
 	err = s.blockSyntacticValidation(b, fetch)
 	r.NoError(err)
 	assert.Equal(t, 2, fetch.getBlockCalled[block1ID])
+
+	b.ATXID = goldenATXID
+	err = s.blockSyntacticValidation(b, fetch)
+	r.NoError(err)
+	assert.Equal(t, 3, fetch.getBlockCalled[block1ID]) // incremented by 1 as golden ATX is skipped
```
noamnelke: This is wrong - a block can never have the `GoldenATXID` as the associated ATX. It doesn't ever make anyone eligible to produce blocks. A block like this should be considered syntactically invalid.

blocks/blockoracle.go
```diff
@@ -70,6 +70,10 @@ func (bo *Oracle) BlockEligible(layerID types.LayerID) (types.ATXID, []types.Blo
 
 	epochNumber := layerID.GetEpoch()
 	bo.log.Info("asked for eligibility for epoch %d (cached: %d)", epochNumber, bo.proofsEpoch)
+	if epochNumber.IsGenesis() {
+		bo.log.Warning("asked for eligibility for epoch 0, cannot create blocks here")
```
noamnelke: Why is this a warning?

I see two options: either this is not supposed to happen (we don't check for eligibility in genesis) and then this check is not needed OR this happens and we handle the empty ATX ID and then this shouldn't be a warning, but perhaps only printed as an FYI (`info` instead of `warning`).

activation/atxdb.go
```diff
@@ -422,8 +422,9 @@ func (db *DB) SyntacticallyValidateAtx(atx *types.ActivationTx) error {
 		}
 	} else {
 		publicationEpoch := atx.PubLayerID.GetEpoch()
-		if !publicationEpoch.IsGenesis() {
-			return fmt.Errorf("no positioning atx found")
+		fmt.Printf("epoch: %v\n", publicationEpoch)
```
noamnelke: I assume this was for local debugging and should be removed. Otherwise, replace it with a proper log print.

noamnelke (changes requested):
Although I had a couple of comments, this is going in the right direction.

I still feel that more unit tests are needed. One example: I don't think we're testing the situation where an ATX references the golden ATX to assert that we're not trying to sync it (i.e. not adding it to the sync queue).

activation/atxdb.go
nkryuchkov: Yes, this was for local debugging and I forgot to remove it

nkryuchkov:
bors try

layerfetcher/blocks.go
```diff
@@ -47,9 +47,7 @@ func (l *Logic) fetchRefBlock(block *types.Block) error {
 func (l *Logic) fetchAllReferencedAtxs(blk *types.Block) error {
 	var atxs []types.ATXID
 
-	if blk.ATXID != l.goldenATXID {
-		atxs = append(atxs, blk.ATXID)
-	}
+	atxs = append(atxs, blk.ATXID)
```
noamnelke: You should do the same as what you did here:

https://github.com/spacemeshos/go-spacemesh/blob/4206af445035c7afcb03eee5bf4691835344248d/blocks/blocks.go#L191-L192

noamnelke (approved):

layerfetcher/blocks.go
nkryuchkov: Thanks, done

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #2219, opened 2020-12-02
replace json dump comparison with specific, type-safe comparison in transaction processing tests

### LINKED ISSUE #2191: Improve transaction processing tests
## Motivation

Several transaction processing tests in `processor_test.go` are implemented by comparing strings of formatted json output of the db dump with an expected json formatted string. This is fragile and should be avoided. A better approach is instead of using processor.Dump() to get a json formatted string and compare it with a hard-coded string is to use processor.RawDump() to get a typed go data collection and compare its members in a type-safe way with expected values.

shibe-dev:
## Motivation
fixes #2191 

## Changes
for tests using ApplyTransaction, changed test from checking/comparing json dumps to discrete assert.Equals checking all relevant values

### DIFF
--- a/state/processor_test.go
+++ b/state/processor_test.go
@@ -88,31 +88,14 @@ func (s *ProcessorStateSuite) TestTransactionProcessor_ApplyTransaction() {
 	assert.NoError(s.T(), err)
 	assert.True(s.T(), failed == 0)
 
-	got := string(s.processor.Dump())
-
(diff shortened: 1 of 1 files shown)


shibe-dev:
@avive @noamnelke - tagging for visibility + PR review request (I don't think I've access to add a reviewer to a PR)

antonlerner:
good refactor - too bad it's not merged and inactive - closing it

PR #2220, opened 2020-12-03
[Merged by Bors] - Profiler CLI Option

web3rover:
## Motivation

Instead of separate build for managed nodes we will configure profiler in the same home smeshers build.

This will require update to spacecraft to use CLI option instead of separate docker tag.

## Changes

Added `--profiler` flag

## Test Plan

I will update spacecraft and create a new network for testing


cmd/root.go
```diff
@@ -58,6 +58,8 @@ func AddCommands(cmd *cobra.Command) {
 		config.BlockCacheSize, "size in layers of meshdb block cache")
 	cmd.PersistentFlags().StringVar(&config.PublishEventsURL, "events-url",
 		config.PublishEventsURL, "publish events to this url; if no url specified no events will be published")
+	cmd.PersistentFlags().BoolVar(&config.Profiler, "profiler",
+		config.Profiler, "enable GCP profiler")
```
noamnelke: pprof isn't actually a GCP specific profiler, although GCP supports it. It's a general purpose Go profiler that can be used locally, as well. I'd remove the reference to GCP from the comment here.

cmd/multi_node_sim/config.toml
```diff
@@ -15,6 +15,7 @@ genesis-time = "2019-02-13T17:02:00+00:00"
 layer-duration-sec = "5"
 hdist = "5"
 coinbase = "0x1234"
+profiler = false
```
noamnelke: Since the value is `false` by default, what's the purpose of adding this here?

main.go
```diff
@@ -6,8 +6,14 @@ import (
...
 
+	"cloud.google.com/go/profiler"
 	"github.com/spacemeshos/go-spacemesh/cmd"
 	"github.com/spacemeshos/go-spacemesh/cmd/node"
+	cfg "github.com/spacemeshos/go-spacemesh/config"
+)
+
+var (
+	config = cfg.DefaultConfig()
 )
```
noamnelke: This is no good, it means that you'll always look at the **default** config, ignoring the **actual** config (e.g. config file, command line flags). For this reason, the call to `profiler.Start()` cannot happen in `main.go` and must be moved to after the config has been resolved.

I would put it in `cmd/node/node.go`, inside the `Initialize()` method, after the call to `cmdp.EnsureCLIFlags()`.

noamnelke (changes requested):

cmd/root.go
narayanprusty: Done

noamnelke (approved):
LGTM.

**Reminder: this requires a change to Spacecraft config!**

web3rover:
bors merge

noamnelke:
bors merge

noamnelke:
bors try

AmitShaul:
bors try

noamnelke:
bors merge

PR #2221, opened 2020-12-03
[Merged by Bors] - NTP Servers List Configurable via CLI

### LINKED ISSUE #2063: NTP sync failing in China

web3rover:
## Motivation

Fixes #2063 
Replaces #2142

Currently we create a separate build during testnet release for Chinese smeshers as our default NTP servers are blocked in china. 

## Changes
Now smeshers can provide a list of NTP servers via `--ntp-servers` CLI option

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -116,6 +116,8 @@ func AddCommands(cmd *cobra.Command) {
 		config.TIME.DefaultTimeoutLatency, "Default timeout to ntp query")
 	cmd.PersistentFlags().DurationVar(&config.TIME.RefreshNtpInterval, "refresh-ntp-interval",
 		config.TIME.RefreshNtpInterval, "Refresh intervals to ntp")
+	cmd.PersistentFlags().StringVar(&config.TIME.NTPServersFile,
+		"ntp-servers", config.TIME.NTPServersFile, "Load NTP servers URLs from file")
 	cmd.PersistentFlags().IntVar(&config.P2P.MsgSizeLimit, "msg-size-limit",
 		config.P2P.MsgSizeLimit, "The message size limit in bytes for incoming messages")
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -118,7 +118,7 @@ func AddCommands(cmd *cobra.Command) {
 	cmd.PersistentFlags().DurationVar(&config.TIME.RefreshNtpInterval, "refresh-ntp-interval",
 		config.TIME.RefreshNtpInterval, "Refresh intervals to ntp")
 	cmd.PersistentFlags().StringSliceVar(&config.TIME.NTPServers,
-		"ntp-servers", config.TIME.NTPServers, "Load NTP servers URLs. The total NTP server URLs should be greater than number of NTP queries")
+		"ntp-servers", config.TIME.NTPServers, "A list of NTP servers to query (e.g., 'time.google.com'). Overrides the list in config. Must contain more servers than the number of ntp-queries.")
 	cmd.PersistentFlags().IntVar(&config.P2P.MsgSizeLimit, "msg-size-limit",
 		config.P2P.MsgSizeLimit, "The message size limit in bytes for incoming messages")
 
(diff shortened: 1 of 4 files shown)


noamnelke (commented):
I'm wondering if using a separate file for this is the best strategy. I think we should be consistent with how we implement configuration.

We have a framework for this and we have a config file. I think this should be another value in the config, and not a separate file.

narayanprusty:
> I'm wondering if using a separate file for this is the best strategy. I think we should be consistent with how we implement configuration.
> 
> We have a framework for this and we have a config file. I think this should be another value in the config, and not a separate file.

web3rover:
@noamnelke I agree using the `--config` option would be the best practice. I have updated the PR to adhere to the framework we have currently.

noamnelke (approved):
:shipit:

y0sher:
just small comment I would also add it to the `config.toml` file

config.toml
```diff
@@ -55,6 +55,7 @@ json-port = 9093
 max-allowed-time-drift = "10s"
 ntp-queries = 5
 default-timeout-latency = "10s"
+ntp-servers = ["time1.cloud.tencent.com", "0.cn.pool.ntp.org"]
```
noamnelke: Note that this will override the default config. If we include it, we should put the entire list here.

noamnelke (changes requested):

lrettig (changes requested):
@narayanprusty what happens if a user passes the `--ntp-servers` CLI flag but only specifies, say, a single NTP server? Does it replace the others? Do we want to enforce a minimum number? It would be good to make this clear in the CLI flag description

web3rover:
@lrettig Looks like the total NTP servers URLs should be greater than the number of NTP queries required to be done. I have expanded the CLI flag description accordingly.

cmd/root.go
```diff
@@ -116,6 +117,8 @@ func AddCommands(cmd *cobra.Command) {
 		config.TIME.DefaultTimeoutLatency, "Default timeout to ntp query")
 	cmd.PersistentFlags().DurationVar(&config.TIME.RefreshNtpInterval, "refresh-ntp-interval",
 		config.TIME.RefreshNtpInterval, "Refresh intervals to ntp")
+	cmd.PersistentFlags().StringSliceVar(&config.TIME.NTPServers,
+		"ntp-servers", config.TIME.NTPServers, "Load NTP servers URLs. The total NTP server URLs should be greater than number of NTP queries")
```
lrettig: I think this message could be a little clearer, how's this?

```suggestion
		"ntp-servers", config.TIME.NTPServers, "A list of NTP servers to query (e.g., 'time.google.com'). Overrides the list in config. Must contain more servers than the number of ntp-queries.")
```

lrettig (changes requested):

lrettig (approved):
LGTM thanks!

noamnelke (approved):
:shipit:

lrettig:
Quick question: assuming that we intend this to replace #2142, how do we intend to package this up so that it still works in places like China that need a different set of NTP servers? Should Smapp detect location and set NTP servers accordingly? Or something else?

CC @avive , @IlyaVi

noamnelke:
@lrettig we can still package a separate SMApp, with a different config which would be the same user experience as today, but with a single `go-spacemesh` version (the purpose of this PR is to eliminate the extra version).

Eventually we want to both make the list configurable and make handling of failures smarter (so we can start with a large list of servers by default and eliminate ones that don’t work or have high latency).

Eventually-eventually we want to not rely on NTP at all and have some internal mechanism like bitcoin does.

web3rover:
bors merge

noamnelke:
bors merge

PR #2223, opened 2020-12-07
[Merged by Bors] - Continuous Deployment for Releases

web3rover:
## Motivation

Automate release builds for go-spacemesh

## Changes

Trigger CD pipeline on tag creation to build binary (OS X, Linux and Windows) and docker images for the release and create draft release. Upload builds to GCP storage 

## Test Plan

Create a test tag to test the release and deployment the latest builds by creating a spacecraft dev network.

## TODO


Makefile
```diff
@@ -12,11 +12,19 @@ export GO111MODULE = on
...
 else
        INTERACTIVE := $(shell [ -t 0 ] && echo 1)
 endif
 
+ifdef GO_SPACEMESH_VERSION
+	VERSION=$(GO_SPACEMESH_VERSION)
+endif
+
+ifdef WINDOWS_BUILD_DIR
+	BIN_DIR_WIN=$(WINDOWS_BUILD_DIR)
```
lrettig: I think you can avoid this clause entirely by changing line 8 to:

`BIN_DIR_WIN ?= $(CURR_DIR_WIN)/build`

(and passing this value in as `BIN_DIR_WIN` instead of `WINDOWS_BUILD_DIR`)

Makefile
```diff
@@ -12,11 +12,19 @@ export GO111MODULE = on
 ifeq ($(OS),Windows_NT)
        # Just assume we're in interactive mode on Windows
        INTERACTIVE = 1
-       VERSION = $(shell type version.txt)
+			 VERSION = $(shell type version.txt)
 else
        INTERACTIVE := $(shell [ -t 0 ] && echo 1)
 endif
 
+ifdef GO_SPACEMESH_VERSION
```
lrettig: I'd rather avoid introducing another variable here. Why not just pass this in as `VERSION=...`?

Makefile
```diff
@@ -12,11 +12,19 @@ export GO111MODULE = on
 ifeq ($(OS),Windows_NT)
        # Just assume we're in interactive mode on Windows
        INTERACTIVE = 1
-       VERSION = $(shell type version.txt)
+			 VERSION = $(shell type version.txt)
```
lrettig:
```suggestion
       VERSION = $(shell type version.txt)
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+        run: make install
+      
+      - name: Read version.txt
+        id: version
+        uses: juliangruber/read-file-action@v1
+        with:
+          path: ./version.txt
+      
+      - name: Build go-spacemesh
+        run: make build GO_SPACEMESH_VERSION=${{ steps.version.outputs.content }} WINDOWS_BUILD_DIR=./build
```
lrettig: Maybe add a comment here to indicate that `make build` reads `version.txt` by default, but you're passing in `VERSION` here to be explicit

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+          export_default_credentials: true
+
+      - if: matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest'
+        name: Upload macos and ubuntu build
+        uses: google-github-actions/upload-cloud-storage@main
+        with:
+          path: ./build/go-spacemesh
+          destination: ${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/${{ matrix.os }}
+      
+      - if: matrix.os == 'windows-latest'
```
lrettig: Rather than having two separate tasks here, could you make `path` a variable?

lrettig (changes requested):
Cool! See questions I added. Also, it looks like this is intended to supersede the existing `dockerpush` workflow - please make sure you remove that if you're including it in the new workflow.

web3rover:
@lrettig The `dockerpush` workflow we have already is for develop branch. So it doesn't collide with the release workflow.

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+          if_true: ./build/go-spacemesh.exe
+          if_false: ./build/go-spacemesh
+
+      - name: Upload build
+        uses: google-github-actions/upload-cloud-storage@main
+        with:
+          path: ${{ steps.buildpath.outputs.value }}
+          destination: ${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/${{ matrix.os }}
+      
+  docker-push:
```
lrettig: Could you add a comment here explaining why this is necessary/why it doesn't conflict with the existing dockerpush workflow?

(And, are you sure it's cleaner to do it this way, rather than modifying the existing dockerpush workflow to push new tags, too? I guess it's easier/safer since there's a dependency in the next step--can mention that in the comment here.)

lrettig (changes requested):
LGTM, just needs one explanatory comment

.github/workflows/release.yml
web3rover: @lrettig Actually we can just use the existing workflow in dockerhub.yml. We want to create a draft release only after all builds are success and this will still remain same as `make dockerpush` will also fail if `make build` will fail.

I have updated the code to use the existing dockerpush workflow and made it to trigger during tag push also.

.github/workflows/release.yml
```diff
@@ -63,17 +63,7 @@ jobs:
...
-      - name: Check out Git repository
-        uses: actions/checkout@v2
-      
-      - name: push to dockerhub
-        run: make dockerpush
-
+  
   release:
     runs-on: ubuntu-latest
     needs: docker-push
```
lrettig: I think you need to fix this dependency

lrettig (changes requested):

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+          path: ./version.txt
+      
+      # `make build` reads version from version.txt but in Windows it doesn't work therefore passing version explicitly
+      - name: Build go-spacemesh
+        run: make build VERSION=${{ steps.version.outputs.content }} BIN_DIR_WIN=./build
+
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
```
lrettig, line 50: Secrets need to be set as an input or an env var to be accessible in a workflow. Have you tested this? I guess the `with` section specifies inputs--are you sure you can access the secrets directly here?

Otherwise you may need to add these three GCP_* secrets to the env section at the top. See https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#accessing-your-secrets for more.

lrettig (approved):
LGTM, but please see question I added re: referencing secrets. Also, how are we notified when this runs and cuts a new release? Do you want to add a slack notification here? (See main ci.yml for example.)

.github/workflows/release.yml
web3rover: @lrettig Yes I have tested the complete pipeline. It successfully builds the binaries and uploads them to our gcloud storage and then creates a draft release. So the secrets are being accessed correctly. Here is the pipeline result: https://github.com/spacemeshos/go-spacemesh/runs/1583958931

web3rover: > LGTM, but please see question I added re: referencing secrets. Also, how are we notified when this runs and cuts a new release? Do you want to add a slack notification here? (See main ci.yml for example.)

We don't need any notifications. We only create tags for releases and it creates a draft release so until someone reviews it and fills change log we won't publish it.

lrettig:
Got it. Sounds good. Thanks!

>

noamnelke (approved):
Approved. Let's discuss how this affects the tagging process and what still needs to be done manually (if at all).

web3rover:
bors merge

Issue #2075, opened 2020-07-16
Closing API streams


lrettig:
## Description

Each of the API Stream endpoints uses a `select{}` statement to poll for incoming events to report, and to check if the channels that it's listening to are closed. Right now this is done rather naively: if any of the channels it's listening on are closed, the entire endpoint returns and closes the stream. (The GRPC service can also pass in a ctx.Done, in which case it closes, but I think this would be passed from the remote end of the stream, not from the local node itself.)

- Do we want to close the entire stream if only one of multiple channels closes? (in practice this will probably never happen - they will all be closed together when the node shuts down)
- Do we want to add another case to the `select{}` to listen for a context.Done being passed down from an outer context? (in practice this is probably not necessary since, when the node shuts down, the channels will be shut down, so it will receive the signal that way, but I don't know enough about golang to understand if this risks leaking a channel)

## Affected code
All of the *Stream endpoints in the API, e.g.:

https://github.com/spacemeshos/go-spacemesh/blob/aa7eb73e121e7746de3468d1a365523dc729e3fb/api/grpcserver/globalstate_service.go#L282-L287

lrettig:
@howydev before trying to fix this, I think we need to answer this question:

> Do we want to add another case to the select{} to listen for a context.Done being passed down from an outer context? (in practice this is probably not necessary since, when the node shuts down, the channels will be shut down, so it will receive the signal that way, but I don't know enough about golang to understand if this risks leaking a channel)

We need to understand how the "kill" signal propagates down from the top-level App down to the API channel listeners.

shibe-dev:
@lrettig Makes sense, I'll spend time looking into that before continuing. I assume there's no implementation/usage of the API thus far?

lrettig:
> I assume there's no implementation/usage of the API thus far?

Not sure what you mean - you can run the API on current testnet nodes. We have public API nodes running, and the explorer and dash are using it too!

moshababo:
@lrettig still relevant?

lrettig:
Yes, but low priority/nice to have. We should make sure the API closes cleanly in all scenarios.

PR #2227, opened 2020-12-08
Don't close API streams when one data channel closes

shibe-dev:
## Motivation
fix for issue #2075 

## Changes
improved the channel closing logic (setting a channel to nil stops it from being selected). This design would be an improvement over the current one but I'm unsure if this is sufficient/final (or if its even what you're looking for) - would love feedback @lrettig. Will leave as draft PR

### DIFF
--- a/api/grpcserver/globalstate_service.go
+++ b/api/grpcserver/globalstate_service.go
@@ -230,16 +230,14 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 		channelReceipt = events.GetReceiptChannel()
 	}
 
-	for {
+	accountOk, rewardOk, receiptOk := true, true, true
+
+	for accountOk || rewardOk || receiptOk {
 		select {
-		case updatedAccount, ok := <-channelAccount:
-			if !ok {
-				// we could handle this more gracefully, by no longer listening
-				// to this stream but continuing to listen to the other stream,
-				// but in practice one should never be closed while the other is
-				// still running, so it doesn't matter
-				log.Info("account channel closed, shutting down")
-				return nil
+		case updatedAccount, accountOk := <-channelAccount:
+			if !accountOk {
+				log.Info("account channel closed")
+				channelAccount = nil
 			}
 			// Apply address filter
 			if updatedAccount == addr {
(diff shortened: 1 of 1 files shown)


api/grpcserver/globalstate_service.go
```diff
@@ -230,16 +230,14 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
-				// we could handle this more gracefully, by no longer listening
-				// to this stream but continuing to listen to the other stream,
-				// but in practice one should never be closed while the other is
-				// still running, so it doesn't matter
-				log.Info("account channel closed, shutting down")
-				return nil
+		case updatedAccount, accountOk := <-channelAccount:
+			if !accountOk {
+				log.Info("account channel closed")
+				channelAccount = nil
```
lrettig, line 240: if `accountOk` is false, does that mean `updatedAccount` is nil? The next line reads `updatedAccount`, will this work okay if the channel has just been closed or do you need to add a check for this case? (here and in the others, below)

api/grpcserver/globalstate_service.go
```diff
@@ -317,6 +307,8 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 		// that the service needs to shut down?
 		// See https://github.com/spacemeshos/go-spacemesh/issues/2075
 	}
+
+	return nil
```
lrettig, line 311: Please add a comment here explaining what this line does and why/when it's necessary, the context is lost here

lrettig (commented):
See https://github.com/spacemeshos/go-spacemesh/issues/2075#issuecomment-742014594 - I don't think this task is well-defined yet

api/grpcserver/globalstate_service.go
shibe-dev: im still rather new to golang but it seems like [setting a channel a to nil stops it from being selected](https://golangbyexample.com/select-with-nil-channel-golang/) which seems to be the logic you were suggesting that the API should go for

lrettig: Yes, that may be true for the _next_ loop iteration, but what about the current iteration?

antonlerner:
closed due to inactivity

PR #2229, opened 2020-12-09
use Hash20 methods for BlockID

noamnelke:
## Motivation
@avive reported that the API returns block IDs in the following format:
```
"blocks": [
   {
    "id": "JYtHaDEC3+vFXE+J1tt3r/iCWV4AAAAAAAAAAAAAAAA="
   },
   {
    "id": "M+L13e+kWtfenEvf3lSWjoVod+sAAAAAAAAAAAAAAAA="
   },
   {
    "id": "75FxQYyoh1mWlN03AgGD3wJa4MgAAAAAAAAAAAAAAAA="
   }
]
```
The zero padding at the end (represented as `A`s) isn't what we want.

## Changes
Because `BlockID` is a `Hash20`, using `Hash32` methods causes the result to be zero padded. I replaced the usage of `Hash32` methods with equivalent `Hash20` methods. The method used by the API code is `Bytes()` and this just returns 20 bytes now (unpadded).

Additionally, the `String()` method used to return a `ShortString()` which is confusing, so I fixed that, too.

## Test Plan
No new tests needed.

## External effects
This changes the output format of blocks and may affect the dashboard and explorer.

### DIFF
--- a/common/types/block.go
+++ b/common/types/block.go
@@ -17,14 +17,19 @@ import (
 // BlockID is a 20-byte sha256 sum of the serialized block, used to identify it.
 type BlockID Hash20
 
-// String returns a short prefix of the hex representation of the ID.
+// String returns the hex representation of the ID.
 func (id BlockID) String() string {
-	return id.AsHash32().ShortString()
+	return id.AsHash20().String()
+}
+
+// ShortString returns a short prefix of the hex representation of the ID.
+func (id BlockID) ShortString() string {
+	return id.AsHash20().ShortString()
 }
 
 // Field returns a log field. Implements the LoggableField interface.
 func (id BlockID) Field() log.Field {
-	return log.String("block_id", id.AsHash32().ShortString())
+	return log.String("block_id", id.ShortString())
 }
 
 // Compare returns true if other (the given BlockID) is less than this BlockID, by lexicographic comparison.
@@ -35,6 +40,11 @@ func (id BlockID) Compare(other BlockID) bool {
 // AsHash32 returns a Hash32 whose first 20 bytes are the bytes of this BlockID, it is right-padded with zeros.
 func (id BlockID) AsHash32() Hash32 {
 	return Hash20(id).ToHash32()
+}
+
+// AsHash20 returns this BlockID as a Hash20.
+func (id BlockID) AsHash20() Hash20 {
+	return Hash20(id)
 }
 
 var layersPerEpoch int32
@@ -228,7 +238,7 @@ func (b Block) Hash32() Hash32 {
 
 // ShortString returns a the first 5 characters of the ID, for logging purposes.
 func (b Block) ShortString() string {
-	return b.id.AsHash32().ShortString()
+	return b.id.ShortString()
 }
 
 // MinerID returns this block's miner's Edwards public key.
--- a/common/types/encode.go
+++ b/common/types/encode.go
@@ -10,7 +10,7 @@ import (
 )
 
 // Bytes returns the BlockID as a byte slice.
-func (id BlockID) Bytes() []byte { return id.AsHash32().Bytes() }
+func (id BlockID) Bytes() []byte { return id.AsHash20().Bytes() }
 
 // Bytes returns the byte representation of the LayerID, using little endian encoding.
 func (l LayerID) Bytes() []byte { return util.Uint64ToBytes(uint64(l)) }
--- a/events/reporter.go
+++ b/events/reporter.go
@@ -118,7 +118,7 @@ func ReportRewardReceived(r Reward) {
 // ReportNewBlock reports a new block
 func ReportNewBlock(blk *types.Block) {
 	Publish(NewBlock{
-		ID:    blk.ID().String(),
+		ID:    blk.ID().ShortString(),
 		Atx:   blk.ATXID.ShortString(),
 		Layer: uint64(blk.LayerIndex),
 	})
@@ -127,7 +127,7 @@ func ReportNewBlock(blk *types.Block) {
 // ReportValidBlock reports a valid block
 func ReportValidBlock(blockID types.BlockID, valid bool) {
 	Publish(ValidBlock{
-		ID:    blockID.String(),
+		ID:    blockID.ShortString(),
 		Valid: valid,
 	})
 }
--- a/api/grpcserver/grpcserver_test.go
+++ b/api/grpcserver/grpcserver_test.go
@@ -1513,7 +1513,7 @@ func TestTransactionServiceSubmitUnsync(t *testing.T) {
 		context.Background(),
 		&pb.SubmitTransactionRequest{Transaction: serializedTx},
 	)
-	require.Error(err)
+	require.EqualError(err, "rpc error: code = FailedPrecondition desc = Cannot submit transaction, node is not in sync yet, try again later")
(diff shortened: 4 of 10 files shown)


common/types/encode.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 // Bytes returns the BlockID as a byte slice.
-func (id BlockID) Bytes() []byte { return id.AsHash32().Bytes() }
+func (id BlockID) Bytes() []byte { return id.AsHash20().Bytes() }
```
antonlerner, line 13: this is not correct I think, and will interfere with current sync and DBconventions. 
currently all IDs in all DBs are of Hash32 format and i think we should keep it as such since it unified our ID in DB format

noamnelke: I'm not sure you're right about this.

Database keys are variable-size byte slices and the blocks database only has blocks in it, so there's no advantage for conformity there. The only way to fetch from this database is by using a getter that accepts a `BlockID` as an argument (not a byte slice).

The sync protocol requires a `Hash32` as a key and it still gets it (justifiably doesn't use `.Bytes()`). The block request handler converts the `Hash32` it receives to a `Hash20` and then to a `BlockID` before fetching from the database anyway (this was before my change).

All the tests are passing.

noamnelke: BTW, if we really do need to often get the padded 32-byte representation (I don't think we do) then we should create a method named `.Bytes32()` which would be explicit and less confusing than padding the returned byte slice from `.Bytes()`.

antonlerner: the new sync relies on the keys of all DBs to be the size of hash32

noamnelke: Can you be more specific? It's hard for me to respond on such a high level...

Did you make the new sync access the blocks database directly, not via the setter and getter as it is in `develop` now?

Can you send me a link to the code so I can take a look and understand?

noamnelke: This is the current sync block request handler:

https://github.com/spacemeshos/go-spacemesh/blob/f76c8b99427a4a9f559da87fced6cf5aa3219266/sync/handler.go#L89-L123

As you can see, the list of `blockids` it receives is a list of `Hash32`s, which it then converts to `BlockID`s before fetching the blocks.

What's the equivalent code that is sensitive to the format of the database key?

common/types/block.go
```diff
@@ -17,14 +17,19 @@ import (
 // BlockID is a 20-byte sha256 sum of the serialized block, used to identify it.
 type BlockID Hash20
 
-// String returns a short prefix of the hex representation of the ID.
+// String returns the hex representation of the ID.
 func (id BlockID) String() string {
-	return id.AsHash32().ShortString()
+	return id.AsHash20().String()
```
lrettig, line 22: Does this actually increase the length of the return value from this method? I think this was intentionally shortened (using ShortString) a few months ago to reduce the size of the logs

events/reporter.go
```diff
@@ -118,7 +118,7 @@ func ReportRewardReceived(r Reward) {
 // ReportNewBlock reports a new block
 func ReportNewBlock(blk *types.Block) {
 	Publish(NewBlock{
-		ID:    blk.ID().String(),
+		ID:    blk.ID().ShortString(),
```
lrettig, line 121: While I think the code here does the right thing, it's a bit misleading because at a glance `ShortString` sounds like it abbreviates the block ID. We might want to make this more explicit.

lrettig (commented):
From the perspective of the API I think this is fine. My only additional comment is that I can't tell if `ShortString` is actually shortening/abbreviating the block ID. If it is, then I think it should only be used in debug output (and not in, e.g., the API). If it isn't, then the name is misleading.

noamnelke:
@lrettig the previous implementation of `BlockID`'s `.String()` method returned a shortened string.

I changed the implementation to return the full string, so I changed all the explicit usages of `.String()` to `.ShortString()` (which returns exactly what the previous `.String()` returned). There may be implicit usages that are hard to find, but the `.Field()` method (used for logs) of `BlockID` uses `.ShortString()`. This means that it's now clear what's returned here, but nothing changed in the actual behavior - if there's an issue, it was there before (if that's the case we should fix it, I'm just not sure).

`.String()` isn't used in the API anyway. What's used in the API is `.Bytes()`. That used to return a byte slice of length 32, even though the `BlockID` is only 20 bytes long. The last 12 bytes were padded with zeros. Now `.Bytes()` (correctly) returns a byte slice of length 20, matching the actual length of the object.

common/types/block.go
noamnelke: This increases the length of the return value, but I changed all usages of this method to `ShortString()` to make it clear that they're getting the shortened version.

events/reporter.go
noamnelke: It DOES return the abbreviated version. If this is not what we want we should change it back to use `.String()` (which used to return the shortened version, but now does what it says).

noamnelke: I just realized that this is for the API... I said in another comment that `.String()` was not used in the API and I was wrong. Why do we use strings here at all? Why not bytes directly? That's what we do here:

https://github.com/spacemeshos/go-spacemesh/blob/f76c8b99427a4a9f559da87fced6cf5aa3219266/api/grpcserver/mesh_service.go#L320-L323

lrettig:
Thanks @noamnelke, sounds good to me. `Publish` goes out to pubsub, which is only used internally/for testing, not out to the API, which (as you pointed out) uses `Bytes`. I'm happy here.

Will let @antonlerner follow up on the database/sync questions here and let him approve this.

events/reporter.go
noamnelke: ok, so for the record - this is NOT for the API - it's for pubsub. Leaving it as is.

common/types/encode.go
antonlerner: this is the new fetch code, as you can see it is generic and assumes same size of hash for all hashes
https://github.com/spacemeshos/go-spacemesh/pull/2213/commits/d1aa40236595002e4541db8b46c315c7ef937dd5#diff-f03683d1546e71521624eea58e910c92a2525ec1a0c6d216c8cad6b3a9d26a8aR252-R299

noamnelke: Yes, this code will break that design.

It seems we're already too far along on that path to discuss if it's a good idea or not. I intuitively don't like allowing remote queries directly into the node's databases. I need to give it some thought if it's a security issue or not (given that all the data in these DBs is public anyway). That's regardless this specific fix.

In order to simplify the discussion I'll create an alternative fix for Aviv's issue that's local to the API rather than fixing (what I consider to be) the underlying problem.

noamnelke: Assuming we stay with your current design, I think we should introduce a `.Bytes32()` method and implement that on all key objects. This will make it explicit that this method always returns 32 bytes, and not necessarily the actual bytes of the object - I think it's currently confusing.

I also suggest creating a wrapper of our current database wrapper, where the `.Get()` and `.Put()` methods take a `Hash32` as the key - I'd then make the sync code only work with that kind of database and not the generic version where keys can be any length. I think the current implementation is possibly confusing and introduces unnecessary risk of coding errors.

noamnelke:
Closing this in favor of #2233 because making database keys other than 32-bytes would break #2213.

PR #2232, opened 2020-12-14
Fix flaky blocks-add-node test

### LINKED ISSUE #2222: Fix flaky blocks-add-node and other tests
## Description
Sometimes add_nodes test fails when some node does not create the correct amount of blocks per epoch

this happens here:
https://github.com/spacemeshos/go-spacemesh/runs/1428531039

the error is usually:
```
>           assert blocks_per_layer == wanted_res, ass_err
E           AssertionError: node 3314eab40ebc14bdc912d2b03309b002869308a96a2860757024ee6f35a3d675 failed creating the avg block size
E           blocks created per layer 1.0, wanted average block per node 0.8
```

## Steps To Reproduce
run `make dockertest-blocks-add-node`
or 
`python3 -m pytest -s -v ./tests/sync/test_sync.py --tc-file=./tests/sync/config.yaml --tc-format=yaml` 

## Expected Behavior
all test should pass, a miner should create 1 block per layer in average at least

## how to investigate
we should look at their declared active set and see it matches other miners active set 
we should also at the blockOracle and the blockBuilder to make sure they don’t have any errors and they are active at all times

nkryuchkov:
## Motivation

Closes #2222 

## Changes

- Fix flaky `blocks-add-node` test

## Test Plan

- System tests

## TODO

- [X] Fix flaky blocks-add-node test
- [ ] Finish created todo's
- [ ] Document changes


noamnelke:
@nkryuchkov I saw you reformatted a few files - I'm generally not against this, but at this point in time it may cause avoidable conflicts with other PRs, so I suggest reverting those changes and possibly making them in a separate PR after we've mostly emptied our PR queue.

Only the following files should be part of this PR:
```
tests/block_atx/add_node/test_blocks_add_node.py
tests/queries.py
tests/utils.py
```

nkryuchkov:
@noamnelke Thanks, I agree.

nkryuchkov:
Superseded by https://github.com/spacemeshos/go-spacemesh/pull/2239

PR #2234, opened 2020-12-17
New CI infrastructure to support private ELK per test

AmitShaul:
## Motivation
added new CI infra
there's another PR open #2197, the difference is that in #2197 there is a code duplication of the tests (tests_elk) folder which doesn't exists here.

## Changes
Node pool for each test
ELK cluster per each test
Both node pool and ELK cluster are deployed using new pytest fixtures

## Overview
**new test:**
When starting a new test a namespace will be created for that test and that test only.
First an ELK cluster will be created over that namespace, ELK is built from Fluent-bit, Logstash, Kibana and Elasticsearch.
After setting the ELK cluster we wait for ES and Logstash to be ready, this process might take a couple of minutes.
When ELK is ready we add the node pool that will serve the Spacemesh nodes, this node pool is necessary in order to get all of the resources needed (for our SM clients) in advance so our deployment will happen in a shorter and more predictable time.
The process of creating\deleting\altering a node pool is blocking since only one action can be made simultaneously which means that if someone just started a test and wants to create a node pool he will have to wait for previous started action to end.
Node pool creation\deletion is polling every 10 seconds to see if it can start and in case it can't (for the reason mentioned above) it  creates a single "Error" message for the first time it tries, this message can be ignored for it will start once GCP is ready to take another action.

**Teardown**:
After a test was done a process of teardown will start in the opposite order, first the tests' node pool will be deleted (where most of the tests' resources are) and after that the namespace will be deleted deleting the ELK with it.

**Logging**:
After a test was finished logs will be dumped to the main ES server located on 'logging' namespace.
Logs will be dumped under two conditions:
1. The test has failed.
2. 'is_dump: True' was added to the tests' config file.

In any other case those logs will be lost with the namespace deletion.
Until namespace is deleted logs are accessible in the tests' private ELK, the IPs can be listed by running:
'kubectl get services -n YOUR_NAMESPACE'


AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

web3rover:
bors try

web3rover:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

tests/config.py
```diff
@@ -3,6 +3,16 @@
...
 CLIENT_STATEFULSET_FILE = './k8s/client-w-conf-ss.yml'
 CURL_POD_FILE = './k8s/curl.yml'
+ELASTIC_CONF_DIR = './elk/elastic/'
+FILEBEAT_CONF_DIR = './elk/filebeat/'
+FLUENT_BIT_CONF_DIR = './elk/fluent-bit/'
+KIBANA_CONF_DIR = './elk/kibana/'
+LOGSTASH_CONF_DIR = './elk/logstash/'
+
+ES_USER_LOCAL = "elastic"
+ES_PASS_LOCAL = "gavradon"
```
y0sher: it's better not to include the password here as plain text

tests/config.py
```diff
@@ -3,6 +3,16 @@
...
+MAIN_ES_URL = "35.197.82.152:9200"
```
y0sher: also maybe create a k8s service for that like we have now ?

tests/conftest.py
```diff
@@ -243,12 +247,36 @@ def save_log_on_exit(request):
...
+def add_node_pool():
+    """
+    memory should be represented by number of megabytes, \d*M
+
+    :return:
+    """
+    deployer = NodePoolDep(testconfig)
+    _, time_elapsed = deployer.add_node_pool()
+    yield time_elapsed
+    deployer.remove_node_pool()
```
y0sher: maybe add some indicator about time passed so we have some feeling of how much time we're waiting ?

y0sher: also wouldn't this code remove the node_pool right after its created?

AmitShaul: added a print for time_elapsed.
and about your question, no.
everything after the yield command is working as a destructor and will run only when this fixture will perform teardown.

tests/config.py
AmitShaul: is base64 adequate?

lrettig: no, we should not store passwords in text files. It should be a GitHub secret attached to the repository and passed in through the environment.

.github/workflows/ci.yml
```diff
@@ -157,9 +159,9 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v2
-      - name: hare+mining system tests
-        run: make dockertest-hare-mining
```
lrettig, line 161: Was this test removed? Or replaced?

.github/workflows/ci.yml
```diff
@@ -168,9 +170,9 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v2
-      - name: sync+blocks remove node test
-        run: make dockertest-sync-blocks-remove-node
```
lrettig, line 172: This one too, was it removed or replaced? I don't see it anywhere else

Makefile
```diff
@@ -227,104 +236,103 @@ DELIM=::
 endif
 
 
-dockerrun-p2p:
+dockerrun-p2p-elk:
```
lrettig, line 241: How important is it to rename this and all the other make recipes to add `-elk`? I think you could just modify the recipes to use `dockerbuild-test-elk` without renaming them all no?

tests/analyse.py
```diff
@@ -53,10 +53,11 @@ def analyze_mining(deployment, last_layer, layers_per_epoch, layer_avg_size, tot
     print("atx created " + str(total_atxs))
     print("blocks created " + str(total_blocks) + " first epoch blocks " + str(first_epoch_blocks))
 
+    up_pods = queries.get_nodes_up(deployment)
     # not enough pods deployed or pods restarted
-    assert queries.get_nodes_up(deployment) == total_pods
+    assert up_pods == total_pods, f"up pods: {up_pods}, total: {total_pods}"
     # not all nodes created blocks
-    assert total_pods == len(blockmap)
+    assert total_pods == len(blockmap), f"total pods={total_pods}, number of nodes created blocks{len(blockmap)}"
```
lrettig:
```suggestion
    assert total_pods == len(blockmap), f"total pods={total_pods}, number of nodes that created blocks={len(blockmap)}"
```

tests/context.py
```diff
@@ -21,16 +20,19 @@ def wrapper(*args, **kwargs):
...
-        ctxt = Context()
-        cluster_name = ctxt.get_cluster_name()
-        es_url = generate_elastic_url(cluster_name)
-        print("ES_PASSWD is {0}".format(ES_PASSWD))
-        self.es = Elasticsearch(es_url, http_auth=("spacemesh", ES_PASSWD), port=80, timeout=90)
+    def __init__(self, namespace):
+        self.namespace = namespace
+        self.es_ip = self.get_elastic_ip()
+        self.es = Elasticsearch(
+            self.es_ip, port=9200, timeout=30, max_retries=3, retry_on_timeout=True, http_auth=('elastic', 'gavradon')
```
lrettig: don't hardcode username and password here either

tests/convenience.py
```diff
@@ -21,12 +21,16 @@ def convert_ts_to_datetime(ts):
...
+    if msg:
+        print(f"\n{msg}")
+    else:
+        print(f"\nsleeping for {tts} seconds\n")
     while tts != 0:
         tts -= 1
-        print(f" {tts} seconds left     ", end="\r")
+        print(f" {tts} seconds left             ", end="\r")
         time.sleep(1)
+    print("\033[A \033[A")
```
lrettig, line 34: What does this print? Could you add a comment to make it clear from the code?

tests/k8s_handler.py
```diff
@@ -0,0 +1,197 @@
...
+            if dep['kind'] == 'StatefulSet':
+                k8s_client = client.AppsV1Api()
+                k8s_client.delete_namespaced_stateful_set(name=name, namespace=namespace)
+            elif dep['kind'] == 'DaemonSet':
+                k8s_client = client.AppsV1Api()
+                k8s_client.delete_namespaced_daemon_set(name=name, namespace=namespace)
+            elif dep['kind'] == 'Deployment':
+                k8s_client = client.AppsV1Api()
+                k8s_client.delete_namespaced_deployment(name=name, namespace=namespace)
+            # elif dep['kind'] == 'Service':
```
lrettig: Why add commented-out code here?

tests/node_pool_deployer.py
```diff
@@ -0,0 +1,120 @@
...
+                 '--no-enable-autoupgrade --enable-autorepair --max-surge-upgrade 1 --max-unavailable-upgrade 0'
+
+    def __init__(self, testconfig):
+        self.testconfig = testconfig
+        self.namespace = testconfig["namespace"]
+        self.pool_name = f"pool-{self.namespace}"
+        self.cluster_name = "spacemesh-cluster-elk"
+        self.zone = "us-west1-a"
+        self.default_config = {
+            "project_name": "spacemesh-198810",
```
lrettig, line 29: This (and the `zone`, above) should probably also not be hardcoded, this should probably be passed in from the env so it's easier to change, or stored in a config file or something

tests/node_pool_deployer.py
```diff
@@ -0,0 +1,120 @@
...
+        ut.exec_wait(self.gcloud_delete.format(pool_name=self.pool_name, cluster_name=self.cluster_name, zone=self.zone)
+                     , retry=80, interval=10)
+
+    def get_total_cpu_and_mem(self):
+        _, cpu, mem = self.get_spec_resources("bootstrap")
+        total_pods = int(self.testconfig["total_pods"])
+        total_cpu = total_pods * cpu
+        total_mem = total_pods * mem
+        if total_cpu % 2:
+            print("adding additional CPU, total CPU was not divide by 2, total CPUs:", total_cpu)
```
lrettig:
```suggestion
            print("adding additional CPU, total CPU count not divisible by 2, total CPUs:", total_cpu)
```

tests/node_pool_deployer.py
```diff
@@ -0,0 +1,120 @@
...
+            total_cpu += 1
+        if total_mem % 512:
+            raise Exception("total memory is not a 512 multiplication")
```
lrettig:
```suggestion
            raise Exception("total memory is not divisible by 512")
```

lrettig (changes requested):

.github/workflows/ci.yml
AmitShaul: we don't run tests together in pairs anymore

.github/workflows/ci.yml
AmitShaul: you can see them run separately

Makefile
AmitShaul: tests before were named without the elk and because I had to change them and keep the old ones I added the "-elk"
I don't think it's something I need to change, if anyone has a strong objection I can consider it.

tests/convenience.py
AmitShaul: it removed the last printed line.
added a comment

tests/k8s_handler.py
AmitShaul: uncommented

tests/node_pool_deployer.py
AmitShaul: fixed

tests/analyse.py
AmitShaul: fixed

tests/node_pool_deployer.py
AmitShaul: fixed

PR #2237, opened 2020-12-22
[Merged by Bors] - fixed different active set size in block creation

antonlerner:
We've had some issues with a race creating blocks and declaring active set size 
the race was when we were syncing and trying to create a block at the same time

this caused the active set size to sometimes be different from the time we calculate eligibility proofs to the time we actually declare the active set size

now we will use the same active set to calculate and to declare in blocks

### DIFF
--- a/blocks/blockoracle.go
+++ b/blocks/blockoracle.go
@@ -34,6 +34,7 @@ type Oracle struct {
 	nodeID               types.NodeID
 
 	proofsEpoch       types.EpochID
+	epochAtxs         []types.ATXID
 	eligibilityProofs map[types.LayerID][]types.BlockEligibilityProof
 	atxID             types.ATXID
 	isSynced          func() bool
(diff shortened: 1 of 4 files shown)


y0sher (approved):
looks good and I think this is more correct than what we had before.

blocks/blockoracle.go
```diff
@@ -60,23 +60,23 @@ func NewMinerBlockOracle(committeeSize uint32, genesisActiveSetSize uint32, laye
...
-		err := bo.calcEligibilityProofs(epochNumber)
-		if err != nil {
-			bo.log.Error("failed to calculate eligibility proofs for epoch %v : %v", epochNumber, err)
-			return *types.EmptyATXID, nil, err
-		}
+
+	activeSet, err := bo.calcEligibilityProofs(epochNumber)
+	if err != nil {
+		bo.log.Error("failed to calculate eligibility proofs for epoch %v : %v", epochNumber, err)
+		return *types.EmptyATXID, nil, nil, err
```
noamnelke: This may be problematic. It essentially cancels caching and calculates the miner's eligibility every time this method is called. Calculating the eligibility is very expensive. It takes several seconds with testnet parameters (288 layers per epoch with 50 blocks each = 14,400 blocks per epoch) it will be even heavier in mainnet (14x = 201,600 blocks per epoch).

The plan that I had in mind was to calculate this during the last layer of the epoch for the next epoch. Doing the calculation in real-time at the start of each layer is wasteful and not practical.

If you want to know the `activeSet` you should cache it during eligibility proof calculation and get it from the cache instead of canceling the cache.

noamnelke (changes requested):

blocks/blockoracle.go
antonlerner: the active set is not calculated using mesh traversal so calculating active set size becomes a matter of reading ATX ids from database

noamnelke: `calcEligibilityProofs` isn't expensive because of traversal, it performs a lot of signing. The new algo in var post may be more efficient, but I haven't benchmarked it yet, so can't say for sure.

I think that adding the active set to the cache is trivial - let's do that.

antonlerner: fixed, pls approve :)

noamnelke (approved):
Looks good!

antonlerner:
bors merge

PR #2239, opened 2020-12-29
[Merged by Bors] - work parallel worker queue work

antonlerner:
## Motivation
try to run the queue in parallel to not deadlock itself
## TODO


antonlerner:
bors try

antonlerner:
bors merge

antonlerner:
bors try

sync/fetch_queue.go
```diff
@@ -103,9 +112,9 @@ func (fq *fetchQueue) addToPending(ids []types.Hash32) []chan bool {
 	for _, id := range ids {
 		ch := make(chan bool, 1)
 		deps = append(deps, ch)
-		if _, ok := fq.pending[id]; !ok {
-			idsToAdd = append(idsToAdd, id)
-		}
+		//if _, ok := fq.pending[id]; !ok {
+		idsToAdd = append(idsToAdd, id)
+		//}
```
noamnelke: Not sure why the conditional was removed here...

It seems that without the conditional `idsToAdd` is an exact copy of `ids`.

If this change was accidental, let's revert it. Otherwise, let's remove the commented out code and also the `idsToAdd` variable and use `ids` instead.

sync/fetch_queue.go
```diff
@@ -67,27 +67,36 @@ func (fq *fetchQueue) work() error {
...
 	output := fetchWithFactory(newFetchWorker(fq.workerInfra, runtime.NumCPU(), fq.batchRequestFactory, fq.queue, fq.name))
-	for out := range output {
-		fq.Debug("new batch out of queue")
-		if out == nil {
-			fq.Info("close queue")
-			return nil
-		}
+	wg := sync.WaitGroup{}
+	wg.Add(runtime.NumCPU())
+	for i := 0; i < runtime.NumCPU(); i++ {
```
noamnelke: `fq.handleFetch(bjb)` is the actual work happening in these threads? I assume it's mostly CPU-bound and that's why you use `runtime.NumCPU()` as the concurrency factor?

In any case, I'd make a constant/variable for this instead of calling `runtime.NumCPU()` twice. Would make the code clearer IMO and be slightly less error prone.

noamnelke (approved):
LGTM, only minor comments.

y0sher:
bors try

antonlerner:
bors merge

antonlerner:
bors merge

y0sher:
bors merge

y0sher:
bors merge

PR #2251, opened 2021-01-10
Add new transactions abstraction (Smip23/1)

sudachen:
## Motivation

This PR adds new transactions abstraction allowing to operate with different classes of transactions.

## Changes

1) Added 3 new types of transactions with 2 signature schemes each. Also, there is one type for the old transactions layout.
2) Client code is adoped to use the new transactions abstraction.

## TODO

- [ ] Add NetworkID calculation
- [x] Sync with current development branch
- [x] Add documentation comments


avive:
This is a very cool pr! Can you please add short go-style docs to all new/modified types and functions? This will greatly help implementing work in other languages such as c / typescript / etc... I think that tests fail due to lack of comments...

api/grpcserver/mesh_service.go
```diff
@@ -255,12 +255,18 @@ func (s MeshService) getTxIdsFromMesh(minLayer types.LayerID, addr types.Address
...
 }
 
-func convertTransaction(t *types.Transaction) *pb.Transaction {
+func convertTransaction(t types.Transaction) *pb.Transaction {
+	scheme := func() pb.Signature_Scheme {
+		if t.Type().IsEdPlus() {
+			return pb.Signature_SCHEME_ED25519_PLUS_PLUS
+		}
+		return pb.Signature_SCHEME_ED25519
+	}
```
noamnelke: This is a function to avoid memory allocation? I assume it's inlined by the compiler.

Maybe worth adding a comment to explain.

sync/mocks.go
```diff
@@ -135,16 +135,18 @@ func (*validatorMock) VerifyPost(signing.PublicKey, *types.PostProof, uint64) er
 
 type mockTxMemPool struct{}
 
-func (mockTxMemPool) Get(types.TransactionID) (*types.Transaction, error) {
-	return &types.Transaction{}, nil
+var alice = signing.NewEdSignerSeed("alice")
```
noamnelke, line 138: where is this used?

common/types/encode.go
```diff
@@ -54,19 +54,17 @@ func NIPSTChallengeToBytes(challenge *NIPSTChallenge) ([]byte, error) {
...
+func BytesToTransaction(buf []byte) (Transaction, error) {
+	return SignedTransaction(buf).Decode()
 }
 
 // BytesToInterface deserializes any type.
 // ⚠️ Pass the interface by reference
 func BytesToInterface(buf []byte, i interface{}) error {
+	if _, ok := i.(*Transaction); ok {
+		panic("use SignedTransaction.Decode()")
+	}
```
noamnelke, line 67: 💯

mesh/meshdb.go
```diff
@@ -382,38 +382,12 @@ func getTransactionDestKeyPrefix(l types.LayerID, account types.Address) []byte
...
-			return fmt.Errorf("could not write tx %v to database: %v", t.ID().ShortString(), err)
-		}
-		if err := batch.Put(getTransactionDestKey(l, t), t.ID().Bytes()); err != nil {
-			return fmt.Errorf("could not write tx %v to database: %v", t.ID().ShortString(), err)
+		if err := m.WriteTransaction(l, t); err != nil {
+			return err
 		}
-		m.Debug("wrote tx %v to db", t.ID().ShortString())
 	}
 	err := batch.Write()
```
noamnelke: This method originally wrote transactions atomically, if the batch failed, nothing was written.

I'm not 100% sure we fully implemented the logic for what happens on failure, but this change feels a little like a step backwards. Worth reconsidering.

noamnelke (approved):
Because this changes the binary format of transactions, I expect this to fail the sync test which relies on pre-generated binary data stored in GCP. There's a way to generate this data which I never remember. Let me get back to you about how to do it, I need to search my Slack history.

noamnelke:
BTW, 👍 on the transaction hierarchy - looks good!

sync/mocks.go
sudachen: https://github.com/spacemeshos/go-spacemesh/blob/0556bf84df74406c12aede6afb14cab2a79fda05/sync/mocks.go#L141

api/grpcserver/mesh_service.go
sudachen: Simplified
https://github.com/spacemeshos/go-spacemesh/blob/80b86b1c9f9f93da13bcca40d663417ccec48603/api/grpcserver/mesh_service.go#L259

mesh/meshdb.go
sudachen: Fixed
https://github.com/spacemeshos/go-spacemesh/blob/80b86b1c9f9f93da13bcca40d663417ccec48603/mesh/meshdb.go#L385

sudachen:
bors try

sudachen:
created the new updated PR

PR #2259, opened 2021-01-17
[Merged by Bors] - Fix GHA Automated Builds

web3rover:
Currently the release builds are failing because protoc requires $GOPATH/bin to be added to $PATH. This PR adds this path so that protoc binaries can be found during `make build`

**Note that for builds to work the version.txt file shouldn't have a new line character.**


.github/workflows/release.yml
```diff
@@ -81,4 +97,4 @@ jobs:
             - Linux: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/ubuntu-latest/go-spacemesh
             - Mac: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/macos-latest/go-spacemesh
           draft: true
-          prerelease: true
+          prerelease: true
```
noamnelke: why remove the end-of-file line break?

noamnelke (approved):
Let's restore the end-of-file line break before merging?

noamnelke (changes requested):
Actually... just noticed the comment... if you don't want a line break in `version.txt` - then let's fix it here.

noamnelke (approved):

web3rover:
bors merge

lrettig:
protoc is no longer used in production - it was removed a long time ago - it's only needed for the current build cycle because it's using very old code. Should this change be reverted?

noamnelke:
once the testnet is caught up with the new code we can consider it, but I think puting `$GOPATH/bin` in `$PATH` is standard practice, so it's not a big deal if it stays this way, isn't it?

lrettig:
> once the testnet is caught up with the new code we can consider it, but I think puting `$GOPATH/bin` in `$PATH` is standard practice, so it's not a big deal if it stays this way, isn't it?

True, it's probably no big deal, but the `ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'` caught my eye and I'm not 100% sure it's something we want in production forever - I guess that's up to @narayanprusty. This change _could_ be removed from `develop` and just cherry-picked into future testnet builds (until testnet catches up with the new API code).

noamnelke:
>This change _could_ be removed from `develop` and just cherry-picked into future testnet builds (until testnet catches up with the new API code).

Yes, I agree.

@narayanprusty WDYT?

We could create a PR that reverts this now and merge it, since it's already in the current testnet and future testnets will either be based on the current one, or on `develop` once the SMApp supports API v2 (and then we don't need this change anyway).

web3rover:
Having $GOPATH/bin in $PATH is already deprecated. https://blog.golang.org/modules2019 As we have protoc in old code therefore we needed it. 

`ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'` doesn't cause any security issues as it's defined for a specific step and not at global level. 

Regardless of that, I can create another PR to revert it and re-open this PR so that we can cherry-pick this PR until we have old code.

web3rover:
@noamnelke @lrettig Here is the PR to revert the code: #2265

noamnelke:
No need to reopen this PR - approved the other one.

PR #2260, opened 2021-01-17
Testnet prints

y0sher:
## Motivation
We had bugs in testnet that we couldn't trace, add more verbosity to logs so it'll be easier to find next time.


noamnelke (approved):
This adds ~5k prints per layer (2 messages per processed block per miner). It's not the end of the world, but do you think it's worth cleaning up after we find what the issue is?

y0sher:
for sure we shouldn't include this in next testnet. we have the collected logs offline for deeper investigation.

PR #2262, opened 2021-01-18
[Merged by Bors] - Add SmesherID Index to Reward DB

### LINKED ISSUE #2068: Add db index for reward smesher ID
For the API we need to know which smesher and coinbase received each reward (https://github.com/spacemeshos/api/blob/ff4a9a6b913a23a83cf07a0a7b4b7c16605ce9c0/proto/spacemesh/v1/types.proto#L100-L107). Right now, we can find rewards by account that received the reward but there is no smesher ID or coinbase attached to it:

https://github.com/spacemeshos/go-spacemesh/blob/61eafeeb2a9e3ec785113cf8284eec8bf8ef6ee9/mesh/meshdb.go#L436

Similar to #2064

sidsankhe:
## Motivation

Closes #2068 

## Changes

1.  Added `SmesherID` and `Coinbase` fields to `struct Reward` for efficient look-up. Used this to complete the `AccountDataQuery` and `AccountDataStream` endpoints in the API code. 
2. Changed the database index for storing Rewards to a key of the form `reward_<coinbase>_<smesherID>_<layerID>` because multiple SmesherIDs could be associated with the same Coinbase. Added a secondary key for Rewards lookups by SmesherID. 
3. Changed the `accumulateRewards` function in `mesh/mesh.go` so to support handling multiple smesherIDs associated with a single Coinbase. Moved event reporting of the Rewards from the processor to the mesh code. 
4. Added tests to `mesh/meshdb_test.go` to test changes made. 
5. Implemented the API endpoints for `SmesherRewardStream` and `SmesherDataQuery` using the endpoints created.
6. Since we now have that there can be multiple `SmesherID`s associated with a particular `Coinbase`, changed the way the rewards are calculated. Now the `blockTotalReward` and `blockLayerReward` are multiplied by the number of blocks produced by a particular `Coinbase`-`SmesherID` pair. 
## Test Plan

Wrote multiple unit tests for both the APIs that I implemented, and the meshdb code that I changed. 
## TODO


api/node.go
```diff
@@ -55,6 +56,7 @@ type TxAPI interface {
 	GetBalance(types.Address) uint64
 	GetNonce(types.Address) uint64
 	GetAllAccounts() (*types.MultipleAccountsState, error)
+	GetRewardsBySmesherID(types.NodeID) ([]types.Reward, error)
```
lrettig, line 60: Maybe add a comment here indicating that the misalignment between `SmesherID` and `NodeID` will be fixed in #2269

common/types/block.go
```diff
@@ -80,6 +81,7 @@ func (l LayerID) Uint64() uint64 {
 func (l LayerID) Field() log.Field { return log.Uint64("layer_id", uint64(l)) }
 
 // NodeID contains a miner's two public keys.
+// Sid : change the structs to fixed size byte arrays
```
lrettig: Remove this comment (or actually make the change 🙂)

common/types/block.go
```diff
@@ -105,8 +111,18 @@ func (id NodeID) ShortString() string {
 	return Shorten(name, 5)
 }
 
+// BytesToNodeID deserializes a byte slice into a NodeID
+func BytesToNodeID(b []byte) NodeID {
```
lrettig: Can you specify fixed-length input here?

common/types/block.go
```diff
@@ -105,8 +111,18 @@ func (id NodeID) ShortString() string {
...
+	vrfKey := b[32:]
+	return NodeID{
+		Key:          util.Bytes2Hex(pubKey),
+		VRFPublicKey: vrfKey,
+	}
+}
+
 // Field returns a log field. Implements the LoggableField interface.
-func (id NodeID) Field() log.Field { return log.String("node_id", id.Key) }
+func (id NodeID) Field() log.Field { return log.String("node_id", string(id.Key[:])) }
```
lrettig: Why do you need to take a slice copy here? `id.Key` is already a string, right?

events/events.go
```diff
@@ -170,8 +170,9 @@ func (ValidTx) GetChannel() ChannelID {
 
 // RewardReceived signals reward has been received
 type RewardReceived struct {
-	Coinbase string
-	Amount   uint64
+	Coinbase  string
```
lrettig, line 173: Why don't we include Layer in this struct?

events/reporter.go
```diff
@@ -442,7 +445,7 @@ type Reward struct {
 	// TODO: We don't currently have a way to get these two.
```
lrettig: Remove this TODO and link as you're addressing the Smesher part. Please make sure there's an open issue to track the TODO for adding LayerComputed (or open one if there isn't already one), and link that one here instead.

mesh/reward_test.go
```diff
@@ -151,7 +152,7 @@ func createLayer(t testing.TB, mesh *Mesh, id types.LayerID, numOfBlocks, maxTra
 	for i := 0; i < numOfBlocks; i++ {
 		block1 := types.NewExistingBlock(id, []byte(rand.String(8)), nil)
 		nodeID := types.NodeID{Key: strconv.Itoa(i), VRFPublicKey: []byte("bbbbb")}
-		coinbase := types.HexToAddress(nodeID.Key)
+		coinbase := types.HexToAddress(string(nodeID.Key))
```
lrettig: Confused by this as well since I thought `nodeID.Key` is already a string.

api/grpcserver/globalstate_service.go
```diff
@@ -138,6 +140,7 @@ func (s GlobalStateService) AccountDataQuery(_ context.Context, in *pb.AccountDa
 					// TODO: There is currently no way to get this for a reward.
```
lrettig: Remove TODO

api/grpcserver/globalstate_service.go
```diff
@@ -184,13 +187,52 @@ func (s GlobalStateService) AccountDataQuery(_ context.Context, in *pb.AccountDa
 }
 
 // SmesherDataQuery returns historical info on smesher rewards
-func (s GlobalStateService) SmesherDataQuery(context.Context, *pb.SmesherDataQueryRequest) (*pb.SmesherDataQueryResponse, error) {
+func (s GlobalStateService) SmesherDataQuery(_ context.Context, in *pb.SmesherDataQueryRequest) (*pb.SmesherDataQueryResponse, error) {
 	log.Info("GRPC GlobalStateService.SmesherDataQuery")
 
 	// TODO: implement me! We don't currently have a way to read rewards per-smesher.
```
lrettig: Remove TODO

api/grpcserver/globalstate_service.go
```diff
@@ -184,13 +187,52 @@ func (s GlobalStateService) AccountDataQuery(_ context.Context, in *pb.AccountDa
...
+	if in.SmesherId == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "`Id` must be provided")
+	}
+	if in.SmesherId.Id == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "`Id.Id` must be provided")
+	}
+
+	smesherIDBytes := in.SmesherId.Id
+	smesherID := types.BytesToNodeID(smesherIDBytes)
+	res := &pb.SmesherDataQueryResponse{}
```
lrettig: You can use [named return variables](https://yourbasic.org/golang/named-return-values-parameters/) here instead, which may be a little cleaner than instantiating another new object

api/grpcserver/globalstate_service.go
```diff
@@ -184,13 +187,52 @@ func (s GlobalStateService) AccountDataQuery(_ context.Context, in *pb.AccountDa
...
+	dbRewards, err := s.Mesh.GetRewardsBySmesherID(smesherID)
+
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting rewards data")
```
lrettig, line 210: Best practice here is to return a generic error message (as you're doing) but to print a more comprehensive `log.Error` message just before this, for debugging - please include the `smesherID` in the internal error message.

api/grpcserver/globalstate_service.go
```diff
@@ -184,13 +187,52 @@ func (s GlobalStateService) AccountDataQuery(_ context.Context, in *pb.AccountDa
...
+	}
+
+	res.TotalResults = uint32(len(res.Rewards))
+	offset := in.Offset
+
+	if offset > uint32(len(res.Rewards)) {
+		return &pb.SmesherDataQueryResponse{}, nil
+	}
+
+	maxResults := in.MaxResults
```
lrettig, line 240: This logic is a bit hard to understand without context or explanation (e.g., the fact that `maxResults==0` is equivalent to infinite results). And it's copied from another method. Could you turn this into a function, including the explanatory comments, and just call that here instead?

api/grpcserver/globalstate_service.go
```diff
@@ -278,7 +320,7 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 						Coinbase: &pb.AccountId{Address: addr.Bytes()},
 						// TODO: There is currently no way to get this for a reward.
```
lrettig: Remove TODO

api/grpcserver/globalstate_service.go
```diff
@@ -320,13 +362,55 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 }
 
 // SmesherRewardStream exposes a stream of smesher rewards
-func (s GlobalStateService) SmesherRewardStream(*pb.SmesherRewardStreamRequest, pb.GlobalStateService_SmesherRewardStreamServer) error {
+func (s GlobalStateService) SmesherRewardStream(in *pb.SmesherRewardStreamRequest, stream pb.GlobalStateService_SmesherRewardStreamServer) error {
 	log.Info("GRPC GlobalStateService.SmesherRewardStream")
 
 	// TODO: implement me! We don't currently have a way to read rewards per-smesher.
```
lrettig: Remove TODO

api/grpcserver/globalstate_service.go
```diff
@@ -417,7 +501,7 @@ func (s GlobalStateService) GlobalStateStream(in *pb.GlobalStateStreamRequest, s
 					Coinbase: &pb.AccountId{Address: reward.Coinbase.Bytes()},
 					// TODO: There is currently no way to get this for a reward.
```
lrettig: Remove TODO

api/grpcserver/grpcserver_test.go
```diff
@@ -237,6 +241,18 @@ func (t *TxAPIMock) GetRewards(types.Address) (rewards []types.Reward, err error
 			Layer:               layerFirst,
 			TotalReward:         rewardAmount,
 			LayerRewardEstimate: rewardAmount,
+			SmesherID:           nodeID,
```
lrettig, line 240: shouldn't this be `smesherID`?

api/grpcserver/grpcserver_test.go
```diff
@@ -237,6 +241,18 @@ func (t *TxAPIMock) GetRewards(types.Address) (rewards []types.Reward, err error
...
+	}, nil
+}
+
+func (t *TxAPIMock) GetRewardsBySmesherID(types.NodeID) (rewards []types.Reward, err error) {
+	return []types.Reward{
+		{
+			Layer:               layerFirst,
+			TotalReward:         rewardAmount,
+			LayerRewardEstimate: rewardAmount,
+			SmesherID:           nodeID,
```
lrettig, line 252: shouldn't this be `smesherID`?

mesh/mesh.go
```diff
@@ -754,11 +758,14 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
 			msh.With().Warning("Atx from block not found in db", log.Err(err), bl.ID(), bl.ATXID)
 			continue
 		}
-		ids = append(ids, atx.Coinbase)
-
+		coinbases = append(coinbases, atx.Coinbase)
+		//Sid: add a mapping between accounts and nodeID, pass this into the function
+		if _, exists := smeshers[atx.Coinbase]; !exists {
```
lrettig: Unfortunately this won't work since one coinbase may be associated with multiple smeshers (although the _reverse_ won't happen in a single layer). Use parallel arrays, or a struct, or some other solution to store a smesherID/nodeID for each block/coinbase address.

mesh/meshdb.go
```diff
@@ -362,6 +363,18 @@ func getRewardKeyPrefix(account types.Address) []byte {
 	return []byte(str)
 }
 
+// adding functions for getting the reward key for a smesherID
+// format for the index "reward_<smesherid>_<layerid> -> reward_<accountid>_<layerid> -> the actual reward"
+func getSmesherRewardKey(l types.LayerID, smesherID types.NodeID) []byte {
+	str := string(getSmesherRewardKeyPrefix(smesherID)) + "_" + strconv.FormatUint(l.Uint64(), 10)
```
lrettig: I think you need to account for endianness here to ensure sort works across platforms, using `binary.BigEndian.PutUint*`. See Noam's comment here: https://github.com/spacemeshos/go-spacemesh/pull/2176#issuecomment-766951415

mesh/meshdb.go
```diff
@@ -362,6 +363,18 @@ func getRewardKeyPrefix(account types.Address) []byte {
...
+	return []byte(str)
+}
+
+func getSmesherRewardKeyPrefix(smesherID types.NodeID) []byte {
+	str := "reward_" + smesherID.String()
```
lrettig: Probably more efficient to use `binary.BigEndian.PutUint*` here as well

Also, this seems to conflict with what `getRewardKeyPrefix` does, above - do we need two separate methods?

mesh/meshdb.go
```diff
@@ -443,25 +456,31 @@ func (m *DB) WriteTransaction(l types.LayerID, t *types.Transaction) error {
...
-	for account, cnt := range actBlockCnt {
-		reward := dbReward{TotalReward: cnt * totalReward.Uint64(), LayerRewardEstimate: cnt * layerReward.Uint64()}
+	for account, cnt := range accountBlockCount {
+		smesherID := smeshers[account]
+		reward := dbReward{TotalReward: cnt * totalReward.Uint64(), LayerRewardEstimate: cnt * layerReward.Uint64(), SmesherID: smesherID, Coinbase: account}
 		if b, err := types.InterfaceToBytes(&reward); err != nil {
 			return fmt.Errorf("could not marshal reward for %v: %v", account.Short(), err)
 		} else if err := batch.Put(getRewardKey(l, account), b); err != nil {
 			return fmt.Errorf("could not write reward to %v to database: %v", account.Short(), err)
+		} else if err := batch.Put(getSmesherRewardKey(l, smesherID), getRewardKey(l, account)); err != nil {
```
lrettig: See note above about `getRewardKeyPrefix` conflicting with `getSmesherRewardKey`. I don't think you need both here.

mesh/meshdb.go
```diff
@@ -443,25 +456,31 @@ func (m *DB) WriteTransaction(l types.LayerID, t *types.Transaction) error {
...
+			return fmt.Errorf("could not write smesherID to %v to datavase: %v", smesherID.ShortString(), err)
```
lrettig:
```suggestion
			return fmt.Errorf("could not write reward key for smesherID %v to database: %v", smesherID.ShortString(), err)
```

mesh/meshdb.go
```diff
@@ -489,6 +508,43 @@ func (m *DB) GetRewards(account types.Address) (rewards []types.Reward, err erro
...
+	it := m.transactions.Find(getSmesherRewardKeyPrefix(smesherID))
+	for it.Next() {
+		if it.Key() == nil {
+			break
+		}
+		str := string(it.Key())
+		strs := strings.Split(str, "_")
+		layer, err := strconv.ParseUint(strs[2], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("wrong key in db %s: %v", it.Key(), err)
```
lrettig:
```suggestion
			return nil, fmt.Errorf("error parsing db key %s: %v", it.Key(), err)
```

mesh/meshdb.go
```diff
@@ -489,6 +508,43 @@ func (m *DB) GetRewards(account types.Address) (rewards []types.Reward, err erro
...
+		}
+		//find the key to the actual reward, which is in it.Value()
+		var reward dbReward
+		rewardBytes, err := m.transactions.Get(it.Value())
+
+		if err != nil {
+			return nil, fmt.Errorf("wrong key in db %s: %v", it.Value(), err)
+		}
+		err = types.BytesToInterface(rewardBytes, &reward)
```
lrettig: You can combine this line and the next line using the `if err = foo; err != nil {` syntax

mesh/meshdb_test.go
```diff
@@ -478,33 +479,237 @@ func getTxns(r *require.Assertions, mdb *DB, origin types.Address) []TinyTx {
...
+	}, rewards)
+
+	rewards, err = mdb.GetRewardsBySmesherID(smesher1)
+	r.NoError(err)
+	r.Equal([]types.Reward{
+		{Layer: 1, TotalReward: 10000, LayerRewardEstimate: 9000, SmesherID: smesher1, Coinbase: addr1},
+		{Layer: 2, TotalReward: 20000, LayerRewardEstimate: 19000, SmesherID: smesher1, Coinbase: addr1},
+	}, rewards)
+
+	rewards, err = mdb.GetRewards(addr4)
```
lrettig: Shouldn't this call `GetRewardsBySmesherID`?

lrettig (changes requested):
Good work! See comments - there are one or two issues to resolve, also a bunch of comments and commented-out code to cleanup (I didn't highlight every case). Please fill in the initial comment in the PR (list of changes, test plan, etc.).

common/types/block.go
sidsankhe: I think that would be better left for the PR on resolving the NodeID type, because currently a lot of the tests use arbitrary length VRF keys so the length of a NodeID isn't fixed in the tests. What do you think?

api/grpcserver/globalstate_service.go
sidsankhe: Is the main utility of a function call here reducing duplication of code? One of the comments mentions that generics are not supported in Go, and so do we still need to add a new function?

api/grpcserver/grpcserver_test.go
sidsankhe: I removed smesherID, it was redundant!

mesh/mesh.go
sidsankhe: Resolving this through our discussion

mesh/meshdb.go
sidsankhe: I added two separate methods so that we could do a prefix lookup on the SmesherRewardKey (which is a secondary key in this abstraction) to obtain all the rewards for a given Smesher

api/grpcserver/globalstate_service.go
sidsankhe: Added the comments in based on our discussion.

events/events.go
sidsankhe: The only place this RewardReceived struct is used is for the Pubsub. It did not contain a layer before, so I didn't change it. Since this is not really used in the DB, is the layer needed at all?

mesh/meshdb.go
sidsankhe: Currently the key is stored as a byte representation of a string where the layerId is being written as a string representation of a base 10 number. How does endianness come into play here, since each integer character will take only 1 byte here.

events/events.go
lrettig: Okay. No, it's fine, we're not using pubsub in production for anything right now. It's confusing that we have multiple structs but let's leave this alone for now.

api/grpcserver/globalstate_service.go
```diff
@@ -184,13 +185,58 @@ func (s GlobalStateService) AccountDataQuery(_ context.Context, in *pb.AccountDa
...
+		return nil, status.Errorf(codes.InvalidArgument, "`Id` must be provided")
+	}
+	if in.SmesherId.Id == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "`Id.Id` must be provided")
+	}
+
+	smesherIDBytes := in.SmesherId.Id
+	smesherID := types.BytesToNodeID(smesherIDBytes)
+	dbRewards, err := s.Mesh.GetRewardsBySmesherID(smesherID)
+	res := &pb.SmesherDataQueryResponse{}
```
lrettig: As elsewhere, you could just name the return var in the function signature and leave this line out completely

api/grpcserver/globalstate_service.go
```diff
@@ -184,13 +185,58 @@ func (s GlobalStateService) AccountDataQuery(_ context.Context, in *pb.AccountDa
...
+
+	if err != nil {
+		log.Error("unable to fetch projected reward state for smesher %v: %s", smesherID, err)
```
lrettig: please use the `log.With()` syntax here, we can discuss on a call if this is unclear

```suggestion
		log.With().Error("unable to fetch projected reward state for smesher", log.String("smesher", smesherID), log.Error(err))
```

common/types/block.go
```diff
@@ -105,6 +106,16 @@ func (id NodeID) ShortString() string {
 	return Shorten(name, 5)
 }
 
+// BytesToNodeID deserializes a byte slice into a NodeID
+func BytesToNodeID(b []byte) NodeID {
```
lrettig: What happens if the input is the wrong length? Too long/too short/empty? Do you need error handling here?

mesh/mesh.go
```diff
@@ -742,8 +744,20 @@ func (msh *Mesh) GetOrphanBlocksBefore(l types.LayerID) ([]types.BlockID, error)
...
+			return true
+		}
+	}
+	return false
+}
+
 func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
-	ids := make([]types.Address, 0, len(l.Blocks()))
+	coinbases := make([]types.Address, 0, len(l.Blocks()))
+	coinbasesAndSmeshers := make(map[types.Address]map[string]uint64)
```
lrettig, line 761: why use a defined type (`types.Address`) for the first dimension and a string for the second dimension (rather than the NodeID type)?

mesh/mesh.go
```diff
@@ -754,11 +768,25 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
 			msh.With().Warning("Atx from block not found in db", log.Err(err), bl.ID(), bl.ATXID)
 			continue
 		}
-		ids = append(ids, atx.Coinbase)
-
+		coinbases = append(coinbases, atx.Coinbase)
+		//create a map where we count the number of coinbase-smesherID pairs appearing
```
lrettig: Please make this comment a bit clearer. Why are there two separate loops/maps here?

mesh/mesh.go
```diff
@@ -754,11 +768,25 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
...
+			coinbasesAndSmeshers[atx.Coinbase] = make(map[string]uint64)
+			coinbasesAndSmeshers[atx.Coinbase][atx.NodeID.String()]++
+		} else {
+			coinbasesAndSmeshers[atx.Coinbase][atx.NodeID.String()]++
+		}
+		if _, exists := smeshers[atx.Coinbase]; !exists {
+			smeshers[atx.Coinbase] = make([]types.NodeID, 0)
+			smeshers[atx.Coinbase] = append(smeshers[atx.Coinbase], atx.NodeID)
+		} else {
+			if !checkIfPresent(atx.NodeID, smeshers[atx.Coinbase]) {
```
lrettig: You're reimplementing set arithmetic. Consider using a set! https://pkg.go.dev/github.com/golang-collections/collections/set#Set.Insert

mesh/mesh.go
```diff
@@ -754,11 +768,25 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
...
+				smeshers[atx.Coinbase] = append(smeshers[atx.Coinbase], atx.NodeID)
+			}
+		}
 	}
 
-	if len(ids) == 0 {
+	if len(coinbasesAndSmeshers) == 0 {
```
lrettig: Why does this use `coinbasesAndSmeshers` but a few lines below you get `numBlocks` from `coinbases`?

mesh/mesh.go
```diff
@@ -790,11 +819,27 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
 		log.Uint64("total_reward_remainder", blockTotalRewardMod.Uint64()),
 		log.Uint64("layer_reward_remainder", blockLayerRewardMod.Uint64()),
 	)
-	err := msh.writeTransactionRewards(l.Index(), ids, blockTotalReward, blockLayerReward)
+	//report the rewards for each coinbase and each smesherID within each coinbase
+	//this can be thought off as a partition of the reward amongst all the smesherIDs
```
lrettig:
```suggestion
	//this can be thought of as a partition of the reward amongst all the smesherIDs
```

mesh/mesh.go
```diff
@@ -790,11 +819,27 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
...
+	//that added the Coinbase into the block
+	for account, smesherAccountEntry := range coinbasesAndSmeshers {
```
lrettig, line 825: I think this logic is wrong. The first map is a map from coinbase -> smesher -> number of blocks. This should be all the data you need to aggregate and report rewards. Why do you need to reference `smeshers` at all here? Can't you just loop through `coinbasesAndSmeshers` and report one reward for each (coinbase, smesher) tuple?

mesh/meshdb.go
```diff
@@ -352,13 +354,31 @@ func (m *DB) getLayerMutex(index types.LayerID) *layerMutex {
 	return ll
 }
 
-func getRewardKey(l types.LayerID, account types.Address) []byte {
-	str := string(getRewardKeyPrefix(account)) + "_" + strconv.FormatUint(l.Uint64(), 10)
+// Schema: "reward_<coinbase>_<smesherId>_<layerId> -> reward struct"
+func getRewardKey(l types.LayerID, account types.Address, smesherID types.NodeID) []byte {
+	byteArray := make([]byte, 8)
+	binary.BigEndian.PutUint64(byteArray, l.Uint64())
```
lrettig: looks like `byteArray` is unused here?

mesh/meshdb.go
```diff
@@ -352,13 +354,31 @@ func (m *DB) getLayerMutex(index types.LayerID) *layerMutex {
 	return ll
 }
 
-func getRewardKey(l types.LayerID, account types.Address) []byte {
-	str := string(getRewardKeyPrefix(account)) + "_" + strconv.FormatUint(l.Uint64(), 10)
+// Schema: "reward_<coinbase>_<smesherId>_<layerId> -> reward struct"
```
lrettig:
```suggestion
// Schema: "r_<coinbase>_<smesherId>_<layerId> -> reward struct"
```

mesh/meshdb.go
```diff
@@ -352,13 +354,31 @@ func (m *DB) getLayerMutex(index types.LayerID) *layerMutex {
...
 }
 
 func getRewardKeyPrefix(account types.Address) []byte {
-	str := "reward_" + account.String()
+	str := "r_" + account.String()
+	return []byte(str)
+}
+
+// adding functions for getting the reward key for a smesherID
+// format for the index "r_<smesherid>_<accountid>_<layerid> -> s_<accountid>_<smesherid>_<layerid> -> the actual reward"
```
lrettig:
```suggestion
// format for the index "s_<smesherid>_<accountid>_<layerid> -> r_<accountid>_<smesherid>_<layerid> -> the actual reward"
```

mesh/meshdb.go
```diff
@@ -443,25 +463,29 @@ func (m *DB) WriteTransaction(l types.LayerID, t *types.Transaction) error {
 	return nil
 }
 
+//Reason we're not using the existing reward type because the layer is implicit in the key
```
lrettig:
```suggestion
// We're not using the existing reward type because the layer is implicit in the key
```

mesh/meshdb.go
```diff
@@ -443,25 +463,29 @@ func (m *DB) WriteTransaction(l types.LayerID, t *types.Transaction) error {
...
+	Coinbase            types.Address
 	// TotalReward - LayerRewardEstimate = FeesEstimate
 }
 
-func (m *DB) writeTransactionRewards(l types.LayerID, accounts []types.Address, totalReward, layerReward *big.Int) error {
-	actBlockCnt := make(map[types.Address]uint64)
-	for _, account := range accounts {
-		actBlockCnt[account]++
-	}
+func (m *DB) writeTransactionRewards(l types.LayerID, accountBlockCount map[types.Address]map[string]uint64, totalReward, layerReward *big.Int, smeshers map[types.Address][]types.NodeID) error {
```
lrettig: Please split this into two lines

lrettig (changes requested):
I wasn't sure whether this PR is ready for review or not. There's still some lingering "to do" comments and stuff that needs to be cleaned up. I added a few more comments here. Let's discuss on Monday!

api/grpcserver/globalstate_service.go
sidsankhe: I tried that, but I believe named pointers get automatically initialized to the zero value, which is `nil`. I was thus getting segfaults, and so I reverted back to using non-named return params with explicit initialization

mesh/mesh.go
sidsankhe: I tried using a types.NodeID for the second dimension, but I was getting the following error: `invalid map key type types.NodeID`

mesh/mesh.go
sidsankhe: The code is functionally correct because of the invariant, but I agree that its pretty unclean. I think with the new map, we can do away with the `smeshers` map altogether, since we only really need that when obtaining the array of `SmesherID`s for a given `Coinbase`. I think I'll remove it from the code

common/types/block.go
sidsankhe: Yes, I'll add that in. What are current best practices for the return value of erroneous functions?

lrettig: Add `err` as a second return value, then:

> 			return nil, fmt.Errorf(errmsg)

mesh/mesh.go
sidsankhe: Resolved by adding the deserialization method for the `NodeID`

mesh/mesh.go
sidsankhe: Made the code cleaner

mesh/mesh.go
```diff
@@ -769,21 +768,14 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
 			continue
 		}
 		coinbases = append(coinbases, atx.Coinbase)
-		//create a map where we count the number of coinbase-smesherID pairs appearing
+		//create a 2 dimensional map where the entries are
+		//coinbasesAndSmeshers[coinbase_id][smesher_id] = number of blocks this pair has created
 		if _, exists := coinbasesAndSmeshers[atx.Coinbase]; !exists {
```
lrettig, line 775: You could clean this logic up a little further:

```
		if _, exists := coinbasesAndSmeshers[atx.Coinbase]; !exists {
			coinbasesAndSmeshers[atx.Coinbase] = make(map[string]uint64)
		}
		coinbasesAndSmeshers[atx.Coinbase][atx.NodeID.String()]++
```

mesh/mesh.go
```diff
@@ -754,11 +767,18 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
...
 			continue
 		}
-		ids = append(ids, atx.Coinbase)
-
+		coinbases = append(coinbases, atx.Coinbase)
+		//create a 2 dimensional map where the entries are
+		//coinbasesAndSmeshers[coinbase_id][smesher_id] = number of blocks this pair has created
+		if _, exists := coinbasesAndSmeshers[atx.Coinbase]; !exists {
+			coinbasesAndSmeshers[atx.Coinbase] = make(map[string]uint64)
+			coinbasesAndSmeshers[atx.Coinbase][atx.NodeID.String()]++
```
lrettig: I still find this counterintuitive. Could you add a comment here explaining why `atx.Coinbase` can be used as a map key but `atx.NodeID` cannot, and why you need to turn it into a string here?

common/types/block.go
lrettig: I'm not sure I agree with the way you're handling input that's too short here. It seems counterintuitive to ignore what's passed in, print/return no error, and just set an empty NodeID. It should probably be an error. Also, you're handling input that's too short but not input that's too long.

common/types/block.go
```diff
@@ -105,6 +106,42 @@ func (id NodeID) ShortString() string {
...
+		VRFPublicKey: vrfKey,
+	}
+}
+
+//StringToNodeID deserializes a string into a NodeID
+func StringToNodeID(s string) (NodeID, error) {
+	strLen := len(s)
+	if strLen < 64 {
+		var emptySlice []byte
+		return NodeID{
```
lrettig: Why not just `return nil, err`?

common/types/block.go
```diff
@@ -105,6 +106,42 @@ func (id NodeID) ShortString() string {
...
+	return NodeID{
+		Key:          util.Bytes2Hex(pubKey),
+		VRFPublicKey: vrfKey,
+	}
+}
+
+//StringToNodeID deserializes a string into a NodeID
+func StringToNodeID(s string) (NodeID, error) {
+	strLen := len(s)
+	if strLen < 64 {
```
lrettig, line 133: Need to handle the case where input is too long as well

common/types/block_test.go
```diff
@@ -37,3 +40,19 @@ func TestFields(t *testing.T) {
 	b.ActiveSet = &[]ATXID{atx1, atx2, atx3}
 	log.With().Info("got new block", b.Fields()...)
 }
+
+func TestStringToNodeID(t *testing.T) {
```
lrettig, line 43: Please add test cases for input that's too long/short as well

mesh/mesh.go
```diff
@@ -210,6 +211,7 @@ func (msh *Mesh) SetLatestLayer(idx types.LayerID) {
 		events.ReportNewLayer(events.NewLayer{
 			Layer:  layer,
 			Status: events.LayerStatusTypeUnknown,
+			//Rewards: list of reward type
```
lrettig: What's the purpose of this comment?

mesh/mesh.go
```diff
@@ -742,8 +744,19 @@ func (msh *Mesh) GetOrphanBlocksBefore(l types.LayerID) ([]types.BlockID, error)
 	return idArr, nil
 }
 
+func checkIfPresent(target types.NodeID, container []types.NodeID) bool {
```
lrettig: Is this still being used?

mesh/meshdb.go
```diff
@@ -443,25 +458,29 @@ func (m *DB) WriteTransaction(l types.LayerID, t *types.Transaction) error {
...
+	for account, smesherAccountEntry := range accountBlockCount {
+		for smesherString, cnt := range smesherAccountEntry {
+			smesherEntry, _ := types.StringToNodeID(smesherString)
+			reward := dbReward{TotalReward: cnt * totalReward.Uint64(), LayerRewardEstimate: cnt * layerReward.Uint64(), SmesherID: smesherEntry, Coinbase: account}
+			if b, err := types.InterfaceToBytes(&reward); err != nil {
+				return fmt.Errorf("could not marshal reward for %v: %v", account.Short(), err)
+			} else if err := batch.Put(getRewardKey(l, account, smesherEntry), b); err != nil {
+				return fmt.Errorf("could not write reward to %v to database: %v", account.Short(), err)
+			} else if err := batch.Put(getSmesherRewardKey(l, smesherEntry, account), getRewardKey(l, account, smesherEntry)); err != nil {
+				return fmt.Errorf("could not write reweard key for smesherID %v to datavase: %v", smesherEntry.ShortString(), err)
```
lrettig:
```suggestion
				return fmt.Errorf("could not write reward key for smesherID %v to database: %v", smesherEntry.ShortString(), err)
```

lrettig (changes requested):
Getting closer! See comments and please also resolve merge conflicts with develop branch

common/types/block.go
sidsankhe: The `nil` was giving compile-time errors because it isn't of type NodeID, and can't be cast to a NodeID as well, so I just decided to return an empty NodeID struct

api/grpcserver/globalstate_service.go
lrettig: as discussed on our last call, try using pointers instead?

common/types/block.go
```diff
@@ -105,6 +106,40 @@ func (id NodeID) ShortString() string {
 	return Shorten(name, 5)
 }
 
+// BytesToNodeID deserializes a byte slice into a NodeID
+func BytesToNodeID(b []byte) (*NodeID, error) {
+	if len(b) < 32 {
```
lrettig, line 113: Why do we allow a range of lengths (32-64 bytes) here, rather than requiring a fixed length?

common/types/block.go
```diff
@@ -105,6 +106,40 @@ func (id NodeID) ShortString() string {
...
+	return &NodeID{
+		Key:          util.Bytes2Hex(pubKey),
+		VRFPublicKey: vrfKey,
+	}, nil
+}
+
+//StringToNodeID deserializes a string into a NodeID
+func StringToNodeID(s string) (*NodeID, error) {
+	strLen := len(s)
+	if strLen < 64 {
```
lrettig, line 133: Same question as above

lrettig (commented):
I cleaned up the tests a little bit. See one question I posted.

lrettig:
This rewritten test is legitimately failing:

```
2021-02-18T17:49:58.9146029Z === RUN   TestBytesToNodeID
2021-02-18T17:49:58.9147078Z     block_test.go:82: 
2021-02-18T17:49:58.9148130Z         	Error Trace:	block_test.go:82
2021-02-18T17:49:58.9149166Z         	Error:      	Not equal: 
2021-02-18T17:49:58.9150175Z         	            	expected: []byte{0x32, 0x32, 0x32, 0x32, 0x32}
2021-02-18T17:49:58.9151207Z         	            	actual  : []byte{0x22, 0x22}
2021-02-18T17:49:58.9152016Z         	            	
2021-02-18T17:49:58.9152770Z         	            	Diff:
2021-02-18T17:49:58.9153929Z         	            	--- Expected
2021-02-18T17:49:58.9154771Z         	            	+++ Actual
2021-02-18T17:49:58.9155792Z         	            	@@ -1,3 +1,3 @@
2021-02-18T17:49:58.9156927Z         	            	-([]uint8) (len=5) {
2021-02-18T17:49:58.9158402Z         	            	- 00000000  32 32 32 32 32                                    |22222|
2021-02-18T17:49:58.9159282Z         	            	+([]uint8) (len=2) {
2021-02-18T17:49:58.9160392Z         	            	+ 00000000  22 22                                             |""|
2021-02-18T17:49:58.9161060Z         	            	 }
2021-02-18T17:49:58.9161773Z         	Test:       	TestBytesToNodeID
2021-02-18T17:49:58.9162705Z         	Messages:   	NodeID VRF Key does not match
2021-02-18T17:49:58.9164027Z --- FAIL: TestBytesToNodeID (0.00s)
```

There's something wrong in the logic of `BytesToNodeID`. It has to do with the fact that the two keys are stored differently: one as a string, the other as a bytearray. These should really both be the same type. See if you can clean this up for now. Feel free to do something simple for now, and we can sort out the types properly in #2269.

api/grpcserver/grpcserver_test.go
```diff
@@ -2230,13 +2303,15 @@ func checkAccountDataQueryItemAccount(t *testing.T, dataItem interface{}) {
 	}
 }
 
+//Add a line here
```
lrettig: What's this?

api/grpcserver/grpcserver_test.go
```diff
@@ -688,18 +702,77 @@ func TestGlobalStateService(t *testing.T) {
...
 		{"SmesherDataQuery", func(t *testing.T) {
-			_, err := c.SmesherDataQuery(context.Background(), &pb.SmesherDataQueryRequest{})
-			require.Error(t, err)
-			statusCode := status.Code(err)
-			require.Equal(t, codes.Unimplemented, statusCode)
-		}},
-		{"SmesherRewardStream", func(t *testing.T) {
-			stream, err := c.SmesherRewardStream(context.Background(), &pb.SmesherRewardStreamRequest{})
-			// We expect to be able to open the stream but for it to fail upon the first request
+			res, err := c.SmesherDataQuery(context.Background(), &pb.SmesherDataQueryRequest{
```
lrettig, line 719: Please add tests for `SmesherDataQuery` without required args (`Id`, `Id.Id`, as below)

common/types/block.go
```diff
@@ -105,6 +106,41 @@ func (id NodeID) ShortString() string {
 	return Shorten(name, 5)
 }
 
+// BytesToNodeID deserializes a byte slice into a NodeID
+func BytesToNodeID(b []byte) (*NodeID, error) {
```
lrettig, line 112: Please add another note here that the length of the input here will be worked out when we finalize node vs. smesher ID (with a link to the same issue)

events/reporter.go
```diff
@@ -403,6 +405,7 @@ const (
 
 // NewLayer packages up a layer with its status (which a layer does not
 // ordinarily contain)
+// add it here as well
```
lrettig: What's this?

lrettig (changes requested):
Few more small things to tweak, see comments

lrettig (approved):
LGTM! Made a few small tweaks, this is good to go

sidsankhe:
bors merge

Issue #1019, opened 2019-06-20
Blocks conflict in epoch beacon


barakshani:
# Overview / Motivation
Validity to produce blocks is based (among others) on the Epoch Beacon.
This beacon is computed locally and should be the same for all nodes. Its value is published in each block.
We need to address the case when different blocks (of different miners) have different beacons.

# The Task
For blocks in the **same** epoch,
if layer(x) > layer(b), and b is in the view of x,
if the epoch beacon is different,
when when counting votes on b, we consider the vote of x to have weight 0.

almogdepaz:
@barakshani milestone ?

barakshani:
next one, where we finish all protocol tasks

antonlerner:
@nkryuchkov close this if you address it in your pr

PR #2267, opened 2021-01-24
[Merged by Bors] - Tortoise Beacon

### LINKED ISSUE #29: Tortoise Beacon
Research forum thread: https://community.spacemesh.io/t/tortoise-beacon-based-on-bad-vrf/100
SMIP: https://github.com/spacemeshos/SMIPS/issues/29
Issue: https://github.com/spacemeshos/go-spacemesh/issues/2249
PR: https://github.com/spacemeshos/go-spacemesh/pull/2267

- [X] Implement ticker
- [X] Implement message exchanger
- [X] Implement message classifier 
- [X] Create initial Tortoise Beacon implementation based on https://github.com/spacemeshos/go-spacemesh/issues/2249
- [X] Adjust the implementation to https://community.spacemesh.io/t/tortoise-beacon-based-on-bad-vrf/100
- [X] Make changes discussed in the research call
- [X] Tie all components together and ensure that they work correctly
- [x] Implement communication between nodes
- [X] Unit tests
- [x] System tests
- [x] Fix bugs found by unit and system tests

### LINKED ISSUE #2249: Create timed ticker and message exchanger component
In order to implement the beacon we must agree on a set of ATXs to calculate the beacon from this set

the proposed solution is 

> The proposed solution for creating a beacon is as follows:
> 
> 1. Create a proposal for a beacon:
>    
>    * take all ATXs received in last epoch (i -1)
>    * concat them into a single proposal message and sign it
>    * Each smesher partitions the valid proposals received in the previous epoch into three sets:
>      
>      * Timely proposals: received up to δ after the end of the previous epoch.
>      * Delayed proposals: received between δ and 2δ after the end of the previous epoch.
>      * Late proposals: more than 2δ after the end of the previous epoch. Note that honest users cannot disagree on timing by more than δ, so if a proposal is timely for any honest user, it cannot be late for any honest user (and vice versa).
>    * A party’s initial input to the consensus algorithm is the set of timely proposals.
> 2. Reach consensus over set of ATXs to construct a beacon from
>    
>    * for K rounds : in each round that lasts δ, wait for proposals to come in
>    * create a voting message that references all seen proposals within δ time frame and send it
>    * for next rounds wait for δ time, and construct a message that points to all messages from previous round received by δ
>    * After K rounds had passed, tally up votes for proposals using simple tortoise vote counting
> 3. Create random beacon coin
>    
>    * for all messages that passed the global voting threshold, take the ATXs they have pointer on
>    * hash all ATX into a single value
>    * Beacon value received
https://github.com/spacemeshos/SMIPS/issues/29

for this solution we need to have a generic protocol ticker capable of sending an inital message and following messages - aslo classifying them into "timely" and "late"

nkryuchkov:
## Motivation

Closes #2249 
Closes https://github.com/spacemeshos/pm/issues/29

## Changes

- Implement Tortoise Beacon

## Test Plan

To be tested when finished

## TODO

- [x] Finish the implementation
- [x] Implement unit tests
- [x] Implement system tests


tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+
+	config Config
+
+	closed chan struct{} // closed if TortoiseBeacon is Close()'d.
+
+	messageReceiver  messageReceiver
+	messageSender    messageSender
+	beaconCalculator beaconCalculator
+
+	atxDB *activation.DB
```
antonlerner: can be an interface as well

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+	tb.beaconsMu.RLock()
+	beacon, ok := tb.beacons[epochID]
+	tb.beaconsMu.RUnlock()
+
+	return beacon, ok
+}
+
+func (tb *TortoiseBeacon) WaitAndGet(epochID types.EpochID) types.Hash32 {
+	tb.beaconsMu.RLock()
+	ch := tb.beaconsCh[epochID]
```
antonlerner: what happens if the key is not present?

antonlerner: this also means that only one caller can WaitAndGet right?

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+	}()
+
+	go tb.cleanupLoop()
+
+	return nil
+}
+
+func (tb *TortoiseBeacon) Close() error {
+	tb.Log.Info("Closing %v", protoName)
+	close(tb.closed)
```
antonlerner: wait until all routines have shut down before returning from this function, or at least that `listenMessages` is done

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+
+	return nil
+}
+
+func (tb *TortoiseBeacon) Get(epochID types.EpochID) (types.Hash32, bool) {
+	tb.beaconsMu.RLock()
+	beacon, ok := tb.beacons[epochID]
+	tb.beaconsMu.RUnlock()
+
+	return beacon, ok
```
antonlerner: perhaps returning `(types.Hash32, error)` is more suitable

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+		select {
+		case <-tb.closed:
+			return
+		case <-ticker.C:
+			tb.cleanup()
+		}
+	}
+}
+
+func (tb *TortoiseBeacon) cleanup() {
```
antonlerner: I think that all beacons - once calculated should be stored persistently since we woudln't want to re calculate them when syncing

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+}
+
+func (tb *TortoiseBeacon) classifyMessage(m Message, epoch types.EpochID) MessageType {
+	tb.currentRoundsMu.Lock()
+	currentRound := tb.currentRounds[epoch]
+	tb.currentRoundsMu.Unlock()
+
+	round := 0
+	if m, ok := m.(VotingMessage); ok {
```
antonlerner: if not `VotingMessage` ?

nkryuchkov: @antonlerner Then `round` is `0`

tortoisebeacon/tortoisebeacon.go
nkryuchkov: @antonlerner I need to return an error If the key is not present. Right, only one caller can `WaitAndGet`. Should many be able to call it?

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+// for K rounds : in each round that lasts δ, wait for proposals to come in
+func (tb *TortoiseBeacon) roundTicker(epoch types.EpochID) {
+	ticker := time.NewTicker(tb.networkDelta)
+	defer ticker.Stop()
+
+	tb.currentRoundsMu.Lock()
+	tb.currentRounds[epoch] = 0
+	tb.currentRoundsMu.Unlock()
+
+	// round 0 already happened at this point, starting from round 1
```
antonlerner: what should happen in round 0?

nkryuchkov: @antonlerner 
1. take all ATXs received in last epoch (i -1)
2. concat them into a single proposal message and sign it
3. send the proposal message

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+				// create a voting message that references all seen proposals within δ time frame and send it
+				tb.timelyProposalsMu.RLock()
+				proposals := tb.timelyProposalsList[epoch]
+				tb.timelyProposalsMu.RUnlock()
+
+				for _, p := range proposals {
+					hash := hashATXList(p)
+
+					m := NewVotingMessage(epoch, round, hash)
+					tb.messageSender.Send(m)
```
antonlerner: this will send multiple messages - why?

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+					EpochID: epoch,
+					Round:   round - 1, // proposals from the previous round are needed
+				}
+
+				tb.timelyProposalsMu.Lock()
+				votes := tb.votes[key]
+				tb.timelyProposalsMu.Unlock()
+
+				for _, voteHash := range votes {
+					m := NewVotingMessage(epoch, round, voteHash)
```
antonlerner: this will send multuple messages as well
If I understand corectly - here you should construct a message that votes for all `votes` at once

tortoisebeacon/tortoisebeacon.go
nkryuchkov: Sending in one message is not pushed yet

tortoisebeacon/tortoisebeacon.go
nkryuchkov: @antonlerner Yes, that's correct. I will push this change

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,450 @@
...
+		for _, hash := range m.HashList() {
+			tb.votesMu.Lock()
+			tb.votes[pair] = append(tb.votes[pair], hash)
+			tb.votesMu.Unlock()
+		}
+
+		return tb.beaconCalculator.CountBatchVotes(m) // TODO(nkryuchkov): count votes properly
+
+	default:
+		// TODO(nkryuchkov): handle other types
```
antonlerner: log for now

tortoisebeacon/db.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+	"sync"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/database"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+// DB hold the atxs received from all nodes and their validity status
```
antonlerner: wrong comment?

nkryuchkov: Yes

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,652 @@
...
+func (tb *TortoiseBeacon) Close() error {
+	tb.Log.Info("Closing %v", protoName)
+	tb.Closer.Close()
+	tb.backgroundWG.Wait() // Wait until background goroutines finish
+
+	return nil
+}
+
+func (tb *TortoiseBeacon) Get(epochID types.EpochID) (types.Hash32, error) {
+	if tb.tortoiseBeaconDB != nil {
```
antonlerner: why would this be nil? isn't it initialised in constructor?

nkryuchkov: It is, but `nil` may be passed to the constructor. In this case, DB is not used. For example, we can use `nil` in tests, where we need to test parts that don't require DB.

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,652 @@
...
+		case <-tb.CloseChannel():
+			return
+		case <-ticker.C:
+			tb.cleanup()
+		}
+	}
+}
+
+func (tb *TortoiseBeacon) cleanup() {
+	// TODO(nkryuchkov): implement a better solution, consider https://github.com/golang/go/issues/20135
```
antonlerner, line 238: also cleaning of beacons maps must be done once in a while and not just when closing, since it will run for a long time and eventually contain all epoch beacons

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,652 @@
...
+	tb.beaconsReady[epoch] = make(chan struct{})
+	tb.beaconsMu.Unlock()
+
+	// round 0
+	// take all ATXs received in last epoch (i -1)
+	atxList := tb.epochATXGetter.GetEpochAtxs(epoch - 1)
+
+	// concat them into a single proposal message
+	m := NewProposalMessage(epoch, atxList)
+	tb.messageSender.Send(m)
```
antonlerner: since this message is sent through gossip - wouldn't the sender receive it as well?

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,652 @@
...
+				timelyProposals := tb.timelyProposalsList[epoch]
+				tb.timelyProposalsMu.RUnlock()
+
+				hashesFor := make([]types.Hash32, 0, len(timelyProposals))
+				for _, p := range timelyProposals {
+					hashesFor = append(hashesFor, hashATXList(p))
+				}
+
+				hashesAgainst := make([]types.Hash32, 0, len(timelyProposals))
+				for _, p := range timelyProposals {
```
antonlerner: you are iterating the timelyProposals again, As I see it, since you don't store late messages there is not need for `VotesAgains` at all, only vote `For` proposals so that we can reach consensus about what are the correct values for beacon

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,652 @@
...
+		}
+	}
+
+	tb.timelyProposalsMu.Unlock()
+
+	votesFor = make([]types.Hash32, 0)
+	votesAgainst = make([]types.Hash32, 0)
+
+	for k, v := range votesForMap {
+		if v > threshold() || tb.WeakCoin(epoch, round) {
```
antonlerner: you also vote yes for timely proposals that have been received I believe

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,652 @@
...
+			votesFor = append(votesFor, k)
+		}
+	}
+
+	for k, v := range votesAgainstMap {
+		if v > threshold() || !tb.WeakCoin(epoch, round) {
```
antonlerner: theoretically you need to vote against messages that came in late - but as said before I think we can drop this part

tortoisebeacon/tortoisecore.go
```diff
@@ -0,0 +1,72 @@
+package tortoisebeacon
+
+import "github.com/spacemeshos/go-spacemesh/common/types"
+
+const (
+	hDist = 1 // TODO: change
```
antonlerner: put in config

antonlerner:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

tortoisebeacon/beaconcalc.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+func (tb *TortoiseBeacon) calcBeacon(epoch types.EpochID) {
+	tb.Log.With().Info("Calculating beacon",
+		log.Uint64("epoch_id", uint64(epoch)))
+
+	allHashes := tb.calcTortoiseBeaconHashList(epoch)
+
+	stringHashes := make([]string, 0, len(allHashes))
```
antonlerner: you can refactor this to a `hashList.asStrings()`

tortoisebeacon/beaconcalc.go
```diff
@@ -0,0 +1,76 @@
...
+	allHashes := make(hashList, 0)
+
+	lastRound := epochRoundPair{
+		EpochID: epoch,
+		Round:   tb.lastRound(),
+	}
+
+	votes, ok := tb.ownVotes[lastRound]
+	if !ok {
+		_, _ = tb.calcVotesDelta(epoch, lastRound.Round)
```
antonlerner: why is the output ignored here?

tortoisebeacon/votescalc.go
```diff
@@ -0,0 +1,287 @@
...
+		log.String("against", fmt.Sprint(votesAgainst)))
+
+	return votesFor, votesAgainst
+}
+
+func (tb *TortoiseBeacon) calcVotesDelta(epoch types.EpochID, round types.RoundID) (forDiff, againstDiff hashList) {
+	tb.votesMu.Lock()
+	defer tb.votesMu.Unlock()
+
+	votesCount := tb.countFirstRoundVotes(epoch)
```
antonlerner: rename to `firstRoundVotes` or something

tortoisebeacon/beaconcalc.go
nkryuchkov: @antonlerner In this case, `calcVotesDelta` is used to re-calculate `tb.ownVotes`, so its output is not needed

tortoisebeacon/votescalc.go
```diff
@@ -0,0 +1,287 @@
...
+
+func (tb *TortoiseBeacon) countFirstRoundVotes(epoch types.EpochID) votesCountMap {
+	firstRoundInThisEpoch := epochRoundPair{
+		EpochID: epoch,
+		Round:   1,
+	}
+
+	tb.votesCache[firstRoundInThisEpoch] = make(map[p2pcrypto.PublicKey]votesSetPair)
+
+	firstRoundVotes := tb.incomingVotes[firstRoundInThisEpoch]
```
antonlerner: concurrent read can occur - this needs to be RLocked

tortoisebeacon/votescalc.go
```diff
@@ -0,0 +1,287 @@
...
+	return votesFor, votesAgainst
+}
+
+func (tb *TortoiseBeacon) countFirstRoundVotes(epoch types.EpochID) votesCountMap {
+	firstRoundInThisEpoch := epochRoundPair{
+		EpochID: epoch,
+		Round:   1,
+	}
+
+	tb.votesCache[firstRoundInThisEpoch] = make(map[p2pcrypto.PublicKey]votesSetPair)
```
antonlerner: this need to be RLocked as well, unless there is a global lock - at which case please add a comment about it

tortoisebeacon/votescalc.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+
+		// copy to cache
+		tb.votesCache[firstRoundInThisEpoch][pk] = votesSetPair{
+			VotesFor:     firstRoundVotesFor,
+			VotesAgainst: firstRoundVotesAgainst,
+		}
+	}
+
+	tb.votesCountCache[firstRoundInThisEpoch] = make(map[types.Hash32]int)
```
antonlerner: `votesCountCache` is not filled here

tortoisebeacon/votescalc.go
```diff
@@ -0,0 +1,287 @@
...
+	for k, v := range tb.votesCache[firstRoundInThisEpoch] {
```
antonlerner: why is this necessary? perhaps this should write into `votesCountCache` ?

tortoisebeacon/votescalc.go
```diff
@@ -0,0 +1,287 @@
...
+	return votesCount
+}
+
+func (tb *TortoiseBeacon) calcOwnFirstRoundVotes(epoch types.EpochID, votesCount votesCountMap) votesSetPair {
+	ownFirstRoundsVotes := votesSetPair{
+		VotesFor:     make(hashSet),
+		VotesAgainst: make(hashSet),
+	}
+
+	for vote, count := range votesCount {
```
antonlerner: rename `votesCount` to `firstRoundVotes` or other that will explain what is this vote count

tortoisebeacon/votescalc.go
```diff
@@ -0,0 +1,287 @@
...
+		switch {
+		case count >= tb.threshold():
+			ownFirstRoundsVotes.VotesFor[vote] = struct{}{}
+		case count <= -tb.threshold():
+			ownFirstRoundsVotes.VotesAgainst[vote] = struct{}{}
+		case tb.weakCoin.Get(epoch, 1):
```
antonlerner: isn't there another case which you vote `FOR` the most recent messages which has been received in time?

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,538 @@
...
+	}
+
+	tb.Log.With().Info("Handling epoch",
+		log.Uint64("epoch_id", uint64(epoch)))
+
+	tb.beaconsMu.Lock()
+	if _, ok := tb.seenEpochs[epoch]; ok {
+		tb.beaconsMu.Unlock()
+
+		// Already handling this epoch
```
antonlerner: this will probably happen a lot, but maybe log here

collector/collector.go
```diff
@@ -206,6 +213,17 @@ loop:
 			if err != nil {
 				log.Error("cannot write message %v", err)
 			}
+		case data := <-tortoiseBeacons:
+			var e events.TortoiseBeaconCalculated
+			err := types.BytesToInterface(data[size:], &e)
+			if err != nil {
+				log.Error("cannot parse received message %v", err)
```
antonlerner, line 220: wrong log

common/util/closer.go
```diff
@@ -0,0 +1,22 @@
+package util
+
+// Closer adds the ability to close objects.
+type Closer struct {
```
antonlerner, line 6: I'd implement an `IsClosed` getter - i think this is the main functionality in this util right?

tests/tortoise_beacon/test_tortoise_beacon.py
```diff
@@ -0,0 +1,64 @@
...
+    print(f"wait until next epoch to layer {last_layer}")
+    _ = q.wait_for_latest_layer(init_session, last_layer, layers_per_epoch, num_miners + 1)
+
+    # ========================== epoch i+4 ==========================
+    curr_epoch += 1
+    print("\n\n-------- current epoch", curr_epoch, "--------")
+
+    last_layer = layers_per_epoch * (curr_epoch + 2)
+    print(f"wait 2 epochs for layer {last_layer}")
+    _ = q.wait_for_latest_layer(init_session, last_layer, layers_per_epoch, num_miners + 1)
```
antonlerner, line 37: there is no tests happening between `wait_for_latest_layer` so why not just wait for the last layer of 4 epochs since start?

tests/utils.py
```diff
@@ -151,6 +153,27 @@ def validate_blocks_per_nodes(block_map, from_layer, to_layer, layers_per_epoch,
     print("\nvalidation succeeded!\n")
 
 
+def validate_beacons(log_messages):
+    epoch_messages = defaultdict(dict)
+
+    assert len(log_messages) > 0, f"no log messages"
+
+    for log in log_messages:
```
antonlerner, line 163: check if beacon is not 0 or null as well

tortoisebeacon/handlers.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+
+		tb.delayedProposals[epoch][m.Proposals().Hash()] = struct{}{}
+
+		tb.delayedProposalsMu.Unlock()
+
+		return nil
+
+	case LateMessage:
+		tb.Log.With().Info("Received late ProposalMessage",
```
antonlerner: I'd change to Warning - this is something we would potentially look for when debugging nodes

tortoisebeacon/handlers.go
```diff
@@ -0,0 +1,213 @@
...
+
+		tb.incomingVotes[thisRound][from] = votesSetPair{
+			VotesFor:     votesFor,
+			VotesAgainst: votesAgainst,
+		}
+
+		return nil
+
+	case DelayedMessage, LateMessage:
+		tb.Log.With().Info(fmt.Sprintf("Received %v VotingMessage, ignoring it", mt.String()),
```
antonlerner: same here - change log level to warning

tortoisebeacon/handlers.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+}
+
+func (tb *TortoiseBeacon) classifyMessage(m message, epoch types.EpochID) MessageType {
+	tb.currentRoundsMu.Lock()
+	currentRound := tb.currentRounds[epoch]
+	tb.currentRoundsMu.Unlock()
+
+	messageRound := types.RoundID(1)
+	if vm, ok := m.(VotingMessage); ok {
```
antonlerner: if not ok - log error

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,559 @@
...
+		}
+	}
+
+	return beacon, nil
+}
+
+// GetBeacon waits until a Tortoise Beacon value is ready for a certain epoch and returns it as []byte.
+func (tb *TortoiseBeacon) GetBeacon(epochNumber types.EpochID) []byte {
+	if err := tb.Wait(epochNumber); err != nil {
+		return nil
```
antonlerner: log or return error

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,559 @@
...
+
+	tb.beaconsMu.RLock()
+	ch, ok := tb.beaconsReady[epochID]
+	tb.beaconsMu.RUnlock()
+
+	if !ok {
+		return ErrBeaconNotCalculated
+	}
+
+	<-ch
```
antonlerner: could this be stuck forever? if so, maybe add a timeout

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,559 @@
...
+			log.Uint64("epoch_id", uint64(epoch)))
+
+		return
+	}
+
+	tb.Log.With().Info("Handling epoch",
+		log.Uint64("epoch_id", uint64(epoch)))
+
+	tb.beaconsMu.Lock()
+	if _, ok := tb.seenEpochs[epoch]; ok {
```
antonlerner: what happens when a miner joins the network in the middle of the epoch? will it start sending proposals? as I recall this consensus protocol should start at the beginning of the epoch, if not - wait for the start of the next one

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+
+	tb.waitAfterLastRoundStarted()
+
+	return nil
+}
+
+func (tb *TortoiseBeacon) runProposalPhase(epoch types.EpochID) error {
+	// take all ATXs received in last epoch (i - 1)
+	atxList := ATXIDList(tb.epochATXGetter.GetEpochAtxs(epoch - 1))
```
antonlerner: check if list is not empty - if so log or error

tortoisebeacon/tortoisebeacon.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+
+	return nil
+}
+
+func (tb *TortoiseBeacon) sendProposalVote(epoch types.EpochID) error {
+	tb.setCurrentRound(epoch, firstRound)
+	// round 1, send hashed proposal
+	// create a voting message that references all seen proposals within δ time frame and send it
+	votesFor, votesAgainst := tb.calcVotesFromProposals(epoch)
```
antonlerner: generally - when creating a hash from a list we must sort it beforehand so that for same values we'll get the same hash
I don't see an ordering issue with your code but please double check

tortoisebeacon/votescalc.go
nkryuchkov: It's locked by `tb.votesMu` outside of `firstRoundVotes` in `calcVotesDelta`, I've added a comment

tortoisebeacon/votescalc.go
nkryuchkov: It's locked by `tb.votesMu` in `calcVotesDelta`, I've added a comment

common/util/closer.go
nkryuchkov: I moved it from the `hare` package without any modifications because I need it as well. Actually, I just need to block until the channel is closed, so `CloseChannel` is enough for me, same in the `hare` package. I've added `IsClosed` in case it's needed in the future.

tortoisebeacon/handlers.go
nkryuchkov: `classifyMessage` can be used for both `ProposalMessage` and `VotingMessage`. 

For `ProposalMessage`, `messageRound == 1`. For `VotingMessage`, `messageRound` is extracted from message.

nkryuchkov:
bors try

nkryuchkov:
bors try

tortoisebeacon/tortoisebeacon.go
```diff
@@ -182,94 +182,34 @@ func (tb *TortoiseBeacon) Close() error {
 	return nil
 }
 
-// Get returns a Tortoise Beacon value as types.Hash32 for a certain epoch or an error if it doesn't exist.
-// TODO(nkryuchkov): Consider making unexported.
-func (tb *TortoiseBeacon) Get(epochID types.EpochID) (types.Hash32, error) {
+// Get returns a Tortoise Beacon value as []byte for a certain epoch or an error if it doesn't exist.
+func (tb *TortoiseBeacon) Get(layerID types.LayerID) ([]byte, error) {
```
antonlerner: why `[]byte` ?

tortoisebeacon/tortoisebeacon.go
```diff
@@ -182,94 +186,39 @@ func (tb *TortoiseBeacon) Close() error {
...
-}
-
-// Wait waits until beacon for this epoch becomes ready.
-func (tb *TortoiseBeacon) Wait(epochID types.EpochID) error {
-	tb.waitUntilStarted()
-
 	if tb.tortoiseBeaconDB != nil {
-		if _, ok := tb.tortoiseBeaconDB.GetTortoiseBeacon(epochID); ok {
-			return nil
+		if err := tb.tortoiseBeaconDB.SetTortoiseBeacon(epochID, beacon); err != nil {
```
antonlerner: every time you will get the beacon you will try to insert it to db? won't it cause an error every time since you put it in the first time?

nkryuchkov: If it already exists in a DB for a certain epoch, it's returned from the DB. So, it's put in a DB once every epoch.

tortoisebeacon/tortoisebeacon.go
nkryuchkov: It's easier to mock. However, I'm still in doubt whether it should be `types.Hash32` or `[]byte`

tortoisebeacon/tortoisebeacon.go
```diff
@@ -588,3 +530,22 @@ func (tb *TortoiseBeacon) votingThreshold() int {
 func (tb *TortoiseBeacon) atxThresholdFraction(totalWeight int) float64 {
 	return 1 - math.Pow(2.0, -(float64(tb.config.Kappa)/((1.0-tb.config.Q)*float64(totalWeight))))
 }
+
+// TODO(nkryuchkov): add unit tests
+func (tb *TortoiseBeacon) beaconAlreadyCalculated(layerID types.LayerID) bool {
```
antonlerner: rename - this function tells us if new beacon should be ready right? it doesn't guarantee it's ready

nkryuchkov: Yes, it does this. Should it also check for a beacon value explicitly?

nkryuchkov:
### 4 May research call changes:

- [x] Proposal messages should not be signed
- [x] Voting messages should be signed
- [x] `T_ave` config parameter is not needed, total ATX weight in the previous epoch should be used instead
- [x] Round should contain two sub-rounds:
    - [x] First is for sending voting messages
    - [x] Second is for sending weak coin messages
    - [x] Their lengths may be different
    - [x] Adversary should be forced to vote before a weak coin value is known: votes should be sent at the beginning of the round, weak coin should be sent after votes for this round are received
- [x] Weak coin messages should be sent starting from the second voting message (2nd round)
- [x] Count all voting messages within an epoch, no matter what round is; if message round is greater than current round, store it and wait for it, count it later
- [x] Two messages for the same round from one miner are invalid
- [x] Change logic of proposal message classification
    - [x] If a signature is invalid, the message is invalid
    - [x] When a proposal is received, check if the proposal sender has an ATX received within the previous epoch up to 30 seconds after the epoch ended
        - [x] If it doesn't exist or is received after 30 seconds after the epoch ended, the proposal is invalid
        - [x] If it is received between the end of the epoch and 30 seconds after, the proposal is potentially valid
        - [x] If it is received before the end of the epoch, it's valid
- [x] Check logic of voting message classification
    - [x] Count all voting messages received before we generated our vote
- [x] First and next voting messages should be different
    - [x] First message contains votes for, votes against, signature; epoch could be implicit
    - [x] Next messages contain bit vector, signature, round; epoch is implicit 
- [x] Instead of ATXs in the previous epoch, `VRF(prefix+epoch_id)` should be sent in a proposal message; it should be sent only if it passes a threshold
- [x] If a received proposal does not pass a threshold, it should be rejected
- [x] Proposal message should contain a pointer to ATX (ATX ID/miner's unique ID)
- [x] Epoch can be removed from proposal message
- [x] Add encoder and decoder entities to make working with messages easier
- [x] Change voting messages encoding:
    - [x] Very 1st voting message is different:
        - [x] Send ids of all potential votes to vote for, i.e. send a list of votes for and potential votes for (a subset of possible votes), and also a list of votes for.
    - [x] Next rounds, a voting message is a bit vector, which length is equal to the length of the subset of possible votes.
        - [x] A threshold in the config for max total votes is needed
- [x] Use node ID instead of Public Key
- [x] Vote weight comes from an ATX, same as ATX weight
- [x] Vote calculation should be encoding-independent
- [x] Delayed proposals should be renamed to potentially valid proposals
- [ ] Votes margin should be calculated in two steps: before applying weak coin, after applying weak coin
- [x] Check VRF of weak coin messages, if it's invalid, ignore the message
- [ ] https://github.com/spacemeshos/go-spacemesh/issues/1019

collector/collector.go
antonlerner: return here? don't store beacon if we failed to deserialise

common/types/block.go
```diff
@@ -63,6 +63,16 @@ func (l LayerID) GetEpoch() EpochID {
 	return EpochID(uint64(l) / uint64(getLayersPerEpoch()))
 }
 
+// OrdinalInEpoch returns layer ordinal in epoch.
+func (l LayerID) OrdinalInEpoch() uint64 {
```
antonlerner, line 67: perhaps this should return epochId? this would be more clear

collector/collector.go
nkryuchkov: @antonlerner Should I do the same fix for other `case`s in switch while I'm fixing it here?

tortoisebeacon/beaconcalc.go
```diff
@@ -0,0 +1,75 @@
...
+
+	votes, ok := tb.ownVotes[lastRound]
+	if !ok {
+		// re-calculate votes
+		v, err := tb.calcVotes(epoch, lastRound.Round)
+		if err != nil {
+			return nil, fmt.Errorf("recalculate votes: %w", err)
+		}
+
+		votes = v
```
antonlerner: should this be assigned to `ownVotes` ?

tortoisebeacon/beaconcalc_test.go
```diff
@@ -0,0 +1,301 @@
+package tortoisebeacon
+
+//func TestTortoiseBeacon_calcBeacon(t *testing.T) {
```
antonlerner: enable these?

tortoisebeacon/votesdecoder.go
```diff
@@ -0,0 +1,45 @@
+package tortoisebeacon
```
antonlerner: missing unit tests

nkryuchkov: added

tortoisebeacon/beaconcalc_test.go
nkryuchkov: enabled

tortoisebeacon/beaconcalc.go
nkryuchkov: added an assignment

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

activation/atxdb.go
```diff
@@ -602,6 +622,19 @@ func (db *DB) addNodeAtxToEpoch(epoch types.EpochID, nodeID types.NodeID, atx *t
...
 
+func (db *DB) addAtxTimestamp(timestamp time.Time, atx *types.ActivationTx) error {
+	db.log.Info("added atx %v timestamp %v", atx.ID().ShortString(), timestamp)
+
+	b := make([]byte, 8)
+	binary.LittleEndian.PutUint64(b, uint64(timestamp.UnixNano()))
+
+	err := db.atxs.Put(getAtxTimestampKey(atx.ID()), b)
+	if err != nil {
+		return fmt.Errorf("failed to store atx ID for node: %v", err)
```
antonlerner: change log message or return `err`

go.sum
```diff
@@ -36,6 +36,8 @@ cloud.google.com/go/storage v1.10.0/go.mod h1:FLPqc6j+Ki4BU591ie1oL6qBQGu2Bl/tZ9
 code.cloudfoundry.org/bytefmt v0.0.0-20180906201452-2aa6f33b730c h1:VzwteSWGbW9mxXTEkH+kpnao5jbgLynw3hq742juQh8=
 code.cloudfoundry.org/bytefmt v0.0.0-20180906201452-2aa6f33b730c/go.mod h1:wN/zk7mhREp/oviagqUXY3EwuHhWyOvAdsn5Y4CzOrc=
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
+github.com/ALTree/bigfloat v0.0.0-20201218142103-4a33235224ec h1:tPYYUIlnk8STUTx7TEcZqCgk0zvxL6KH6eDSTa38zBI=
+github.com/ALTree/bigfloat v0.0.0-20201218142103-4a33235224ec/go.mod h1:+NaH2gLeY6RPBPPQf4aRotPPStg+eXc8f9ZaE4vRfD4=
```
antonlerner, line 44: where is this used? we shouldn't rely on external floating point imlementations since it may cause different values in different implementations and may affect consensus

tests/utils.py
```diff
@@ -151,6 +153,33 @@ def validate_blocks_per_nodes(block_map, from_layer, to_layer, layers_per_epoch,
...
+
+        if log.beacon not in epoch_messages[log.epoch_id]:
+            epoch_messages[log.epoch_id][log.beacon] = 0
+
+        epoch_messages[log.epoch_id][log.beacon] += 1
+
+        assert log.beacon != '0x0000000000000000000000000000000000000000000000000000000000000000', \
+            f"beacon in epoch {log.epoch_id} is 0x00...00: {log.beacon}"
+
+        assert log.beacon != '0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855', \
```
antonlerner, line 175: how is it that this is the same beacon in every run? shouldn't it be a hash of all ATXs?

tortoisebeacon/atxid_list.go
```diff
@@ -0,0 +1,27 @@
+package tortoisebeacon
+
+import (
+	"github.com/spacemeshos/sha256-simd"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+)
+
+// ATXIDList defines ATX ID list.
+type ATXIDList []types.ATXID
```
antonlerner: put this in types

go.sum
nkryuchkov: It's used in https://github.com/spacemeshos/go-spacemesh/blob/tortoise-ticker/tortoisebeacon/tortoise_beacon.go#L808
I use it because I need to calculate 2 to the power of `big.Float`, and the `Pow` method doesn't exist on https://golang.org/pkg/math/big/#Float

tests/utils.py
nkryuchkov: The beacon should not be equal to `0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`, because this number is `sha256("")`

tortoisebeacon/handlers.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	sender := m.MinerID.VRFPublicKey // hash of a miner PK
+	ok := tb.vrfVerifier(sender, currentEpochProposal, m.VRFSignature)
+	if !ok {
+		// TODO: attach telemetry
+		tb.Log.With().Warning("Received malformed proposal message: VRF is not verified",
+			log.String("sender", m.MinerID.String()),
+			log.String("sender_short", m.MinerID.ShortString()))
+		return nil
```
antonlerner, line 78: shouldn't this return an error as well?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -0,0 +1,917 @@
...
+	}
+
+	if (epochID - 1).IsGenesis() {
+		return genesisBeacon.Bytes(), nil
+	}
+
+	tb.beaconsMu.RLock()
+	defer tb.beaconsMu.RUnlock()
+
+	beacon, ok := tb.beacons[epochID-1]
```
antonlerner, line 257: will this hold all beacons all the time?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -0,0 +1,917 @@
...
+	}
+}
+
+func (tb *TortoiseBeacon) cleanup() {
+	// TODO(nkryuchkov): implement a better solution, consider https://github.com/golang/go/issues/20135
+	tb.beaconsMu.Lock()
+	defer tb.beaconsMu.Unlock()
+
+	for e := range tb.beacons {
+		if tb.epochIsOutdated(e) {
```
antonlerner, line 297: I'd do it when inserting a new beacon - this will eliminate the need for an additional gouroutine and simply follow a LRU heuristic

tortoisebeacon/tortoise_beacon.go
```diff
@@ -0,0 +1,917 @@
...
+
+	if err := tb.runProposalPhase(ctx, epoch); err != nil {
+		tb.Log.With().Error("Failed to send proposal",
+			log.Uint64("epoch_id", uint64(epoch)),
+			log.Err(err))
+
+		return
+	}
+
+	if err := tb.runConsensusPhase(ctx, epoch); err != nil {
```
antonlerner, line 393: should we continue to net phase if error was received here?

tortoisebeacon/votes_calc.go
```diff
@@ -0,0 +1,256 @@
...
+		thisRoundVotes := tb.incomingVotes[thisRound]
+
+		for pk, votesList := range thisRoundVotes {
+			voteWeight, err := tb.voteWeight(pk, epoch)
+			if err != nil {
+				return fmt.Errorf("vote weight: %w", err)
+			}
+
+			for vote := range votesList.ValidVotes {
+				votesMargin[vote] += int(voteWeight)
```
antonlerner, line 206: could there be an overflow here?

tortoisebeacon/votes_encoder.go
```diff
@@ -0,0 +1,42 @@
+package tortoisebeacon
+
+import "github.com/bits-and-blooms/bitset"
+
+func (tb *TortoiseBeacon) encodeVotes(currentRound votesSetPair, firstRound firstRoundVotes) (votesBitVector []uint64) {
```
antonlerner, line 5: does need to be a method of `TortoiseBeacon`?

tortoisebeacon/votes_decoder.go
```diff
@@ -0,0 +1,45 @@
+package tortoisebeacon
+
+import "github.com/bits-and-blooms/bitset"
+
+func (tb *TortoiseBeacon) decodeVotes(votesBitVector []uint64, firstRound firstRoundVotes) votesSetPair {
```
antonlerner, line 5: does this need to be a method of TortoiseBeacon?

tortoisebeacon/votes_encoder.go
nkryuchkov: Assuming the current state — no. I left it as a method to be able to add logs easily if we need them.

tortoisebeacon/tortoise_beacon.go
```diff
@@ -0,0 +1,917 @@
...
+	signature := tb.vrfSigner.Sign(p)
+	tb.Log.With().Debug("Calculated signature",
+		log.Uint64("epoch_id", uint64(epoch)),
+		log.String("proposal", util.Bytes2Hex(p)),
+		log.String("signature", util.Bytes2Hex(signature)))
+
+	return signature, nil
+}
+
+func (tb *TortoiseBeacon) calcEligibilityProof(message interface{}) ([]byte, error) {
```
antonlerner: this does not calculates eligibility proof, it signs the message - rename it

tortoisebeacon/tortoise_beacon.go
```diff
@@ -0,0 +1,917 @@
...
+	encoded, err := types.InterfaceToBytes(message)
+	if err != nil {
+		return nil, err
+	}
+
+	return tb.edSigner.Sign(encoded), nil
+}
+
+func (tb *TortoiseBeacon) calcProposal(epoch types.EpochID) ([]byte, error) {
```
antonlerner: rename to "buildProposal" since there is no calculation done here

tortoisebeacon/tortoise_beacon.go
```diff
@@ -0,0 +1,917 @@
...
+	maxPossibleNumberBigInt := new(big.Int).Exp(two, signatureLengthBigInt, nil)
+	maxPossibleNumberBigFloat := new(big.Float).SetInt(maxPossibleNumberBigInt)
+
+	thresholdBigFloat := new(big.Float).Mul(maxPossibleNumberBigFloat, fraction)
+	threshold, _ := thresholdBigFloat.Int(nil)
+
+	return threshold, nil
+}
+
+func (tb *TortoiseBeacon) calcProposalSignature(epoch types.EpochID) ([]byte, error) {
```
antonlerner: rename to getSignedProposal

tortoisebeacon/votes_calc.go
```diff
@@ -0,0 +1,256 @@
...
+		firstRoundVotesFor := make(hashSet)
+		firstRoundVotesAgainst := make(hashSet)
+
+		voteWeight, err := tb.voteWeight(nodeID, epoch)
+		if err != nil {
+			return nil, fmt.Errorf("get vote weight: %w", err)
+		}
+
+		for vote := range votesList.ValidVotes {
+			firstRoundVotesMargin[vote] += int(voteWeight)
```
antonlerner, line 126: can this overflow?

nkryuchkov: @antonlerner I think so. Should I use `big.Int`?

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

antonlerner (approved):

nkryuchkov:
bors merge

PR #2270, opened 2021-01-28
[Merged by Bors] - Make testnet join instructions generic

lrettig:
Update link and instructions for joining the latest testnet

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,9 +177,10 @@ Assuming tn1.json is a testnet config file saved in the same directory as go-spa
 
 4. Use the CLI wallet to check your coinbase account balance and to transact
 
-#### Joining Spacemesh ([TweedleDee](https://testnet.spacemesh.io/#/?id=what-is-spacemesh-01-tweedledee)) Testnet (net id 118)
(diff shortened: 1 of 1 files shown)


noamnelke (approved):
Good enough for now, let's merge it.

lrettig:
bors merge

avive:
It is good we are updating but you want to add how to get a CLIWallet that works with go-sm - something like what I wrote at the bottom of this: https://github.com/spacemeshos/go-spacemesh/releases/tag/v0.1.22 . There is a specific CLIWallet release that is kind of verified to work with the TN.

lrettig:
This is blocking on #2272

avive:
Why not just edit readme on github in place in develop branch for now?

avive:
Sorry, looks like this is blocked as well because develop is a protected branch.

web3rover:
If there is no specific reason to show the latest tweedledee version then I would prefer to change the link to https://github.com/spacemeshos/go-spacemesh/releases/latest and explicitly mention that follow the steps in the latest release to join the network. This way we don't need to keep updating the readme.

lrettig:
Exactly. Unfortunately there's no way around bors, even for simple
documentation changes. That's why documentation like this that's updated
frequently should be moved out of go-spacemesh into another repo where we
can change it more easily.

lrettig:
bors merge

PR #2272, opened 2021-01-28
[Merged by Bors] - Mark CI auto-success on documentation-only changes

### LINKED ISSUE #2179: GitHub Actions not triggering on push without changes
Opening this to document a strange issue that @noamnelke discovered today. It appears that, when we `bors try` a PR and the tests fail, then we `bors try` again, GH Actions fails to trigger the `push` workflow the second time if there are no changes to the code. I confirmed that making a tiny, superficial change to the code causes it to work again. It appears that Actions has an optimization where, if there's a subsequent push to the same branch with no change to the code, it doesn't run again. I can't find this documented anywhere, but it may be an issue with bors and flakey tests.

Bors sort of doesn't play nice with Actions so this is not terribly surprising.

At the very least we should understand what's going on and make sure it's documented. We should consider workarounds, too. Random ideas:

- cause the relevant workflow to make a superficial commit to the source branch if a push to staging fails (e.g., a file that tracks the last commit hash or something)
- use a different branch name for each try (e.g., `staging-XXXXXX` where `XXXXXX` is derived from the time, source branch, latest commit hash, or something) - most workflows support wildcards in branch names so this might work
- "force" a workflow to run using the GH API
- make bors smart enough to _rerun_ the previous workflow instead of creating a new push

### LINKED ISSUE #2271: Unable to merge markdown-only changes
## Description
Right now, GHA totally ignores PRs and pushes that change only markdown code. This is my fault, the original intention here was to skip the CI/bors process entirely for superficial, documentation-only changes. Unfortunately, as a result, we can't actually merge such PRs today (ex.: #2270) since bors waits forever for CI to pass, but CI never actually runs.

https://github.com/spacemeshos/go-spacemesh/blob/35432834b1bcc8aa4110e033c9d4e2e9701bc7d7/.github/workflows/ci.yml#L20-L31

lrettig:
## Motivation
Closes #2271 

## Changes
Adds a new CI workflow that is triggered only on changes exclusively to markdown files, which instantly marks both CI stages as succeeded

### DIFF
--- a/.github/workflows/ci-markdown.yml
+++ b/.github/workflows/ci-markdown.yml
@@ -1,4 +1,9 @@
-# Much simpler CI for changes only to markdown documentation
+# Much simpler CI for changes only to markdown documentation.
+# If **only** markdown files were changed, the main CI workflow
+# does not run. Instead, this workflow runs and immediately
+# marks the stages as succeeded so that CI and bors immediately
+# pass. If other files were changed, both workflows run, but this
+# one does nothing.
 name: CI-markdown
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-markdown.yml
```diff
@@ -0,0 +1,34 @@
...
+# Trigger the workflow on the pushes that bors cares about
+on:
+  push:
+    branches:
+      - staging
+      - trying
+    # Only trigger if updating only docs
+    paths:
+      - '**.md'
+      - '!**.md'
```
noamnelke: This seems wrong... Based on [this](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths), this workflow will only trigger when any file with extension `md` is included, excluding any file with extension `md`, in other words, it will never run.

I couldn't find a way to achieve what we need using GHA filters. You could create a bash command that achieves it if no better solution comes to mind.

noamnelke (changes requested):

.github/workflows/ci-markdown.yml
lrettig: You're right. I'll figure this out.

noamnelke (approved):
LGTM

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

>

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

PR #2279, opened 2021-02-02
[Merged by Bors] - Optimizing TransactionsStateStream

### LINKED ISSUE #2076: Optimize filtering transactions in TransactionsStateStream
## Description
The TransactionService currently implements filtering of transactions in the simplest, most naive way possible, leading to this awful triple-nested for loop. Similar to #2073, we need to optimize this, but it might require changing our database indices: we need a way to find txs matching a set of txids in a given layer.

## Affected code
https://github.com/spacemeshos/go-spacemesh/blob/d4d76b6eaf550f7732a10e208ceb7001d0b34664/api/grpcserver/transaction_service.go#L219-L258

sidsankhe:
## Motivation

Closes #2076 

## Changes

Edited the function to create a set (in the form of a map) and then use that to efficiently check for membership, rather than using a doubly nested for loop. 
## Test Plan

Since no new functionality was added, no new tests were added, but all the existing tests verify correctness.
## TODO


noamnelke:
@sidsankhe pull requests should have a proper title and description from the moment of creation, or it becomes very hard to keep track of- and manage all open PRs. If you don't want to invest the time to describe your work up front, just don't create a PR for your branch until later - there's no benefit to a PR other than the ability to easily review (your own and others') code.

sidsankhe:
@noamnelke So sorry for this, I will keep this in mind for future PRs. I have updated the name.

api/grpcserver/transaction_service.go
```diff
@@ -226,42 +226,48 @@ func (s TransactionService) TransactionsStateStream(in *pb.TransactionsStateStre
...
-							if err := stream.Send(res); err != nil {
-								return err
+				//create a set for the block transaction IDs
+				for _, txid := range b.TxIDs {
+					blockTXIDSet[txid] = struct{}{}
+				}
+
+				for _, inputTxID := range in.TransactionId {
+					var arrayID [32]byte
+					copy(arrayID[:], inputTxID.Id[:])
```
lrettig, line 254: This copy seems unnecessary

api/grpcserver/transaction_service.go
```diff
@@ -226,42 +226,48 @@ func (s TransactionService) TransactionsStateStream(in *pb.TransactionsStateStre
...
+						case events.LayerStatusTypeApproved:
+							txstate = pb.TransactionState_TRANSACTION_STATE_MESH
+						case events.LayerStatusTypeConfirmed:
+							txstate = pb.TransactionState_TRANSACTION_STATE_PROCESSED
+						default:
+							txstate = pb.TransactionState_TRANSACTION_STATE_UNSPECIFIED
+						}
+						res := &pb.TransactionsStateStreamResponse{
+							TransactionState: &pb.TransactionState{
+								Id: &pb.TransactionId{
```
lrettig: You can simplify this:

```
								Id: inputTxID,
```

lrettig (changes requested):

api/grpcserver/transaction_service.go
sidsankhe: I can only convert it into a TransactionID if it's a fixed size array, and so I created a fixed sized array from the slice

lrettig: You're right. The underlying issue here is that data items including txid do not have fixed lengths in our protobuf API. I opened https://github.com/spacemeshos/api/issues/130 to investigate if there's anything we can do about this.

Could you please add an explanatory comment here? Something to the effect of, "Since the txid coming in from the API does not have a fixed length, we need to convert it from a slice into an array here before we can convert it into a TransactionID object. We don't need to worry about error-handling here since `copy()` intelligently copies only as much as it can. If the resultant TransactionID is invalid, an error will be returned below."

api/grpcserver/transaction_service.go
```diff
@@ -226,42 +226,46 @@ func (s TransactionService) TransactionsStateStream(in *pb.TransactionsStateStre
...
+				}
+
+				for _, inputTxID := range in.TransactionId {
+					var arrayID [32]byte
+					copy(arrayID[:], inputTxID.Id[:])
+					txid := types.TransactionID(arrayID)
+					// if there is an ID corresponding to inputTxID in the block
+					if _, exists := blockTXIDSet[txid]; exists {
+						var txstate pb.TransactionState_TransactionState
+						switch layer.Status {
```
lrettig: Another potential optimization here: this `switch` statement only depends on the `layer.Status`. It doesn't need to run for every tx. Could you move this outside the loop?

lrettig (changes requested):
See comments about adding one more optimization and [adding a comment](https://github.com/spacemeshos/go-spacemesh/pull/2279/files#r575816224). Then feel free to mark this as ready for review. Thanks!

lrettig (approved):
LGTM!

sidsankhe:
bors merge

sidsankhe:
bors merge

PR #2282, opened 2021-02-08
[Merged by Bors] - fix for missing early logs and late arrival of logs in the middle of the test

AmitShaul:
* upgraded fluent bit from 1.6.4 to 1.6.10
* added another replica of logstash (2 instead of 1)
* added a sleeping time of 15 seconds in the harness in order to enable fluent bit container to be created first

## Motivation
early logs were missing, seems that fluent bit was not ready when bootstrap pod was already up.
logs were arriving late in the middle of late_nodes test. this was caused by high pressure on our logstash service what caused http 429 warnings (too many requests).
added another replica of logstash in order to fix this.


AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

tests/node_pool_deployer.py
```diff
@@ -70,18 +70,19 @@ def calculate_cpu_and_mem_per_node(self):
...
         dividor = 6
         for i in range(1, 6):
-            if total_cpu / i > max_cpus_per_node:
+            if total_cpu / i > max_cpus_per_node + extra_cpus_per_machine:
                 continue
             dividor = i
             break
 
-        cpu_per_node = ceil(total_cpu / dividor)
+        cpu_per_node = ceil(total_cpu / dividor) + extra_cpus_per_machine
```
noamnelke: The value here doesn't match the name of the variable - please update one of them to match the other.

What's this variable used for? It seems to me that leaving this line as it was before makes the most sense.

noamnelke (commented):
We're waiting 15 seconds before the pod starts (`SLEEP_TIME_BEFORE_POD_START = 15`) and another 15 seconds after the pod starts and before the node process is started by the harness (sleep in `harness.go`)? Why split this waiting period and not wait 30 seconds in one of these places?

tests/node_pool_deployer.py
AmitShaul: fixed, and also found a bug there with the calculation, thanks

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

antonlerner (approved):

AmitShaul:
bors merge

lrettig:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

tests/k8s_handler.py
```diff
@@ -198,3 +199,34 @@ def wait_for_namespaced_deletion(name, namespace, deletion_func, list_func, time
...
+
+def wait_for_daemonset_to_be_ready(name, namespace, timeout=None):
+    wait_for_to_be_ready("daemonset", name, namespace, timeout=timeout)
+
+
+def resolve_read_status_func(obj_name):
+    if obj_name == "daemonset":
+        return client.AppsV1Api().read_namespaced_daemon_set_status
+    else:
+        raise ValueError(f"resolve_read_status_func: {obj_name} is not a valid value")
```
antonlerner, line 212: if this error occurs it will crash the entire run - is this function something we'd like to retry?

antonlerner (approved):

tests/k8s_handler.py
AmitShaul: it's a setup for other waiting functions to be included here (deployment and statefulset) though I have other tasks to attend to at the moment so I will join those in the future.
right now this function is only utilised for daemonset and that's the only value to be given here so no worries about exceptions at this point.
we already now failing tests if pods are not ready in time so it seemed right to throw an exception here but that might be changed once I change this func to include other waiting functions as well.

AmitShaul:
bors merge

PR #2284, opened 2021-02-09
[Merged by Bors] - Correct path matching in CI-markdown

lrettig:
## Motivation
This was wrong in #2272, so when only markdown files are changed, currently, CI fails to run completely.

This happened because the glob syntax [used by GitHub](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) slightly differs from that [used by picomatch](https://github.com/micromatch/picomatch#basic-globbing) (used under the hood in the library we're using here).

## Changes
- fix picomatch glob syntax to match all md files regardless of path

## Test Plan
N/A

### DIFF
--- a/.github/workflows/ci-markdown.yml
+++ b/.github/workflows/ci-markdown.yml
@@ -31,7 +31,7 @@ jobs:
     - uses: dorny/paths-filter@v2
       id: filter
       with:
-        # this pattern matches using picomatch syntax (used by this Action), which is slightly
+        # this pattern matches using picomatch syntax (used by this third party Action), which is slightly
         # different than the GitHub syntax above: it also matches any file in any path ending in '.md'
         filters: |
           nondoc:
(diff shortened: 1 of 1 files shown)


noamnelke (changes requested):
I think we also need to update line 20 in the same file. No?

lrettig:
> I think we also need to update line 20 in the same file. No?

No. Line 20 matches using the GitHub matching engine. It's confusing, but the syntax is slightly different. See the links I posted [above](https://github.com/spacemeshos/go-spacemesh/pull/2284#issue-570401456).

I'll a comment to this effect inline to make this a little clearer.

noamnelke (approved):

noamnelke:
Wow, confusing indeed...

lrettig:
bors merge

PR #2286, opened 2021-02-11
Run systemtests on self-hosted runners

valar999:
## Motivation

We would like to test the idea that some of systemtests failure occur due to GitHub Actions runners.

## Changes

Change where systemtests run.

## Test Plan

Tested more that 20 times on these self-hosted runners.

## TODO

Autoscaler for self-hosted runners group have no suitable metrics.


noamnelke (approved):
I don't think that this is the reason for the issues, but I don't mind testing your hypothesis.

valar999:
No difference found when using self-hosted runners.

PR #2288, opened 2021-02-16
[Merged by Bors] - Api exception while sending a HTTP request to a miner

AmitShaul:
## Motivation
Following a recurring bug in the CI involving 'search_phrase_in_pod_log' automation function.
mining test would sometimes fail due to a 'kubernetes.client.rest.ApiException: (500)' exception.
this is caused because of an unhandled exception.
added a retry mechanism to handle those exceptions.

### DIFF
--- a/tests/pod.py
+++ b/tests/pod.py
@@ -97,7 +97,7 @@ def check_for_restarted_pods(namespace, specific_deployment_name=''):
     return restarted_pods
 
 
-def search_phrase_in_pod_log(pod_name, name_space, container_name, phrase, time_out=60, group=None, retry=3):
+def search_phrase_in_pod_log(pod_name, name_space, container_name, phrase, timeout=60, group=None, retry=3):
     sleep_interval = 0.3
     total_sleep = 0
     _retry = retry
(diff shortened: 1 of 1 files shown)


tests/pod.py
```diff
@@ -97,19 +97,31 @@ def check_for_restarted_pods(namespace, specific_deployment_name=''):
     return restarted_pods
 
 
-def search_phrase_in_pod_log(pod_name, name_space, container_name, phrase, time_out=60, group=None):
-
-    match = None
+def search_phrase_in_pod_log(pod_name, name_space, container_name, phrase, timeout=60, group=None, retry=3):
+    sleep_interval = 0.3
     total_sleep = 0
+    _retry = retry
```
y0sher: why this? we can't just change retry?

y0sher: not really important though but confusing

y0sher (approved):

tests/pod.py
noamnelke: @y0sher this var is reset in line 125 if a successful call was made but no match found.

@AmitShaul I would consider renaming `_retry` to `retries_remaining` and `retry` to `retries`.

noamnelke (approved):

AmitShaul:
bors try

AmitShaul:
bors merge

PR #2293, opened 2021-02-23
[Merged by Bors] - Fix cPool flaky unit test

y0sher:
## Motivation
Unit test was failing randomly .

## Changes

better check the errors and wait for connections to close before ending the test.

### DIFF
--- a/p2p/connectionpool/connectionpool_test.go
+++ b/p2p/connectionpool/connectionpool_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	net2 "net"
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


noamnelke (approved):
I'm not a big fan of randomness in unit tests, but if you believe that removing the seeding was "necessary evil" we could add it back, but print the seed. This way, if it randomly fails we can hardcode the random seed into the test to reproduce the failure.

Approved with/without the change I proposed. Up to you.

y0sher:
> I'm not a big fan of randomness in unit tests, but if you believe that removing the seeding was "necessary evil" we could add it back, but print the seed. This way, if it randomly fails we can hardcode the random seed into the test to reproduce the failure.
> 
> Approved with/without the change I proposed. Up to you.

I think you missed something or misread the PRs, in this PR I haven't changed anything related to randomness.

y0sher:
bors merge

p2p/connectionpool/connectionpool_test.go
```diff
@@ -268,7 +269,6 @@ func TestRandom(t *testing.T) {
 	nMock.SetDialDelayMs(50)
 	nMock.SetDialResult(nil)
 	cPool := NewConnectionPool(nMock.Dial, generatePublicKey(), log.NewDefault(t.Name()))
-	rand.Seed(time.Now().UnixNano())
```
noamnelke, line 271: by not seeding you removed the randomness, no?

y0sher:
bors merge

PR #2294, opened 2021-02-23
Discovery Ping Implementation

### LINKED ISSUE #718: Finish discovery "Ping" feature
# Overview / Motivation
*TODO*: choose a better name
we want to be able to ensure nodes are listening where they claim to. nodes advertise themselves with a `ping` message with their IPs and ports, we should validate they're listening with another ping message to this address.

# The Task

Implement a mechanism that remembers which peers we pinged lately (and got ponged back), condition sending `Addresses` messages to peers only after we pinged them and got ponged. this serves to protect other peers from reflected denial of service where an attacker could spoof source address in order to make honest peers sending many large `Addresses` messages to a victim.

# Implementation Notes

should be implemented around the `verifyPinger` method.

# Contribution Guidelines
Important: Issue assignment to developers will be by the order of their application and proficiency level according to the tasks complexity. We will not assign tasks to developers who have'nt introduced themselves on our Gitter [dev channel](https://gitter.im/spacemesh-os/Lobby)

1. Introduce yourself on go-spacemesh [dev chat channel](https://gitter.im/spacemesh-os/Lobby) - ask our team any question you may have about this task
2. Fork branch `develop` to your own repo and work in your repo
3. You must document all methods, enums and types with [godoc comments](https://blog.golang.org/godoc-documenting-go-code)
4. You must write go unit tests for all types and methods when submitting a component, and integration tests if you submit a feature
5. When ready for code review, submit a PR from your repo back to branch `develop`
6. Attach relevant issue to PR

sidsankhe:
## Motivation

Closes #718 

## Changes

- Added a `lastPing` field to the `knownAddress` struct in order to determine when a peer in the address book was pinged last. 
- On a `getAddresses` request, if the node hasn't been pinged in the last `pingInterval`, the receiving node will request a `ping` from the node. 

## Test Plan

- Added unit tests to make sure that nodes issuing a `getAddresses` request that hadn't been pinged recently received a ping. 
- WIP: need to move the `lastDiscoveryPing` down into the testing code rather than it being in production code. 

## TODO

## DevOps Notes


p2p/server/msgserver.go
```diff
@@ -108,10 +109,10 @@ func (p *MessageServer) readLoop() {
 				return
 			}
 			p.workerCount.Add(1)
-			p.workerLimiter <- struct{}{}
+			//p.workerLimiter <- struct{}{}
```
y0sher, line 112: should probably un-comment but can you still explain please?

p2p/discovery/ping.go
```diff
@@ -17,18 +18,19 @@ func (p *protocol) newPingRequestHandler() func(msg server.Message) []byte {
...
 			return nil
 		}
-
 		if err := p.verifyPinger(msg.Metadata().FromAddress, pinged); err != nil {
 			plogger.Error("msg contents were not valid err=", err)
 			return nil
 		}
-
+		//only update the last pinged time once we know everything is valid
+		p.lastDiscoveryPing = time.Now()
```
y0sher: is this used for anything other than tests?

p2p/discovery/ping.go
```diff
@@ -17,18 +18,19 @@ func (p *protocol) newPingRequestHandler() func(msg server.Message) []byte {
 	return func(msg server.Message) []byte {
 		plogger := p.logger.WithFields(log.String("type", "ping"), log.String("from", msg.Sender().String()))
 		plogger.Debug("handle request")
+		plogger.Warning("In ping request handler")
```
y0sher: switch to debug and remove one of the lines..

p2p/discovery/ping.go
```diff
@@ -38,6 +40,7 @@ func (p *protocol) newPingRequestHandler() func(msg server.Message) []byte {
 		}
 
 		plogger.Debug("Sending pong message")
+		plogger.Warning("Sending pong message")
```
y0sher: same thing with logs..

p2p/discovery/ping.go
```diff
@@ -38,6 +40,7 @@ func (p *protocol) newPingRequestHandler() func(msg server.Message) []byte {
...
 		return payload
 	}
 }
```
y0sher: we should probably check if a peer `NeedsPing` after we got pinged from him, this is used to make sure the pinger with his provided identity actually owns the IP from the incoming ping, otherwise he could use it to create a reflected DoS attack and make many peers try and reach specific address, so we check this before we save an address in the addressbook.

sidsankhe:
bors merge

p2p/discovery/ping.go
sidsankhe: This makes sense. If the peer who's sending the ping `NeedsPing` within the `AddressBook`, what should the desired behavior be? We should not issue another ping request right, because that would risk an infinite loop of pings. Should the desired behavior then be to not reply to the ping request?

p2p/discovery/ping.go
sidsankhe: No, I was trying to figure out a better way to test if a peer has been pinged without adding this flag to the protocol that is only used for tests. Any thoughts?

p2p/server/msgserver.go
sidsankhe: There was a race condition that caused a deadlock in the following test: `TestFindNodeProtocol_FindNode_Concurrency` in `p2p/discovery/protocol_test.go`. I worked on this for a while and commenting the above line was a fix to it. I can keep looking into it more.

p2p/discovery/ping.go
y0sher: it could probably be done by mocking out some part of the function like `verifyPinger` and making it run specific test logic and mutate test state. there might be other parts that could be used to achieve the same thing.

p2p/server/msgserver.go
y0sher: we shouldn't merge before working this out, might be a test specific bug or some deadlock that can happen in real-time. I'll take a look once this will be the only thing left..

p2p/discovery/ping.go
y0sher: I see your point, we might need to check this only before sending a GetAdderss\Addresses message because these are bigger messages, anyway this was inspired from ethereum's `Bond` mechanism, please take a look here -
https://github.com/ethereum/go-ethereum/blob/4adf0b9187060539e8c0dd9eda2ad5a7d3c53cac/p2p/discover/v4_udp.go#L570

p2p/discovery/ping.go
sidsankhe: Pushed this logic down to the sim layer, when a message is being sent

p2p/discovery/ping.go
sidsankhe: We currently check whether the Peer needs ping before the GetAddresses. If the peer needs the ping, then we send a ping and without a successful response, we don't respond.

nkryuchkov:
@y0sher @lrettig 
Should we close this?

y0sher:
If we're moving to libp2p (@dshulyak) or even if we plan (which is the plan to ease of unroutable clients) to drop udp discovery we can skip this PR.

PR #2296, opened 2021-02-25
[Merged by Bors] - Logging, whitespace, comment cleanup

lrettig:
## Motivation
Minor improvements made in the process of testnet debugging. Primarily improves logging and makes some log output cleaner and clearer.

## Changes
- Makes log messages clearer (for testnet debugging)
- Some minor whitespace and comment cleanup
- Some very, very minor logic optimizations

## Test Plan
None, no logic changes

## DevOps Notes

- [x] Does this code make changes to log messages that our monitoring infrastructure may rely on?

### DIFF
--- a/blocks/blockoracle.go
+++ b/blocks/blockoracle.go
@@ -25,6 +25,8 @@ type vrfSigner interface {
 	Sign(msg []byte) ([]byte, error)
 }
 
+const DefaultProofsEpoch = ^types.EpochID(0)
+
 // Oracle is the oracle that provides block eligibility proofs for the miner.
 type Oracle struct {
 	committeeSize        uint32
@@ -46,7 +48,6 @@ type Oracle struct {
 
 // NewMinerBlockOracle returns a new Oracle.
 func NewMinerBlockOracle(committeeSize uint32, genesisActiveSetSize uint32, layersPerEpoch uint16, atxDB activationDB, beaconProvider *EpochBeaconProvider, vrfSigner vrfSigner, nodeID types.NodeID, isSynced func() bool, log log.Log) *Oracle {
-
 	return &Oracle{
 		committeeSize:        committeeSize,
 		genesisActiveSetSize: genesisActiveSetSize,
@@ -55,7 +56,7 @@ func NewMinerBlockOracle(committeeSize uint32, genesisActiveSetSize uint32, laye
 		beaconProvider:       beaconProvider,
 		vrfSigner:            vrfSigner,
 		nodeID:               nodeID,
-		proofsEpoch:          ^types.EpochID(0),
+		proofsEpoch:          DefaultProofsEpoch,
 		isSynced:             isSynced,
 		log:                  log,
 	}
@@ -69,7 +70,15 @@ func (bo *Oracle) BlockEligible(layerID types.LayerID) (types.ATXID, []types.Blo
 	}
 
 	epochNumber := layerID.GetEpoch()
-	bo.log.With().Info("asked for eligibility", epochNumber, log.FieldNamed("cached", bo.proofsEpoch))
+	var cachedEpochDescription log.Field
+	if bo.proofsEpoch == DefaultProofsEpoch {
+		cachedEpochDescription = log.String("cached_epoch_id", "(none)")
+	} else {
+		cachedEpochDescription = log.FieldNamed("cached_epoch_id", bo.proofsEpoch.Field())
+	}
+	bo.log.With().Info("asked for eligibility",
+		log.FieldNamed("epoch_id", epochNumber),
+		cachedEpochDescription)
 	if epochNumber.IsGenesis() {
 		bo.log.Info("asked for eligibility for genesis epoch, cannot create blocks here")
 		return *types.EmptyATXID, nil, nil, nil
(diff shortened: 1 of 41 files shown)


lrettig:
bors try

lrettig:
bors try

cmd/node/node.go
```diff
@@ -326,11 +326,11 @@ func (app *SpacemeshApp) getAppInfo() string {
 }
 
 // Cleanup stops all app services
-func (app *SpacemeshApp) Cleanup(cmd *cobra.Command, args []string) {
-	log.Info("App Cleanup starting...")
+func (app *SpacemeshApp) Cleanup(_ *cobra.Command, _ []string) {
```
noamnelke: Removing the names entirely is even cleaner IMO.

```suggestion
func (app *SpacemeshApp) Cleanup(*cobra.Command, []string) {
```

common/types/block.go
```diff
@@ -191,7 +191,7 @@ func (b *Block) Fields() []log.LoggableField {
 		log.FieldNamed("miner_id", b.MinerID()),
 		log.Int("view_edges", len(b.ViewEdges)),
 		log.Int("vote_count", len(b.BlockVotes)),
-		b.ATXID,
+		log.String("atxid", b.ATXID.ShortString()),
```
noamnelke: I don't like this change. If we want to print short ATX IDs that's fine, but then we should do it consistently, namely we should change the behaviour of:
```go
func (t ATXID) Field() log.Field { return t.Hash32().Field("atx_id") }
```

Perhaps all `Hash32`s should be printed shortly in logs, but even if not, we should be consistent with ATX IDs.

hare/algorithm.go
```diff
@@ -394,22 +415,32 @@ func (proc *consensusProcess) handleMessage(m *Msg) {
...
+			log.Int32("current_k", proc.k),
+			log.Int32("msg_k", m.InnerMsg.K),
+			types.LayerID(proc.instanceID))
 		return
 	}
 
 	// warn on late pre-round msgs
 	if m.InnerMsg.Type == pre && proc.k != -1 {
-		proc.With().Warning("encountered late PreRound message", log.String("sender_id", m.PubKey.ShortString()), types.LayerID(proc.instanceID))
+		proc.With().Warning("encountered late PreRound message",
```
```suggestion
		proc.With().Warning("encountered late preround message",
```

hare/algorithm.go
```diff
@@ -555,29 +589,30 @@ func (proc *consensusProcess) beginNotifyRound() {
...
-		proc.With().Warning("Begin notify round: not enough commits", log.Int("expected", proc.cfg.F+1), log.Int("actual", proc.commitTracker.CommitCount()))
+		proc.With().Warning("begin notify round: not enough commits",
+			log.Int("expected", proc.cfg.F+1), log.Int("actual", proc.commitTracker.CommitCount()))
 		return
 	}
 
 	cert := proc.commitTracker.BuildCertificate()
 	if cert == nil {
-		proc.Error("Begin notify round: Build certificate returned nil")
+		proc.Error("begin notify round: Build certificate returned nil")
```
```suggestion
		proc.Error("begin notify round: BuildCertificate returned nil")
```

noamnelke (approved):
❤️❤️❤️

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merg

lrettig:
bors merge

lrettig:
I think system tests are failing due to flakiness (#2222, #2281)

lrettig:
bors merge

lrettig:
bors merg

lrettig:
bors merge

lrettig:
bors try

lrettig:
bors merge

lrettig:
bors merge

blocks/blockoracle.go
```diff
@@ -69,16 +71,24 @@ func (bo *Oracle) BlockEligible(layerID types.LayerID) (types.ATXID, []types.Blo
 	}
 
 	epochNumber := layerID.GetEpoch()
-	bo.log.Info("asked for eligibility for epoch %d (cached: %d)", epochNumber, bo.proofsEpoch)
+	var cachedEpochDescription log.Field
+	if bo.proofsEpoch == DefaultProofsEpoch {
+		cachedEpochDescription = log.String("cached_epoch_id", "(none)")
+	} else {
+		cachedEpochDescription = log.FieldNamed("cached_epoch_id", bo.proofsEpoch.Field())
+	}
```
noamnelke, line 79: @lrettig I know I already approved this and it's already merged, but I just realized that this approach is problematic.

We sometimes put a string (`"(none)"`) and sometimes a numeric value (`bo.proofsEpoch.Field()` uses `log.Uint64()`) in a field named `cached_epoch_id`.

I think that Elasticsearch doesn't like it when a field has more than one data type and it causes indexing issues. Not sure exactly what the implications are.

@narayanprusty is this really an issue? Is there some better practice for cases like this?

lrettig: I agree we need to be careful about this - I didn't realize this was an issue previously, but it caused problems with another field (an error about late message arrival in hare) in the last testnet. It would be good to know the best practice for cases like this.

noamnelke: let's wait for @narayanprusty to chime in

web3rover: If we have different data type for a particular field then ES raises exception and the log is never stored. We have to fix it in code. We can convert uint64 to integer string.

Other solution is to transform the data type using logstash data type transformation plugin. But that isn't a good idea as we have to put install and configure the plugin in CI, spacecraft, local testnet and all other systems that push logs to ELK. In long term it will become difficult to manage this.

noamnelke: @lrettig why is it important to have this `"(none)"` string? I know Kibana let's you filter for explicitly missing fields - can't we just omit the field if it's not cached?

lrettig: Opened #2367, let's follow up there

Issue #2263, opened 2021-01-18
block * from layer * has no explicit voting, something went wrong


sl-he:
## Description
Node is crashed with this error log
```
2021-01-18T15:18:27.460Z        ERROR   8101d.trtl              Fatal: goroutine panicked. Stacktrace: goroutine 2734 [running]:
runtime/debug.Stack(0x4fba72, 0xc00edb0a1f, 0x114ca3f)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x9f
github.com/spacemeshos/go-spacemesh/log.Log.Panic(0xc000dc7380, 0xc000040320, 0x4d, 0x0, 0x0, 0x0)
        /root/source/go-spacemesh/log/zap.go:42 +0xd7
github.com/spacemeshos/go-spacemesh/tortoise.(*ninjaTortoise).addPatternVote.func1(0x47c1caaf98ef23d0, 0xbe18462bcf2ebc66, 0xc070583d39)
        /root/source/go-spacemesh/tortoise/ninja_tortoise.go:440 +0xb66
github.com/spacemeshos/go-spacemesh/tortoise.(*ninjaTortoise).handleIncomingLayer(0xc00002c2c0, 0xc00e3fed40)
        /root/source/go-spacemesh/tortoise/ninja_tortoise.go:567 +0x5ff
github.com/spacemeshos/go-spacemesh/tortoise.(*tortoise).HandleIncomingLayer(0xc00911d320, 0xc00e3fed40, 0x0, 0x0)
        /root/source/go-spacemesh/tortoise/algorithm.go:69 +0xa6
github.com/spacemeshos/go-spacemesh/mesh.(*validator).ValidateLayer(0xc00006b600, 0xc00e3fed40)
        /root/source/go-spacemesh/mesh/mesh.go:280 +0xcb
github.com/spacemeshos/go-spacemesh/sync.(*Syncer).getAndValidateLayer(0xc00033cb40, 0x392, 0x0, 0x0)
        /root/source/go-spacemesh/sync/syncer.go:1248 +0xe4
github.com/spacemeshos/go-spacemesh/sync.(*Syncer).handleCurrentLayer(0xc00033cb40, 0x111fa00, 0x15)
        /root/source/go-spacemesh/sync/syncer.go:415 +0x1dc
github.com/spacemeshos/go-spacemesh/sync.(*Syncer).handleWeaklySynced(0xc00033cb40)
[...]

Issue #2088, opened 2020-07-29
panic: block x from layer x has no explicit voting, something went wrong


lrettig:
## Description
My node is dying every time I try to sync the devnet (network id 2). Oddly enough, I resynced from scratch using a different datadir and the sync ran fine.

The same error was reported in a few other issues, looks like it's still an issue:

- #1671
- #1756
- #1999 
- #2010
- #2211
- #2263

## Steps To Reproduce
1. Run go-spacemesh using the devnet config file

## Environment
Please complete the following information:
 - OS: macOS
 - Node Version: `v0.0.0-unreleased+40fbf77ec348a55f24aee1133af1a29b17b67589
`

## Additional Resources 

Partial log below, full log here: https://gist.github.com/lrettig/ecbd9636923fed9f5bb39be557878620

```
2020-07-29T13:06:18.864-0400    INFO    3a2c0.sync              handle layer hash request       {"node_id": "3a2c0f229af198c034a1298195cd8272045425ca8b7492a48961066c2ede5b92", "layer_id": 234}
2020-07-29T13:06:19.968-0400    INFO    3a2c0.sync              handle layer hash request       {"node_id": "3a2c0f229af198c034a1298195cd8272045425ca8b7492a48961066c2ede5b92", "layer_id": 235}
2020-07-29T13:06:20.206-0400    INFO    3a2c0.sync              Node is out of sync setting gossip-synced to false and starting sync        {"node_id": "3a2c0f229af198c034a1298195cd8272045425ca8b7492a48961066c2ede5b92"}
2020-07-29T13:06:20.206-0400    INFO    3a2c0.sync              syncing layer   {"node_id": "3a2c0f229af198c034a1298195cd8272045425ca8b7492a48961066c2ede5b92", "current_sync_layer": 2189, "last_ticked_layer": 2618}
2020-07-29T13:06:20.206-0400    INFO    3a2c0.sync         .peers               now connected   {"node_id": "3a2c0f229af198c034a1298195cd8272045425ca8b7492a48961066c2ede5b92", "n_peers": 8}
2020-07-29T13:06:20.206-0400    INFO    3a2c0.sync              fetch layer hash        {"node_id": "3a2c0f229af198c034a1298195cd8272045425ca8b7492a48961066c2ede5b92", "layer_id": 2189}
[...]

Issue #1856, opened 2020-02-25
Verifying Tortoise


barakshani:
Consider implementing a verifying version of the tortoise, that only counts blocks that are inline with the Hare results. In quiet times that would be sufficient, and efficient. As a fallback we would need to run the current Tortoise implementation.

almogdepaz:
not sure i understand the benefits of implementing this, the current tortoise is not a bottleneck at "quiet times" and preforms rather well.

barakshani:
See relevant part here: https://github.com/spacemeshos/research/issues/33

PR #2297, opened 2021-02-25
Don't panic on block in view without explicit vote

lrettig:
## Motivation
This happens sporadically on every testnet. See #2088 and, most recently, #2263.

This does not need to be a panic. It happens in isolated cases when a block that's been voted for (i.e., that's in a view) hasn't been processed by hare for some reason. Reduce it to an error.

## Changes
Reduce panic to an error

## Test Plan
N/A


antonlerner (approved):
Pls decide if you want to merge or discard this change 
IMO this should be left out

lrettig:
I think this can go into develop - @antonlerner why do you think it should be left out?

lrettig:
Component was replaced in #1856
And this was added to the testnet branch long ago

PR #2298, opened 2021-02-25
Make tortoise less reliant on hare for testnet [DO NOT MERGE]

lrettig:
## Motivation
In the most recent testnets (123, 124), tortoise has failed because of hare failures. This makes tortoise less reliant on hare succeeding, so that the testnet should be able to continue to advance pbase even if hare fails completely.

## Changes
- Change how blocks vote when hare has failed
- Lower tortoise thresholds

## Test Plan
- Affected tests have been updated: block_builder tests have been updated to expect the right voting pattern, and tortoise tests have been updated to change the number of blocks per layer (and thus votes) proportional to the modified tortoise thresholds

## TODO
- [x] Change voting to vote _against_ all blocks in case of hare failure, per Tal

### DIFF
--- a/tortoise/ninja_tortoise.go
+++ b/tortoise/ninja_tortoise.go
@@ -16,8 +16,9 @@ type vec [2]int
 type patternID uint32 // this hash does not include the layer id
 
 const ( // Threshold
-	window          = 10
-	globalThreshold = 0.6
+	window             = 10
+	globalThreshold    = 0.5
+	newlyGoodThreshold = 0.4
 )
 
 var ( // correction vectors type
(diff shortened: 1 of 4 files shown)


lrettig:
This PR looks huge because I merged #2296 into it. I'll rebase it to remove that to make this one cleaner.

lrettig:
Done. @antonlerner this is ready for a review. I'll leave it marked as a draft to make it clear that it's not meant to be merged into develop. Thanks!

lrettig:
bors try

tortoise/ninja_tortoise.go
```diff
@@ -16,8 +16,9 @@ type vec [2]int
 type patternID uint32 // this hash does not include the layer id
 
 const ( // Threshold
-	window          = 10
-	globalThreshold = 0.6
+	window            = 10
+	globalThreshold   = 0.5
+	newLayerThreshold = 0.5 * (globalThreshold / 0.6) // adjust to match global threshold
```
antonlerner, line 21: I'd explain better why this adjustment is made and why like so (the result of this as I understand is 0.41 and I have no idea why this is the correct value)

antonlerner (approved):

lrettig:
Added a comment thus:

```
const (
	// These two thresholds control how tortoise votes on blocks and layers (patterns).
	// The first is used when voting for or against blocks (in a layer/pattern).
	// The second is used when voting for or against a layer (to promote it to newly-good as a pbase candidate).
	// They should be independent, but the existing tortoise tests expect the second to be a fixed ratio of the first.
	// NOTE: These have been lowered (from 0.6 and 0.5, respectively) for the testnet.
	globalThreshold   = 0.5
	newLayerThreshold = 0.5 * (globalThreshold / 0.6) // adjust to match global threshold
	window            = 10
)
```

Thanks @antonlerner ! Closing this as the relevant code has been included in the testnet release: 6a9bfe49302ffa97c4595c88414ca5c02c8f111c

PR #2305, opened 2021-03-03
[Merged by Bors] - Update default hare params

lrettig:
## Motivation
As discussed on today's R&D call. These are reasonable default params. Per @tal-m, we want hare iterations to be 40-50 in normal network scenarios to meet our security assumptions.

## Changes
- change `RoundDuration` (`hare-round-duration-sec`) to 10sec
- change `LimitIterations` (`hare-limit-iterations`) to 50 by default

## Test Plan
no new tests needed

## TODO


antonlerner:
I think since we've had another conversation and hare max iterations has changed to 5, but maybe this is the more robust setting we want as default now. 
if you decide to discard this change thats ok as well, just close this PR

lrettig:
> I think since we've had another conversation and hare max iterations has changed to 5, but maybe this is the more robust setting we want as default now.

Are you referring to the testnet? Or target mainnet params? I don't recall finalizing this number yet, but 50 seems a lot more reasonable than 1000, which is way too many. (Tal indicated that we need 40-50 in order to satisfy our security assumptions.)

Also, just to be clear, this is not `max-concurrent` consensus processes. This is the `max-iterations` for each CP.

antonlerner:
> > I think since we've had another conversation and hare max iterations has changed to 5, but maybe this is the more robust setting we want as default now.
> 
> Are you referring to the testnet? Or target mainnet params? I don't recall finalizing this number yet, but 50 seems a lot more reasonable than 1000, which is way too many. (Tal indicated that we need 40-50 in order to satisfy our security assumptions.)
> 
> Also, just to be clear, this is not `max-concurrent` consensus processes. This is the `max-iterations` for each CP.

Ok so it looks good, maybe use named parameters for this struct, but you can merge it as is...

lrettig:
@antonlerner this should be good to go, but I can't bors merge until you approve it :) thanks!

lrettig:
bors try

antonlerner (approved):

lrettig:
bors merge

PR #2308, opened 2021-03-03
[Merged by Bors] - Fix failing TestBlockBuilder_CreateBlockFlow

nkryuchkov:
## Motivation

WIP on https://github.com/spacemeshos/pm/issues/40

## Changes

Fix failing `TestBlockBuilder_CreateBlockFlow`

## Test Plan

- [x] `TestBlockBuilder_CreateBlockFlow` passes

## TODO

- [x] Finish

## DevOps Notes

### DIFF
--- a/miner/builder_test.go
+++ b/miner/builder_test.go
@@ -212,69 +212,98 @@ var (
 )
 
 func TestBlockBuilder_CreateBlockFlow(t *testing.T) {
-	net := service.NewSimulator()
-	beginRound := make(chan types.LayerID)
-	n := net.NewNode()
-	receiver := net.NewNode()
+	const iterations = 1000000
 
-	hare := MockHare{res: map[types.LayerID][]types.BlockID{}}
-	hare.res[1] = hareRes
+	for i := 1; i <= iterations; i++ {
+		t.Logf("iteration %v", i)
 
-	txPool := state.NewTxMemPool()
+		net := service.NewSimulator()
+		beginRound := make(chan types.LayerID)
+		n := net.NewNode()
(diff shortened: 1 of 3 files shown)


nkryuchkov:
bors try

antonlerner (approved):

miner/builder_test.go
```diff
@@ -269,7 +271,7 @@ func TestBlockBuilder_CreateBlockFlow(t *testing.T) {
 		assert.True(t, ContainsAtx(b.ATXIDs, atxs[2].ID()))*/
 
 		assert.Equal(t, []types.ATXID{atx1, atx2, atx3, atx4, atx5}, *b.ActiveSet)
-	case <-time.After(1 * time.Second):
+	case <-time.After(1 * time.Minute):
```
noamnelke, line 274: The other fix makes a lot of sense to me. I'm also ok with increasing the timeout, but I'm wondering if it isn't just a leftover from a first attempt at a fix - does it really take more than a second sometimes?

noamnelke (approved):

miner/builder_test.go
nkryuchkov: @noamnelke Thanks for noticing. No, it's not a leftover. Even with the other fix, sometimes, under high CPU load, spawning a goroutine takes multiple seconds, so the timeout occurs. I've run the test around 5M times and it took around 15 seconds around 2 times per 1M iterations. So, I think, 1 minute would give some reserve.

antonlerner (approved):

nkryuchkov:
bors merge

PR #2312, opened 2021-03-04
[Merged by Bors] - Fix failing TestBuilder_PublishActivationTx_HappyFlow

nkryuchkov:
## Motivation

WIP on https://github.com/spacemeshos/pm/issues/40

## Changes

Fix failing `TestBuilder_PublishActivationTx_HappyFlow`

## Test Plan

- [x] `TestBuilder_PublishActivationTx_HappyFlow` passes

## TODO

- [x] Finish

## DevOps Notes

### DIFF
--- a/activation/activation_test.go
+++ b/activation/activation_test.go
@@ -243,7 +243,7 @@ func (l *LayerClockMock) GetCurrentLayer() types.LayerID {
 func (l *LayerClockMock) AwaitLayer(types.LayerID) chan struct{} {
 	ch := make(chan struct{})
 	go func() {
-		time.Sleep(1 * time.Second)
+		time.Sleep(500 * time.Millisecond)
 		close(ch)
 	}()
 	return ch
--- a/activation/activation_test.go
+++ b/activation/activation_test.go
@@ -243,7 +243,7 @@ func (l *LayerClockMock) GetCurrentLayer() types.LayerID {
 func (l *LayerClockMock) AwaitLayer(types.LayerID) chan struct{} {
 	ch := make(chan struct{})
 	go func() {
-		time.Sleep(1 * time.Millisecond)
+		time.Sleep(1 * time.Second)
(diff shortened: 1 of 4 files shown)


nkryuchkov:
bors try

activation/activation_test.go
```diff
@@ -243,7 +243,7 @@ func (l *LayerClockMock) GetCurrentLayer() types.LayerID {
 func (l *LayerClockMock) AwaitLayer(types.LayerID) chan struct{} {
 	ch := make(chan struct{})
 	go func() {
-		time.Sleep(1 * time.Millisecond)
+		time.Sleep(1 * time.Second)
```
noamnelke: For something that's flaky (not failing consistently), is 1000x increase really required? Maybe 50ms will be enough to make it stable?

nkryuchkov: @noamnelke After running a few million iterations, I didn't see anything higher than 150ms. So, I rounded it up to the next power of 10. With lower values, the test can fail rarely, but still can.

nkryuchkov: @noamnelke Do you think something like 200ms or 500ms is worth trying? Or doesn't it make much difference if we increase the timeout so much?

antonlerner (approved):

activation/activation_test.go
noamnelke: This is not a timeout, but an actual sleep. A second isn't that much, so I leave it up to you, but based on what you say, maybe 300-500ms?

noamnelke (approved):

activation/activation_test.go
nkryuchkov: Yes, something times out because of it, so I meant this sleep. 500ms seems to work fine for 10K iterations

nkryuchkov:
bors try

nkryuchkov:
bors merge

Issue #2307, opened 2021-03-03
Integrate nodes API Querying In Automation


AmitShaul:
## Description
Currently automation using ES queries in order to validate tests scenarios, this issue is for integrating API queries instead of some of the ES queries to reduce our dependency in logs.

PR #2316, opened 2021-03-08
New API Integration In Automation 

AmitShaul:
## Motivation
Moving automation to use existing API endpoints instead of ES querying.
Some endpoints are missing especially on the hare side so we won't be able to migrate completely for now.

### issue
#2307


tests/api/api_handler.py
```diff
@@ -0,0 +1,111 @@
...
+            return current_layer
+        while True:
+            if current_layer == layer_num:
+                break
+            if is_timeout and timeout <= 0:
+                raise TimeoutError("wait for layer reached timeout")
+            print(f"waiting for layer {layer_num}")
+            print(f"sleeping for {interval} before querying for current layer again\n")
+            time.sleep(interval)
+            timeout = timeout - interval if is_timeout else timeout
```
antonlerner, line 49: besides my personal preference for the standart `if else` statements you have 2 checks on `if is_timeout` which I think can be refactored if you use this standart

tests/api/api_handler.py
```diff
@@ -0,0 +1,111 @@
...
+        self.sender = Sender(ips, namespace)
+
+    def extend_ips(self, ips):
+        self.sender.extend_ips(ips)
+
+    def remove_ips(self, ips):
+        self.sender.remove_ips(ips)
+
+    # TODO: add timeout decorator
+    def wait_for_layer(self, layer_num, timeout=None, interval=10):
```
antonlerner, line 23: 10 seconds is quite a long time IMO, can reduce to 1 sec

tests/api/api_handler.py
```diff
@@ -0,0 +1,111 @@
...
+            timeout = timeout - interval if is_timeout else timeout
+            current_layer, elapsed_time = self.sender.send_current_layer_timed(timeout)
+            timeout = timeout - elapsed_time if is_timeout else timeout
+        return current_layer
+
+    # TODO: add timeout
+    def wait_for_epoch(self, epoch_num):
+        curr_epoch = self.sender.send_current_epoch()
+        if curr_epoch > epoch_num:
+            raise ValueError(f"epoch requested is {epoch_num} has already passed, current epoch {curr_epoch}")
```
antonlerner, line 58: or we can just return

tests/api/parser.py
```diff
@@ -0,0 +1,40 @@
+
+# ################ MESH ################
+
+
+# TODO: this and the following function should return a list of values and not issue only the first object
+def parse_layer_hash_from_layer_query_res(layer_query_response):
+    # {..., 'layer': [{'hash': '[\d\w=]+', ...}]}
+    return layer_query_response['layer'][0]['hash']
```
antonlerner, line 8: what should happen if the value is not found? just throw an exception?

tests/api/sender.py
```diff
@@ -0,0 +1,139 @@
...
+            p.start()
+        # Exit the completed processes
+        for p in processes:
+            p.join()
+        return return_list
+
+    # epoch_num_layers = "v1/mesh/epochnumlayers"
+    def send_layers_per_epoch(self):
+        res = self.send("{}", "epoch_num_layers")
+        return int(parser.parse_layer_per_epoch_msg(res))
```
antonlerner, line 63: consider handling the exception here if value not found

tests/assertions/mesh_assertion.py
```diff
@@ -0,0 +1,95 @@
...
+    :param num_miners: int, number of miners
+    :param ignore_lst: list, a list of pod names to be ignored
+
+    """
+    # layers count start from 0
+    if from_layer == 0:
+        print(f"refactoring starting layer from 0 to {layers_per_epoch}, not validating first epoch")
+        from_layer = layers_per_epoch * 2
+        if from_layer == to_layer:
+            return
```
antonlerner: print that test will not execute because `from_layer == to_layer1`

tests/api/api_handler.py
noamnelke, line 23: We shouldn't use polling. The new API has a layers stream (`api/grpcserver/mesh_service.go -> LayerStream`).

tests/api/parser.py
noamnelke:
```suggestion
    return [h['hash'] for h in layer_query_response['layer']]
```

This should take care of the TODO, while also not raising an exception if the list in `layer` is empty (just returns an empty list in that case).

It still raises an exception if `layer` is missing or a value in the list is missing the `hash` key. That seems like reasonable behaviour to me, but if you think it shouldn't raise an exception in that case, you can replace the `['layer']` with `.get('layer', [])` and/or the `['hash']` with `.get('hash')`.

noamnelke: You can use the same list comprehension in the following methods, too.

tests/elk/elastic/2bm86_elastic_stateful.yaml
```diff
@@ -0,0 +1,194 @@
...
+            - name: node.ingest
+              value: "true"
+            - name: node.master
+              value: "true"
+            - name: rest.action.multi.allow_explicit_index
+              value: "true"
+            - name: ELASTIC_USERNAME
+              value: elastic
+            - name: ELASTIC_PASSWORD
+              value: aGeWkp56b
```
noamnelke: We must change this password now 😔

I assume this file wasn't supposed to be committed?

tests/late_nodes/delayed_config.yaml
```diff
@@ -41,7 +41,7 @@ bootstrap:
 client:
   image: 'spacemeshos/go-spacemesh:develop'
   command: '/bin/go-harness'
-  replicas: 49
+  replicas: 20
```
noamnelke, line 45: why?

noamnelke (changes requested):

tests/late_nodes/delayed_config.yaml
AmitShaul: this is the value from the original test

tests/elk/elastic/2bm86_elastic_stateful.yaml
AmitShaul: it doesn't really matters since we use HTTP for communication instead of HTTPS

AmitShaul: removing this file

tests/api/api_handler.py
AmitShaul: for that we need a GRPC service instead of the CURL we have now

tests/elk/elastic/2bm86_elastic_stateful.yaml
noamnelke: Not using SSL (using `http` instead of `https`) means that if you're on a non-secure network (e.g. a wifi at a cafe with no password) anyone on the local network can eavesdrop and see your password in plaintext. This also makes man-in-the-middle attacks possible (dns-hijacking, or malicious ISP, etc.).

None of these are very likely, but we should protect against them.

Having the password exposed publicly (even after the file was removed) means that bots that monitor these kinds of things now have it, making it meaningless to have a password in the first place. Let's please replace it or not use a password at all.

AmitShaul: replacing this will take all developers changing their local env variable.
I will change the file in 1pass but everyone should download and source it.
as far as I know some developers use their own passwords.

tests/assertions/mesh_assertion.py
AmitShaul: fixed

tests/api/sender.py
AmitShaul: I can add a general validation for result in 'send' and 'send_all' functions under sender.
this specific endpoint should always return a valid response since all miners should know what epoch they're at.

tests/api/parser.py
AmitShaul: I need only 1 value in this case, the name of the function implies one value as well.
next I will add a multiple value return function that will do what you suggested here.

tests/api/api_handler.py
AmitShaul: I would rather keep this as it is, it's more tight that way, it might float a bug when the developer entered wrong values unintentionally.

tests/api/api_handler.py
AmitShaul: checking if 'is_timeout' to see if the timeout is relevant since its default value is None.
first check is for validating we didn't exceed timeout and the second is for recalculating timeout after sleeping for our interval between retries.

tests/elk/elastic/2bm86_elastic_stateful.yaml
AmitShaul: another issue will be replacing the password in our current main Kibana.

AmitShaul: once I integrate HTTPS I will change all passwords of course

AmitShaul:
bors try

noamnelke:
bors try

noamnelke:
bors try-

noamnelke:
bors try

tests/elk/elastic/2bm86_elastic_stateful.yaml
noamnelke: If you plan to implement SSL in the next few days then fine, but in the meantime we should consider this password to be compromised and the current state of any service that uses this password is that it's unprotected.

If it'll take more than a couple of days, then I think the effort to replace this password is well worth it.

tests/api/api_handler.py
noamnelke: Yes, I agree, we should switch to using gRPC directly instead of using the JSON gateway.

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

lrettig:
> Some endpoints are missing especially on the hare side so we won't be able to migrate completely for now.

Is there a place where I can see the list of unmet dependencies? Maybe in the parent issue?

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try-

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

noamnelke (approved):

AmitShaul:
bors try

AmitShaul:
bors try

nkryuchkov:
@AmitShaul Do we still need to merge this PR? Do only merge conflicts and failing tests block this PR from merging?

PR #2323, opened 2021-03-11
[Merged by Bors] - Contextual logging

### LINKED ISSUE #2322: Contextual logging
Right now, we use per-module loggers, which is nice, but what we really need is contextual logging that lets us track and tie together a single thread of processing (for, e.g., an incoming message). We can achieve this using go contexts (which we're already using in some places). Here are some helpful resources:

- https://blog.gopheracademy.com/advent-2016/context-logging/
- https://dave.cheney.net/tag/logging
- https://github.com/wacul/ulog
- https://www.cockroachlabs.com/blog/enriching-log-messages-using-go-contexts/

lrettig:
## Motivation
Closes #2322 

## Changes
- Threads context through network and message handling for debug purposes
- Adds app context-related helper methods
- Adds ability to add context to logging

## Test Plan
- [x] Add unit tests for log context helpers
- [x] Add an integration test that tracks a requestId from top to bottom

## TODO


p2p/net/network.go
```diff
@@ -290,27 +293,31 @@ func (n *Net) accept(listen net.Listener) {
...
 			err := c.setupIncoming(n.config.SessionTimeout)
 			if err != nil {
-				n.logger.Event().Warning("conn_incoming_failed", log.String("remote", c.remoteAddr.String()), log.Err(err))
+				n.logger.Event().Warning("incoming connection failed",
+					log.String("remote_addr", c.remoteAddr.String()),
+					log.Err(err))
 				return
 			}
-			go c.beginEventProcessing()
+			go c.beginEventProcessing(context.TODO())
```
noamnelke: Why not use `context.Background()`? Do you think we should pass a context from the outside here in the future?

p2p/net/udp.go
```diff
@@ -259,12 +266,13 @@ func (n *UDPNet) listenToUDPNetworkMessages(listener net.PacketConn) {
 			mconn := newMsgConnection(conn, n, pk, ns, n.config.MsgSizeLimit, n.config.DialTimeout, n.logger)
 			n.publishNewRemoteConnectionEvent(mconn, node.NewNode(pk, net.ParseIP(host), 0, uint16(iport)))
 			n.addConn(addr, conn)
-			go mconn.beginEventProcessing()
+			go mconn.beginEventProcessing(context.TODO())
```
noamnelke, line 291: Same here, it's fine to leave `TODO` if you think we should create an app-level context and pass it in here, just wanted to know your thoughts. In general, I think we should use/pass contexts more broadly and we should probably do it as part of the graceful shutdown work (using Tomb).

noamnelke (approved):

noamnelke:
Just noticed that it's still a draft... 🤭

p2p/net/network.go
lrettig: you're right, this one is not a TODO. fixed.

p2p/net/udp.go
lrettig: cleaned this one up too. agree re: graceful shutdown work, context objects can be used for that too, and this work should make that easier, but for now I'm focused just on logging.

lrettig:
bors try

activation/atxdb.go
```diff
@@ -746,38 +747,38 @@ func (db *DB) ValidateSignedAtx(pubKey signing.PublicKey, signedAtx *types.Activ
...
 
 	//todo fetch from neighbour (#1925)
 	if atx.Nipst == nil {
-		db.log.Panic("nil nipst in gossip")
+		logger.Panic("nil nipst in gossip")
 		return fmt.Errorf("nil nipst in gossip")
 	}
 
-	if err := syncer.GetPoetProof(atx.GetPoetProofRef()); err != nil {
+	if err := syncer.GetPoetProof(context.TODO(), atx.GetPoetProofRef()); err != nil {
```
noamnelke:
```suggestion
	if err := syncer.GetPoetProof(ctx, atx.GetPoetProofRef()); err != nil {
```

Or am I missing something?

If I am, I guess it should still not be `TODO()`

activation/poetlistener.go
```diff
@@ -74,7 +75,7 @@ func (l *PoetListener) handlePoetProofMessage(gossipMessage service.GossipMessag
 		return
 	}
 
-	gossipMessage.ReportValidation(PoetProofProtocol)
+	gossipMessage.ReportValidation(context.TODO(), PoetProofProtocol)
```
noamnelke:
```suggestion
	gossipMessage.ReportValidation(ctx, PoetProofProtocol)
```

activation/activation.go
```diff
@@ -244,7 +245,7 @@ func (b *Builder) buildNipstChallenge(currentLayer types.LayerID) error {
 
 // StartPost initiates post commitment generation process. It returns an error if a process is already in progress or
 // if a post has been already initialized
-func (b *Builder) StartPost(rewardAddress types.Address, dataDir string, space uint64) error {
+func (b *Builder) StartPost(ctx context.Context, rewardAddress types.Address, dataDir string, space uint64) error {
```
AmitShaul, line 248: is Context necessary here?
I don't see any use of it in this function

activation/poetlistener.go
```diff
@@ -26,7 +27,7 @@ type PoetListener struct {
 }
 
 // Start starts listening to PoET gossip messages.
-func (l *PoetListener) Start() {
+func (l *PoetListener) Start(ctx context.Context) {
```
AmitShaul, line 30: same here, is it necessary to have Context here without use?

hare/broker.go
```diff
@@ -141,77 +142,70 @@ func (b *Broker) validate(m *Message) error {
...
+			// create an inner context object to handle this message
+			messageCtx := ctx
+
+			// try to read stored context
+			if msg.RequestID() == "" {
+				logger.With().Warning("broker received hare message with no registered requestID")
+			} else {
+				messageCtx = log.WithRequestID(ctx, msg.RequestID())
+			}
+
```
AmitShaul, line 165: I think that moving line 156 into the first if statement on line 159 will be more clear, it's a matter of semantics so it's not an issue if you decide to leave it this way.

activation/activation.go
```diff
@@ -3,6 +3,7 @@
 package activation
 
 import (
+	"context"
```
AmitShaul, line 6: you sometimes import "golang.org/x/net/context" (like in gateway_service.go) and sometimes "context"

hare/proposaltracker.go
```diff
@@ -72,16 +73,19 @@ func (pt *proposalTracker) OnLateProposal(msg *Msg) {
...
 			pt.isConflicting = true
 		}
 	}
 
 	// not equal check rank
 	// lower ranked proposal on late proposal is a conflict
 	if bytes.Compare(msg.InnerMsg.RoleProof, pt.proposal.InnerMsg.RoleProof) < 0 {
-		pt.With().Info("late lower rank detected", log.String("id_malicious", msg.PubKey.String()))
+		pt.With().Info("late lower rank detected",
+			log.String("id_malicious", msg.PubKey.String()))
```
AmitShaul, line 88: why the new line?:)

sync/validation_queue.go
```diff
@@ -253,14 +254,12 @@ func (vq *blockQueue) addDependencies(jobID interface{}, blks []types.BlockID, f
...
+		logger.With().Debug("adding dependencies to pending queue",
+			log.Int("count", len(idsToPush)))
+		vq.addToPending(ctx, idsToPush)
 	}
 
-	vq.With().Debug("finished adding dependencies",
-		log.Int("count", len(dependencies)),
-		log.String("job_id", fmt.Sprintf("%v", jobID)))
+	logger.With().Debug("finished adding dependencies",
+		log.Int("count", len(dependencies)))
```
AmitShaul: did you intentionally remove the job_id?

p2p/gossip/protocol.go
```diff
@@ -105,32 +107,31 @@ func (p *Protocol) markMessageAsOld(h types.Hash12) bool {
 	return p.oldMessageQ.GetOrInsert(h)
 }
 
-func (p *Protocol) processMessage(sender p2pcrypto.PublicKey, protocol string, msg service.Data) error {
+func (p *Protocol) processMessage(ctx context.Context, sender p2pcrypto.PublicKey, protocol string, msg service.Data) error {
 	h := types.CalcMessageHash12(msg.Bytes(), protocol)
-	fields := []log.LoggableField{
+	logger := p.WithContext(ctx).WithFields(
 		log.String("from", sender.String()),
 		log.String("protocol", protocol),
```
AmitShaul: indentation

activation/activation.go
lrettig: Thanks for noticing. Actually, I didn't add the context imports to the GRPC code in this PR. It was there from before. The google GRPC stack may be doing something different with context so I think this can be left alone.

activation/activation.go
lrettig: fixed!

activation/poetlistener.go
lrettig: fixed

hare/broker.go
lrettig: That won't work, it needs to be in scope outside the if statement

hare/proposaltracker.go
lrettig: :shrug:

sync/validation_queue.go
lrettig: it's added to the context on line 214, above, so it will appear in all log lines - the beauty of contextual logging :)

p2p/gossip/protocol.go
lrettig: what do you mean? looks okay to me.

activation/activation.go
noamnelke: It looks like `golang.org/x/net/context` is (unofficially) deprecated. We should probably use the package from the standard library (assuming we require Go 1.7+, otherwise we can't use it at all).

https://github.com/grpc-ecosystem/go-grpc-middleware/issues/175

(doesn't have to be in this PR).

cmd/node/node.go
```diff
@@ -456,7 +449,8 @@ func (app *SpacemeshApp) SetLogLevel(name, loglevel string) error {
 	return nil
 }
 
-func (app *SpacemeshApp) initServices(nodeID types.NodeID,
+func (app *SpacemeshApp) initServices(ctx context.Context,
```
noamnelke, line 453: while we pass the `ctx` along, we don't use it for logs in this method - shouldn't we?

(i.e. `logger := app.log.WithContext(ctx)` + `s/app.log/logger/`)

cmd/node/node.go
```diff
@@ -706,37 +702,33 @@ func (app *SpacemeshApp) HareFactory(mdb *mesh.DB, swarm service.Service, sgn ha
 	return ha
 }
 
-func (app *SpacemeshApp) startServices() {
-	//app.blockListener.Start()
-	go app.startSyncer()
+func (app *SpacemeshApp) startServices(ctx context.Context) {
+	logger := log.AppLog.WithContext(ctx)
```
noamnelke: I think this is the first use of `log.AppLog` outside of testing code. I don't see how this is any worse than using bare calls logging methods with no logger, but just making sure you're aware. Consider passing a "real" logger in here?

layerfetcher/layers.go
```diff
@@ -172,23 +149,23 @@ func (l *Logic) PollLayer(layer types.LayerID) chan LayerPromiseResult {
...
-			l.receiveLayerHash(layer, p, len(peers), b, nil)
+			l.receiveLayerHash(ctx, layer, p, len(peers), b, nil)
 		}
 
 		timeoutFunc := func(err error) {
-			l.receiveLayerHash(layer, p, len(peers), nil, err)
+			l.receiveLayerHash(ctx, layer, p, len(peers), nil, err)
 		}
-		err := l.net.SendRequest(LayerBlocksDB, layer.Bytes(), p, receiveForPeerFunc, timeoutFunc)
+		err := l.net.SendRequest(context.TODO(), LayerBlocksDB, layer.Bytes(), p, receiveForPeerFunc, timeoutFunc)
```
noamnelke: missed a spot 🧽

```suggestion
		err := l.net.SendRequest(ctx, LayerBlocksDB, layer.Bytes(), p, receiveForPeerFunc, timeoutFunc)
```

layerfetcher/layers.go
```diff
@@ -243,15 +220,15 @@ func (l *Logic) receiveLayerHash(id types.LayerID, p p2ppeers.Peer, peers int, d
...
-			l.receiveBlockHashes(id, data, nil)
+			l.receiveBlockHashes(ctx, id, data, nil)
 		}
 		remainingPeers := 0
 		errFunc := func(err error) {
-			l.receiveBlockHashes(id, nil, err)
+			l.receiveBlockHashes(ctx, id, nil, err)
 		}
-		err := l.net.SendRequest(LayerHashDB, hash.Bytes(), peer[remainingPeers], receiveForPeerFunc, errFunc)
+		err := l.net.SendRequest(context.TODO(), LayerHashDB, hash.Bytes(), peer[remainingPeers], receiveForPeerFunc, errFunc)
```
noamnelke: Either you consistently miss this spot, or I'm missing something?

```suggestion
		err := l.net.SendRequest(ctx, LayerHashDB, hash.Bytes(), peer[remainingPeers], receiveForPeerFunc, errFunc)
```

layerfetcher/layers.go
```diff
@@ -331,11 +298,11 @@ func (l *Logic) getPoetResult(hash types.Hash32, data []byte) error {
 
 // blockReceiveFunc handles blocks received via fetch
 func (l *Logic) blockReceiveFunc(data []byte) error {
-	return l.blockHandler.HandleBlockData(data, l)
+	return l.blockHandler.HandleBlockData(context.TODO(), data, l)
```
noamnelke: I guess this should be fixed

p2p/message.go
```diff
@@ -33,6 +33,7 @@ type gossipProtocolMessage struct {
 	sender         p2pcrypto.PublicKey
 	data           service.Data
 	validationChan chan service.MessageValidation
+	requestID      string
```
noamnelke, line 36: If I'm understanding the flow correctly, this means we'll be able to log the request ID of a message that originated in the same node as it passes through the processing pipeline before being gossipped (intra-node).

We'll only be able to track messages inter-node for Hare messages, where you've added the request ID to the message on the wire.

Is this the intended behaviour?

hare/algorithm.go
```diff
@@ -95,7 +96,8 @@ type StateQuerier interface {
 // Messages are sent as type Message. Upon receiving, the public key is added to this wrapper (public key extraction).
 type Msg struct {
 	*Message
-	PubKey *signing.PublicKey
+	PubKey    *signing.PublicKey
+	RequestID string
```
noamnelke, line 100: I'm wondering if the addition of a field to every gossip message on the wire for the sole purpose of easier to follow logs is worth it... 🤔

My intuition is that it isn't worth it and that we should do the best we can with the minimal messages required to operate the protocol.

Have we (the team) discussed this before implementation? Maybe I just don't remember, or was absent, and we did. It seems like a pretty big decision to me, worth discussing.

p2p/net/udp.go
```diff
@@ -84,7 +84,7 @@ func (n *UDPNet) SubscribeClosingConnections(f func(connection ConnectionWithErr
 func (n *UDPNet) publishClosingConnection(connection ConnectionWithErr) {
 	n.clsMutex.RLock()
 	for _, f := range n.closingConnections {
-		f(connection)
+		f(context.TODO(), connection)
```
noamnelke, line 102:
```suggestion
		f(ctx, connection)
```

p2p/net/udp.go
noamnelke: >cleaned this one up

?

sync/block_listener.go
```diff
@@ -36,9 +37,9 @@ func (bl *BlockListener) Close() {
 }
 
 // Start starts the main listening goroutine
-func (bl *BlockListener) Start() {
+func (bl *BlockListener) Start(ctx context.Context) {
```
noamnelke: Why do we keep this commented-out code in the repo? (this entire file)

noamnelke (commented):
My main issue with this change is that it changes the Hare gossip messages on the wire for the purpose of nicer logging. I _think_ that only Hare messages were changed, maybe I missed something?

p2p/net/udp.go
lrettig: cleaned up as in it's no longer a TODO

activation/activation.go
lrettig: okay, will fix this, thanks guys

activation/atxdb.go
lrettig: already caught and fixed this

sync/block_listener.go
lrettig: No idea, that was part of #2128 - @antonlerner ? I'll remove them here, this is sort of already a "general cleanup" project anyway

p2p/net/udp.go
lrettig: Won't work, there is no ctx here. I didn't try to add context _everywhere_, including the full p2p stack. That's what context.TODO is for :) Feel free to finish threading it through this stack... (this can be done incrementally in future - e.g., when debugging issues, where it's very helpful)

hare/algorithm.go
lrettig: No, that wasn't my intention, and no, we didn't discuss this. RequestIDs (and SessionIDs) are strictly local and have no meaning on the wire, or to another node, so they should not be sent on the wire. The field I added here, and to the gossip message, is just used to convey context across different goroutines/contexts locally. Let me rethink this - I might be able to achieve the same thing a different way, or else, I can wrap the on-the-wire messages in a struct that's passed around locally if necessary. Thanks for catching this.

p2p/message.go
lrettig: RequestID isn't added to the message on the wire. Only the `data` member here is serialized.

hare/algorithm.go
lrettig: Took a closer look. As with the above comment (gossipProtocolMessage), RequestID is not sent on the wire. Only the inner `Message` member here is serialized.

lrettig:
> My main issue with this change is that it changes the Hare gossip messages on the wire for the purpose of nicer logging. I _think_ that only Hare messages were changed, maybe I missed something?

See inline replies to questions - I don't think the format of any messages has changed on the wire. These are all local only. I think all the concerns raised here have been addressed.

cmd/node/node.go
lrettig: done, and done throughout this module - but I think there are probably still a few other scattered places where I did the same thing. We're not very consistent with how log objects are passed around today, that still needs to be improved.

p2p/net/udp.go
noamnelke: yes, it is 😬

hare/algorithm.go
noamnelke: You're right, I missed this.

noamnelke (approved):

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
I can't understand for the life of me why the p2p systest is failing here, but it always fails with this error:

```
E       AssertionError: msg_testing: Total gossip messages in ES is not as expected
E         expected 4440, got 4300
E       assert 4440 == 4300
E         -4440
E         +4300
```

(The number of messages received varies from run to run but it's always < 4400.)

I don't see any errors or warnings in the logs. I double-checked all of the p2p-related changes here, and there are no changes to logic that I can find that would cause this sort of error - there are hardly any logic changes at all in this PR, it's almost entirely cosmetic. I tried increasing the timeout, and that also didn't help. I'm stumped, and I'm blocked on doing further investigation by a kibana issue.

lrettig:
bors merge

lrettig:
Walked through all of the code changes today. Confirmed that there are no logic changes, AFAICT, with one small exception: the timeout in fetch. I tried reverting this change, to no avail. Also confirmed that there's no change to any of the test code. I am totally stumped as to what's causing this failure. Either I'm missing something, or else it's a complex interaction, e.g., a timing issue or something. The only thing I can think of trying is a binary search to isolate the breaking change.

lrettig:
bors merge

lrettig:
bors merge-

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge-

lrettig:
bors merge

lrettig:
bors merge

PR #2331, opened 2021-03-11
[Merged by Bors] - Use commithash as docker tag for staging

### LINKED ISSUE #2328: Dockerhub tag name is reused, potentially causing issues
## Description
System tests use an image on Dockerhub that's overwritten with different code each time. When using bors, it's always either `spacemeshos/go-spacemesh-dev:staging` or `spacemeshos/go-spacemesh-dev:trying` and for local runs it's an image with the name of the branch.

This could cause issues, because when running the tests we rely on the assumption that the image there is up to date and there are edge cases where it wouldn't be.

## Affected code

This issue appears in commit hash: `f156138`

https://github.com/spacemeshos/go-spacemesh/blob/f15613812df68543e4cffb27c11049af801daa71/Makefile#L232-L235

lrettig:
## Motivation
Closes #2328

## Changes
For staging branch, i.e., for `bors merge` runs, use commithash rather than branch name (`staging`) as docker image tag to ensure that bors is running tests against the right code

## Test Plan
N/A


noamnelke (changes requested):
The `trying` branch should be treated the same way, for the same reason.

That would also solve a different big issue we're having with `try`s right now: if we have a random failure right now, we must issue a new `try`, since re-running the tests in GHA will use whatever image is currently in the `trying` Dockerhub image, which may no longer be what's relevant. If we submit an image tagged with the commit hash and then use that for the system tests, the GHA run will be reproducible and we won't need to issue a new `bors try`.

lrettig:
Updated. Let's test and merge if it's working.

lrettig:
bors try

lrettig:
bors try

noamnelke (approved):

noamnelke:
bors try

AmitShaul:
bors try-

AmitShaul:
bors try-

noamnelke:
bors try

lrettig:
bors merge

lrettig:
bors merge

PR #2354, opened 2021-03-28
[Merged by Bors] - Terminate old hare processes if max concurrent processes are reached

### LINKED ISSUE #2303: Hare broker should terminate old processes if max iterations reached
This should happen in FIFO fashion: if a new hare process (for a new layer) is kicked off, but the max number of concurrent hare process iterations is reached, then the oldest (i.e., earliest layer) process should be terminated.

sidsankhe:
## Motivation

Closes #2303 

## Changes

Changed the Hare broker so that on registering a protocol, if we have reached the maximum number of concurrent Hare processes, we terminate the process that was started earliest.
## Test Plan

Added unit tests to test the above behavior
## TODO

## DevOps Notes


hare/broker.go
```diff
@@ -308,9 +309,12 @@ func (b *Broker) Register(id instanceID) (chan *Msg, error) {
 		b.updateLatestLayer(id)
 
 		if len(b.outbox) >= b.limit {
-			resErr <- errTooMany
-			resCh <- nil
-			return
+			//unregister the earliest layer to make space for the new layer
+			//cannot call unregister here because unregister blocks and this would cause a deadlock
+			id := b.minDeleted + 1
```
lrettig: Don't overload the `id` variable here, it makes this code hard to read. Use a different name.

hare/broker.go
```diff
@@ -308,9 +309,12 @@ func (b *Broker) Register(id instanceID) (chan *Msg, error) {
...
+			delete(b.outbox, id)
```
lrettig: Don't duplicate the code from the `Unregister` method. Refactor that method to create a new (internal) method that does not block, that you can call here directly.

lrettig (changes requested):

hare/broker.go
```diff
@@ -308,9 +309,11 @@ func (b *Broker) Register(id instanceID) (chan *Msg, error) {
 		b.updateLatestLayer(id)
 
 		if len(b.outbox) >= b.limit {
-			resErr <- errTooMany
-			resCh <- nil
-			return
+			//unregister the earliest layer to make space for the new layer
+			//cannot call unregister here because unregister blocks and this would cause a deadlock
+			instance := b.minDeleted + 1
```
lrettig: I think you only want to unregister the oldest layer after the `b.isSynced(id)` check, below

lrettig (changes requested):
One more change requested. Overall looks good. One more quick thing: are you 100% sure that this code is concurrency-safe? What if, e.g., `Register()` were called back to back in rapid succession, before the first task were completed? I think you either need to add some sort of concurrency safety or, if you're sure you don't need it, add a comment to this effect and explain why.

sidsankhe:
The good thing with this code is that the tasks are submitted into a channel, and an event loop executes it synchronously. So even if `Register` were called quickly in succession, the tasks would be added to the event loop and would be executed synchronously.

lrettig:
> The good thing with this code is that the tasks are submitted into a channel, and an event loop executes it synchronously. So even if `Register` were called quickly in succession, the tasks would be added to the event loop and would be executed synchronously.

But the order isn't guaranteed, is it?

sidsankhe:
No, the order isn't guaranteed. However, since `Register` is called only on ticks, it should be very unlikely that there's a race condition such that it involves a later layer ID being Registered before an earlier layer ID. Even so, the `b.UpdateLatestLayer(id)` would panic if an earlier layer was trying to be Registered after a later layer. Do you think this check needs to be made stronger?

lrettig:
> No, the order isn't guaranteed. However, since `Register` is called only on ticks, it should be very unlikely that there's a race condition such that it involves a later layer ID being Registered before an earlier layer ID. Even so, the `b.UpdateLatestLayer(id)` would panic if an earlier layer was trying to be Registered after a later layer. Do you think this check needs to be made stronger?

I agree with your logic, it seems highly unlikely and therefore okay. Could you add an explanatory comment to this effect to the code? (If the code were refactored later, this assumption might not hold, so let's state our assumptions clearly.)

lrettig (approved):
tweaked the comment wording slightly, otherwise lgtm

sidsankhe:
bors merge

PR #2357, opened 2021-03-30
[Merged by Bors] - Remove circular dependency in Hare active set/eligibility

### LINKED ISSUE #2356: Hare active set size calculation is fragile
## Description
If hare happens to fail once (for any reason) for a layer that's an epoch boundary, it will continue to fail forever beginning with the following epoch. This is due to the way safe layers are used to calculate active set size, which is used in hare eligibility calculations.

When hare fails for the layer in question, all of that layer's blocks are eventually marked as contextually invalid (see #2298, #2337). This is expected behavior and I don't think it will change with self-healing. Hare needs to be able to recover from this scenario.

This caused testnet 126 to fail.

## Expected Behavior
Hare eligibility should not depend on the success of a previous Hare round.

lrettig:
## Motivation
Fixes #2356

## Changes
When calculating active set for a given epoch for hare eligibility, use the Tortoise active set rather than Hare active set. Hare active set consists of the set of unique node IDs that produced contextually valid blocks for a "safe" layer in the past. The problem is that this causes a circular dependency: if Hare fails for the "safe" layer, then the active set size for a future layer will be zero, causing the beacon to fail and eligibility to always be false. The Tortoise active set, by contrast, consists of the set of unique node IDs that produced valid ATXs targeting the epoch for the layer. This is much more robust against Hare failure and removes the circular dependency.

## Test Plan
Updates existing tests accordingly

## TODO

## DevOps Notes


lrettig:
bors try

lrettig:
Close in favor of #2368

lrettig:
Decided to go with this patch after all, see https://github.com/spacemeshos/go-spacemesh/pull/2368#issuecomment-818127924. Reopening.

blocks/blockoracle.go
```diff
@@ -15,10 +15,10 @@ import (
 )
 
 type activationDB interface {
-	GetNodeAtxIDForEpoch(nodeID types.NodeID, targetEpoch types.EpochID) (types.ATXID, error)
```
noamnelke, line 16: I think the variable names here serve as documentation. Specifically `targetEpoch` can reduce a lot of confusion. Alternatively, possibly better, we can change the name of the method to `GetNodeAtxIDForTargetEpoch`.

Just in case it's not obvious, the source of potential confusion is that the ATX has a publication epoch and separate target epoch.

hare/eligibility/oracle.go
```diff
@@ -251,56 +240,44 @@ func (o *Oracle) Proof(ctx context.Context, layer types.LayerID, round int32) ([
...
-			layer.GetEpoch(),
-			log.FieldNamed("safe_layer_id", sl),
-			log.FieldNamed("safe_epoch_id", safeEp))
-		return nil, err
+	// extract the nodeIDs
+	activeMap := make(map[string]struct{}, len(atxs))
+	for _, atxid := range atxs {
+		atx, err := o.atxdb.GetAtxHeader(atxid)
+		if err != nil {
+			o.With().Error("failed to fetch atx for layer in non-genesis epoch",
```
noamnelke: We have a "standard" error message for cases like this (where we have a reference stored, and then the object itself is missing). I also think that mentioning that it's non-genesis is implied/irrelevant - we'll never try to fetch an ATX for a genesis epoch (since there are none).
```suggestion
			o.With().Error("inconsistent state: failed to get atx header",
```

hare/eligibility/oracle.go
```diff
@@ -251,56 +240,44 @@ func (o *Oracle) Proof(ctx context.Context, layer types.LayerID, round int32) ([
...
+				layer,
+				layer.GetEpoch(),
+				log.Err(err))
+			return nil, err
```
noamnelke:
```suggestion
			return nil, fmt.Errorf("inconsistent state: failed to get atx header: %v", err)
```

hare/eligibility/oracle.go
noamnelke: I'm actually wondering if we should log this error here, at all. All consumers of this method log the error if one happens, so this will just be duplicated.

In general, to avoid noisy, redundant logging, I think our policy should be to only log errors where we stop propagating them, unless there's useful context that's not available up the call stack. In this case, the layer is available (and already logged).

noamnelke (approved):
Here's the problem: we'll check for actives the first time Hare runs in an epoch (first layer). We include all known ATXs, not just those referenced in blocks. Here's the "dangerous" scenario: a malicious miner publishes their ATX (which may have substantial weight) late, during the second layer of the target epoch. Online miners will ignore it, because they cached the list of ATXs during the first layer. Miners that were offline for a while and sync to catch up will know about the additional ATX before they fill the cache and will include it. This is a trivial way to cause a byzantine chain-split with no need for delaying messages or knowing anything about the network topology. 

Since there's no need to traverse blocks to get the list of ATXs, calling `actives()` for the first time is now not particularly expensive, so I'm wondering if keeping the `activesCache` is even worth it.

Without the cache it's harder to split the miners, but for a short while it's still possible (by timing the message and distributing to specific peers).

I guess the research team has to decide if we're ok with this change. I'm approving it, assuming you won't merge until we get a final, definitive ok from them.

hare/eligibility/oracle.go
```diff
@@ -251,56 +240,44 @@ func (o *Oracle) Proof(ctx context.Context, layer types.LayerID, round int32) ([
...
 	o.lock.Lock()
 	defer o.lock.Unlock()
 
-	// check cache
-	if val, exist := o.activesCache.Get(safeEp); exist {
-		return val.(map[string]struct{}), nil
+	// we can't read blocks during genesis epochs as there are none
+	if layer.GetEpoch().IsGenesis() {
+		return nil, errGenesis
 	}
```
noamnelke: This probably isn't relevant based on what was discussed today, but I'll still point it out.

This PR removes the need to read any blocks, so there's no need to return this error anymore. Everything below should work during genesis.

lrettig:
Based on our conversation today, this PR will require some changes:

- we introduce two new params, which I'm going to call `adist` and `bdist` (they don't have permanent names yet). the first is the number of layers after the start of a new epoch that we wait to receive blocks for the new epoch (for the purpose of establishing Hare eligibility), and the second is the number of layers we wait for consensus to be established on those blocks. for purposes of illustration, you can think of `adist` and `bdist` of being on the order of 100 layers.
- for the first `adist+bdist` layers of epoch `n`, we use the Hare active set from epoch `n-1`. after this point, we calculate the active set for epoch `n` as the count of unique ATXs referenced by valid blocks in the first `adist` layers in epoch `n`.
- note that this is _not_ a sliding window. we only consider blocks in the first `adist` layers of an epoch in the active set calculation for that epoch, and only after the first `adist+hdist` layers of each epoch.
- to address the circular dependency issue: in the (unlikely) case where we have an entire epoch with no valid blocks, e.g., if Hare fails continually for an entire epoch, then as a fallback we instead use the "local view" of the active set: the set of all ATXs received for the previous epoch. as @noamnelke pointed out, nodes may disagree in their view of this active set, which is a problem for Hare but not a problem for the self-healing weak coin, which relies on Hare eligibility.
- for purposes of bootstrapping at genesis, use this fallback option for the first `adist+hdist` layers of epoch 2 (the third epoch), which is the first epoch with blocks.

noamnelke:
@lrettig great summary, thanks!

Let's better define the trigger for the fallback mechanism: If by the time we reach layer `adist+bdist` of an epoch we've reached consensus on at least one block in a layer up to `adist` we don't need the fallback mechanism, otherwise we do. An _entire epoch_ with no valid blocks is a subset of the cases that require the fallback mechanism.

What does an adversary need to do to trigger the fallback mechanism? Either (a) suppress all blocks before `adist`; or (b) cause the Hare to converge on an empty set during all layers in `adist`; or (c) cause the Hare not to converge on any layer until `adist+bdist` is reached (any combination of the above, as long as no contextually valid block is included before `adist` by the time `adist+bdist` is reached).

Since we need consensus on using the fallback mechanism, `bdist` must be large enough that option c above is impossible (in other words, it must be greater than `maxIterations` expressed in layers), otherwise, if Hare converges after the fallback was triggered, some miners may not use the fallback.

---

Regarding your last point - I thought we decided to use self-healing in genesis, rather than the fallback, no?

There was a problem with bootstrapping self-healing, since it uses Hare messages for the beacon, but we said we'd send a round of Hare messages or a special beacon message and then self-healing would work and be more predictable than optimistically using the fallback mechanism, which might fail and trigger self-healing anyway.

lrettig:
> Regarding your last point - I thought we decided to use self-healing in genesis, rather than the fallback, no?
> 
> There was a problem with bootstrapping self-healing, since it uses Hare messages for the beacon, but we said we'd send a round of Hare messages or a special beacon message and then self-healing would work and be more predictable than optimistically using the fallback mechanism, which might fail and trigger self-healing anyway.

I'm talking about bootstrapping self-healing, not Hare. I think we're on the same page here. I'm not worried about bootstrapping Hare. As you pointed out, self-healing uses Hare preround messages for the weak coin, so there is a chicken-and-egg problem here at genesis. What I meant by "use the fallback option at genesis" is that, at the start of epoch 2, all nodes can use the tortoise active set for epoch 1 to calculate their Hare eligibility, then send Hare messages. Hare might still fail since there might not be consensus on the active set at this point, but this should be enough to bootstrap self-healing. Make sense?

blocks/blockoracle.go
lrettig: Fair point. I generally don't remove them when they are explanatory. Added them back here.

lrettig:
One more change I missed above: as discussed yesterday, previously, we defined the Hare active set as the set of unique ATXs referenced by blocks contained in a _single "safe" layer_ some distance before the current layer:

https://github.com/spacemeshos/go-spacemesh/blob/55957a811bbcdaf6eb2e215120090434d61e720f/hare/eligibility/oracle.go#L254

We now redefine the Hare active set for an epoch to be the set of all unique ATXs referenced by _all blocks in the first `adist+bdist` layers of the epoch._

hare/eligibility/oracle.go
lrettig: My strategy is to wrap the errors with slightly more context and then return them. Sometimes it's definitely helpful for debugging to print the error, with all of the local context, before returning it, too.

hare/eligibility/oracle.go
```diff
@@ -249,58 +269,68 @@ func (o *Oracle) Proof(ctx context.Context, layer types.LayerID, round int32) ([
...
 
-	// update
-	o.activesCache.Add(safeEp, activeMap)
+	// extract the nodeIDs
+	activeMap := make(map[string]struct{}, len(atxs))
+	for _, atxid := range atxs {
+		atx, err := o.atxdb.GetAtxHeader(atxid)
+		if err != nil {
+			return nil, fmt.Errorf("error getting ATX %v for target layer %v: %w", atxid, targetLayer, err)
```
noamnelke: Because we got this `atxid` from the database (vs. some incoming object) I think it's important to note that this is a consistency issue in the database. This should never happen unless we have a bug somewhere, and we may want to add some kind of monitoring for this specific keyword in errors.

```suggestion
			return nil, fmt.Errorf("inconsistent state: error getting ATX %v for target layer %v: %w", atxid, targetLayer, err)
```

lrettig: made this change

hare/eligibility/oracle.go
```diff
@@ -249,75 +276,107 @@ func (o *Oracle) Proof(ctx context.Context, layer types.LayerID, round int32) ([
...
+	// extract the nodeIDs
+	activeMap := make(map[string]struct{}, len(atxs))
+	for _, atxid := range atxs {
+		atx, err := o.atxdb.GetAtxHeader(atxid)
+		if err != nil {
+			return nil, fmt.Errorf("inconsistent state: error getting atx header %v for target layer %v: %w", atxid, targetLayer, err)
+		}
+		activeMap[atx.NodeID.Key] = struct{}{}
+	}
+	logger.With().Debug("got tortoise active set", log.Int("count", len(activeMap)), targetLayer.GetEpoch())
```
lrettig: Remove epoch from log

hare/eligibility/oracle.go
```diff
@@ -249,75 +276,107 @@ func (o *Oracle) Proof(ctx context.Context, layer types.LayerID, round int32) ([
...
+	if len(hareActiveSet) > 0 {
+		logger.With().Debug("successfully got hare active set for layer range",
+			log.Int("count", len(hareActiveSet)))
+		o.activesCache.Add(targetLayer.GetEpoch(), hareActiveSet)
+		return hareActiveSet, nil
+	}
+
+	// if we failed to get a Hare active set, we fall back on reading the Tortoise active set targeting this epoch
+	logger.With().Warning("no hare active set for layer range, reading tortoise set for epoch instead",
+		targetLayer.GetEpoch())
```
lrettig, line 491: Remove epoch from log

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

hare/eligibility/oracle.go
```diff
@@ -398,58 +418,92 @@ func (o *Oracle) actives(layer types.LayerID) (map[string]uint64, error) {
 	o.lock.Lock()
 	defer o.lock.Unlock()
 
-	// check cache
-	if val, exist := o.activesCache.Get(safeEp); exist {
-		return val.(map[string]uint64), nil
+	// check cache first
+	if val, exist := o.activesCache.Get(targetLayer.GetEpoch()); exist {
```
noamnelke: This is problematic. The target layer's epoch alone doesn't determine the safe layers, which determines the number of activations. If the target layer is before the epoch offset, the previous epoch is used (see lines 88-94 in this file).

I suggest we remove lines 88-94 and replace them with a method that determines in which epoch the safe layers should be - then we can use that function to determine the cache key, as well.

hare/eligibility/oracle.go
```diff
@@ -398,58 +418,92 @@ func (o *Oracle) actives(layer types.LayerID) (map[string]uint64, error) {
...
-		o.With().Error("could not calculate hare active set size: no contextually valid blocks",
-			layer,
-			layer.GetEpoch(),
-			log.FieldNamed("safe_layer_id", sl),
-			log.FieldNamed("safe_epoch_id", safeEp))
-		return nil, errNoContextualBlocks
+	if len(hareActiveSet) > 0 {
+		logger.With().Debug("successfully got hare active set for layer range",
+			log.Int("count", len(hareActiveSet)))
+		o.activesCache.Add(targetLayer.GetEpoch(), hareActiveSet)
```
noamnelke: We'd replace the key here, as well, of course.

noamnelke (changes requested):
I think I found what could end up becoming a pretty nasty bug... Let's fix it and also add a unit test that would catch it.

hare/eligibility/oracle.go
lrettig: > replace them with a method that determines in which epoch the safe layers should be

This is exactly what `safeLayerRange()` already does, isn't it? The answer is, `safeLayerStart.GetEpoch()` and `safeLayerEnd.GetEpoch()`.

> we can use that function to determine the cache key, as well

I don't think this is necessary. Think of the cache as memoizing this method call. If the method is called twice with the same input, it should return the same output. The cache key does _not_ need to agree with the safe layer.

noamnelke: Let's take an example. Say we have 10 layer epochs and 13 layer safety param, epoch offset is 3.

`targetLayer = 44`. We get `safeLayerStart = 20, safeLayerEnd = 23`. We cache the result for `targetLayer.GetEpoch() = 4`.

Then we get to `targetLayer = 47`. We get a cache hit for `targetLayer.GetEpoch() = 4` which was calculated with `safeLayerStart = 20, safeLayerEnd = 23` but the real result we want should be calculated with `safeLayerStart = 30, safeLayerEnd = 33`.

Added comments with actual values to the code I'm referring to:
```golang
	safeLayer := targetLayer - safetyParam               // safeLayer = 44 - 13 = 31   | 47 - 13 = 34
	safeEpoch := safeLayer.GetEpoch()                    // safeEpoch = 3              | 3
	safeLayerStart = safeEpoch.FirstLayer()              // safeLayerStart = 30        | 30
	safeLayerEnd = safeLayerStart + epochOffset          // safeLayerEnd = 30 + 3 = 33 | 33

	// If the safe layer is in the first epochOffset layers of an epoch,
	// return a range from the beginning of the previous epoch
	if safeLayer < safeLayerEnd {                        // 31 < 33 => true            | 34 < 33 => false
		// prevent overflow
		if safeLayerStart <= layersPerEpoch {
			return types.GetEffectiveGenesis(), types.GetEffectiveGenesis()
		}
		safeLayerStart = safeLayerStart - layersPerEpoch // safeLayerStart = 30 - 10 = 20
		safeLayerEnd = safeLayerEnd - layersPerEpoch     // safeLayerEnd = 33 - 10 = 23
	}
```

lrettig: thanks for the clear example. fixed this. please have another look!

hare/eligibility/oracle.go
```diff
@@ -55,61 +61,72 @@ type Oracle struct {
...
-	if sl >= roundedLayer { // the safe layer is after the rounding threshold
-		return roundedLayer // round to threshold
+// Returns a range of safe layers that should be used to construct the Hare active set for a given target layer
+// Safe layer is a layer prior to the input layer on which w.h.p. we have agreement (i.e., on its contextually valid
+// blocks), defined to be confidence param number of layers prior to the input layer.
+func safeLayerRange(
+	targetLayer types.LayerID,
+	safetyParam types.LayerID,
+	layersPerEpoch types.LayerID,
+	epochOffset types.LayerID,
```
noamnelke: I consider `LayerID` to be analogous to Go's datetime module's `Time` and `uint16` to be analogous to `Duration`.

For that reason I think it makes little sense to make these arguments of type `LayerID`.

noamnelke (changes requested):
Fix for the previous issue is 💯 .

There's a very minor issue remaining. LMK if you disagree.

hare/eligibility/oracle.go
lrettig: I changed them to uint16. The types we use to represent layers and layer diffs in configs and throughout the code are still messy and should still be cleaned up. Hopefully we can do that in #2489. But, you're right, we shouldn't make it worse here.

lrettig:
bors try

noamnelke (approved):

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
The system test errors make no sense to me. Log output hasn't changed. The line in question is printed in `p2p/switch.go`, which hasn't been touched, and no code that touches P2P or the public key has been changed. I suspect a logging issue.

```
        match = search_phrase_in_pod_log(bs_pod['name'], namespace, 'bootstrap',
                                         r"local node identity.*\"key\"\s*:\s*\"(?P<bootstrap_key>\w+)")
    
        if not match:
>           raise Exception("Failed to read container logs in {0}".format('bootstrap'))
E           Exception: Failed to read container logs in bootstrap
```

lrettig:
The error message appearing in the system tests is a red herring. For posterity, here's the full error message (found in tests kibana, then recreated locally):

```
> go run . --config /etc/config/config.toml --oracle_server http://10.40.5.12:3030 --grpc gateway,globalstate,transaction --test-mode --metrics --start-mining --metrics-port 2020 --pprof-server --poet-server 127.0.0.1:80 --genesis-time 2021-06-30T23:41:39+00:00 --randcon 8 --hare-committee-size 800 --hare-max-adversaries 399 --hare-round-duration-sec 7 --hare-exp-leaders 10 --layer-duration-sec 60 --layer-average-size 50 --hare-wakeup-delta 20 --layers-per-epoch 3 --coinbase 0x4321 --eligibility-confidence-param 5 --eligibility-epoch-offset 0 --post-space 1024 --space-to-commit 1024 --genesis-total-weight 51200 --acquire-port=false
2021-06-30T20:32:14.350-0400    WARN    00000.defaultLogger     failed loading config from /etc/config/config.toml trying ./config.toml
2021-06-30T20:32:14.352-0400    INFO    00000.defaultLogger     app cleanup starting...
2021-06-30T20:32:14.352-0400    INFO    00000.defaultLogger     app cleanup completed


panic: reflect.Set: value of type uint is not assignable to type uint16

goroutine 1 [running]:
reflect.Value.assignTo(0xec2ea0, 0x17e9548, 0x87, 0x10797f7, 0xb, 0xec2ee0, 0x0, 0x5, 0x0, 0x0)
        /usr/lib/go-1.15/src/reflect/value.go:2435 +0x405
reflect.Value.Set(0xec2ee0, 0xc00035e2b0, 0x189, 0xec2ea0, 0x17e9548, 0x87)
        /usr/lib/go-1.15/src/reflect/value.go:1564 +0xbd
github.com/spacemeshos/go-spacemesh/cmd.EnsureCLIFlags.func1(0x11f1b20, 0xf6a720, 0xf6a720, 0xc00035e2b0, 0x199, 0x108c044, 0x1c)
        go-spacemesh/cmd/base.go:139 +0x345
github.com/spacemeshos/go-spacemesh/cmd.EnsureCLIFlags.func2(0xc0003efa40)
        go-spacemesh/cmd/base.go:180 +0xb53
github.com/spf13/pflag.(*FlagSet).VisitAll(0xc000268d00, 0xc00022db70)
        go/pkg/mod/github.com/spf13/pflag@v1.0.5/flag.go:290 +0x94
github.com/spacemeshos/go-spacemesh/cmd.EnsureCLIFlags(0x18b3700, 0xc00035e000, 0x0, 0x1)
        go-spacemesh/cmd/base.go:146 +0xa9
github.com/spacemeshos/go-spacemesh/cmd/node.(*SpacemeshApp).Initialize(0xc0003a17c0, 0x18b3700, 0xc000032600, 0x0, 0x2f, 0xf13660, 0x0)
        go-spacemesh/cmd/node/node.go:264 +0x18d
github.com/spacemeshos/go-spacemesh/cmd/node.glob..func1(0x18b3700, 0xc000032600, 0x0, 0x2f)
        go-spacemesh/cmd/node/node.go:101 +0xd9
github.com/spf13/cobra.(*Command).execute(0x18b3700, 0xc000032310, 0x2f, 0x2f, 0x18b3700, 0xc000032310)
        go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0x18b3700, 0xc000000180, 0xc00022df78, 0x408025)
        go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
        go-spacemesh/main.go:23 +0xaf
exit status 2
```

lrettig:
bors merge

PR #2362, opened 2021-04-04
Don't include vrf pubkey in nodeID for rewards API

### LINKED ISSUE #2228: Some rewards are not sent by the api
Looks like not all rewards are sent via the api for a layer. See logs here: https://github.com/spacemeshos/explorer-frontend/issues/102 . This what causes the under-reporting of rewards in dash and explore.

### LINKED ISSUE #2366: unable to convert bytes to nodeid
I'm seeing this error from all nodes in the testnet (v0.1.28, tn127):

```
{
  "_index": "sm-2021.04.06",
  "_type": "_doc",
  "_id": "Y28MqXgBFRy6v6uwCc25",
  "_score": 1,
  "_source": {
    "@version": "1",
    "sm": {
      "errmsg": "Invalid input length, input too long",
      "N": "00000.defaultLogger",
      "L": "ERROR",
      "M": "unable to convert bytes to nodeid"
    },
    "name": "miner-39",
    "@timestamp": "2021-04-06T21:17:19.991Z",
    "stream": "stdout"
  },
  "fields": {
    "@timestamp": [
      "2021-04-06T21:17:19.991Z"
    ]
  }
}
```

Likely due to a bug in #2262 (or to an issue it uncovered).

sidsankhe:
## Motivation

Closes #2228
Closes #2366 

## Changes

There was an issue with the deserialization of the NodeID which resulted in no rewards being returned for specific miners. 
## TODO

## DevOps Notes


common/types/block_test.go
```diff
@@ -44,25 +44,19 @@ func TestStringToNodeID(t *testing.T) {
...
 	r := require.New(t)
 	r.NoError(err, "Error converting string to NodeID")
 	r.Equal(nodeID1.Key, reversed.Key, "Node ID deserialization Key does not match")
-	r.Equal(nodeID1.VRFPublicKey, reversed.VRFPublicKey, "Node ID deserialization VRF Key does not match")
 
 	// Test too short
 	_, err = StringToNodeID(string(pubkey[:10]))
 	r.Error(err, "Expected error converting too-short string to NodeID")
 
-	// Test too long
```
lrettig, line 61: You should leave a "too long" test for two reasons:

- if we're really ignoring everything after the first 32 bytes, then you should test passing in > 32 bytes and making sure you only get the first 32 back
- we want a regression test, i.e., one that would've failed before this fix but passes now

mesh/mesh.go
```diff
@@ -350,6 +350,7 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
 }
 
 func (msh *Mesh) applyState(l *types.Layer) {
+	msh.With().Info("applying state of the layer")
```
lrettig: Don't use `With()` without passing additional LoggableFields - or, better yet, add the layer here

mesh/mesh.go
```diff
@@ -765,6 +766,7 @@ func (msh *Mesh) GetOrphanBlocksBefore(l types.LayerID) ([]types.BlockID, error)
 }
 
 func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
+	msh.With().Info("Accumulating rewards")
```
lrettig: same comment as above

mesh/meshdb.go
```diff
@@ -519,6 +519,7 @@ func (m *DB) writeTransactionRewards(l types.LayerID, accountBlockCount map[type
 	for account, smesherAccountEntry := range accountBlockCount {
 		for smesherString, cnt := range smesherAccountEntry {
 			smesherEntry, err := types.StringToNodeID(smesherString)
+			m.Info("writing reward to database")
```
lrettig: add some more context to this log message to make it more useful: layer, reward amount, smesher, etc.

lrettig (changes requested):

common/types/block_test.go
sidsankhe: Makes sense, I added in the too-long tests for both the string and bytes deserialization

lrettig:
I improved the logs here for the testnet branch and cherry-picked the change onto this branch as well

common/types/block_test.go
```diff
@@ -57,6 +58,12 @@ func TestStringToNodeID(t *testing.T) {
 	_, err = StringToNodeID(string(pubkey[:10]))
 	r.Error(err, "Expected error converting too-short string to NodeID")
 
+	//Test too long : make sure that everything after the first 64 characters is ignored
+	longString := nodeID1.String() + string(pubkey2[:])
+	reversed, err = StringToNodeID(string(longString))
```
lrettig: isn't `longString` already a string? why do you need to use `string()` here?

common/types/block_test.go
```diff
@@ -57,6 +58,12 @@ func TestStringToNodeID(t *testing.T) {
 	_, err = StringToNodeID(string(pubkey[:10]))
 	r.Error(err, "Expected error converting too-short string to NodeID")
 
+	//Test too long : make sure that everything after the first 64 characters is ignored
+	longString := nodeID1.String() + string(pubkey2[:])
```
lrettig, line 62: Insert after this line:
`r.Greater(len(longString), len(nodeID1.String()))`

(it's fairly obvious but it makes the test more readable and helps protect against future changes to the test - it's always good to be extra verbose in tests)

common/types/block_test.go
```diff
@@ -79,4 +86,12 @@ func TestBytesToNodeID(t *testing.T) {
 	rand.Read(x[:])
 	_, err = BytesToNodeID(x[:])
 	r.Error(err, "Expected error converting too-short byte array to NodeID")
+
+	// Test too long
+	longSlice := append(bytes, x[:]...)
```
lrettig, line 92: as above, add a `t.Greater` test

lrettig (changes requested):
Just some minor tweaks in the test and then I think this is good to go

lrettig:
Please merge in develop and resolve merge conflicts. Probably also best to rebase this on #2411 before finalizing it.

lrettig (changes requested):
Let's rebase on #2411 before finalizing

common/types/block_test.go
```diff
@@ -42,27 +42,29 @@ func TestFields(t *testing.T) {
...
-	rand.Read(x[:])
-	_, err = StringToNodeID(string(x[:]))
-	r.Error(err, "Expected error converting too-long string to NodeID")
+	//Test too long : make sure that everything after the first 64 characters is ignored
+	longString := nodeID1.String() + string(pubkey2[:])
+	r.Greater(len(longString), len(nodeID1.String()))
+	reversed, err = StringToNodeID(longString)
+	r.NoError(err, "Error converting too-long string to NodeID")
+	r.Equal(nodeID1.Key, reversed.Key, "NodeID does not match for too-long string")
+	r.Equal([]byte{}, reversed.VRFPublicKey, "VRF Key is not empty for too-long string")
```
lrettig, line 67: "is" empty, not "is not" empty, right?

lrettig: also, is there a difference between `[]byte{}` and `[]byte(nil)` (which you use below)?

common/types/block_test.go
```diff
@@ -88,8 +89,11 @@ func TestBytesToNodeID(t *testing.T) {
...
-	var y [65]byte
-	rand.Read(y[:])
-	_, err = BytesToNodeID(y[:])
-	r.Error(err, "Expected error converting too-long byte array to NodeID")
+	longSlice := append(bytes, x[:]...)
+	r.Greater(len(longSlice), len(bytes))
+	reversed, err = BytesToNodeID(longSlice)
+	r.NoError(err, "Error converting too-long byte array to NodeID")
+	r.Equal(nodeID1.Key, reversed.Key, "NodeID Key does not match for too-long")
+	r.Equal([]byte{}, reversed.VRFPublicKey, "VRF Key is not empty for too-long")
```
lrettig, line 97: is there a difference between `[]byte{}` and `[]byte(nil)` (which you use below)?

common/types/block.go
```diff
@@ -134,15 +133,14 @@ func StringToNodeID(s string) (*NodeID, error) {
...
 	// portion of the string corresponding to the Edwards public key
+	// Note that we don't care when the input is too long, because we are only deserializing the first 32
+	// bytes into the Edwards public key. The VRFKey is not used by the deserializaiton at all, and thus
+	// it is returned as empty
 	pubKey := s[:64]
-	vrfKey := s[64:]
 	return &NodeID{
 		Key:          pubKey,
-		VRFPublicKey: []byte(vrfKey),
+		VRFPublicKey: []byte{},
```
lrettig, line 143: `[]byte{}` or `[]byte(nil)`?

common/types/block.go
```diff
@@ -114,15 +115,13 @@ func BytesToNodeID(b []byte) (*NodeID, error) {
...
-
+	// Note that we don't care when the input is too long, because we are only deserializing the first 32
+	// bytes into the Edwards public key. The VRFKey is not used by the deserializaiton at all, and thus
+	// it is returned as empty
 	pubKey := b[0:32]
-	vrfKey := b[32:]
 	return &NodeID{
 		Key:          util.Bytes2Hex(pubKey),
-		VRFPublicKey: []byte(util.Bytes2Hex(vrfKey)),
+		VRFPublicKey: []byte{},
```
lrettig, line 124: `[]byte{}` or `[]byte(nil)`?

lrettig (approved):
Approving, but see comments about minor changes

nkryuchkov:
@lrettig Should we fix your suggestions, merge develop into this branch and merge the PR?

dshulyak:
We will use ecvrf with a single key at some point, no need to do anything about this change

PR #2368, opened 2021-04-07
Make hare active set/eligibility check less fragile (take 2)

### LINKED ISSUE #2356: Hare active set size calculation is fragile

lrettig:
## Motivation
Closes #2356

## Changes
When calculating active set for a given epoch for hare eligibility, if there are no contextually valid blocks for the safe layer (e.g., if hare failed for this layer), just use zero as the active set size.

This issue caused the demise of testnet 126. Per Tal, this is the safest workaround for now, until eventually the tortoise beacon allows us to remove the dependency of hare beacon upon hare itself.

## Test Plan
Adds a regression test

## TODO

## DevOps Notes

### DIFF
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -19,7 +19,6 @@ const activesCacheSize = 5 // we don't expect to handle more than two layers con
 
 var (
 	errGenesis            = errors.New("no data about active nodes for genesis")
-	errNoContextualBlocks = errors.New("no contextually valid blocks")
 )
 
 type valueProvider interface {
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -269,9 +269,9 @@ func (o *Oracle) actives(layer types.LayerID) (activeMap map[string]struct{}, er
 	}
 
 	// build a map of all blocks on the current layer
-	mp, err := o.blocksProvider.ContextuallyValidBlock(sl)
-	if err != nil {
-		return
+	mp, err2 := o.blocksProvider.ContextuallyValidBlock(sl)
+	if err2 != nil {
+		return nil, err2
 	}
 
 	// no contextually valid blocks: for now we just fall back on an empty active set. this will go away when we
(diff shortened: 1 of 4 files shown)


hare/eligibility/oracle.go
```diff
@@ -270,20 +270,21 @@ func (o *Oracle) actives(layer types.LayerID) (map[string]struct{}, error) {
...
+	// upgrade hare eligibility to use the tortoise beacon.
 	if len(mp) == 0 {
-		o.With().Error("could not calculate hare active set size: no contextually valid blocks",
+		o.With().Warning("no contextually valid blocks in layer, using active set of size zero",
 			layer,
 			layer.GetEpoch(),
 			log.FieldNamed("safe_layer_id", sl),
 			log.FieldNamed("safe_epoch_id", safeEp))
-		return nil, errNoContextualBlocks
+		return activeMap, nil
```
sudachen: Not sure, but is it possible that hare will have no eligible nodes for consensus for the whole epoch?
https://github.com/spacemeshos/go-spacemesh/blob/develop/hare/eligibility/oracle.go#L208

lrettig:
As @sudachen pointed out, this solution won't work since we cannot calculate eligibility using an active set of size zero (even temporarily, as a hack). We need to have an active set size or else we'll end up with a committee (or no. of leaders) that's too big or too small.

https://github.com/spacemeshos/go-spacemesh/blob/51a69a81d6eafa78b7b4efdf47ff0c653022fca0/hare/eligibility/oracle.go#L206-L210

One option would be to count contextually invalid blocks (if there are no contextually valid ones). But I think we should just go with my original patch, in #2357. I'll close this in favor of that PR.

Issue #2333, opened 2021-03-18
Return true current sync layer in API


lrettig:
In #2299 we patched the API to return the current (tip) layer rather than statically returning the first layer of the third epoch as the current layer. But this still isn't really accurate. What we really want is for the API to instead return the latest _synced_ layer. This information is currently not exposed to the API, but it could be, and it definitely lives in the sync code somewhere, since sync prints log lines like these:

```
2021-03-14T22:04:13.399+0300    INFO    bfda3.sync              syncing layer   {"node_id": "...", "current_sync_layer": 67, "last_ticked_layer": 3375}
```

The `NodeService.Status` should return the latest _synced_ layer (not latest layer in state) during genesis (and maybe always.)

dshulyak:
Latest synced layer is stored as a ProcessedLayer in the validator, so we can just read it (note that ProcessedLayer+1 might be *current* layer that is being synced, but definitely not *latest*).

@lrettig can you elaborate why would we want to use it during genesis? as I understand your comment is still valid
```
// We cannot get meaningful data from the mesh during the genesis epochs since there are no blocks in these
// epochs, so just return the current layer instead```

lrettig:
> @lrettig can you elaborate why would we want to use it during genesis? as I understand your comment is still valid

`NodeService.Status` returns three layer values:

https://github.com/spacemeshos/go-spacemesh/blob/631d56f11c25aa36074e84fbbf5879fd504cf32b/api/grpcserver/node_service.go#L81-L83

While syncing genesis layers, the intuitive thing that should happen here is that `SyncedLayer` and `VerifiedLayer` should both be set to the most recently synced layer. There's no state, no blocks, nothing to process or verify during genesis (i.e., all of these layers are "zero layers" from the perspective of syncer).

`TopLayer` should be set to the _current layer height_ (based on time), as it is now.

Right now, the code sets `VerifiedLayer` and `SyncedLayer` to equal `TopLayer`. From the perspective of the user, this is confusing, since it seems like the node is already at layer 575 (the final genesis layer) when it first launches, and then gets stuck at this layer for a while.

Does this make sense?

dshulyak:
[...]

PR #2379, opened 2021-04-14
true-current-layer

cong-or:
@lrettig 

see issue [#2333](https://github.com/spacemeshos/go-spacemesh/issues/2333)

After investigating the code base, it looks like `s.ProcessedLayer() + 1` is equivalent to the latest synced layer `currentSyncLayer` .

However, I am curious if the following are [truer](https://github.com/spacemeshos/go-spacemesh/blob/develop/api/grpcserver/mesh_service.go#L378) current sync layers.

`// Get the latest layers that passed both consensus engines.`
	`lastLayerPassedHare := s.Mesh.LatestLayerInState()`
	`lastLayerPassedTortoise := s.Mesh.ProcessedLayer()`

I was unable to modify the tests to check if this hunch was right as I do not have access to the protos 
`pb "github.com/spacemeshos/api/release/go/spacemesh/v1"`

### DIFF
--- a/api/grpcserver/node_service.go
+++ b/api/grpcserver/node_service.go
@@ -87,13 +87,15 @@ func (s NodeService) Status(context.Context, *pb.StatusRequest) (*pb.StatusRespo
 func (s NodeService) getLayers() (curLayer, latestLayer, verifiedLayer uint32) {
 	// We cannot get meaningful data from the mesh during the genesis epochs since there are no blocks in these
 	// epochs, so just return the current layer instead
+	latestSyncedLayer := s.Mesh.ProcessedLayer() + 1
+
 	curLayerObj := s.GenTime.GetCurrentLayer()
 	curLayer = uint32(curLayerObj)
 	if curLayerObj.GetEpoch().IsGenesis() {
-		latestLayer = curLayer
+		latestLayer = uint32(latestSyncedLayer)
 		verifiedLayer = curLayer
 	} else {
-		latestLayer = uint32(s.Mesh.LatestLayer())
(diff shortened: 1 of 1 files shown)


lrettig:
Thanks for the suggestion!

> I was unable to modify the tests to check if this hunch was right as I do not have access to the protos
> `pb "github.com/spacemeshos/api/release/go/spacemesh/v1"`

These are just a compiled version of the code in https://github.com/spacemeshos/api/tree/master/proto/spacemesh/v1

cong-or:
It may be best to just add a field to [node status](https://github.com/spacemeshos/api/blob/master/proto/spacemesh/v1/node_types.proto#L36) 

`message NodeStatus {`
    `uint64 connected_peers = 1; // number of connected neighbors`
    `bool   is_synced = 2; // true when meshed is synced`
    `LayerNumber synced_layer = 3; // the last layer node has synced`
    `LayerNumber top_layer = 4; // top layer is the tip`
    `LayerNumber verified_layer = 5; // the last layer node has verified`
     `LayerNumber latest_synced_layer = 6; //  latest synced later`
}

The previous commit caused too many side effects based on the tests reaction, it might be better to gradually introduce the `latest_synced _layer` and use it when needed rather than replacing one of the other fields.

nkryuchkov:
@lrettig @dshulyak Was this PR superseded with https://github.com/spacemeshos/go-spacemesh/pull/2482?

dshulyak:
@nkryuchkov thanks, i think this is the case

nkryuchkov:
@mycorrhizas We've decided to implement #2333 differently but we appreciate your contribution. Thank you

PR #2387, opened 2021-04-19
[Merged by Bors] - Removed logstash from ELK and added support for setting namespace name and  ns deletion from cli

AmitShaul:
## Motivation
logstash have created a bottle neck while shipping logs to ES.
removed logstash from ELK and upgraded fluent bit to latest version (1.7.4).

make it easier for developers to control the namespace name or whether to delete it at the end of the run.
prevent logs loss in case ES dump has failed.

## Changes
removed logstash from ELK cluster

added 2 cmd line arguments
namespace - the namespace name
delns - whether or not to delete the namespace at the end of the run

in case the ES dump has failed do not delete the namespace for log exploration


AmitShaul:
bors try

AmitShaul:
bors try

Makefile
```diff
@@ -260,37 +260,53 @@ DELIM=::
...
+get-delns-cmd-arg:
+ifdef delns
+    EXTRA_PARAMS:=$(EXTRA_PARAMS) --delns=$(delns)
+endif
+.PHONY: get-delns-cmd-arg
+
+
+get_extra_params: get-namespace-cmd-arg get-delns-cmd-arg
+
+
```
noamnelke: This is unneeded and fragile.

I don't think `EXTRA_PARAMS` should be constructed in a special target, but in the root of the makefile.

tests/conftest.py
```diff
@@ -99,48 +126,50 @@ def set_docker_images():
...
+                       f"kubectl delete namespace {session_id}"
+            print(exit_msg)
 
-    request.addfinalizer(fin)
-    return _setup_namespace()
+    setup_namespace()
+    yield
+    # delete namespace if was not mentioned otherwise in the command line args and if global arg is set to
+    # True (pytest.delete_namespace)
+    fin([delete_ns, pytest.delete_namespace])
```
noamnelke: This code is organized in a VERY confusing way. Defining functions and then calling them immediately is bad form.

We should put the code directly in `set_namespace()` - first the code that's currently in `setup_namespace()` (replacing the `return` there with an `else`), then the `yield` keyword, along with a comment like 
```python
# pytest considers any code after `yield` to be teardown code
# see: https://docs.pytest.org/en/reorganize-docs/yieldfixture.html
```

and then the code that's currently in `fin()`.

noamnelke (changes requested):

tests/conftest.py
noamnelke: Alternatively, define the two methods outside of `set_namespace()` and make it just call `setup_namespace()`, then `yield` + the comment, then `fin()`.

That's actually my preferred solution.

AmitShaul:
bors try

Makefile
```diff
@@ -260,37 +260,53 @@ DELIM=::
 endif
 
 
+get-namespace-cmd-arg:
+ifdef namespace
+    EXTRA_PARAMS:=$(EXTRA_PARAMS) --namespace=$(namespace)
```
antonlerner: you can create another parameter called "NAMESPACE" and just assign `NAMSPACE=$(namespace)`, then you can still add this as parameter to commands - if it doesn't exist nothing will be affected 
in case this parameter doesn't exist it will do nothing

Makefile
```diff
@@ -260,37 +260,53 @@ DELIM=::
...
+endif
+.PHONY: get-namespace-cmd-arg
+
+get-delns-cmd-arg:
+ifdef delns
+    EXTRA_PARAMS:=$(EXTRA_PARAMS) --delns=$(delns)
```
antonlerner: same here

tests/elk/logstash/statefulset.yaml
```diff
@@ -30,15 +30,15 @@ spec:
...
-        podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            - labelSelector:
-                matchExpressions:
-                  - key: app
-                    operator: In
-                    values:
-                      - "logstash"
-              topologyKey: kubernetes.io/hostname
+#        podAntiAffinity:
```
y0sher, line 33: if we removed logstash shouldn't we remove this whole folder?

AmitShaul:
bors try

Makefile
AmitShaul: so it's the same as it already is, if it doesn't exists nothing will be affected

tests/elk/logstash/statefulset.yaml
AmitShaul: leaving it for future utilisation, in case we will be needing it.
it's in a separated folder so I don't think it's an issue leaving it.

tests/conftest.py
AmitShaul: About the code - I just added the yield command, my plan was to rewrite this function so I didn't bother unpacking the nested functions.

I've used yield keyword in other fixtures as well, I don't see a reason to add a comment on python syntax,
lmk if you insist we can talk about it.

AmitShaul: fixed

tests/elk/logstash/statefulset.yaml
AmitShaul: if you'd like I can delete this

AmitShaul:
bors try

AmitShaul:
bors try

noamnelke (approved):

AmitShaul:
bors try

AmitShaul:
bors merge

AmitShaul:
bors try

AmitShaul:
bors try-

AmitShaul:
bors ping

AmitShaul:
bors merge

PR #2388, opened 2021-04-20
Streaming api patch

avive:
Temp fix to reporter to always report data to streaming clients in a blocking way. This addresses the issue of explorer backend missing rewards but doesn't solve the issue of multiple clients on streams that use same reporter channels.


avive:
going to do this pr from develop and not from tag/0.1.29

PR #2389, opened 2021-04-20
patch reporter to address streaming api issue

### LINKED ISSUE #2384: Streaming apis implementation can miss sending data to clients
## Description
As the streaming api is currently implemented, there can be multiple consumers for a single go channel of events created by the event reporter.  For example, in the globalState service implantation, there will be a consumer of data from the reporter's singleton rewardsChannel, per subscribed client, per one of the 3 streaming methods that use this channel - GlobalStateStream, SmesherRewardStream and AccountDataStream. 

This will cause an issue where for example, not all rewards events will be sent to all registered clients in the case that two or more clients are registered to receive rewards using any of the 3 implemented streaming methods, or even  when only client is registered on two or more of the streams because the first select loop which will receive the event from the channel will consume it.

We need to implement pub-sub here to support multiple producers and multiple consumers pattern.

avive:
Fixes #2384


events/reporter.go
```diff
@@ -77,17 +68,34 @@ func ReportNewActivation(activation *types.ActivationTx) {
...
-				log.With().Debug("reported activation", activation.Fields(len(innerBytes))...)
-			default:
-				log.With().Debug("not reporting activation as no one is listening", activation.Fields(len(innerBytes))...)
+		reporter.channelActivation <- activation
+		log.With().Debug("reported activation", activation.Fields(len(innerBytes))...)
+	}
+}
+
+// SubscribeToRewards subscribes a channel to rewards events
+func SubscribeToRewards (subscriber chan Reward) {
```
lrettig: This method should take no args, create the channel, and return it

events/reporter.go
```diff
@@ -77,17 +68,34 @@ func ReportNewActivation(activation *types.ActivationTx) {
...
+	mu.RLock()
+	defer mu.RUnlock()
+	if reporter != nil {
+		reporter.rewardsSubs = append(reporter.rewardsSubs, subscriber)
```
lrettig: The channel should be stored in a map, not an array

events/reporter.go
```diff
@@ -77,17 +68,34 @@ func ReportNewActivation(activation *types.ActivationTx) {
...
+
+// UnsubscribeFromRewards unsubscribes a channel from rewards events
+func UnsubscribeFromRewards(subscriber chan Reward) {
+	mu.RLock()
+	defer mu.RUnlock()
+
+	if reporter != nil {
+		for i, s := range reporter.rewardsSubs {
+			if s == subscriber {
+				reporter.rewardsSubs[i] = reporter.rewardsSubs[len(reporter.rewardsSubs)-1]
```
lrettig: Much easier to delete from a map rather than splicing an array here

events/reporter.go
```diff
@@ -102,18 +110,13 @@ func ReportRewardReceived(r Reward) {
...
-			select {
-			case reporter.channelReward <- r:
-				log.Debug("reported reward: %v", r)
-			default:
-				log.Debug("not reporting reward as no one is listening: %v", r)
-			}
+	if reporter != nil && len(reporter.rewardsSubs) > 0 {
+		log.Info("about to report reward: %v", r)
+		for _, sub := range reporter.rewardsSubs {
+			sub <- r
```
lrettig: Need to use a `select{}` statement here to make the send non-blocking - it must never block. And print a Debug/Info if no subscribers, or if any of the channels is full.

events/reporter.go
lrettig: Or: could accept a single arg, which is the buffer size of the channel

lrettig (changes requested):
Generally agree with this approach. We should do this for all of the reporter channels. See comments.

avive:
> Generally agree with this approach. We should do this for all of the reporter channels. See comments.

this is just a sketch WIP PR - I didn't mean this code should be merged - it is a proof of concept.

events/reporter.go
avive: I think that this means that if the channel is full because the client is slow pulling rewards from it (super slow network connection), then the client will miss reported rewards. Ideally, slow clients won't miss any events.

avive: Also - won't this code block in line 121 in case that the channel is full?

avive: Didn't you mean to remove this line as the send happens in line 117?

sidsankhe: Yes, I meant to remove the line apologies. If we don't want slow clients to miss rewards, should we then try sends in goroutines instead? So that the reporter itself won't block, but the individual goroutine will, until the client is ready

avive: I think we should but it needs to be done with care because the subscribers list is reporter state and can change. So, I think that each specific send needs to be done in a go routine and not 1 go routine with iteration over subscribers list inside of it. Also curious what @noamnelke thinks on the best design for this.

lrettig: The solution here is _not_ one goroutine per send. This is exactly the use case for buffered channels. Just set a capacity for the channel, write to the channel, and let the consumer consume it. If the channel is full, the write should not block, it should drop the data (and print a log).

avive: Also - we want to apply this not just to rewards but to all streaming apis that smapp 0.2 needs.

sidsankhe: Which other channels does this need to be applied to? I see `channelAccount` of type `types.Address,` `channelReceipt` of type `events.TxReceipt`. Are there any others that I'm missing?

events/reporter.go
```diff
@@ -102,18 +127,16 @@ func ReportRewardReceived(r Reward) {
 		SmesherID: r.Smesher.ToBytes(),
 	})
 
-	if reporter != nil {
-		if reporter.blocking {
-			reporter.channelReward <- r
-			log.Debug("reported reward: %v", r)
-		} else {
+	if reporter != nil && len(reporter.rewardsSubs) > 0 {
```
lrettig: You can remove the second part of the check here, `len(reporter.rewardsSubs) > 0` - this is implicit in the loop

events/reporter.go
```diff
@@ -91,6 +91,31 @@ func ReportNewActivation(activation *types.ActivationTx) {
...
+		newChan := make(chan Reward, bufsize)
+		reporter.rewardsSubs[newChan] = struct{}{}
+		return newChan
+	}
+	return nil
+}
+
+// UnsubscribeFromRewards unsubscribes a channel from rewards events
+// do we need to close the subscriber channel here?
+func UnsubscribeFromRewards(subscriber chan Reward) {
```
lrettig: the `subscriber` arg should be renamed `subscription` or `subscriptionChannel` or something like that

lrettig (changes requested):

events/reporter.go
avive: yes, please go over api services definitions in the api repo and collect all streaming methods. For example, NodeService has 2 important ones. All should support multiple slow clients.

api/grpcserver/node_service.go
```diff
@@ -123,7 +125,11 @@ func (s NodeService) Shutdown(context.Context, *pb.ShutdownRequest) (*pb.Shutdow
 // StatusStream exposes a stream of node status updates
 func (s NodeService) StatusStream(_ *pb.StatusStreamRequest, stream pb.NodeService_StatusStreamServer) error {
 	log.Info("GRPC NodeService.StatusStream")
-	statusStream := events.GetStatusChannel()
+	var statusStream chan struct{}
+	if s.statusChannel == nil {
+		statusStream = events.SubscribeToStatus(30)
```
lrettig: Please turn this hardcoded number 30 into a param. You might want the default value to live in the reporter rather than in the API, or to be used across the API.

events/reporter.go
```diff
@@ -203,17 +203,15 @@ func ReportError(err NodeError) {
 	defer mu.RUnlock()
 
 	if reporter != nil {
-		if reporter.blocking {
-			reporter.channelError <- err
-			log.Debug("reported error: %v", err)
-		} else {
+		log.Info("about to report error updates for error : %v", err)
```
lrettig: please use the new log format:

```suggestion
		log.With().Info("about to report error updates for error", log.Err(err))
```

events/reporter.go
```diff
@@ -203,17 +203,15 @@ func ReportError(err NodeError) {
...
+		for sub := range reporter.errorSubs {
 			select {
-			case reporter.channelError <- err:
-				log.Debug("reported error: %v", err)
+			case sub <- err:
 			default:
-				log.Debug("not reporting error as buffer is full: %v", err)
+				log.Debug("reporter would block on subscriber %v", sub)
```
lrettig: use new log format here, no "%v"

events/reporter.go
```diff
@@ -234,17 +232,15 @@ func ReportNodeStatusUpdate() {
 	defer mu.RUnlock()
 
 	if reporter != nil {
-		if reporter.blocking {
-			reporter.channelStatus <- struct{}{}
-			log.Debug("reported status update")
-		} else {
+		log.Info("about to report status updates")
```
lrettig: make this log statement more useful, with more context, or leave it out

api/grpcserver/globalstate_service.go
```diff
@@ -274,13 +278,25 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 		channelReceipt chan events.TxReceipt
 	)
 	if filterAccount {
-		channelAccount = events.GetAccountChannel()
+		if s.accountChannel == nil {
+			channelAccount = events.SubscribeToAccounts(30)
```
lrettig: As mentioned elsewhere, please turn this hardcoded number into a param

api/grpcserver/globalstate_service.go
```diff
@@ -448,13 +475,29 @@ func (s GlobalStateService) GlobalStateStream(in *pb.GlobalStateStreamRequest, s
 		channelLayer   chan events.NewLayer
 	)
 	if filterAccount {
-		channelAccount = events.GetAccountChannel()
+		if s.accountChannel == nil {
+			channelAccount = events.SubscribeToAccounts(30)
+		} else {
+			channelAccount = s.accountChannel
```
lrettig: This channel may be used multiple places, by multiple streams. I know this is only for debugging, but it could cause unintended errors - same as we're seeing now, where multiple subscribers to the same channel will not see everything published to the channel. The simplest fix might be to create a different struct member for each stream, e.g., for this stream rename `s.accountChannel` to `s.globalStateStreamAccountChannel`.

api/grpcserver/globalstate_service.go
```diff
@@ -274,13 +278,25 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
+		}
 	}
 	if filterReward {
-		channelReward = events.GetRewardChannel()
+		if s.rewardsChannel == nil {
+			channelReward = events.SubscribeToRewards(30)
```
lrettig: It looks like you're not handling the error case here, where this returns nil, as you are elsewhere

api/grpcserver/globalstate_service.go
```diff
@@ -448,13 +475,29 @@ func (s GlobalStateService) GlobalStateStream(in *pb.GlobalStateStreamRequest, s
...
 	if filterReward {
-		channelReward = events.GetRewardChannel()
+		// needs to be buffered to avoid reporter from being blocked on slow streaming connections
+		if s.rewardsChannel == nil {
+			channelReward = events.SubscribeToRewards(30)
+		} else {
+			channelReward = s.rewardsChannel
+		}
+		if channelReward == nil {
+			// what to do here
```
lrettig, line 521: This means the reporter isn't running. Log an error to this effect and return.

api/grpcserver/grpcserver_test.go
```diff
@@ -2089,6 +2096,15 @@ func TestAccountDataStream_comprehensive(t *testing.T) {
 	wg := sync.WaitGroup{}
```
lrettig, line 2166: Using a series of waitgroups like this is an antipattern. You should use a (unbuffered) channel instead. The listener reads from the channel, the sender sends into the channel to release it.

api/grpcserver/grpcserver_test.go
```diff
@@ -2101,33 +2117,39 @@ func TestAccountDataStream_comprehensive(t *testing.T) {
...
+
 		res, err = stream.Recv()
 		require.NoError(t, err, "got error from stream")
 		checkAccountDataItemReward(t, res.Datum.Datum)
 
+		wg3.Done()
+
 		res, err = stream.Recv()
+		if err != nil {
+			log.Error("Error thrown : %v", err)
```
lrettig: this check isn't necessary since you have `require.NoError` below

go.sum
```diff
@@ -22,12 +22,15 @@ cloud.google.com/go/bigquery v1.3.0/go.mod h1:PjpwJnslEMmckchkHFfq+HTD2DmtT67aNF
 cloud.google.com/go/bigquery v1.4.0/go.mod h1:S8dzgnTigyfTmLBfrtrhyYhwRxG72rYxvftPBK2Dvzc=
 cloud.google.com/go/bigquery v1.5.0/go.mod h1:snEHRnqQbz117VIFhE8bmtwIDY80NLUZUMb4Nv6dBIg=
 cloud.google.com/go/bigquery v1.7.0/go.mod h1://okPTzCYNXSlb24MZs83e2Do+h+VXtc4gLoIoXIAPc=
+cloud.google.com/go/bigquery v1.8.0 h1:PQcPefKFdaIzjQFbiyOgAqyx8q5djaE7x9Sqe712DPA=
```
lrettig: Is there a specific reason you need to include this? I know it was added when Aviv first opened the PR. Unless you've explicitly added or updated go modules, it's probably not necessary - see if you can remove it.

lrettig (changes requested):
Few things to clean up: use of waitgroups in tests, overlapping use of channels in API streams, the format of the error messages, the new param, the `go.sum` file, etc. Thanks!

go.sum
avive: I didn't mean to add any dep explicitly when I wrote the poc - I just run `make'.

lrettig: @sidsankhe you can remove go.sum from the git index for this PR, it's probably not necessary

events/reporter.go
```diff
@@ -42,16 +42,19 @@ func ReportTxWithValidity(tx *types.Transaction, valid bool) {
...
-				log.Info("reported tx to subs")
+				log.With().Debug("reported tx to subs", txWithValidity.Transaction.ID(),
+					log.Bool("validity", txWithValidity.Valid))
 			default:
-				log.Debug("reporter would block on subscriber %v", sub)
+				log.Debug("reporter would block on subscriber")
 			}
 		}
-		log.Info("reported tx validity update to subscribers: %v", txWithValidity)
+		log.With().Info("reported tx validity update to subscribers: %v", txWithValidity.Transaction.ID(),
```
lrettig:
```suggestion
		log.With().Debug("reported tx validity update to subscribers", txWithValidity.Transaction.ID(),
```

events/reporter.go
```diff
@@ -125,16 +128,16 @@ func ReportRewardReceived(r Reward) {
 	})
 
 	if reporter != nil {
-		log.Info("about to report reward: %v", r)
+		log.With().Info("about to report reward", r)
 		for sub := range reporter.rewardsSubs {
 			select {
 			case sub <- r:
-				log.Info("reporter send reward %v to subscriber", r)
+				log.With().Info("reporter send reward to subscriber", r)
```
lrettig:
```suggestion
				log.With().Debug("reporter send reward to subscriber", r)
```

events/reporter.go
```diff
@@ -125,16 +128,16 @@ func ReportRewardReceived(r Reward) {
...
 			default:
-				log.Debug("reporter would block on subscriber %v", sub)
+				log.Debug("reporter would block on subscriber")
 			}
 		}
-		log.Info("reported reward to subscriber: %v", r)
+		log.With().Info("reported reward to subscriber", r)
```
lrettig:
```suggestion
		log.With().Debug("reported reward to subscribers", r)
```

events/reporter.go
```diff
@@ -180,15 +183,16 @@ func ReportNewLayer(layer NewLayer) {
...
 			select {
 			case sub <- layer:
+				log.With().Debug("reporter new layer to subscriber", layer)
 			default:
-				log.Debug("reporter would block on subscriber %v", sub)
+				log.Debug("reporter would block on subscriber")
 			}
 		}
-		log.Info("reported layer update to subscribers: %v", layer)
+		log.With().Info("reported layer update to subscribers", layer)
```
lrettig:
```suggestion
		log.With().Debug("reported layer update to subscribers", layer)
```

events/reporter.go
```diff
@@ -198,15 +202,16 @@ func ReportError(err NodeError) {
 	defer mu.RUnlock()
 
 	if reporter != nil {
-		log.Info("about to report error updates for error : %v", err)
+		log.With().Info("about to report error updates for error", log.String("error-msg", err.Msg))
```
lrettig, line 219:
```suggestion
		log.With().Info("about to report error updates for error", log.Err(err))
```

events/reporter.go
```diff
@@ -198,15 +202,16 @@ func ReportError(err NodeError) {
...
 			select {
 			case sub <- err:
+				log.With().Debug("reported error updates to subscriber for error", log.String("error-msg", err.Msg))
 			default:
-				log.Debug("reporter would block on subscriber %v", sub)
+				log.Debug("reporter would block on subscriber")
 			}
 		}
-		log.Info("reported error update to subscribers: %v", err)
+		log.With().Info("reported error update to subscribers: %v", log.String("error-msg", err.Msg))
```
lrettig:
```suggestion
		log.Debug("reported error update to subscribers")
```

events/reporter.go
```diff
@@ -198,15 +202,16 @@ func ReportError(err NodeError) {
 	defer mu.RUnlock()
 
 	if reporter != nil {
-		log.Info("about to report error updates for error : %v", err)
+		log.With().Info("about to report error updates for error", log.String("error-msg", err.Msg))
 		for sub := range reporter.errorSubs {
 			select {
 			case sub <- err:
+				log.With().Debug("reported error updates to subscriber for error", log.String("error-msg", err.Msg))
```
lrettig, line 223:
```suggestion
				log.With().Debug("reported error updates to subscriber for error", log.Err(err.Msg))
```

events/reporter.go
```diff
@@ -231,10 +236,9 @@ func ReportNodeStatusUpdate() {
 			select {
 			case sub <- struct{}{}:
 			default:
-				log.Debug("reporter would block on subscriber %v", sub)
+				log.Debug("reporter would block on subscriber")
 			}
 		}
-		log.Info("reported status update to subscribers: %v")
```
lrettig: why remove this one?

events/reporter.go
```diff
@@ -244,16 +248,16 @@ func ReportReceipt(r TxReceipt) {
...
 			case sub <- r:
-				log.Info("reporter send receipt %v to subscriber", r)
+				log.With().Debug("reporter send receipt to subscriber", r, r.Address)
 			default:
-				log.Debug("reporter would block on subscriber %v", sub)
+				log.Debug("reporter would block on subscriber")
 			}
 		}
-		log.Info("reported receipt to subscribers: %v", r)
+		log.With().Info("reported receipt to subscribers", r, r.Address)
```
lrettig:
```suggestion
		log.With().Debug("reported receipt to subscribers", r, r.Address)
```

events/reporter.go
```diff
@@ -263,16 +267,16 @@ func ReportAccountUpdate(a types.Address) {
 	defer mu.RUnlock()
 
 	if reporter != nil {
-		log.Info("about to report account update for: %v", a)
+		log.With().Info("about to report account update for", a)
 		for sub := range reporter.accountsSubs {
 			select {
 			case sub <- a:
-				log.Info("reporter send account %v to subscriber", a)
+				log.With().Info("reporter send account to subscriber", a)
```
lrettig:
```suggestion
				log.With().Debug("reporter send account to subscriber", a)
```

events/reporter.go
```diff
@@ -263,16 +267,16 @@ func ReportAccountUpdate(a types.Address) {
...
 			default:
-				log.Debug("reporter would block on subscriber %v", sub)
+				log.Debug("reporter would block on subscriber")
 			}
 		}
-		log.Info("reported account update to subscribers: %v", a)
+		log.With().Info("reported account update to subscribers", a)
```
lrettig:
```suggestion
		log.With().Debug("reported account update to subscribers", a)
```

events/reporter.go
```diff
@@ -467,8 +471,8 @@ func UnsubscribeFromReceipts(subscriptionChannel chan TxReceipt) {
 
 // InitializeEventReporter initializes the event reporting interface
 func InitializeEventReporter(url string) error {
-	// By default use zero-buffer channels and non-blocking.
-	return InitializeEventReporterWithOptions(url, 0)
+	// Need to buffer the channels so that slow subscribers can still receive updates
+	return InitializeEventReporterWithOptions(url, 30)
```
lrettig: please make this a parameter rather than hard-coding the number

lrettig (changes requested):
I suggest reducing the verbosity of the logs a little bit - see the suggestions I added. one Info log line per report is enough. the rest should be debug.

events/reporter.go
sidsankhe: Since `err` is of type `NodeError`, `log.Err(err)` does not work since the `NodeError` type does not have an `Error` method. Should I add the `Error` method, or leave this as is?

events/reporter.go
sidsankhe: Same issue as above

events/reporter.go
sidsankhe: It had no information associated with it, because the status update is just a `struct{}{}`. I think this was in response to a previous PR comment

events/reporter.go
sidsankhe: Should I add this to the `Config` struct?

events/reporter.go
noamnelke: I would consider adding the `Field() log.Field` method to this type - then it can be logged directly without wrapping.

events/reporter.go
lrettig: Add it to the default config in config/config.go. Call it something like "EventQueueSize"/`event-queue-size` and give it a reasonable default there, 30 seems okay.

events/reporter.go
```diff
@@ -250,124 +267,224 @@ func ReportAccountUpdate(a types.Address) {
...
+				log.Debug("reporter would block on subscriber")
 			}
 		}
+		log.With().Debug("reported account update to subscribers", a)
 	}
 }
 
 // GetNewTxChannel returns a channel of new transactions
-func GetNewTxChannel() chan TransactionWithValidity {
+// func GetNewTxChannel() chan TransactionWithValidity {
```
lrettig, line 298: removed commented code

lrettig (changes requested):

avive:
I'm trying to build a node from this pr branch but getting an error running `make` or `make install`: https://github.com/protocolbuffers/protobuf/releases/download/v3.6.1/protoc-3.6.1-osx-arm64.zip is requested but seems to be 404 right now. Develop branch make works okay for me. @sidsankhe @lrettig 

Is it okay to just modify the url in isntall-protobuff.sh to:
protoc_url=https://github.com/protocolbuffers/protobuf/releases/download/v3.17.0/protoc-3.17.0-${os}-${arch}.zip

lrettig:
@avive not sure what's going on but there is no protobuf in this branch. It was removed months ago. The only branch that still contains it is the testnet branch (until smapp is fully upgraded to use the new API).

avive:
> @avive not sure what's going on but there is no protobuf in this branch. It was removed months ago. The only branch that still contains it is the testnet branch (until smapp is fully upgraded to use the new API).

`make install` is trying to install an invalid version of protoc. In develop, the version is good. I gues that if there's no protobufs anymore in go-sm then make install should be updated...

lrettig:
> `make install` is trying to install an invalid version of protoc. In develop, the version is good. I gues that if there's no protobufs anymore in go-sm then make install should be updated...

This branch includes the latest develop. Make install was updated. Most likely, you need to delete the `api/pb` directory locally. Best thing to do is to just `git clone` a fresh version of the repository.

avive:
> > `make install` is trying to install an invalid version of protoc. In develop, the version is good. I gues that if there's no protobufs anymore in go-sm then make install should be updated...
> 
> This branch includes the latest develop. Make install was updated. Most likely, you need to delete the `api/pb` directory locally. Best thing to do is to just `git clone` a fresh version of the repository.

OK. Q: is this PR closed to being merged? Eager to test it with interactive clients on testnet or a devnet.

sidsankhe:
@avive Yes, I think this PR is mostly done. Waiting for some final reviews!

api/grpcserver/globalstate_service.go
```diff
@@ -274,13 +283,37 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 		channelReceipt chan events.TxReceipt
 	)
 	if filterAccount {
-		channelAccount = events.GetAccountChannel()
+		if s.accountDataStreamAccountChannel == nil {
+			channelAccount = events.SubscribeToAccounts()
+			if channelAccount == nil {
+				log.Error("Reporter is not running")
```
lrettig, line 289: It's redundant and probably unnecessary to print the error here verbatim right before returning it. Make sure the caller properly handles/surfaces the error, then you can remove all of these `log.Error` statements.

api/grpcserver/globalstate_service.go
```diff
@@ -448,18 +496,54 @@ func (s GlobalStateService) GlobalStateStream(in *pb.GlobalStateStreamRequest, s
...
+			channelReward = events.SubscribeToRewards()
+			if channelReward == nil {
+				log.Error("Reporter is not running")
+				return errors.New("Reporter is not running")
+			}
+		} else {
+			channelReward = s.globalStateStreamRewardsChannel
+		}
+		if channelReward == nil {
+			// what to do here
```
lrettig, line 521: Do you need this check? None of the other channel logic has it. Seems redundant/unnecessary.

blocks/blockoracle_test.go
```diff
@@ -306,7 +306,9 @@ func TestBlockOracleValidatorInvalidProof3(t *testing.T) {
 		// we want a proof with J != 0, so we must have more than one proof in the list
 		if len(proofs) > 1 {
 			proof = proofs[0]
-			// we keep trying until J != 0
+			if len(proofs) <= 1 {
```
lrettig, line 309: Where did this come from? possible merge conflict?

events/reporter.go
```diff
@@ -17,6 +17,9 @@ var reporter *EventReporter
 // we use a mutex to ensure thread safety
 var mu sync.RWMutex
 
+// EventQueueSize is a parameterized length of the channel buffers
+const EventQueueSize int = 30 //TODO: replace this with reading from a file
```
lrettig, line 21:
```suggestion
const EventQueueSize int = 30 // TODO: move this to config
```

Please open an issue to track this and link it here

events/reporter.go
```diff
@@ -250,124 +270,224 @@ func ReportAccountUpdate(a types.Address) {
...
+				log.Debug("reporter would block on subscriber")
 			}
 		}
+		log.With().Debug("reported account update to subscribers", a)
 	}
 }
 
 // GetNewTxChannel returns a channel of new transactions
-func GetNewTxChannel() chan TransactionWithValidity {
+// func GetNewTxChannel() chan TransactionWithValidity {
```
lrettig, line 298: Remove comments

lrettig (changes requested):

nkryuchkov:
@lrettig @avive Will this PR be ready for merging after merge conflicts are resolved?

lrettig:
> @lrettig @avive Will this PR be ready for merging after merge conflicts are resolved?

Good question, need to take a deeper look. There is definitely still an underlying issue here that needs to be resolved.

PR #2390, opened 2021-04-21
[Merged by Bors] - Hare Termination Certification

### LINKED ISSUE #1966: termination round
after termination, a committee signs the hare results.
This should not be part of the Hare protocol. It is merely a step that should be done after consensus is terminated.
This is part of protocol 2.0.

sidsankhe:
## Motivation

Closes #1966 

## TODO

## DevOps Notes

### DIFF
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -70,6 +70,19 @@ func (cpo procReport) Completed() bool {
 	return cpo.completed
 }
 
+type certification struct {
+	id          types.LayerID
+	eligibility uint16
+}
+
+func (c *certification) ID() types.LayerID {
+	return c.id
+}
+
+func (c *certification) Eligibility() uint16 {
+	return c.eligibility
+}
+
 func (proc *consensusProcess) report(completed bool) {
 	proc.terminationReport <- procReport{proc.instanceID, proc.s, proc.preRoundTracker.coinflip, completed}
 }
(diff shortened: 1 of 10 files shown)


lrettig:
Regarding the test that's timing out, I suspect a timer issue. A new round was added to Hare, but the timeout here hasn't been increased:

https://github.com/spacemeshos/go-spacemesh/pull/2390/files#diff-8fcfad77e6ef36a7ff45977166f2e978a81defa5d536f837adc179e848ce0ad6R177

Why is this set to 30 secs?

moshababo:
@sudachen will continue to work on this.

sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

hare/algorithm.go
```diff
@@ -332,15 +353,36 @@ PreRound:
...
 		select {
 		case msg := <-proc.inbox: // msg event
+
 			proc.handleMessage(ctx, msg)
-			if proc.terminating {
+
+			if proc.terminating || (proc.completed && proc.certified) {
+				// TODO: possible error close closed channel
+				// TODO: but tests are awaiting for proc.CloseChannel()...
+				close(proc.CloseChannel())
```
```suggestion
				proc.Close()
```

hare/algorithm.go
```diff
@@ -332,15 +353,36 @@ PreRound:
...
 	for {
 		select {
 		case msg := <-proc.inbox: // msg event
+
 			proc.handleMessage(ctx, msg)
-			if proc.terminating {
+
+			if proc.terminating || (proc.completed && proc.certified) {
+				// TODO: possible error close closed channel
+				// TODO: but tests are awaiting for proc.CloseChannel()...
```
noamnelke: All calls to `close(proc.CloseChannel())` (should be replaced by `proc.Close()`) are immediately followed by `return`. The `eventLoop` is only called once per process.

I don't think this error is possible...

hare/algorithm.go
```diff
@@ -332,15 +353,36 @@ PreRound:
...
 				return
 			}
+
 		case <-endOfRound: // next round event
+
 			proc.onRoundEnd(ctx)
+
+			if proc.terminating || (proc.completed && proc.certified) || proc.k == certifyRound {
+				// TODO: possible error close closed channel
+				// TODO: but tests are awaiting for proc.CloseChannel()...
```
noamnelke: Same

hare/algorithm.go
```diff
@@ -332,15 +353,36 @@ PreRound:
...
+				close(proc.CloseChannel())
```
```suggestion
				proc.Close()
```

hare/algorithm.go
```diff
@@ -350,17 +392,17 @@ PreRound:
 					log.Int("limit", proc.cfg.LimitIterations),
 					log.Uint32("current_k", k),
 					proc.instanceID)
-				proc.report(notCompleted)
+				close(proc.CloseChannel())
```
```suggestion
				proc.Close()
```

hare/algorithm.go
```diff
@@ -537,11 +581,15 @@ func (proc *consensusProcess) onRoundEnd(ctx context.Context) {
 
 // advances the state to the next round.
 func (proc *consensusProcess) advanceToNextRound(ctx context.Context) {
-	k := proc.addToK(1)
-	if k >= 4 && k%4 == 0 {
-		proc.WithContext(ctx).Event().Warning("starting new iteration",
-			log.Uint32("current_k", k),
-			proc.instanceID)
+	if proc.completed && proc.currentRound() == notifyRound {
+		proc.k = certifyRound
```
noamnelke: I've noticed that `proc.k` has accessors that use `atomic` operations (`addToK`, `getK` and `setK`) that you've stopped using. In my Hare round clock PR I've also used `k` directly. I think this might be a mistake and we should switch back to using only the accessors. If you think `k` can't be accessed concurrently then we should clean up and remove the accessors.

hare/certifytracker_test.go
```diff
@@ -0,0 +1,14 @@
+package hare
```
noamnelke, line 1: no actual tests

hare/consensus_test.go
```diff
@@ -151,15 +151,24 @@ func createConsensusProcess(tb testing.TB, isHonest bool, cfg config.Config, ora
...
 	proc.SetInbox(c)
 
+	go func() {
+		for range certs {
+		}
+	}()
+	go func() {
+		for range output {
+		}
+	}()
```
noamnelke: I assume this is just to consume the channels - should add a comment to make it clear.

noamnelke (dismissed):
I think the only major issue is with the concurrent access to `proc.k` - looks great otherwise!

sudachen:
bors try

sudachen:
Bors merge

hare/algorithm.go
```diff
@@ -330,17 +351,36 @@ PreRound:
...
 
 	for {
 		select {
 		case msg := <-proc.inbox: // msg event
+
 			proc.handleMessage(ctx, msg)
-			if proc.terminating {
+
+			if proc.terminating || (proc.completed && proc.certified) {
+				// TODO: possible error close closed channel
```
noamnelke, line 369: I think this is not possible, because there's a return after every close and the event loop only runs once. If you think it's possible - please explain (we can fix it by checking for a closed channel in `proc.Close()` before closing, but I don't think it's needed today). Otherwise, let's remove the `TODO`.

hare/algorithm.go
```diff
@@ -330,17 +351,36 @@ PreRound:
...
+				proc.Close()
 				return
 			}
+
 		case <-endOfRound: // next round event
+
 			proc.onRoundEnd(ctx)
+
+			if proc.terminating || (proc.completed && proc.certified) || proc.getK() == certifyRound {
+				// TODO: possible error close closed channel
```
noamnelke, line 379: same

noamnelke (approved):

noamnelke:
bors merge

Issue #2067, opened 2020-07-09
Remove time.Sleep from tests


lrettig:
## Description
We use `time.Sleep` throughout our tests while e.g. waiting for a goroutine or service to spin up. In general this isn't a great idea because it's fragile, and also because it slows down the tests quite a bit. Most or all of these Sleep statements can probably be replaced with concurrency tools like channels, waitgroups, mutexes, etc.

See one example of such a fix here: 9bc6794f29f7bb7d0deab7aae2870af2c11adae1

## Affected code
Many tests, grep for `time.Sleep`

u5surf:
@lrettig I would like to try to this issue

problematic case is e.g. 
https://github.com/spacemeshos/go-spacemesh/blob/b49a23f5d839f0d2d24b5b4f34f2fd372eeb97e8/cmd/sync/sync_test.go#L24

lrettig:
Welcome to give it a shot! Not sure how much experience you have with concurrency in go but it would be great practice. I haven't analyzed this test in depth so I can't explain exactly what's going on here or how I'd fix it, but at a glance, it looks like the syncer should notify a channel every time it processes a layer, and the test should pass after 20 layers have been processed, or fail if the 20 secs passes first instead.

lrettig:
Here's another good example: https://github.com/spacemeshos/go-spacemesh/commit/4f35f7b7e899788bfbd945aff3abd507ed4f54c3

lrettig:
Another example: #2422

PR #2393, opened 2021-04-23
[Merged by Bors] - Weak coin for self healing

### LINKED ISSUE #1857: Weak coin implementation
As described in:
https://github.com/spacemeshos/SMIPS/issues/28

Todays verifying tortoise algorithm simply verifies that honest blocks vote the same way to make it easier to tally votes for each honest block.
Voting is done by tallying up past votes and voting accordingly - if there are enough prior votes to support a block vote the for the block as well. this is done by supporting blocks who view the same as our node.

This protocol is “fragile”, in the sense that its security depends on all of our underlying assumptions holding continuously since genesis. For example, if at layer i the Hare protocol fails (e.g.,due to some very low probability event), honest parties may not agree on the validity of the blocks in layer i. Suppose honest parties are evenly split about the validity of some block B in layer i - half accept it as valid and half don’t.In subsequent layers, the vote margin for B will be very close to 0 - and an adversary can succeed in a balancing attack: the adversary keeps the margin small indefinitely by always voting “with the minority”; since the margin is already small, the adversary does not have to expend many resources in order to “pad”the minority so that the total number of positive and negative votes remains equal. Thus, honest parties will never reach consensus onB’s validity.

In case there is no consensus on how to vote for a specific block, meaning the for and against votes are pretty balanced we need to make a decision on how to vote.
For this we need a shared coin that will cause all blocks in a certain layer to vote in the same direction - to faster converge on a value for all blocks.

Weak coin requirements
The main idea is to randomise honest parties’ votes when the vote margin is small - but in a coordinated way. For intuition, suppose we had a sequence of coin flips that is:
– In Consensus: at the start of layer i+ 1, all honest parties agree on the result of the i-th coin flip.
– Independent and Unbiased: each coin comes up heads and tails with equal probability.
– Unpredictable: the adversary cannot guess the result of the i-th coin flip until the end of layer i

A coin is considered weak because the adversary can guess the result of the coin flip early some of the time, or bias the result so that heads and tails are no longer equally likely. A weak coin is sufficient as the adversary can’t predict the result with more than constant probability, and can’t bias it “too far”.

lrettig:
## Motivation
Closes #1857

## Changes
- Hare calculates a weak coin for each layer based on VRF messages received in the pre-round, and sends it to the meshdb (regardless of whether hare succeeded or not)
- Meshdb stores this in memory (not actually in the DB)

## Test Plan
- Adds unit tests for hare pre-round tracker, hare, and mesh
- Adds a weakcoin mocker to the superhare
- Adds a multinode integration app test that makes sure superhare writes a weak coin for each layer to meshdb, and that the coin values agree across all nodes

## TODO

## DevOps Notes


hare/hare.go
```diff
@@ -68,6 +71,7 @@ type Hare struct {
 	outputChan chan TerminationOutput
 	mu         sync.RWMutex
 	outputs    map[types.LayerID][]types.BlockID
+	coinflips  map[types.LayerID]bool
```
lrettig: Remove this

cmd/node/app_test.go
```diff
@@ -220,7 +223,7 @@ type TestScenario struct {
 func txWithUnorderedNonceGenerator(dependencies []int) TestScenario {
 	acc1Signer, err := signing.NewEdSignerFromBuffer(util.FromHex(apicfg.Account2Private))
 	if err != nil {
-		log.Panic("Could not build ed signer err=%v", err)
+		log.Panic("could not build ed signer err=%v", err)
```
lrettig: Log.err

turbohare/hare.go
```diff
@@ -5,39 +5,68 @@ import (
...
-
+	close(h.closeChannel)
 }
 
 // GetResult is the implementation for receiving consensus process result
 func (h *SuperHare) GetResult(id types.LayerID) ([]types.BlockID, error) {
-	blks, err := h.blocks.LayerBlockIds(id)
+	blks, err := h.mesh.LayerBlockIds(id)
 	if err != nil {
-		log.Error("WTF SUPERHARE?? %v err: %v", id, err)
```
lrettig, line 40: Use logger

cmd/node/app_test.go
```diff
@@ -321,11 +324,37 @@ func reachedEpochTester(dependencies []int) TestScenario {
...
+		// there will be no coin value for layer zero since ticker delivers only new layers
+		// and there will be no coin value for the last layer
+		for layerID := types.LayerID(1); layerID < lastLayer; layerID++ {
+			coinflip, exists := app.mesh.DB.GetCoinflip(context.TODO(), layerID)
+			if !exists {
+				suite.Fail("no weak coin value", "node %v layer %v last layer %v", i, layerID, lastLayer)
+				continue
+			}
+			if _, layerExists := globalCoin[layerID]; layerExists {
+				suite.Equal(globalCoin[layerID], coinflip, "bad weak coin value on node %v layer %v last layer %v", i, layerID, lastLayer)
```
noamnelke:
```suggestion
			if gc, layerExists := globalCoin[layerID]; layerExists {
				suite.Equal(gc, coinflip, "bad weak coin value on node %v layer %v last layer %v", i, layerID, lastLayer)
```

cmd/node/app_test.go
```diff
@@ -423,7 +452,9 @@ func (suite *AppTestSuite) validateBlocksAndATXs(untilLayer types.LayerID) {
 
 		for i := types.LayerID(5); i <= untilLayer; i++ {
 			lyr, err := ap.mesh.GetLayer(i)
-			assert.NoError(suite.T(), err, "couldn't get a validated layer from db layer %v", i)
+			if err != nil {
+				log.With().Panic("couldn't get a validated layer from db", i, log.Err(err))
+			}
```
noamnelke: wouldn't a `require.NoError()` instead of `assert.NoError()` do the trick here? I'm not sure how it behaves when called in a goroutine, so it's a genuine question.

hare/algorithm.go
```diff
@@ -52,12 +52,12 @@ const (
...
-// It consists of the layer id, the set we agreed on (if available) and a flag to indicate if the CP completed.
+// procReport is the termination report of the CP
 type procReport struct {
-	id        instanceID
-	set       *Set
-	completed bool
+	id        instanceID // layer id
+	set       *Set       // agreed-upon set
+	coinflip  bool       // weak coin value
+	completed bool       // whether the CP completed
```
noamnelke, line 62: renaming the fields would be nicer 😢

hare/algorithm_test.go
```diff
@@ -464,42 +470,34 @@ func TestConsensusProcess_currentRound(t *testing.T) {
...
-	select {
-	case <-timeout.C:
-		r.Equal(0, len(proc.pending))
-	case <-tk.C:
-		if 0 == len(proc.pending) {
-			return
-		}
-	}
+	time.Sleep(100 * time.Microsecond)
+	r.Zero(len(proc.pending))
```
noamnelke: looks like the original code was meant to be run in a loop... the bad implementation may have caused the test to falsely succeed, I wonder if the new implementation won't cause it to falsely fail sometimes (e.g. on weak machines with a single CPU)?

noamnelke:
```suggestion
	r.Len(proc.pending, 0)
```
This will produce a nicer error message (incl. the actual length).

hare/algorithm_test.go
```diff
@@ -226,35 +226,40 @@ func TestConsensusProcess_handleMessage(t *testing.T) {
 	r := require.New(t)
 	net := &mockP2p{}
 	broker := buildBroker(net, t.Name())
-	broker.Start(context.TODO())
+	_ = broker.Start(context.TODO())
```
noamnelke:
```suggestion
	r.NoError(broker.Start(context.TODO()))
```

noamnelke (approved):

hare/algorithm_test.go
lrettig: We have these all over our code. They're messy. See #2067. I replaced this with [`Eventually`](https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually) which is a bit cleaner.

hare/algorithm.go
lrettig: You mean rather than adding the explanatory comments?

cmd/node/app_test.go
lrettig: It doesn't work inside a goroutine. But this isn't inside a goroutine. I may have missed that, or I may have just wanted failure here to short-circuit execution of the rest of the test. I fixed it.

lrettig:
bors merge

lrettig:
bors merge

PR #2397, opened 2021-04-25
Prevent sync from stalling

lrettig:
## Motivation
Add two more timeouts to sync, so that it doesn't get stuck forever, e.g., when the internet connection is temporarily lost

## Changes
Adds two sync timeouts, one on the high-level request to sync a layer, another on a lower-level request to fetch block data. Both of these can hang indefinitely waiting for a callback which may never be called. (I'm not totally sure why the callback is sometimes not called, but I've experienced this when my internet connection at home is temporarily interrupted.)

## Test Plan
N/A

## TODO

## DevOps Notes

### DIFF
--- a/blocks/blocks.go
+++ b/blocks/blocks.go
@@ -201,7 +201,6 @@ func (bh *BlockHandler) fastValidation(block *types.Block) error {
 		return err
 	}
 	return nil
-
 }
 
 func validateUniqueTxAtx(b *types.Block) error {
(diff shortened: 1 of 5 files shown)


nkryuchkov:
@lrettig What's the status of this PR?

dshulyak:
i think this pr lost its relevance long time ago (cc @countvonzero )

PR #2394, opened 2021-04-23
[Merged by Bors] - Modify Hare eligibility to use Tortoise beacon

lrettig:
## Motivation
Part of self-healing. Related to #2356. Removes another circular dependency in Hare eligibility: rather than relying on a layer pattern based on a set of contextually valid blocks from a safe layer, instead use the Tortoise beacon value.

## Changes
- Adds a simple, mock Tortoise beacon provider to node (waiting for #2267)
- Update Hare eligibility beacon to use this rather than a layer pattern for safe layer

## Test Plan
Existing tests updated. No new tests are required, as the Tortoise beacon is just mocked for now.

## TODO

- [x] Explain motivation or link existing issue(s)
- [x] Test changes and document test plan
- [ ] Update documentation as needed

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

Issue #2369, opened 2021-04-09
blocks, tortoise: merge for/against exceptions into a single list, vote neutral on layers instead of blocks


lrettig:
As described in https://github.com/spacemeshos/SMIPS/issues/2, the initial version of verifying tortoise includes three explicit lists of exception votes:

```
Block {
   Base_Block: <block_id>
   For_Exception_List: [<block_id>, <block_id>, <block_id>...] // List of blocks I vote For and Base_Block doesn't
   Against_Exception_List: [<block_id>, <block_id>, <block_id>...] // List of blocks I vote Against and Base_Block doesn't
   Neutral_Exception_List: [<block_id>, <block_id>, <block_id>...] // List of blocks I vote neutral and Base_Block doesn't or blocks that just come after Base_Block.
}
```

As discussed on the R&D call with Tal this week, we can optimize this in a few ways:

- blocks only ever vote neutral on an entire layer, never on individual blocks in a layer (and only in the case where they're still waiting for a Hare result for that layer). So there's no need to list individual block ids in the neutral list, just layer IDs.
- the For and Against exception lists can be collapsed down to a single list. For any block on this list, the block's vote is implicitly the _opposite_ of the base block's vote.

Regarding neutral votes in the base block, by default a block citing it should also be considered to vote neutral for the same layers, unless the block includes _any_ explicit for or against votes for _any_ blocks in that layer (this may already be how things work, I'm not sure)

CC @y0sher, @sudachen 

Addendum: a block does not need to encode an explicit vote _for_ its base block, this can be made automatic and implicit. ([discussion](https://spacemesh.slack.com/archives/C89QJJY3G/p1635846350093700))

lrettig:
Related: in block fast validation, make sure the exception lists are valid (i.e., don't contain contradictions), and invalidate a block if they do:

https://github.com/spacemeshos/go-spacemesh/blob/9d3a379994e3652eb0378d2cc555588bba91005e/blocks/blocks.go#L153-L159

dshulyak:
should also incorporate https://github.com/spacemeshos/go-spacemesh/issues/2674 when this will be implemented

PR #2400, opened 2021-04-28
[Merged by Bors] - Self-healing

### LINKED ISSUE #2203: differentiate between not terminated hare and terminated with empty set in tortoise and sync
## Description

with the verifying tortoise we are now syncing on hare results (aka vote input vector) and using them for verifying. 
we need to be able to differentiate between not terminated hare and terminated with empty set in both, 
if terminated with empty set we should know to vote against every block in that layer
if not terminated we abstain until terminated.

currently we always act as if the hare was not terminated and abstain on all blocks when we get empty set from sync or hare itself.

## Affected code
- hare
- tortoise
- sync

### LINKED ISSUE #2687: How to weight implicit base block votes in tortoise
if a base block has weight X and a block based on it has weight Y, what weight should we use for the "implicit" opinion of X that's included in Y's vote?

lrettig:
## Motivation
See https://github.com/spacemeshos/SMIPS/issues/46

Requires #2394, #2393, #2357
Closes #2203
Closes #2687

## Changes
- adds zdist param: whereas hdist is the "Hare lookback" distance (the number of layers for which we consider hare results/local input vector rather than global opinion), zdist is the "Hare result wait" distance (the number of layers we're willing to wait for Hare to finish before reverting to invalidating layers with no input vector)
- Hare explicitly reports failed CPs to the mesh, with the results stored in memory

## Test Plan
TBD

## TODO

- [x] Finish rescoring block goodness after healing
- [x] Rescoring block unit test
- [x] Block voting weight unit test
- [x] Make sure healing -> verifying tortoise handoff is working
- [x] App test (for decay in active set size)
- [x] Split-then-rejoin test (two way split, three way split w/o majority fork)
- [x] Late blocks unit test(s)
- [x] Correctly weight votes from blocks with late atxs/those with a bad beacon value (will follow up in separate issue: #2540)
- [x] Multi tortoise unit tests

## DevOps Notes


tortoise/verifying_tortoise.go
```diff
@@ -22,6 +23,14 @@ type blockDataProvider interface {
 	Retrieve(key []byte, v interface{}) (interface{}, error)
 }
 
+//func blockArrayToMap(ar []types.BlockID) (mp map[types.BlockID]struct{}) {
```
lrettig: Remove

tortoise/verifying_tortoise.go
```diff
@@ -156,33 +173,35 @@ func (t *turtle) getSingleInputVectorFromDB(lyrid types.LayerID, blockid types.B
...
 	for _, exceptionBlockID := range diffList {
 		if exceptionBlock, err := t.bdp.GetBlock(exceptionBlockID); err != nil {
-			t.logger.Panic("can't find %v from block's %v diff ", exceptionBlockID, sourceBlockID)
+			logger.With().Error("inconsistent state: can't find block from block's diff",
+				log.FieldNamed("exception_block_id", exceptionBlockID))
+			return false
 		} else if exceptionBlock.LayerIndex < layerIndex {
-			t.logger.With().Debug("not marking block good because it points to block older than base block",
-				sourceBlockID,
+			// TODO: can a base block exception list point to blocks in the same layer, or must they all be newer?
```
lrettig: Update with latest info from Tal

tortoise/verifying_tortoise.go
```diff
@@ -392,29 +441,35 @@ func (t *turtle) processBlock(block *types.Block) error {
...
+	// TODO: this logic would be simpler if For and Against were a single list
 	// TODO: save and vote against blocks that exceed the max exception list size (DoS prevention)
 	opinion := make(map[types.BlockID]vec)
-	for blk, vote := range baseBlockOpinion.BlocksOpinion {
-		opinion[blk] = vote
-	}
-
 	for _, b := range block.ForDiff {
-		opinion[b] = opinion[b].Add(support.Multiply(t.BlockWeight(block.ID(), b)))
+		opinion[b] = support.Multiply(t.BlockWeight(block.ID(), b))
```
lrettig: Check for collisions here? Maybe in malicious blocks

tortoise/verifying_tortoise.go
```diff
@@ -423,179 +478,180 @@ func (t *turtle) HandleIncomingLayer(newlyr *types.Layer) (pbaseOld, pbaseNew ty
...
-		rawLayerInputVector, err := t.bdp.GetLayerInputVector(i)
+		// get the local opinion (input vector) for this layer. below, we calculate the global opinion on each block in
+		// the layer and check if it agrees with this local opinion.
+		rawLayerInputVector, err := t.bdp.GetLayerInputVector(candidateLayer)
 		if err != nil {
-			// this sets the input to abstain
-			t.logger.With().Warning("input vector abstains on all blocks", i)
+			if errors.Is(err, mesh.ErrInvalidLayer) {
+				// Hare already failed for this layer, so we want to vote against all blocks in the layer: an empty
+				// map will do the trick
```
lrettig: List not map

tortoise/verifying_tortoise.go
```diff
@@ -423,179 +478,180 @@ func (t *turtle) HandleIncomingLayer(newlyr *types.Layer) (pbaseOld, pbaseNew ty
...
+				logger.With().Warning("input vector abstains on all blocks", candidateLayer)
+			}
 		}
-		layerVotes := t.inputVectorForLayer(layerBlockIds, rawLayerInputVector)
-		if len(layerVotes) == 0 {
-			t.logger.With().Warning("no blocks in input vector, can't verify layer", i)
+		localLayerOpinionVec := t.inputVectorForLayer(layerBlockIds, rawLayerInputVector)
+		if len(localLayerOpinionVec) == 0 {
+			logger.With().Warning("no blocks in input vector, can't verify layer", candidateLayer)
 			break
```
lrettig, line 523: Why break not continue?

tortoise/verifying_tortoise.go
```diff
@@ -392,214 +450,406 @@ func (t *turtle) processBlock(block *types.Block) error {
...
+	return nil
+}
+
+// return the set of blocks we currently consider valid for the layer. factors in both local and global opinion,
+// depending how old the layer is.
+func (t *turtle) layerOpinionVector(ctx context.Context, layerID types.LayerID) (opinionVec []types.BlockID, err error) {
+	logger := t.logger.WithContext(ctx).WithFields(layerID)
+
+	// for layers older than hdist, we vote according to global opinion
+	if layerID < t.Last-t.Hdist {
```
lrettig: To do: what if this layer hasn't been verified yet?

lrettig: Need to calculate global opinion

lrettig:
bors try

tortoise/verifying_tortoise.go
```diff
@@ -261,68 +314,95 @@ func (t *turtle) BaseBlock() (types.BlockID, [][]types.BlockID, error) {
...
 
-		layerBlockIds, err := t.bdp.LayerBlockIds(i)
+		layerBlockIds, err := t.bdp.LayerBlockIds(layerID)
 		if err != nil {
 			if err != leveldb.ErrClosed {
-				// todo: empty layer? maybe skip verify differently
-				t.logger.Panic("can't find old layer %v", i)
+				// this should not happen! we only look at layers up to the last processed layer, and we only process
+				// layers with valid block data.
+				logger.Error("no block ids for layer in database")
```
lrettig: print err

tortoise/verifying_tortoise_test.go
```diff
@@ -361,65 +368,221 @@ func TestTurtle_Recovery(t *testing.T) {
...
+	mdb, teardown := getPersistentMesh()
 	defer func() {
-		if r := recover(); r != nil {
-			t.Log("Recovered from", r)
-		}
+		require.NoError(t, teardown())
+	}()
+
+	getHareResults := func(l types.LayerID) ([]types.BlockID, error) {
+		return mdb.LayerBlockIds(l)
```
lrettig: Is this necessary? Do these methods have different signatures? Looks like no. Remove the wrapper

tortoise/verifying_tortoise.go
```diff
@@ -362,244 +450,497 @@ func (t *turtle) persist() error {
...
+	// this is the full range of unverified layers that we might possibly be able to verify at this point.
+	// Note: t.Verified is initialized to the effective genesis layer, so the first candidate layer here necessarily
+	// follows and is post-genesis. There's no need for an additional check here.
+candidateLayerLoop:
+	for candidateLayerID := t.Verified+1; candidateLayerID < targetLayerID; candidateLayerID++ {
+		logger := logger.WithFields(log.FieldNamed("candidate_layer", candidateLayerID))
+
+		// it's possible that self-healing already validated a layer
+		if t.Verified >= candidateLayerID {
+			logger.Info("self-healing already validated this layer")
```
lrettig: verify not validate

tortoise/verifying_tortoise.go
```diff
@@ -362,244 +450,497 @@ func (t *turtle) persist() error {
...
-			if globalOpinion != vote {
-				// TODO: trigger self healing after a while?
-				t.logger.With().Warning("global opinion on block differs from our vote, cannot verify layer",
-					blk,
-					log.String("global_opinion", globalOpinion.String()),
-					log.String("vote", vote.String()))
-				break layerLoop
+				// short-circuit processing of this layer, but allow verification of later layers to continue after
+				// self-healing has finished
+				break candidateLayerLoop
```
lrettig: continue not break

tortoise/util.go
```diff
@@ -21,15 +20,36 @@ func (a vec) Field() log.Field {
...
 func (a vec) Multiply(x uint64) vec {
-	a[0] = a[0] * x
-	a[1] = a[1] * x
+	one := a[0] * x
+	two := a[1] * x
+	// prevent overflow/wraparound
+	if x != 0 && one/x != a[0] || two/x != a[1] {
+		panic("vector arithmetic overflow")
+	}
 	return a
```
noamnelke: You're never actually doing the multiplication, only checking for overflow and then discarding the result.

You're missing either (in-place multiplication):
```go
a[0] = one
a[1] = two
```
or returning a copy of the vector with the multiplication result.

tortoise/util.go
```diff
@@ -21,15 +20,36 @@ func (a vec) Field() log.Field {
...
 }
 
+func (a vec) netVote() int64 {
+	// prevent overflow/wraparound
+	one := int64(a[0])
+	two := int64(a[1])
+	diff := one - two
+	if diff > one {
+		panic("vector arithmetic overflow")
+	}
```
noamnelke: A `uint64` isn't guaranteed to fit into an `int64` which is why you need a check here, but the check you do isn't enough. If `two` overflows it will be negative, causing `diff` to be greater than `one`, but if `one` overflows and `two` doesn't the method won't panic.

You can separately check if each component overflows before doing the subtraction, but that would mean that we just don't support any values in vectors that don't fit in a `uint32` - so a better solution would be to change to `uint32` in vectors and fail when first trying to place an overflowing value there. Then the conversion from `uint32` to `int64` is guaranteed to succeed and no check is necessary.

Alternatively, we can do a calculation that only overflow when the result doesn't fit in an `int64` (solution 1) or return a `uint64` + sign bit (solution 2).

Solution 1 (no code changes outside `netVote`):
```go
abs := uint64()
sign := 1
if a[0] > a[1] {
	res = a[0] - a[1]
} else {
	res = a[1] - a[0]
	sign = -1
}
if int64(res) < 0 {
	panic("vector arithmetic overflow")
}
return int64(res) * sign
```

Solution 2 (`netVote` returns `absNetVote` + `positiveNetVote` - requires code changes, but cannot overflow):
```go
if a[0] > a[1] {
	return a[0] - a[1], true
} else {
	return a[1] - a[0], false
}
```

tortoise/util.go
```diff
@@ -53,7 +73,7 @@ func (a vec) String() string {
 
 func calculateOpinionWithThreshold(logger log.Log, v vec, layerSize int, theta uint8, delta float64) vec {
 	threshold := float64(theta) / 100 * delta * float64(layerSize)
-	netVote := float64(v[0] - v[1])
+	netVote := float64(v.netVote())
```
noamnelke: How is it ok that we use floats here? I thought that they are unacceptable in consensus code due to varying implementation / behaviour across processor architectures 🤔

noamnelke (dismissed):

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

hare/hare.go
```diff
@@ -61,7 +62,7 @@ type Hare struct {
 
 	sign Signer
 
-	msh     meshProvider
+	mesh    meshProvider
```
y0sher, line 65: `msh` was used because `mesh` is a package name hence can be confusing.

hare/hare.go
```diff
@@ -156,20 +165,14 @@ var ErrTooLate = errors.New("consensus process finished too late")
...
+	blocks := make([]types.BlockID, 0, len(set.values))
 	for v := range set.values {
-		blocks[i] = v
-		i++
-	}
-
-	// check validity of the collected output
-	if !h.validate(blocks) {
-		h.WithContext(ctx).Error("failed to validate the collected output set")
+		blocks = append(blocks, v)
```
y0sher: at least in older versions of go, creating the slice and assigning each index when the size is known was more efficient than append. though I get append is more readable.

tortoise/util.go
```diff
@@ -1,79 +1,95 @@
...
-	return log.String("vote_vector", fmt.Sprint(a))
+	return log.String("vote_vector", fmt.Sprintf("%s (+%d, -%d)", a, a[0], a[1]))
 }
 
 func (a vec) Add(v vec) vec {
-	return vec{a[0] + v[0], a[1] + v[1]}
+	a[0] += v[0]
+	a[1] += v[1]
+	// prevent overflow/wraparound
+	if a[0] < v[0] || a[1] < v[1] {
```
y0sher, line 28: maybe we could check before adding instead of check that overflow has happened?

tortoise/verifying_tortoise.go
```diff
@@ -36,11 +61,38 @@ func blockMapToArray(m map[types.BlockID]struct{}) []types.BlockID {
 type turtle struct {
 	logger log.Log
 
-	bdp blockDataProvider
+	atxdb atxDataProvider
+	bdp   blockDataProvider
+	clock layerClock
+
+	// TODO: can some or all of these be unexported? is this an issue for serialization?
```
y0sher: it is a serialization issue, w/e you want to be serialized (at least with current xdr impl) has to be exported.

tortoise/verifying_tortoise.go
```diff
@@ -49,73 +101,132 @@ type turtle struct {
...
 		BlockOpinionsByLayer: make(map[types.LayerID]map[types.BlockID]Opinion, hdist),
 		MaxExceptions:        int(hdist) * avgLayerSize * 100,
+		RerunInterval:        rerun,
 	}
-	return t
 }
 
-func (t *turtle) init(genesisLayer *types.Layer) {
+// cloneTurtle creates a new verifying tortoise instance using the params of this instance
+func (t *turtle) cloneTurtle() *turtle {
```
y0sher: the name got me confused at first, thought you were cloning the tables as well, maybe mention you're cloning or copying configuration in the name.

tortoise/verifying_tortoise.go
```diff
@@ -130,19 +241,23 @@ func blockIDsToString(input []types.BlockID) string {
 	return str
 }
 
-// TODO: cache but somehow check for changes (hare that didn't finish finishes..) maybe check hash?
-func (t *turtle) getSingleInputVectorFromDB(lyrid types.LayerID, blockid types.BlockID) (vec, error) {
-	if !lyrid.After(types.GetEffectiveGenesis()) {
+// returns the binary local opinion on the validity of a block in a layer (support or against).
+// abstain is not an option here: if the current opinion on a layer is abstain, an opinion on
+// a block in that layer will be against.
+// TODO: does this ever need to return abstain?
```
y0sher: if we know hare wasn't terminated yet we should return abstain, this is why the error case is here. I assumed a terminated hare with empty set is perfectly valid so no err but if there are no results return an error to indicate it's still running.

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
 
-		// TODO: how to differentiate between missing Hare results, and Hare failed/gave up waiting?
-		layerInputVector, err := t.bdp.GetLayerInputVectorByID(i)
+		// helper function for adding diffs
+		addDiffs := func(bid types.BlockID, voteClass string, voteVec vec, diffMap map[types.BlockID]struct{}) {
+			if v, ok := baseBlockOpinion.BlockOpinions[bid]; !ok || simplifyVote(v) != voteVec {
+				logger.With().Debug("added vote diff",
+					log.FieldNamed("diff_block", bid),
+					log.String("diff_class", voteClass))
+				if layerID.Before(baseBlockLayerID) {
```
y0sher, line 441: should we allow this to happen? why not add the diff only if the layer is newer? it sounds to me it might be required for self healing.

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
-					log.String("diff", "against"))
-				againstDiff[b] = struct{}{}
-			}
-		}
+		// Finally, we need to consider the case where the base block supports a block in this layer that is not in our
+		// input vector (e.g., one we haven't seen), by adding a diff against the block.
+		// We do not explicitly add votes against blocks that the base block does _not_ support, since by not voting to
+		// support a block in an old layer, we are implicitly voting against it. But if the base block does explicitly
+		// support a block and we disagree, we need to add a vote against here.
+		// TODO: this is not currently possible since base block opinions aren't indexed by layer. See
```
y0sher, line 483: you can always get full block from db and check the layer though it's quite a lot of overhead.

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
 }
 
-func (t *turtle) BlockWeight(voting, voted types.BlockID) int {
-	return 1
+// voteWeight returns the weight to assign to one block's vote for another.
+// Note: weight depends on more than just the weight of the voting block. It also depends on contextual factors such as
+// whether or not the block's ATX was received on time, and on how old the layer is.
+// TODO: for now it's probably sufficient to adjust weight based on whether the ATX was received on time, or late, for
+//   the current epoch. See https://github.com/spacemeshos/go-spacemesh/issues/2540.
+// TODO: does blockVotedOn ever matter here? can we remove it?
```
y0sher: unless you see some future use, remove blockVotedOn, I assumed it's needed when creating the stub but it was just a guess.

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
 
-// RecoverVerifyingTortoise retrieve latest saved tortoise from the database
+// RecoverVerifyingTortoise retrieves the latest saved tortoise from the database
 func RecoverVerifyingTortoise(mdb retriever) (interface{}, error) {
 	return mdb.Retrieve(mesh.TORTOISE, &turtle{})
 }
 
-func (t *turtle) processBlock(block *types.Block) error {
+func (t *turtle) processBlock(ctx context.Context, block *types.Block) error {
+	logger := t.logger.WithContext(ctx).WithFields(
```
y0sher, line 562: won't this add double key fields when logging line L563 or L564?

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
+
+	calcWeightedVote := func(block *types.Block, votedOn types.BlockID, baseVote vec) (vec, error) {
+		weight, err := t.voteWeight(ctx, block, votedOn)
+		if err != nil {
+			return baseVote, fmt.Errorf("error processing block %v in block %v for diff list: %w", votedOn, block.ID(), err)
+		}
+		return baseVote.Multiply(weight), nil
 	}
 
+	// TODO: this logic would be simpler if For and Against were a single list
```
y0sher, line 597: agreed

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
+				"block in base block opinion list not in db! "+
+					"voting_block_id: %v, voting_block_layer: %v, base_block_id: %v, base_block_layer: %v, block_id: %v",
+				block.ID().String(), block.LayerIndex, baseBlock.ID(), baseBlock.LayerIndex, blk.String())
+		}
+		if fblk.LayerIndex.Before(t.LastEvicted) {
+			continue
+		}
+
+		// add base block vote only if there weren't already exceptions support/against/abstain for this block
+		if _, exists := opinion[blk]; !exists {
```
y0sher, line 635: I was under the impression that the baseblock votes are "added" implicitly.. 
also, do we need to weight these votes as well?

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
-				log.FieldNamed("base_block", b.BaseBlock))
-		} else if baseBlock, err := t.bdp.GetBlock(b.BaseBlock); err != nil {
-			t.logger.Panic(fmt.Sprint("block not found ", b.BaseBlock, "err", err))
-		} else if t.checkBlockAndGetInputVector(b.ForDiff, "support", support, b.ID(), baseBlock.LayerIndex) &&
-			t.checkBlockAndGetInputVector(b.AgainstDiff, "against", against, b.ID(), baseBlock.LayerIndex) &&
-			t.checkBlockAndGetInputVector(b.NeutralDiff, "abstain", abstain, b.ID(), baseBlock.LayerIndex) {
-			// (2) all diffs appear after the base block and are consistent with the input vote vector
-			t.logger.With().Debug("marking good block", b.ID(), b.LayerIndex)
+		if err := t.processBlock(ctx, b); err != nil {
+			logger.With().Error("error processing block", log.Err(err))
```
y0sher, line 686: should blocks that we had errors processing be added to `filteredBlocks`? if not we should probably continue here.

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
+	return nil
+}
+
+func (t *turtle) scoreBlocks(ctx context.Context, blocks []*types.Block) {
+	logger := t.logger.WithContext(ctx)
+	logger.With().Debug("marking good blocks", log.Int("count", len(blocks)))
+	numGood := 0
+	for _, b := range blocks {
+		if t.determineBlockGoodness(ctx, b) {
+			// note: we have no way of warning if a block was previously marked as not good
```
y0sher, line 719: indeed we can't check the list, but if we received this block previously and it isn't in the list, doesn't it mean we thought it wasn't good?

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
+		block.ID(),
+		block.LayerIndex,
+		log.FieldNamed("base_block_id", block.BaseBlock))
+	// Go over all blocks, in order. Mark block i "good" if:
+	// (1) the base block is marked as good
+	if _, good := t.GoodBlocksIndex[block.BaseBlock]; !good {
+		logger.Debug("base block is not good")
+	} else if baseBlock, err := t.bdp.GetBlock(block.BaseBlock); err != nil {
+		logger.With().Error("inconsistent state: base block not found", log.Err(err))
+	} else if true &&
```
y0sher, line 748: I don't see why you need this `true` here?
also I know it's a style preference but I think 2 ifs and returns are clearer than `if else if else if`

tortoise/verifying_tortoise.go
```diff
@@ -194,428 +309,936 @@ func (t *turtle) checkBlockAndGetInputVector(
...
+				// later layers to continue
+				if t.Verified.After(lastVerified) {
+					// rescore goodness of blocks in all intervening layers on the basis of new information
+					// TODO: this is inefficient and we can probably do better
+					// see https://github.com/spacemeshos/go-spacemesh/issues/2522
+					for layerID := lastVerified.Add(1); !layerID.After(t.Last); layerID = layerID.Add(1) {
+						// TODO LANE: this operation can be very expensive and very slow after a few layers,
+						//   since no caching is currently performed. it needs to be optimized and cached.
+						if err := t.scoreBlocksByLayerID(ctx, layerID); err != nil {
+							// if we fail to process a layer, there's probably no point in trying to recore blocks
```
y0sher:
```suggestion
							// if we fail to process a layer, there's probably no point in trying to rescore blocks
```

tortoise/util.go
lrettig: not sure what you mean by "check before adding", what's the difference? you mean like turn this into an error in the code that calls Add?

tortoise/verifying_tortoise.go
lrettig: got it, I thought so, thanks. updated the comment accordingly.

lrettig:
bors try

tortoise/verifying_tortoise.go
lrettig: removed

tortoise/verifying_tortoise.go
lrettig: fixed this, thanks for highlighting

lrettig:
bors try

tortoise/verifying_tortoise.go
```diff
@@ -130,492 +242,1011 @@ func blockIDsToString(input []types.BlockID) string {
...
+					// for a layer before hare finishes for that layer, i.e., before the weak coin value is ready for
+					// the layer.
+
+					// TODO: if we rescore old blocks, it's very likely that newly-created blocks will contain
+					//   exceptions for those blocks, and their opinion will differ from their base blocks for blocks
+					//   older than the base blocks, which will cause those blocks not to be marked good. is there
+					//   anything we can do about this? e.g., explicitly pick base blocks that agree with the new
+					//   opinion, or pick older base blocks.
+					//   see https://github.com/spacemeshos/go-spacemesh/issues/2678
+					if coin, exists := t.bdp.GetCoinflip(ctx, t.Last.Sub(1)); exists {
```
y0sher, line 1070: shouldn't we get the coinflip for the current `layerID` we're checking ?

tortoise/util.go
y0sher: I was thinking preventing the overflow before it happened but I know see there'
s not much difference. resolving this.

tortoise/verifying_algorithm.go
```diff
@@ -11,52 +13,104 @@ import (
...
+	Clock           layerClock
+	Hdist           uint32 // hare lookback distance: the distance over which we use the input vector/hare results
+	Zdist           uint32 // hare result wait distance: the distance over which we're willing to wait for hare results
+	ConfidenceParam uint32 // confidence wait distance: how long we wait for global consensus to be established
+	GlobalThreshold uint8  // threshold required to finalize blocks and layers (0-100)
+	LocalThreshold  uint8  // threshold that determines whether a node votes based on local or global opinion (0-100)
+	WindowSize      uint32 // tortoise sliding window: how many layers we store data for
+	Log             log.Log
+	Recovered       bool
+	RerunInterval   time.Duration // how often to rerun from genesis
```
y0sher, line 36: `RerunInterval` is only used in `tortoise/verifying_algorithm.go` but it's still part of `turtle`. is it because we want to persist this setting?

mesh/meshdb.go
```diff
@@ -324,17 +331,36 @@ func (m *DB) ContextualValidity(id types.BlockID) (bool, error) {
...
+	return m.inputVector.Put(hash.Bytes(), bytes)
+}
+
+// InvalidateLayer receives notification from Hare that it failed for a layer. If Hare explicitly fails for a
+// layer, we consider all blocks in that layer to be invalid. Note that simply not having received a layer from Hare
+// as _validated_ is not sufficient information since we might still be waiting for Hare to finish for the layer. This
+// method lets us know that Hare has given up.
+func (m *DB) InvalidateLayer(ctx context.Context, layerID types.LayerID) {
+	m.WithContext(ctx).With().Info("recording hare invalidated layer in mesh", layerID)
+	m.invalidatedLayers[layerID] = struct{}{}
```
y0sher, line 362: this map is accessed from multiple code paths, shouldn't it be thread-safe?

y0sher (approved):
Good job, I especially liked how detailed the documentation is, it made reviewing this a breeze. I couldn't find anything major so I'm approving. please look at my comments and consider them, they are mainly suggestions/questions.

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

tortoise/util.go
lrettig: Let's continue the conversation in #2629

lrettig:
bors try

tortoise/util.go
lrettig: This code has been updated. Let's continue the conversation in #2629.

hare/hare.go
lrettig: I see your point, but I'd rather optimize for readability. That package is not imported here, so there is no conflict, at least in this module. We could probably clean this up globally.

lrettig:
bors try

tortoise/verifying_tortoise.go
lrettig: for a given base block candidate, it's essential that the diffs be added, that's how the verifying tortoise algorithm works (otherwise, the votes would be counted incorrectly). what we _could_ do instead is do a better job of picking base blocks, so as to minimize the length of the exception list. that's a to do item in #2402.

tortoise/verifying_tortoise.go
lrettig: yeah, I'm afraid that would be too expensive. this is a high priority followup item and will be addressed in #2424.

tortoise/verifying_tortoise.go
lrettig: yup, but I think it's a small price to pay for coherence in the logs throughout the code block

tortoise/verifying_tortoise.go
lrettig: also a high priority follow up item, in #2369

tortoise/verifying_tortoise.go
lrettig: this is precisely the operation where the base block votes are implicitly added to the block's vote! the base block's opinions should already have had the weight factored in here.

actually, this raises an interesting question: if the base block has weight X and a block based on it has weight Y, what weight should we use for the "implicit" opinion of X that's included in Y's vote? I'll ask Tal.

lrettig: #2687

tortoise/verifying_tortoise.go
lrettig: if we received and processed the block previously, yes. but I don't think we have an easy way of checking that here, either...

tortoise/verifying_tortoise.go
lrettig: it's also a style thing, to keep the three conditionals on the following lines aligned

tortoise/verifying_tortoise.go
lrettig: nope, we use the coinflip for the top layer we've seen. I clarified this with Tal. I'll add an explanatory comment to this effect.

tortoise/verifying_algorithm.go
lrettig: yup

mesh/meshdb.go
lrettig: it's only accessed in one place, the Hare's `outputCollectionLoop`, and there should only be one of those running at a time. do you think we still need to protect it?

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

syncer/syncer.go
```diff
@@ -268,12 +269,12 @@ func (s *Syncer) synchronize(ctx context.Context) bool {
 		close(vQueue)
 		<-vDone
 		s.setStateAfterSync(ctx, success)
-		logger.With().Info(fmt.Sprintf("finished sync run #%v", s.run),
+		logger.With().Debug(fmt.Sprintf("finished sync run #%v", s.run),
```
countvonzero: i'd rather this to be the Info and line 304 to be Debug.

countvonzero: but no fuss. i can change it after you merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

syncer/syncer.go
lrettig: done!

dshulyak:
please use logtest.New(t) in tests and not log.NewDefault

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

tortoise/verifying_tortoise.go
```diff
@@ -130,492 +242,1007 @@ func blockIDsToString(input []types.BlockID) string {
...
+	atxTimestamp, err := t.atxdb.GetAtxTimestamp(votingBlock.ATXID)
+	if err != nil {
+		return 0, err
+	}
+	atxEpoch := atxHeader.PubLayerID.GetEpoch()
+	nextEpochStart := t.clock.LayerToTime((atxEpoch + 1).FirstLayer())
+
+	// check if the ATX was received on time
+	// TODO: add an exception for sync, when we expect everything to be received late
+	if atxTimestamp.Before(nextEpochStart) {
```
dshulyak: @lrettig this appears to be the reason why the verified layer was stuck on my new devnet. it might be as well the reason for latenodes failure, or i misunderstood something?

lrettig: good point, I'm testing without this

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
Here are the logs for the latest `latenodes` systest failure (with some filters): http://34.83.138.105:5601/goto/760e4d8aa78c69d21358608525d7ffb9

Here are the first few relevant errors/warnings (with filters applied): http://34.83.138.105:5601/goto/91a1cc5a33cfa8ec699fa98c69f381d3

I'm seeing the following errors and warnings:
- `hare | error checking if identity is active | empty active set for layer 6 in epoch 2`
- `blockBuilder | failed to check for block eligibility | failed to get latest atx for node in epoch 2: failed to get atx id for target epoch 2: atx for node 08d24 with publication epoch 1: leveldb: not found`
- `hareOracle | error getting active set | empty active set for layer 6 in epoch 2`
- `blockOracle | failed to calculate eligibility proofs | failed to get latest atx for node in epoch 2: failed to get atx id for target epoch 2: atx for node 08d24 with publication epoch 1: leveldb: not found`
- `mesh | requested unknown layer hash`
- `tBeacon | failed to send proposal message | proposalPassesEligibilityThreshold: atxThreshold: zero epoch weight provided`
- `tBeacon | failed to handle proposal message | get node ATXID for epoch (miner ID 78bc0): atx for node 78bc0 with publication epoch 1: leveldb: not found`
- `tBeacon | failed to handle first voting message | verify first voting message: get node ATXID for epoch (miner ID f57a2): atx for node f57a2 with publication epoch 1: leveldb: not found`
- `hare | error processing hare tick | not starting hare since node is not synced`

It appears we still have issues syncing ATXs.

@countvonzero @nkryuchkov I could use your help troubleshooting the sync and beacon related issues here.

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

dshulyak:
bors merge

dshulyak:
weird. different unit test is failing every time, it didn't even come to latenodes in last 4 tries

bors merge

dshulyak:
bors merge

dshulyak:
i don't understand why unit tests are so flaky, there are now 3 failures in a single run:
- TestMesh_AddBlockWithTxs_PushTransactions_UpdateUnappliedTxs
- TestMesh_updateStateWithLayer
- TestFindNodeProtocol_FindNode_Concurrency
i just restarted them on develop in my branch and they passed from the first try https://github.com/spacemeshos/go-spacemesh/pull/2683

bors merge

dshulyak:
bors merge

dshulyak:
There must be something in this branch that makes those tests crash consistently on CI

lrettig:
bors merge

lrettig:
> i don't understand why unit tests are so flaky, there are now 3 failures in a single run:
> 
> * TestMesh_AddBlockWithTxs_PushTransactions_UpdateUnappliedTxs
> * TestMesh_updateStateWithLayer
> * TestFindNodeProtocol_FindNode_Concurrency
>   i just restarted them on develop in my branch and they passed from the first try [Refactor trtl.Persist to use its own store and do the task more efficiently #2683](https://github.com/spacemeshos/go-spacemesh/pull/2683)

These failures were due to a bugfix I made in the mesh code, just needed to update those tests - `TestFindNodeProtocol_FindNode_Concurrency` is flaky but the others were not a flakiness issue

mesh/reward_test.go
```diff
@@ -211,57 +211,85 @@ func TestMesh_updateStateWithLayer(t *testing.T) {
...
+	// this should be played until numOfLayers-1 if we want to compare states
+	for i := 0; i < numOfLayers-1; i++ {
+		layerID := startLayer.Add(uint32(i))
+		blockIds := copyLayer(t, mesh, mesh2, atxDB2, layerID)
+		mesh2.HandleValidatedLayer(context.TODO(), layerID, blockIds)
 	}
+
 	// test states are the same when one input is from tortoise and the other from hare
-	assert.Equal(t, s.Txs, s2.Txs)
+	require.NotEqual(t, s.Txs, s2.Txs)
```
countvonzero, line 235: @lrettig 
is this correct? if we want the same result, it should be equal right?

mesh/reward_test.go
```diff
@@ -211,57 +211,85 @@ func TestMesh_updateStateWithLayer(t *testing.T) {
...
+
+	layerIDFinal := startLayer.Add(uint32(numOfLayers - 1))
+	copyLayer(t, mesh, mesh2, atxDB2, layerIDFinal)
+	l, err := mesh.GetLayer(layerIDFinal)
+	require.NoError(t, err)
+	mesh2.ValidateLayer(context.TODO(), l)
+
+	// test states are the same when one input is from tortoise and the other from hare
+	require.Equal(t, s.Txs, s2.Txs)
```
countvonzero, line 244: why not do `for i := 0; i < numOfLayers; i++` in line 228 and `require.Equal(t, s.Txs, s2.Txs)` directly?
why care about this one last layer difference?

mesh/reward_test.go
lrettig: the comment is a little misleading (see identical comment a few lines down), but the change is intentional and this should be correct. all layers _except_ the latest layer have been applied here, so the state should not be the same. then the latest layer is applied, and then the state should be the same. that's what this is testing.

mesh/reward_test.go
lrettig: the test is more thorough if it isolates the final layer application: the states are _not_ equal before the final layer is applied, and they _are_ equal after.

PR #2401, opened 2021-04-30
[Merged by Bors] - Push metrics to remote server

valar999:
## Motivation

To get metrics from external nodes we have to add push mechanism.

Part of https://github.com/spacemeshos/pm/issues/37

## Changes

- adds flag `--metrics-push` to specify remote server url
- `metrics-push-period` specify frequency of pushing

## Test Plan

Test infrastructure is ready and receiving of metrics was tested.

## DevOps Notes


cmd/node/node.go
```diff
@@ -1072,6 +1073,12 @@ func (app *SpacemeshApp) Start(*cobra.Command, []string) {
 		metrics.StartCollectingMetrics(app.Config.MetricsPort)
```
y0sher: this code starts a regular Promethues client (server for pulling the data from the node). shouldn't we NOT start it when using push metrics?

y0sher: might need a rename to "StartMetricsServer" because apparently it doesn't actually collects the metrics but just exposes the server.

valar999: > this code starts a regular Promethues client (server for pulling the data from the node). shouldn't we NOT start it when using push metrics?

Metrics push is a separate component that is triggered by a flag. It is intended for collecting metrics from community nodes, and prometheus client collect metrics from spacecraft nodes so they can work in parallel with the prometheus client.

y0sher: I meant renaming the Pull component "MetricsServer" since this one is pushing and ain't a server.

valar999: You are right.

We have a server to pull metrics from, function that start it is called `StartCollectingMetrics`. It could be renamed to `StartMetricsServer`.

Pushing metrics starts with `StartMetricsServer`. Let's rename it to `StartPushingMetrics`.

y0sher (approved):

valar999:
bors merge

valar999:
bors merge

valar999:
bors merge

Issue #2405, opened 2021-05-03
Late blocks


lrettig:
## Description
Very similar to the issue in #2386, we've been seeing blocks arrive late sporadically in the testnet, even between managed miners. In the screengrab below you can see a delay of nearly two minutes between when a node attempts to gossip a new block, and when it's received by the block handler and finally gossiped. I can see only two possible explanations for this: either the gossip queue is so full that it's taking a very long time to process, or else the block handler is taking a very long time to initialize the block data in this routine: https://github.com/spacemeshos/go-spacemesh/blob/9d96750851538865162745ef7745aef02814a5e5/blocks/blocks.go#L82

## Environment
Please complete the following information:
 - OS: probably all
 - Node Version: v0.1.29

## Additional Resources 
![image](https://user-images.githubusercontent.com/3316532/116836243-c9da4800-ab93-11eb-948a-d8684dc4d74b.png)

lrettig:
For reference here are links to the logs in kibana. The first drills down to the block creation, the second filters on requestId to follow the thread after that.

- http://104.154.113.206:5601/app/kibana#/discover/context/572455b0-a140-11eb-ad90-9794d01888da/-mowKHkBZi5D8iFxnyis?_a=(columns:!(name,sm.L,sm.M,sm.N,sm.layer_id,sm.to,sm.errmsg,sm.block_id),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'572455b0-a140-11eb-ad90-9794d01888da',key:name,negate:!f,params:(query:miner-95),type:phrase),query:(match_phrase:(name:miner-95))),('$state':(store:appState),meta:(alias:!n,disabled:!t,index:'572455b0-a140-11eb-ad90-9794d01888da',key:sm.M,negate:!f,params:(query:'block%20created'),type:phrase),query:(match_phrase:(sm.M:'block%20created')))),predecessorCount:15,sort:!('@timestamp',desc),successorCount:5)&_g=(filters:!())
[...]

Issue #2386, opened 2021-04-19
Late hare messages


lrettig:
## Description
We've seen for months, over several generations of testnets, that Hare messages are sometimes arriving very late, 30secs+, even among managed miners running on the same cloud infrastructure. Directly or indirectly, this has caused the failure of several testnets, since it can cause Hare to fail.

We finally have enough logs to investigate what's going on. As you can see in the screengrab below, Hare is correctly enqueuing an outgoing message using `network.Broadcast` / `gossip.Broadcast`, and the message is being added to the gossip protocol message queue, but then it's not picked up by the handler for that protocol for ~30 secs. This appears to be a Hare broker issue: it probably has a large backlog of incoming messages to process.

The right way to deal with this is probably to refactor the Hare broker to prioritize outgoing messages, and/or to broadcast self-generated gossip messages immediately (#1157).

I also see late blocks on the testnet, also among managed nodes, and I suspect something similar is going on, but due to logging/ELK stack issues I haven't been able to isolate the issue as well.

## Expected Behavior
We should only see Hare messages arriving late in extreme cases, such as network connectivity issues or nodes with overwhelmed resources.

## Environment
 - OS: probably all
 - Node Version: v0.1.29 (and all prior versions)

## Additional Resources 
![image](https://user-images.githubusercontent.com/3316532/115284502-c0c09400-a11a-11eb-9d48-86d5bf29e7fc.png)

lrettig:
This is most likely due to a combination of slowness processing Hare messages due to BLS signature validation, and the queuing issue described above.

As discussed on this morning's dev call, we'll take a two-pronged approach to address this:

1. Replace BLS with ed for signatures (#2269)
2. Implement a priority queue for _incoming_ Hare messages (so that a node can prioritize outbound messages that it generated) (#2391)

lrettig:
Even with the inclusion of #2391, I'm still seeing late hare messages between managed miners in the current testnet (tn129). Here's an example just now, where an outbound hare message took ~32 secs to be broadcast:

![image](https://user-images.githubusercontent.com/3316532/117734655-e4c74080-b1c1-11eb-8276-37499b492ac9.png)

The node was quiet for most of this period, and then suddenly there was a burst of activity:

![image](https://user-images.githubusercontent.com/3316532/117740033-8902b480-b1cd-11eb-9e0d-74feaeed1a0a.png)

[...]

PR #2411, opened 2021-05-05
[Merged by Bors] - Bye bye BLS

### LINKED ISSUE #1648: Replace BLS implementation
Our BLS implementation is highly non-optimsed. Moreover, the repo is not maintained anymore.

We should try and replace it with a different implementation: herumi / filecoin / chia-eth

Algorand should have an implantation as well (maybe they use one of the above); they also have a different VRF construction, which is the unique signature ed25519.
We will probably need it in the future, and if no other BLS implementation works for us, this VRF is the usual suspect

noamnelke:
## Motivation
Closes #1648 

Possibly addresses #2405, #2386

This is a temporary fix that replaces BLS with standard ED. Research is pending about which implementation of ED we should use for mainnet.

## Changes
Replace BLS signatures for VRF with regular ED signatures.

## Test Plan
Regular unit- and system-tests. Some were adapted to this change.

## DevOps Notes

### DIFF
--- a/signing/vrfsigner.go
+++ b/signing/vrfsigner.go
@@ -8,12 +8,12 @@ import (
 
 // VRFSigner is a signer for VRF purposes
 type VRFSigner struct {
-	privKey []byte
+	privateKey []byte
 }
 
 // Sign signs a message for VRF purposes
-func (s VRFSigner) Sign(m []byte) ([]byte, error) {
-	return ed25519.Sign(s.privKey, m), nil
+func (s VRFSigner) Sign(msg []byte) []byte {
+	return ed25519.Sign(s.privateKey, msg)
 }
 
 // NewVRFSigner creates a new VRFSigner from a 32-byte seed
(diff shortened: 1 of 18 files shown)


noamnelke:
bors try

signing/vrfsigner.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	vrfPub, vrfPriv, err := ed25519.GenerateKey(bytes.NewReader(seed))
+	if err != nil {
+		return nil, nil, err
+	}
+	return &VRFSigner{privKey: vrfPriv}, vrfPub, nil
+}
+
+// VRFVerify verifies a message and signature, given a public key
+func VRFVerify(msg, sig, pub []byte) (bool, error) {
```
lrettig: Why does this method need to return an error in its signature? Does it need to conform to some interface?

lrettig (commented):
This may be the most beautiful PR I've ever seen. It makes me so happy I want to cry 😍

No need for unit tests for the new signing package?

avive:
> This may be the most beautiful PR I've ever seen. It makes me so happy I want to cry 😍
> 
> 
> 
> No need for unit tests for the new signing package?

even bors likes it!

signing/vrfsigner.go
noamnelke: Sort of, the BLS verification method returned an error and the code around it expects it. I'd remove this, but this is a temporary implementation (we need to replace the standard ED with some unique-signature variant) and while I guess the final implementation will also not need to return an error, I'm not 100% sure, so I prefer to defer this change to when we know for sure.

But now that you mention it - I'll add a TODO.

noamnelke: On second thought, I'll just change the signature now, it's easy and not worth the TODO.

noamnelke:
I originally didn't add tests because this is a temp implementation, but the tests don't need to change when we switch to the final implementation, so I added them.

noamnelke:
bors try

lrettig (approved):
let's do this :)

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

Issue #2311, opened 2021-03-04
Hare behavior if preround fails


lrettig:
Right now, if the preround fails, Hare prints an error but otherwise keeps on running. What should happen in this situation? Should Hare give up completely? Should it iterate and try the preround again? Should it just keep going?

[Per Tal](https://spacemesh.slack.com/archives/CG91F13C7/p1614870102017900), for the testnet, the entire Hare process for that layer should just give up and the node should vote for the zero pattern for that layer. (See #2298 where the latter part of this was implemented as a fix for the testnet.)

What do we want to do in general?

PR #2417, opened 2021-05-10
Var post empty set

### LINKED ISSUE #1967: Support empty set hare result
Hare-wise: should support reaching consensus on an empty set.
State-wise: update the state with the blocks from the hare, so nothing to update in that particular layer
Block-wise: vote on the blocks from the hare, so no votes for that particular layer.

sudachen:
## Motivation
Closes #1967 #2311 #2311

## Changes
Hare always finished with a result

If preround fails empty set is used for hare continue
If hare fails empty set is considered as hare result

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -230,10 +230,12 @@ func (b *Builder) buildNipstChallenge() error {
 	if err != nil {
 		return fmt.Errorf("failed to get positioning ATX: %v", err)
 	}
+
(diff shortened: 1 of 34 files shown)


sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

sudachen:
Moved to https://github.com/spacemeshos/go-spacemesh/pull/2438

PR #2419, opened 2021-05-12
Enable gc trace

antonlerner:
## Motivation
This PR allows adding GC prints to the logs which will tell how mush time we spend in GC


cmd/integration/harness.go
```diff
@@ -42,11 +43,13 @@ func newHarnessDefaultServerConfig(args []string) (*Harness, error) {
...
 	// next will be exec path value
 	execPath := args[execPathInd+1]
+
+	enableGcTrace := Contains(args, setGcTrace) != -1
 	// remove executable path label and value
 	args = append(args[:execPathInd], args[execPathInd+2:]...)
 	args = append(args, "--acquire-port=false")
 	// set servers' configuration
-	cfg, errCfg := DefaultConfig(execPath)
+	cfg, errCfg := DefaultConfig(execPath, enableGcTrace)
```
y0sher: it's probably a personal preference but I think you could do it without changing the `DefaultConfig` signature, something along these lines - 
```go
cfg, errCfg := DefaultConfig(execPath)
cfg.gcTrace = enableGcTrace
```

y0sher: I know it was a given but probably it'll be easier to understand and cleaner code if we would've done that with the `execPath` as well.

y0sher (approved):

antonlerner:
bors merge

nkryuchkov:
@y0sher @antonlerner Do we still need it?

dshulyak:
this is integrated in spacecraft using go env. but i don't think that we need that

PR #2423, opened 2021-05-18
Remove short nodeID from log lines

lrettig:
## Motivation
The main motivation for the change is that elasticsearch [wildcard queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html#wildcard-query-ex-request) don't work well with a wildcard _prefix_. Right now, our `sm.N` log field consists of `nodeID.moduleName`, and this is very problematic in expensive searches for all output from a particular module. It's also redundant since every log line also includes the full nodeID.

## Changes
- removes nodeID prefix from log name

## Test Plan
relevant tests have been updated

## TODO

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -472,10 +472,8 @@ func (app *SpacemeshApp) initServices(ctx context.Context,
 
 	app.nodeID = nodeID
 
-	name := nodeID.ShortString()
-
 	// This base logger must be debug level so that other, derived loggers are not a lower level.
-	lg := log.NewWithLevel(name, zap.NewAtomicLevelAt(zapcore.DebugLevel)).WithFields(nodeID)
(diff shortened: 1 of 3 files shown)


noamnelke:
The reason the logs are formatted this way is for local unit tests. The logs from all nodes in the app test and multi node sim are mixed together and this node id prefix is the only thing that lets us know where a log line came from.

Perhaps we can find a better solution for this? We shouldn't merge this PR until we do.

noamnelke:
If you don't want to waste time on this right now, you can perhaps just flip the order?

lrettig:
I found a much simpler way to accomplish the same thing: https://github.com/spacemeshos/go-spacemesh/pull/2427/commits/5dab96717a4905f5e9d66faeb2ec6fc4150acfd7 (in #2427)

PR #2433, opened 2021-05-26
[Merged by Bors] - p2p: use the same shutdown context for all components

### LINKED ISSUE #1313: Make sure shutdown sequence interrupts getting more peers
# Overview / Motivation

`morePeersReq` channel races with the shutdown channel, we should make sure that we'll not try getting more peers if we're already shutting down.

countvonzero:
## Motivation

Closes #1313 

## Changes

make all p2p components share the same shutdown context so they can be in sync wrt to shutdown timeing.

## Test Plan

unit tests

## TODO

## DevOps Notes

### DIFF
--- a/p2p/connectionpool/connectionpool.go
+++ b/p2p/connectionpool/connectionpool.go
@@ -37,26 +37,22 @@ type ConnectionPool struct {
 	dialWait    sync.WaitGroup
 	logger      log.Log
 
-	shutdownCtx    context.Context
-	shutdownCancel context.CancelFunc
-	shutdown       bool
+	shutdownCtx  context.Context
+	shutdownOnce sync.Once
 }
 
 // NewConnectionPool creates new ConnectionPool
-func NewConnectionPool(dialFunc DialFunc, lPub p2pcrypto.PublicKey, logger log.Log) *ConnectionPool {
-	shutdownCtx, cancel := context.WithCancel(context.Background())
+func NewConnectionPool(ctx context.Context, dialFunc DialFunc, lPub p2pcrypto.PublicKey, logger log.Log) *ConnectionPool {
 	cPool := &ConnectionPool{
-		localPub:       lPub,
-		dialFunc:       dialFunc,
-		connections:    make(map[p2pcrypto.PublicKey]net.Connection),
-		connMutex:      sync.RWMutex{},
-		pending:        make(map[p2pcrypto.PublicKey][]chan dialResult),
-		pendMutex:      sync.Mutex{},
-		dialWait:       sync.WaitGroup{},
-		logger:         logger,
-		shutdown:       false,
-		shutdownCtx:    shutdownCtx,
-		shutdownCancel: cancel,
+		localPub:    lPub,
+		dialFunc:    dialFunc,
+		connections: make(map[p2pcrypto.PublicKey]net.Connection),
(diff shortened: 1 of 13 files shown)


countvonzero:
@y0sher i accidentally changed go.sum in this PR. i will revert it in the next iteration after your first round of review.

p2p/connectionpool/connectionpool.go
```diff
@@ -84,30 +80,23 @@ func (cp *ConnectionPool) OnClosedConnection(ctx context.Context, cwe net.Connec
...
-		cp.logger.Error("shutdown was already called")
-		return
-	}
-	cp.shutdown = true
-	cp.connMutex.Unlock()
-
-	cp.dialWait.Wait()
-	// we won't handle the closing connection events for these connections since we exit the loop once the teardown is done
-	cp.closeConnections()
+	cp.shutdownOnce.Do(func() {
```
y0sher, line 95: I guess this is being called on `switch.go`.. I thought the advantage of using a context means we don't have to call this explicitly  but that will probably require a running routine or maybe calling that once when isShuttingDown goes down the `true` path, consider switching to one of those.. context is passed down so we could cancel it only once above and it will trigger shutting down all the way down every component.

p2p/gossip/protocol.go
```diff
@@ -207,7 +217,7 @@ func (p *Protocol) propagationEventLoop(ctx context.Context) {
 			}
 			metrics.PropagationQueueLen.Set(float64(len(p.propagateQ)))
 
-		case <-p.shutdown:
+		case <-p.shutdownCtx.Done():
```
y0sher, line 220: so if it was shutdown out of the select or during message processing we might never get to this path and close `pq`.

p2p/net/network.go
```diff
@@ -285,6 +285,15 @@ func (n *Net) Shutdown() {
 	}
 }
 
+func (n *Net) isShuttingDown() bool {
```
y0sher, line 288: we could probably refactor this to a general `IsShuttingDown(context)` but this is really matter of preference since this is not a lot of code to duplicate, just a lot of places, if we ever add changes to this method it's worth extracting it out.

y0sher (commented):
Looks good, please check my comments :)

p2p/connectionpool/connectionpool.go
countvonzero: that's the route (i hope) we will eventually take. however, at this time, it seems `Switch.Shutdown()` dictates the sequence of shutting down for p2p components. i don't want to make too much change in one PR. will experiment with the paradigm you described in a later PR.

p2p/gossip/protocol.go
countvonzero: moved into the select case. thanks

p2p/net/network.go
countvonzero: i was more worried about circular dep (where to put this util func, in p2p/ or in its own package.
a util package for this seems overkill. since i will continue working on this package for the Tomb package, i prefer to defer this till later changes.

y0sher: eventually will need its own package inside p2p/whatever or util in order to prevent the circular dep..

y0sher (approved):
Great job!

countvonzero:
bors merge

Issue #1868, opened 2020-03-03
Single failed message postpones other messages


barakshani:
A miner failed sending a single message to a single peer (peer seems to be disconnected). After the timeout, we see that other messages that the miner should have sent while waiting for a response on the aforementioned message were postponed and sent only after the timeout. 

Here we see the timeout at 01:10:27 (only one message seems to be in the queue)
<img width="996" alt="Screen Shot 2020-03-03 at 9 50 58 AM" src="https://user-images.githubusercontent.com/35497694/75755802-0617db80-5d38-11ea-9bc8-32302fe0b816.png">

Here other miners declare a late Hare message from the same miner (pod `q54fp` is miner `d8554`) at the same time
<img width="1051" alt="Screen Shot 2020-03-03 at 9 51 51 AM" src="https://user-images.githubusercontent.com/35497694/75755879-31022f80-5d38-11ea-85b9-c92ddee7ac26.png">

lrettig:
Might be solved by the broker described in #1911

lrettig:
This was resolved through a combination of #2435 and #2450 (not merged yet)

PR #2435, opened 2021-05-26
[Merged by Bors] - Don't hold mutex while sending on connection

lrettig:
## Motivation
Partially addresses #1868, #2386 (and maybe #2405)

## Changes
The underlying connection objects contain a mutex that's used to protect the `closed` struct member. They hold the mutex while trying to send on the socket. But the send operation may be delayed or time out, which has a ripple effect up the stack: it blocks the operation that enqueues outbound messages, which in turns blocks propagation of gossip messages (#1868) (since gossip waits for outbound messages to all peers to be enqueued before it processes the next outbound gossip message). There are various places in this stack where we could add more concurrency or queues, but ultimately this appears to be the bottleneck, and fixing it should address the gossip message bottleneck.

## Test Plan
Adds a test that makes sure that a blocked Write operation does not block Send

## TODO

## DevOps Notes


countvonzero (commented):
it seems the underlying io.Writer is not thread-safe tho. i don't know enough to be sure there will be no concurrent reads/writes to this network connection.

lrettig:
Fair point, but writing on the connection is already synchronized with the `messages` channel.

p2p/net/conn_test.go
```diff
@@ -48,8 +48,57 @@ func TestSendMessage(t *testing.T) {
...
+}
+
+func TestSendMutex(t *testing.T) {
+	netw := NewNetworkMock()
+	rwcam := NewReadWriteCloseAddresserMock()
+	rPub := p2pcrypto.NewRandomPubkey()
+	rwcam.writeWaitChan = make(chan []byte) // unbuffered channel, write is blocking
+	conn := newConnection(rwcam, netw, rPub, &networkSessionImpl{}, msgSizeLimit, time.Second, netw.logger)
+	go conn.beginEventProcessing(context.TODO())
+	msg := "hello"
```
moshababo, line 64: Perhaps add a new line below this, to distinguish the declarations from the test logic.

p2p/net/conn_test.go
```diff
@@ -48,8 +48,57 @@ func TestSendMessage(t *testing.T) {
...
+	done := make(chan struct{})
+	go func() {
+		err := conn.Send(context.TODO(), []byte(msg))
+		assert.NoError(t, err)
+		close(done)
+	}()
+	select {
+	case <-done:
+		// success
+		break
```
moshababo, line 91: Is `break` needed?

p2p/net/conn_test.go
```diff
@@ -48,8 +48,57 @@ func TestSendMessage(t *testing.T) {
...
+	rPub := p2pcrypto.NewRandomPubkey()
+	rwcam.writeWaitChan = make(chan []byte) // unbuffered channel, write is blocking
+	conn := newConnection(rwcam, netw, rPub, &networkSessionImpl{}, msgSizeLimit, time.Second, netw.logger)
+	go conn.beginEventProcessing(context.TODO())
+	msg := "hello"
+	err := conn.Send(context.TODO(), []byte(msg))
+	assert.NoError(t, err)
+
+	// Make sure SendSock received the message and tried to send it (and is blocking on Write)
+	for i := 0; i < 10; i++ {
```
moshababo: Perhaps this would be cleaner?
````go
timeout := time.After(1 * time.Second)
for rwcam.WriteCount() < 1 {
    select {
    case <-timeout:
        assert.Fail(t, "timed out waiting for SendSock to block")
    case <-time.After(100 * time.Millisecond):
    }
}
````

File: p2p/net/conn_test.go [hunk 4]
```diff
@@ -48,8 +48,57 @@ func TestSendMessage(t *testing.T) {
...
+		if rwcam.WriteCount() >= 1 {
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	if rwcam.WriteCount() < 1 {
+		assert.Fail(t, "timed out waiting for SendSock to block")
```
moshababo: "timed out waiting for SendSock to ~~block~~ call Write"?

Writer's blocking is assumed, and is used for asserting the next msg sending, but it's not tested here.

p2p/net/conn_test.go
```diff
@@ -48,8 +48,57 @@ func TestSendMessage(t *testing.T) {
...
+		assert.Fail(t, "timed out waiting for message to be sent")
+	}
+
+	// make sure we can read both messages
+	for i := 0; i < 2; i++ {
+		select {
+		case <-rwcam.writeWaitChan:
+			rcvmsg := <-rwcam.writeWaitChan
+			assert.Equal(t, rcvmsg, []byte(msg))
+		case <-time.After(2 * time.Second):
```
moshababo: Looks like you can replace this with `default`, because it's guaranteed to _not_ be selected if any other channel returns.

moshababo:
`net.Conn` seems to be thread safe according to the [documentation](https://golang.org/pkg/net/#Conn), so concurrent calls to the underlying `io.Writer` are expected to be serialized. This obviously doesn't cover the interface per se, but we need to optimize here for the concrete implementation, since this is a hot spot.
 
Also see this clarification regarding the documentation: https://github.com/golang/go/issues/27203#issuecomment-534386117

p2p/net/conn.go
```diff
@@ -276,21 +276,19 @@ func (c *FormattedConnection) SendSock(m []byte) error {
 		c.wmtx.Unlock()
 		return fmt.Errorf("connection was closed")
 	}
+	c.wmtx.Unlock()
```
moshababo: It might be that the original scope of the lock was to avoid writing to a closed socket. 
I'm not sure it was helpful, because the socket can be closed from the other side or for internal reasons which `c.closed` doesn't cover, so the write operation must handle this condition gracefully anyway.

If so, do we even need to check `c.closed` when entering? 
`SendSock` can be assumed to be a private method (and if so we should lowercase it), and called only when processing `c.messages` until `c.stopSending` returns, in a higher-level function (`SendListener`). 

If there's some benefit that i'm missing to `c.closed` besides protecting against double close and we still need to check it here, the mutex can be upgraded to RW mutex, because there's no reason for concurrent calls to `SendSock` to block each other on this code block.

moshababo:
@lrettig since the underlying `io.Writer` can still block or time out and thus concurrent send operations to block, is this alone really expected to fix the described ripple effect?

moshababo:
@lrettig since the underlying `io.Writer` can still block or time out, and calls to `sendSock` are already serialized, is this alone really expected to fix the described ripple effect?

moshababo:
@lrettig since the underlying `io.Writer` can still block or time out, and calls to `SendSock` are already serialized, is this alone really expected to fix the described ripple effect?

p2p/net/conn.go
countvonzero: the interface doc https://golang.org/src/io/io.go?s=4279:4319#L94 says
```
// The behavior of Close after the first call is undefined.
// Specific implementations may document their own behavior.
```
so i guess holding a read lock instead will be safer. otoh, how about using a atomic int to hold `c.closed`?

p2p/net/conn_test.go
y0sher: this is not accurate, check this example - 
https://play.golang.org/p/7o3gQMl-kMi

p2p/net/conn.go
y0sher: Original mutex was to serialize messages, we now have the `sendListener` `messages` queue for that. it was and still used to prevent double closing and to make sure one and only codepath reports the closing of the connection. if everything works fine I don't see anything wrong with the new approach.

y0sher:
@lrettig so I understand that contention can happen around the mutex you moved, though you should know that what you wrote above about the whole p2p stack waiting for every message to be sent to all the peers is wrong. the `Send` method on `conn` holds the mutex only for checking closed and then sends it into `messages` queue meaning that the method is returned right away unless the `messages` chan is full (current size 250 per peer).
so the change you've done might get the queue free faster, but we could also try increasing its size..

p2p/net/conn_test.go
lrettig: No, but I try to optimize for readability, not character count :)

lrettig:
> the `Send` method on `conn` holds the mutex only for checking closed and then sends it into `messages` queue meaning that the method is returned right away unless the `messages` chan is full (current size 250 per peer).

The point is that, if `SendSock` holds the mutex, then `Send` gets blocked until `SendSock` finishes, or times out - regardless of the channel capacity. That's not the right design pattern. `Send` should never block, regardless of what `SendSock` is doing. The issue here has nothing to do with the channel capacity, and increasing it won't fix the issue. See the test I added - it fails in develop (i.e., it's a regression test). It should be clear what's going on from the test.

p2p/net/conn_test.go
lrettig: right, I don't think that would work here since we need to read from the channel twice in succession, and if we don't block the second time, we'll miss the second message and fail prematurely

p2p/net/conn.go
lrettig: I renamed `sendSock` to make it a private method. I don't see anything else that needs to change here. I think the mutex should still work fine, as is, to protect against double-close. Let me know if I've missed anything.

moshababo:
> `Send` should never block, regardless of what `SendSock` is doing. The issue here has nothing to do with the channel capacity, and increasing it won't fix the issue

Right, the fixed issue has nothing to do with it, but `Send` still depends on `SendSock` to consume the channel, and as a secondary potential issue it may block if full capacity was reached. So if the side-effects of a blocking `Send` are particularly detrimental, we should increase it, but i'd rather get some analytics info first before trying to find the optimal size.

p2p/net/conn.go
moshababo: It still not clear to be why we need to check for `c.closed` when entering `sendSock`.

You've added a secondary, limited-scope protection against double close. So the first check seems to be just to prevent writing to a closed channel, but as I mentioned, it isn't helpful for that anyway. 


 
You've added a limited-scoped check for that in case we get an error and want to close, to prevent double close, 

 (and so to avoid double close), so if the initial check is for avoid writing to a closed channel,  I mentioned

moshababo: It's still unclear to me why do we need to check for `c.closed` when entering `sendSock`.

You've added a secondary, limited-scope protection against the double close. So the first check seems to be just to prevent writing to a closed channel, but as I mentioned, it isn't helpful for that anyway.

p2p/net/conn_test.go
moshababo: It might harm the readability of a Gopher ;)

p2p/net/conn_test.go
moshababo: > this is not accurate, check this example -
https://play.golang.org/p/7o3gQMl-kMi

The non-`default` case doesn't return (for 2 sec), so `default` is selected. What's not accurate?

> right, I don't think that would work here since we need to read from the channel twice in succession, and if we don't block the second time, we'll miss the second message and fail prematurely

You are entering the `for` loop after the second `Send` returned, so the channel is expected to already have the 2 batches waiting in the channel, and `default` to not be selected. There's still some place for discrepancies due to Go's non-deterministic scheduler, but I usually favor trying to eliminate magic-duration waits.

p2p/net/conn.go
moshababo: > otoh, how about using a atomic int to hold c.closed?

If you mean to use it instead of the mutex lock, It would be sufficient for the initial status check (which I don't think we need), but not for preventing double close, because the close operation is not atomic.

p2p/net/conn.go
```diff
@@ -269,28 +269,26 @@ func (c *FormattedConnection) Send(ctx context.Context, m []byte) error {
...
 		return err
 	}
-	_, err = c.w.WriteRecord(m)
-	if err != nil {
-		cerr := c.closeUnlocked()
-		c.wmtx.Unlock()
-		if cerr != ErrAlreadyClosed {
+	if _, err := c.w.WriteRecord(m); err != nil {
+		c.wmtx.Lock()
+		if err := c.closeUnlocked(); err != ErrAlreadyClosed {
```
moshababo: It will probably be cleaner to put the lock inside `c.closeUnlocked`, which isn't safe anyway without it.

lrettig:
> Right, the fixed issue has nothing to do with it, but `Send` still depends on `SendSock` to consume the channel, and as a secondary potential issue it may block if full capacity was reached. So if the side-effects of a blocking `Send` are particularly detrimental, we should increase it, but i'd rather get some analytics info first before trying to find the optimal size.

You read my mind :) It turns out that the side-effects of a blocking Send are quite detrimental. See #2450 for a fix. It'll be included in the next testnet build as well.

p2p/net/conn_test.go
lrettig: @moshababo you're right, there's no concurrency issue here, the message sends are already synchronized and they should be immediately available. will clean up.

p2p/net/conn.go
lrettig: fixed

lrettig:
Thanks for feedback. @moshababo @y0sher @countvonzero this is ready for final review.

p2p/net/conn_test.go
lrettig: Actually, there is a concurrency issue here after all. `Send` is called synchronously, but under the hood the sending mechanism is concurrent, so we need the timeout after all. See [this failure](https://github.com/spacemeshos/go-spacemesh/runs/2796067538?check_suite_focus=true).

moshababo (approved):

lrettig:
Just finished merging in latest develop and resolving merge conflicts, and this has evolved a bit. @moshababo mind taking another look?

p2p/net/conn.go
```diff
@@ -285,27 +284,25 @@ func (c *FormattedConnection) Send(ctx context.Context, m []byte) error {
...
 		return err
 	}
-	_, err = c.w.WriteRecord(m)
-	if err != nil {
-		cerr := c.closeNoWait()
-		c.wmtx.Unlock()
-		if cerr != ErrAlreadyClosed {
+
+	// not entirely clear whether the underlying IOWriter here is goroutine safe, so we serialize writes to be safe
+	// see https://github.com/spacemeshos/go-spacemesh/pull/2435#issuecomment-851039112
```
moshababo: As per the comment, shouldn't we assume that the underlying `io.Writer` is already thread-safe, and not impose the additional safety overhead?

p2p/net/conn.go
```diff
@@ -285,27 +284,25 @@ func (c *FormattedConnection) Send(ctx context.Context, m []byte) error {
...
-	if c.Closed() {
-		return ErrClosed
-	}
-	c.wmtx.Lock()
-	err := c.deadliner.SetWriteDeadline(time.Now().Add(c.deadline))
-	if err != nil {
-		c.wmtx.Unlock()
+// sendSock sends a message directly on the socket without waiting for the queue.
+func (c *FormattedConnection) sendSock(m []byte) error {
+	// no need to check if closed here, the caller just checked
```
moshababo: I found this comment a bit misleading -- we're not _not_ checking because the caller just checked. It can still get closed in the intermediate stage, before or during the write, and so the write operation must handle this condition gracefully anyway.

p2p/net/conn.go
```diff
@@ -247,7 +247,7 @@ func (c *FormattedConnection) sendListener() {
 				log.Int("queue_length", len(c.messages)))
 
 			//todo: we are hiding the error here...
```
moshababo: Can we get rid of this comment? (it was somehow handled by https://github.com/spacemeshos/go-spacemesh/pull/2554)

p2p/net/msgcon.go
```diff
@@ -161,8 +162,8 @@ func (c *MsgConnection) sendListener() {
 				log.String("requestId", m.reqID),
 				log.Int("queue_length", len(c.messages)))
 
-			//todo: we are hiding the error here...
-			if err := c.SendSock(m.payload); err != nil {
+			// TODO: do we need to propagate this error upwards or add a callback?
```
moshababo, line 182: The current design makes it a bit tricky to be implemented. We can keep this as an open question, but if so It should be similar in `conn.go`.

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
This is done and ready to be merged, but stuck on flaky tests. Assigning to @antonlerner to make sure it gets merged. Let me know if I've missed anything or if there are any questions. Thanks!

lrettig:
bors merge

lrettig:
bors merge

PR #2436, opened 2021-05-27
Grpc graceful shutdown

countvonzero:
## Motivation

allow grpc server to shutdown gracefully - allow any in-flight requests to complete and return status code Unavailable for any new requests.

## Changes

for grpc/http servers, the graceful shutdown can block for a while. launch the shutdown routine and shutdown other components before waiting for grpc/http servers to finish shutting down.

## Test Plan

make test

## TODO

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/debug"
+	gsync "sync"
 	"time"
 
 	"github.com/spacemeshos/go-spacemesh/fetch"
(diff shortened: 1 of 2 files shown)


cmd/node/node.go
```diff
@@ -820,17 +821,28 @@ func (app *SpacemeshApp) stopServices() {
...
+		}()
 	}
 
 	if app.grpcAPIService != nil {
-		log.Info("stopping grpc service")
-		// does not return any errors
-		_ = app.grpcAPIService.Close()
+		wg.Add(1)
+		go func() {
+			log.Info("stopping grpc service")
```
noamnelke, line 841: Not your fault, but let's remove this redundant log.
https://github.com/spacemeshos/go-spacemesh/blob/5d488e5d4bfdd85a0f316624b1d4679e40e8cfd0/api/grpcserver/grpc.go#L62

In the JSON server this log is a DEBUG, so either make it an INFO and remove from here as well, or make it DEBUG in the gRPC server and leave both here. Let's just be consistent.

noamnelke (commented):
There's a possible problem that I see. Maybe it's not a problem, if you thought it through, lmk.

Since we issue the shutdown commands for the JSON and gRPC servers in goroutines, there's no guarantee that the JSON server will receive the command first. So there's a possible case where the gRPC server is already down, while the JSON server is still up. What happens when we call an endpoint on the JSON server? I assume it either errors or times out. It could be fine, especially if it's a rare condition. Just want to make sure it won't cause a panic or some other undesirable outcome.

What could be even worse is if we stop other services while the gRPC server is still up. Again could be fine (not sure it is) but it's even harder to make sure of this.

Am I overthinking this? Or could it be better to leave the shutdown process synchronous, even if it takes longer? What could help the decision is to know how long it currently takes to shut down the services after the gRPC server. If it's a couple of seconds or less then I vote to keep everything synchronous.

cmd/node/node.go
```diff
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/debug"
+	gsync "sync"
```
lrettig, line 16: This is a nitpick, but I'd rather you "import as" our sync library than the standard sync lib, for readability. Call it "syncer" maybe. `gsync` sounds like something google-related!

lrettig (changes requested):
I'm curious why you chose to attack the grpc services and only the grpc services. Other services, like sync and p2p, could also hang while, e.g., waiting for a send attempt to timeout for as long as 30 secs.

I sort of agree with Noam's concerns: without strict control on the order that services are shut down, there's the possibility of getting into a weird state where service A, which depends on service B, is still running while B has been shut down.

It seems to me like the right way to attack this is to switch to using context, as we discussed - to fire off a single `CancelFunc` so that all services receive the message at the same time.

avive:
It is awesome to see grpc improvements - but this issue has SO much more impact on the end users experience than this one: #2533 - I hope it can be implemented soon in a sprint and in time for 0.2 open devnet....

countvonzero:
closing as the PR is too old.

PR #2437, opened 2021-05-27
[Merged by Bors] - Only report receiving a block once

noamnelke:
## Motivation
We've seen cases in system tests where nodes report getting several multiples of the number of blocks that are supposed to be in a layer. It appears to happen because we log receiving a block before checking if we know it already, at which point block processing stops.

## Changes
Log `got new block` only after checking for duplicates.

## Test Plan
Since this only affects logging, no new tests are needed.

## DevOps Notes


antonlerner (approved):

noamnelke:
bors merge

PR #2438, opened 2021-05-28
[Merged by Bors] - Empty Set and Failures support for Hare consensus

### LINKED ISSUE #1967: Support empty set hare result

### LINKED ISSUE #2311: Hare behavior if preround fails

### LINKED ISSUE #2404: Hare should check sync status after wake-up delay
## Description
rather than before wake-up delay. It probably doesn't make a huge difference but: (1) we may not be in sync before, but if we’re in sync after - it should be enough (assuming we trust our neighbors, which we need for trusting gossip anyway) (2) if we were synced before and lost sync- we shouldn’t participate.

Discussed with @noamnelke , @sudachen and we all agree.

This issue appears in commit hash: 9d96750851538865162745ef7745aef02814a5e5

https://github.com/spacemeshos/go-spacemesh/blob/develop/hare/hare.go#L209

sudachen:
## Motivation
Closes #1967
Closes #2311
Closes #2404

## Changes
Hare always finished with a result

If preround fails empty set is used for hare continue
If hare fails empty set is considered as hare result


sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

hare/messagevalidation.go
```diff
@@ -238,21 +238,23 @@ func (v *syntaxContextValidator) SyntacticallyValidateMessage(ctx context.Contex
...
-			log.String("msg_type", m.InnerMsg.Type.String()))
-		return false
-	}
+		if len(m.InnerMsg.Values) == 0 {
+			logger.With().Warning("syntax validation failed: set is empty",
+				log.String("sender_id", m.PubKey.ShortString()),
+				types.LayerID(m.InnerMsg.InstanceID),
+				log.String("msg_type", m.InnerMsg.Type.String()))
+			return false
+		}*/
```
noamnelke: Let's remove the commented-out code.

hare/hare.go
```diff
@@ -160,21 +160,23 @@ var ErrTooLate = errors.New("consensus process finished too late")
...
 func (h *Hare) collectOutput(ctx context.Context, output TerminationOutput) error {
-	set := output.Set()
-	blocks := make([]types.BlockID, len(set.values))
-	i := 0
-	for v := range set.values {
-		blocks[i] = v
-		i++
-	}
+	var blocks []types.BlockID
+	if output.Completed() {
```
noamnelke: This seems redundant to me.

If `output.Completed() == false` then `len(output.Set()) == 0` and then the loop runs zero times and with an empty blocks list `h.validate(blocks)` does nothing and returns `true`.

hare/hare.go
```diff
@@ -241,8 +255,9 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) {
 	// retrieve set from orphan blocks
 	blocks, err := h.msh.LayerBlockIds(h.lastLayer)
 	if err != nil {
-		logger.With().Error("no blocks for consensus", log.Err(err))
-		return
+		// we have to try achieve consensus on empty set or fail, otherwise it will be the delayed consensus process forever
+		logger.With().Error("", log.Err(err))
```
noamnelke: Let's add a message here. Maybe `no blocks found for hare, using empty set`?

hare/hare.go
```diff
@@ -254,12 +269,13 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) {
 	instID := instanceID(id)
 	c, err := h.broker.Register(ctx, instID)
 	if err != nil {
-		logger.With().Warning("could not register consensus process on broker", log.Err(err))
+		logger.With().Error("could not register consensus process on broker", log.Err(err))
+		err = errors.New("could not register consensus process on broker")
```
noamnelke: What's the point of changing the error here? We do nothing with it other than check for nilness.

hare/hare.go
```diff
@@ -278,6 +295,9 @@ var (
 // GetResult returns the hare output for the provided range.
 // Returns error iff the request for the upper is too old.
 func (h *Hare) GetResult(lid types.LayerID) ([]types.BlockID, error) {
+
+	h.With().Warning("GetResult", lid)
```
noamnelke: I assume this is a warning for debugging purposes - let's change to info/debug (feels like a debug to me, but up to you).

hare/hare.go
```diff
@@ -297,12 +317,9 @@ func (h *Hare) outputCollectionLoop(ctx context.Context) {
...
 		case out := <-h.outputChan:
-			if out.Completed() { // CP completed, collect the output
-				if err := h.collectOutput(ctx, out); err != nil {
-					h.WithContext(ctx).With().Warning("error collecting output from hare", log.Err(err))
-				}
+			if err := h.collectOutput(ctx, out); err != nil {
+				h.WithContext(ctx).With().Warning("error collecting output from hare", log.Err(err))
 			}
-
 			// either way, unregister from broker
```
noamnelke: This comment is no longer relevant.

hare/algorithm_test.go
```diff
@@ -186,13 +186,15 @@ func TestConsensusProcess_Start(t *testing.T) {
...
-	proc.s = NewDefaultEmptySet()
 	inbox, _ := broker.Register(context.TODO(), proc.ID())
 	proc.SetInbox(inbox)
-	err := proc.Start(context.TODO())
-	assert.Equal(t, "instance started with an empty set", err.Error())
+	/*
+		proc.s = NewDefaultEmptySet()
+		err := proc.Start(context.TODO())
+		assert.Equal(t, "instance started with an empty set", err.Error())
+	*/
```
noamnelke: Remove comment.

hare/messagevalidation_test.go
```diff
@@ -166,7 +168,7 @@ func TestMessageValidator_Aggregated(t *testing.T) {
 	msgs[0].Sig = tmp
 	inner := msgs[0].InnerMsg.Values
 	msgs[0].InnerMsg.Values = nil
-	r.Equal(errInnerSyntax, validator.validateAggregatedMessage(context.TODO(), agg, funcs))
+	//r.Equal(errInnerSyntax, validator.validateAggregatedMessage(context.TODO(), agg, funcs))
```
noamnelke: Remove comment.

hare/xfailure_test.go
```diff
@@ -0,0 +1,332 @@
+package hare
```
noamnelke: The name of this file is not great. It tests both failures and successes.

At first I thought you missed writing success tests as I reviewed this file last. It might be confusing for someone looking for a test in the future.

noamnelke (approved):
Approved. Had just minor comments that should be fixed before merging.

activation/activation.go
```diff
@@ -386,22 +386,32 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 		b.log.With().Info("building new atx challenge", b.currentEpoch())
 		err := b.buildNipstChallenge()
 		if err != nil {
+			b.log.With().Error("failed to build new atx challenge", log.Err(err))
 			return err
 		}
 	}
+
+	b.log.With().Info("chellenge ready")
```
lrettig:
```suggestion
	b.log.With().Info("challenge ready")
```

hare/hare.go
```diff
@@ -206,16 +209,19 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) {
...
-		return
-	}
-
 	if id.GetEpoch().IsGenesis() {
 		logger.Info("not starting hare since we are in genesis epoch")
 		return
 	}
 
+	defer func() {
+		// it must don't return without starting consensus process or mark result as fail
```
lrettig:
```suggestion
		// it must not return without starting consensus process or mark result as fail
```

hare/hare.go
```diff
@@ -241,8 +255,9 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) {
 	// retrieve set from orphan blocks
 	blocks, err := h.msh.LayerBlockIds(h.lastLayer)
 	if err != nil {
-		logger.With().Error("no blocks for consensus", log.Err(err))
-		return
+		// we have to try achieve consensus on empty set or fail, otherwise it will be the delayed consensus process forever
+		logger.With().Error("", log.Err(err))
+		//return
```
lrettig: Are you sure you don't want to return here?

hare/hare.go
```diff
@@ -241,8 +255,9 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) {
 	// retrieve set from orphan blocks
 	blocks, err := h.msh.LayerBlockIds(h.lastLayer)
 	if err != nil {
-		logger.With().Error("no blocks for consensus", log.Err(err))
-		return
+		// we have to try achieve consensus on empty set or fail, otherwise it will be the delayed consensus process forever
```
lrettig: I don't understand this comment. Why would it "otherwise delay the consensus process forever"? If hare never runs for layer N, it will still try to run for layer N+1

lrettig (dismissed):

hare/hare.go
sudachen: Not sure. But If there is no difference, why not?

hare/hare.go
sudachen: It was before I added defer finishing hare with "notCompleted" state. So it was leaving this hare run incomplete for the tortoise.

sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

AmitShaul:
bors try

sudachen:
bors try

sudachen:
bors try

sudachen:
bors try

sudachen:
bors merge

sudachen:
bors merge

PR #2439, opened 2021-05-30
Add Pyroscope profiler to CI

y0sher:
## Motivation

we want to be able to track memory and cpu using the infra structure

## Changes
- Start pyroscope server in each test
- Pass the server IP to the node
- use nodeid as application name in pyrocope

## DevOps Notes

@narayanprusty - this means in spacecraft the change will be the nodeid and not the pod-name, before merging I might include a line to use the parameter only if passed otherwise use the nodeid.
Note: This is the only way this can work in our CI since we use one deployment for all the miners meaning they must have the same spec and arguments passed to them. so we can't pass a distinct name for each miner.

@antonlerner @noamnelke @sudachen , @other team members - Q:
do you think you'd prefer miner name vs nodeid? we might decide to invest in a workaround for this ..

The Pyroscope server IP:PORT printed on the test logs and currently lives only during the test, soon we'll add an option to keep it running.

### DIFF
--- a/tests/config.py
+++ b/tests/config.py
@@ -11,6 +11,7 @@ FILEBEAT_CONF_DIR = './elk/filebeat/'
 FLUENT_BIT_CONF_DIR = './elk/fluent-bit/'
 KIBANA_CONF_DIR = './elk/kibana/'
 LOGSTASH_CONF_DIR = './elk/logstash/'
+PYROSCOPE_CONF_DIR = './elk/pyroscope/'
 
 ES_USER_LOCAL = ut.get_env("ES_USER")
 ES_PASS_LOCAL = ut.get_env("ES_PASS")
@@ -20,4 +21,5 @@ MAIN_ES_URL = f"{MAIN_ES_IP}:9200"
 
 BOOTSTRAP_PORT = 7513
 ORACLE_SERVER_PORT = 3030
+PROFILER_SERVER_PORT = 4040
 POET_SERVER_PORT = 80
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -323,10 +323,8 @@ def add_elk(init_session, request):
 @pytest.fixture(scope='module')
 def add_pyroscope(init_session, request):
     # get today's date for filebeat data index
-    index_date = datetime.utcnow().date().strftime("%Y.%m.%d")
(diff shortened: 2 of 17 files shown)


tests/config.py
```diff
@@ -11,6 +11,7 @@
 FLUENT_BIT_CONF_DIR = './elk/fluent-bit/'
 KIBANA_CONF_DIR = './elk/kibana/'
 LOGSTASH_CONF_DIR = './elk/logstash/'
+PYROSCOPE_CONF_DIR = './elk/pyroscope/'
```
AmitShaul, line 14: I don't think pyroscope belongs under the same directory as the elk

tests/conftest.py
```diff
@@ -316,6 +320,15 @@ def add_elk(init_session, request):
     pytest.delete_namespace = dump_es_to_main_server(init_session, index_date, request.session.testsfailed)
 
 
+@pytest.fixture(scope='module')
+def add_pyroscope(init_session, request):
+    # get today's date for filebeat data index
+    index_date = datetime.utcnow().date().strftime("%Y.%m.%d")
+    add_pyroscope_cluster(init_session)
+
+    # yield
```
AmitShaul: remove redundant lines

AmitShaul (approved):

tests/config.py
y0sher: I just wanted to reduce mess instead of recreating the same structure somewhere else, I think we can rename to monitoring..

y0sher:
bors merge

y0sher:
bors merge

noamnelke:
> do you think you'd prefer miner name vs nodeid? we might decide to invest in a workaround for this ..

I generally think we should phase out the use of `NodeID` as a node identifier. The name is misleading, it should have been called `MinerID` and the reason I think it should be phased out is twofold:
1. We want to support non-mining nodes, where creating a keypair is pointless.
2. In order to support the SMApp flow we must generate the mining keypair / ID only after node startup. Today we have a workaround for this that isn't great.

The SMApp flow should be:
1. launch node
2. start syncing / listening as a non-mining node
3. at some point, the user "sets up mining" and decides on a location for the PoST files
4. generate an ID / read it from disk if it exists

y0sher:
bors merge

PR #2440, opened 2021-05-31
Handle SIGTERM

countvonzero:
## Motivation

k8s `kubctl delete pod` issue a SIGTERM to the binary. we want to listen to this signal to be able to shutdown gracefully

## Changes

add SIGTERM, in addition to os.interrupt, to the set of signals the binary listens to for shutting down

## Test Plan

make test

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/debug"
+	"syscall"
 	"time"
 
 	"github.com/spacemeshos/go-spacemesh/fetch"
@@ -243,7 +244,7 @@ func (app *SpacemeshApp) introduction() {
 func (app *SpacemeshApp) Initialize(cmd *cobra.Command, args []string) (err error) {
 	// exit gracefully - e.g. with app Cleanup on sig abort (ctrl-c)
 	signalChan := make(chan os.Signal, 1)
-	signal.Notify(signalChan, os.Interrupt)
+	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
 
 	// Goroutine that listens for Ctrl ^ C command
 	// and triggers the quit app


y0sher:
wouldn't that result in k8s restarting the node anyway? usually in CI we start the node wrapped with the `harness` which can also receive signals and can make the process just hang instead of restart. 
I'm not sure this was the original intention of this PR, but to test nodes dropping in CI the harness is probably our best chance since k8s eagerly tries to keep all pods up and running..

countvonzero:
> wouldn't that result in k8s restarting the node anyway? usually in CI we start the node wrapped with the `harness` which can also receive signals and can make the process just hang instead of restart.
> I'm not sure this was the original intention of this PR, but to test nodes dropping in CI the harness is probably our best chance since k8s eagerly tries to keep all pods up and running..

the original goal is to facilitate `kubectl delete pod` so that the node has a chance of clean shutdown. as @narayanprusty shared on slack #dev channel that he simply deletes the pod and let k8s restart the pod. and k8s sends SIGTERM (https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace).
i think in this case it's okay for k8s to restart as long as it shutdown cleanly.

i don't understand "to test nodes dropping in CI". that wasn't my original intent. 
tho while adding this signal, i looked for a way to test this change and found harness a useful place. do you mean to add a go routine that listens to os.Interrupt / syscall.SIGTERM and call `server.shutdown()`?
the current code seems to block on `s.cmd.Wait()` and relies on any os interrupt/kill signal to be propagated to the child process (server) when the harness process is interrupted/killed.

cmd/node/node.go
```diff
@@ -243,7 +244,7 @@ func (app *SpacemeshApp) introduction() {
 func (app *SpacemeshApp) Initialize(cmd *cobra.Command, args []string) (err error) {
 	// exit gracefully - e.g. with app Cleanup on sig abort (ctrl-c)
 	signalChan := make(chan os.Signal, 1)
-	signal.Notify(signalChan, os.Interrupt)
+	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
```
noamnelke, line 247: Let's try to be consistent, either use `os.Interrupt, os.Kill` or `syscall.SIGINT, syscall.SIGTERM` (see [here](https://golang.org/pkg/os/#Signal)).

I personally prefer the second option as I find it less abstract and that it makes it clearer what we're doing.

noamnelke (changes requested):
@y0sher makes an important point - we should make sure the harness "delegates" signals to the node. If a SIGTERM isn't listened for by the harness I think it would still result in the process just dying - not 100% sure how it would work, though.

Do we use the harness in spacecraft? Or run the node directly in a pod?

cmd/node/node.go
nkryuchkov: @noamnelke 

Consistent values look better, however, looking from the other side, `os.Kill` differs from `syscall.SIGTERM` and `syscall.SIGTERM` [is not generated in Windows](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-160&viewFallbackFrom=vs-2017), so `os.Interrupt, syscall.SIGTERM` kind of indicates that `os.Interrupt` is intended to be caught on all systems and `syscall.SIGTERM` is needed only for systems that use it.

Also, in theory, `os.Interrupt` may be a dummy signal on an abstract system that doesn't support `SIGINT`.

I think it's a matter of taste and we can use any option.

noamnelke: Looks like you're right and it's my bad, `OS.Kill` can't be caught:
https://groups.google.com/g/golang-nuts/c/t2u-RkKbJdU

countvonzero:
closing as it's too old

PR #2441, opened 2021-05-31
Limit markdown files line length to ~80 chars

daonb:
## Motivation
It's hard to read files with long lines using `less` & `vi` 

## Changes
I haven't changed the text at all - just broke the lines close to the 80 char

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -2,45 +2,71 @@
 
 ## Our Pledge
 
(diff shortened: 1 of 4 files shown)


noamnelke (changes requested):
I feel that adding this without some linter, or other mechanism for validating line length ongoingly, is doing half the work and the situation will gradually deteriorate.

One of the following can be used in a script to check for long lines:
```shell
sed -ne '/.\{80\}/p'
awk 'length($0) > 80'
grep '.\{80\}'
```

Other comments:
1. Not sure we should make lines 80 chars, modern monitors are easily wide enough to support 120.
2. This PR leaves some longer lines, and those cases are reasonable, but would fail an automated check. Need to decide how to handle this.

PR #2442, opened 2021-06-02
[Merged by Bors] - moved teardown process to GCP using 'google cloud tasks'

AmitShaul:
## Motivation
making tests finish faster

## Changes
moved automation teardown to GCP using 'google cloud tasks'
this includes the following tasks to be executed on the cloud:
* remove node pool
* delete fluent bits clusterrole binding object
* dumping ES logs to main ES if necessary
* removing namespace

added app engine code under automation 'tests' directory

on the node side:
changed 'cached_epoch_id' from string to int where -1 replaced '(none)'


AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

AmitShaul:
bors try

blocks/blockoracle.go
```diff
@@ -70,7 +70,7 @@ func (bo *Oracle) BlockEligible(layerID types.LayerID) (types.ATXID, []types.Blo
 	epochNumber := layerID.GetEpoch()
 	var cachedEpochDescription log.Field
 	if bo.proofsEpoch == DefaultProofsEpoch {
-		cachedEpochDescription = log.String("cached_epoch_id", "(none)")
+		cachedEpochDescription = log.Int("cached_epoch_id", -1)
```
noamnelke, line 73: I don't like this change. It's kicking the can down the road instead of fixing the issue. IIUC we have many other places where we log more than one type with the same field name, but those places are edge-cases that don't happen as often. Even if we fix them all, there's nothing preventing someone from accidentally introducing it back in the future.

As I see it, by allowing Elasticsearch to fail when we violate a field's type and at the same time having no mechanism for validating that the code uses consistent types for each field name is setting ourselves up for flaky tests and hard-to-track-down errors.

We can take one of two approaches:
1. Change ES so it doesn't break when we log different field types with the same name.
2. Introduce a mechanism to prevent logging different field types with the same name. One option that comes to mind is adding an `_n` to every numeric field name and an `_s` to every string.
Analyzing the code for bad usage would be orders of magnitude harder.

cc: @lrettig @narayanprusty @y0sher

tests/k8s_handler.py
```diff
@@ -16,7 +16,8 @@ def remove_clusterrole_binding(shipper_name, crb_name):
         k8s_client.delete_cluster_role_binding(crb_name)
         print(f"\nsuccessfully deleted: {crb_name}")
     except Exception as e:
-        print(f"\n{shipper_name} cluster role binding deletion has failed, manually delete {crb_name}")
+        print(f"\n{shipper_name} cluster role binding deletion has failed, please manually delete {crb_name}:")
```
noamnelke, line 19: how polite 😄

tests/node_pool_deployer.py
```diff
@@ -6,9 +6,9 @@
 
 class NodePoolDep:
     resources = ["bootstrap", "client"]
-    gcloud_delete = 'yes | gcloud container --project="{project_name}" node-pools delete {pool_name} ' \
+    gcloud_delete = 'gcloud --quiet container --project="{project_name}" node-pools delete {pool_name} ' \
                     '--cluster={cluster_name} --zone={zone}'
-    gcloud_cmd = 'yes | gcloud beta container --project "{project_name}" node-pools create "{pool_name}" ' \
+    gcloud_cmd = 'gcloud --quiet beta container --project "{project_name}" node-pools create "{pool_name}" ' \
```
noamnelke, line 11: 👏🏻

blocks/blockoracle.go
noamnelke: Also, this change doesn't seem related to this PR...

noamnelke (changes requested):

blocks/blockoracle.go
y0sher: I like the second idea with adding a prefix/suffix with types better. 
We use consistently use certain features of Elasticsearch that would render obsolete if we ignore the types, it seems kind of abuse of the logs and a waste of all the hard work we invested in creating structured detailed logs.
IMO It shouldn't be too hard to keep our logs using consistent types and if sounds good practice to be anyway, and if fixing one type once in a while saves us flattening Elasticsearch indexing it is worth it. 
Idea: this could probably be easily found with some custom linting code in the CI.

AmitShaul:
bors try

AmitShaul:
bors try

blocks/blockoracle.go
noamnelke: I'm not opposed to the suffix idea (I suggested it, after all) but I have two comments on what you said, @y0sher.
1. What ES features do we use that depend on field types? I was under the impression that fields can be cast and anything can be done directly on strings with only minor impact on performance. If I'm wrong about this than you're possibly right to prefer not to flatten.
2. I proposed the suffix idea because I tried to think how to implement linting and found it very hard. Especially due to the `.FieldNamed()` method, which accepts a field (of any type) and changes its name. We could make it easier if we split this method into `.NumericFieldNamed()` and `.StringNamed()` and then also create two types of fields: `StringField` and `NumericField`. Not the end of the world, but not trivial either.

AmitShaul:
bors try

blocks/blockoracle.go
AmitShaul: so what do you guys suggest we do from here?
can I at least push this change for now and we attend this issue on another PR? (kicking it down the road)

noamnelke: I'll approve the PR so you can merge it. @y0sher and I can keep discussing.

noamnelke (approved):

blocks/blockoracle.go
y0sher: I agree we can push this down the road as long as there are no problematic logs we know of rn

AmitShaul:
bors try

AmitShaul:
bors merge

Issue #2385, opened 2021-04-19
Drop peers after failing to send messages to them


lrettig:
## Description
I'm seeing ~200k/hr p2p warnings about "failed sending" messages from the managed testnet nodes (see screenshot below). This happens when a node tries to propagate a gossip message to a peer, and fails. I think this error is happening so often because nodes never drop peers after such an error. Ideally, a node should drop a peer--in the case of connection not found in cpool, probably right away, and replace it with a healthy peer.

## Affected code

https://github.com/spacemeshos/go-spacemesh/blob/a3ba021480a2aa5ed9b88d9d2c1f62dbddf7c1f1/p2p/gossip/protocol.go#L145

This issue appears in commit hash: v0.1.29, a3ba021480a2aa5ed9b88d9d2c1f62dbddf7c1f1

Related files (optionally with line numbers): 

![image](https://user-images.githubusercontent.com/3316532/115275524-e4caa800-a10f-11eb-8053-817654d8e580.png)

dshulyak:
So in the codebase if writer fails https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/net/conn.go#L285-L297 it will notify switch https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/switch.go#L217-L218 which will call Disconnect. We definitely don't need to add another disconnect in the protocol, as this mechanism already exists.

But there is probably something wrong with it, the code didn't change much between https://github.com/spacemeshos/go-spacemesh/commit/a3ba021480a2aa5ed9b88d9d2c1f62dbddf7c1f1 and current develop.

dshulyak:
For reads there is similar code flow that will propagate notifications and close connection if reader failed.
One suspicious part is https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/switch.go#L687-L696

Notification is actually non-blocking so if there were a lot of disconnects (more than 30) and peer manager goroutine wasn't scheduled to run some events might get dropped. And this will result in the behaviour that you are seeing (connection in cpool doesn't exist, but peer manager is not updated properly).

dshulyak:
I experimented with the change https://github.com/spacemeshos/go-spacemesh/pull/2490 on a devnet with 50 nodes. By disconnecting multiple peers periodically. "expired peer" log was consistently observed, but never "failed sending". 

[...]

PR #2450, opened 2021-06-10
[Merged by Bors] - Drop peers when outbound message queue is full

lrettig:
## Motivation
Closely related to #2435
Partially addresses #1868, #2386 (and maybe #2405)

## Changes
When the outbound message queue for a given connection is full, rather than blocking on `Send`, begin dropping messages and printing errors (rather than allowing the blocking to propagate up the gossip stack)

It would be good to go one step further and drop peers that are hanging, but this is a quick, simple fix for the acute issue we're seeing on the testnets. See #2385.

## Test Plan
Includes a new unit test for both `conn` and `msgconn` types

## TODO

## DevOps Notes


dshulyak:
This queue is not only for broadcast, but also for direct requests (msg server and fetcher), is it the case? If so we can't simply drop messages, we must disconnect peer and return with error.

lrettig:
> This queue is not only for broadcast, but also for direct requests (msg server and fetcher), is it the case? If so we can't simply drop messages, we must disconnect peer and return with error.

This is correct. I updated this to drop peers if the outbound queue is full. Could you please take a look and see if this is what you had in mind?

dshulyak:
bors try

dshulyak (approved):
yes, thanks! 

i am wondering how it can fail. are we ever sending bursts of data over short period of time,? if we do it will naturally take some time and queue may fill up with smaller messages.

lrettig:
Thanks. Not sure why the apptests failed, I can't recreate the `TestShutdown` failure locally. It looks like it's this goroutine but I don't see anything wrong with the shutdown logic:

```
        goroutine 313099 [select]:
        github.com/spacemeshos/go-spacemesh/activation.(*Builder).waitOrStop(0xc01a486000, 0x132b6a0, 0xc02159a300, 0xc01a47cb40, 0xc01a47bee0, 0xc02159bb80)
        	/home/runner/work/go-spacemesh/go-spacemesh/activation/activation.go:283 +0xbd
        github.com/spacemeshos/go-spacemesh/activation.(*Builder).loop(0xc01a486000, 0x132b6a0, 0xc02159a300)
        	/home/runner/work/go-spacemesh/go-spacemesh/activation/activation.go:309 +0x1ff
        created by github.com/spacemeshos/go-spacemesh/activation.(*Builder).StartSmeshing.func1
        	/home/runner/work/go-spacemesh/go-spacemesh/activation/activation.go:246 +0x15f
```

lrettig:
bors try

lrettig:
> i am wondering how it can fail. are we ever sending bursts of data over short period of time,? if we do it will naturally take some time and queue may fill up with smaller messages.

Yes, definitely. This pattern is pretty common due to gossip: every node does something, such as sending a hare message or creating blocks, then gossips them at the same time, so there's a massive explosion of data across the network in a brief period. If the node has a stable connection to its peers then the queue should not fill up since it should be able to send messages pretty quickly, but we do need to parameterize this so that the queue channel is large enough to handle these bursts.

lrettig:
bors try

lrettig:
bors try

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

>

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge-

lrettig:
bors merge

lrettig:
bors merge

dshulyak:
this change clearly makes p2p systest fail more often

lrettig:
It's probably the debug logs, will try to comment those out, I've seen this issue before

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
This is done and ready to be merged, but stuck on flaky tests. Assigning to @antonlerner to make sure it gets merged. Let me know if I've missed anything or if there are any questions. Thanks!

dshulyak:
bors try

lrettig:
bors merge

PR #2452, opened 2021-06-12
[Merged by Bors] - Require golang >= 1.15

lrettig:
## Motivation
Golang 1.15 has some [nice features](https://golang.org/doc/go1.15) such as `testing.TempDir`.

## Changes
This PR updates CI and scripts to require go >= 1.15.

## Test Plan
N/A

## TODO

## DevOps Notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
 env:
-  go-version: '1.14.6'
+  go-version: '1.15.13'
   GCLOUD_KEY: ${{ secrets.GCLOUD_KEY }}
   PROJECT_NAME: ${{ secrets.PROJECT_NAME }}
   CLUSTER_NAME: ${{ secrets.CLUSTER_NAME }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Inspired by https://container-solutions.com/faster-builds-in-docker-with-go-1-11/
 # Base build image
-FROM golang:1.14.4-alpine3.12 AS build_base
+FROM golang:1.15.13-alpine AS build_base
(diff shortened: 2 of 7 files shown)


dshulyak:
@lrettig any reason not to go for 1.16?

noamnelke:
> any reason not to go for 1.16?

In general, we want to support older versions, as long as we don't use features from the newer version. This is not to say that you can't use newer versions, or that we shouldn't support those, too. That's why #2475 is a good idea.

noamnelke (approved):

dshulyak:
> In general, we want to support older versions, as long as we don't use features from the newer version. This is not to say that you can't use newer versions, or that we shouldn't support those, too. That's why #2475 is a good idea.

I don't understand why, to be honest. It is not like it is hard to update golang locally. And there are always improvements/fixes under the hood, even if those are not new APIs. For instance, this is a very useful change for monitoring (among other changes in [runtime](https://golang.org/doc/go1.16#runtime)):

```
On Linux, the runtime now defaults to releasing memory to the operating system promptly (using MADV_DONTNEED), rather than lazily when the operating system is under memory pressure (using MADV_FREE). This means process-level memory statistics like RSS will more accurately reflect the amount of physical memory being used by Go processes. Systems that are currently using GODEBUG=madvdontneed=1 to improve memory monitoring behavior no longer need to set this environment variable. 
```

noamnelke:
@dshulyak the improvement you mention really seems useful and we should therefore upgrade to Go 1.16 in our testnet infra, but it doesn't depend on us requiring it... If the code works on 1.15, that should be the minimum.

While we don't see that many contributions from the community, we're still an open source project and we don't want to force developers to upgrade their Go for no reason, other than "a newer version exists".

lrettig:
bors merge

PR #2457, opened 2021-06-16
[Merged by Bors] - Introduce Hare-round clock

noamnelke:
## Motivation
Round timing in the Hare was based on timers and tickers integrated in the flow. This has two main downsides:
1. Round timing depends on the time it takes to run the code before the timer is set (this is mitigated by running most code in goroutines, reducing the time and variability of time before setting timers, but in environments with tight resources this could still be an issue).
2. It's practically impossible to mock/patch the timing mechanism for accelerated or tightly controlled tests.

## Changes
This PR introduces a Hare-round clock that calculates when a round ends based on the time the layer starts and then sets precise timers internally based on those calculations. Each "consensus process" (Hare for a particular layer) has its own clock instance which is generated by a clock factory that can be easily mocked and patched in test code to achieve desired behaviors.

## Test Plan
- New unit tests for the new clock.
- Hare unit tests now take advantage of the ability to patch the clock.
- Existing Hare and general system tests.

## TODO

## DevOps Notes


noamnelke:
bors try

noamnelke:
bors try

noamnelke:
bors try

hare/hare.go
```diff
@@ -328,10 +351,18 @@ func (h *Hare) outputCollectionLoop(ctx context.Context) {
...
 // listens to new layers.
 func (h *Hare) tickLoop(ctx context.Context) {
-	for {
+	for layer := h.layerClock.GetCurrentLayer(); ; layer++ {
 		select {
-		case layer := <-h.beginLayer:
-			go h.onTick(ctx, layer)
+		case <-h.layerClock.AwaitLayer(layer):
+			if h.layerClock.LayerToTime(layer).Sub(time.Now()) > (time.Duration(h.config.WakeupDelta) * time.Second) {
+				h.With().Warning("missed hare window, skipping layer", layer)
```
antonlerner, line 404: this should be an error - it is very critical

noamnelke: Under normal circumstances this can happen to regular users. For instance, if your computer sleeps and wakes up again - this warning will be printed for every layer that the node missed.

I think you're imagining the node running in a perfect environment (like a cloud node) and you're thinking that if the clock didn't tick on time then something terrible must have happened - but as I see it, this is something the node should be able to handle and recover from and it's only a problem when it happens unexpectedly. This is exactly what warnings are for, IMO: giving you a heads up when you try to investigate something, but communicating that it's not necessarily an issue.

---

Having said that, if you feel that this being a warning will make your life harder when investigating - I'll change it. I think that at some point we must decide (as a team) on a coherent strategy for what logs should belong to each level and then change the existing logs to match that (not a small task).

noamnelke:
bors try

sudachen:
bors try

noamnelke:
bors try

hare/haretypes.go
```diff
@@ -112,16 +117,25 @@ func (s *Set) Clone() *Set {
...
 func (s *Set) Add(id types.BlockID) {
+	s.m.RLock()
 	if _, exist := s.values[id]; exist {
+		s.m.RUnlock()
 		return
 	}
+	s.m.RUnlock()
+
+	s.m.Lock()
+	defer s.m.Unlock()
```
countvonzero: is this really thread-safe? upgrading the lock doesn't guarantee thread-safety. why not simply obtain a writer lock from the beginning?

hare/haretypes.go
```diff
@@ -130,16 +144,25 @@ func (s *Set) Add(id types.BlockID) {
...
 // It has no effect if the value doesn't exist in the set.
 func (s *Set) Remove(id types.BlockID) {
+	s.m.RLock()
 	if _, exist := s.values[id]; !exist {
+		s.m.RUnlock()
 		return
 	}
+	s.m.RUnlock()
+
+	s.m.Lock()
```
countvonzero: ditto

sudachen:
bors try

hare/haretypes.go
noamnelke: While this doesn't guarantee an atomic upgrade (not sure there's a simple way to do that) I think it's fine.

If the value was in the set when the method was called - we'll avoid obtaining an expensive write lock.
If the value wasn't there, we'll (non-atomically) upgrade the lock. This means that in the time between unlocking and locking the value may have been introduced to the set. Since adding a value to the set is idempotent - this will have no functional effect, other that unnecessarily obtaining the write lock (which we would anyway in your solution).

hare/haretypes.go
noamnelke: `delete()` is also idempotent, so the same applies here. We're optimistically saving ourselves the need to obtain a write lock. In the worst case we'll obtain it twice, but usually not.

hare/hare_test.go
```diff
@@ -150,39 +149,60 @@ func TestHare_GetResult2(t *testing.T) {
...
+	_, ok = h.outputs[mockid]
+	require.True(t, ok, "failure to validate should only log an error and succeed")
+}
+
 func TestHare_collectOutput(t *testing.T) {
 	sim := service.NewSimulator()
 	n1 := sim.NewNode()
 
-	h := createHare(n1, logtest.New(t).WithName(t.Name()))
+	h := createHare(n1, newMockClock(), log.NewDefault(t.Name()))
```
dshulyak: logtest

hare/hare_test.go
```diff
@@ -150,39 +149,60 @@ func TestHare_GetResult2(t *testing.T) {
...
 
 	_, err = h.GetResult(types.LayerID{})
 	require.Equal(t, err, errTooOld)
 }
 
+func TestHare_collectOutputCheckValidation(t *testing.T) {
+	sim := service.NewSimulator()
+	n1 := sim.NewNode()
+
+	h := createHare(n1, newMockClock(), log.NewDefault(t.Name()))
```
dshulyak: logtest

hare/haretypes.go
dshulyak: isn't access to the Set is serialized using a channel for each separate hare instance?

noamnelke: hmmm.... I think you're right.

I was getting concurrency issues in the tests, so now I'm thinking the problem may have been in the tests rather than the code...

@sudachen - can you try to remove the lock from `Set` and see what breaks?

countvonzero: @noamnelke apart from whether there's a need to add a lock here, i think i'll argue how "expensive" obtaining a writer lock is.
getting a writer lock is only expensive in the sense that it blocks other reads, which i don't see critical for this struct.
CPU-wise, it's cheaper to obtain a writer lock instead of a reader lock and optionally writer lock each time.
but the extra CPU cost is minuscule compared to the added cost of code maintenance.
in my view, simplicity out-weighs minor performance gain, which is arguable in this case.

noamnelke: Given that there shouldn't be any concurrent use of this - I can't argue with what you're saying. What I meant about an "expensive" write lock was exactly what you said - it means nobody can read or write while it's held (while a read lock allows others to read in parallel).

I also agree about maintainability being more important than performance (to an extent) but I disagree that upgrading the lock adds significant complexity.

Anywayyyyy... I hope we can just remove this synchronization logic and then everyone will agree and be happy!

noamnelke:
bors try

noamnelke:
bors try

hare/haretypes.go
noamnelke: fyi - I removed the lock and nothing broke. All's well that ends well.

hare/algorithm_test.go
```diff
@@ -207,7 +212,7 @@ func TestConsensusProcess_handleMessage(t *testing.T) {
 	r := require.New(t)
 	net := &mockP2p{}
 	broker := buildBroker(t, net, t.Name())
-	r.NoError(broker.Start(context.TODO()))
+	broker.Start(context.TODO())
```
countvonzero: restore `r.NoError` check

hare/algorithm_test.go
```diff
@@ -503,8 +507,16 @@ func TestConsensusProcess_onEarlyMessage(t *testing.T) {
...
-	r.Eventually(func() bool { return len(proc.pending) == 0 },
-		time.Second, 100*time.Millisecond, "expected proc.pending to have zero length")
+	timeout := time.NewTimer(1 * time.Second)
+	tk := time.NewTicker(100 * time.Microsecond)
+	select {
+	case <-timeout.C:
+		r.Len(proc.pending, 0)
+	case <-tk.C:
+		if 0 == len(proc.pending) {
+			return
```
countvonzero: i think `break` is more future prove just in case someone added more tests after the select block

noamnelke:
bors try

countvonzero (approved):
merge looks correct

hare/algorithm_test.go
noamnelke: you didn't look at the latest commit - I already did ☺️

hare/algorithm_test.go
noamnelke: This is very broken code that I don't think I wrote - I think it's another weird merge artifact. I'm glad you drew my attention to it.

Here's what happens: the `select` is not in a loop or anything, so the `tk` ticker will always tick before the timeout (100 microseconds < 1 second). It will then check if `proc.pending` is empty. If it is - it returns, otherwise the function ends and nothing happens. So this entire section of code basically just waits 100 microseconds.

Fixing now...

noamnelke:
bors try

noamnelke:
bors merge

PR #2461, opened 2021-06-17
use last ticked layer to calculate ticker.timeSinceLastTick()

countvonzero:
## Motivation
ticker.timeSinceLastTick() currently only returns the diff between time.now() and start of the current layer
the purpose of this utility function is to figure out whether our last ticked has be delayed by ticker.sendTickThreshold.
using ticker.lastTickedLayer to yield the current delay since last tick.

## Changes
use ticker.lastTickedLayer instead of converting time.Now() to layer number to calculate ticker.timeSinceLastTick() 

## Test Plan
added a test that failed before the change and passed after the change.

## DevOps Notes

### DIFF
--- a/timesync/ticker_test.go
+++ b/timesync/ticker_test.go
@@ -1,11 +1,12 @@
 package timesync
 
 import (
-	"github.com/spacemeshos/go-spacemesh/common/types"
-	"github.com/stretchr/testify/require"
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/stretchr/testify/require"
 )
 
 var conv = LayerConv{
@@ -75,7 +76,7 @@ func TestTicker_StartNotifying(t *testing.T) {
 	r.True(tr.started)
 	_, e = tr.Notify()
 	r.NoError(e)
-	r.Equal(3, mc.countToLayer)
+	r.Equal(2, mc.countToLayer)
 }
 
 func TestTicker_Notify(t *testing.T) {
@@ -143,7 +144,12 @@ func TestTicker_timeSinceLastTick(t *testing.T) {
 		duration: 100 * time.Millisecond,
 		genesis:  c.Now().Add(-320 * time.Millisecond),
 	})
-	r.True(tr.timeSinceLastTick() >= 20)
+	// 20ms is the diff between time.now() and start of the current layer i.e. t.LayerToTime(currentLayer)
+	sinceStartOfLayer := 20 * time.Millisecond
+	r.True(tr.timeSinceLastTick() == sinceStartOfLayer)
+	c.advance(sendTickThreshold)
+	r.True(tr.timeSinceLastTick() == sinceStartOfLayer+sendTickThreshold)
+
 }
 
 func TestTicker_AwaitLayer(t *testing.T) {
--- a/timesync/ticker.go
+++ b/timesync/ticker.go
@@ -145,7 +145,7 @@ func (t *Ticker) Notify() (int, error) {
 // TimeSinceLastTick returns the duration passed since the last layer that we ticked
 // note: the call is not lock-protected
 func (t *Ticker) timeSinceLastTick() time.Duration {
-	timeOfLastTick := t.LayerToTime(t.TimeToLayer(t.clock.Now()))
+	timeOfLastTick := t.LayerToTime(t.lastTickedLayer)
 	return t.clock.Now().Sub(timeOfLastTick)
 }
 


timesync/ticker_test.go
```diff
@@ -75,7 +76,7 @@ func TestTicker_StartNotifying(t *testing.T) {
 	r.True(tr.started)
 	_, e = tr.Notify()
 	r.NoError(e)
-	r.Equal(3, mc.countToLayer)
+	r.Equal(2, mc.countToLayer)
```
countvonzero, line 79: note that this is a count of # of calls to converter.TimeToLayer(). since there is one less call from timeSinceLastTick(), hence 1 less count.

timesync/ticker_test.go
```diff
@@ -143,7 +144,12 @@ func TestTicker_timeSinceLastTick(t *testing.T) {
 		duration: 100 * time.Millisecond,
 		genesis:  c.Now().Add(-320 * time.Millisecond),
 	})
-	r.True(tr.timeSinceLastTick() >= 20)
+	// 20ms is the diff between time.now() and start of the current layer i.e. t.LayerToTime(currentLayer)
+	sinceStartOfLayer := 20 * time.Millisecond
+	r.True(tr.timeSinceLastTick() == sinceStartOfLayer)
```
noamnelke, line 149:
```suggestion
	r.Equal(tr.timeSinceLastTick(), sinceStartOfLayer)
```

Will produce nicer error messages.

timesync/ticker_test.go
```diff
@@ -143,7 +144,12 @@ func TestTicker_timeSinceLastTick(t *testing.T) {
...
+	c.advance(sendTickThreshold)
+	r.True(tr.timeSinceLastTick() == sinceStartOfLayer+sendTickThreshold)
```
noamnelke, line 151:
```suggestion
	r.Equal(tr.timeSinceLastTick(), sinceStartOfLayer+sendTickThreshold)
```

noamnelke (approved):

countvonzero:
i am having second thought about this PR.
the comment on sendTickThreshold seems to think that it really did want to calculate the time diff between now and the start of the current layer. 
```
// the limit on how late a notify can be
// an attempt to notify later than sendTickThreshold from the expected tick time will resulted in a missed tick error
const sendTickThreshold = 500 * time.Millisecond
```
the failure in app_test.go is real. this change stops notifying each tick due to errMissedTickTime. i don't understand the purpose of this threshold. why do we stop notifying in this case? i'd argue we should get rid of this error.... 

if it really is intended to calculate the delay since the start of the current layer, it does need a better method name than timeSinceLastTick()

noamnelke:
I think your doubts are justified, and I was confused earlier. I just looked at the only place in the code where this function is used:
https://github.com/spacemeshos/go-spacemesh/blob/37ea7f6ae6b1208b63a3b9808a9ea907e6d22a71/timesync/ticker.go#L107-L113

What this method is used for is to tell, _before the clock initiates a tick_, if it's already too late, since too much time has passed since the time it was supposed to be sent. After your change, instead of measuring the time passed from when the tick was supposed to be sent, we're measuring the time since the last tick that was successfully sent (the previous one, not the current one) - so it's always after the threshold.

The behaviour before your change was correct, but I agree that the function name is misleading. I would call it `timeSinceLayerStart()`, wdyt?

noamnelke:
BTW, it's worth noting why this situation would even arise. Go cannot guarantee waking a goroutine that awaits a timer/ticker at exactly the requested time. It can and does only guarantee waking it _after_ the requested time. If there's pressure, the delay can be considerable, especially on machines/VMs with few cores. This protection is there to allow users of the clock to assume that if they receive a tick - it's within a certain time range of the given layer's tick time OR the layer is skipped altogether.

I believe that it's safer and cleaner to pass this responsibility to the caller for several reasons:
1. Each caller may have a different threshold they care about. Some examples:
  a. Blocks must be published within a certain time after the layer starts.
  b. Hare starts some cache filling at layer start, but only really cares when the "wakeup delta" passes.
  c. ATX builder doesn't care at all - it just needs to know the epoch started.
2. It should be a Spacemesh developer's habit and responsibility to validate the time at critical points in the code. Even if tick time is guaranteed, Spacemesh has many processes happening in goroutines that may have started later than expected due to load and it should be common to check for timeouts before performing actions like publishing a Hare message or block or when "locking in" a layer's block IDs for voting in the Hare.

countvonzero:
thanks for the discussion @noamnelke 
closing PR as it's not in any way a priority for now.

PR #2463, opened 2021-06-18
Parallelize processing of incoming gossip messages

lrettig:
## Motivation
Fixes an issue spotted in tn132: processing an inbound block may take a very long time (> 1m), during which time all other inbound blocks (messages on the same gossip channel) are blocking

## Changes
- Adds concurrency to processing of inbound gossip messages (per protocol/channel), with a limit on the degree of concurrency
- Add timeout to syncer layer fetch

## Test Plan
Includes regression tests

## TODO

## DevOps Notes


p2p/service/gossip_listener_test.go
```diff
@@ -112,8 +118,54 @@ func Test_AddListener_notSynced(t *testing.T) {
...
+	l := NewListener(n1, &syncMock{true, listenFn}, func() bool { return true }, conf, log.NewDefault(n1.Info.ID.String()))
+	defer l.Stop()
+
+	var channelCount int32
 
-	l.Stop()
+	releaseChan := make(chan struct{})
+	handlerFn := func(ctx context.Context, data GossipMessage, syncer Fetcher) {
+		<-releaseChan
+		atomic.AddInt32(&channelCount, 1)
```
countvonzero: is channelCount needed here?

p2p/service/gossip_listener_test.go
```diff
@@ -112,8 +118,54 @@ func Test_AddListener_notSynced(t *testing.T) {
...
+		}
+	}
+	assert.Eventually(t, checkVal(2), time.Second, 10*time.Millisecond)
+
+	// release one handler
+	releaseChan <- struct{}{}
+	assert.Eventually(t, checkVal(3), time.Second, 10*time.Millisecond)
+	releaseChan <- struct{}{}
+	assert.Eventually(t, checkVal(4), time.Second, 10*time.Millisecond)
+	releaseChan <- struct{}{}
```
countvonzero, line 144: curious what the significance of adding data to releaseChan manually and check that listenCount doesn't increase?

p2p/service/gossip_listener.go
```diff
@@ -78,8 +78,41 @@ func (l *Listener) Stop() {
...
+		}
+		<-tokenChan
+
+		l.WithContext(ctx).With().Info("got gossip message, forwarding to data handler",
+			log.String("protocol", channel),
+			log.Int("queue_length", len(gossipChannel)))
+		if !l.syncer.ListenToGossip() {
+			// not accepting data
+			l.WithContext(ctx).Info("not currently listening to gossip, dropping message")
+			return
```
countvonzero, line 101: this looks like token leaking. syncer.ListenToGossip() could become true in the next layer.
i think this check is redundant. it has been checked right before you call handleMsg()

countvonzero (changes requested):
nice cleanup

p2p/service/gossip_listener.go
```diff
@@ -78,8 +78,41 @@ func (l *Listener) Stop() {
...
+		}
+		go func() {
```
antonlerner, line 103: this design will cause many gouroutines to be created and waiting on `tokenChan`. I wonder if this is the desired behaviour? another option is to block listenToGossip until we have a worker ready

p2p/service/gossip_listener.go
lrettig: it's being checked again here because we may have blocked while waiting for an available token. what do you mean by "token leaking"?

p2p/service/gossip_listener.go
lrettig: agree, I updated this along the lines you suggested

p2p/service/gossip_listener_test.go
lrettig: listenCount is a running count of how many listeners ran. we queue up several and given a concurrency factor of 2, we initially expect this to be set to 2. then we release them one by one and expect this to tick up to 4 and stay there. does that make sense?

p2p/service/gossip_listener.go
```diff
@@ -67,6 +70,46 @@ func (l *Listener) Stop() {
 
 func (l *Listener) listenToGossip(ctx context.Context, dataHandler GossipDataHandler, gossipChannel chan GossipMessage, stop chan struct{}, channel string) {
 	l.WithContext(ctx).With().Info("start listening to gossip", log.String("protocol", channel))
+
+	// fill the channel with tokens to limit number of concurrent routines
+	tokenChan := make(chan struct{}, l.config.MaxGossipRoutines)
+	for i := 0; i < l.config.MaxGossipRoutines; i++ {
+		tokenChan <- struct{}{}
```
dshulyak: if you reverse the logic there will be no need for filling tokenChan

p2p/service/gossip_listener.go
```diff
@@ -67,6 +70,46 @@ func (l *Listener) Stop() {
...
+	}
+
+	waitForGossipToken := func(ctx context.Context) {
+		// get a token to create a new channel
+		l.WithContext(ctx).With().Debug("waiting for available slot for gossip handler",
+			log.Int("available_slots", len(tokenChan)),
+			log.Int("total_slots", cap(tokenChan)))
+
+		if len(tokenChan) == 0 {
+			l.WithContext(ctx).Error("no available slots for gossip handler, blocking")
```
dshulyak, line 85: do we need to notify user about this error? if not this is a warning

p2p/config/config.go
```diff
@@ -20,6 +20,8 @@ const (
 	defaultTCPPort = 7513
 	// defaultTCPInterface is the inet interface that P2P listens on by default
 	defaultTCPInterface = "0.0.0.0"
+	// max number of parallel gossip goroutines per channel (protocol)
+	defaultMaxGossipRoutines = 50
```
dshulyak, line 25: a default needs to be based on what the task is doing. if this is a cpu task the number of goroutines shouldn't be higher than runtime.NumCPU, if writing to the database it probably should be runtime.NumCPU as well since we are not flushing to the disk currently. otherwise it will introduce inefficiencies with scheduling

p2p/config/config.go
```diff
@@ -56,6 +58,7 @@ type Config struct {
 	SwarmConfig           SwarmConfig   `mapstructure:"swarm"`
 	BufferSize            int           `mapstructure:"buffer-size"`
 	MsgSizeLimit          int           `mapstructure:"msg-size-limit"` // in bytes
+	MaxGossipRoutines     int           `mapstructure:"max-gossip-routines"`
```
dshulyak, line 62: user shouldn't be responsible/able to configure stuff like that

p2p/server/msgserver.go
```diff
@@ -147,15 +147,17 @@ func NewMsgServer(ctx context.Context, network Service, name string, requestLife
 		cancel:             cancel,
 		workerLimiter:      make(chan struct{}, runtime.NumCPU()),
 	}
-	p.tg.Go(p.readLoop)
+	if err := p.tg.Go(p.readLoop); err != nil {
+		logger.WithContext(ctx).With().Error("error creating message server read loop", log.Err(err))
```
dshulyak: should be panic then

cmd/node/multi_node.go
```diff
@@ -169,6 +169,9 @@ func getTestDefaultConfig(numOfInstances int) *config.Config {
 	cfg.LAYERS.RequestTimeout = 10
 	cfg.GoldenATXID = "0x5678"
 
+	// increase from 2s to 10s: for some reason, the multi node test requires a long timeout
+	cfg.SyncRequestTimeout = 10000
```
countvonzero: i am unable to see where this variable is used in the codebase. is this still relevant?

syncer/syncer.go
```diff
@@ -366,33 +367,37 @@ func (s *Syncer) getLayerFromPeers(ctx context.Context, layerID types.LayerID) (
...
+	select {
+	case res := <-bch:
+		if res.Err != nil {
+			if res.Err == layerfetcher.ErrZeroLayer {
+				return types.NewLayer(layerID), nil
+			}
+			return nil, fmt.Errorf("getLayerFromPeers: %w", res.Err)
 		}
-		return nil, fmt.Errorf("PollLayerBlocks: %w", res.Err)
+	case <-time.After(s.fetcher.GetTimeout()):
```
countvonzero: i don't understand why we need this. fetcher/msgserver already has timeout mechanism. does it make sense to adjust the value there instead?

p2p/service/gossip_listener.go
countvonzero: i must have misread the code. my bad.

p2p/service/gossip_listener.go
lrettig: maybe eventually this could be reduced to a warning, but for now, if this is happening in the devnet/testnet, we want to know about it right away since it will have ripple effects and may cause problems for downstream processes such as tortoise/hare

p2p/config/config.go
lrettig: it's hard to say. handling a gossip message can do many things, and the message may be part of one of several protocols: tortoise beacon, hare, block, etc. what's more, the arrival of a gossip message can trigger a bunch of other things: database reads and writes, fetches of other data, etc. I don't think determining whether this is, say, CPU or IO bound, is that straightforward.

if we wanted to be really proper here, we'd use a different limit per protocol. but for now I think we can pick a reasonable number and research this further in future.

lrettig: let's follow up in #2708

p2p/config/config.go
lrettig: most users won't touch these config values (which is why we need a reasonable default), but the config is useful so we can test different values in, e.g., different devnets/test networks or other test environments

cmd/node/multi_node.go
lrettig: good catch! I cleaned this up. Its usage may have been removed recently.

syncer/syncer.go
lrettig: I don't see a timeout mechanism. I traced this down several levels:

LayerFetcher.PollLayerBlocks: blocks on MsgServer.SendRequest
MsgServer.SendRequest: blocks on Switch.sendWrappedMessage
Switch.sendWrappedMessage: blocks on Switch.sendMessageImpl
Switch.sendMessageImpl: blocks on conn.Send
FormattedConnection.Send: blocks on writing data to a channel

What am I missing?

countvonzero: https://github.com/spacemeshos/go-spacemesh/blob/07f9602c2f06094180a84962722116ba486c9cd6/p2p/server/msgserver.go#L216

countvonzero: @lrettig

lrettig: That's a callback, but it doesn't change the stack I described above: the request itself still blocks. It's possible that the error callback gets called while the request goroutine is still blocking. This should be cleaned up - to be clear, what's being done in this PR is just a messy workaround.

lrettig: In other words, `MessageServer.cleanStaleMessages` does not unblock the request goroutine

countvonzero: after RequestTimeout elapsed, the end result is that  an error will be delivered to `bch` channel and unblock.
the code in layerfetcher uses the same handler for good data and error, and the responses got aggregated and sent to `bch` channel. the new timeout code just added another level of timeout.

lrettig: I see where the error callback gets called. But I don't see where the original request gets unblocked. Could you explain this part to me?

countvonzero: https://github.com/spacemeshos/go-spacemesh/blob/389ab13e4270364f1a892f90f003ca20da92ab9e/layerfetcher/layers.go#L416
is the error callback for this block IDs request `SendRequest()`

https://github.com/spacemeshos/go-spacemesh/blob/389ab13e4270364f1a892f90f003ca20da92ab9e/layerfetcher/layers.go#L485
is the gorountine launched after all expected peers responded and all blocks fetched.

https://github.com/spacemeshos/go-spacemesh/blob/389ab13e4270364f1a892f90f003ca20da92ab9e/layerfetcher/layers.go#L521
is where all the `bch`s for the same layer hash are unblocked.

countvonzero: `:s/for the same layer hash/for the same layer ID/` in my last sentence

lrettig: I don't think we're on the same page here. These are two different levels of abstraction. You're describing what happens once the callbacks are called inside the message server. I'm pointing out that the message server itself may block on a stuck connection, and may never call those callbacks. Maybe it would be easier to look over the code together on a call?

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

lrettig:
bors try

syncer/syncer.go
countvonzero: if the msg server itself block on stuck connection, why is the timeout mechanism placed here and not in msg server?

does that mean all clients of msg server need to do its timeout handling? the design is weird that the caller already provide a timeout hander when calling SendRequest yet have to handle the timeout anyway?

can you point me to the line of code where the msg server connection may get stuck?

lrettig:
bors try

syncer/syncer.go
lrettig: As I mentioned in https://github.com/spacemeshos/go-spacemesh/pull/2463/files#r694103854, it's inside MsgConnection/FormattedConnection. The `Send` method of both of these queues up outbound messages into a channel:

https://github.com/spacemeshos/go-spacemesh/blob/e213df4eb0b87d7c41cd38ba9e622c812d7050c6/p2p/net/conn.go#L281

This will block if the channel is full. The channel is consumed inside `sendListener()`, which passes the outbound messages to `SendSock()`, which blocks until the message is successfully written to the wire:

https://github.com/spacemeshos/go-spacemesh/blob/e213df4eb0b87d7c41cd38ba9e622c812d7050c6/p2p/net/conn.go#L299

This has a timeout, but it's set pretty high, 60 sec by default, and that's per message, so if a connection is totally stuck, then everything will get backed up and block. This is what we saw happen on the old testnet.

#2450 is also part of addressing this, by dropping stuck peers to prevent these things from blocking.

I see your point about weird design, let me take another look and see if there isn't another option.

lrettig: I took a closer look and I think that, with #2450, this timeout is no longer necessary since this call can no longer block. I've removed it. Thanks!

lrettig:
bors try

countvonzero (approved):
thanks for helping me understand

lrettig:
The `latenodes` system test does appear to be consistently failing here. Not sure whether it's due to bad luck and flakiness, or due to a change in this PR. Let's try again.

bors merge

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

p2p/server/msgserver.go
```diff
@@ -162,34 +164,40 @@ func (p *MessageServer) Close() {
...
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-timer.C:
-			p.tg.Go(func(ctx context.Context) error {
+			if err := p.tg.Go(func(ctx context.Context) error {
 				p.cleanStaleMessages()
 				return nil
-			})
+			}); err != nil {
+				logger.With().Panic("error from cleanStaleMessages taskgroup", log.Err(err))
```
dshulyak: you shouldn't panic here. you can panic only in initializer, because stop can't be executed concurrently with initializer. in this case this is valid for tg.Go to return an error that tg was terminated.

p2p/server/msgserver.go
```diff
@@ -162,34 +164,40 @@ func (p *MessageServer) Close() {
...
 			select {
 			case p.workerLimiter <- struct{}{}:
-				p.tg.Go(func(ctx context.Context) error {
+				if err := p.tg.Go(func(ctx context.Context) error {
 					p.handleMessage(sctx, msg.(Message)) // pass session ctx to log session id
 					<-p.workerLimiter
 					return nil
-				})
+				}); err != nil {
+					logger.With().Panic("error from handleMessage taskgroup", log.Err(err))
```
dshulyak: the same. tbh i don't under5tand why you need to check this error. this is a part of the contract that this error is optional

dshulyak (changes requested):

p2p/server/msgserver.go
dshulyak: to explain a bit. this error is needed only if you want to know if goroutine was started or not. and for example if it wasn't started you know that tg was stopped and can exit immediately. printing this information is not useful at all

lrettig: Good to know, but I'm still a bit confused. Don't we want to know if the taskgroup was stopped, and therefore the task failed to schedule? I've reduced this from a panic to info, is that what you had in mind?

dshulyak (approved):

dshulyak:
bors try

lrettig:
bors merge

dshulyak:
bors try

dshulyak:
- failure in mining test is due to https://github.com/spacemeshos/go-spacemesh/issues/2764
- new failure in latenodes 
```
 E           2: file "assert_hare.py", line 95, in expect_consensus_process()
E           layer=21 expected=61 actual=60
E           
E           3: file "assert_hare.py", line 104, in expect_consensus_process()
E           layer=21 expected=61 actual=60
```

previously i didn't see this one. @countvonzero did you see this flake before?

dshulyak:
bors try

dshulyak:
This is definitely something related to the changes in this PR. latenodes haven't failed for me in develop at least 10 times in a row.

countvonzero:
@dshulyak agreed. this is a new failure for me too. have not seen it before. took a look of the first failure and none of the nodes fall out of sync. so data seems to be synced across nodes.

lrettig:
bors merge

lrettig:
bors merge

lrettig:
bors merge

dshulyak (changes requested):
latenodes doesn't fail so often in develop. please no need to retry until it succeeds

nkryuchkov:
bors merge

nkryuchkov (approved):

nkryuchkov:
bors merge

nkryuchkov:
bors try

nkryuchkov:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/3914832778)

https://github.com/spacemeshos/go-spacemesh/issues/2874

nkryuchkov:
bors try

nkryuchkov:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/3915010391)

late-nodes still fails

dshulyak:
@nkryuchkov what is the point of merging this? this code is replaced in my change

nkryuchkov:
Will be overridden by https://github.com/spacemeshos/go-spacemesh/pull/2801

PR #2449, opened 2021-06-09
Testnet 132/133 candidate

lrettig:
## Motivation
Candidate build for next testnet (132)

Closes #2454 

## Changes
- perform non-blocking send on msg channels, report error if the channels are full (as further mitigation of p2p send blocking gossip stack)
- don't falsely mark layer as having no blocks when layer hash requests fail (note: will probably not be ported over to develop, as the syncer code has already been rewritten)
- improved log output, error handling for debugging purposes
- parallelize processing of incoming gossip messages

PR #2464, opened 2021-06-18
[Merged by Bors] - continue processing a layer even when only one hash has blocks

countvonzero:
## Motivation
the code, as is, doesn't work when we receive layer blocks in the following order
t1: for layer hash1, we receive valid block data (err == nil)
t2: for layer hash2, we receive error 
sync.getLayerFromNeighbors() will determine there is error for this layer and stop processing.

## Changes
- make sync process a layer as long as there is a layer hash that returns non-empty blocks
- when there is an empty layer hash, and all other layer hashes return errors, determine this layer as empty

## Test Plan
added several unit tests. in particular, `Test_notifyLayerPromiseResult_OneHasBlockData()` fails before this change.

## DevOps Notes

### DIFF
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -75,7 +75,7 @@ type Logic struct {
 	fetcher              fetch.Fetcher
 	net                  network
 	layerHashResults     map[types.LayerID]map[p2ppeers.Peer]*types.Hash32
-	blockHashResults     map[types.LayerID][]bool //list of results from peers - true if blocks were received
+	blockHashResults     map[types.LayerID][]error
 	layerResultsChannels map[types.LayerID][]chan LayerPromiseResult
 	poetProofs           poetDb
 	atxs                 atxHandler
@@ -110,7 +110,7 @@ func NewLogic(ctx context.Context, cfg Config, blocks blockHandler, atxs atxHand
 		fetcher:              fetcher,
 		net:                  srv,
 		layerHashResults:     make(map[types.LayerID]map[p2ppeers.Peer]*types.Hash32),
-		blockHashResults:     make(map[types.LayerID][]bool),
+		blockHashResults:     make(map[types.LayerID][]error),
 		layerResultsChannels: make(map[types.LayerID][]chan LayerPromiseResult),
 		poetProofs:           poet,
 		atxs:                 atxs,
@@ -322,25 +322,24 @@ func (l *Logic) receiveLayerHash(ctx context.Context, id types.LayerID, p p2ppee
 		return
 	}
 
-	l.log.Info("got hashes %v for layer, now querying peers", len(hashes))
+	l.log.Info("got %v hashes for layer, now querying peers", len(hashes))
 
 	// send a request to get blocks from a single peer if multiple peers declare same hash per layer
 	// if the peers fails to respond request will be sen to next peer in line
 	//todo: think if we should aggregate or ask from multiple peers to have some redundancy in requests
-	for hash, peer := range hashes {
+	for hash, peers := range hashes {
 		if hash == emptyHash {
 			l.receiveBlockHashes(ctx, id, nil, len(hashes), ErrZeroLayer)
 			continue
 		}
-		//build receiver function
+		// build receiver function
 		receiveForPeerFunc := func(data []byte) {
 			l.receiveBlockHashes(ctx, id, data, len(hashes), nil)
 		}
-		remainingPeers := 0
 		errFunc := func(err error) {
 			l.receiveBlockHashes(ctx, id, nil, len(hashes), err)
 		}
-		err := l.net.SendRequest(ctx, LayerBlocksMsg, hash.Bytes(), peer[remainingPeers], receiveForPeerFunc, errFunc)
+		err := l.net.SendRequest(ctx, LayerBlocksMsg, hash.Bytes(), peers[0], receiveForPeerFunc, errFunc)
 		if err != nil {
 			l.receiveBlockHashes(ctx, id, nil, len(hashes), err)
 		}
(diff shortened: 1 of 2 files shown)


layerfetcher/layers.go
```diff
@@ -322,67 +322,70 @@ func (l *Logic) receiveLayerHash(ctx context.Context, id types.LayerID, p p2ppee
...
 
+func (l *Logic) determinePromiseResult(layer types.LayerID, errs []error) *LayerPromiseResult {
+	hasZeroBlockHash := false
+	for _, e := range errs {
+		if e == nil {
+			// at least one layer hash contains blocks. not a zero block layer
+			return &LayerPromiseResult{Layer: layer, Err: nil}
+		}
+		if e == ErrZeroLayer {
+			hasZeroBlockHash = true
```
antonlerner, line 358: you can put `return &LayerPromiseResult{Layer: layer, Err: ErrZeroLayer}` here

antonlerner (approved):

lrettig:
Please see these related changes from the testnet branch: https://github.com/spacemeshos/go-spacemesh/pull/2449/files#diff-14565cc7a7d654724cd76622a3373579a10b2ef86d8d2e52787d7f0c46089b87L1142-R1170

layerfetcher/layers.go
```diff
@@ -322,67 +322,70 @@ func (l *Logic) receiveLayerHash(ctx context.Context, id types.LayerID, p p2ppee
...
+		}
+	}
+	if hasZeroBlockHash {
+		// all other non-empty layer hashes returned errors. use the best information we've got
```
lrettig, line 362: Why "all other"? Either this comment is misleading or else I'm misunderstanding something. Wouldn't _one_ such result be enough to trigger this logic?

layerfetcher/layers_test.go
```diff
@@ -198,24 +199,114 @@ func Test_receiveLayerHash(t *testing.T) {
...
+
+	// test empty layer response
+	l.receiveLayerHash(context.TODO(), 1, net.peers[0], numOfPeers, nil, fmt.Errorf("error"))
+	for i := 1; i < numOfPeers; i++ {
+		l.receiveLayerHash(context.TODO(), 1, net.peers[i], numOfPeers, emptyHash.Bytes(), nil)
+	}
+	// zero-block layer should not incur additional send
+	assert.Equal(t, 4, net.sendCalled)
+
+	// test two many errors
```
lrettig: "two" or "many"?

layerfetcher/layers_test.go
```diff
@@ -198,24 +199,114 @@ func Test_receiveLayerHash(t *testing.T) {
...
+	result := make(chan LayerPromiseResult, 1)
+	l.layerResM.Lock()
+	l.layerResultsChannels[layer] = append(l.layerResultsChannels[1], result)
+	l.layerResM.Unlock()
+	l.notifyLayerPromiseResult(layer, 3, nil)
+	l.notifyLayerPromiseResult(layer, 3, nil)
+	l.notifyLayerPromiseResult(layer, 3, nil)
+	res := <-result
+	assert.Equal(t, layer, res.Layer)
+	assert.Equal(t, nil, res.Err)
```
lrettig:
```suggestion
	assert.Nil(t, res.Err)
```

lrettig (changes requested):
Could you help me understand why a _single_ zero layer result trumps even many other non-zero layer results? I'm not totally convinced this is the correct logic

layerfetcher/layers.go
countvonzero: i mean all other hashes that's not empty hash returned errors.

because we want to return a valid result if there is one. hence the early return at line 355.
so seeing one such result (ErrZeroLayer) isn't enough. we need to make sure there is no valid result before we determine this is an zero-block layer. see test case Test_notifyLayerPromiseResult_OneHasBlockData

layerfetcher/layers_test.go
countvonzero: too many :D

countvonzero: it tests the case where numErrors > peers/2, then it errors out without doing anything.

countvonzero:
> Could you help me understand why a _single_ zero layer result trumps even many other non-zero layer results? I'm not totally convinced this is the correct logic

@lrettig that's not the logic. the logic is, if there is any valid non-zero layer results, it returns immediately. see [line 355](https://github.com/spacemeshos/go-spacemesh/pull/2464/files/af7189824b1d1f7990d4a6842ab10d6684eb0a12#diff-a0088fa89f106feef536e183048c1b7b443daf2b16951b445caec755e7a23b9aR355).
 [line 355]. after you exit the loop, you are sure all that's left are errors. and if there is any ErrZeroLayer, it trumps all other types of errors.

countvonzero:
> > Could you help me understand why a _single_ zero layer result trumps even many other non-zero layer results? I'm not totally convinced this is the correct logic

@lrettig this is exactly i am fixing. the test i added Test_notifyLayerPromiseResult_OneHasBlockData(). this test fails before this change and passes afterwards.

layerfetcher/layers.go
```diff
@@ -392,7 +395,7 @@ func (l *Logic) notifyLayerPromiseResult(id types.LayerID, expectedResults int,
 func (l *Logic) receiveBlockHashes(ctx context.Context, layer types.LayerID, data []byte, expectedResults int, extErr error) {
 	//if we failed getting layer data - notify
 	if extErr != nil {
-		l.log.Error("received error for layer id %v", extErr)
+		l.log.Error("received error %v for layer id %v", extErr, layer)
```
lrettig:
```suggestion
		l.log.With().Error("received error", log.Err(extErr), layer)
```

lrettig (approved):
thanks for the explanation!

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #2466, opened 2021-06-20
[Merged by Bors] - Refactor metrics

valar999:
## Motivation

Put metrics in order. Use one library for all metrics.
https://github.com/spacemeshos/pm/issues/37

## Changes

Refactor metrics in `sync`, `tortoise`, `hare` and `p2p` packages.

## TODO
- [x] Set metrics.Enable=true when Server or push-metrics are enabled 
- [ ] Decide is it good to use go-kit
- [ ] Find out what is MeteredCache is for

## DevOps Notes


valar999:
A few comments about `go-kit`:
- it has no `Summary` and replace it with `Histogram`. But these are completely different things.
- go-kit allows to use variable in label names. We will have to remove this kind of variable https://github.com/spacemeshos/go-spacemesh/blob/v0.1.41/p2p/metrics/prometheus.go#L18 if we decide to get rid of go-kit

@noamnelke please consider

tortoise/metrics.go
```diff
@@ -1,24 +1,42 @@
...
+		[]string{"validity"},
+	)
+	invalidBlocks = blockVotes.With("validity", "invalid")
+
+	validBlocks = metrics.NewGauge(
+		"validity",
+		MetricsSubsystem,
+		"Valid",
+		nil,
+	)
```
noamnelke: Why are these two so different?

noamnelke (approved):

tortoise/metrics.go
valar999: Thank you. It was a mistake. Fixed

valar999:
bors merge

valar999:
bors merge

valar999:
bors merge

valar999:
bors merge

PR #2468, opened 2021-06-21
[Merged by Bors] - events: address possible issues in pubsub

dshulyak:
## Motivation
There are several issues with pubsub, that could lead to goroutines leak and unnecessary memory allocations.
Found them while getting familiar with the code. If you prefer I can open bugs for each issue separately.

## Changes

- avoid mem allocation for the payload in the debug log when message is sent or received
- allow subscriber to exit if subscriber was closed and send is blocking
- test that empty payload will not crash subscriber
- add utilities to find the free port and wait until the subscriber is connected
  with publisher

## Test Plan
unit tests are sufficient

## DevOps Notes


dshulyak:
@antonlerner cant add you as a reviewer for some reason, but please take a look

avive:
Q: Is the pubsub code currently used in the node anywhere? I thought we have phased it out in favor of the grpc api? Maybe it is used for some tests only?

dshulyak:
@avive This is probably the case. Initially, I thought that this code is extensively used because it was imported everywhere. But it looks like that the subscriber is only initialized with SetupMultiNode, which is used only in a couple of tests and multi node simulation.

antonlerner (approved):

dshulyak:
bors merge

PR #2473, opened 2021-06-22
updated Makefile & Dockerfiles

sudachen:
# Motivation
Since gpu-setup shared library for linux is linked for glibc and can't be used on alpine we need to create docker images based on ubuntu dist. Also it's required to download and unpack appropriate set of binaries for selected build target. 

# Changes
New Dockefiles based on Ubuntu
Makefile changes related to gpu-post binaries setup.


scripts/check-go-version.go
```diff
@@ -0,0 +1,28 @@
...
+	flag.IntVar(&minMajor, "major", 1, "required major version")
+	flag.IntVar(&minMinor, "minor", 14, "required minor version")
+}
+
+func main() {
+	flag.Parse()
+	ver := strings.Split(runtime.Version()[2:], ".")
+	major, _ := strconv.Atoi(ver[0])
+	minor, _ := strconv.Atoi(ver[1])
+	if major < minMajor || minor < minMinor {
```
noamnelke:
```suggestion
	if major < minMajor || (major == minMajor && minor < minMinor) {
```

Dockerfile
```diff
@@ -1,7 +1,58 @@
...
+    bash \
+    sudo \
+    unzip \
+    make \
+    curl \
+    procps \
+    net-tools \
+    apt-transport-https \
+    # -- it allows to start with nvidia-docker runtime --
+    libnvidia-compute-390 \
```
noamnelke: I assume we need all of these deps on the machine where we build the binary, but we surely don't need all of this to RUN the binary!

If we do - will home users need to also install all of this to run Spacemesh?

If we don't - let's install this on the `golang` image instead of `linux` to make the shipped artifact leaner.

noamnelke (commented):
Haven't reviewed everything, but added a couple of comments.

Dockerfile
sudachen: Yeh. I used it for trubleshutting

tests/config.yaml
```diff
@@ -41,14 +41,19 @@ bootstrap:
...
     tortoise-beacon-votes-limit: '100'
     layers-per-epoch: '3'
-    coinbase: '0x4321'
     eligibility-confidence-param: '5'
     eligibility-epoch-offset: '0'
-    post-space: '1024'
-    space-to-commit: '2048'
     executable-path: '/bin/go-spacemesh'
-#    grpc-port: '19092'
-#    json-port: '19093'
```
noamnelke, line 51: why?

tests/config.yaml
```diff
@@ -81,14 +86,19 @@ client:
...
     tortoise-beacon-votes-limit: '100'
     layers-per-epoch: '3'
-    coinbase: '0x1234'
     eligibility-confidence-param: '5'
     eligibility-epoch-offset: '0'
-    post-space: '1024'
-    space-to-commit: '2048'
     executable-path: '/bin/go-spacemesh'
-#    grpc-port: '19092'
-#    json-port: '19093'
```
noamnelke, line 91: also why?

Makefile
```diff
@@ -61,11 +56,12 @@ endif
 # This prevents "the input device is not a TTY" error from docker in CI
 DOCKERRUNARGS := --rm \
 	-e GOOGLE_APPLICATION_CREDENTIALS=./spacemesh.json \
-	-e CLUSTER_NAME=$(CLUSTER_NAME) \
+ 	-e CLUSTER_NAME=$(CLUSTER_NAME) \
```
noamnelke, line 59: remove extra space

Makefile
```diff
@@ -211,73 +161,68 @@ lint:
...
-
 dockerbuild-test:
-	docker build -f DockerFileTests --build-arg GCLOUD_KEY="$(GCLOUD_KEY)" \
+	docker build -f DockerFileTests \
+	             --build-arg GCLOUD_KEY="$(GCLOUD_KEY)" \
 	             --build-arg PROJECT_NAME="$(PROJECT_NAME)" \
-	             --build-arg CLUSTER_NAME="$(CLUSTER_NAME)" \
-	             --build-arg CLUSTER_ZONE="$(CLUSTER_ZONE)" \
+	             --build-arg CLUSTER_NAME="$(CLUSTER_NAME_ELK)" \
+	             --build-arg CLUSTER_ZONE="$(CLUSTER_ZONE_ELK)" \
```
noamnelke, line 198: Please revert this, we're deprecating the `_ELK` suffix. The updated env currently has the same value with/without the suffix and we want to remove the version with the suffix soon.

DockerfilePrebuiltBinary
```diff
@@ -1,10 +1,42 @@
...
+    locales \
+    git \
+    bash \
+    sudo \
+    unzip \
+    make \
+    curl \
+    procps \
+    net-tools \
+    apt-transport-https \
```
noamnelke: This makes no sense... why would all this be needed in runtime? We're not gonna require users to install all these dev tools...

noamnelke (changes requested):
@sudachen we're putting all the dev tools in the final docker - this should be changed. It should be a minimal installation of only dependencies for runtime.

tests/config.yaml
sudachen: it does not exists in develop.

tests/config.yaml
sudachen: it does not exists in develop.

DockerfilePrebuiltBinary
sudachen: I removed all staff except porcps (ps command) and net-tools (ping command)

PR #2479, opened 2021-06-24
[Merged by Bors] - Use uint32 to work with network id

### LINKED ISSUE #2113: Increase network ID range
## Description
Right now the `network-id` in the config cannot be larger than 128 since it's stored in `int8`.

We should have a larger range to make it easier to segment into different types of networks, e.g., testnet is `1xx`, devnet `2xx`, local networks `1xxx`, etc.

Also I'm not sure why this is an int rather than uint, since I don't think we ever want it to be negative.

CC @avive @noamnelke @y0sher any thoughts to add here? Any reason not to change this?

## Affected code
https://github.com/spacemeshos/go-spacemesh/blob/b4a7bcdd152d5f4af3da00abbadab2439b9ce90f/cmd/root.go#L81

This issue appears in commit hash: b4a7bcdd152d5f4af3da00abbadab2439b9ce90f

dshulyak:
## Motivation
Int8 is too short and forces reuse of the same ids for testnets.
Closes: https://github.com/spacemeshos/go-spacemesh/issues/2113

## Changes
- NetworkID is stored internally as uint32 instead of int8.
- handshake data will use uint32 instead of int32 for consistency

## Test Plan
Existing unit and system tests.

## TODO

## DevOps Notes

### DIFF
--- a/api/grpcserver/mesh_service.go
+++ b/api/grpcserver/mesh_service.go
@@ -19,7 +19,7 @@ type MeshService struct {
 	Mempool          api.MempoolAPI
 	GenTime          api.GenesisTimeAPI
 	LayersPerEpoch   int
-	NetworkID        int8
+	NetworkID        uint32
 	LayerDurationSec int
 	LayerAvgSize     int
 	TxsPerBlock      int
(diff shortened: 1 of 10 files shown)


antonlerner (approved):

lrettig (approved):

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #2482, opened 2021-06-25
[Merged by Bors] - Use latest received layer in NodeService.Status response during genesis

### LINKED ISSUE #2333: Return true current sync layer in API

dshulyak:
## Motivation
Closes https://github.com/spacemeshos/go-spacemesh/issues/2333

## Changes
During genesis, NodeService Status will return the latest received layer as the Synced and Verified layer.

## Test Plan
Existing tests

## TODO

## DevOps 

### DIFF
--- a/api/grpcserver/grpcserver_test.go
+++ b/api/grpcserver/grpcserver_test.go
@@ -567,9 +567,9 @@ func TestNodeService(t *testing.T) {
 			require.NoError(t, err)
 			require.Equal(t, uint64(0), res.Status.ConnectedPeers)
 			require.Equal(t, false, res.Status.IsSynced)
-			require.Equal(t, uint32(layerCurrent), res.Status.SyncedLayer.Number)
+			require.Equal(t, uint32(layerLatest), res.Status.SyncedLayer.Number)
 			require.Equal(t, uint32(layerCurrent), res.Status.TopLayer.Number)
-			require.Equal(t, uint32(layerCurrent), res.Status.VerifiedLayer.Number)
+			require.Equal(t, uint32(layerLatest), res.Status.VerifiedLayer.Number)
 
 			// Now do a mock check post-genesis
 			layerCurrent = oldCurLayer
--- a/api/grpcserver/node_service.go
+++ b/api/grpcserver/node_service.go
@@ -91,8 +91,8 @@ func (s NodeService) getLayers() (curLayer, latestLayer, verifiedLayer uint32) {
 	curLayerObj := s.GenTime.GetCurrentLayer()
 	curLayer = uint32(curLayerObj)
 	if curLayerObj.GetEpoch().IsGenesis() {
-		latestLayer = curLayer
-		verifiedLayer = curLayer
+		latestLayer = uint32(s.Mesh.LatestLayer())
+		verifiedLayer = latestLayer
 	} else {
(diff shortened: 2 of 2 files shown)


lrettig (changes requested):
Please see my comment here: https://github.com/spacemeshos/go-spacemesh/issues/2333#issuecomment-869013979

This change only needs to happen when `curLayerObj.GetEpoch().IsGenesis()` is true, i.e., the first part of this conditional, not the second. It needs to update both `latestLayer` and `verifiedLayer`. And the value should not be read from `s.Mesh`. It needs to come from the state of the syncer, which is not currently injected into the API.

Let me know if anything is unclear. Thanks!

dshulyak:
@lrettig syncer updates the state on the mesh ([ref](https://github.com/spacemeshos/go-spacemesh/blob/develop/sync/syncer.go#L492-L524)). if you are worried that in the future syncer may start to store this value in its own state i can surely extend syncer

lrettig:
> @lrettig syncer updates the state on the mesh ([ref](https://github.com/spacemeshos/go-spacemesh/blob/develop/sync/syncer.go#L492-L524)). if you are worried that in the future syncer may start to store this value in its own state i can surely extend syncer

During genesis, when syncer receives a zero-block layer, it does not even call `ValidateLayer` since there's nothing to validate. It just calls `SetZeroBlockLayer` in the mesh:

https://github.com/spacemeshos/go-spacemesh/blob/9116d1fd4f6f4fadef986d12212a938a62f87e97/sync/syncer.go#L512-L517

This calls `SetLatestLayer` but not `SetProcessedLayer`.

All we need to do here is ensure that the API code can fetch the latest synced layer and report that during genesis.

The title of this PR should probably be changed to something like, "Use latest synced layer in NodeService response during genesis"

dshulyak:
@lrettig as i can see ValidateLayer is called in any case. this condition https://github.com/spacemeshos/go-spacemesh/blob/9116d1fd4f6f4fadef986d12212a938a62f87e97/sync/syncer.go#L512-L517 doesn't interrupt the loop

lrettig:
If the syncer is always setting the latest received layer in the mesh, then it might be enough to return this from the API!

dshulyak:
as i mentioned, syncer **sometimes** updates Latest (when no blocks), and **always** updates Processed (but not always persists it on disk, which seems to be a bug from my perspective). i will try to clarify if this is the intented behavior. to me it seems that it doesn't matter if we return Latest or Processed as they are always updated synchronously one after the other when a layer without blocks is received

lrettig (approved):
lgtm

dshulyak:
bors merge

PR #2483, opened 2021-06-27
[Merged by Bors] - syncer rewrite

### LINKED ISSUE #2471: simplify sync logic and add tests
## description
current sync code has two code paths:
- steady-state: weakly synced -> fully synced
- catch-up sync: from layer 0 -> fully synced + 1 layer of gossip-sync -> fully synced + fully gossip

issues:
- these two code path only behave differently when they get close to the current layer, but they use different sync logic and code sharing is minimum. this make it challenging to understand and make future code changes
- tests are not in a good state. this further reduces confidence when making code changes
- the sync state is not very well defined in the code. there should be a clear definition of what the node can do in a given sync state

plan: 
- simplify the sync flow to use the same sync logic as much as possible
- add unit tests / integration tests

reference design:
https://github.com/spacemeshos/SMIPS/issues/43

countvonzero:
## Motivation
Closes #2471

## Changes
simplify synchronize logic
- use one loop to capture both stead-state sync and catch up sync
- separate logic for data syncing and layer validation. this paves the road for parallelize data syncing while keeping validating layers in order

## Test Plan
unit test + make test + CI

## DevOps Notes

### DIFF
--- a/cmd/sync/sync.go
+++ b/cmd/sync/sync.go
@@ -8,13 +8,10 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"cloud.google.com/go/storage"
-	"github.com/spf13/cobra"
-	"google.golang.org/api/iterator"
-	"google.golang.org/api/option"
-
 	"github.com/spacemeshos/go-spacemesh/activation"
 	cmdp "github.com/spacemeshos/go-spacemesh/cmd"
 	"github.com/spacemeshos/go-spacemesh/common/types"
@@ -24,9 +21,10 @@ import (
 	"github.com/spacemeshos/go-spacemesh/mesh"
 	"github.com/spacemeshos/go-spacemesh/p2p"
 	"github.com/spacemeshos/go-spacemesh/state"
-	"github.com/spacemeshos/go-spacemesh/sync"
-	"github.com/spacemeshos/go-spacemesh/timesync"
-	"strings"
+	"github.com/spacemeshos/go-spacemesh/syncer"
+	"github.com/spf13/cobra"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
 )
 
 // Sync cmd
@@ -68,8 +66,9 @@ func init() {
 
 type syncApp struct {
 	*cmdp.BaseApp
-	sync  *sync.Syncer
-	clock *timesync.Ticker
+	sync   *syncer.Syncer
+	msh    *mesh.Mesh
+	logger log.Log
 }
 
 func newSyncApp() *syncApp {
@@ -79,11 +78,11 @@ func newSyncApp() *syncApp {
 func (app *syncApp) Cleanup() {
 	err := os.RemoveAll(app.Config.DataDir())
 	if err != nil {
-		app.sync.With().Error("failed to cleanup sync", log.Err(err))
+		app.logger.With().Error("failed to cleanup sync", log.Err(err))
 	}
 }
 
-func (app *syncApp) start(cmd *cobra.Command, args []string) {
+func (app *syncApp) start(_ *cobra.Command, _ []string) {
 	// start p2p services
 	lg := log.NewDefault("sync_test")
 	lg.With().Info("------------ Start sync test -----------",
@@ -105,18 +104,6 @@ func (app *syncApp) start(cmd *cobra.Command, args []string) {
 	}
 
 	goldenATXID := types.ATXID(types.HexToHash32(app.Config.GoldenATXID))
-
-	conf := sync.Configuration{
-		Concurrency:     4,
-		AtxsLimit:       200,
-		LayerSize:       app.Config.LayerAvgSize,
-		RequestTimeout:  time.Duration(app.Config.SyncRequestTimeout) * time.Millisecond,
-		SyncInterval:    2 * 60 * time.Millisecond,
-		Hdist:           app.Config.Hdist,
-		ValidationDelta: 30 * time.Second,
-		LayersPerEpoch:  uint16(app.Config.LayersPerEpoch),
-		GoldenATXID:     goldenATXID,
-	}
 	types.SetLayersPerEpoch(int32(app.Config.LayersPerEpoch))
 	lg.Info("local db path: %v layers per epoch: %v", path, app.Config.LayersPerEpoch)
 
@@ -146,17 +133,36 @@ func (app *syncApp) start(cmd *cobra.Command, args []string) {
 	}
 
 	txpool := state.NewTxMemPool()
-	atxpool := activation.NewAtxMemPool()
 
-	app.sync = sync.NewSyncWithMocks(atxdbStore, mshdb, txpool, atxpool, swarm, poetDb, conf, goldenATXID, types.LayerID(expectedLayers), poetDbStore)
+	app.logger = log.NewDefault("sync_test")
+	app.logger.Info("new sync tester")
+
+	layersPerEpoch := app.Config.LayersPerEpoch
+	atxdb := activation.NewDB(atxdbStore, &mockIStore{}, mshdb, uint16(layersPerEpoch), goldenATXID, &validatorMock{}, lg.WithOptions(log.Nop))
+	dbs := &allDbs{
+		atxdb:       atxdb,
+		atxdbStore:  atxdbStore,
+		poetDb:      poetDb,
+		poetStorage: poetDbStore,
+		mshdb:       mshdb,
+	}
+	msh := createMockMesh(dbs, txpool, app.logger)
+	app.msh = msh
+	layerFetch := createMockFetcher(dbs, msh, swarm, app.logger)
+	layerFetch.Start()
+	syncerConf := syncer.Configuration{
+		SyncInterval:    2 * 60 * time.Millisecond,
+		ValidationDelta: 30 * time.Second,
+	}
+	app.sync = createSyncer(syncerConf, msh, layerFetch, types.LayerID(expectedLayers), app.logger)
 	if err = swarm.Start(cmdp.Ctx); err != nil {
 		log.With().Panic("error starting p2p", log.Err(err))
 	}
 
-	i := conf.LayersPerEpoch * 2
+	i := layersPerEpoch * 2
 	for ; ; i++ {
 		lg.With().Info("getting layer", types.LayerID(i))
-		if lyr, err2 := app.sync.GetLayer(types.LayerID(i)); err2 != nil || lyr == nil {
+		if lyr, err2 := msh.GetLayer(types.LayerID(i)); err2 != nil || lyr == nil {
 			l := types.LayerID(i)
 			if l > types.GetEffectiveGenesis() {
 				lg.With().Info("finished loading layers from disk",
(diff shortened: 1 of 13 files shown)


syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+			log.FieldNamed("latest", s.mesh.LatestLayer()),
+			log.FieldNamed("validated", s.mesh.ProcessedLayer()))
+		events.ReportNodeStatusUpdate()
+		// notify subscribes
+		s.awaitSyncedMu.Lock()
+		defer s.awaitSyncedMu.Unlock()
+		if newState == synced {
+			close(s.awaitSyncedCh)
+		} else if oldState == synced {
+			s.awaitSyncedCh = make(chan struct{})
```
antonlerner: since you are replacing the channel maybe we should change how `awaitSynced` works because this switch might cause other processes to listen to the wrong channel

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+// Close stops the syncing process and the goroutines syncer spawns.
+func (s *Syncer) Close() {
+	// TODO: ensure goroutines are all terminated before shutting down
+	s.cancelFunc()
+}
+
+// AwaitSynced returns the channel that will be notified whenever the node enters synced state.
+func (s *Syncer) AwaitSynced() chan struct{} {
```
antonlerner: I think we should change this function to read from the channel and listen to result, since `s.awaitSyncedCh` can be switched out

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+		if newState == synced {
+			close(s.awaitSyncedCh)
+		} else if oldState == synced {
+			s.awaitSyncedCh = make(chan struct{})
+		}
+	}
+}
+
+// startNewSyncProcess returns true if syncer is not currently synchronizing.
+func (s *Syncer) startNewSyncProcess() bool {
```
antonlerner: i'd rename it, this function does not start a new sync process :)

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+	return atomic.CompareAndSwapUint32(&s.isBusy, 0, 1)
+}
+
+func (s *Syncer) endSyncProcess() {
+	atomic.StoreUint32(&s.isBusy, 0)
```
antonlerner, line 215: same here

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+func (s *Syncer) setTargetSyncedLayer(ctx context.Context, layerID types.LayerID) {
```
antonlerner, line 219: add a comment here as well to what a "target sync layer" is

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+		logger.With().Info("node is synced. validation done",
+			log.FieldNamed("current", s.ticker.GetCurrentLayer()),
+			log.FieldNamed("latest", s.mesh.LatestLayer()),
+			log.FieldNamed("validated", s.mesh.ProcessedLayer()))
+		logger.Info("finished sync run #%v", s.run)
+		s.endSyncProcess()
+	}()
+
+	// using ProcessedLayer() instead of LatestLayer() so we can validate layers on a best-efforts basis.
+	// TODO: is it safe to skip layer 0
```
antonlerner: our clock starts ticking from 1 so yes, you can skip layer 0

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+			log.FieldNamed("behindThreshold", outOfSyncThreshold))
+		s.setSyncState(ctx, notSynced)
+	}
+}
+
+func (s *Syncer) setStateAfterSync(ctx context.Context) {
+	currSyncState := s.getSyncState()
+	current := s.ticker.GetCurrentLayer()
+	// if we have gossip-synced to the target synced layer, we are ready to participate in the consensus
+	if currSyncState == gossipSync && s.getTargetSyncedLayer() <= current {
```
antonlerner, line 326: this will set state to synced even if we did not validate all the layers in validation queue and I think this might cause an inconsistent state
I think that coming out of the sync - all nodes should have the same state

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+func (s *Syncer) syncLayer(ctx context.Context, layerID types.LayerID) (*types.Layer, error) {
+	if s.isClosed() {
+		return nil, errors.New("shutdown")
+	}
+
+	layer, err := s.getLayerFromPeers(ctx, layerID)
+	if err != nil {
+		s.logger.WithContext(ctx).With().Error("failed to get layer from neighbors", layerID, log.Err(err))
```
antonlerner: do we have a test for this case? failure to sync

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+	return s.mesh.GetLayer(layerID)
+}
+
+func (s *Syncer) getATXs(ctx context.Context, layerID types.LayerID) {
+	if layerID.GetEpoch() == 0 {
+		s.logger.WithContext(ctx).Info("skip getting ATX in epoch 0")
+		return
+	}
+	lastLayerOfEpoch := (layerID.GetEpoch() + 1).FirstLayer() - 1
+	if layerID != lastLayerOfEpoch {
```
antonlerner: we need to sync ATX also for the current epoch
consider that we might become synced in the middle of an epoch so this condition will not met but still - all ATXs published in this epoch must be synced.

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+				return false
+			}
+		}
+
+		if s.shouldValidateLayer(layerID) {
+			vQueue <- layer
+		}
+		logger.Info("finished data sync for layer %v", layerID)
+	}
+	s.setStateAfterSync(ctx)
```
antonlerner: I think we need to sync ATXs here before becoming "synced" 
see my comment in `getATXs`

syncer/syncer.go
```diff
@@ -0,0 +1,402 @@
...
+	s.run++
+	logger.Info("staring sync run #%v", s.run)
+
+	s.setStateBeforeSync(ctx)
+	// start a dedicated process for validation.
+	// do not use a unbuffered channel for vQueue. we don't want it to block if the receiver isn't ready. i.e.
+	// if validation for the last layer is still running
+	vQueue := make(chan *types.Layer, s.ticker.GetCurrentLayer())
+	vDone := make(chan struct{})
+	go s.startValidating(ctx, s.run, vQueue, vDone)
```
antonlerner, line 257: this will leak go routines as each time `synchronize` is called a new routine will be created
Perhaps move this code to constructor

antonlerner (changes requested):

syncer/syncer.go
countvonzero: you are right.
changed to use client supplied channel for notification.
activation.go wants to select on a channel so i don't want to make AwaitSynced() block.

syncer/syncer.go
countvonzero: done. but changed to `RegisterChForSynced` and uses client's channel to notify. the only caller of this is activation.go and it really wants a channel to select on. and i don't want to change logic in activation.go either.

syncer/syncer.go
countvonzero: done

syncer/syncer.go
countvonzero: done

syncer/syncer.go
countvonzero: done

syncer/syncer.go
countvonzero: thanks

syncer/syncer.go
countvonzero: changed to update state ONLY after validation is done.

syncer/syncer.go
countvonzero: yes. i renamed it to `TestSynchronize_getLayerFromPeersFailed` to be clear.

syncer/syncer.go
countvonzero: done. 

note that this is different behavior from the old syncer impl. but `dockertest-sync-elk` passed on this change.

syncer/syncer.go
countvonzero: done.

note that this is also different behavior from the old syncer impl. but dockertest-sync-elk also passed on this change.

syncer/syncer.go
countvonzero: i don't believe this will cause goroutine leak. the `defer` call below close `vQueue` and block on `vDone`. once `vQueue` is closed, the range for loop in startValidating() will terminate and exit.

doing it in the constructor is also a viable option. my preference is to keep each sync run independent for ease of logging/debugging. if you think using just one validation goroutine is a better design, i will move it into the constructor.

countvonzero:
@antonlerner thank you so much for the speedy review. PTAL.

also added a test case where the syncer goes from notSynced->gossipSync->notSynced->gossipSync->synced

countvonzero:
bors try

syncer/syncer.go
```diff
@@ -0,0 +1,419 @@
...
+func (s *Syncer) getATXs(ctx context.Context, layerID types.LayerID) error {
+	if layerID.GetEpoch() == 0 {
+		s.logger.WithContext(ctx).Info("skip getting ATX in epoch 0")
+		return nil
+	}
+	epoch := layerID.GetEpoch()
+	s.logger.WithContext(ctx).With().Info("getting ATXs", epoch, layerID)
+	ctx = log.WithNewRequestID(ctx, layerID.GetEpoch())
+	if err := s.fetcher.GetEpochATXs(ctx, layerID.GetEpoch()); err != nil {
+		// TODO: is this really expected? we should have ATXs in epoch 1
```
countvonzero: @antonlerner 

we already skip ATX for epch 0. should we ignore error for epoch 1? i'd imagine we will have ATX in epoch 1 when miners do the PoET work during epoch 0.

syncer/syncer.go
antonlerner: I stand corrected, missed the `close(vQueue)`

syncer/syncer.go
antonlerner: yes we need to ignore only in epoch 0

syncer/syncer.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+	return s.mesh.GetLayer(layerID)
+}
+
+func (s *Syncer) getATXs(ctx context.Context, layerID types.LayerID) error {
+	if layerID.GetEpoch() == 0 {
+		s.logger.WithContext(ctx).Info("skip getting ATX in epoch 0")
+		return nil
+	}
+	epoch := layerID.GetEpoch()
```
antonlerner, line 369: now we get ATXs for each layer of an epoch, we can however optimise this, if we are syncing an epoch that is not the current epoch we can sync only once
If we are in the current epoch we will need to query for each layer or try to sync atxs when we sync the last layer we need to sync in current epoch. - if implemented this way - be careful since syncing atxs might take a while

syncer/syncer.go
countvonzero: fixed. thanks

syncer/syncer.go
countvonzero: changed to what you suggested.
- get ATXs in the last layer of an old epoch
- get ATXs for every layer of current epoch to amortize the latency

countvonzero:
bors try

antonlerner (approved):

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/2949988273)

note that i cancelled this try run.

countvonzero:
bors try

countvonzero:
bors try

cmd/sync/sync.go
```diff
@@ -146,17 +133,36 @@ func (app *syncApp) start(cmd *cobra.Command, args []string) {
...
+	dbs := &allDbs{
+		atxdb:       atxdb,
+		atxdbStore:  atxdbStore,
+		poetDb:      poetDb,
+		poetStorage: poetDbStore,
+		mshdb:       mshdb,
+	}
+	msh := createMockMesh(dbs, txpool, app.logger)
+	app.msh = msh
+	layerFetch := createMockFetcher(dbs, msh, swarm, app.logger)
```
antonlerner: The purpose of the sync test is to test actual syncing of data on a live network so I think we should use actual fetcher instead

countvonzero: this is a real Fetcher with a mockTxProcessor. do you want me to rename the method instead?

countvonzero: offline conversation. resolved.

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

PR #2480, opened 2021-06-24
[Merged by Bors] - Log hostname when spacemesh node is started

dshulyak:
## Motivation
Make it easier to identify the node.
Closes: https://github.com/spacemeshos/go-spacemesh/issues/2318

## Changes
Example log 
```
starting spacemesh	{"data-dir": "/home/dd/.spacemesh-data/1", "post-dir": "/home/dd/post/data", "hostname": "dd-MS-7C37"}
```

## Test Plan
Existing test suite

## TODO
- [x] Explain motivation or link existing issue(s)
- [x] Test changes and document test plan
- [x] Update documentation as needed

## DevOps Notes
- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [ ] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #2488, opened 2021-06-29
[Merged by Bors] - cmd/node: exit with os.Exit(1) on error

dshulyak:
depends on: https://github.com/spacemeshos/go-spacemesh/pull/2480/
## Motivation
The code that handles exit on error was inconsistent, sometimes it was printing an error, sometimes exited with stacktrace, and in one case exited but with exit code set to 0.

## Changes
If any component fails to start exit from spacemesh node with non zero exit code, and without stacktrace.

## Test Plan
Existing teests

## TODO

## DevOps Notes


avive:
Please keep in mind to always report any node panic / critical error via the node api so Smapp can display it to user and display restart node button in network screen.

cmd/node/node.go
```diff
@@ -991,7 +994,7 @@ func (app *SpacemeshApp) Start(*cobra.Command, []string) {
 		defer srv.Shutdown(ctx)
 		go func() {
 			if err := srv.ListenAndServe(); err != nil {
-				logger.With().Error("cannot start pprof http server", log.Err(err))
+				logger.With().Fatal("cannot start pprof http server", log.Err(err))
```
noamnelke: Not sure this should be fatal

cmd/node/node.go
```diff
@@ -1005,7 +1008,7 @@ func (app *SpacemeshApp) Start(*cobra.Command, []string) {
 			// by default all profilers are enabled,
 		})
 		if err != nil {
-			logger.With().Error("cannot start profiling client")
+			logger.With().Fatal("cannot start profiling client")
```
noamnelke: This should also not be fatal

dshulyak:
@avive i don't see any errors that are reported via API. maybe smapp checks exit code?
i changed errors to Fatal only during initialization of the node, i obviously didn't touch any that occur when application is running

avive:
This is why I mentioned it. I'm not sure if this should be part of this PR but go-sm should report an error via the API before *any* panic in order to provide a good user experience to Smapp users. This is one of the top usability issues with Smapp - node dying under the hood and user has no knowledge of it and can't restart the node with 1-click...  So, anytime before the node gives up and crashes, an error event should be sent over the ErrorStream server-side streaming method. Smapp listens on this and updates the Network screen UI based on it. @IlyaVi @moshababo

noamnelke (changes requested):

cmd/node/node.go
dshulyak: then it should be disabled. otherwise if developer wants to run the node he should not search the logs to understand if pprof was started succesfully

noamnelke:
@avive the node should probably report such errors via the API where it can provide rich details, but regardless, the SMApp MUST handle panics (even without an error on the API) because the node might always have an unexpected/unhandled panic that was not initiated in our code. Whenever the node exits with a non-zero exit code - the SMApp must handle it gracefully, even if no error was reported via the API.

cmd/node/node.go
dshulyak: @noamnelke  i don't agree with the logic that this service is in some sense optional. if it is enabled the intention is pretty clear. imagine if you are running two clients, and you forgot about the first one but they share a port. the second one will fail, but you are not guaranteed to notice it, and will simply collect info from the first one

avive:
> @avive the node should probably report such errors via the API where it can provide rich details, but regardless, the SMApp MUST handle panics (even without an error on the API) because the node might always have an unexpected/unhandled panic that was not initiated in our code. Whenever the node exits with a non-zero exit code - the SMApp must handle it gracefully, even if no error was reported via the API.

Can you please clarify the design of critical error handling in the node?
Specifically, are there critical errors that do NOT cause a panic but require the user to restart the node? If yes, then these errors MUST and not just probably be reported via the ErrorStream method. Otherwise the user will not know about them and the need to restart the node.

Smapp will also need to catch panics in the node process. @IlyaVi - in case of a node panic, please display node crashed in the network screen and an enabled 'restart node' button.

We know from Smapp 0.1 that having the managed node crash silently and Smapp not notifying the user about it and giving him a chance to restart the node without having to restart Smapp is one of the most annoying app behavior for users.

noamnelke:
> Can you please clarify the design of critical error handling in the node?
Specifically, are there critical errors that do NOT cause a panic but require the user to restart the node? If yes, then these errors MUST and not just probably be reported via the ErrorStream method. Otherwise the user will not know about them and the need to restart the node.

When an unrecoverable condition occurs in the node we panic. When we panic on purpose we can do it "gracefully" (e.g. report an error before terminating). However, there can be situations (mostly due to a bug) where a panic occurs where we didn't explicitly choose to panic (e.g. division by zero, concurrent map writes, etc.) - we can't blanket catch all of these cases because of how Go is designed (a panic in any goroutine terminates the entire process and a goroutine may be started by some library code, for example). So when an unhandled panic occurs the SMApp must reflect this to the user.

cmd/node/node.go
noamnelke: cool, you convinced me.

noamnelke (approved):

lrettig (commented):
What's the difference between panic and fatal?

lrettig:
> @avive i don't see any errors that are reported via API.

All errors are reported over the NodeService.ErrorStream if it's running. I'm actually not sure what happens in case of a panic, whether the error would make it over the stream before the node shut down, but I believe there's a test for this case and I believe it works.

noamnelke:
> What's the difference between panic and fatal?

I looked it up, didn't know either...

https://quasilyte.dev/blog/post/log-fatal-vs-log-panic/

The bottom line is that unlike `Panic`, which calls `panic`, `Fatal` calls `os.Exit(1)` which prevents any deferred code from running.

I'm actually not sure that's what we want... When a Go program `panic`s - doesn't it also cause a non-zero exit code to be emitted? Why would we prefer to skip deferred code?

dshulyak:
i haven't thought about defer. the reason why i used fatal is to avoid printing stack trace. as for the panic it will execute deferred code only in the goroutine that panicked.

i think the cleanest way would be is to terminate already started services (and goroutines) gracefully and then exit with the initial error using fatal, but it would be too much for this change.

lrettig:
> i haven't thought about defer. the reason why i used fatal is to avoid printing stack trace. as for the panic it will execute deferred code only in the goroutine that panicked.

Why don't we want a stack trace?

dshulyak:
the stack trace is useful for programmer errors. like for example with the layer overflow (we will see which code has led to the issue). but for the error that HTTP server can't start on some port it is just noise.

lrettig:
In cases like this, ideally, we do want deferred funcs to run but we
probably don't need the stack trace. On balance, I'd rather have the stack
trace than skip deferred funcs (so, e.g., temp dirs are cleaned up).

>

dshulyak:
makes sense, deferred code should be preserved. i made some changes to exit cleanly anyway by returning error from initializers

lrettig:
@dshulyak every time you force push on a branch like this, 1. the links in github emails stop working, and 2. I lose the ability to see incremental changes since the last time I viewed changes. We try to avoid force pushing except in case of an emergency, just FYI!

cmd/node/node.go
```diff
@@ -344,9 +351,8 @@ func (app *SpacemeshApp) setupGenesis(state *state.TransactionProcessor, msh *me
 	for id, acc := range conf.InitialAccounts {
 		bytes := util.FromHex(id)
 		if len(bytes) == 0 {
-			// todo: should we panic here?
 			app.log.With().Error("cannot read config entry for genesis account", log.String("acct_id", id))
-			continue
+			return errors.New("id in initial accounts is empty")
```
lrettig: It's a bad idea to log one errmsg here and return another one. Probably best to just return the existing errmsg (which is more informative) and print it at a higher level.

cmd/node/node.go
```diff
@@ -360,8 +366,9 @@ func (app *SpacemeshApp) setupGenesis(state *state.TransactionProcessor, msh *me
 
 	_, err := state.Commit()
 	if err != nil {
-		log.Panic("cannot commit genesis state")
+		app.log.With().Error("cannot commit genesis state", log.Err(err))
```
lrettig: Same comment as above: we usually don't want to print the err _and_ return it, it should be printed once where the error is handled

cmd/node/node.go
```diff
@@ -724,30 +733,34 @@ func (app *SpacemeshApp) HareFactory(ctx context.Context, mdb *mesh.DB, swarm se
...
 }
 
-func (app *SpacemeshApp) startServices(ctx context.Context, logger log.Log) {
+func (app *SpacemeshApp) startServices(ctx context.Context, logger log.Log) error {
 	app.layerFetch.Start()
 	go app.startSyncer(ctx)
 
 	if err := app.hare.Start(ctx); err != nil {
-		logger.Panic("cannot start hare")
+		logger.With().Error("cannot start hare", log.Err(err))
```
lrettig: Print or return the error, not both

cmd/node/node.go
```diff
@@ -1098,9 +1125,13 @@ func (app *SpacemeshApp) Start(*cobra.Command, []string) {
...
-	<-ctx.Done()
+	select {
+	case <-ctx.Done():
+	case err = <-pprofErr:
+	}
 	events.ReportError(events.NodeError{
 		Msg:   "node is shutting down",
 		Level: zapcore.InfoLevel,
 	})
+	return err
```
lrettig: What does this return in the `<-ctx.Done()` case?

cmd/node/node.go
```diff
@@ -96,15 +95,23 @@ var Cmd = &cobra.Command{
...
 		app := NewSpacemeshApp()
-		defer app.Cleanup(cmd, args)
-
-		err := app.Initialize(cmd, args)
+		starter := func() error {
+			if err := app.Initialize(cmd, args); err != nil {
+				log.With().Error("Failed to initialize node.", log.Err(err))
+				return err
+			}
+			// This blocks until the context is finished
```
lrettig:
```suggestion
			// This blocks until the context is finished or until an error is produced
```

lrettig: Are you sure this won't print a misleading error message if the app exits cleanly?

lrettig (changes requested):
As I indicated a few times in individual comments, I don't think we should print an error and then return it and print it again at a higher level. It's almost always better to _just_ return the error and print it where it's handled.

cmd/node/node.go
dshulyak: nil

cmd/node/node.go
dshulyak: node exits without error only if ctx is terminated

dshulyak:
bors try

cmd/node/node.go
lrettig: Yes, I see that now, but in this case `err` still holds the value it got on line 1113 above, right? It would be good to make this safer and more explicit by using a different variable name, or else using `err := ...` syntax (rather than `err = ...`). I say "safer" because someone could add code above this that sets `err` to something else, later.

lrettig: To make this more readable, consider something like:

```
	reportErr := func() {
		events.ReportError(events.NodeError{
			Msg:   "node is shutting down",
			Level: zapcore.InfoLevel,
		})
	}
	select {
	case <-ctx.Done():
		reportErr()
		return nil
	case err := <-pprofErr:
		reportErr()
		return err
	}
```

lrettig (approved):
Good use of wrapped errors! See one suggested change.

dshulyak:
bors merge

Issue #2082, opened 2020-07-27
LayerID uint64 -> uint32


lrettig:
See https://github.com/spacemeshos/api/issues/88 for background. With 32 bits and five minute layers we get nearly 41,000 years. Even if the layer time were to speed up to one minute, it would still be enough for ~8100 years.

For the record here is the math:

```
# max_uint32 divided by number of five-minute layers per year
In [3]: 2**32 / (60 / 5. * 24 * 365.25)
Out[3]: 40829.79025020914

# max_uint32 divided by number of minutes per year
In [4]: 2**32 / (60 * 24 * 365.25)
Out[4]: 8165.958050041828
```

PR #2489, opened 2021-06-29
[Merged by Bors] - common/types: add safe sub/add/mul for LayerID type

### LINKED ISSUE #2414: LayerID type should protect against overflow
## Description
Int types in golang [do not panic on overflow](https://stackoverflow.com/questions/54605650/arithmetic-on-uint8-int8) and instead wrap around. Internally, our LayerID type is stored as a uint (see also #2082). In practice this means that if you inadvertently subtract layer x from layer y, where x > y, you'll get a very large layer as a result (the opposite could happen if adding two very large layers, but this is less likely).

It is of course possible to add checks to every bit of layer arithmetic, but this is dangerous and error-prone. It would be safer if the LayerID type did not wrap an integer type, as it does now, and therefore did not natively allow arithmetic operations. Instead, the type should implement basic safe math functions such as Add, Sub, Comp, Inc, Dec, etc. that prevent overflow and wrap around (or panic if it's attempted).

## Affected code
common/types/block.go LayerID type, and everything that relies on it
many places that perform layer arithmetic, e.g., the new Tortoise code

This issue appears in commit hash: e36cb051371df4e3a35c4b3e91f9e6fe5e49e527

dshulyak:
## Motivation
closes: https://github.com/spacemeshos/go-spacemesh/issues/2414

## Changes
LayerID Add/Sub/Mul operations will panic on wraparound. 

## Test Plan
Unit tests

## TODO

## DevOps Notes


antonlerner (approved):

common/types/block.go
```diff
@@ -68,9 +68,33 @@ func GetEffectiveGenesis() LayerID {
 	return LayerID(atomic.LoadInt32(&EffectiveGenesis))
 }
 
-// Add returns the LayerID that's layers (the param passed into this method) after l (this LayerID).
-func (l LayerID) Add(layers uint16) LayerID {
-	return LayerID(uint64(l) + uint64(layers))
+// Add layers to the layer. Panics on wraparound.
+func (l LayerID) Add(layers LayerID) LayerID {
```
noamnelke: `Add` (and all other new methods here) should accept a `uint16` (which should be our standard for layer-deltas). This is similar to `Time.Add()` accepting a duration and not another time ([see here](https://golang.org/pkg/time/#Time.Add)).

noamnelke (changes requested):

common/types/block.go
dshulyak: the thing is that in some places two LayerIDs were added together. i didn't want to cast uint64 to uint16 just to perform "safe" arithmetic

noamnelke:
In addition to the comment I added, which should save us a lot of casting, the purpose of the original issue was to prevent the possibility of overflows. The idea was to change the `LayerID` type from an alias of a `uint64` to a struct that wraps a `uint64` (that should not be exported). The way it's currently implemented doesn't prevent future developers to misuse the type and do calculations on it directly.

noamnelke:
This also requires implementing `Before(LayerID)` or `After(LayerID)` methods (I think one of them is enough).

common/types/block.go
dshulyak: - https://github.com/spacemeshos/go-spacemesh/pull/2489/files#diff-c420f14b462324bfc31c771f62e5f9014f50251ba32793485e6bad0f307478bfR143
- https://github.com/spacemeshos/go-spacemesh/pull/2489/files#diff-de1c771222d45ea483c2a6212af7b40473ea6c8c814ecec398deae0ffaddb0d7R231

dshulyak: @noamnelke to make it safe to use i need to store those values as uint16 then

noamnelke: `bufferSize` is currently an `int`

https://github.com/spacemeshos/go-spacemesh/blob/d16d26f4c68062b06ce4f5b2c8281034e2e89f6b/hare/hare.go#L71

`HDist` is currently a `LayerID`

https://github.com/spacemeshos/go-spacemesh/blob/d16d26f4c68062b06ce4f5b2c8281034e2e89f6b/tortoise/verifying_tortoise.go#L41

I think both should become `uint16`.

Alternatively, we could create a special type for this, like Go's datetime's `Duration`, but I think that's overkill and not worth the effort.

noamnelke: Actually, I think both "layer duration" and the underlying layer itself (the number inside the struct) should become `uint32`. That would give us enough range for ~40,000 years (assuming 5 minute layers) and save us the need to convert types.

noamnelke: While I have high aspirations for Spacemesh, I don't believe this code will run unmodified for more than 40k years 😉

lrettig: Layers should be uint32. See #2082.

common/types/block_test.go
```diff
@@ -93,3 +93,33 @@ func TestBytesToNodeID(t *testing.T) {
 	_, err = BytesToNodeID(y[:])
 	r.Error(err, "Expected error converting too-long byte array to NodeID")
 }
+
+func TestLayerIDWraparound(t *testing.T) {
+	max := LayerID(^uint64(0))
```
lrettig: Consider using this for readability:

```suggestion
	max := LayerID(math.MaxUint64)
```

lrettig (changes requested):
Agree with Noam's comments. It might be nice to introduce an explicit `LayerDuration` type of uint16. It should not be possible to directly perform arithmetic operations on the LayerID type. This may also be a good opportunity to take care of #2082.

noamnelke:
Making the underlying number in `LayerDuration` the same type as the underlying number in `LayerID` will save us some conversions. I can't think of any good reason to make `LayerDuration` any smaller than a `uint32` (it won't save memory due to alignment, anyway).

In case you're asking "on a 64-bit system a `uint64` and `uint32` also take up the same memory, so why not go with `uint64`?" - it's because unlike `LayerDuration`, the `LayerID` is serialized and included in almost every message on the wire and in the mesh and there is potential for significant space savings there.

Let's just make both `uint32`.

dshulyak:
To be honest feels like code becomes distinctly worse for almost no benefit. I am referring to the suggested change for the LayerID type. I would rather keep it as a simple alias for uint32 and rely on developers sanity to use safe math functions.

noamnelke:
What do you feel that makes the code after this change worse? The need to use `.Equals()`, `.Before()` and `.After()` instead of `=`, `<` and `>`? What else would change?

dshulyak:
i disliked that i had to cast manually to LayerID where previously the type was just inferred from the definition. and the loops became a bit harder to read because of the After/Before. but I reviewed the code and from the second look it doesn't seem as bad as it was initially

lrettig:
I think Noam is right here, I'm for explicitness and safety. This is an easy and obvious potential source of bugs. I've found a number of cases of unsafe layer arithmetic already.

dshulyak:
I found one setback with xdr codec. It works using reflection and doesn't have support for extensions (e.g. fields that implement BinaryMarshal/BinaryUnmarshal methods). It is not hard to add such feature. However, for now I had to make Value of the LayerID public (see commit https://github.com/spacemeshos/go-spacemesh/pull/2489/commits/b667549c39754eb7fb4ee2f37fc8dec3cc4bacb2) which somewhat defeats the purpose of this refactoring.

Otherwise, this change is ready for review. I changed EpochID to uint32 as well to avoid casting between uint32 and uint64 in several places. Changed "layer duration" to uint32, but without the LayerDuration type.

dshulyak:
bors try

common/types/block.go
```diff
@@ -39,47 +40,107 @@ func (id BlockID) AsHash32() Hash32 {
 	return Hash20(id).ToHash32()
 }
 
-var layersPerEpoch int32
+var (
+	layersPerEpoch uint32
+	// effectiveGenesis marks when actual blocks would start being crated in the network, this will take account the first
```
lrettig:
```suggestion
	// effectiveGenesis marks when actual blocks would start being created in the network, this will take into account the first
```

common/types/block.go
```diff
@@ -39,47 +40,107 @@ func (id BlockID) AsHash32() Hash32 {
...
+func (l LayerID) Before(other LayerID) bool {
+	return l.Value < other.Value
+}
+
+// After returns true if this layer is higher than the other.
+func (l LayerID) After(other LayerID) bool {
+	return l.Value > other.Value
+}
+
+// Duration returns different between current and other layer.
```
lrettig:
```suggestion
// Duration returns the difference between current and other layer.
```

common/types/block.go
```diff
@@ -39,47 +40,107 @@ func (id BlockID) AsHash32() Hash32 {
...
+func (l LayerID) Duration(other LayerID) uint32 {
```
lrettig: Can we call this `Difference` or maybe `LayersBetween` instead?

common/types/block.go
```diff
@@ -39,47 +40,107 @@ func (id BlockID) AsHash32() Hash32 {
...
+	return l.Value - other.Value
 }
 
 // Field returns a log field. Implements the LoggableField interface.
-func (l LayerID) Field() log.Field { return log.Uint64("layer_id", uint64(l)) }
+func (l LayerID) Field() log.Field { return log.Uint32("layer_id", l.Value) }
+
+// String returns string representation of the layer id numeric value.
+func (l LayerID) String() string {
+	return strconv.FormatUint(uint64(l.Value), 10)
```
lrettig, line 154: Why convert to uint64 here?

hare/broker.go
```diff
@@ -57,13 +57,13 @@ type Broker struct {
 	inbox          chan service.GossipMessage
 	queue          priorityq.PriorityQueue
 	queueChannel   chan struct{} // used to synchronize the message queues
-	syncState      map[instanceID]bool
```
lrettig, line 40: Are we sure we want to eliminate the hare instanceID type entirely? @noamnelke what do you think?

lrettig (dismissed):
Just a preliminary review for now

common/types/block.go
dshulyak: i thought that duration fits because it was used in the previous messages, but the difference sounds fine to me too

common/types/block.go
dshulyak: strconv doesn't provide an alternative function that accepts uint32. I can format with fmt.Sprintf but it converts to uint64 internally too

hare/broker.go
dshulyak: this type didn't provide any utility, the only method that it had was Bytes() and it was doing exactly the same thing as LayerID.Bytes(). but now I also had to cast it to LayerID every time I had to modify it, and then back to instanceID to store it

noamnelke: I'm with @dshulyak on this. I've hated that type for ages and wanted to eliminate it, too. Other than adding unjustified code complexity, I find that it complicates the mental model of what the Hare is doing. Fewer concepts to keep in mind == better.

common/types/block.go
```diff
@@ -39,47 +40,107 @@ func (id BlockID) AsHash32() Hash32 {
...
+// NewLayerID creates LayerID from uint32.
+func NewLayerID(value uint32) LayerID {
+	return LayerID{Value: value}
+}
+
+// LayerID is representing a layer number. Zero value is safe to use, and means 0.
+// Internally it is a simple wrapper over uint32 and should be considered immutable
+// the same way as any integer.
+type LayerID struct {
+	Value uint32
```
noamnelke, line 76: Let's add a comment on this field explaining why it's exported and advising developers not to modify it directly.

noamnelke (changes requested):
I wish there was a cleaner solution to the XDR issue, but I can't think of one.

I wonder if a struct with a single field is serialized the same way as just a field. If the serialization changes we may want to reconsider this change. @dshulyak do you know? Can you look it up?

dshulyak:
@noamnelke yes xdr encodes it in the same way. in both cases, LayerID is encoded as 4 bytes big endian. i double-checked when debugging Test_VrfSignVerify. Note, there is still a minor tweak in that test, but that's because uint64 was changed to uint32.

antonlerner (approved):

events/reporter.go
```diff
@@ -64,7 +64,7 @@ func ReportNewActivation(activation *types.ActivationTx) {
 
 	Publish(NewAtx{
 		ID:      activation.ShortString(),
-		LayerID: uint64(activation.PubLayerID.GetEpoch()),
+		LayerID: uint32(activation.PubLayerID.GetEpoch()),
```
noamnelke: @antonlerner is there a special reason that we report the epoch ID in the `LayerID` field? Maybe we should change the field name to `EpochID`?

(not related to this PR, it just drew my attention to it)

events/events.go
```diff
@@ -180,7 +180,7 @@ func (RewardReceived) GetChannel() ChannelID {
 type AtxCreated struct {
 	Created bool
 	ID      string
-	Layer   uint64
+	Layer   uint32
```
noamnelke: In practice this is the `Epoch`, not `Layer` - @antonlerner should we rename this field?

events/events.go
```diff
@@ -120,7 +120,7 @@ func (ValidBlock) GetChannel() ChannelID {
 // NewAtx signals that a new ATX has been received
 type NewAtx struct {
 	ID      string
-	LayerID uint64
+	LayerID uint32
```
noamnelke: In practice this is the `EpochID`, not `LayerID` - @antonlerner should we rename this field?

hare/algorithm.go
```diff
@@ -462,7 +463,7 @@ func (proc *consensusProcess) processMsg(ctx context.Context, m *Msg) {
 	proc.WithContext(ctx).With().Debug("processing message",
 		log.String("msg_type", m.InnerMsg.Type.String()),
 		log.Int("num_values", len(m.InnerMsg.Values)))
-	metrics.MessageTypeCounter.With("type_id", m.InnerMsg.Type.String(), "layer", strconv.FormatUint(uint64(m.InnerMsg.InstanceID), 10), "reporter", "processMsg").Add(1)
+	metrics.MessageTypeCounter.With("type_id", m.InnerMsg.Type.String(), "layer", strconv.FormatUint(uint64(m.InnerMsg.InstanceID.Uint32()), 10), "reporter", "processMsg").Add(1)
```
```suggestion
	metrics.MessageTypeCounter.With("type_id", m.InnerMsg.Type.String(), "layer", m.InnerMsg.InstanceID.String(), "reporter", "processMsg").Add(1)
```

hare/broker.go
```diff
@@ -236,7 +235,10 @@ func (b *Broker) eventLoop(ctx context.Context) {
 				log.Int("queue_length", len(b.inbox)))
 
 			msgInstID := hareMsg.InnerMsg.InstanceID
-			metrics.MessageTypeCounter.With("type_id", hareMsg.InnerMsg.Type.String(), "layer", strconv.FormatUint(uint64(msgInstID), 10), "reporter", "brokerHandler").Add(1)
+			metrics.MessageTypeCounter.With(
+				"type_id", hareMsg.InnerMsg.Type.String(),
+				"layer", strconv.FormatUint(uint64(msgInstID.Uint32()), 10),
```
noamnelke:
```suggestion
				"layer", msgInstID.String(),
```

lrettig:
> I wish there was a cleaner solution to the XDR issue, but I can't think of one.

Here are a few options:

- Make the Layer struct wrap an (exported) internal struct that stores the actual layer number. This adds one layer of indirection and makes it a bit less likely that a developer modifies the data directly.
- Use custom [encoders](https://pkg.go.dev/github.com/davecgh/go-xdr/xdr2?utm_source=godoc#NewEncoder) and [decoders](https://pkg.go.dev/github.com/davecgh/go-xdr/xdr2?utm_source=godoc#NewDecoder) for the Layer type - they can marshal and unmarshal unexported struct members
- Fork the xdr repo and make a slight change to marshal unexported fields - it looks like it would be a [small change](https://github.com/davecgh/go-xdr/blob/e6a2ba005892b6a5b27cb5352f64c2e96942dd28/xdr/encode.go#L437-L441)

I prefer the second option

dshulyak:
> Make the Layer struct wrap an (exported) internal struct that stores the actual layer number. This adds one layer of indirection and makes it a bit less likely that a developer modifies the data directly.

@lrettig can you elaborate, please? I am not sure that I follow. Let's say LayerID is stored like

```go
type Block struct {
  Layer LayerID
}
```

Then we pass it to xdr encoder. When the wrapping will occur?

> Use custom encoders and decoders for the Layer type - they can marshal and unmarshal unexported struct members

LayerID is usually not encoded separately. it is always a part of some bigger struct. Encoder recursively traverses the struct and encodes fields of the struct according to internal rules. I don't see how we can use encoder for our use case, maybe I am missing something

> Fork the xdr repo and make a slight change to marshal unexported fields - it looks like it would be a small change

The common solution would be to check for an interface like BinaryMarshal/BinaryUnmarshal if we are talking about the fork. Encoding unexported fields may result unexpected behavior for other structs.

noamnelke:
Encoding unexported fields is not an option - we rely on this behavior to store data that we don't want serialized on some types.

> ...check for an interface like BinaryMarshal/BinaryUnmarshal...

I like this solution from a clean code perspective, but I wonder if forking the XDR lib wouldn't defeat the purpose of using "battle tested" XDR instead of rolling our own encoding? cc: @antonlerner

hare/hare.go
```diff
@@ -105,10 +105,9 @@ func New(conf config.Config, p2p NetworkService, sign Signer, nid types.NodeID,
 	h.networkDelta = time.Duration(conf.WakeupDelta) * time.Second
 	// todo: this should be loaded from global config
 	h.bufferSize = LayerBuffer // XXX: must be at least the size of `hdist`
-	h.lastLayer = 0
 	h.outputChan = make(chan TerminationOutput, h.bufferSize)
-	h.outputs = make(map[types.LayerID][]types.BlockID, h.bufferSize) // we keep results about LayerBuffer past layers
-	h.factory = func(conf config.Config, instanceId instanceID, s *Set, oracle Rolacle, signing Signer, p2p NetworkService, terminationReport chan TerminationOutput) Consensus {
+	h.outputs = make(map[types.LayerID][]types.BlockID, h.bufferSize) //  we keep results about LayerBuffer past layers
```
noamnelke:
```suggestion
	h.outputs = make(map[types.LayerID][]types.BlockID, h.bufferSize) // we keep results about LayerBuffer past layers
```

antonlerner:
I think we shouldn't fork the XDR just for this implementation, as Noam said we chose to rely on this code because it is battle tested. also generally I support moving to .x files which will make xdr encoding / decoding much more efficient so this could be done as part of that change #2545. 
Dmitry also is right here, we cannot write our custom marshalling methods because we use the generic xdr encoder that uses reflection to serialise primitive types.
IMO there is no better solution at the moment for the time we want to spend on this task

lrettig:
> LayerID is usually not encoded separately. it is always a part of some bigger struct. Encoder recursively traverses the struct and encodes fields of the struct according to internal rules. I don't see how we can use encoder for our use case, maybe I am missing something

Are you sure that, if we specify a custom encoder/decoder for the Layer/LayerID type, it won't be triggered (during the recursive traversal) when a struct it's embedded in is marshalled?

dshulyak:
> Are you sure that, if we specify a custom encoder/decoder for the Layer/LayerID type, it won't be triggered (during the recursive traversal) when a struct it's embedded in is marshalled?

it doesn't look like that you can register an Encoder for a type. You can just use it to encode fields separately

noamnelke (commented):
Partial review - just releasing what I have so far.

events/events.go
antonlerner: Yes

events/events.go
antonlerner: Yes

events/reporter.go
antonlerner: Yes

dshulyak:
@noamnelke @lrettig let me know if you want to finish review before merge. otherwise I will try to merge it today

dshulyak:
bors merge

dshulyak:
bors merge

noamnelke (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
flake in systemtest-p2p. can't be related to this PR

dshulyak:
bors merge

PR #2490, opened 2021-06-29
[Merged by Bors] - p2p/peers: ensure that peer events are received by peer manager

### LINKED ISSUE #2385: Drop peers after failing to send messages to them

dshulyak:
## Motivation
Sometimes connections are removed from cpool but peers are not removed from peers store that is formed by listening to notifications. Notifications were non-blocking, so in certain goroutine scheduling scenario it was possible that peers event listener will miss some events. I am not totally sure that this was the case, but I don't see any other problem in this code. 

closes: https://github.com/spacemeshos/go-spacemesh/issues/2385

## Changes
- make peer notifications blocking with a buffer that can fit a potential burst of connected or expired peers 
- don't copy and stop shuffling peers when GetPeers is called, this adds a significant pressure on gc for nothing
  there are 3 cases when GetPeers is used, in two of them all peers are receiving messages, and in the other one random peer is selected from the array

## Test Plan
TODO

## TODO

## DevOps Notes


dshulyak:
@antonlerner take a look please, i will experiment a bit with this change on real environment before moving forward. but maybe you will have some other suggestions

p2p/switch.go
```diff
@@ -677,7 +677,7 @@ func (s *Switch) publishNewPeer(peer p2pcrypto.PublicKey) {
 	for _, p := range s.newPeerSub {
 		select {
 		case p <- peer:
-		default:
+		case <-s.shutdownCtx.Done():
```
antonlerner, line 680: return here?

antonlerner:
Good catch of this edge case
Another thing is to remove the duplicate peer lists we have throughout the p2p code, but it is a lot to refactor so I suggest we test this fix and see if it reduces the p2p errors seen

dshulyak:
> Another thing is to remove the duplicate peer lists we have throughout the p2p code, but it is a lot to refactor so I suggest we test this fix and see if it reduces the p2p errors seen

@antonlerner can you show me what you have in mind? i thought that all calls were proxied to Peers object

antonlerner:
> > Another thing is to remove the duplicate peer lists we have throughout the p2p code, but it is a lot to refactor so I suggest we test this fix and see if it reduces the p2p errors seen
> 
> @antonlerner can you show me what you have in mind? i thought that all calls were proxied to Peers object

this is the duplication I'm talking about, there are inbound and outbound lists of peers in `switch.go`, IMO there is no real need to keep another list in `peers.go`

antonlerner (approved):

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
build merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

Issue #2478, opened 2021-06-24
Hare inbound gossip message processing delays


lrettig:
## Description
Nodes are taking as long as a minute to receive inbound Hare gossip messages and report them as valid (after which they're gossiped on to other peers).

I can't tell exactly what's going on under the hood without full debug logs. It could be an issue with the Hare broker priority queue, message validation, or something else. I suspect it's happening during the Hare broker `eventLoop`.

## Environment
Please complete the following information:
 - OS: Linux
 - Node Version: v0.1.41

## Additional Resources 
![image](https://user-images.githubusercontent.com/3316532/123186844-02e1bb00-d467-11eb-9174-803aad81651d.png)
from https://testnet-kibana.spacemesh.io/goto/29beacaf0fe3b5c3757deed8d6f8788a

lrettig:
Dug a bit deeper using debug logs:

![image](https://user-images.githubusercontent.com/3316532/123360706-c59b2d00-d53b-11eb-9b82-0635e6b20aae.png)
full data: 
[hare backlog (1).csv](https://github.com/spacemeshos/go-spacemesh/files/6713146/hare.backlog.1.csv)
[hare backlog.csv](https://github.com/spacemeshos/go-spacemesh/files/6713147/hare.backlog.csv)
from https://testnet-kibana.spacemesh.io/goto/91ecf641e57a09491279144a2f4d0d67

The hare broker is getting stuck somewhere in here:

https://github.com/spacemeshos/go-spacemesh/blob/3645848f68a00d8b96f348402d61dbdb97b2db09/hare/broker.go#L232-L269

It seems like it's either the eligibility validator or the `newMsg` call. Both of these rely on the Hare oracle, which does use a mutex. And there's one or two other parallel hare goroutines also using the Hare oracle from the main consensus process code in algorithm.go, including:

https://github.com/spacemeshos/go-spacemesh/blob/3645848f68a00d8b96f348402d61dbdb97b2db09/hare/algorithm.go#L286-L303

But I can't find any blocking calls that would cause it to get stuck for as long as 60 secs.

dshulyak:
[...]

PR #2496, opened 2021-07-02
Testnet hotfix: Warm active set cache for hare

lrettig:
## Motivation
Testnet hotfix workaround for #2478

Unclear yet whether we want to apply this workaround to develop, or address the underlying issue in another way

## Changes
- When hare gets active set, it checks if it's near a safe epoch boundary, and if so, it warms the cache for the next layer/safe epoch ahead of time

## Test Plan
- Adds regression test

## TODO

## DevOps Notes

### DIFF
--- a/hare/eligibility/oracle_test.go
+++ b/hare/eligibility/oracle_test.go
@@ -560,17 +560,22 @@ func TestOracleActiveSetCacheWarming(t *testing.T) {
 		return mp, nil
 	}
 
-	firstLayer := types.LayerID(73)
+	// set firstLayer to two layers before a safe epoch boundary
+	r.Equal(2, cacheWarmingDistance)
+	firstLayer := types.LayerID(72)
 	sl1 := roundedSafeLayer(firstLayer, types.LayerID(o.cfg.ConfidenceParam), o.layersPerEpoch, types.LayerID(o.cfg.EpochOffset))
 	r.Equal(60, int(sl1))
 	sl2 := roundedSafeLayer(firstLayer.Add(1), types.LayerID(o.cfg.ConfidenceParam), o.layersPerEpoch, types.LayerID(o.cfg.EpochOffset))
 	r.Equal(60, int(sl2))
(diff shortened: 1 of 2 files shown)


hare/eligibility/oracle_test.go
```diff
@@ -542,3 +543,100 @@ func TestOracle_Eligible2(t *testing.T) {
...
+	types.SetLayersPerEpoch(defLayersPerEpoch)
+	o := New(&mockValueProvider{1, nil}, nil, nil, nil, defLayersPerEpoch, genActive, mockBlocksProvider{}, cfg, log.NewDefault(t.Name()))
+	mc := newMockCacher()
+	o.activesCache = mc
+	mapSize := 9
+	mp := createMapWithSize(mapSize)
+	callCount := int32(0)
+	o.getActiveSet = func(ctx context.Context, epoch types.EpochID, blocks map[types.BlockID]struct{}) (map[string]struct{}, error) {
+		// simulate slow IO
+		time.Sleep(time.Second)
```
dshulyak, line 558: what sleep is for? test passes without sleep as well
do you expect something to happen during the sleep? is it possible to use synchronization in some way e.g.

```go
syncPoint := make(chan struct{})
go func() {
    syncPoint <- struct{}{} // before IO
    // getActiveSet
    syncPoint <- struct{}{} // after IO
}()
// some setup code
<-syncPoint // allow getActiveSet to execute
// more setup code
<-syncPoint
```

dshulyak (commented):
it seems to depend on another component to trigger cache warmup at the correct time. i don't know if this is ok to rely on this fact for correctness, but even if that works it is a fragile dependency. wouldn't it be better to use either:
1. notifications over channels and trigger cache warmup when safe layer is observed
2. since the whole protocol relies on time synchrony it should be fine to trigger cache warmup based on time

lrettig:
@dshulyak thanks for taking a look. I agree this solution is not bulletproof and is not completely production-grade. We'll want to do something more robust to solve this issue for develop branch. This is a rough-and-ready hotfix for the testnet, and it does appear to be working.

Another reason this isn't a bulletproof solution: it'll only work for nodes that are already synced and running hare continuously.

dshulyak (approved):
@lrettig great, i definitely don't mind if it will get merged for testnet

PR #2518, opened 2021-07-06
[Merged by Bors] - Remove genesis-total-weight

noamnelke:
## Motivation
Now that we have a fallback from the in-consensus active set for the Hare, we can have a proper genesis flow without pre-configuring how many miners or how much weight is allocated.

## Changes
Remove the `GenesisTotalWeight` config (formerly `GenesisActiveSet`) and all code that uses it. The Tortoise active set will be used immediately after genesis.

## Test Plan
Existing unit- and system tests. The genesis flow is already well tested, the logic just changed now, but behavior remains identical.

## DevOps Notes

  Yes, it does. The `genesis-total-weight` param should be removed.
  It also introduces the `tortoise-beacon-wait-after-epoch-start-ms`.


noamnelke:
bors try

noamnelke:
bors try

hare/eligibility/oracle.go
```diff
@@ -408,11 +396,6 @@ func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID) (map[st
 		log.FieldNamed("target_layer_epoch", targetLayer.GetEpoch()))
 	logger.Debug("hare oracle getting active set")
 
-	// we can't read blocks during genesis epochs as there are none
-	if targetLayer.GetEpoch().IsGenesis() {
```
lrettig, line 412: consider adding a similar conditional on [this line](https://github.com/spacemeshos/go-spacemesh/pull/2518/files#diff-a2fa964ab3766e7996ed48abf1f4db1a8f864e897bf510e899fc5dd3d4d131d0R473) so that a warning is not printed during genesis (maybe just INFO)

hare/eligibility/oracle.go
```diff
@@ -521,10 +504,6 @@ func (o *Oracle) IsIdentityActiveOnConsensusView(ctx context.Context, edID strin
 	}()
 	actives, err := o.actives(ctx, layer)
 	if err != nil {
-		if err == errGenesis { // we are in genesis
```
lrettig, line 524: Where is `errGenesis` defined? Is it still used anywhere? Do we still need it?

lrettig (approved):
lgtm, added a couple of questions/ideas for minor improvements

for the record I do not fully understand the context here--have we finalized the genesis flow for block production and hare eligibility? is it documented somewhere? in other words: hare falls back on tortoise active set during genesis, but how is _that_ bootstrapped in the first place?

hare/eligibility/oracle.go
noamnelke: Forgot to remove it, thanks for noticing - took care of it now.

hare/eligibility/oracle.go
noamnelke: Your link is not to anything specific so I don't really understand what you're referring to... where would a warning be printed in genesis? IIUC nothing in the code should call `actives()` in genesis anyway, so this code was zombie code before this change. Maybe I missed something - I can look at the logs to check.

noamnelke:
bors try

noamnelke:
@lrettig I removed significantly more code since your review (found out that the active set from view calculation was never used and it had a lot of supporting code like a cache and mutex). Could you take another look?

noamnelke:
bors try

lrettig (approved):

hare/eligibility/oracle.go
lrettig: I made the change myself, take a look: cbf0623

noamnelke:
@lrettig now I understand what you were referring to - thanks for the commit 😉

noamnelke:
bors try

noamnelke:
bors merge

noamnelke:
@lrettig 
> hare falls back on tortoise active set during genesis, but how is _that_ bootstrapped in the first place?

The Tortoise active set is simply the set of all ATXs received in the previous epoch. During epoch 0, miners start working on their ATX and they publish it during epoch 1. Blocks are first published in epoch 2, which is also when the first Hare runs. If it falls back to the Tortoise active set, it means it will use all ATXs published in epoch 1 and received by the node. If ATXs are published late, less than 30 seconds before the Hare takes a snapshot of the active set, there could be a Byzantine failure, where some nodes see one active set while others see another. IIRC, we decided that this is ok, despite not being extremely unlikely, because the worst that can happen is that we immediately go into self-healing, which is the next best alternative for the genesis flow anyway. We're being optimistic and trying the fast track, if it fails, we fall back to slow-but-guaranteed success - after genesis, as in any other time.

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

noamnelke:
bors merge

PR #2521, opened 2021-07-06
[Merged by Bors] - Reduce API unit tests duration

### LINKED ISSUE #2398: Reduce node_test (API test) test run time
## Description
currently API tests take 130 seconds, this could probably be reduced dramatically

nkryuchkov:
## Motivation

Closes #2398 

## Changes

- Reduce duration of all API unit tests to ~2 seconds

## Test Plan

- API unit tests

## DevOps Notes


api/grpcserver/grpcserver_test.go
```diff
@@ -466,7 +466,28 @@ func launchServer(t *testing.T, services ...ServiceAPI) func() {
...
+		select {
+		case <-timer.C:
+		case <-jsonStarted:
+		}
+	case <-jsonStarted:
+		select {
+		case <-timer.C:
+		case <-grpcStarted:
+		}
+	}
```
dshulyak, line 479: it doesn't matter which one we will select first? in that case you could iterate over array with channels. it will make a bit prettier i believe
```go
for _, start := range []chan struct{}{json, grpc} {
  select {
  ...
  }
}
```

nkryuchkov: @dshulyak I agree that this doesn't look pretty. 

No, the order does not matter. What we need is to wait until both channels are closed or until a 3-second timeout. I'm not sure I'm following your thought about `range`. Could you please add the implementation of the `select` clause body?

dshulyak: nothing changes in the select
```go

	for _, ch := range []chan struct{}{grpc, json} {
		select {
		case <-ch:
		case <-timer.C:
			panic("timeout")
		}
	}
```

dshulyak:
one note regarding the change in general. i had issues in the past with services being not immediately available when the listener socket is opened. usually, the error-proof practice was to poke some endpoint (like /ping) until it returns a successful response. but the test passed so I guess it is fine

api/grpcserver/grpcserver_test.go
nkryuchkov: Now I see. So many `select`s are not needed indeed. Thanks

nkryuchkov:
@dshulyak I was going to implement it the way you suggested, but way more changes were needed for that. Tests seemed to always pass even with the current implementation, so I left it. If tests fail in the future, I'll improve the checks.

dshulyak (approved):

antonlerner (approved):

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #2532, opened 2021-07-08
[Merged by Bors] - activation: handle retries better if atx building process fails 

dshulyak:
## Motivation
If activation builder process fails for whatever reason it will retry only in the next layer. On the devnet this interval is equal to 200s.

## Changes
Main changes here is handling of two distinct error types:
- ErrATXChallengeExpired. Can be retried immediately with a new challenge
- ErrPoetServiceUnstable. Can be retreied after short delay (hardcoded 5 seconds at the moment).

For now if proof was received, but our challenge is not returned this error is handled as ErrPoetServiceUnstable

Other than that I made some refactoring:
- replaced self-written poet service mock with generated mock to handle expectations better
- replaced the creation of unnecessary goroutines by a channel with a buffer for one item in SubscribeToProofRef
- removed duplicated log (logging error and returning it up the stack where it is logged again)
- updated testify module to 1.7.0 for require.ErrorIs check

## Test Plan
Unit tests. Maybe will run this PR on devnet with chaos-mesh, simulating network issues with poet server, if there will be time.

## TODO

## DevOps Notes


activation/activation.go
```diff
@@ -394,47 +443,82 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
 		err := b.buildNipstChallenge(ctx)
 		if err != nil {
-			b.log.With().Error("failed to build new atx challenge", log.Err(err))
+			return fmt.Errorf("failed to build new atx challenge: %w", err)
+		}
+	}
+
+	if b.pendingATX == nil {
+		var err error
+		b.pendingATX, err = b.createAtx(ctx)
```
antonlerner, line 449: why not handle poet errors here?

antonlerner (approved):

activation/activation.go
dshulyak: if we will retry here we will miss the code block on top that discards stale challenge and builds a new one

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #2534, opened 2021-07-09
[Merged by Bors] - activation: allow to update poet server during runtime

dshulyak:
## Motivation
Allows updating the poet server that is used during ATX building process.

depends on: https://github.com/spacemeshos/go-spacemesh/pull/2532

## Changes
- safe interface for activation.Builder that can be used concurrently with PublishActivationTx. The new server will be used only after current build finishes
- unsafe interface for activation.NiPSTBuilder (that type should not be exported at all, but it requires more refactoring)
- API change. NodeServer.UpdatePoetClient (https://github.com/spacemeshos/api/pull/154)

## Test Plan
Unit tests

## TODO

## DevOps Notes


activation/activation.go
```diff
@@ -195,35 +240,64 @@ func (b *Builder) loop(ctx context.Context) {
...
+			poetRetryTimer.Stop()
+		}
+	}()
+	defer b.log.Info("atx builder is stopped")
 	for {
-		select {
-		case <-b.stop:
-			return
-		default:
+		client := atomic.LoadPointer(&b.pendingPoetClient)
```
antonlerner, line 257: consider encapsulating poet `UpdatePoETServer` within poetClient. As I see it there is no need for this functionality to be part AtxBuilder as it does not use it internally, we can export this functionality to API from poet client wrapper.

dshulyak: this is in AtxBuilder because it controls the concurrency of the execution. We can't update poet just at any time, because It can be used several times during build process, and we know that it is not used only before execution is started.

for example here https://github.com/spacemeshos/go-spacemesh/blob/develop/activation/nipst.go#L162-L174

we first fetch PoetServiceID, then Submit and after that, we are waiting for a proof from a PoetServiceID that we fetched. If it was updated after PoetServiceID is fetched we will just timeout later. Besides that, if we are going to rotate between multiple clients this will be the only safe place to do that as well.

I considered doing some locking in NiPSTBuilder, but it in results in error-prone code. Much safer to assume that we always interact with the same client until an error has occurred.

activation/nipst.go
```diff
@@ -148,6 +148,15 @@ func NewNIPSTBuilder(
 	}
 }
 
+// UpdatePoETProver updates poetProver reference. It should not be executed concurently with BuildNIPST.
```
antonlerner: how do we enforce this rule? I mean,  I assume it will not be uncommon to switch poet servers, perhaps while the node is running
Perhaps we can check if we need to switch clients at the beginning of BuildNIPST?

dshulyak: i don't think that this object should be public at all. it stores some state already so even now if someone will try to BuildNiPST from 2 goroutines the behavior is not safe

dshulyak: actually i will make this method private for this module

antonlerner (approved):

dshulyak:
bors merge

dshulyak:
failed in systemtest-mining

bors merge

dshulyak:
bors merge

PR #2536, opened 2021-07-09
[Merged by Bors] - mesh: fix origin/destination index key to avoid unintentional range queries

### LINKED ISSUE #2530: Duplicated results in MeshService.AccountMeshDataQuery
Repro: 
```bash
grpcurl -d '{"filter" : {"accountId":{"address":"uo8EjA4XCaFWQpCuz1xr6vihXhw="}, "accountMeshDataFlags":1}, "minLayer":{"number":0},"maxResults":0, "offset":0}' api-34.spacemesh.io:443 spacemesh.v1.MeshService.AccountMeshDataQuery
```

Result: 8 results of 2 duplicated transactions.
Expected: 2 results, one for each unique transaction.

dshulyak:
## Motivation
closes: https://github.com/spacemeshos/go-spacemesh/issues/2530

## Changes
- changed keys encoding for origin/destination transaction index to use big endian instead of strconv
if there are multiple keys with '2' in it, such as '2', '20', '200'. utils.BytePrefix (that is used in datatabase iterator) won't terminate on key '2', but instead scan all of them. For '20' it will scan '20', '200' and so on.

I changed it to big endian specifically (and not little endian) because in future we may use it for range queries instead of iterating over all layers and making point queries.

## Test Plan
unit tests

## TODO

## DevOps Notes


antonlerner (approved):

dshulyak:
bors merge

PR #2538, opened 2021-07-09
Testnet 135 candidate

lrettig:
Candidate for next testnet

## Changes
- properly close msgserver on exit (#2539)
- add context to msgserver logging
- #2490 
- #2479 
- #2433
- try a range of layers for active set (for hare eligibility) (in case there is one bad layer)


lrettig:
This is no longer relevant as we've retired the 0.1 testnet line

PR #2539, opened 2021-07-09
Properly close message server on exit

lrettig:
## Motivation
Msgserver sometimes hangs on node shutdown

## Changes
Fixes a concurrency issue in how the msgserver is shut down
Also improves logging and adds context

## Test Plan
tbd

## TODO

## DevOps Notes

### DIFF
--- a/p2p/server/msgserver.go
+++ b/p2p/server/msgserver.go
@@ -105,9 +105,15 @@ func NewMsgServer(ctx context.Context, network Service, name string, requestLife
 
 // Close stops the MessageServer
 func (p *MessageServer) Close() {
-	close(p.exit)
-	<-p.closed
-	p.workerCount.Wait()
+	select {
+	case <-p.exit:
+		// channel already closed
+		// (since nothing is ever sent on this channel)
+	default:
+		close(p.exit)
+		<-p.closed
+		p.workerCount.Wait()
+	}
 }
 
 // readLoop reads incoming messages and matches them to requests or responses.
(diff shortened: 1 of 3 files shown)


p2p/server/msgserver.go
```diff
@@ -103,29 +105,31 @@ func NewMsgServer(ctx context.Context, network Service, name string, requestLife
 
 // Close stops the MessageServer
 func (p *MessageServer) Close() {
-	p.exit <- struct{}{}
-	<-p.exit
+	close(p.exit)
```
dshulyak: Close will panic if called repetitively. better to make it noop
```go
select {
  case <-p.exit:
  default:
    close(p.exit)
    <-p.closed
    Wait
}
```

lrettig: this is a clever trick! thanks for the tip.

dshulyak (approved):

p2p/server/msgserver.go
```diff
@@ -103,29 +105,37 @@ func NewMsgServer(ctx context.Context, network Service, name string, requestLife
...
-	p.exit <- struct{}{}
-	<-p.exit
-	p.workerCount.Wait()
+	select {
+	case <-p.exit:
+		// channel already closed
+		// (since nothing is ever sent on this channel)
+	default:
+		close(p.exit)
+		<-p.closed
```
y0sher, line 114: I know this is a draft, though I'm not sure why we need both `closed` and `exit`.. actually since context is passed in, can't we just use `context.Done` for all of that?

dshulyak:
this was addressed in develop. will close

PR #2529, opened 2021-07-07
[Merged by Bors] - Add a small delay after epoch start for sending tortoise beacon messages

nkryuchkov:
## Motivation

Closes #2528 

Epoch start times can differ by a few milliseconds on different nodes. If the tortoise beacon starts sending messages immediately after the epoch start time, some nodes can receive those messages while they are still in a previous epoch, which leads to message verification failures. This PR adds a small configurable delay before the tortoise beacon starts sending messages.

## Changes

- Add a small delay after epoch start for sending tortoise beacon messages

## Test Plan

Manual testing, `TestMultiNode`

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #2543, opened 2021-07-10
Fix tortoise beacon system test in #2518

### LINKED ISSUE #2531: Flaky tortoise beacon system test
## Description

The tortoise beacon system test is flaky and fails sometimes because a beacon in some epoch is equal to `sha256("")`. This likely means that not enough votes were received.

## Steps To Reproduce
1. Run tortoise beacon system test until the error appears

## Expected Behavior

The tortoise beacon system test always passes

## Environment
Commit hash: `662420d`

## Additional Resources 

https://github.com/spacemeshos/go-spacemesh/runs/3009254762

nkryuchkov:
## Motivation

Closes #2531

The tortoise beacon system test fails in #2518 because the removed `genesis-total-weight` flag is used in the test. This PR removes `genesis-total-weight` from the tortoise beacon system test config and applies a fix from #2529 for flaky unit tests. Also, it sets epoch weight to `1` if it equals `0`.

## Changes

- Remove `genesis-total-weight` flag in tortoise beacon system test
- Set epoch weight to `1` if it equals `0`
- Apply fix from #2529

## Test Plan

Tortoise beacon system test

## DevOps Notes


nkryuchkov:
bors try

activation/atxdb.go
```diff
@@ -636,6 +636,11 @@ func (db *DB) GetEpochWeight(epochID types.EpochID) (uint64, []types.ATXID, erro
 		}
 		weight += atxHeader.GetWeight()
 	}
+
+	if weight == 0 {
+		weight = 1
+	}
```
noamnelke: I don't like this change. I assume it's there to prevent division by zero, however it should not be done here, but at the division site. You actually have a check for this condition here:

https://github.com/spacemeshos/go-spacemesh/blob/b341df7b58bc98f0d86b680de0a4c286f0500234/tortoisebeacon/tortoise_beacon.go#L814-L816

Due to this change, the check above (that would actually print an error message) can never catch this condition.

tests/tortoise_beacon/config.yaml
```diff
@@ -84,7 +86,6 @@ client:
     eligibility-epoch-offset: '0'
     post-space: '1024'
     space-to-commit: '1024'
-    genesis-total-weight: '51200'
```
noamnelke, line 87: How did I miss this? 🤦🏻‍♂️

tortoisebeacon/tortoise_beacon.go
```diff
@@ -364,16 +366,22 @@ func (tb *TortoiseBeacon) handleLayer(ctx context.Context, layer types.LayerID)
...
 
-	tb.handleEpoch(ctx, epoch)
+	epochStartTimer := time.NewTimer(tb.proposalDuration + tb.gracePeriodDuration)
+	defer epochStartTimer.Stop()
+
+	select {
+	case <-tb.CloseChannel():
+	case <-epochStartTimer.C:
+		tb.handleEpoch(ctx, epoch)
+	}
```
noamnelke: Based on [the docs](https://golang.org/pkg/time/#Timer.Stop), there's no need to stop a timer in case it already fired. Additionally, if you do have to stop it, the channel should be drained.

```suggestion

	select {
	case <-tb.CloseChannel():
		if !epochStartTimer.Stop() {
			<-epochStartTimer.C
		}
	case <-epochStartTimer.C:
		tb.handleEpoch(ctx, epoch)
	}
```

noamnelke (changes requested):

noamnelke (approved):

PR #2553, opened 2021-07-13
[Merged by Bors] - Buffer proposal messages

### LINKED ISSUE #2549: Buffer tortoise beacon proposal messages from future epochs
## Description

https://github.com/spacemeshos/go-spacemesh/pull/2529#issuecomment-877772213
@antonlerner commented:
> what happens when nodes receive messages before the epoch start? maybe use some buffer to store these messages until the epoch starts for this node as well, from our experience although adding some delay helps sometimes, but at other times the delay differs causing these problems to re occur, add this fix, but consider adding more resilience to code itself

## Affected code

https://github.com/spacemeshos/go-spacemesh/tree/develop/tortoisebeacon

`d0377f9`

nkryuchkov:
## Motivation

Closes #2549 

## Changes

- Buffer tortoise beacon proposal messages from future epochs

## Test Plan

## DevOps Notes


tortoisebeacon/handlers.go
```diff
@@ -22,95 +23,92 @@ const TBFirstVotingProtocol = "TBFirstVotingGossip"
...
-	}
+	return nil
+}
 
-	if !passes {
-		tb.Log.With().Warning("Rejected proposal message which doesn't pass threshold")
-		return nil
-	}
+func (tb *TortoiseBeacon) classifyProposalMessage(m ProposalMessage, atxID types.ATXID, currentEpoch types.EpochID) error {
+	receivedTimestamp := time.Now()
```
dshulyak: shouldn't we use adjusted clock?

nkryuchkov:
bors try

tortoisebeacon/tortoise_beacon.go
```diff
@@ -915,11 +953,15 @@ func (tb *TortoiseBeacon) sendToGossip(ctx context.Context, channel string, data
 		return fmt.Errorf("serializing: %w", err)
 	}
 
-	return tb.net.Broadcast(ctx, channel, serialized)
+	if err := tb.net.Broadcast(ctx, channel, serialized); err != nil {
+		return fmt.Errorf("InterfaceToBytes: %w", err)
```
dshulyak: the error msg is confusing

tortoisebeacon/handlers.go
```diff
@@ -22,95 +23,92 @@ const TBFirstVotingProtocol = "TBFirstVotingGossip"
...
+		gossip:  data,
+	}
 
-	receivedTimestamp := time.Now()
+	go func() {
+		select {
+		case <-tb.CloseChannel():
+		case ch <- extendedMessage:
+		}
+	}()
```
dshulyak: creating a goroutine just to push a message is a bad practice. it is a like an unlimited queue, which shouldn't be used. unlimited queue is probably even cheaper

if you can it is better to think what is a max number of pending messages can be allowed, and then decide what you will do with them (pop last, pop first, or return error for example)

dshulyak (commented):
some of the new lines seem to be a bit random :) it is a style question, but couldn't help myself to ignore them

tortoisebeacon/tortoise_beacon.go
nkryuchkov: Yes, a copy-paste mistake. Thanks

tortoisebeacon/handlers.go
nkryuchkov: @dshulyak What do you mean by the adjusted clock? Actually, the timestamp should be gotten earlier, I implemented it here: https://github.com/spacemeshos/go-spacemesh/pull/2553/commits/ee07b71b510bd57fa4ff397083f6d1ca35cb4d61. But I guess you meant something different?

tortoisebeacon/handlers.go
nkryuchkov: @dshulyak Well, it may look not that great, however, `HandleSerializedProposalMessage` is not run in a separate goroutine (https://github.com/spacemeshos/go-spacemesh/blob/feature/buffer-proposal-messages/p2p/service/gossip_listener.go#L83), so I don't want to block it. The channel is buffered and has some capacity, but I don't want either to drop messages that don't fit the buffer or to make the channel capacity extremely big.

nkryuchkov:
bors try

tortoisebeacon/handlers.go
dshulyak: my bad, i thought that we are adjusting our local clock with ntp, but instead we only check that time.Now is in sync with NTP

tortoisebeacon/handlers.go
dshulyak: i think that you are creating an unlimited queue instead. i am pretty sure that this pattern must be avoided, but it is up to you

in my opinion, correct system behavior would be to block producer (peer that sends messages to us) until the queue has enough capacity to accept the message, do you know if we have a listener for each peer or a global one?

nkryuchkov:
bors try

tortoisebeacon/handlers.go
```diff
@@ -148,16 +147,51 @@ func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalM
 	return nil
 }
 
-func (tb *TortoiseBeacon) isPotentiallyValidProposalMessage(atxTimestamp time.Time, nextEpochStart time.Time, receivedTimestamp time.Time) bool {
+func (tb *TortoiseBeacon) verifyProposalMessage(m ProposalMessage, currentEpoch types.EpochID) (done bool, err error) {
```
antonlerner: why do we need `done` boolean here? seems like in every other option we return an error

tortoisebeacon/handlers.go
```diff
@@ -148,16 +147,51 @@ func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalM
...
+
+	passes, err := tb.proposalPassesEligibilityThreshold(m.VRFSignature, epochWeight)
+	if err != nil {
+		return true, fmt.Errorf("proposalPassesEligibilityThreshold: %w", err)
+	}
+
+	if !passes {
+		tb.Log.With().Warning("Rejected proposal message which doesn't pass threshold")
+
+		return true, nil
```
antonlerner: this case will cause us to print an error log message without actual message

tortoisebeacon/handlers.go
nkryuchkov: I agree that we should try to avoid that. However, blocking a producer may deliver a message late or not deliver it which could affect consensus. As far as I know, we have a global listener which puts messages in queue, and each queue handles its messages in a separate goroutine: https://github.com/spacemeshos/go-spacemesh/blob/feature/buffer-proposal-messages/p2p/switch.go#L510

tortoisebeacon/handlers.go
```diff
@@ -214,20 +248,26 @@ func (tb *TortoiseBeacon) HandleSerializedFollowingVotingMessage(ctx context.Con
...
 }
 
-func (tb *TortoiseBeacon) handleFirstVotingMessage(ctx context.Context, message FirstVotingMessage) error {
+func (tb *TortoiseBeacon) handleFirstVotingMessage(message FirstVotingMessage) error {
 	currentEpoch := tb.currentEpoch()
 	from := message.MinerID
 
 	_, err := tb.atxDB.GetNodeAtxIDForEpoch(from, currentEpoch-1)
-	if err != nil {
+	if errors.Is(err, database.ErrNotFound) {
```
antonlerner, line 286: why do we need to check if miner has created an ATX to verify proposal message? miners who did not create ATX still need to have beacon value right?

tortoisebeacon/handlers.go
```diff
@@ -311,28 +353,34 @@ func (tb *TortoiseBeacon) handleFirstVotingMessage(ctx context.Context, message
...
+func (tb *TortoiseBeacon) handleFollowingVotingMessage(message FollowingVotingMessage) error {
+	// TODO(nkryuchkov): use tb.currentEpoch()
+	// currentEpoch := tb.currentEpoch()
 	currentEpoch := message.EpochID
 	messageRound := message.RoundID
 	from := message.MinerID
 
 	_, err := tb.atxDB.GetNodeAtxIDForEpoch(from, currentEpoch-1)
-	if err != nil {
+	if errors.Is(err, database.ErrNotFound) {
```
antonlerner, line 394: same here

tortoisebeacon/handlers.go
nkryuchkov: It means that message processing is done. If `done == true`, we need to return from the outer function. We may need to return from the outer function even if there's no error, e.g. if a proposal does not pass an eligibility threshold.

tortoisebeacon/tortoise_beacon.go
```diff
@@ -422,7 +417,49 @@ func (tb *TortoiseBeacon) handleEpoch(ctx context.Context, epoch types.EpochID)
...
+	if prevCh, ok := tb.proposalChans[epoch]; ok {
+		select {
+		case <-prevCh:
+		default:
+			close(prevCh)
+		}
+	}
+}
+
+func (tb *TortoiseBeacon) ensureProposalChanExists(epoch types.EpochID) chan extendedProposalMessage {
```
antonlerner: maybe rename to `getOrCreateProposalChannel`

tortoisebeacon/handlers.go
nkryuchkov: There's no error in this case. It means that proposal is less than a threshold, so we don't need to continue handling it. Should proposal and threshold values be printed?

tortoisebeacon/handlers.go
antonlerner: proposal channel `ch` is a buffered channel of 1024 items, so there is some inherent buffering here
I think that since there can be many proposals gossiped it is better to block on writing to this channel until it is available, otherwise we might have a lot of goroutines pending
Lets consider however refactoring the gossip listener code to have a worker pool or some additional buffering

tortoisebeacon/handlers.go
antonlerner: https://github.com/spacemeshos/go-spacemesh/pull/2553/files#diff-c301b2c1645d6588238bfe213b34a50d1ab81819fb56e68857529e9ec87eca3fR86
will return with error without inner error, is this expected? perhaps return it as a different error for consistency?

antonlerner (changes requested):

tortoisebeacon/handlers.go
nkryuchkov: No, it was a mistake. I don't like the signature of `verifyProposalMessage`, changed it

tortoisebeacon/handlers.go
nkryuchkov: It was decided on a research call. Yes, they still need but their proposals cannot be counted

tortoisebeacon/handlers.go
antonlerner: if we are done `handleProposalMessage` will exit with error message anyways so calling function will receive error in any case so `handleProposalMessage` can basically check if error was received and return it in

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try-

nkryuchkov:
bors try

layerfetcher/layers.go
```diff
@@ -241,9 +241,14 @@ func (l *Logic) tortoiseBeaconReqReceiver(ctx context.Context, msg []byte) []byt
 	epoch := types.EpochID(util.BytesToUint32(msg))
 	l.log.WithContext(ctx).Debug("got tortoise beacon request for epoch %v", epoch)
 
-	beacon, ok := l.tbDB.GetTortoiseBeacon(epoch)
-	if !ok {
-		l.log.WithContext(ctx).Warning("cannot find tortoise beacon for epoch %v", epoch)
+	beacon, err := l.tbDB.GetTortoiseBeacon(epoch)
+	if errors.Is(err, database.ErrNotFound) {
+		l.log.WithContext(ctx).Debug("still no tortoise beacon for epoch %v", epoch)
```
antonlerner: Is there a case where a request to a beacon that is not yet calculated can be made?
`ErrNotFound` doesn't necessarily mean that we didn't calculate the beacon yet- it could mean that something is wrong

tortoisebeacon/handlers.go
```diff
@@ -25,92 +26,99 @@ const TBFollowingVotingProtocol = "TBFollowingVotingGossip"
...
+	case ch <- extendedMessage:
+	}
+}
+
+func (tb *TortoiseBeacon) handleProposalMessage(m ProposalMessage, receivedTime time.Time) error {
 	currentEpoch := tb.currentEpoch()
 
 	atxID, err := tb.atxDB.GetNodeAtxIDForEpoch(m.MinerID, currentEpoch-1)
-	if err != nil {
+	if errors.Is(err, database.ErrNotFound) {
```
antonlerner, line 84: I think you need to add the minerID and current epoch to debug log -this will help finding issues if there is a consensus failure

tortoisebeacon/handlers.go
```diff
@@ -25,92 +26,99 @@ const TBFollowingVotingProtocol = "TBFollowingVotingGossip"
...
 		tb.Log.With().Warning("Miner has no ATXs in the previous epoch")
+
 		return nil
 	}
 
-	currentEpochProposal, err := tb.buildProposal(currentEpoch)
 	if err != nil {
-		return fmt.Errorf("calculate proposal: %w", err)
+		return fmt.Errorf("get node ATXID for epoch: %w", err)
```
antonlerner: add minerID

tortoisebeacon/handlers.go
```diff
@@ -25,92 +26,99 @@ const TBFollowingVotingProtocol = "TBFollowingVotingGossip"
...
-	if !ok {
-		// TODO(nkryuchkov): attach telemetry
-		tb.Log.With().Warning("Received malformed proposal message: VRF is not verified",
-			log.String("sender", m.MinerID.Key))
-
-		// TODO(nkryuchkov): add a test for this case
-		return ErrMalformedProposal
+	err = tb.verifyProposalMessage(m, currentEpoch)
+	if errors.Is(err, ErrProposalDoesntPassThreshold) {
+		// not a handling error
```
antonlerner, line 97: log this at least, so that we can debug failures better

tortoisebeacon/handlers.go
```diff
@@ -25,92 +26,99 @@ const TBFollowingVotingProtocol = "TBFollowingVotingGossip"
...
-		return nil
-	}
+	return nil
+}
 
+func (tb *TortoiseBeacon) classifyProposalMessage(m ProposalMessage, atxID types.ATXID, currentEpoch types.EpochID, receivedTime time.Time) error {
 	atxHeader, err := tb.atxDB.GetAtxHeader(atxID)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get ATX header: %w", err)
```
antonlerner: add miner and / or atx info to error

tortoisebeacon/handlers.go
```diff
@@ -25,92 +26,99 @@ const TBFollowingVotingProtocol = "TBFollowingVotingGossip"
...
 	}
 
 	atxTimestamp, err := tb.atxDB.GetAtxTimestamp(atxID)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get ATX timestamp: %w", err)
```
antonlerner: add atx info

tortoisebeacon/handlers.go
```diff
@@ -148,16 +156,51 @@ func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalM
...
+		tb.Log.With().Warning("Received malformed proposal message: VRF is not verified",
+			log.String("sender", m.MinerID.Key))
+
+		// TODO(nkryuchkov): add a test for this case
+		return ErrMalformedProposal
+	}
+
+	epochWeight, _, err := tb.atxDB.GetEpochWeight(currentEpoch)
+	if err != nil {
+		return fmt.Errorf("get epoch weight: %w", err)
```
antonlerner: add epoch num

tortoisebeacon/handlers.go
```diff
@@ -148,16 +156,51 @@ func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalM
...
+	}
+
+	passes, err := tb.proposalPassesEligibilityThreshold(m.VRFSignature, epochWeight)
+	if err != nil {
+		return fmt.Errorf("proposalPassesEligibilityThreshold: %w", err)
```
antonlerner: add weight and maybe short string of proposal

tortoisebeacon/handlers.go
```diff
@@ -148,16 +156,51 @@ func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalM
...
+	}
+
+	if !passes {
+		tb.Log.With().Warning("Rejected proposal message which doesn't pass threshold")
```
antonlerner: add short string of proposal

tortoisebeacon/handlers.go
```diff
@@ -214,20 +257,26 @@ func (tb *TortoiseBeacon) HandleSerializedFollowingVotingMessage(ctx context.Con
...
 	_, err := tb.atxDB.GetNodeAtxIDForEpoch(from, currentEpoch-1)
-	if err != nil {
+	if errors.Is(err, database.ErrNotFound) {
 		tb.Log.With().Warning("Miner has no ATXs in the previous epoch")
+
 		return nil
 	}
 
+	if err != nil {
+		return fmt.Errorf("get node ATXID for epoch: %w", err)
```
antonlerner: add node info

tortoisebeacon/message.go
```diff
@@ -22,6 +25,12 @@ func (p ProposalMessage) String() string {
 	return string(bytes)
 }
 
+type extendedProposalMessage struct {
```
antonlerner: why is this named extended? it is not clear from its name what its purpose

tortoisebeacon/tortoise_beacon.go
```diff
@@ -253,13 +235,18 @@ func (tb *TortoiseBeacon) Close() error {
...
+		if err == nil {
 			return val.Bytes(), nil
 		}
+
+		if !errors.Is(err, database.ErrNotFound) {
+			tb.Log.Error("Failed to get tortoise beacon for epoch %v from DB: %v", epochID-1, err)
+		}
 	}
 
 	if (epochID - 1).IsGenesis() {
```
antonlerner, line 255: `if epochID >0 && (epochID - 1).IsGenesis()`

tortoisebeacon/tortoise_beacon.go
```diff
@@ -253,13 +235,18 @@ func (tb *TortoiseBeacon) Close() error {
 // GetBeacon returns a Tortoise Beacon value as []byte for a certain epoch or an error if it doesn't exist.
 func (tb *TortoiseBeacon) GetBeacon(epochID types.EpochID) ([]byte, error) {
 	if tb.tortoiseBeaconDB != nil {
-		if val, ok := tb.tortoiseBeaconDB.GetTortoiseBeacon(epochID - 1); ok {
+		val, err := tb.tortoiseBeaconDB.GetTortoiseBeacon(epochID - 1)
```
antonlerner, line 243: check that epochID > 0

tortoisebeacon/handlers.go
```diff
@@ -214,20 +257,26 @@ func (tb *TortoiseBeacon) HandleSerializedFollowingVotingMessage(ctx context.Con
...
 
-func (tb *TortoiseBeacon) handleFirstVotingMessage(ctx context.Context, message FirstVotingMessage) error {
+func (tb *TortoiseBeacon) handleFirstVotingMessage(message FirstVotingMessage) error {
 	currentEpoch := tb.currentEpoch()
 	from := message.MinerID
 
 	_, err := tb.atxDB.GetNodeAtxIDForEpoch(from, currentEpoch-1)
-	if err != nil {
+	if errors.Is(err, database.ErrNotFound) {
 		tb.Log.With().Warning("Miner has no ATXs in the previous epoch")
```
antonlerner: add miner ID

tortoisebeacon/tortoise_beacon.go
```diff
@@ -393,21 +400,15 @@ func (tb *TortoiseBeacon) handleEpoch(ctx context.Context, epoch types.EpochID)
 	tb.Log.With().Info("Handling epoch",
 		log.Uint64("epoch_id", uint64(epoch)))
 
-	if err := tb.runProposalPhase(ctx, epoch); err != nil {
-		tb.Log.With().Error("Failed to send proposal",
-			log.Uint64("epoch_id", uint64(epoch)),
-			log.Err(err))
+	tb.proposalChansMu.Lock()
+	tb.closePrevChan(epoch)
```
antonlerner: won't this close the channel and then return the closed channel?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -421,7 +422,49 @@ func (tb *TortoiseBeacon) handleEpoch(ctx context.Context, epoch types.EpochID)
...
+
+				return
+			}
+
+			em.gossip.ReportValidation(ctx, TBProposalProtocol)
+		}
+	}
+}
+
+func (tb *TortoiseBeacon) closePrevChan(epoch types.EpochID) {
```
antonlerner: why is this called close prev channel?  it closes proposal channel for epoch `epoch`

nkryuchkov:
bors try

tortoisebeacon/tortoise_beacon.go
```diff
@@ -252,14 +235,23 @@ func (tb *TortoiseBeacon) Close() error {
...
+	}
+
 	if tb.tortoiseBeaconDB != nil {
-		if val, ok := tb.tortoiseBeaconDB.GetTortoiseBeacon(epochID - 1); ok {
+		val, err := tb.tortoiseBeaconDB.GetTortoiseBeacon(epochID - 1)
+		if err == nil {
 			return val.Bytes(), nil
 		}
+
+		if !errors.Is(err, database.ErrNotFound) {
```
antonlerner, line 248: if the error is not `database.ErrNotFound` we need to calculate it's value ?
If this is the expected behaviour add it as a comment

nkryuchkov: I think that an error other than `database.ErrNotFound` should mean that something is wrong with the DB and should not normally happen. In this case, beacon still may be calculated. I'm not sure if we need to calculate it in this case though. I'll return an error in this case.

antonlerner (approved):

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #2557, opened 2021-07-14
[Merged by Bors] - fix duplicate SendRequest for the same hash

### LINKED ISSUE #2556: multiple SendRequest for the same hash
## Description
in fetch.go, we keep both unhandled and outstanding requests in the same list.
this means we may send out multiple requests for the same hash in consecutively.

for example, first call of requestHashBatchFromPeers() sent out a batched request for requests N, ... N+M
before we get the response for the first request, if the clock ticks or a new request arrives, it will cause the 2nd call of requestHashBatchFromPeers(), which in turn will send out the same batched request for requests N, ..., N+M

countvonzero:
## Motivation

Closes #2556

## Changes
- added a unit test that fails before this change and passes afterwards
- create a separate pending list for outstanding requests waiting for peer responses
- calculate the actual hash at most once for each response for a hash

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -177,10 +177,14 @@ type network interface {
 
 // Fetch is the main struct that contains network peers and logic to batch and dispatch hash fetch requests
 type Fetch struct {
-	cfg                  Config
-	log                  log.Log
-	dbs                  map[Hint]database.Store
-	activeRequests       map[types.Hash32][]*request
+	cfg Config
+	log log.Log
+	dbs map[Hint]database.Store
+	// activeRequests contains requests that are not processed
+	activeRequests map[types.Hash32][]*request
+	// pendingRequests contains requests that have been processed and are waiting for responses
+	pendingRequests map[types.Hash32][]*request
+	// activeBatches contains batches of requests in pendingRequests.
 	activeBatches        map[types.Hash32]requestBatch
 	net                  network
 	requestReceiver      chan request
@@ -204,6 +208,7 @@ func NewFetch(ctx context.Context, cfg Config, network service.Service, logger l
 		log:             logger,
 		dbs:             make(map[Hint]database.Store),
 		activeRequests:  make(map[types.Hash32][]*request),
+		pendingRequests: make(map[types.Hash32][]*request),
 		net:             srv,
 		requestReceiver: make(chan request),
 		batchTimeout:    time.NewTicker(time.Millisecond * 50),
@@ -231,7 +236,11 @@ func (f *Fetch) Stop() {
 	close(f.stop)
 	f.activeReqM.Lock()
 	for _, batch := range f.activeRequests {
-
+		for _, req := range batch {
+			close(req.returnChan)
+		}
+	}
+	for _, batch := range f.pendingRequests {
 		for _, req := range batch {
 			close(req.returnChan)
 		}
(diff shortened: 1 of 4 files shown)


p2p/server/msgserver.go
```diff
@@ -103,9 +103,11 @@ func NewMsgServer(ctx context.Context, network Service, name string, requestLife
 
 // Close stops the MessageServer
 func (p *MessageServer) Close() {
-	p.exit <- struct{}{}
-	<-p.exit
+	p.With().Info("closing MessageServer")
+	close(p.exit)
```
dshulyak, line 107: i think the point was to wait until readLoop exits. it is started without workerCount.Add(1). otherwise it is possible that that Add(1) in readLoop will be called after Wait here

countvonzero: great catch!!! thanks for the review. much appreciated.

fetch/fetch.go
```diff
@@ -268,16 +277,22 @@ func (f *Fetch) loop() {
 	for {
 		select {
 		case req := <-f.requestReceiver:
-			// group requests by hash
+			sendNow := req.priority > Low
 			f.activeReqM.Lock()
-			f.activeRequests[req.hash] = append(f.activeRequests[req.hash], &req)
+			// group requests by hash
+			if sendNow {
```
antonlerner, line 283: what happens if we have one hash sent with high priority and then with low priority, will this will still result in 2 calls made to other peers?

fetch/fetch.go
```diff
@@ -177,10 +177,14 @@ type network interface {
...
-	log                  log.Log
-	dbs                  map[Hint]database.Store
-	activeRequests       map[types.Hash32][]*request
+	cfg Config
+	log log.Log
+	dbs map[Hint]database.Store
+	// activeRequests contains requests that are not processed
+	activeRequests map[types.Hash32][]*request
+	// pendingRequests contains requests that have been processed and are waiting for responses
+	pendingRequests map[types.Hash32][]*request
```
antonlerner, line 186: why add this distinction? an active request is a request that hasn't received a response yet. why do we need to distinct between active requests and requests waiting for response?

countvonzero: because we don't delete the requests from activeRequests when we call requestHashBatchFromPeers(). so when we call requestHashBatchFromPeers() is called again before the first one got response back, the same set of hashes will be requested again. however, we do need a place to hold the outstanding requests, hence a separate list.

you can see it in the test i added that SendRequest() is called twice for the same hash before the first requestHashBatchFromPeers() got response.

fetch/fetch.go
countvonzero: prior to the change, for a high priority hash req1, we will send out two requests, one now, and one within the batch.
now it just send out a request for req1 now, and not in the batch. 

but if someone send in req1 (high priority) and req2 (low priority), req1 will be sent right away and req2 will be batched up.

in short. old behavior: high priority req got sent twice. new behavior, high priority got sent just once.

antonlerner: >but if someone send in req1 (high priority) and req2 (low priority), req1 will be sent right away and req2 will be batched up.

we can even have this case optimised if we check if there are pending requests with this hash right?

fetch/fetch.go
```diff
@@ -416,12 +433,15 @@ func (f *Fetch) receiveResponse(data []byte) {
...
+				// put the request back to the active list
+				f.activeRequests[req.hash] = append(f.activeRequests[req.hash], req)
 			}
 		}
-		if invalidatedRequests == len(reqs) {
-			delete(f.activeRequests, h)
-		}
+		// the remaining requests in pendingRequests is either invalid (exceed MaxRetriesForRequest) or
+		// put back to the active list.
+		// TODO should we notify channels of the invalid requests?
```
antonlerner, line 443: this happens in line 430

antonlerner (approved):

fetch/fetch.go
countvonzero: oh right. silly me.... thanks

fetch/fetch.go
countvonzero: good idea. i think that requires a bigger code change and more tests to change both high and low priority requests. i will do it in the next PR together with reduce logging

countvonzero:
bors merge

PR #2561, opened 2021-07-15
[Merged by Bors] - reduce logging level for high frequency events

countvonzero:
## Motivation
- reduce logging level for high frequency events to reduce CI log size
- following up on review feedback for https://github.com/spacemeshos/go-spacemesh/pull/2557

## Changes
- change INFO -> DEBUG level for some high frequency events
- in fetcher.go, for high priority request, also make sure multiple SendRequest is not called.

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/blocks/blockoracle.go
+++ b/blocks/blockoracle.go
@@ -111,7 +111,7 @@ func (bo *Oracle) calcEligibilityProofs(epochNumber types.EpochID) (map[types.La
 
 	bo.log.With().Info("Got beacon",
 		log.Uint64("epoch_id", uint64(epochNumber)),
-		log.Err(err))
+		log.String("epoch_beacon", fmt.Sprint(epochBeacon)))
 
 	var weight uint64
 	// get the previous epoch's total weight
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -193,14 +193,14 @@ func (l *Logic) AddDBs(blockDB, AtxDB, TxDB, poetDB, IvDB, tbDB database.Store) 
 func (l *Logic) layerHashReqReceiver(ctx context.Context, msg []byte) []byte {
 	lyr := types.NewLayerID(util.BytesToUint32(msg))
 	h := l.layerDB.GetLayerHash(lyr)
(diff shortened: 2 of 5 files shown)


layerfetcher/layers.go
```diff
@@ -216,9 +216,10 @@ func (l *Logic) layerHashBlocksReqReceiver(ctx context.Context, msg []byte) []by
 
 	blocks := l.layerDB.GetLayerHashBlocks(h)
 	vector, err := l.layerDB.GetLayerInputVector(h)
```
antonlerner, line 218: I think @y0sher told me getting input vector here is redundant, @y0sher ?

antonlerner (approved):

layerfetcher/layers.go
countvonzero: i will double check with @y0sher
i want to remove this separately. i actually prefer keeping it here and don't make an extra network RTT at the end of PollLayerBlocks()

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

blocks/blockoracle.go
```diff
@@ -111,7 +111,7 @@ func (bo *Oracle) calcEligibilityProofs(epochNumber types.EpochID) (map[types.La
 
 	bo.log.With().Info("Got beacon",
 		log.Uint64("epoch_id", uint64(epochNumber)),
-		log.Err(err))
+		log.String("epoch_beacon", fmt.Sprint(epochBeacon)))
```
dshulyak, line 114: is this a binary value? it will be better to use log.Binary, it will encode it in base64 and won't copy memory until the moment when it needs to be printed. unfortunately it is not that easy to extend zap to encode binary to hexary instead of base64

countvonzero: actually i copied it from a Debug log below. i'm going to change it to
```
types.BytesToHash(epochBeacon).ShortString()
```
to be consistent with the rest of the code. thanks for the reminder 💯

dshulyak: that is still not perfect btw :) this will copy and encode this field even if it is not printed later

PR #2562, opened 2021-07-15
[Merged by Bors] - meshdb: do not recompute ID and MinerID when block is loaded from database

dshulyak:
## Motivation

blk.Initialize function is very slow, see the difference in benchmark:

```
benchmark                      old ns/op     new ns/op     delta
BenchmarkGetBlockHeader-16     119538        2810          -97.65%
```

It computes MinerID from signature and ID from encoded block every time when we load block from database.

related: https://github.com/spacemeshos/go-spacemesh/issues/2478

## Changes
Introduce DBBlock type that will encode ID and MinerID as a part of the block when block is saved to the local database.

## Test Plan
unit tests
## TODO

## DevOps Notes

### DIFF
--- a/mesh/meshdb.go
+++ b/mesh/meshdb.go
@@ -202,10 +202,11 @@ func (m *DB) GetBlock(id types.BlockID) (*types.Block, error) {
 	if err != nil {
 		return nil, err
 	}
-	mbk := &types.Block{}
-	err = types.BytesToInterface(b, mbk)
-	mbk.Initialize()
-	return mbk, err
+	mbk := &types.DBBlock{}
+	if err := types.BytesToInterface(b, mbk); err != nil {
+		return nil, err
+	}
+	return mbk.ToBlock(), nil
 }
 
 // LayerBlocks retrieves all blocks from a layer by layer index
@@ -396,7 +397,13 @@ func (m *DB) SaveLayerHashInputVector(h types.Hash32, data []byte) error {
 }
 
 func (m *DB) writeBlock(bl *types.Block) error {
-	if bytes, err := types.InterfaceToBytes(bl); err != nil {
+	block := &types.DBBlock{
+		MiniBlock: bl.MiniBlock,
+		ID:        bl.ID(),
+		Signature: bl.Signature,
+		MinerID:   bl.MinerID().Bytes(),
+	}
+	if bytes, err := types.InterfaceToBytes(block); err != nil {
 		return fmt.Errorf("could not encode block: %w", err)
 	} else if err := m.blocks.Put(bl.ID().AsHash32().Bytes(), bytes); err != nil {
 		return fmt.Errorf("could not add block %v to database: %w", bl.ID(), err)
--- a/mesh/meshdb.go
+++ b/mesh/meshdb.go
@@ -159,19 +159,19 @@ func (m *DB) Close() {
 //todo: for now, these methods are used to export dbs to sync, think about merging the two packages
 
 // Blocks exports the block database
-func (m *DB) Blocks() database.Database {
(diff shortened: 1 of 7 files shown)


dshulyak:
@antonlerner it appears that fetcher can actually request any data that is stored in the database in the raw format. how I can change the way block is stored in  the database without causing changes all over the place?

dshulyak:
I made a wrapper that protects block store from internal changes. It still relies on a fact that the key will remain the same, but I think this is fine.

mesh/meshdb.go
```diff
@@ -1069,3 +1076,23 @@ func (m *DB) cacheWarmUpFromTo(from types.LayerID, to types.LayerID) error {
...
+
+// BlockFetcherDB implements API that allows fetcher to get a block from a remote database.
+type BlockFetcherDB struct {
+	mdb *DB
+}
+
+// Get types.Block encoded in byte using hash.
+func (db *BlockFetcherDB) Get(hash []byte) ([]byte, error) {
+	id := types.BlockID(types.BytesToHash(hash).ToHash20())
+	blk, err := db.mdb.GetBlock(id)
```
lrettig, line 1088: Combine with the following line

mesh/meshdb_test.go
```diff
@@ -975,3 +975,24 @@ func TestMesh_FindOnce(t *testing.T) {
 		}
 	})
 }
+
+func BenchmarkGetBlockHeader(b *testing.B) {
+	// blocks is set to be twice as large as cache to avoid hitting the cache
+	cache := layerSize
+	blocks := make([]*types.Block, cache*2)
+	// 1 is passed as a cache size because it is actually multiplied by layerSize
```
lrettig: I don't understand this comment without context. Why is it multiplied by layer size?

lrettig (commented):
I don't see any new unit tests (other than the benchmark, which is great). Are the new code paths covered by existing tests? I feel like probably at least one new unit test should be added to test the new code.

mesh/meshdb_test.go
dshulyak: internally size of the cache will be multiplied by a constant (layerSize = 200). will update the comment

mesh/meshdb.go
dshulyak: did you mean to combine id initialization with this line?

lrettig: I mean move the if... conditional up one line and combine it with this line (line 1093)

antonlerner (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

noamnelke:
I know this is merged already and I think it was a good "quick fix", but I wonder if we can't implement for blocks the same mechanism we have for ATXs? @dshulyak 

The gist of it is that we always index ATXs by ID, so whenever we fetch an ATX from the DB we already know its ID - so we can just set it without calculating it. The only place we calculate an ATX ID is when receiving it over the network and then it's unavoidable (trust issue).

My question is - can we use the same principle for blocks (saving some space and an additional type + being more consistent) or is there some other scenario with blocks that makes this impossible? If it's not impossible - can you please create a tech-debt issue for this?

dshulyak:
@noamnelke i noticed that we can do that. it is only 32b, and given that whole block was 10kb on a testnet saving space didn't seem important to me. i decided that it is better to have a future-proof id representation that won't change if index has to change. but overall it doesn't matter to me much, if you want i can remove it from DBBlock and recover from ID.

PR #2571, opened 2021-07-19
[Merged by Bors] - relax sync failure on ATX failure

countvonzero:
## Motivation
when syncing current epoch, sync fails whenever it fails to fetch an ATX, and it will take the node from synced state to notSynced state. this is possibly too aggressive for current epoch, as ATXs are not fully gossiped in the network.

## Changes
for current epoch, only fail sync on ATX failure at last layer. 
for previous epoch, sync fails as soon as it fails to fetch ATXs for that epoch.

## Test Plan
unit test

## DevOps Notes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -392,17 +392,22 @@ func (s *Syncer) getATXs(ctx context.Context, layerID types.LayerID) error {
 		return nil
 	}
 	epoch := layerID.GetEpoch()
-	currentEpoch := s.ticker.GetCurrentLayer().GetEpoch()
+	atCurrentEpoch := epoch == s.ticker.GetCurrentLayer().GetEpoch()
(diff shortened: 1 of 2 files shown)


antonlerner:
I assume that this PR is a bit less relevant after #2573 
Would you still like to merge this change? @countvonzero

antonlerner (approved):

countvonzero:
@antonlerner 

yes. i think this PR is still useful. for getting epoch ATXs, layer fetcher simply pick one peer to send request. if network is flaky we will fall out-of-sync and require gossip-sync to recover. i think this is too severe.

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #2572, opened 2021-07-19
[Merged by Bors] - sync: fetch ATXs from peers before syncing everything else

countvonzero:
## Motivation
part of https://github.com/spacemeshos/go-spacemesh/issues/3003

## Changes
- sync ATX to the current epoch before syncing all other data
- once ATXs are synced, the node will start listening to ATX gossip

## Test Plan
unit tests, systest, manual testing

## DevOps Notes


noamnelke:
> ATX messages can be verified without historical data

This is not true. It may be wise to start gossiping ATXs before other things, but they still rely on the previous ATX, one epoch back.

The reliance on the positioning ATX is different, since it's an ATX published close to the publication of the ATX referencing it, so even if we need to fetch the positioning ATX on-demand, it won't go too much back.

countvonzero:
thanks @noamnelke 

i wasn't expressing the motivation correctly. i think the rationale is that gossiping ATXs, and fetching the referenced ATXs, previous or positioning, should be an independent chain of event aside from syncer. i.e. the node's sync state should not affect how the node gossip about ATXs.

> ATX messages can be verified without historical data and should be gossiped all the time.

so i'll rephrase the motivation like i restated above

i think this PR is still valid as is. always gossip about ATXs regardless of the node's sync state. do you see any problems/potential attacks?

activation/atxlistener.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// loop listens to gossip messages until ctx is canceled.
+func (al *ATXListener) loop() error {
+	al.logger.Info("start listening to ATX gossip")
+	for {
+		select {
+		case atx := <-al.atxMessages:
+			al.tomb.Go(func() error {
+				al.processor.HandleGossipAtx(al.tomb.Context(nil), atx, al.fetcher)
```
dshulyak: i think this needs to be limited (e.g. after max number of handlers are spawned loop should either drop the ATX or block until the worker is available). if I read it right the previous code was always blocking until the previous ATX was Handled, but unlimited number of workers is also not a correct approach

activation/atxlistener.go
```diff
@@ -0,0 +1,71 @@
...
+const AtxProtocol = "AtxGossip"
+
+type atxProcessor interface {
+	HandleGossipAtx(context.Context, service.GossipMessage, service.Fetcher)
+}
+
+// ATXListener handles ATX gossip messages.
+type ATXListener struct {
+	startOnce   sync.Once
+	tomb        tomb.Tomb
```
dshulyak: are you sure that tomb makes sense here? it looks like it is the same as errgroup and supposed to manage goroutines that must be killed if one of them exits with error

from doc

> If any of the tracked goroutines returns a non-nil error, or the Kill or Killf method is called by any goroutine in the system (tracked or not), the tomb Err is set, Alive is set to false, and the Dying channel is closed to flag that all tracked goroutines are supposed to willingly terminate as soon as possible.

dshulyak (commented):
is it mostly a refactoring PR? i looked up an it seems that gossip.AddListener will do the same as ATXListner.Start

UPD: didn't notice before, AddListener has a hidden condition that uses syncer.ListenToGossip to decide if the message should be propagated or not. in my opinion it would be better to disable that behavior with an additional option passed to the AddListener.

activation/atxlistener.go
dshulyak: i see now that you always return nil

activation/atxlistener.go
noamnelke: This is probably not the right place for this discussion, but I don't mind explaining it again later.

I think that we do a bad job of managing concurrency in the node. We don't currently have a way to prioritize urgent tasks (e.g. participating in Hare) over less urgent ones (e.g. processing an incoming ATX). The closest thing we have is a priority queue for gossip, which I think is not enough and we see the effects when running nodes on weak hardware on the testnet and tests.

I think there are clear priorities and those change through the course of the layer. Here's a rough outline:
image

So the challenge is transitioning from having each module manage its own workers that consume an internal queue of tasks to code that manages centralized (within the node) queue of tasks and prioritizes them dynamically and a set of workers that consume tasks from the queue. The number of workers would be configurable, but by default it can be chosen by the node based on the machine its running on.

I believe this would considerably reduce the variability of the node's resource consumption and as a major added bonus, it would make debugging issues much easier since we can expose which tasks are being handled at any given moment making rationalizing what's going on in the node much easier (I imagine a "task log" with "task X started" and "task X completed" type messages).

activation/atxlistener.go
```diff
@@ -0,0 +1,71 @@
...
+func (al *ATXListener) Start() {
+	al.startOnce.Do(func() {
+		al.tomb.Go(al.loop)
+	})
+}
+
+// Stop shutdown the listener gracefully.
+func (al *ATXListener) Stop() error {
+	al.logger.Info("stopping ATX gossip processing")
+	al.tomb.Kill(nil)
```
antonlerner: what will happen if this is called twice?

noamnelke: [Nothing](https://pkg.go.dev/gopkg.in/tomb.v2#Tomb.Kill)

activation/atxlistener.go
countvonzero: @noamnelke would you like to capture this in an issue, or have you already?

dshulyak: it is hard to tell for me if it is a good idea or not, i believe it requires careful design. i also would like to find any good project in golang that is doing goroutine prioritization and runs well.

but as for this change, cpu is mostly idle on miners with an exception is actually handling atxs (you can see that in pyroscope history). so this is clearly a cpu bound task and we shouldn't blindly allocate a worker for every received atx. i suggest to limit number of workers that can process atx to NumCPU/2. 
Screenshot from 2021-08-07 09-09-15

dshulyak (changes requested):

activation/atxlistener.go
noamnelke: I haven't yet. I've given it some thought and created this illustration, but I still need to sit down and dive into the details. If I won't get to it soon, I'll just start with what I've got and flesh it out later.

countvonzero:
a follow up discussion in the R&D meeting was that it's a good idea to sync ATXs before anything else (blocks/TXs).
combined with profiling data @dshulyak provided that syncing ATXs is actually CPU bound, i think the next step for syncing ATXs would be 
A. assign a dedicated worker pool w. configurable num of CPUs for handling ATXs (both in catchup sync and steady-state sync)
B.  in the case of catchup sync, we give ATXs a headstart before everything else.

@noamnelke @dshulyak sounds reasonable to you?
if so, any suggestion on libraries that provides `worker pool w configurable num of CPUs`?

countvonzero:
bors try

cmd/node/node.go
```diff
@@ -618,7 +618,22 @@ func (app *App) initServices(ctx context.Context,
 	app.host.Register(beacon.FollowingVotesProtocol,
 		pubsub.ChainGossipHandler(syncHandler, beaconProtocol.HandleFollowingVotes))
 	app.host.Register(proposals.NewProposalProtocol, pubsub.ChainGossipHandler(syncHandler, proposalListener.HandleProposal))
-	app.host.Register(activation.AtxProtocol, pubsub.ChainGossipHandler(syncHandler, atxHandler.HandleGossipAtx))
+
+	// ValidatePost is CPU intensive. limit its number of workers.
+	numWorkers := runtime.NumCPU()
+	if numWorkers > 1 {
+		numWorkers /= 2
+	}
```
dshulyak: i remember that it was discussed before, but we shouldn't enforce this limit. we have a fetcher codepath in every handler, which can take unpredictable amount of time

countvonzero: my rationale is that once we enter gossip mode, fetcher should fetch minimal amount of ATXs in a steady state. (unless we are doing a network rejoin).

so when we start syncing, we spend most of the CPUs processing ATX and nothing else. once ATXs are synced, then the majority of ATX processing is on half of the CPUs.

in any case, i am fine removing the CPUs part from this PR and deal with it later when it becomes a problem.

countvonzero:
bors try

cmd/node/node.go
countvonzero: one interesting thing: with the 1/2 CPUs dedicated to ATX gossip via p2p validator, systest failed across the board. i looked briefly and see a lot more hare/beacon errors because miner has no ATXs. that seems to hint that the node wasn't able to process all the ATXs in time.

dshulyak: they may have been dropped as well. i don't remember what exactly is controlled with a variable that you changed. and what happens on overflow

countvonzero: i used WithValidatorConcurrency(), which is probably wrong. the default is 1024.

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7077551719?check_suite_focus=true)

unit test timedout

countvonzero:
bors merge

countvonzero:
bors merge

Issue #2574, opened 2021-07-19
standardize mesh hash calculation


countvonzero:
## Description
there are several issues with mesh.go w.r.t. simple layer hash and aggregated layer hash

1. mesh.GetLayerHash() != mesh.GetLayer().Hash() 
former is what we persisted in the db, latter is calculated at runtime.
we should change it to:
for layers that tortoise has reached an opinion, we just use the hash it persisted (only ctxual valid blocks). otherwise use all blocks we've have for the layer.

2. mesh.layerHash vs running layer hash
mesh.layerHash is essentially hash32(prev mesh.layerHash+"valid" blocks in this layer).
running layer hash (accumulated layer hash) is hash32(prev running layer hash + blocks in this layer, valid or not)
these two really should have the same value. i.e. mesh.layerHash should have the same value as the last processed layer's running hash.

PR #2585, opened 2021-07-21
[Merged by Bors] - mesh: save processed layer and hash atomically

countvonzero:
## Motivation
part I of resolving https://github.com/spacemeshos/go-spacemesh/issues/2574

## Changes
- remove unused mesh public methods
- make sure processed layer and accumulated hash are saved/persisted together
- make sure processed layer are set in ascending order

## Test Plan
unit test

## DevOps Notes

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -5,7 +5,6 @@ package mesh
 import (
 	"context"
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"math/big"
 	"math/rand"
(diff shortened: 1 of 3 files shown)


countvonzero:
bors try

mesh/mesh.go
```diff
@@ -239,23 +232,66 @@ func (msh *Mesh) GetLayer(i types.LayerID) (*types.Layer, error) {
...
-	defer vl.lvMutex.RUnlock()
-	vl.lvMutex.RLock()
+	return vl.getProcessedLayer().layerID
+}
+
+func (vl *validator) ProcessedLayerHash() types.Hash32 {
+	return vl.getProcessedLayer().layerHash
+}
+
+func (vl *validator) SetProcessedLayer(layerID types.LayerID, hash types.Hash32) {
```
antonlerner: this function is used only internally and can be unexported

mesh/mesh.go
```diff
@@ -239,23 +232,66 @@ func (msh *Mesh) GetLayer(i types.LayerID) (*types.Layer, error) {
...
 	events.ReportNodeStatusUpdate()
-	defer vl.lvMutex.Unlock()
-	vl.lvMutex.Lock()
-	vl.processedLayer = lyr
+	vl.Event().Info("processed layer set", lyr.layerID, log.String("layer_hash", lyr.layerHash.Hex()))
+}
+
+func (vl *validator) persistProcessedLayer() {
+	lyr := vl.getProcessedLayer()
+	data, err := types.InterfaceToBytes(lyr)
```
antonlerner, line 270: `InterfaceToBytes` only serialises exported fields, in our case  `internalLayer`s fields are unexported 
Please add a unit test to verify persist and store

mesh/mesh.go
```diff
@@ -239,23 +232,66 @@ func (msh *Mesh) GetLayer(i types.LayerID) (*types.Layer, error) {
 	return l, nil
 }
 
+type internalLayer struct {
+	layerID   types.LayerID
```
antonlerner: fields need to be exported for serialisation

mesh/mesh.go
```diff
@@ -473,19 +506,7 @@ func (msh *Mesh) logStateRoot(layerID types.LayerID) {
 
 func (msh *Mesh) calcValidLayerHash(layer *types.Layer) types.Hash32 {
```
antonlerner, line 518: add comment that layer hash is aggregated hash

mesh/mesh.go
countvonzero: done

mesh/mesh.go
countvonzero: done

mesh/mesh.go
countvonzero: done

we didn't check the number of bytes marshaled/unmarshaled so InterfaceToBytes/BytesToInterface fails silently in this case. i think we should at least print an warning if not return an error of zerobyte and let the caller decide whether it's expected...

mesh/mesh.go
countvonzero: done

countvonzero:
bors try

antonlerner (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #2590, opened 2021-07-22
[Merged by Bors] - timesync: compare system clock with peers clock

### LINKED ISSUE #2577: timesync: collect samples to calculate effective offset
SMIP: https://github.com/spacemeshos/SMIPS/issues/55

Initiator must send a message to a set of peers, such as:

```go
struct SyncRequest struct {
    ID uint64
}
```

Response should be sent immediatly that has following info:

```go
struct SyncResponse struct {
    ID uint64
    ReceiveTimestamp uint64
    ResponseTimestamp uint64
}
```

- [ ] why NEM uses two timestamps? they will be identical, unless goroutine that sends a message is suspended

In case of any failure initiator will retry after round is expired, with incremented ID. Note that Round is not incremented. ID is used to discard replies for earlier failed tries, it is fine for ID to wrap around.

### LINKED ISSUE #2578: timesync: compute offset according to rules described in SMIP
SMIP: https://github.com/spacemeshos/SMIPS/issues/55

dshulyak:
## Motivation

closes: https://github.com/spacemeshos/go-spacemesh/issues/2577
closes: https://github.com/spacemeshos/go-spacemesh/issues/2578

## Changes
TODO

## Test Plan
unit tests, testing on devnet with chaos-mesh and hijacking time

## TODO

## DevOps Notes


dshulyak:
We currently don't know the difference between peers that connected to us and peers that we dialed. This might be important if we want to avoid very simple eclipse scenarios. I will make changes for that in next PR

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+		MaxOffsetErrors:     10,
+		RequiredResponses:   3,
+		ResponsesBufferSize: 20,
+	}
+}
+
+// Config for Sync.
+type Config struct {
+	Disable            bool          `mapstructure:"disable"`
+	RoundRetryInterval time.Duration `mapstructure:"round-retry-interval"`
```
antonlerner, line 77: add comments about time units

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+				)
+				atomic.StoreUint32(&s.errCnt, 0)
+			}
+			timeout = s.config.RoundInterval
+		} else {
+			s.log.With().Error("failed to fetch offset from peers", log.Err(err))
+			timeout = s.config.RoundRetryInterval
+		}
+
+		round++
```
antonlerner, line 257: why do we need round counter?

timesync/peersync/sync.go
dshulyak: it allows to specify any time.Duration e.g 5s/10m/6h

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+	if errors.Is(err, context.Canceled) {
+		return nil
+	}
+	return err
+}
+
+func (s *Sync) run() error {
+	var (
+		timer *time.Timer
+		round uint64
```
antonlerner, line 214: rename to roundCounter

timesync/peersync/peers.go
```diff
@@ -0,0 +1,70 @@
...
+type waitPeersReq struct {
+	ch  chan []p2pcrypto.PublicKey
+	min int
+}
+
+type peersWatcher struct {
+	requests chan *waitPeersReq
+}
+
+func (p *peersWatcher) waitPeers(ctx context.Context, n int) ([]p2pcrypto.PublicKey, error) {
```
antonlerner, line 18: perhaps this code will be better located in p2p package as a utility for others to use as well

timesync/peersync/sync.go
dshulyak: initially i needed it to avoid counting delayed votes. i know that msgServer will already prevent it, as it have the unique id for each request. but decided to keep it anyway so that protocol can work over any connection without req/response server. can remove it if you prefer it

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+	defer s.log.With().Debug("exiting sync background worker")
+	for {
+		prs, err := s.peersWatcher.waitPeers(s.ctx, s.config.RequiredResponses)
+		if err != nil {
+			return err
+		}
+		s.log.With().Info("starting time sync round with peers",
+			log.Uint64("round", round),
+			log.Int("peers_count", len(prs)),
+			log.Uint32("errors_count", atomic.LoadUint32(&s.errCnt)),
```
antonlerner, line 226: why does this need to be atomic? is it run in parallel?

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+			added, expired := s.network.SubscribePeerEvents()
+			return s.peersWatcher.run(s.ctx, added, expired)
+		})
+	})
+}
+
+// Stop background workers.
+func (s *Sync) Stop() {
+	s.cancel()
+	s.srv.Close()
```
antonlerner, line 198: `srv` is not safe against calling stop twice

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+		}, func(error) {}); err != nil {
+			s.log.Debug("can't send request to peer", log.Err(err))
+			errCnt++
+		}
+		if d := len(prs) - errCnt; d < s.config.RequiredResponses || d == 0 {
+			return 0, fmt.Errorf("%w: failed to send requests to %d peers from %d", ErrTimesyncFailed, errCnt, len(prs))
+		}
+	}
+
+	wait := func() {
```
antonlerner, line 310: if `len(prs)-errCnt` < `s.config.RequiredResponses` we don't even have to wait for responses

dshulyak: we are not. i return on L306

timesync/peersync/sync.go
dshulyak: the idea is that the user can reset this counter after seeing a notification (or error in logs, depending on how they run a node). using an API call

timesync/peersync/round.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type round struct {
+	ID                uint64
+	Timestamp         int64
+	RequiredResponses int
+	responses         []timedResponse
+}
+
+func (r *round) AddResponse(resp Response, timestamp int64) {
```
antonlerner, line 20: can `AddResponse` and `Ready` br called at the same time? should this struct be thread safe?

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+		s.log.Debug("can't decode request", log.Binary("request", buf), log.Err(err))
+		return nil
+	}
+	resp := Response{
+		ID:        request.ID,
+		Timestamp: s.time.Now().UnixNano(),
+	}
+	buf, err := types.InterfaceToBytes(&resp)
+	if err != nil {
+		s.log.Panic("can't encode response", log.Binary("response", buf), log.Err(err))
```
antonlerner, line 177: should we panic here if response can't be encoded? maybe an error will suffice

timesync/peersync/sync.go
```diff
@@ -0,0 +1,330 @@
...
+		round     = round{
+			ID:                id,
+			Timestamp:         s.time.Now().UnixNano(),
+			RequiredResponses: s.config.RequiredResponses,
+		}
+		request = Request{ID: id}
+	)
+	buf, err := types.InterfaceToBytes(&request)
+	if err != nil {
+		s.log.With().Panic("can't encode request to bytes", log.Err(err))
```
antonlerner, line 284: maybe an error is enough

timesync/peersync/sync.go
dshulyak: pretty sure we should panic. this error can occur only because of the programmer error. like passing nil value to the function

timesync/peersync/round.go
dshulyak: it doesn't need to be thread safe

timesync/peersync/peers.go
dshulyak: makes sense. will move it

timesync/peersync/sync.go
dshulyak: :+1: will change that in MsgServer i think

dshulyak:
bors try

antonlerner (approved):

antonlerner (approved):

timesync/peersync/peers.go
dshulyak: i will open a tech debt issue for it and address it soon

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #2591, opened 2021-07-22
[Merged by Bors] - cobra command splitted from app execution logic

### LINKED ISSUE #2594: decouple node startup structure from `Cobra`
## Description

We want to be able to test spacemesh with `testground` which enables almost unit-test level testing as we can directly access go. currently `node.go` is written in a way that the `SpacemeshApp` struct cannot be started without being executed by `Cobra` .

## TODO 

Change the code so it can be run without using cobra.

## Affected code
`node.go`, `node_test.go`

y0sher:
## Motivation
We want to be able to access the node package without being tied to using cobra.

Closes #2594 

## Changes

separate the methods that initialize and start `go-spacemesh` from the structs in `cobra`

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -106,10 +106,6 @@ var Cmd = &cobra.Command{
 				log.With().Error("Failed to initialize node.", log.Err(err))
 				return err
 			}
-			if err := app.Initialize(); err != nil {
-				log.With().Error("Failed to initialize node.", log.Err(err))
-				return err
-			}
 			// This blocks until the context is finished or until an error is produced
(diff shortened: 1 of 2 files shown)


nkryuchkov (approved):

dshulyak (approved):
just a nit: maybe InitializeCmd can call Initialize inside it?

y0sher:
> just a nit: maybe InitializeCmd can call Initialize inside it?

gothcha, this makes sense since the only occurrence of InitializeCmd is followed by Initialize.

y0sher:
bors merge

PR #2592, opened 2021-07-23
[Merged by Bors] - make sure rand is not seeded more than once in prod binary

countvonzero:
## Motivation
for true pseudo-random, make sure we don't seed it multiple times in production code path

## Changes
apart from tests that uses special seed for rand, remove all other Seed calls that expect pseudo random.

## Test Plan
unit test

## DevOps Notes


dshulyak (approved):

countvonzero:
bors merge

PR #2599, opened 2021-07-27
[Merged by Bors] - Make tortoise beacon weak coin up-to-date

nkryuchkov:
## Motivation

Tortoise beacon weak coin was outdated, so the code didn't even compile. This PR makes it work and adds improvements for it.

## Changes

- Fix compilation issues
- Check VRF of incoming messages
- Fix some bugs
- Testing basic functionality
- Fix unit tests
- Add weak coin check to tortoise beacon system test

## Test Plan

## DevOps Notes


tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -59,27 +58,42 @@ type weakCoin struct {
...
 
 type signer interface {
-	Sign(msg []byte) ([]byte, error)
+	Sign(msg []byte) []byte
+}
+
+// DefaultThreshold defines default weak coin threshold.
+// It's equal to (2^256) / 2
+// (hex: 0x8000000000000000000000000000000000000000000000000000000000000000).
+func DefaultThreshold() *big.Int {
```
dshulyak, line 70: maybe we can just parse it from the hex since it is a constant?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -206,34 +239,55 @@ func (wc *weakCoin) calculateWeakCoin(epoch types.EpochID, round types.RoundID)
...
+		stringProposals = append(stringProposals, current.String())
 
 		if current.Cmp(smallest) == -1 {
 			smallest = current
 		}
 	}
 
-	result := &big.Int{}
-	result.And(smallest, big.NewInt(1))
+	coinBit := new(big.Int).And(smallest, big.NewInt(1))
```
dshulyak, line 266: smallest.Bit(0)

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -131,41 +152,49 @@ func (wc *weakCoin) PublishProposal(ctx context.Context, epoch types.EpochID, ro
...
+		log.Uint64("epoch_id", uint64(epoch)),
+		log.Uint64("round_id", uint64(round)),
+		log.String("proposal", types.BytesToHash(signedProposal).ShortString()))
+
 	return nil
 }
 
 func (wc *weakCoin) proposalExceedsThreshold(proposal []byte) bool {
-	proposalInt := new(big.Int).SetBytes(proposal[:])
+	proposalInt := new(big.Int).SetBytes(proposal)
```
dshulyak, line 164: bytes.Compare ?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -106,20 +120,27 @@ func (wc *weakCoin) Get(epoch types.EpochID, round types.RoundID) bool {
 }
 
 func (wc *weakCoin) PublishProposal(ctx context.Context, epoch types.EpochID, round types.RoundID) error {
-	p, err := wc.generateProposal(epoch, round)
+	wc.Log.With().Info("Publishing proposal",
+		log.Uint64("epoch_id", uint64(epoch)),
```
dshulyak, line 124: nit: epoch was changed to uin32

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,57 +1,118 @@
...
+
+			var wc weakCoin
+			wc.threshold = DefaultThreshold()
+			got := wc.proposalExceedsThreshold(tc.proposal)
+			r.Equal(tc.expected, got)
+		})
+	}
+}
+
+func Test_weakCoin_coinBitToBool(t *testing.T) {
```
dshulyak, line 75: i would remove this utility completely. bit can't be any other value except 0 or 1. just use Bit(0), or it can be selected manually from []byte slice without big.Int wrapper

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,57 +1,118 @@
...
+	require.True(t, ok)
+
+	tt := []struct {
+		name      string
+		proposal  []byte
+		threshold []byte
+		expected  bool
+	}{
+		{
+			name:      "Case 1",
```
dshulyak, line 49: name: Exceeds

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,57 +1,118 @@
...
+			proposal:  proposal1.Bytes(),
+			threshold: threshold.Bytes(),
+			expected:  true,
+		},
+		{
+			name:      "Case 2",
```
dshulyak, line 55: name: "NotExceeds"

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -131,41 +152,49 @@ func (wc *weakCoin) PublishProposal(ctx context.Context, epoch types.EpochID, ro
...
+	proposal := bytes.Buffer{}
 
-	msg.WriteString(proposalPrefix)
+	proposal.WriteString(proposalPrefix)
 
 	epochBuf := make([]byte, 8)
 	binary.LittleEndian.PutUint64(epochBuf, uint64(epoch))
-	if _, err := msg.Write(epochBuf); err != nil {
-		return nil, err
+	if _, err := proposal.Write(epochBuf); err != nil {
```
dshulyak, line 176: it must be epoch.ToBytes, this type was changed to uint32

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -131,41 +152,49 @@ func (wc *weakCoin) PublishProposal(ctx context.Context, epoch types.EpochID, ro
...
+		return nil, fmt.Errorf("write epoch to buffer: %w", err)
```
dshulyak, line 177: please panic here. it shouldn't happen with buffer

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -131,41 +152,49 @@ func (wc *weakCoin) PublishProposal(ctx context.Context, epoch types.EpochID, ro
...
 	}
 
 	roundBuf := make([]byte, 8)
 	binary.LittleEndian.PutUint64(roundBuf, uint64(round))
-	if _, err := msg.Write(roundBuf); err != nil {
-		return nil, err
+	if _, err := proposal.Write(roundBuf); err != nil {
+		return nil, fmt.Errorf("write round to buffer: %w", err)
```
dshulyak, line 183: panic

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -206,34 +239,55 @@ func (wc *weakCoin) calculateWeakCoin(epoch types.EpochID, round types.RoundID)
...
 
 	wc.Log.With().Info("Calculated weak coin",
 		log.Uint64("epoch_id", uint64(epoch)),
-		log.Uint64("round", uint64(round)),
-		log.Bool("weak_coin", coin))
+		log.Uint64("round_id", uint64(round)),
+		log.Bool("weak_coin", weakCoinValue))
+}
+
+func coinBitToBool(coinBit *big.Int) bool {
```
dshulyak, line 284: can be removed

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,57 +1,118 @@
...
+			proposal:  proposal2.Bytes(),
+			threshold: threshold.Bytes(),
+			expected:  false,
+		},
+	}
+
+	for _, tc := range tt {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
dshulyak, line 65: initialize new r with inner t instance

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,57 +1,118 @@
...
+			name:          "Panic",
+			coinBit:       big.NewInt(3),
+			expectedPanic: true,
+		},
+	}
+
+	for _, tc := range tt {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
dshulyak, line 106: initialize new r with inner t instance

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -206,34 +239,55 @@ func (wc *weakCoin) calculateWeakCoin(epoch types.EpochID, round types.RoundID)
...
 	}
 
-	smallest := new(big.Int).SetBytes(wc.proposals[pair][0][:])
+	smallest := new(big.Int).SetBytes(proposals[0])
+	stringProposals := []string{smallest.String()}
 
-	for i := 1; i < len(wc.proposals[pair]); i++ {
-		current := new(big.Int).SetBytes(wc.proposals[pair][i][:])
+	for i := 1; i < len(proposals); i++ {
+		current := new(big.Int).SetBytes(proposals[i])
```
dshulyak, line 258: no need to initialize big.Int. bytes.Compare

dshulyak (commented):
who is allowed to send vrf proposals? where do we verify it?

dshulyak:
some other things that can be improved:
- no reason to store collection of proposals for each round, store only smallest
- we have a map for activeRounds, but it is cleaned after each round has ended. we can store only one activeRound as value on WeakCoin

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: It was implemented as `"0x80" + strings.Repeat("00", 31))`, but I thought it might be confusing. Also, it's hard to say how many zeroes are in `0x8000000000000000000000000000000000000000000000000000000000000000` when looking at it. So I decided that `2^256 / 2` looks best. However, feel free to change it if you don't agree

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: Indeed. I didn't think of it. Looks way better

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: I missed that change. Then this needs to be fixed

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: As `SetBytes` expects a big-endian number, it looks to me like another way to do the same. Feel free to change if you want

tortoisebeacon/weakcoin/weak_coin_test.go
nkryuchkov: Well, I added it to just indicate that calculations are correct. With `Bit(0)` it's not needed because `Bit` is tested in the standard library

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: Yes, panic looks better here

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: i think it is a common practice to have a copyable constant for these things. like if you have a several implementations in different languages. it is probably can be also defined in genesis config.

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -106,20 +120,27 @@ func (wc *weakCoin) Get(epoch types.EpochID, round types.RoundID) bool {
...
+
+	proposal, err := wc.generateProposal(epoch, round)
 	if err != nil {
 		return err
 	}
 
-	if wc.proposalExceedsThreshold(p) {
+	signedProposal := wc.signProposal(proposal)
+
+	if wc.proposalExceedsThreshold(proposal) {
```
dshulyak, line 134: @nkryuchkov this should be proposalExceedsThreshold(signature) , isn't it? otherwise we can completely skip generating proposals for the a particular epoch/round combination

tortoisebeacon/weakcoin/weak_coin_test.go
nkryuchkov: Yes, needs to be done

nkryuchkov:
@dshulyak 
> who is allowed to send vrf proposals? 

Everyone who participates in tortoise beacon and whose proposal passes threshold

> where do we verify it?

https://github.com/spacemeshos/go-spacemesh/blob/fix/make-weak-coin-up-to-date/tortoisebeacon/weakcoin/handlers.go#L52

> no reason to store collection of proposals for each round, store only smallest
> we have a map for activeRounds, but it is cleaned after each round has ended. we can store only one activeRound as value on WeakCoin

Yes, I didn't make any optimizations yet before reviewing the code with research, because something may need to be changed after review. If this approach remains as it is, these optimizations need to be done

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: @dshulyak Yes, this is a mistake. Also, it would be better to rename it to something like `proposalPassesEligibilityThreshold` like in the `tortoisebeacon` package and negate its output

dshulyak:
@nkryuchkov i was trying to figure out if that verification is implemented or not. it seems that https://github.com/spacemeshos/go-spacemesh/blob/77f6e2f5106606995e8a4e7cb23942f2c58b29e2/tortoisebeacon/weakcoin/handlers.go#L91 only verifies that the signature matches miner id (this is also can be improved, miner id can be recovered from signature).

nkryuchkov:
@dshulyak 
Yes, this is correct. Also, miner ID should be recovered from signature in `tortoisebeacon`. Moreover, a check that proposal passes the eligibility threshold needs to be added. Now only the sender checks it (https://github.com/spacemeshos/go-spacemesh/blob/fix/make-weak-coin-up-to-date/tortoisebeacon/weakcoin/weak_coin.go#L134), which is enough for quick testing but not enough for production.

There's a list of what needs to be done in weak coin, feel free to add your thoughts here: https://github.com/spacemeshos/SMIPS/issues/28#issuecomment-887092949

dshulyak (approved):
@nkryuchkov is it possible to merge this change? will be easier to review my changes in a separate pr

nkryuchkov:
@dshulyak sure

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #2600, opened 2021-07-27
set cgo ldflafs to environment, use $(make ldflags) command

sudachen:
It sets CGO_LDFLAGS environment variable. Use it before start go test ... 

```
[sudachen/go-spacemesh|ldflags] ../go-spacemesh$ $(make ldflags)
[sudachen/go-spacemesh|ldflags] ../go-spacemesh$ go test ./activation
ok  	github.com/spacemeshos/go-spacemesh/activation	25.937s
```

To run tests in goland you need to setup environment variable in setting. To know the CGO_LDFLAGS value run in terminal `make print-ldflags`

image

### DIFF
--- a/Makefile.Inc
+++ b/Makefile.Inc
@@ -100,3 +100,8 @@ ldflags:
 	$(shell $(MAKE) -s get-gpu-setup 1>&2)
 	@echo 'export CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)"' > /tmp/.go-spacemesh.ldflags
 	@echo . /tmp/.go-spacemesh.ldflags
+.PHONY: ldflags
+
+print-ldflags: get-gpu-setup
+	@echo $(CGO_TEST_LDFLAGS)
+.PHONY: print-ldflags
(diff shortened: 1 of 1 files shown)


noamnelke (approved):

noamnelke:
@sudachen shouldn't we call the exported variable `CGO_TEST_LDFLAGS`?
No, go uses CGO_LDFLAGS to find hoe to build with cgo extentions.

noamnelke (changes requested):
Let's not merge this.

I just learned about `go env -w` - read about it [here](https://pkg.go.dev/cmd/go#hdr-Print_Go_environment_information).
This means we can set the default value for `CGO_LDFLAGS` in a global way. We can make the ldflags make command call
```
go env -w CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)"
```
Actually we can call it something more generic, like `set-go-env`. As long as we run this make command before using `go build`/`go test` it Just Works.

Regarding GoLand - it has a feature where you can always run one config before running another - so this will even prevent devs from forgetting to run the make command.

image

dshulyak:
would be nice if we can merge some version of this change

noamnelke:
> would be nice if we can merge some version of this change

I'll do something now.

noamnelke:
Superseded by #2627

PR #2603, opened 2021-07-28
[Merged by Bors] - Introduce logtest.New and a env variable to control verbosity during tests

### LINKED ISSUE #2598: Disable logs during test run
## Description

It is really hard to find out what test failed when there is a wall of text. Most of the time test output doesn't even fit in my terminal history and I can't find the test failure message. 

Logs might be necessary to debug the failure in some cases and we can keep them on CI, but there should be an option to disable them completely (reduce verbosity level).

Solution:
- Register a command line option to set a log level for tests
- Use common logger initializer across all tests

```go
var testLoglevel = flag.String("testlevel", "info", "")

func NewTestLog(testName string) Log {
     // set testLogLevel to a new logger with
}
```

dshulyak:
## Motivation

Closes https://github.com/spacemeshos/go-spacemesh/issues/2598

## Changes

- it is possible to control verbosity of all tests during test run with TEST_LOG_LEVEL env variable
In order to allow `go test ./...` TEST_LOG_LEVEL needs to be used instead of a flag. Otherwise module that do not import logtest will fail with unknown flag.
- if tests are executed without `-v` then only logs from failed tests will be printed

Example:

```
go test ./mesh/ -run=TestMeshDB
--- FAIL: TestMeshDB_GetStateProjection (0.00s)
    logger.go:130: 2021-07-28T08:36:26.971+0300	DEBUG	added block 91adc7dc3a to layer 5
    logger.go:130: 2021-07-28T08:36:26.971+0300	DEBUG	save contextual validity 91adc7dc3a true
    logger.go:130: 2021-07-28T08:36:26.972+0300	INFO	SaveLayerInputVectorByID: Saving input vector	{"layer_id": 5, "hash": "2594b6a92ebfb1c3312deb7d01c015fb95e9fbe9bd7bc6b527af07813ec7b910", "name": ""}
    meshdb_test.go:333: 
        	Error Trace:	meshdb_test.go:333
        	Error:      	Should not be: 2
        	Test:       	TestMeshDB_GetStateProjection
FAIL
FAIL	github.com/spacemeshos/go-spacemesh/mesh	0.102s
FAIL
```

```
go test ./mesh/ -run=TestMeshDB -log=info
--- FAIL: TestMeshDB_GetStateProjection (0.00s)
    logger.go:130: 2021-07-28T08:39:36.374+0300	INFO	SaveLayerInputVectorByID: Saving input vector	{"layer_id": 5, "hash": "2594b6a92ebfb1c3312deb7d01c015fb95e9fbe9bd7bc6b527af07813ec7b910", "name": ""}
    meshdb_test.go:333: 
        	Error Trace:	meshdb_test.go:333
        	Error:      	Should not be: 2
        	Test:       	TestMeshDB_GetStateProjection
FAIL
FAIL	github.com/spacemeshos/go-spacemesh/mesh	0.104s
FAIL
```

## TODO

## DevOps Notes


dshulyak:
I refactored one file meshdb_test.go to use this logger. I can change other files if the change is ok.

noamnelke (commented):
I think that by default logs should be turned off (return a `noop` logger) and it should be easy to turn them on when investigating.

Ideally we could add an argument when running the test without changing any code, but I don't know how I'd implement this, so changing the level in the code sounds reasonable to me.

I'm not sure your solution will work for all cases, I think we may create the logger inside some prod code, but I'm not sure. Even if that's the case - this can be a good opportunity to rectify that.

dshulyak:
@noamnelke i can return a noop if level is empty. but apart from this i think we should continue running tests with debug level until existing issues are fixed. also, i don't understand if you are suggesting to modify code every time we are seeing a random failure, because this is clearly not a good approach

dshulyak:
> I'm not sure your solution will work for all cases, I think we may create the logger inside some prod code, but I'm not sure. Even if that's the case - this can be a good opportunity to rectify that.

if this is the case it must be fixed separately

noamnelke (approved):
Fair enough, since we're currently in the process of fixing flaky tests, having the logs is probably more useful than not.

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
failed because of some deployment hiccup:

```
     def wait_to_statefulset_to_be_ready(statefulset_name, name_space, time_out=None):
        start = datetime.now()
        while True:
            resp = client.AppsV1Api().read_namespaced_stateful_set(name=statefulset_name, namespace=name_space)
            total_sleep_time = (datetime.now()-start).total_seconds()
            if resp.status.replicas == resp.status.ready_replicas:
                ready_replicas = resp.status.ready_replicas
                print("Total time waiting for statefulset {0} [size: {1}]: {2} sec\n".format(statefulset_name,
                                                                                             ready_replicas,
                                                                                             total_sleep_time))
                break
            print("{0}/{1} pods ready {2} sec               ".format(resp.status.ready_replicas, resp.status.replicas,
                                                                     total_sleep_time), end="\r")
            time.sleep(1)
    
            if time_out and total_sleep_time > time_out:
>               raise Exception("Timeout waiting for statefulset to be ready")
E               Exception: Timeout waiting for statefulset to be ready
```

dshulyak:
again same deployment error

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #2605, opened 2021-07-28
[Merged by Bors] - Finalize weak coin implementation

dshulyak:
## Motivation

Closes #

## Changes

- vrf proposals are made based on units of spacetime

For this message is extended with `Unit uint64` and included when computing signature.
Miner broadcast only smallest proposal if the has more than 1 unit.

- messages for next rounds are buffered and used when the round is started

- changed interface for the WeakCoin

## TODO

## DevOps Notes


tortoisebeacon/handlers.go
```diff
@@ -183,7 +183,7 @@ func (tb *TortoiseBeacon) verifyProposalMessage(m ProposalMessage, currentEpoch
 		return fmt.Errorf("calculate proposal: %w", err)
 	}
 
-	if !tb.vrfVerifier(m.MinerID.VRFPublicKey, currentEpochProposal, m.VRFSignature) {
+	if !tb.vrfVerifier.Verify(signing.NewPublicKey(m.MinerID.VRFPublicKey), currentEpochProposal, m.VRFSignature) {
```
dshulyak, line 186: @nkryuchkov this part is not finished, right? we need to recover public key from signature here and verify that miner is in the active set

tortoisebeacon/tortoise_beacon.go
```diff
@@ -606,32 +614,41 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
+		tb.weakCoin.CompleteRound()
 	}
 
-	tb.Log.With().Debug("Following voting message sender finished",
-		log.Uint64("epoch_id", uint64(epoch)))
-
-	tb.waitAfterLastRoundStarted()
-	tb.weakCoin.OnRoundFinished(epoch, tb.lastPossibleRound())
-
+	tb.weakCoin.CompleteEpoch()
```
dshulyak: @nkryuchkov i didn't understand why do we need to wait twice. in the loop above we are waiting already. was it intentional?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -662,10 +679,6 @@ func (tb *TortoiseBeacon) sendFollowingVotesLoopIteration(ctx context.Context, e
 		log.Uint64("round_id", uint64(round)))
 
 	go func(epoch types.EpochID, round types.RoundID) {
-		if round > firstRound+1 {
-			tb.weakCoin.OnRoundFinished(epoch, round-1)
-		}
-
```
dshulyak, line 668: moved this part after ticker.C in the consensusPhase

dshulyak:
I think there is still work that needs to be done to interact with tortoise beacon in a correct way. But I need more information about it

tortoisebeacon/handlers.go
nkryuchkov: @dshulyak It works with the `MinerID` field too, but the field is redundant, we can recover it from the signature. There's a check that miner has an ATX in a previous epoch [here](https://github.com/dshulyak/go-spacemesh/blob/finalize-weakcoin/tortoisebeacon/handlers.go#L83).

tortoisebeacon/tortoise_beacon.go
nkryuchkov: @dshulyak The for loop finishes after the last following message is sent, i.e. at the beginning of the last round (sending is done in a goroutine in `sendFollowingVotesLoopIteration`). Then `waitAfterLastRoundStarted` waits for one round until the last round is definitely finished to be sure that all last round votes are received before starting the beacon calculation.

tortoisebeacon/handlers.go
dshulyak: great, i will leave it as is then for this PR. i thought that needs to be done because you mentioned it previously

tortoisebeacon/tortoise_beacon.go
nkryuchkov: Actually, something like `waitUntilLastRoundFinished` would be a much better and less confusing name.

dshulyak: sorry right, i changed that. now the loop always sends messages, then waits for voting and weak coin duration, completes round and only then exits. so it wasn't waiting twice before.

nkryuchkov: Yeah, waiting in a loop looks like a clearer solution

cmd/node/node.go
```diff
@@ -554,11 +550,16 @@ func (app *SpacemeshApp) initServices(ctx context.Context,
 	atxdb := activation.NewDB(atxdbstore, idStore, mdb, layersPerEpoch, goldenATXID, validator, app.addLogger(AtxDbLogger, lg))
 	tBeaconDB := tortoisebeacon.NewDB(tBeaconDBStore, app.addLogger(TBeaconDbLogger, lg))
 
-	// TODO(nkryuchkov): Enable weak coin when finished.
-	// wc := weakcoin.NewWeakCoin(weakcoin.DefaultThreshold(), nodeID, swarm, signing.VRFVerify, vrfSigner, app.addLogger(WeakCoinLogger, lg))
-	wc := weakcoin.ValueMock{Value: false}
+	wcConfig := weakcoin.DefaultConfig()
```
nkryuchkov: What do you think of something like this?
```suggestion
	wcConfig := weakcoin.DefaultConfig(weakcoin.WithMaxRound(app.Config.TortoiseBeacon.RoundsNumber))
```

cmd/node/node.go
```diff
@@ -554,11 +550,16 @@ func (app *SpacemeshApp) initServices(ctx context.Context,
...
+	wcConfig.MaxRound = types.RoundID(app.Config.TortoiseBeacon.RoundsNumber)
+	wc := weakcoin.New(swarm,
+		vrfSigner, signing.VRFVerifier{},
```
nkryuchkov, line 554: I'd add a constructor for `signing.VRFVerifier`.

```suggestion
		vrfSigner, signing.NewVRFVerifier(),
```

cmd/node/node.go
```diff
@@ -554,11 +550,16 @@ func (app *SpacemeshApp) initServices(ctx context.Context,
...
+		weakcoin.WithLog(app.addLogger(WeakCoinLogger, lg)),
+		weakcoin.WithConfig(wcConfig),
```
nkryuchkov: Great change!

cmd/node/node.go
```diff
@@ -554,11 +550,16 @@ func (app *SpacemeshApp) initServices(ctx context.Context,
...
+	)
+
 	ld := time.Duration(app.Config.LayerDurationSec) * time.Second
-	tBeacon := tortoisebeacon.New(app.Config.TortoiseBeacon, nodeID, ld, swarm, atxdb, tBeaconDB, sgn, signing.VRFVerify, vrfSigner, wc, clock, app.addLogger(TBeaconLogger, lg))
+	tBeacon := tortoisebeacon.New(app.Config.TortoiseBeacon, nodeID, ld, swarm, atxdb, tBeaconDB, sgn, signing.VRFVerifier{}, vrfSigner, wc, clock, app.addLogger(TBeaconLogger, lg))
```
nkryuchkov, line 560:
```suggestion
	tBeacon := tortoisebeacon.New(app.Config.TortoiseBeacon, nodeID, ld, swarm, atxdb, tBeaconDB, sgn, signing.NewVRFVerifier(), vrfSigner, wc, clock, app.addLogger(TBeaconLogger, lg))
```

signing/interfaces.go
```diff
@@ -0,0 +1,15 @@
+package signing
+
+//go:generate mockgen -package=mocks -destination=./mocks/mocks.go -source=./interfaces.go Signer,Verifier
+
+// Signer ...
```
nkryuchkov: Please add a comment

signing/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+type Signer interface {
+	Sign([]byte) []byte
+	PublicKey() *PublicKey
+}
+
+// Verifier ...
```
nkryuchkov: Please add a comment

tortoisebeacon/votes_calc.go
```diff
@@ -165,11 +164,7 @@ func (tb *TortoiseBeacon) calcOwnFirstRoundVotes(epoch types.EpochID, votesMargi
 		case margin <= -votingThreshold:
 			ownFirstRoundsVotes.InvalidVotes[vote] = struct{}{}
 		default:
-			if coinToss {
-				ownFirstRoundsVotes.ValidVotes[vote] = struct{}{}
-			} else {
-				ownFirstRoundsVotes.InvalidVotes[vote] = struct{}{}
-			}
+			ownFirstRoundsVotes.InvalidVotes[vote] = struct{}{}
```
nkryuchkov, line 165: As `case margin <= -votingThreshold:` and `default:` have the same code, the former can be removed:
```go
	for vote, margin := range votesMargin {
		switch {
		case margin >= votingThreshold:
			ownFirstRoundsVotes.ValidVotes[vote] = struct{}{}
		default:
			ownFirstRoundsVotes.InvalidVotes[vote] = struct{}{}
		}
	}
```

tortoisebeacon/weakcoin/handler.go
```diff
@@ -0,0 +1,67 @@
...
+func (wc *WeakCoin) HandleSerializedMessage(ctx context.Context, data service.GossipMessage, sync service.Fetcher) {
+	wc.logger.With().Debug("received weak coin message",
+		log.String("from", data.Sender().String()),
+		log.Binary("data", data.Bytes()),
+	)
+
+	var message Message
+	if err := types.BytesToInterface(data.Bytes(), &message); err != nil {
+		wc.logger.With().Debug("received invalid weak coin message",
+			log.Binary("message", data.Bytes()),
```
nkryuchkov, line 22: What do you think of encoding to hex instead of base64? Hex is a bit human-readable because it's easy to detect null-bytes or numbers

tortoisebeacon/tortoise_beacon.go
```diff
@@ -577,34 +585,21 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
-			log.Uint64("epoch_id", uint64(epoch)))
-
-		if err := tb.sendVotes(firstVoteCtx, epoch, firstRound); err != nil {
-			tb.Log.With().Error("Failed to send first voting message",
-				log.Uint64("epoch_id", uint64(epoch)),
-				log.Err(err))
+	// we need to pass a map with spacetime unit allowances before any round is started
+	_, atxs, err := tb.atxDB.GetEpochWeight(epoch)
+	if err != nil {
+		tb.Log.With().Panic("can't load list of atxs", log.Err(err))
```
nkryuchkov, line 591: Why do we need to panic here?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -577,34 +585,21 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
-		return
-	case <-firstVoteCtx.Done():
-		break
+		ua[string(header.NodeID.VRFPublicKey)] += uint64(header.NumUnits)
 	}
-
-	tb.Log.With().Debug("Starting following voting message sender",
-		log.Uint64("epoch_id", uint64(epoch)))
+	tb.weakCoin.StartEpoch(epoch, ua)
+	defer tb.weakCoin.CompleteEpoch()
```
nkryuchkov: What do you think of `FinishEpoch`?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -613,25 +608,27 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
 	ticker := time.NewTicker(tb.votingRoundDuration + tb.weakCoinRoundDuration)
 	defer ticker.Stop()
 
-	tb.sendFollowingVotesLoopIteration(ctx, epoch, firstRound+1)
+	for round := firstRound; round <= tb.lastPossibleRound(); round++ {
+		go tb.startWeakCoin(ctx, epoch, round)
```
nkryuchkov: We've already discussed the change here, but I'm leaving a comment not to forget about it.

tortoisebeacon/tortoise_beacon.go
```diff
@@ -613,25 +608,27 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
+		go func(epoch types.EpochID, round types.RoundID) {
+			if err := tb.sendVotes(ctx, epoch, round); err != nil {
+				tb.Log.With().Error("Failed to send voting messages",
+					log.Uint64("epoch_id", uint64(epoch)),
```
nkryuchkov: It's not a scope of your PR, but as `types.EpochID` is `uint32` now, and as you're touching this, it can be changed to `uint32`
```suggestion
					log.Uint32("epoch_id", uint32(epoch)),
```

tortoisebeacon/weakcoin/handler.go
```diff
@@ -0,0 +1,67 @@
...
+			log.Uint32("epoch_id", uint32(message.Epoch)),
+			log.Uint64("round_id", uint64(message.Round)))
+		return
+	}
+	data.ReportValidation(ctx, GossipProtocol)
+}
+
+func (wc *WeakCoin) receiveMessage(message Message) error {
+	if wc.exceedsThreshold(message.Signature) {
+		return fmt.Errorf("proposal %x exceeds threshold", message.Signature)
```
nkryuchkov: As you're touching this, could you please negate the logic of this function and rename it to `passesThreshold`? There was the same change in the `tortoisebeacon` package: https://github.com/dshulyak/go-spacemesh/blob/finalize-weakcoin/tortoisebeacon/tortoise_beacon.go#L942

```suggestion
	if !wc.passesThreshold(message.Signature) {
		return fmt.Errorf("proposal %x doesn't pass threshold", message.Signature)
```

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
+	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
 const (
 	// Prefix defines weak coin proposal prefix.
 	proposalPrefix = "WeakCoin"
 	// GossipProtocol is weak coin Gossip protocol name.
 	GossipProtocol = "WeakCoinGossip"
+
+	defaultThreshold = "0x8000000000000000000000000000000000000000000000000000000000000000"
```
nkryuchkov, line 24: Could you please note that it's `2^256 / 2`? It's not easy to count zeroes here

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
 )
 
-type epochRoundPair struct {
-	EpochID types.EpochID
-	Round   types.RoundID
+// DefaultConfig ...
```
nkryuchkov: Could you please add a comment?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
+func DefaultConfig() Config {
+	return Config{
+		Threshold:           util.FromHex(defaultThreshold),
+		VRFPrefix:           proposalPrefix,
+		NextRoundBufferSize: 10000, // ~1mb given the size of Message is ~100b
+		MaxRound:            100,
+	}
+}
+
+// Config ...
```
nkryuchkov: Could you please add a comment?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
-type epochRoundPair struct {
-	EpochID types.EpochID
-	Round   types.RoundID
+// DefaultConfig ...
+func DefaultConfig() Config {
+	return Config{
+		Threshold:           util.FromHex(defaultThreshold),
+		VRFPrefix:           proposalPrefix,
+		NextRoundBufferSize: 10000, // ~1mb given the size of Message is ~100b
+		MaxRound:            100,
```
nkryuchkov: Why is it 100?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
+				log.Err(err),
+				log.Uint32("epoch_id", uint32(msg.Epoch)),
+				log.Uint64("round_id", uint64(msg.Round)))
+		}
+		wc.nextRoundBuffer[i].Epoch = 0
+		wc.nextRoundBuffer[i].Round = 0
+		wc.nextRoundBuffer[i].Unit = 0
+		wc.nextRoundBuffer[i].Signature = nil
+	}
+	wc.nextRoundBuffer = wc.nextRoundBuffer[:0]
```
nkryuchkov, line 187: Why do you use `wc.nextRoundBuffer[:0]` instead of `make`'ing a new one?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
+	return wc.publishProposal(ctx, epoch, round)
+}
 
-	proposal, err := wc.generateProposal(epoch, round)
+func (wc *WeakCoin) updateProposal(message Message) error {
+	buf := wc.encodeProposal(message.Epoch, message.Round, message.Unit)
+	miner, err := wc.verifier.Extract(buf, message.Signature)
 	if err != nil {
-		return err
+		return fmt.Errorf("can't recover miner id from signature: %w", err)
```
nkryuchkov: Do we need to log signature (maybe debug level) here to know what failed to extract?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
+	if wc.smallestProposal == nil {
+		wc.logger.With().Error("completed round without valid proposals",
+			log.Uint32("epoch_id", uint32(wc.epoch)),
+			log.Uint64("round_id", uint64(wc.round)),
+		)
+		return
 	}
-
-	return proposal.Bytes(), nil
+	coinflip := wc.smallestProposal[0]&1 == 1
```
nkryuchkov, line 283: The weak coin should be the least significant bit of the signature. Why do you assume that a signature is always stored as a little-endian number?

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
-	case 0:
-		return false
-	case 1:
-		return true
-	default:
-		panic("bad input")
+	buf := make([]byte, 8)
+	binary.LittleEndian.PutUint64(buf, unit)
+	if _, err := proposal.Write(buf); err != nil {
+		wc.logger.Panic("can't write uint16 to a buffer", log.Err(err))
```
nkryuchkov: I guess you meant `uint64` here
```suggestion
		wc.logger.Panic("can't write uint64 to a buffer", log.Err(err))
```

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
 
-	proposal, err := wc.generateProposal(epoch, round)
+func (wc *WeakCoin) updateProposal(message Message) error {
+	buf := wc.encodeProposal(message.Epoch, message.Round, message.Unit)
+	miner, err := wc.verifier.Extract(buf, message.Signature)
 	if err != nil {
-		return err
+		return fmt.Errorf("can't recover miner id from signature: %w", err)
 	}
+	allowed, exists := wc.allowances[string(miner.Bytes())]
```
nkryuchkov, line 199: Could you please add new lines after error checks? So the pieces of logic would be split and be easier to read

nkryuchkov (commented):
Great job! The code looks much better than it used to

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: i am not sure that i understand the question. signature doesn't depend on the architecture in any way. the representation is always the same

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: it is cheaper and safe

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: it is just random number. it doesn't have to be a valid number for spacemesh protocol. valid number must be configured during initalization

tortoisebeacon/tortoise_beacon.go
dshulyak: what are you referring to? i wasn't planning to make other changes here

cmd/node/node.go
dshulyak: but this is already a valid initializer. why do you want to have a function?

cmd/node/node.go
dshulyak: i don't like adding options to config. it would be a bit messy. i can get rid from config completely and just pass WithMaxRound to weakcoin.New

tortoisebeacon/tortoise_beacon.go
nkryuchkov: Sorry, I didn't check the body of `tb.startWeakCoin` before leaving this comment and thought that you start weak coin before voting. What do you think of putting `go tb.startWeakCoin(ctx, epoch, round)` below the goroutine below to show that its actual code is run after sending votes? Just for readability

tortoisebeacon/tortoise_beacon.go
dshulyak: because nothing can be done if local database fails. i am not aware of any temporary errors that will be fixed by themself

tortoisebeacon/tortoise_beacon.go
dshulyak: could be finish if your prefer

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: i mean making a new one is safe too obviously. but if we can reuse without any complexity then why not?

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: i guess your question was about how library that produces signature treats them (as big endian or little endian). i haven't thought about this. but found https://datatracker.ietf.org/doc/html/draft-josefsson-eddsa-ed25519-03 Section 
5.2.  Encoding

>  All values are coded as octet strings, and integers are coded using
   little endian convention.  I.e., a 32-octet string h h[0],...h[31]
   represents the integer h[0] + 2^8 h[1] + ... + 2^248 h[31].

dshulyak: it may become invalid if we will switch to another library. but idk if it is worth it to add it to the interface of the signer/verifier

nkryuchkov: > i am not sure that i understand the question. signature doesn't depend on the architecture in any way. the representation is always the same

I meant that to extract an LSB from bytes, we need to get `wc.smallestProposal[0]` if a proposal is stored as a little-endian number and `wc.smallestProposal[len(wc.smallestProposal)-1]` if the proposal is stored as a big-endian number. So now we assume that a proposal is a little-endian number. Of course, if we get not LSB but some other byte consistently on all nodes, the algorithm would work. But I'd add a comment for this.

nkryuchkov: > i guess your question was about how library that produces signature treats them (as big endian or little endian)

Yes, that's what I meant. Thanks for checking. So it would work fine. Could you please add a comment with something like this?

```suggestion
	// ED25519 assumes that byte order is little-endian. For another algorithm, the endianness might differ.
	coinflip := wc.smallestProposal[0]&1 == 1
```

cmd/node/node.go
nkryuchkov: I just prefer initializing via a constructor function if some struct with methods is created. But it's a matter of taste

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: I thought that using `make` would have less cognitive complexity. But your choice is fine if you prefer it

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+	bc := mocks.NewMockbroadcaster(ctrl)
+	bc.EXPECT().Broadcast(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes().Return(nil)
+	return bc
+}
 
-	var wc weakCoin
-	epoch := types.EpochID(3)
-	round := types.RoundID(1)
-	expected := "5765616b436f696e03000000000000000100000000000000"
+func broadcastedMessage(tb testing.TB, ctrl *gomock.Controller, msg weakcoin.Message) *servicemocks.MockGossipMessage {
```
nkryuchkov, line 28: I think `tb.Helper()` might be added to the beginning of the function

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+	require.NoError(t, wc.StartRound(context.TODO(), nextRound))
+	wc.CompleteRound()
+	require.True(t, wc.Get(epoch, nextRound))
+}
 
-	for _, tc := range tt {
-		tc := tc
-		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
+func TestWeakCoinExchangeProposals(t *testing.T) {
```
nkryuchkov, line 367: Can we add `t.Parallel()` to the beginning?

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
-		{
-			name:          "Panic",
-			coinBit:       big.NewInt(3),
-			expectedPanic: true,
-		},
+	wc.StartEpoch(epoch, nil)
+	require.False(t, wc.Get(epoch, round))
+}
+
+func TestWeakCoinNextRoundBufferOverflow(t *testing.T) {
```
nkryuchkov, line 293: Can we add `t.Parallel()` to the beginning?

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+	verifier.EXPECT().Extract(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(msg, sig []byte) (*signing.PublicKey, error) {
+			return signing.NewPublicKey(sig), nil
+		}).AnyTimes()
+	return verifier
 }
 
-func Test_weakCoin_proposalExceedsThreshold(t *testing.T) {
-	t.Parallel()
+func TestWeakCoin(t *testing.T) {
```
nkryuchkov, line 57: Can we add `t.Parallel()` to the beginning?

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+	"github.com/spacemeshos/go-spacemesh/tortoisebeacon/weakcoin/mocks"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func Test_weakCoin_generateProposal(t *testing.T) {
-	t.Parallel()
-
-	r := require.New(t)
+func noopBroadcaster(tb testing.TB, ctrl *gomock.Controller) *mocks.Mockbroadcaster {
```
nkryuchkov, line 22: I think `tb.Helper()` might be added to the beginning of the function

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+	buf, err := types.InterfaceToBytes(&msg)
+	require.NoError(tb, err)
+	mm.EXPECT().Bytes().Return(buf).AnyTimes()
+	mm.EXPECT().ReportValidation(gomock.Any(), gomock.Any()).AnyTimes()
+	return mm
+}
 
-	p, err := wc.generateProposal(epoch, round)
-	r.NoError(err)
+func staticSigner(tb testing.TB, ctrl *gomock.Controller, sig []byte) *smocks.MockSigner {
```
nkryuchkov, line 39: `tb.Helper()`

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+	signer := smocks.NewMockSigner(ctrl)
+	signer.EXPECT().Sign(gomock.Any()).Return(sig).AnyTimes()
+	signer.EXPECT().PublicKey().Return(signing.NewPublicKey(sig)).AnyTimes()
+	return signer
+}
 
-	r.EqualValues(expected, hex.EncodeToString(p))
+func pubkeyFromSigVerifier(tb testing.TB, ctrl *gomock.Controller) *smocks.MockVerifier {
```
nkryuchkov, line 47: `tb.Helper()`

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+
+			require.Equal(t, tc.coinflip, wc.Get(tc.startedEpoch, tc.startedRound+1))
+			wc.CompleteEpoch()
+		})
+	}
+	for _, tc := range tcs {
 		tc := tc
-		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
+		t.Run("BufferingNextEpochAfterCompletion/"+tc.desc, func(t *testing.T) {
```
nkryuchkov, line 249: `t.Parallel()`

tortoisebeacon/weakcoin/weak_coin_test.go
```diff
@@ -1,118 +1,378 @@
...
+
+			require.Equal(t, tc.coinflip, wc.Get(tc.startedEpoch+1, tc.startedRound))
+			wc.CompleteEpoch()
 		})
 	}
 }
 
-func Test_weakCoin_coinBitToBool(t *testing.T) {
-	t.Parallel()
+func TestWeakCoinGetPanic(t *testing.T) {
```
nkryuchkov, line 274: `t.Parallel()`

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,303 @@ import (
...
+	if !exists {
 		return nil
 	}
-
-	message := Message{
-		Epoch:        epoch,
-		Round:        round,
-		MinerID:      wc.minerID,
-		VRFSignature: signedProposal,
+	for unit := uint64(0); unit < allowed; unit++ {
```
nkryuchkov, line 216: What do you think of extracting a function for finding the smallest?

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: Maybe 300 then? https://github.com/dshulyak/go-spacemesh/blob/finalize-weakcoin/tortoisebeacon/config.go#L23

dshulyak:
bors try

tortoisebeacon/weakcoin/handler.go
dshulyak: the problem is that i don't want to encode always. otherwise someone can send any junk and waste node cpu encoding it to hex. with log.Binary cpu will be spent only if log level is Debug, which shouldn't be set to it on production node

tortoisebeacon/weakcoin/handler.go
dshulyak: i will rename but i don't think that it matters how simple helper is called in two different places

dshulyak: @nkryuchkov or do you mean that the signature must be above threshold now? should be strictly above or above or equal?

nkryuchkov: @dshulyak If the signature is strictly below the threshold, it passes the threshold and needs to be handled. If it's equal or above the threshold, it doesn't pass the threshold and should be rejected

dshulyak: i think it is very confusing to call it passes, when it actually means that it verifies that it actually needs to be below the threshold. exceeds is already better, but it can be named even if `aboveThreshold() { return err }`

nkryuchkov: It was a request from the research team to call it "passes". But we can leave it as "exceeds"

tortoisebeacon/weakcoin/weak_coin_test.go
dshulyak: i created single mocking controller instance. i don't know if it will result in any concurrency issues. but the difference in run time is nelgibile. 0.110 s vs 0.100 s

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

signing/vrfsigner.go
```diff
@@ -27,10 +35,29 @@ func NewVRFSigner(seed []byte) (*VRFSigner, []byte, error) {
...
 }
 
 // VRFVerify verifies a message and signature, given a public key
 func VRFVerify(pub, msg, sig []byte) bool {
 	return ed25519.Verify(pub, msg, sig)
 }
+
+var _ Verifier = VRFVerifier{}
+
+// VRFVerifier ...
```
nkryuchkov: Could you please add a comment here?

tortoisebeacon/weakcoin/handler.go
```diff
@@ -0,0 +1,67 @@
...
+		if wc.isNextRound(message.Epoch, message.Round) && len(wc.nextRoundBuffer) < cap(wc.nextRoundBuffer) {
+			wc.nextRoundBuffer = append(wc.nextRoundBuffer, message)
+			return nil
+		}
+		return fmt.Errorf("message for the wrong round %v/%v", message.Epoch, message.Round)
+	}
+	return wc.updateProposal(message)
+}
+
+func (wc *WeakCoin) isNextRound(epoch types.EpochID, round types.RoundID) bool {
```
nkryuchkov, line 54: What do you think of adding a test for this method? There're some calculations here and a test could help ensure that they are correct

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,325 @@ import (
...
+		if err := wc.updateProposal(msg); err != nil {
+			wc.logger.With().Debug("received invalid proposal",
+				log.Err(err),
+				log.Uint32("epoch_id", uint32(msg.Epoch)),
+				log.Uint64("round_id", uint64(msg.Round)))
+		}
+		wc.nextRoundBuffer[i].Epoch = 0
+		wc.nextRoundBuffer[i].Round = 0
+		wc.nextRoundBuffer[i].Unit = 0
+		wc.nextRoundBuffer[i].Signature = nil
```
nkryuchkov: Why did you prefer this over the code below?
```suggestion
		wc.nextRoundBuffer[i] = Message{}
```

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -4,290 +4,325 @@ import (
...
-func (wc *weakCoin) generateProposal(epoch types.EpochID, round types.RoundID) ([]byte, error) {
-	proposal := bytes.Buffer{}
-
-	proposal.WriteString(proposalPrefix)
-
-	epochBuf := make([]byte, 8)
-	binary.LittleEndian.PutUint64(epochBuf, uint64(epoch))
-	if _, err := proposal.Write(epochBuf); err != nil {
-		return nil, fmt.Errorf("write epoch to buffer: %w", err)
+func (wc *WeakCoin) prepareProposal(epoch types.EpochID, round types.RoundID) (broadcast, smallest []byte) {
```
nkryuchkov, line 208: Do we need a unit test for this?

tortoisebeacon/weakcoin/weak_coin_test.go
nkryuchkov: Let's leave it as it is then

tortoisebeacon/weakcoin/handler.go
dshulyak: this method is tested. the coverage is close to 100. creating unit tests for non public methods is not a good practice and should be avoided.

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: do you know if this will allocate a new memory or overwrite existing? doesn't matter much to me, but this is what i was thinking when made that change

tortoisebeacon/weakcoin/weak_coin.go
dshulyak: this code is tested. i can add regression test to verify that encoding doesn't change

signing/vrfsigner.go
dshulyak: added

tortoisebeacon/weakcoin/weak_coin.go
nkryuchkov: It won't allocate anything because of no pointers. It will do `runtime.typedmemclr`.

nkryuchkov (approved):

dshulyak:
bors merge

PR #2608, opened 2021-07-28
[Merged by Bors] - Fix most data races

### LINKED ISSUE #2226: Add -race flag to tests and fix data races
## Description
Neither unit tests, nor system tests use the `-race` flag. Using it would enable Go's race detector and let us find some existing data races easily. As a result of a read/write operation is undefined in case of a data race, they might cause some unexpected behavior.

I ran unit tests with the flag, and there are many races in multiple packages.

`f790a44`

https://github.com/spacemeshos/go-spacemesh/blob/develop/Makefile#L137
https://github.com/spacemeshos/go-spacemesh/blob/develop/Makefile#L142
https://github.com/spacemeshos/go-spacemesh/blob/develop/Makefile#L142
https://github.com/spacemeshos/go-spacemesh/blob/develop/Dockerfile#L31

nkryuchkov:
## Motivation

WIP on #2226

Depends on https://github.com/spacemeshos/post/pull/45

Some data races still exist and are triggered by logs that appear after tests have finished their execution. This still doesn't allow to enable the `-race` flag. However, most issues in the non-test code seem fixed. 

## Changes

- Fix most data races

## Test Plan

- Unit tests
- System tests

## DevOps Notes


cmd/base.go
```diff
@@ -29,6 +30,9 @@ var (
 	// Commit is the git commit used to build the app. Designed to be overwritten by make.
 	Commit string
 
+	// Mu is the mutex for accessing global variables in this package.
+	Mu sync.RWMutex
```
dshulyak: i don't like introducing global mutex. it was a mistake to introduce global context and cancel function in the first place, maybe it is possible to refactor code to avoid global context? or it is too many changes?

hare/algorithm.go
```diff
@@ -85,7 +87,7 @@ var _ TerminationOutput = (*procReport)(nil)
 
 // State holds the current state of the consensus process (aka the participant).
 type State struct {
-	k           uint32       // the round counter (k%4 is the round number)
+	k           uint32       // the round counter (k%4 is the round number); it should be first in struct for alignment because atomics are used
```
dshulyak, line 81: it is better to understand when k can be updated and read concurrently. the race here might be the symptom of the real issue

hare/algorithm.go
```diff
@@ -844,49 +851,57 @@ func (proc *consensusProcess) shouldParticipate(ctx context.Context) bool {
...
+		k, expectedCommitteeSize(k, proc.cfg.N, proc.cfg.ExpectedLeaders), proc.nid, proof)
 	if err != nil {
-		logger.With().Error("failed to check eligibility", log.Err(err), types.LayerID(proc.instanceID))
+		logger.With().Error("failed to check eligibility", log.Err(err), proc.instanceID)
 		return passive
 	}
 
+	proc.mu.Lock()
 	proc.eligibilityCount = eligibilityCount
+	proc.mu.Unlock()
```
dshulyak, line 898: also not clear how the eligibility count can be read and updated concurrently

p2p/connectionpool/connectionpool.go
```diff
@@ -224,7 +232,10 @@ func (cp *ConnectionPool) GetConnection(ctx context.Context, address inet.Addr,
 	cp.pending[remotePub] = append(cp.pending[remotePub], pendChan)
 	if !found {
 		// No one is waiting for a connection with the remote peer, need to call Dial
-		cp.dialWait.Add(1)
+		cp.dialWaitMu.Lock()
+		cp.dialWait.Add(1) //  The first increment must be synchronized with Wait.
```
dshulyak: what if Wait completes and we call Add(1) afterwards? i think this is the bug and we need to check if connection pool is closed before Add(1)

tortoise/verifying_tortoise_test.go
```diff
@@ -2476,37 +2476,37 @@ func TestVectorArithmetic(t *testing.T) {
...
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 1, 1, 1))
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 10, 1, 1))
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 1, 10, 1))
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 1, 1, 10))
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 10, 1, 10))
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 10, 10, 1))
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 1, 10, 10))
-	r.Equal(abstain, calculateOpinionWithThreshold(log.AppLog, against.Multiply(10), 10, 10, 10))
-	r.Equal(against, calculateOpinionWithThreshold(log.AppLog, against.Multiply(11), 10, 10, 10))
+	r.Equal(abstain, calculateOpinionWithThreshold(log.GetLogger(), abstain, 1, 1, 1))
```
dshulyak: logtest.New(t)

dshulyak (commented):
there are several places where race is probably a symptom of the bug. for example if global context is overwritten and we call cancel on older context. in hare there are several suspicious places as well. i think it is better to understand why it can be called concurrently and reduce usage of shared memory, and only if the this is not possible then wrap everything in mutex.

otherwise we will simply hide the issues without addressing them

cmd/base.go
nkryuchkov: not too many changes, fixed

hare/algorithm.go
nkryuchkov: IIRC, k was concurrently updated in the test, therefore data races occurred

nkryuchkov: 1. 
```
Write at 0x00c000120318 by goroutine 25:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).addToK()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:927 +0x84
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).advanceToNextRound()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:545 +0x58
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:337 +0xf66
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).Start·dwrap·1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:253 +0x58

Previous read at 0x00c000120318 by goroutine 26:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).getK()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:917 +0x479
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).sendMessage()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:503 +0x4e0
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop.func1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:300 +0x284

```

2. 
```
Write at 0x00c000266318 by goroutine 170:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).addToK()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:927 +0x84
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).advanceToNextRound()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:545 +0x58
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:337 +0xf66
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).Start·dwrap·1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:253 +0x58

Previous read at 0x00c000266318 by goroutine 173:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).getK()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:917 +0x33e
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).shouldParticipate()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:866 +0x3a5
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop.func1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:292 +0x85
```

hare/algorithm.go
nkryuchkov:
```
Write at 0x00c00019e8ea by goroutine 174:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).currentRole()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:903 +0x5a8
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).shouldParticipate()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:864 +0x325
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).beginStatusRound()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:559 +0x2f7
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).onRoundBegin()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:694 +0x8d
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:341 +0xf84
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).Start·dwrap·1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:254 +0x58

Previous write at 0x00c00019e8ea by goroutine 193:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).currentRole()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:903 +0x5a8
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).shouldParticipate()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:864 +0x325
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop.func1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:293 +0x85
```

hare/algorithm.go
nkryuchkov: @dshulyak 

It looks like the race detector also detects unsynchronized access at different times when there's actually no data race but there's a potential of a data race when timing differs. E. g., I increased the timeout in `TestConsensusProcess_TerminationLimit` to 10 seconds, and here's how logs look like:
```
==================
WARNING: DATA RACE
Write at 0x00c0002ce018 by goroutine 19:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).addToK()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:939 +0x84
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).advanceToNextRound()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:558 +0x58
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:344 +0x10d0
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).Start·dwrap·1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:253 +0x58

Previous read at 0x00c0002ce018 by goroutine 20:
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).getK()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:929 +0x479
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).sendMessage()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:516 +0x4e0
  github.com/spacemeshos/go-spacemesh/hare.(*consensusProcess).eventLoop.func1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/hare/algorithm.go:304 +0x325
[...]
==================
sendMessage started at 2021-10-07 00:22:26.019217 +0300 MSK m=+0.073492008
[...]
sendMessage finished at 2021-10-07 00:22:26.019383 +0300 MSK m=+0.073658050
advanceToNextRound started at 2021-10-07 00:22:36.01865 +0300 MSK m=+10.072839518
advanceToNextRound finished at 2021-10-07 00:22:36.019665 +0300 MSK m=+10.073854166
```

So, here `advanceToNextRound` starts 10 seconds after `sendMessage` finishes, but the race detector still throws an error. I also tried a 1-minute timeout, the results are the same.

As I understand, the `ThreadSanitizer` library which is used as the Go race detector, logs every `atomic` operation by calling `__tsan_go_atomic32_load` (https://github.com/golang/go/blob/master/src/runtime/race_amd64.s#L200). Regarding mutexes, the compiler calls `__tsan_acquire` (https://github.com/golang/go/blob/master/src/sync/mutex.go#L76, https://github.com/golang/go/blob/master/src/runtime/race.go#L520) which establishes some relationships of what happens before what (https://github.com/golang/go/blob/master/src/runtime/race.go#L30) which ensures the lack of data race. So, what's not explicitly reported to TSAN is considered a data race.

I think we should just make the race detector happy and leave the access synchronized. If we need to ensure that the operations aren't done simultaneously, we should consider using other options

hare/algorithm.go
nkryuchkov: https://github.com/spacemeshos/go-spacemesh/pull/2608#discussion_r723713505

p2p/connectionpool/connectionpool.go
nkryuchkov: @dshulyak Could you please check if https://github.com/spacemeshos/go-spacemesh/pull/2608/commits/355c210289d00403834bc736a7239c0680b9be43 is what you meant?

nkryuchkov:
Tests should pass after https://github.com/spacemeshos/post/pull/45 is merged

activation/activation.go
```diff
@@ -137,7 +137,7 @@ type Builder struct {
 	// pendingATX is created with current commitment and nipst from current challenge.
 	pendingATX            *types.ActivationTx
 	layerClock            layerClock
-	mtx                   sync.Mutex
+	mu                    sync.RWMutex
```
countvonzero: is there any particular reason to use RWMutex instead of Mutex? 

in my view, this is one form of premature optimization. RWMutex is suitable when there are more reads than writes. so in case where we only write, we should simply use Mutex for the following reasons:
- there is a slight performance penalty, but i can ignore that
- i have seen in huge codebase, debugging become more difficult with different kinds of lock and multiple lock holders.

i opt for simplicity for now and only use RWMutex when there are indeed a lot more readers than writers

activation/post.go
```diff
@@ -68,23 +68,26 @@ var _ PostSetupProvider = (*PostSetupManager)(nil)
...
 type PostSetupManager struct {
+	mu sync.RWMutex
+
 	id     []byte
 	cfg    PostConfig
 	logger log.Log
 
-	stopMtx       sync.Mutex
-	initStatusMtx sync.Mutex
+	// stopMu  sync.Mutex
```
countvonzero: rm

activation/post.go
```diff
@@ -299,11 +311,11 @@ func (mgr *PostSetupManager) StartSession(opts PostSetupOpts) (chan struct{}, er
 // StopSession stops the current Post setup data creation session
 // and optionally attempts to delete the data file(s).
 func (mgr *PostSetupManager) StopSession(deleteFiles bool) error {
-	mgr.stopMtx.Lock()
-	defer mgr.stopMtx.Unlock()
+	// mgr.stopMu.Lock()
+	// defer mgr.stopMu.Unlock()
```
countvonzero: rm

activation/post.go
```diff
@@ -130,14 +133,14 @@ var errNotComplete = errors.New("not complete")
 // Status returns the setup current status.
 func (mgr *PostSetupManager) Status() *PostSetupStatus {
```
countvonzero, line 130: does it make more sense to just get the lock once for this method, instead of 4 times, especially if this method is called frequently?

activation/post.go
```diff
@@ -213,25 +221,21 @@ func (mgr *PostSetupManager) Benchmark(p PostSetupComputeProvider) (int, error)
 // It supports resuming a previously started session, as well as changing the Post setup options (e.g., number of units)
 // after initial setup.
 func (mgr *PostSetupManager) StartSession(opts PostSetupOpts) (chan struct{}, error) {
```
countvonzero, line 227: i think it make more sense to obtain the lock once for this method.
for simplicity, i'll even venture to use just 1 lock for everything in this file.

hare/consensus_test.go
```diff
@@ -174,7 +175,7 @@ func TestConsensusFixedOracle(t *testing.T) {
 	}
 	test.Create(totalNodes, creationFunc)
 	test.Start()
-	test.WaitForTimedTermination(t, 30*time.Second)
+	test.WaitForTimedTermination(t, 40*time.Second)
```
countvonzero: do you know why this requires extra time?

hare tests already take too long as is, now we are adding 10 seconds in many more tests?

countvonzero (commented):
this is one tough PR to create. kudos to your effort.

enabling --race for tests is very important. i wholeheartedly support this effort.
however, i feel this blanketed approach with mutex is excessive. there are other tools such as making immutable struct (lower-case struct members+data hiding) or the atomic package.

i guess this is a good first step to force --race on all tests. but the code is unnecessarily bloated IMO.

nkryuchkov:
@countvonzero 

Generally, I agree with your point of view. Most data races were fixed by using a mutex or atomics and their usage may look excessive in some places. However, this approach works, it gets rid of data races and is relatively easy to test. 

Deeper modifications would look better but they require much more time spent on the PR, we will need to resolve new merge conflicts constantly and will be having data races in `develop` until the PR is merged. Also, deeper modifications should be tested much better than just adding mutexes and atomics. A lot of packages are modified, so we'll need to test all parts of the projects, which is very risky and time-consuming.

In my opinion, it would be better for us to have smaller and shorter-living PRs with the least scope of changes, and to use an iterative approach, and to merge this PR as it is (after similar PR to PoET is merged). It's already better than what we have now in `develop`. We can continue making improvements in the next PRs.

activation/activation.go
nkryuchkov: You're right, `sync.RWMutex` is a premature optimization here, I'll change it to `sync.Mutex`

activation/post.go
nkryuchkov: thanks, I forgot to remove it

activation/post.go
nkryuchkov: I tried to reduce the number of mutex locks, but tests hang then. I tried to fix hanging but was able to do that only by reverting the number of locks. I think we need to leave it as is now and make deeper changes to the package and tests in the future

activation/post.go
nkryuchkov: https://github.com/spacemeshos/go-spacemesh/pull/2608#discussion_r727415921

countvonzero:
> @countvonzero
> 
> Generally, I agree with your point of view. Most data races were fixed by using a mutex or atomics and their usage may look excessive in some places. However, this approach works, it gets rid of data races and is relatively easy to test.
> 
> Deeper modifications would look better but they require much more time spent on the PR, we will need to resolve new merge conflicts constantly and will be having data races in `develop` until the PR is merged. Also, deeper modifications should be tested much better than just adding mutexes and atomics. A lot of packages are modified, so we'll need to test all parts of the projects, which is very risky and time-consuming.
> 
> In my opinion, it would be better for us to have smaller and shorter-living PRs with the least scope of changes, and to use an iterative approach, and to merge this PR as it is (after similar PR to PoET is merged). It's already better than what we have now in `develop`. We can continue making improvements in the next PRs.

i agree with all your stmts with the exception of the iterative approach. IMO, the iterative approach should be breaking down this issue into multiple packages and change them one at a time, instead of a blanketed change like this, which suffers exactly what you pointed out, never-ending merges.

what i am going to say is a little brutal. please pardon my lack of diplomacy.

in some way i feel huge changes like this is effectively putting a knife on others' throat and say "hey, i already done this much work, you need to accept it in order to move forward". i strongly discourage big changes like this going forward.

on that note, i'll approve and let's move on.

countvonzero (approved):

nkryuchkov:
bors try

nkryuchkov:
bors merge

PR #2611, opened 2021-07-29
add tomb to manage goroutines for syncer

countvonzero:
## Motivation
POC: uses https://github.com/go-tomb/tomb/tree/v2 to manage goroutines for syncer

## Changes
- use tomb to manage goroutines for graceful shutdown
- instead of monitoring top-level context for cancellation, use tomb to signal shutdown

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -15,6 +15,7 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/mesh"
 	"github.com/spacemeshos/go-spacemesh/p2p/peers"
+	"gopkg.in/tomb.v2"
 )
 
 type layerTicker interface {
@@ -32,10 +33,12 @@ type layerFetcher interface {
 // Configuration is the config params for syncer
 type Configuration struct {
 	SyncInterval time.Duration
-	// the sync process will try validate the current layer if ValidationDelta has elapsed.
+	// the sync process will try to validate the current layer if ValidationDelta has elapsed.
 	ValidationDelta time.Duration
 	AlwaysListen    bool
 }
+
+var errShuttingDown = errors.New("shutting down")
 
 const (
 	outOfSyncThreshold  uint32 = 2 // see notSynced
@@ -91,16 +94,18 @@ type Syncer struct {
 	awaitSyncedCh []chan struct{}
 	awaitSyncedMu sync.Mutex
 
-	shutdownCtx context.Context
-	cancelFunc  context.CancelFunc
+	// tomb is used to manage goroutines for graceful shutdown.
+	// note that tomb kills itself after its last goroutine exits. syncer's loop is the long-living goroutine
+	// that keeps tomb alive before the node shuts down.
+	// see https://github.com/go-tomb/tomb/issues/28
+	tomb tomb.Tomb
 
 	// recording the run # since started. for logging/debugging only.
 	run uint64
 }
 
 // NewSyncer creates a new Syncer instance.
-func NewSyncer(ctx context.Context, conf Configuration, ticker layerTicker, mesh *mesh.Mesh, fetcher layerFetcher, logger log.Log) *Syncer {
-	shutdownCtx, cancel := context.WithCancel(ctx)
+func NewSyncer(conf Configuration, ticker layerTicker, mesh *mesh.Mesh, fetcher layerFetcher, logger log.Log) *Syncer {
 	return &Syncer{
 		logger:            logger,
 		conf:              conf,
@@ -111,15 +116,16 @@ func NewSyncer(ctx context.Context, conf Configuration, ticker layerTicker, mesh
 		syncTimer:         time.NewTicker(conf.SyncInterval),
 		targetSyncedLayer: unsafe.Pointer(&types.LayerID{}),
 		awaitSyncedCh:     make([]chan struct{}, 0),
-		shutdownCtx:       shutdownCtx,
-		cancelFunc:        cancel,
 	}
 }
 
 // Close stops the syncing process and the goroutines syncer spawns.
-func (s *Syncer) Close() {
-	// TODO: ensure goroutines are all terminated before shutting down
-	s.cancelFunc()
+func (s *Syncer) Close() error {
+	s.logger.Info("start terminating syncer goroutines")
+	s.tomb.Kill(errShuttingDown)
+	err := s.tomb.Wait()
+	s.logger.Info("all syncer goroutines terminated")
+	return err
 }
 
 // RegisterChForSynced registers ch for notification when the node enters synced state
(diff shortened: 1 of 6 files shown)


syncer/syncer.go
```diff
@@ -91,16 +94,18 @@ type Syncer struct {
 	awaitSyncedCh []chan struct{}
 	awaitSyncedMu sync.Mutex
 
-	shutdownCtx context.Context
-	cancelFunc  context.CancelFunc
+	// tomb is used to manage goroutines for graceful shutdown.
+	// note that tomb kills itself after its last goroutine exits. syncer's loop is the long-living goroutine
+	// that keeps tomb alive before the node shuts down.
+	// see https://github.com/go-tomb/tomb/issues/28
+	tomb tomb.Tomb
```
dshulyak, line 101: i think https://pkg.go.dev/golang.org/x/sync/errgroup doesn't have this issue. very popular and it also has a simpler API

syncer/syncer.go
```diff
@@ -111,15 +116,16 @@ func NewSyncer(ctx context.Context, conf Configuration, ticker layerTicker, mesh
...
 }
 
 // Close stops the syncing process and the goroutines syncer spawns.
-func (s *Syncer) Close() {
-	// TODO: ensure goroutines are all terminated before shutting down
-	s.cancelFunc()
+func (s *Syncer) Close() error {
+	s.logger.Info("start terminating syncer goroutines")
+	s.tomb.Kill(errShuttingDown)
+	err := s.tomb.Wait()
```
dshulyak, line 126: will it exit with errShuttingDown or with nil?

syncer/syncer.go
```diff
@@ -151,36 +157,37 @@ func (s *Syncer) IsSynced(ctx context.Context) bool {
...
-
-func (s *Syncer) isClosed() bool {
-	select {
-	case <-s.shutdownCtx.Done():
-		return true
-	default:
-		return false
-	}
+	// synchronize only allows one process to proceed so no need to worry about spawning too many goroutines
+	s.tomb.Go(func() error {
```
dshulyak, line 187: this is probably not in the scope of this PR, but maybe it should pass a notification that will be handled by the loop?

dshulyak (approved):

syncer/syncer.go
countvonzero: errShuttingDown.

syncer/syncer.go
countvonzero: i think the original idea was that if this RPC call comes in during the a synchronize run, it will be ignored instead of being processed after the current run.
i don't see a huge advantage one way over the other.

syncer/syncer.go
countvonzero: @dshulyak it looks promising! i am quite annoyed with this issue.

so during shutdown, i call 
```
g.Go(func() error {
    return errShuttingDown
})
g.Wait()
```
did i read that correctly?
if so, the API is indeed simpler.

syncer/syncer.go
dshulyak: i think it should be nil or have no error at all. otherwise, caller have to check it and can log it unnecessarily for example

syncer/syncer.go
dshulyak: it is usually used with context or another termination mechanism (a channel), so the code looks like:

```go
ctx, cancel := context.WIthCancel(context.Background())
g.Go(func() error {
    <-ctx.Done()
    return ctx.Err()
})
g.Wait()
```

But there is another detail that i missed. With errgroup, if Go called after ctx is canceled and Wait is called it will panic. and with tomb it will not try to spawn a goroutine and just exit, right? so the tomb is probably better for this use case, unless we want to make a small wrapper around errgroup

noamnelke: I actually think this is not how Tomb was intended to be used. The way it should work is that you instantiate a tomb in main and then start all modules with it. Then it makes perfect sense that the tomb kills itself when the last goroutine dies.

To implement this concept incrementally, I would start by instantiating the tomb in main, but only use it to start the sync module. This way, the goroutine that runs the sync "event loop" will keep the tomb alive and you can then start other goroutines from that event loop.

I didn't know about `errgroup` before, which seems to perform a similar function, putting more of the heavy lifting on the context itself. It's more recently updated and more widely used, making me thing we should consider using it in place of tomb. I think it can pretty easily replace tomb in the code - unless either of you sees any issue with it, I would give it a try and see if it doesn't introduce new issues.

syncer/syncer.go
```diff
@@ -32,11 +33,13 @@ type layerFetcher interface {
 // Configuration is the config params for syncer
 type Configuration struct {
 	SyncInterval time.Duration
-	// the sync process will try validate the current layer if ValidationDelta has elapsed.
+	// the sync process will try to validate the current layer if ValidationDelta has elapsed.
 	ValidationDelta time.Duration
 	AlwaysListen    bool
 }
 
+var errShuttingDown = errors.New("shutting down")
```
noamnelke, line 41: I don't think you're supposed to do this. I think you should be using `tomb.ErrDying`. This means that if you kill the tomb then it will return no error if all errors returned by goroutines are `tomb.ErrDying`. If at least one other error is returned by a goroutine then you'll get an aggregated error from the tomb which is the first (non `tomb.ErrDying`) error returned.

You could check for this error yourself, but that's not exactly the same - tomb remembers and returns the first error that isn't `tomb.ErrDying`. So if the first error is `errShuttingDown` and then another error is returned - you'll get `errShuttingDown` as the aggregate (not what you probably want).

noamnelke (changes requested):

syncer/syncer.go
dshulyak: yes errgroup will fit the described use case. 

but it would be nice to have a goroutine manager that can prevent spawning new goroutines after Wait is called (Go method will silently exit instead). it should cover the use cases in this PR, and many others I've seen in this repo

noamnelke: I don't think it should be the responsibility of the goroutine manager to "prevent spawning new goroutines". The shutdown procedure might require spawning processes and that's fine. This should be our responsibility and the tool provides us a way to know if we're shutting down so WE can decide not to spawn. In Tomb you check if the `dying` channel is closed and in errgroup you check the `context.Done()` channel.

dshulyak: i agree. but in order not to repeat ourselves (with selects on ctx.Done() before calling Go) we can write a small wrapper around errgroup that will do it always for us.

also i think it is not as simple as a single select, select on ctx.Done() and calling Go should be atomic, otherwise there will be a race with Wait. so some locking will be required.

noamnelke: I don't think we should necessarily check `ctx.Done()` before starting goroutines.

As I see it, every `select` statement we put in the code should include a read from `ctx.Done()` and the context should be passed to any long-running method that can be canceled.

You want to be able to do shutdown procedures in parallel (e.g. flushing state to disk).

noamnelke: I briefly considered implementing our own `select`, but it [looks like](https://stackoverflow.com/a/19992525/1225467) it would be a syntactic nightmare...

dshulyak: the problem is that if you call Go after Wait is called errgroup will panic (the same as regular sync.WaitGroup). what you described regarding select is obviously needed but not enough if we will use the same group to manage goroutines that are spawned by other goroutines

noamnelke: A WaitGroup will only panic if a positive `Add()` happens while there's a pending `Wait()` _and the counter is zero_. ([see docs](https://pkg.go.dev/sync#WaitGroup))

I don't know enough about errgroup, but if the problem is that it uses an underlying WaitGroup then it shouldn't be an issue: as long as there are running goroutines, more goroutines can be started (since the counter isn't zero) and once all goroutines have completed, there's no risk of some goroutine spawning another goroutine. Right?

dshulyak: i believe there is an issue. see this code for an example:

```go
var g errgroup.Group
g.Go(func() {
    g.Go(func() {}) 
})

g.Wait() // you cannot guarantee that both goroutines will exit
```

dshulyak: this is from the source code of Add

```
       if w != 0 && delta > 0 && v == int32(delta) {
		panic("sync: WaitGroup misuse: Add called concurrently with Wait")
	}
	if v > 0 || w == 0 {
		return
	}
	// This goroutine has set counter to 0 when waiters > 0.
	// Now there can't be concurrent mutations of state:
	// - Adds must not happen concurrently with Wait,
	// - Wait does not increment waiters if it sees counter == 0.
	// Still do a cheap sanity check to detect WaitGroup misuse.
	if *statep != state {
		panic("sync: WaitGroup misuse: Add called concurrently with Wait")
	}
```

countvonzero:
@noamnelke @dshulyak 

thank you both for the feedback. i'm going to change this to use taskgroup package that @dshulyak added. ok with both of you?

countvonzero:
https://github.com/spacemeshos/go-spacemesh/pull/2744 merged instead

PR #2612, opened 2021-07-30
[Merged by Bors] - syncer don't work on current layer

### LINKED ISSUE #2610: Blocks are not produced because node goes out of sync
## Description
It seems that nodes go out of sync sometimes and not produce blocks at some layers

```2021-07-20T04:36:03.3731003Z 2021-07-20T04:32:24.107Z	INFO	bb576.sync         	setting layer 6 to zero-block	{"node_id": "bb5760ec4abb906c71caa5b6b039fde593651432686141a64ec39dfdd3b4a177", "module": "sync"}```
```2021-07-20T04:36:03.3741370Z 2021-07-20T04:32:24.107Z	ERROR	bb576.sync         	failed to set zero-block for layer	{"node_id": "bb5760ec4abb906c71caa5b6b039fde593651432686141a64ec39dfdd3b4a177", "module": "sync", "layer_id": 6, "errmsg": "layer has blocks", "name": "sync"}```

this error triggers the “not synced” state

Error is caused because we try to sync the current layer while accepting blocks from gossip, while we do that there is a race between synced and blocks received by gossip - if a block was received after we have synced layer with peers and received empty response it will error 

## Expected Behavior
Node should be synced all of the time
Node shouldn't go out of sync if an error is received when syncing current layer, maybe even not sync the current layer

## Environment
 - OS: 
 
 - Node Version: 
 
 - App Version (if applicable):

### PR DESCRIPTION
**Pull request by countvonzero**
## Motivation
Closes #2610

## Changes
- syncer sync up to current-1 layer
- syncer doesn't validate a layer that hare plans to

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -38,7 +38,7 @@ type Configuration struct {
 }
 
 const (
-	outOfSyncThreshold  uint32 = 2 // see notSynced
+	outOfSyncThreshold  uint32 = 3 // see notSynced
 	numGossipSyncLayers uint32 = 2 // see gossipSync
 )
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by countvonzero** [comment]
bors try

**Comment by countvonzero** [comment]
bors try

**Comment by countvonzero** [comment]
bors try

**Comment by countvonzero** [comment]
bors try

**Review by antonlerner — APPROVED** [review]

File: syncer/syncer.go [hunk 1]
```diff
@@ -292,7 +292,6 @@ func (s *Syncer) synchronize(ctx context.Context) bool {
 			logger.Info("setting layer %v to zero-block", layerID)
 			if err := s.mesh.SetZeroBlockLayer(layerID); err != nil {
```
antonlerner, line 293: I think we shouldn't fail sync if there are already blocks in the layer... I wonder if this can happen now after your previous fix

countvonzero: i think it's still possible in the case where
- we received a late block before our peers
- the peer we connect to is not fully synced, and other peers happen to return error
the chance is there but small. 

either way, like you said, i think we still should not fail sync for this.

nkryuchkov (approved):

countvonzero:
bors merge

PR #2617, opened 2021-07-31
[Merged by Bors] - standardizing aggregated hash calculation

### LINKED ISSUE #2574: standardize mesh hash calculation

countvonzero:
## Motivation
Closes #2574

## Changes
- make sure all aggregated hash is calculated the same way
- make sure aggregated hash is updated in order
- set and persist processed layer together

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -163,7 +163,7 @@ func NewRecoveredMesh(db *DB, atxDb AtxDB, rewardConfig Config, trtl tortoise, t
 	msh.With().Info("recovered mesh from disk",
 		log.FieldNamed("latest_layer", msh.LatestLayer()),
 		log.FieldNamed("validated_layer", msh.ProcessedLayer()),
-		log.String("layer_hash", msh.ProcessedLayerHash().Hex()),
+		log.String("layer_hash", msh.ProcessedLayerHash().ShortString()),
 		log.String("root_hash", pr.GetStateRoot().String()))
 
 	return msh
@@ -256,7 +256,7 @@ func (msh *Mesh) setProcessedLayerFromRecoveredData(pLayer *ProcessedLayer) {
 	msh.mutex.Lock()
 	defer msh.mutex.Unlock()
 	msh.processedLayer = *pLayer
-	msh.Event().Info("processed layer set from recovered data", pLayer.ID, log.String("layer_hash", pLayer.Hash.Hex()))
+	msh.Event().Info("processed layer set from recovered data", pLayer.ID, log.String("layer_hash", pLayer.Hash.ShortString()))
 }
 
 func (msh *Mesh) setProcessedLayer(layer *types.Layer) {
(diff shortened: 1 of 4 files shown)


countvonzero:
bors try

mesh/mesh.go
```diff
@@ -249,21 +252,58 @@ func (msh *Mesh) getProcessedLayer() ProcessedLayer {
...
-			log.FieldNamed("new_processed_layer", layerID))
+	msh.processedLayer = *pLayer
+	msh.Event().Info("processed layer set from recovered data", pLayer.ID, log.String("layer_hash", pLayer.Hash.Hex()))
+}
+
+func (msh *Mesh) setProcessedLayer(layer *types.Layer) {
+	msh.mutex.Lock()
+	defer msh.mutex.Unlock()
+	if !layer.Index().After(msh.processedLayer.ID) {
+		msh.With().Warning("attempt to set processed layer to an older layer",
```
antonlerner: if we try to set a processed layer with new blocks than perhaps we should allow this and also calculate aggregated hashes for all subsequent layers as well.

mesh/mesh.go
```diff
@@ -249,21 +252,58 @@ func (msh *Mesh) getProcessedLayer() ProcessedLayer {
...
 	}
+	msh.processedLayer = lastProcessed
 	events.ReportNodeStatusUpdate()
-	msh.Event().Info("processed layer set", layerID, log.String("layer_hash", hash.Hex()))
+	msh.Event().Info("processed layer set", msh.processedLayer.ID, log.String("layer_hash", msh.processedLayer.Hash.Hex()))
+
+	if err := msh.persistProcessedLayer(&lastProcessed); err != nil {
+		msh.With().Error("failed to persist processed layer",
+			log.FieldNamed("processed_layer", lastProcessed.ID),
+			log.String("processed_layer_hash", lastProcessed.Hash.Hex()),
```
antonlerner: maybe print short hex of hash

mesh/mesh.go
countvonzero: done. also added test

mesh/mesh.go
countvonzero: done for the whole file

antonlerner (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #2625, opened 2021-08-03
[Merged by Bors] - Exit tortoise beacon cleanly by using taskgroup.Group

### LINKED ISSUE #2621: Ignore weak coin if votes margin result is sufficient for consensus
## Description

In the implementation, vote margin in round 2 will always be computed using weak coin computed during round 1 (and so on for every next round). Even if vote margin is high enough to ensure that the consensus is reached. This process needs to be changed to ignore weak coin in the latter cases. 

## Affected code

tortoisebeason/beacon_calc.go

cc @antonlerner @nkryuchkov please check if this is accurate

One question that i have, why should we ignore weak coin after it was already computed? Maybe we should compute weak coin **only** when vote margin is not high enough?

dshulyak:
## Motivation

Tortoise beacon creates multiple goroutines during its lifetime, not all of them are guaranteed to exit when Close is called. I need it for my logging refactoring, and it should be default behavior anyway.

closes: https://github.com/spacemeshos/go-spacemesh/issues/2621
depends on change in: https://github.com/spacemeshos/go-spacemesh/pull/2605

## Changes

I added new module called taskgroup, it behaves identically to errgroup, with a difference that it is safe to call Go concurrently with Wait. If the group didn't terminate Wait will ensure that concurrent Go call completes. If it did Go call will exit immediately with Terminated error (this error should be ignored unless callers counts number of goroutine for some reason).

## Test Plan
unit tests

## TODO

## DevOps Notes

### DIFF
--- a/taskgroup/group.go
+++ b/taskgroup/group.go
@@ -6,8 +6,8 @@ import (
 	"sync"
 )
 
-// Terminated returnedif Group was already terminated.
-var Terminated = errors.New("taskgroup: terminated")
+// ErrTerminated returned if Group was already terminated.
+var ErrTerminated = errors.New("taskgroup: terminated")
 
 // Option to modify Group.
 type Option func(g *Group)
@@ -19,7 +19,7 @@ func WithContext(ctx context.Context) Option {
 	}
 }
 
-// WithContext return instance of the Group.
+// New returns instance of the Group.
 func New(opts ...Option) *Group {
 	g := &Group{
 		waitErr: make(chan struct{}),
@@ -50,12 +50,12 @@ type Group struct {
 }
 
 // Go spawns new goroutine that will run f, unless Group was already terminated.
-// In the latter case Terminated error will be returned.
+// In the latter case ErrTerminated error will be returned.
 func (g *Group) Go(f func(ctx context.Context) error) error {
 	g.mu.Lock()
 	defer g.mu.Unlock()
 	if g.err != nil {
-		return Terminated
+		return ErrTerminated
 	}
 	g.wg.Add(1)
 	go func() {
--- a/taskgroup/group.go
+++ b/taskgroup/group.go
@@ -1,0 +1,90 @@
+package taskgroup
+
+import (
+	"context"
+	"errors"
+	"sync"
+)
+
+// Terminated returnedif Group was already terminated.
+var Terminated = errors.New("taskgroup: terminated")
+
+// Option to modify Group.
+type Option func(g *Group)
+
+// WithContext passes parent context to use for the Group.
+func WithContext(ctx context.Context) Option {
+	return func(g *Group) {
+		g.ctx = ctx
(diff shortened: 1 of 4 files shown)


dshulyak:
@noamnelke @countvonzero we had a conversation about it in another PR, do you have anything against using new module to handle task of managing goroutines?

taskgroup/group.go
```diff
@@ -0,0 +1,90 @@
...
+	g.mu.Lock()
+	if g.err != nil {
+		g.mu.Unlock()
+		return g.err
+	}
+	g.mu.Unlock()
+
+	<-g.waitErr
+	// at this point the error is set to a non-nil value, won't be ever overwritten,
+	// and all Go calls will terminated immediatly, therefore re-locking here is not required.
```
nkryuchkov: `will be terminated immediately`

tortoisebeacon/tortoise_beacon.go
```diff
@@ -204,41 +145,107 @@ func New(
...
+	proposalChans   map[types.EpochID]chan *proposalMessageWithReceiptData
+
+	layerMu   sync.RWMutex
+	lastLayer types.LayerID
+}
+
+// Start starts listening for layers and outputs.
+func (tb *TortoiseBeacon) Start(ctx context.Context) error {
+	if !atomic.CompareAndSwapUint64(&tb.closed, 0, 1) {
+		return nil
```
nkryuchkov, line 212: This case shouldn't usually happen, so I'd log it.

dshulyak:
bors test

dshulyak:
bors try

nkryuchkov (approved):

dshulyak:
bors try

dshulyak:
bors try

taskgroup/group.go
```diff
@@ -0,0 +1,90 @@
...
+func (g *Group) Go(f func(ctx context.Context) error) error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	if g.err != nil {
+		return ErrTerminated
+	}
+	g.wg.Add(1)
+	go func() {
+		err := f(g.ctx)
+		g.wg.Done()
```
countvonzero: should this be in the last line of go func, say line 72?

countvonzero (approved):
no objections. as a matter of fact, happier to have our own.

taskgroup/group.go
dshulyak: i thought that it is correct to have here as well, because of the additional waitErr. but probably it will be more correct to execute Done as a very last thing. will change, thanks

tortoisebeacon/tortoise_beacon.go
dshulyak: sure will do, added it just for consistency with Stop

dshulyak:
bors merge

PR #2644, opened 2021-08-09
[Merged by Bors] - add aggregated layer hash in layer hash response

### LINKED ISSUE #2524: add aggregated hashes in layer response
## Description
part of #2507

currently nodes exchange simple layer hashes. if two nodes simply exchange layer hashes for the past X layers, there may be multiple layers with different hashes and one is still none the wiser about where the point of divergence is. the only way is to exchange aggregated layer hashes where 

aggregatedHash(N) = hash(aggregatedHash(N-1) + hash(N)) 

to ensure there is a point of divergence between two peers.

countvonzero:
## Motivation
Closes #2524

## Changes
- only persist layer hash after validity is determined
- add aggregated layer hash and processed layer in layer hash response

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -38,9 +38,11 @@ type TxProcessor interface {
 // layerDB is an interface that returns layer data and blocks
 type layerDB interface {
 	GetLayerHash(ID types.LayerID) types.Hash32
+	GetAggregatedLayerHash(ID types.LayerID) types.Hash32
 	GetLayerHashBlocks(hash types.Hash32) []types.BlockID
 	GetLayerInputVector(hash types.Hash32) ([]types.BlockID, error)
 	SaveLayerInputVectorByID(ctx context.Context, id types.LayerID, blks []types.BlockID) error
+	ProcessedLayer() types.LayerID
 }
 
 type atxIDsDB interface {
@@ -92,7 +94,7 @@ type peerResult struct {
 
 // LayerHashResult is the result of fetching hashes for each layer from peers.
 type LayerHashResult struct {
-	Hashes map[types.Hash32][]peers.Peer
+	Hashes map[layerHash][]peers.Peer
 	Err    error
 }
 
@@ -191,11 +193,24 @@ func (l *Logic) AddDBs(blockDB, AtxDB, TxDB, poetDB, IvDB, tbDB database.Getter)
 // layerHashReqReceiver returns the layer hash for the specified layer.
 func (l *Logic) layerHashReqReceiver(ctx context.Context, msg []byte) []byte {
 	lyr := types.NewLayerID(util.BytesToUint32(msg))
-	h := l.layerDB.GetLayerHash(lyr)
-	l.log.WithContext(ctx).With().Debug("responded layer hash request",
-		lyr,
-		log.String("layerHash", h.ShortString()))
-	return h.Bytes()
+	lyrHash := &layerHash{
+		ProcessedLayer: l.layerDB.ProcessedLayer(),
+		SimpleHash:     l.layerDB.GetLayerHash(lyr),
+		AggHash:        l.layerDB.GetAggregatedLayerHash(lyr),
+	}
+	out, err := types.InterfaceToBytes(lyrHash)
+	if err != nil {
+		l.log.WithContext(ctx).With().Error("failed to serialize layer hash",
+			lyr,
+			log.String("simpleHash", lyrHash.SimpleHash.ShortString()),
+			log.String("aggHash", lyrHash.AggHash.ShortString()))
+	} else {
+		l.log.WithContext(ctx).With().Debug("responded layer hash request",
+			lyr,
+			log.String("simpleHash", lyrHash.SimpleHash.ShortString()),
+			log.String("aggHash", lyrHash.AggHash.ShortString()))
+	}
+	return out
 }
 
 // epochATXsReqReceiver returns the ATXs for the specified epoch.
@@ -341,13 +356,18 @@ func notifyLayerHashResult(layerID types.LayerID, channels []chan LayerHashResul
 	logger.With().Debug("got layer hashes. now aggregating", layerID)
 	numErrors := 0
 	// aggregate the peerResult by data
-	hashes := make(map[types.Hash32][]peers.Peer)
+	hashes := make(map[layerHash][]peers.Peer)
 	for peer, res := range peerResult {
 		if res.err != nil {
 			numErrors++
 		} else {
-			hash := types.BytesToHash(res.data)
-			hashes[hash] = append(hashes[hash], peer)
+			var lyrHash layerHash
+			convertErr := types.BytesToInterface(res.data, &lyrHash)
+			if convertErr != nil {
+				logger.Error("received error converting bytes to layerHash", convertErr)
+				numErrors++
+			}
+			hashes[lyrHash] = append(hashes[lyrHash], peer)
 		}
 	}
 	result := LayerHashResult{Hashes: hashes, Err: nil}
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -195,20 +195,20 @@ func (l *Logic) layerHashReqReceiver(ctx context.Context, msg []byte) []byte {
 	lyr := types.NewLayerID(util.BytesToUint32(msg))
 	lyrHash := &layerHash{
 		ProcessedLayer: l.layerDB.ProcessedLayer(),
-		SimpleHash:     l.layerDB.GetLayerHash(lyr),
-		AggHash:        l.layerDB.GetAggregatedLayerHash(lyr),
+		Hash:           l.layerDB.GetLayerHash(lyr),
+		AggregatedHash: l.layerDB.GetAggregatedLayerHash(lyr),
 	}
 	out, err := types.InterfaceToBytes(lyrHash)
 	if err != nil {
-		l.log.WithContext(ctx).With().Error("failed to serialize layer hash",
+		l.log.WithContext(ctx).With().Panic("failed to serialize layer hash",
 			lyr,
-			log.String("simpleHash", lyrHash.SimpleHash.ShortString()),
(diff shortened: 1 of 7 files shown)


countvonzero:
bors try

mesh/mesh_test.go
```diff
@@ -163,6 +163,58 @@ func addLayer(r *require.Assertions, id types.LayerID, layerSize int, msh *Mesh)
...
+	err = msh.SetZeroBlockLayer(lyrID)
+	assert.NoError(t, err)
+	assert.Equal(t, EmptyLayerHash, msh.GetLayerHash(lyrID))
+
+	// it's ok to add to an empty layer
+	lyr = addLayer(r, lyrID, 10, msh)
+	msh.ValidateLayer(lyr)
+	lyr, err = msh.GetLayer(lyrID)
+	assert.NoError(t, err)
+	assert.Equal(t, 10, len(lyr.Blocks()))
```
dshulyak, line 235: this will fail with nil pointer dereference if err is not nil. i think it is desirable to use require.Smth if code below the statement will crash if the assertion didn't pass

layerfetcher/layers.go
```diff
@@ -342,13 +356,18 @@ func notifyLayerHashResult(layerID types.LayerID, channels []chan LayerHashResul
...
 	for peer, res := range peerResult {
 		if res.err != nil {
 			numErrors++
 		} else {
-			hash := types.BytesToHash(res.data)
-			hashes[hash] = append(hashes[hash], peer)
+			var lyrHash layerHash
+			convertErr := types.BytesToInterface(res.data, &lyrHash)
+			if convertErr != nil {
+				logger.Error("received error converting bytes to layerHash", convertErr)
```
dshulyak: we have a popup notification for every error in the code. could you maybe change it to warn or debug? i personally think that any data that was received from the network and can't be unmarshalled is a debug information

layerfetcher/layers.go
```diff
@@ -192,11 +193,24 @@ func (l *Logic) AddDBs(blockDB, AtxDB, TxDB, poetDB, IvDB, tbDB database.Getter)
...
-		log.String("layerHash", h.ShortString()))
-	return h.Bytes()
+	lyrHash := &layerHash{
+		ProcessedLayer: l.layerDB.ProcessedLayer(),
+		SimpleHash:     l.layerDB.GetLayerHash(lyr),
+		AggHash:        l.layerDB.GetAggregatedLayerHash(lyr),
+	}
+	out, err := types.InterfaceToBytes(lyrHash)
+	if err != nil {
+		l.log.WithContext(ctx).With().Error("failed to serialize layer hash",
```
dshulyak: Panic is cleaner here. InterfaceToBytes returns error only if the input is incorrect (and in this case it is formed by the developer)

mesh/mesh_test.go
```diff
@@ -163,6 +163,58 @@ func addLayer(r *require.Assertions, id types.LayerID, layerSize int, msh *Mesh)
...
+	// after a layer is validated, only contextually valid blocks count towards its hash
+	expHash := types.CalcBlocksHash32(types.SortBlockIDs(types.BlockIDs(lyr.Blocks()[1:])), nil)
+	assert.Equal(t, expHash, msh.GetLayerHash(lyrID))
+	assert.NotEqual(t, expHash, lyr.Hash())
+}
+
+func TestMesh_SetZeroBlockLayer(t *testing.T) {
+	r := require.New(t)
+	msh := getMesh(t, "zero block layer")
+	defer msh.Close()
```
dshulyak: t.Cleanup(msh.Close)

mesh/mesh_test.go
```diff
@@ -163,6 +163,58 @@ func addLayer(r *require.Assertions, id types.LayerID, layerSize int, msh *Mesh)
 	return l
 }
 
+func TestMesh_GetLayerHash(t *testing.T) {
+	r := require.New(t)
+	msh := getMesh(t, "get layer hash")
+	defer msh.Close()
```
dshulyak: t.Cleanup

layerfetcher/wire_types.go
```diff
@@ -2,7 +2,17 @@ package layerfetcher
...
 
-// layerBlocks is the response for layer hash
+// layerHash is the response for a given layer
+type layerHash struct {
+	// ProcessedLayer is the latest processed layer from peer
+	ProcessedLayer types.LayerID
+	// SimpleHash is the hash of contextually valid blocks (sorted by block ID) in the given layer
+	SimpleHash types.Hash32
+	// AggHash is the aggregated hash of all layers up to the given layer
+	AggHash types.Hash32
```
dshulyak: could it be a full word please?

layerfetcher/wire_types.go
```diff
@@ -2,7 +2,17 @@ package layerfetcher
 
 import "github.com/spacemeshos/go-spacemesh/common/types"
 
-// layerBlocks is the response for layer hash
+// layerHash is the response for a given layer
+type layerHash struct {
+	// ProcessedLayer is the latest processed layer from peer
+	ProcessedLayer types.LayerID
+	// SimpleHash is the hash of contextually valid blocks (sorted by block ID) in the given layer
+	SimpleHash types.Hash32
```
dshulyak: i think just Hash is easier to understand compared with SimpleHash

layerfetcher/wire_types.go
```diff
@@ -2,7 +2,17 @@ package layerfetcher
 
 import "github.com/spacemeshos/go-spacemesh/common/types"
 
-// layerBlocks is the response for layer hash
+// layerHash is the response for a given layer
```
dshulyak: it seems like a copy from another struct. better to remove it altogether

dshulyak (changes requested):

countvonzero:
bors try

countvonzero:
bors try

mesh/mesh.go
```diff
@@ -236,6 +236,22 @@ func (msh *Mesh) GetLayer(i types.LayerID) (*types.Layer, error) {
...
 
+// GetLayerHash returns layer hash for received blocks
+func (msh *Mesh) GetLayerHash(layerID types.LayerID) types.Hash32 {
+	h, err := msh.recoverLayerHash(layerID)
+	if err == nil {
+		return h
+	}
+	if err == database.ErrNotFound {
+		// layer hash not persisted. i.e. contextual validity not yet determined
+		lyr, err := msh.GetLayer(layerID)
```
antonlerner, line 247: this means that layer hash might be changed when layer validity is determined, is this wanted behaviour?

mesh/mesh.go
```diff
@@ -575,6 +568,15 @@ func (msh *Mesh) persistAggregatedLayerHash(layerID types.LayerID, hash types.Ha
 	}
 }
 
+// GetAggregatedLayerHash returns the aggregated layer hash up to the specified layer
+func (msh *Mesh) GetAggregatedLayerHash(layerID types.LayerID) types.Hash32 {
+	h, err := msh.getAggregatedLayerHash(layerID)
+	if err != nil {
+		return EmptyLayerHash
```
antonlerner, line 575: shouldn't this return an error? event if the requested layer has no blocks the aggregated hash shouldn't be zero

mesh/mesh.go
countvonzero: actually the behavior in this area is mostly unchanged by this PR.

- before this change:
when we received blocks via gossip, the code persists the layer hash as the blocks are received. see 
https://github.com/spacemeshos/go-spacemesh/blob/develop/mesh/meshdb.go#L436
before the layer is validated, it simply calculates the hash of all blocks since validity isn't determined yet. but after a layer is validated, a problem can rise when we receive a late block. because the code immediately persists the layer hash when we get a new block, before it determines whether the block is valid or not.

in short, the persisted hash **may** contains invalid blocks.

- after this change
layer hash is only persisted **after** validation happens. before the validation runs for a given layer, no hash is persisted. that's why i added the code here to simply return hash of all blocks. no validity has been determined. after the layer is validated, then we only let tortoise update the layer hash as it deem a block valid.

in short, the persisted hash **only** contains valid blocks.

in both cases, layer hash can change from `all blocks` to only `valid blocks` after validation.

mesh/mesh.go
countvonzero: exactly. EmptyLayerHash is actually the way to indicate error for aggregated layer. it will **never** be EmptyLayerHash, even for layer 1.

to be honest i was trying to pass around an error between peers but xdr encoding doesn't work with nil 
`xdr:encodeInterface: can't encode nil interface`

layerfetcher/layers.go
countvonzero: done. thanks

layerfetcher/layers.go
countvonzero: what is your opinion in the case of DDoS attack? i guess in this case we have metrics (numErrors). but in general i think warning may be appropriate. changed to debug for now.

layerfetcher/wire_types.go
countvonzero: done

layerfetcher/wire_types.go
countvonzero: done

layerfetcher/wire_types.go
countvonzero: done

mesh/mesh_test.go
countvonzero: done for the whole test

mesh/mesh_test.go
countvonzero: done

mesh/mesh_test.go
countvonzero: thanks. done. and other places (i can find anyway. eyes got blurry after a while)

dshulyak (approved):

antonlerner (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #2649, opened 2021-08-09
[Merged by Bors] - Reduce the amount of memory consumed by tortoise beacon

### LINKED ISSUE #2636: Tortoise beacon doesn't release memory
On current devnet for 0.2 tortoise beacon doesn't release memory. See attached flame graph.

Screenshot from 2021-08-06 10-56-02

More data can be collected from any miner by fetching pprof.

nkryuchkov:
## Motivation

Tortoise beacon cleaned up only the internal `map` with beacons but didn't clean up other internal `map`s, which led to leaking memory. Also, a lot of allocations were done when logging a malformed message and its signature as a string. The scope of this PR is only those changes. It's possible to do further memory allocation improvements in the tortoise beacon (remove some `map`s), but they will be done in other PRs.

Closes #2636

## Changes

- Clean up more old tortoise beacon data
- Remove logging message and its signature if it's malformed

## Test Plan

## DevOps Notes


nkryuchkov:
bors try

tortoisebeacon/tortoise_beacon.go
```diff
@@ -326,7 +318,14 @@ func (tb *TortoiseBeacon) cleanup() {
...
 		// TODO(nkryuchkov): https://github.com/spacemeshos/go-spacemesh/pull/2267#discussion_r662255874
 		if tb.epochIsOutdated(e) {
+			delete(tb.validProposals, e)
+			delete(tb.potentiallyValidProposals, e)
 			delete(tb.beacons, e)
+			delete(tb.proposalPhaseFinishedTimestamps, e)
+			delete(tb.incomingVotes, e)
+			delete(tb.ownVotes, e)
+			delete(tb.firstRoundIncomingVotes, e)
+			delete(tb.firstRoundOutcomingVotes, e)
```
dshulyak: can we release all this data immediately when the beacon for the epoch was calculated? precisely at the end of handleEpoch method, maybe even just a deferred statement to cleanup it if we exit early.

i don't see why we need to store all this stuff in memory for 1000 epochs, with an exception of the beacon itself ofcourse (but should we store it in database actually?).

dshulyak (changes requested):
i also saw that some keys are marshaled like this 
```go
tb.validProposals[epoch][util.Bytes2Hex(proposedSignature)] = struct{}{}
```

can't we just string() this? hexary encoding makes the size of the string twice larger than original

tortoisebeacon/tortoise_beacon.go
antonlerner: whenever a beacon is calculated it needs to be stored in DB and all votes can be discarded at once

nkryuchkov:
@dshulyak Yes, we can avoid hex encoding. It also slows things down. However, it's easier to debug hex strings because of better readability, that's why I implemented it this way for now. I'm going to change it in future PRs. It's worth leaving a comment about it, though. Thanks!

tortoisebeacon/tortoise_beacon.go
nkryuchkov: Yes, we can release data immediately and store the beacon in the DB only. I also want to remove some internal `map`s. I'm working on that now, but I was going to have this in another PR because I wanted to keep the scope of this PR as small as possible to have it ready ASAP and merged before the devnet relaunch. The current implementation should reduce the memory usage by a lot and it introduces a few changes, which are easy to test because they are similar to already implemented beacon cleanup.

tortoisebeacon/tortoise_beacon.go
```diff
@@ -584,15 +555,17 @@ func (tb *TortoiseBeacon) proposalPhaseImpl(ctx context.Context, epoch types.Epo
 		tb.validProposals[epoch] = make(map[proposal]struct{})
 	}
 
-	tb.validProposals[epoch][util.Bytes2Hex(proposedSignature)] = struct{}{}
+	// TODO(nkryuchkov): encode as string instead of hex
```
dshulyak: outdated TODO?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -901,8 +873,8 @@ func (tb *TortoiseBeacon) getSignedProposal(epoch types.EpochID) ([]byte, error)
 	signature := tb.vrfSigner.Sign(p)
 	tb.Log.With().Debug("Calculated signature",
 		log.Uint64("epoch_id", uint64(epoch)),
-		log.String("proposal", util.Bytes2Hex(p)),
-		log.String("signature", util.Bytes2Hex(signature)))
+		log.String("proposal", string(p)),
```
dshulyak: i think you still want to use hex in logs (or log.Binary). string([]byte{..}) won't be readable most of the time

tortoisebeacon/beacon_calc.go
```diff
@@ -35,6 +35,8 @@ func (tb *TortoiseBeacon) calcBeacon(epoch types.EpochID, coinflip bool) error {
 	tb.beacons[epoch] = beacon
 	tb.beaconsMu.Unlock()
 
+	tb.cleanupVotes(epoch)
```
dshulyak: what happens if we exit with error in this method or even before calcBeacon is called?

tortoisebeacon/tortoise_beacon.go
nkryuchkov: Yes, it doesn't exist anymore in new commits. but they are not pushed yet

tortoisebeacon/tortoise_beacon.go
nkryuchkov: You're right, I replaced too much stuff

tortoisebeacon/beacon_calc.go
nkryuchkov: The call needs to be deferred in a parent function

dshulyak (approved):

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bots try

nkryuchkov:
bors try

nkryuchkov:
bors try

tortoisebeacon/beacon_calc.go
```diff
@@ -1,32 +1,35 @@
...
 	"strings"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
-func (tb *TortoiseBeacon) calcBeacon(epoch types.EpochID, coinflip bool) error {
+func (tb *TortoiseBeacon) calcBeacon(ctx context.Context, epoch types.EpochID, lastRoundVotes allVotes) error {
+	select {
```
dshulyak: imo this is unnecessary. calcBeacon is just a computation and suppose not to take long time, right?

tortoisebeacon/proposal_list.go
```diff
@@ -11,8 +11,7 @@ import (
 
 type proposalList []proposal
 
-// Sort sorts proposal list.
-func (hl proposalList) Sort() []proposal {
+func (hl proposalList) sort() []proposal {
```
dshulyak, line 14: unrelated to pr, but isn't this copy below unnecessary?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -592,59 +548,143 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
-			return false
+			return allVotes{}
 		}
+
 		tb.weakCoin.FinishRound()
-		coinflip = tb.weakCoin.Get(epoch, round)
+
+		cf := tb.weakCoin.Get(epoch, round+1)
+		if cf {
+			atomic.StoreUint64(&coinFlip, 1)
```
dshulyak: if atomic is required this is clearly a bug. coinflip from the previous round should be used in current round, concurrent update doesn't make any sense

dshulyak (changes requested):

signing/vrfsigner.go
```diff
@@ -17,7 +17,7 @@ type VRFSigner struct {
 
 // Sign signs a message for VRF purposes
 func (s VRFSigner) Sign(msg []byte) []byte {
-	return ed25519.Sign(s.privateKey, msg)
+	return ed25519.Sign2(s.privateKey, msg)
```
dshulyak: this is actually the reason why i am seeing issues in the devnet. i didn't notice that we can't extract key from vrf signatures. i used ed signer in multi actor test and this is why it passed

tbh i think we should remove vrf signer, this is the same as ed signer but initialized with different key

dshulyak: maybe vrf signer wasn't using ed++ by intention?

tortoisebeacon/beacon_calc.go
nkryuchkov: It is just computation, but I made `runConsensusPhase` return an empty list of votes if the context is cancelled, so I need to check it in `calcBeacon` too and return in this case. But I think it's better to indicate the cancelled context explicitly. What I have now is not clear

tortoisebeacon/proposal_list.go
nkryuchkov: It can be removed

tortoisebeacon/tortoise_beacon.go
```diff
@@ -222,6 +199,7 @@ func (tb *TortoiseBeacon) IsClosed() bool {
 }
 
 // GetBeacon returns a Tortoise Beacon value as []byte for a certain epoch or an error if it doesn't exist.
+// TODO(nkryuchkov): consider not using (using DB instead)
```
antonlerner, line 208: are beacons not persisted now?

tortoisebeacon/tortoise_beacon.go
nkryuchkov: Without it, a data race is reported in `TestTortoiseBeacon`

dshulyak: then we either need to wait until goroutine from previous rounds exited, or copy coinflip before it is passed to a goroutine

```go
t.start(coinflip bool) {
  tb.tg.Go(func() {
   // use coinflip
  })
}
```

this way you will use correct version of the coinflip

dshulyak: the race here means that you can use coinFlip in round 2 that is meant to be used in round 3. so solving the race condition will only hide the problem but not solve it

tortoisebeacon/tortoise_beacon.go
nkryuchkov: @antonlerner They are stored in the `beacons` map and in the tortoise beacon database. But I think we can get rid of the `beacons` map and use the database only

signing/vrfsigner.go
nkryuchkov: Let's clarify that with research team

antonlerner: noam has responded that we intentionally didn't want to use ed++

tortoisebeacon/tortoise_beacon.go
nkryuchkov: @dshulyak Could you please look if https://github.com/spacemeshos/go-spacemesh/pull/2649/commits/48914176493d3a1e488df0996fdb83c55154e4a3 contains what you meant?

dshulyak: this should work too :+1: . i think that it would be cleaner with a function, but whatever you prefer

dshulyak (approved):

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #2650, opened 2021-08-09
[Merged by Bors] - Improve tortoise beacon config

nkryuchkov:
## Motivation

Currently, durations are set in tortoise beacon as amounts of milliseconds, but it's impossible to represent smaller durations, and large durations such as minutes or hours look not clear. Also, `q` is stored as a string, `θ` is a float64 (this may lose precision), but they both should be rational numbers. This PR makes them all look better.

NOTE: This PR changes the names of tortoise beacon config values in config files.

## Changes

- Change types of tortoise beacon config values

## Test Plan

## DevOps Notes


nkryuchkov:
bors try

nkryuchkov:
bors try

cmd/node/node.go
```diff
@@ -205,15 +208,48 @@ func LoadConfigFromFile() (*cfg.Config, error) {
...
+
+		switch f.Kind() {
+		case reflect.String:
+			v, ok := new(big.Rat).SetString(data.(string))
+			if !ok {
+				return nil, errors.New("malformed string representing big.Rat was provided")
+			}
+
+			return v, nil
+		case reflect.Float64:
```
dshulyak, line 243: `f` can't be smaller type?

cmd/root.go
```diff
@@ -187,29 +186,49 @@ func AddCommands(cmd *cobra.Command) {
...
-		config.TortoiseBeacon.Theta, "Ratio of votes for reaching consensus")
-	cmd.PersistentFlags().IntVar(&config.TortoiseBeacon.VotesLimit, "tortoise-beacon-votes-limit",
+	cmd.PersistentFlags().StringVar(&theta, "tortoise-beacon-theta",
+		config.TortoiseBeacon.Theta.String(), "Ratio of votes for reaching consensus")
+	cmd.PersistentFlags().Uint64Var(&config.TortoiseBeacon.VotesLimit, "tortoise-beacon-votes-limit",
 		config.TortoiseBeacon.VotesLimit, "Maximum allowed number of votes to be sent")
 
+	var ok bool
+
+	config.TortoiseBeacon.Q, ok = new(big.Rat).SetString(q)
```
dshulyak: i think using RatVar for Q and Theta that implements pflag.Value will be cleaner

tortoisebeacon/handlers.go
```diff
@@ -133,7 +141,7 @@ func (tb *TortoiseBeacon) classifyProposalMessage(m ProposalMessage, atxID types
 			log.String("atx_timestamp", atxTimestamp.String()),
 			log.String("next_epoch_start", nextEpochStart.String()),
 			log.String("received_time", receivedTime.String()),
-			log.String("grace_period", tb.gracePeriodDuration.String()))
+			log.String("grace_period", tb.config.GracePeriodDuration.String()))
```
dshulyak: there is a log.Duration

dshulyak (approved):

cmd/node/node.go
nkryuchkov: I think it will come from `json`s/`yaml`s/`toml`s as `float64`, so it's enough. Actually, I need string only. Do you think I should add all numeric types?

cmd/root.go
nkryuchkov: In this case, a wrapper-type is needed, which looks less clean to me. But I'll think of it

tortoisebeacon/handlers.go
nkryuchkov: thanks!

nkryuchkov:
bors try

nkryuchkov:
bors try-

cmd/root.go
dshulyak: i think wrapper can be applied only when Q is registered e.g:

```go
cmd.VarP((*RatVar)(config.TortoiseBeacon.Q), ...)
```

cmd/node/node.go
dshulyak: probably unnecessary

cmd/root.go
nkryuchkov: Good idea, thanks.

nkryuchkov:
bors try

nkryuchkov:
bors try-

nkryuchkov:
bors try-

nkryuchkov:
bors try

nkryuchkov:
bors merge

PR #2662, opened 2021-08-12
[Merged by Bors] - Use first round duration again

dshulyak:
## Changes
- first round will use longer duration specified in FirstRoundVotingDuration config parameter

## Test Plan
unit tests

## TODO

## DevOps Notes

### DIFF
--- a/tortoisebeacon/tortoise_beacon.go
+++ b/tortoisebeacon/tortoise_beacon.go
@@ -557,8 +557,8 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
 	// For next rounds,
 	// wait for δ time, and construct a message that points to all messages from previous round received by δ.
 	// rounds 1 to K
-	ticker := time.NewTicker(tb.config.VotingRoundDuration + tb.config.WeakCoinRoundDuration)
-	defer ticker.Stop()
+	timer := time.NewTimer(tb.config.FirstVotingRoundDuration + tb.config.WeakCoinRoundDuration)
+	defer timer.Stop()
 
 	var (
 		coinFlip            bool
@@ -629,14 +629,15 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
 		}
 
 		select {
-		case <-ticker.C:
+		case <-timer.C:
+			timer.Reset(tb.config.VotingRoundDuration + tb.config.WeakCoinRoundDuration)
 		case <-ctx.Done():
 			return allVotes{}, ctx.Err()
 		}
 
 		tb.weakCoin.FinishRound()
 
-		coinFlip = tb.weakCoin.Get(epoch, round+1)
+		coinFlip = tb.weakCoin.Get(epoch, round)
 	}
 
 	tb.Log.With().Debug("Consensus phase finished",
@@ -693,7 +694,11 @@ func (tb *TortoiseBeacon) receivedBeforeProposalPhaseFinished(epoch types.EpochI
 }
 
 func (tb *TortoiseBeacon) startWeakCoinRound(ctx context.Context, epoch types.EpochID, round types.RoundID) {
-	t := time.NewTimer(tb.config.VotingRoundDuration)
+	timeout := tb.config.FirstVotingRoundDuration
+	if round > firstRound {
+		timeout = tb.config.VotingRoundDuration
+	}
+	t := time.NewTimer(timeout)
 	defer t.Stop()
 
 	select {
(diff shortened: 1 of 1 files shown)


dshulyak:
bors try

dshulyak:
bors try

cmd/root.go
```diff
@@ -195,12 +195,8 @@ func AddCommands(cmd *cobra.Command) {
 		config.TortoiseBeacon.GracePeriodDuration, "Grace period duration in milliseconds")
 	cmd.PersistentFlags().DurationVar(&config.TortoiseBeacon.ProposalDuration, "tortoise-beacon-proposal-duration",
 		config.TortoiseBeacon.ProposalDuration, "Proposal duration in milliseconds")
-	cmd.PersistentFlags().DurationVar(&config.TortoiseBeacon.FirstVotingRoundDuration, "tortoise-beacon-first-voting-round-duration",
```
nkryuchkov: I missed that you removed its usage in https://github.com/spacemeshos/go-spacemesh/pull/2605. However, we need to be able to set the first voting round longer than the other ones

tortoisebeacon/config.go
```diff
@@ -9,66 +9,58 @@ import (
...
 type Config struct {
-	Kappa                    uint64        `mapstructure:"tortoise-beacon-kappa"`                       // Security parameter (for calculating ATX threshold)
-	Q                        *big.Rat      `mapstructure:"tortoise-beacon-q"`                           // Ratio of dishonest spacetime (for calculating ATX threshold). It should be a string representing a rational number.
-	RoundsNumber             types.RoundID `mapstructure:"tortoise-beacon-rounds-number"`               // Amount of rounds in every epoch
-	GracePeriodDuration      time.Duration `mapstructure:"tortoise-beacon-grace-period-duration"`       // Grace period duration
-	ProposalDuration         time.Duration `mapstructure:"tortoise-beacon-proposal-duration"`           // Proposal phase duration
-	FirstVotingRoundDuration time.Duration `mapstructure:"tortoise-beacon-first-voting-round-duration"` // First voting round duration
-	VotingRoundDuration      time.Duration `mapstructure:"tortoise-beacon-voting-round-duration"`       // Voting round duration
-	WeakCoinRoundDuration    time.Duration `mapstructure:"tortoise-beacon-weak-coin-round-duration"`    // Weak coin round duration
```
nkryuchkov: Why do you want to remove and not split one round into voting part and weak coin part?

cmd/root.go
dshulyak: why?

tortoisebeacon/config.go
dshulyak: i don't see any reason to split it in current implementation

cmd/root.go
dshulyak: the code complexity and amount of messages is the same, i haven't checked the size but it is probably the same as well

nkryuchkov: We send lists of valid and potentially valid proposals in the first round, but a bit vector with a difference in the following rounds, so the following rounds take less time. Therefore we can reduce the duration of the following rounds

dshulyak: but what is the difference is it like 10kb vs 100b or 10mb vs 100b? if it is the former then simplicity > micro optimizations

nkryuchkov: @tal-m made calculations in [the forum thread](https://community.spacemesh.io/t/tortoise-beacon-based-on-bad-vrf/100):

> ## Consensus phase
> 
> 1. The first round of the consensus phase is by far the most expensive: each party needs to send a list of `m=100` proposals and a signature, hence will download `m⋅W` proposals and `W` signatures, or approximately 31MB of data (assuming each proposal is also 256 bits).
> 2. Each subsequent round has a weak-coin part followed by a tortoise round. In the weak coin part, 100 parties each send a VRF output, so the total download is about 3200 bytes. In the tortoise round, each party needs to send a 100 bit string and a signature, so a total of `W⋅((m+256)/8)` bytes, or approximately 450KB of data per round, and about 110MB of data for all 256 rounds.
>
> Given that we can run the protocol over a two-week period, each round can be over an hour long, so even 31MB of data per round doesn’t sound too bad. The first round can also be optimized by having parties “compress” their proposal lists in the same way we do tortoise encoding: pointing to another first-round message and just listing the differences. This would require first-round messages to be slightly staggered, but given that we have an hour this shouldn’t be a problem.

dshulyak: added first round timeout

dshulyak:
@nkryuchkov i kept only first round in this pr, other things (merging weak coin roundn, already merged changes) are removed

nkryuchkov (approved):

dshulyak:
bors merge

PR #2665, opened 2021-08-15
[Merged by Bors] - add error to peer response in msg server

### LINKED ISSUE #2347: Message server should support error streams
## Description
currently our endpoint function are defined to return `[]byte` as response to requested message,
It would be better if we could return `[]byte, error` instead to signal that there has been an error on receiving end

countvonzero:
## Motivation

Closes #2347

## Changes

add error along with []byte for response from SendRequest

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -180,8 +180,8 @@ func (f MessageNetwork) GetPeers() []peers.Peer {
 
 type network interface {
 	GetPeers() []peers.Peer
-	SendRequest(ctx context.Context, msgType server.MessageType, payload []byte, address p2pcrypto.PublicKey, resHandler func(resp server.Response), failHandler func(err error)) error
-	RegisterBytesMsgHandler(msgType server.MessageType, reqHandler func(ctx context.Context, b []byte) []byte)
+	SendRequest(ctx context.Context, msgType server.MessageType, payload []byte, address p2pcrypto.PublicKey, resHandler func(msg []byte), failHandler func(err error)) error
+	RegisterBytesMsgHandler(msgType server.MessageType, reqHandler func(ctx context.Context, b []byte) ([]byte, error))
 	Close()
 }
 
@@ -333,16 +333,16 @@ func (f *Fetch) loop() {
 
 // FetchRequestHandler handles requests for sync from peersProvider, and basically reads Data from database and puts it
 // in a response batch
-func (f *Fetch) FetchRequestHandler(ctx context.Context, data []byte) []byte {
+func (f *Fetch) FetchRequestHandler(ctx context.Context, data []byte) ([]byte, error) {
 	if f.stopped() {
-		return server.SerializeResponse(f.log, nil, server.ErrShuttingDown)
+		return nil, server.ErrShuttingDown
 	}
 
 	var requestBatch requestBatch
 	err := types.BytesToInterface(data, &requestBatch)
 	if err != nil {
 		f.log.WithContext(ctx).With().Error("failed to parse request", log.Err(err))
-		return server.SerializeResponse(f.log, nil, server.ErrBadRequest)
+		return nil, server.ErrBadRequest
 	}
 	resBatch := responseBatch{
 		ID:        requestBatch.ID,
@@ -380,29 +380,22 @@ func (f *Fetch) FetchRequestHandler(ctx context.Context, data []byte) []byte {
 
 	bts, err := types.InterfaceToBytes(&resBatch)
 	if err != nil {
-		f.log.WithContext(ctx).With().Error("cannot parse message for batch ID",
+		f.log.WithContext(ctx).With().Panic("failed to serialize batch ID",
 			log.String("batchHash", resBatch.ID.ShortString()))
-		return nil
 	}
 	f.log.WithContext(ctx).With().Debug("returning response for batch",
 		log.String("batchHash", resBatch.ID.ShortString()),
 		log.Int("numResponse", len(resBatch.Responses)),
 		log.Int("dataSize", len(bts)))
-	return server.SerializeResponse(f.log, bts, nil)
+	return bts, nil
 }
 
 // receive Data from message server and call response handlers accordingly
-func (f *Fetch) receiveResponse(resp server.Response) {
+func (f *Fetch) receiveResponse(data []byte) {
 	if f.stopped() {
 		return
 	}
-	peerErr := resp.GetError()
-	if peerErr != nil {
-		f.log.With().Warning("received fetch error from peer", log.Err(peerErr))
-		return
-	}
 
-	data := resp.GetData()
 	var response responseBatch
 	err := types.BytesToInterface(data, &response)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


p2p/server/msgserver.go
```diff
@@ -262,7 +319,7 @@ func (p *MessageServer) RegisterBytesMsgHandler(msgType MessageType, reqHandler
 }
 
 // SendRequest sends a request of a specific message.
-func (p *MessageServer) SendRequest(ctx context.Context, msgType MessageType, payload []byte, address p2pcrypto.PublicKey, resHandler func(msg []byte), timeoutHandler func(err error)) error {
+func (p *MessageServer) SendRequest(ctx context.Context, msgType MessageType, payload []byte, address p2pcrypto.PublicKey, resHandler func(resp Response), timeoutHandler func(err error)) error {
```
dshulyak: i think it is more natural to keep resHandler unchanged and change timeoutHandler to errorHandler, and timeout will be just one of the possible errors.

also instead of marshaling response in each separate handler, we could change the signature of the request handler to return an error and encapsulate marshaling in the message server

p2p/server/msgserver.go
```diff
@@ -49,18 +51,68 @@ func extractPayload(m Message) []byte {
...
+
+// SerializeResponse serializes the response data returned by SendRequest
+func SerializeResponse(logger log.Log, data []byte, err error) []byte {
+	resp := response{Data: data}
+	if err != nil {
+		resp.ErrorStr = err.Error()
+	}
+	bytes, err := types.InterfaceToBytes(&resp)
+	if err != nil {
+		logger.With().Panic("failed to serialize response", log.Err(err))
```
dshulyak: could be just `panic`, without passing logger here

p2p/server/msgserver.go
```diff
@@ -49,18 +51,68 @@ func extractPayload(m Message) []byte {
...
 }
 
+// ErrShuttingDown is returned to the peer when the node is shutting down
+var ErrShuttingDown = errors.New("node is shutting down")
+
+// ErrBadRequest is returned to the peer upon failure to parse the request
+var ErrBadRequest = errors.New("unable to parse request")
+
+// Response is the interface type of the peer response returned by SendRequest.
+type Response interface {
```
dshulyak: might be simpler without an interface. we can unmarshal response to some inner type and then convert to public response only once
```go
type Response  struct {
  Data []byte
  Error error
}
```

p2p/server/msgserver.go
```diff
@@ -237,7 +289,12 @@ func (p *MessageServer) handleResponseMessage(ctx context.Context, headers *serv
 	p.pendMutex.RUnlock()
 	p.removeFromPending(headers.ReqID)
 	if okFoo {
-		foo.okCallback(headers.Payload)
+		resp, err := DeserializeResponse(logger, headers.Payload)
+		if err != nil {
+			logger.With().Warning("failed to deserialize response", log.Err(err))
```
dshulyak, line 290: we should notify the requestor that the response was received but it is malformed. maybe it will be more natural if we were using an error handler for errors.

if we call removeFromPending timeoutHandller will be never triggered, right?

dshulyak (changes requested):
if the response is malformed we will call neither resHandler nor timeoutHandler, this will result in leaked request, we need to handle that somehow

fetch/fetch.go
```diff
@@ -388,15 +388,21 @@ func (f *Fetch) FetchRequestHandler(ctx context.Context, data []byte) []byte {
...
 // receive Data from message server and call response handlers accordingly
-func (f *Fetch) receiveResponse(data []byte) {
+func (f *Fetch) receiveResponse(resp server.Response) {
 	if f.stopped() {
 		return
 	}
+	peerErr := resp.GetError()
+	if peerErr != nil {
+		f.log.With().Warning("received fetch error from peer", log.Err(peerErr))
+		return
```
antonlerner: if returned here, this means that the request won't be removed from `pendingRequests`, is this intentional? what should happen in such case? perhaps  the error should be propagated to `req.returnChan` ?

antonlerner (changes requested):

p2p/server/msgserver.go
countvonzero: thanks for the great suggestion. done.

my first edition of this PR is actually done as you suggested. but after some warped debugging i changed to this unnatural version. appreciate you bringing me back to the good path :D

p2p/server/msgserver.go
countvonzero: done

p2p/server/msgserver.go
countvonzero: removed this interface and just leave the internal struct `response`. note that error is not a concrete type so cannot be serialized. so i used a string instead

p2p/server/msgserver.go
countvonzero: you are right. now done (automatically) through the error handler. much more elegant.

fetch/fetch.go
countvonzero: as @dshulyak suggested to handle error in the error handler, this is now automatically handled correctly.

countvonzero:
bors try

layerfetcher/layers.go
```diff
@@ -246,30 +245,30 @@ func (l *Logic) layerHashBlocksReqReceiver(ctx context.Context, msg []byte) []by
 
 	out, err := types.InterfaceToBytes(b)
 	if err != nil {
-		l.log.WithContext(ctx).With().Error("cannot serialize layer blocks response", log.Err(err))
+		l.log.WithContext(ctx).With().Panic("failed to serialize layer blocks response", log.Err(err))
 	}
 
-	return server.SerializeResponse(l.log, out, err)
+	return out, nil
```
antonlerner, line 251: shouldn't we retiurn some sort of error to the caller here? if serialisation fails `out` will be nil

countvonzero: i changed the code so that if we failed to serialize layerBlocks struct, it panics. so at this point `out` is never nil

p2p/server/msgserver.go
```diff
@@ -49,29 +50,73 @@ func extractPayload(m Message) []byte {
...
 	failCallBack func(err error)
 }
 
+// ErrShuttingDown is returned to the peer when the node is shutting down
+var ErrShuttingDown = errors.New("node is shutting down")
+
+// ErrBadRequest is returned to the peer upon failure to parse the request
+var ErrBadRequest = errors.New("unable to parse request")
+
+var errRequestTimeout = errors.New("request timed out")
```
dshulyak: better to make it public so that requestor can identify if request timed out with errors.Is

dshulyak (approved):

antonlerner (approved):

p2p/server/msgserver.go
countvonzero: done

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #2667, opened 2021-08-16
[Merged by Bors] - Add codec module and make sure that it is used everywhere

dshulyak:
## Motivation

This is a preparatory refactoring to simplify future changes to the codec (which will be needed even in case if we will use .x files).
It also has a small performance optimization in form of pooled buffer for encoding.

## Changes
- Encode/Decode and EncodeTo/DecodeFrom for stream encoding
- Encode will use pooled buffer to reduce allocations

## Test Plan
unit tests

## TODO

## DevOps Notes


noamnelke (approved):

hare/builder.go
```diff
@@ -71,13 +63,11 @@ type innerMessage struct {
 
 // Bytes returns the message as bytes.
 func (im *innerMessage) Bytes() []byte {
-	var w bytes.Buffer
-	_, err := xdr.Marshal(&w, im)
+	buf, err := types.InterfaceToBytes(im)
 	if err != nil {
-		log.Panic("could not marshal InnerMsg before send")
+		panic("could not marshal InnerMsg before send")
```
antonlerner: log.panic wraps panic in json, panic outputs the error to err stream, maybe it's best to stay with log.Panic

dshulyak: added it back. i just think that if we created struct ourselves encoding shouldn't fail when running a node, and therefore panic is sufficient, but we can keep log.Panic as well

dshulyak:
bors merge

nkryuchkov (approved):

Issue #2545, opened 2021-07-11
Replace dynamic XDR codec with more efficient encoding


antonlerner:
## Description
Currently we use dynamic encoding / decoding of our structs we should move to use statically defined .x files that describes the structs primitive types for better performance.

avive:
i think that the time to implement this for transactions encoding should probably be on or after the new tx model (based on account abstraction) is implemented in the node which we need to do for 0.3.

dshulyak:
Is there a library that can generate golang code from .x files?

dshulyak:
I suggest we tackle it before MN, current xdr library is not optimal and we should check if xdr using .x files will address its issues. If not we should switch to other options (cbor, flatbuffers, maybe something else).

moshababo:
Pending https://github.com/spacemeshos/go-spacemesh/issues/2686.

PR #2668, opened 2021-08-16
[Merged by Bors] - Refactor keys for database code and use tree operations in meshdb

### LINKED ISSUE #2653: Refactor database keys without fmt.Sprintf
## Description

In multiple places database key is stored like:

```go
func getAtxHeaderKey(atxID types.ATXID) []byte {
	return []byte(fmt.Sprintf("h_%v", atxID.Bytes()))
}
```

This creates a key of length ~120bytes, and this is actually how it looks like:

```
h_[16 132 199 253 72 148 199 72 229 207 249 82 22 182 128 154 193 23 176 72 233 200 38 62 192 45 45 200 109 34 50 188]
```

Note that `[` and empty lines are all separate bytes.

Instead this key should be made like this:

```go
var b bytes.Buffer
b.WriteString("h")
b.Write(atx.ID())
```

Which will result only in 33 bytes. 

Beside this issue the conversion from string to []byte at end copies data additional time.

## Affected code

almost all database code

dshulyak:
## Motivation

Closes: https://github.com/spacemeshos/go-spacemesh/issues/2653
Related to: https://github.com/spacemeshos/go-spacemesh/issues/2545

Additionally in meshdb collection of block ids is now stored using leveldb tree instead of updating a slice every time.

## Changes
- refactor all db keys to remove fmt.Sprintf or other unnecessary operations
- refactor operations on collections of ids in meshdb
- in some places BytesToInterface call was unnecessary as we can simply cast bytes to hash

## Test Plan
unit tests

## TODO

## DevOps Notes

### DIFF
--- a/activation/atxdb.go
+++ b/activation/atxdb.go
@@ -22,12 +22,12 @@ import (
 
 const topAtxKey = "topAtxKey"
 
-func getNodeAtxKey(nodePK string, targetEpoch types.EpochID) []byte {
-	return append(getNodeAtxPrefix(nodePK), util.Uint64ToBytesBigEndian(uint64(targetEpoch))...)
+func getNodeAtxKey(nodeID types.NodeID, targetEpoch types.EpochID) []byte {
+	return append(getNodeAtxPrefix(nodeID), util.Uint64ToBytesBigEndian(uint64(targetEpoch))...)
 }
 
-func getNodeAtxPrefix(nodePK string) []byte {
-	return []byte(fmt.Sprintf("n_%v_", nodePK))
+func getNodeAtxPrefix(nodeID types.NodeID) []byte {
+	return []byte(fmt.Sprintf("n_%v_", nodeID.Key))
 }
 
 func getNodeAtxEpochKey(epoch types.EpochID, nodeID types.NodeID) []byte {
@@ -425,7 +425,7 @@ func (db *DB) StoreAtx(ech types.EpochID, atx *types.ActivationTx) error {
 		return err
 	}
 
-	err = db.addAtxToNodeID(atx.NodeID.Key, atx)
+	err = db.addAtxToNodeID(atx.NodeID, atx)
 	if err != nil {
 		return err
 	}
@@ -534,8 +534,8 @@ func (db *DB) getAtxTimestamp(id types.ATXID) (time.Time, error) {
 }
 
 // addAtxToNodeID inserts activation atx id by node
-func (db *DB) addAtxToNodeID(nodePK string, atx *types.ActivationTx) error {
-	err := db.atxs.Put(getNodeAtxKey(nodePK, atx.PubLayerID.GetEpoch()), atx.ID().Bytes())
+func (db *DB) addAtxToNodeID(nodeID types.NodeID, atx *types.ActivationTx) error {
+	err := db.atxs.Put(getNodeAtxKey(nodeID, atx.PubLayerID.GetEpoch()), atx.ID().Bytes())
 	if err != nil {
 		return fmt.Errorf("failed to store atx ID for node: %v", err)
 	}
@@ -569,7 +569,7 @@ type ErrAtxNotFound error
 
 // GetNodeLastAtxID returns the last atx id that was received for node nodeID
 func (db *DB) GetNodeLastAtxID(nodeID types.NodeID) (types.ATXID, error) {
-	nodeAtxsIterator := db.atxs.Find(getNodeAtxPrefix(nodeID.Key))
+	nodeAtxsIterator := db.atxs.Find(getNodeAtxPrefix(nodeID))
 	// ATX syntactic validation ensures that each ATX is at least one epoch after a referenced previous ATX.
 	// Contextual validation ensures that the previous ATX referenced matches what this method returns, so the next ATX
 	// added will always be the next ATX returned by this method.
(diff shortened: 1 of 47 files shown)


dshulyak:
bors try

activation/atxdb.go
```diff
@@ -20,34 +20,58 @@ import (
...
-const topAtxKey = "topAtxKey"
+const (
+	namespaceAtx       = "a"
+	namespaceEpoch     = "e"
+	namespaceTop       = "p"
+	namespaceHeader    = "h"
+	namespaceTimestamp = "t"
+)
 
 func getNodeAtxKey(nodePK string, targetEpoch types.EpochID) []byte {
```
countvonzero: what do you think about putting all the leveldb keys functions and prefixes in one file? for me it's better readability and potentially easier to prevent collision.

activation/identities.go
```diff
@@ -18,21 +17,19 @@ func NewIdentityStore(db database.Database) *IdentityStore {
 	return &IdentityStore{db}
 }
 
-func getKey(key string) [32]byte {
-	return sha256.Sum256(util.Hex2Bytes(key))
+func getKey(key string) []byte {
```
countvonzero, line 20: since the key can be arbitrary string, can this change cause the size of key become variable?

activation/poetdb.go
```diff
@@ -60,7 +60,9 @@ func (db *PoetDb) ValidateAndStoreMsg(data []byte) error {
 
 // Validate validates a new PoET proof.
 func (db *PoetDb) Validate(proof types.PoetProof, poetID []byte, roundID string, signature []byte) error {
-
+	if len(poetID) < 5 {
```
countvonzero: what is 5? maybe define a constant to explain?

mesh/meshdb.go
```diff
@@ -275,43 +275,40 @@ func (m *DB) ForBlockInView(view map[types.BlockID]struct{}, layer types.LayerID
...
-		return []types.BlockID{}, nil
+	layerBuf := index.Bytes()
+	zero := false
+	ids := []types.BlockID{}
+	it := m.layers.Find(layerBuf)
+	// TODO(dshulyak) iterator must be able to return an error
+	for it.Next() {
+		if len(it.Key()) == len(layerBuf) {
+			zero = true
+			continue
```
countvonzero, line 287: should this be `break` instead? i guess if we have late blocks for this layer we don't clear the zero block entry.

mesh/meshdb.go
```diff
@@ -275,43 +275,40 @@ func (m *DB) ForBlockInView(view map[types.BlockID]struct{}, layer types.LayerID
...
+		}
+		var id types.BlockID
+		copy(id[:], it.Key()[len(layerBuf):])
```
countvonzero, line 290: what's the diff bwtn copy() and BytesToInterface here? i'd imagine the latter is stronger typed than former. i am thinking if the data is corrupted somehow.

mesh/meshdb.go
```diff
@@ -275,43 +275,40 @@ func (m *DB) ForBlockInView(view map[types.BlockID]struct{}, layer types.LayerID
...
+	err := m.layers.Put(index.Bytes(), nil)
 	if err == nil {
 		m.persistLayerHash(index, EmptyLayerHash)
 	}
 	return err
 }
 
 func (m *DB) getBlockBytes(id types.BlockID) ([]byte, error) {
-	return m.blocks.Get(id.AsHash32().Bytes())
+	return m.blocks.Get(id.Bytes())
```
countvonzero, line 313: do you think this should also be prefixed?

mesh/meshdb.go
```diff
@@ -440,43 +428,15 @@ func (m *DB) writeBlock(bl *types.Block) error {
...
-	m.layers.Put(blk.LayerIndex.Bytes(), w)
-	return nil
-}
-
-func (m *DB) getLayerHashKey(layerID types.LayerID) []byte {
-	return []byte(fmt.Sprintf("layerHash_%v", layerID.Bytes()))
+	var b bytes.Buffer
+	b.Write(blk.LayerIndex.Bytes())
+	b.Write(blk.ID().Bytes())
+	return m.layers.Put(b.Bytes(), nil)
```
countvonzero, line 436: this code doesn't seem to preserve previous blocks saved for this layer.

countvonzero (commented):
i like the simplification that removes the locking in meshdb.go

activation/poetdb.go
dshulyak: the 5 was used below for short id, and it was crashing if id was empty

activation/identities.go
dshulyak: we are using public key of the node here, but it also doesn't matter much if the length will be variable

activation/atxdb.go
dshulyak: do you mean like for all databases? because we are using multiple isolated databases in current code

mesh/meshdb.go
dshulyak: :+1:

mesh/meshdb.go
dshulyak: actually, in the previous code, BlockID was always written without BytesToInterface and was unmarshalled using it. it wasn't a problem with xdr because apparently it was noop, but it is not the case with most of the formats.

as for the safety, i don't think that there is anything beneficial in using codec if we have bytes already. db will usually check crc for corruption after loading stuff from disk.

mesh/meshdb.go
dshulyak: it should be. i left it because of dependency in the fetcher, but i can change it in this pr

mesh/meshdb.go
dshulyak: what do you mean? i think all previous writes are untouched, e.g. if we do multiple writes such as:
1. layer_01_block_01
2. layer_01_block_02
all of them will be in the store

activation/atxdb.go
countvonzero: yes. i know we are using multiple isolated databases in current code. but it would be nice to hide those details somewhere else instead of mixing with application logic. but your call.

mesh/meshdb.go
countvonzero: i misread. yes. all others are preserved.

mesh/meshdb.go
countvonzero: a fixme works for me too

activation/poetdb.go
countvonzero: i was thinking changing it to a named constant makes the code more readable. up to you.

countvonzero (approved):

activation/atxdb.go
dshulyak: i think it makes more sense to do it when we will be merging them

mesh/meshdb.go
dshulyak: the TestMesh_SetZeroBlockLayer started to fail. i think i will keep continue to preserve original behavior for now

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

Issue #2629, opened 2021-08-03
Fix vector and threshold arithmetic in verifying tortoise


lrettig:
## Description

Verifying tortoise and self-healing currently use floating-point math to determine whether net votes (for/against a block) cross a threshold:

https://github.com/spacemeshos/go-spacemesh/blob/4c7a6da8ed4ef10c3f2ec4cb48802f0be26eadcb/tortoise/util.go#L75-L95

Refactor this to remove the floating-point math.

In addition, the vector type (same file) stores votes as uint64, because it's downstream of block weights (which are uint64). This makes arithmetic, such as calculating net votes, unsafe and requires extra checks:

https://github.com/spacemeshos/go-spacemesh/blob/4c7a6da8ed4ef10c3f2ec4cb48802f0be26eadcb/tortoise/util.go#L44-L52

We should probably just store votes using uint32 instead (and perform the safety/overflow check once, on instantiation instead).

Thank you to @noamnelke for raising these issues in this review: https://github.com/spacemeshos/go-spacemesh/pull/2400#pullrequestreview-719642912.

## Affected code
- tortoise/verifying_tortoise.go
- tortoise/util.go

This issue appears in commit hash: df7c1c1f34deafc6d107f8c167fb204c11ffae7c

noamnelke:
After giving it more thought, I'm not sure that `uint32` will be enough for the weight - we want to maximize the supported range there since we can't predict or cap the maximal total weight in an epoch. So I think that using `uint64` is better.

I've also made a mistake in my review comment - `int64` doesn't have the same positive range as `uint32`, it has a much larger range. The sign bit halves the positive range, but a `uint64` has double the *bits* vs. `uint32`, not double the range!
```
int32              -2,147,483,648 to              2,147,483,647
uint32                          0 to              4,294,967,295
int64  -9,223,372,036,854,775,808 to  9,223,372,036,854,775,807
uint64                          0 to 18,446,744,073,709,551,615
```

I still think the best solution is to separate the sign bit and keep using `uint64`, but if this complicates other code, then maybe keeping the current implementation (`int64` with overflow check) is not too bad.

dshulyak:
Can someone explain what is the problem with float64/float32 ? I am not very familiar with this topic and would like to know more.
I assumed that this is because it is not deterministic (depends on the environment somehow), but after some googling it seems that it is deterministic - https://stackoverflow.com/questions/42181795/is-ieee-754-2008-deterministic.

[...]

PR #2683, opened 2021-08-17
[Merged by Bors] - tortoise: refactor trtl.Persist to use its own store and do the task more efficiently

dshulyak:
## Motivation
My main motivation was to simplify encoding in turtle.Persist. turtle.Persist also shows up in profiles (both in cpu and allocations) and i thought that it will be more efficient without encoding whole thing at once.

```
BenchmarkStatePersist
BenchmarkStatePersist/New
BenchmarkStatePersist/New-16         	    4899	    252503 ns/op	  140557 B/op	    2625 allocs/op
BenchmarkStatePersist/Repeat
BenchmarkStatePersist/Repeat-16      	   52162	     22840 ns/op	     496 B/op	      16 allocs/op
BenchmarkStatePersist/Old
BenchmarkStatePersist/Old-16         	    2386	    461755 ns/op	  148863 B/op	    5432 allocs/op

```

BenchmarkStatePersist/Repeat is using diff mode. It still has to loop through every diff, we can improve it by writing to the database in HandleIncomingLayer directly but it will require a bit more changes.

related to: https://github.com/spacemeshos/go-spacemesh/issues/2545
## Changes
- added state to the turtle instead of relying on mesh.TORTOISE key, which seemed quite hacky to me
- tested state recovery with testing/quick, also added simple benchmark for new and old Persist version
- simplified initialized
- added support for diff mode, that will persist only non-flushed records from collections and layers

## Test Plan
unit tests

## TODO

## DevOps Notes


dshulyak:
this can wait until self-healing is merged, i will fix conflicts afterwards

dshulyak:
bors try

dshulyak:
@lrettig i think we will want to merge this change into/before self-healing. if i read profiles correctly node will stop crashing. unless there are other ideas

tortoise/state.go
```diff
@@ -0,0 +1,176 @@
...
+		b.Reset()
+	}
+
+	for layer, blocks := range s.BlockOpinionsByLayer {
+		for block1, opinions := range blocks {
+			for block2, opinion := range opinions {
+				if opinion.Flushed && s.diffMode {
+					continue
+				}
+				opinion.Flushed = true
```
antonlerner, line 73: mark as flushed after `batch.Put` succeeds

dshulyak:
bors try

dshulyak:
bors try

tortoise/util.go
```diff
@@ -2,50 +2,61 @@ package tortoise
...
+
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
-type vec [2]int
+// NOTE(dshulyak) there is a bug in xdr. without specifying size xdr will cut the value
+// down to int32
+type vec struct {
+	Support, Against int64
```
lrettig: We use uint64 for this in healing branch. see #2629

tortoise/verifying_tortoise.go
```diff
@@ -357,18 +338,18 @@ func (t *turtle) calculateExceptions(layerid types.LayerID, blockid types.BlockI
 	return []map[types.BlockID]struct{}{againstDiff, forDiff, neutralDiff}, nil
 }
 
-func (t *turtle) BlockWeight(voting, voted types.BlockID) int {
+func (t *turtle) BlockWeight(voting, voted types.BlockID) int64 {
```
lrettig: should be uint64, weight can never be negative

tortoise/verifying_tortoise.go
```diff
@@ -84,13 +67,11 @@ func (t *turtle) init(genesisLayer *types.Layer) {
 	t.BlockOpinionsByLayer[genesisLayer.Index()] = make(map[types.BlockID]Opinion)
 	for _, blk := range genesisLayer.Blocks() {
 		id := blk.ID()
-		t.BlockOpinionsByLayer[genesisLayer.Index()][blk.ID()] = Opinion{
-			BlocksOpinion: make(map[types.BlockID]vec),
-		}
-		t.GoodBlocksIndex[id] = struct{}{}
+		t.BlockOpinionsByLayer[genesisLayer.Index()][id] = Opinion{}
+		t.GoodBlocksIndex[id] = false
```
lrettig: this should be `true` shouldn't it?

tortoise/verifying_tortoise.go
```diff
@@ -357,18 +338,18 @@ func (t *turtle) calculateExceptions(layerid types.LayerID, blockid types.BlockI
...
 func (t *turtle) persist() error {
-	return t.bdp.Persist(mesh.TORTOISE, t)
-}
-
-// RecoverVerifyingTortoise retrieve latest saved tortoise from the database
-func RecoverVerifyingTortoise(mdb retriever) (interface{}, error) {
-	return mdb.Retrieve(mesh.TORTOISE, &turtle{})
+	// TODO(dshulyak) make state eviction a part of regular eviction
+	// HandleIncomingLayer should expose an error for this
+	if err := t.state.Evict(); err != nil {
```
lrettig: why perform eviction inside persist (rather than when verifying a layer, as we do now)?

tortoise/verifying_algorithm.go
```diff
@@ -17,43 +19,26 @@ type ThreadSafeVerifyingTortoise struct {
...
-// NewRecoveredVerifyingTortoise recovers a previously persisted tortoise copy from mesh.DB
-func recoveredVerifyingTortoise(mdb blockDataProvider, lg log.Log) *ThreadSafeVerifyingTortoise {
-	tmp, err := RecoverVerifyingTortoise(mdb)
-	if err != nil {
-		lg.Panic("could not recover tortoise state from disc ", err)
+	if err := alg.trtl.Recover(); err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			alg.trtl.init(mesh.GenesisLayer())
+		} else {
+			cfg.Log.Panic("can't recover turtle state", log.Err(err))
```
lrettig:
```suggestion
			cfg.Log.With().Panic("can't recover turtle state", log.Err(err))
```

tortoise/verifying_tortoise.go
```diff
@@ -84,13 +67,11 @@ func (t *turtle) init(genesisLayer *types.Layer) {
 	t.BlockOpinionsByLayer[genesisLayer.Index()] = make(map[types.BlockID]Opinion)
 	for _, blk := range genesisLayer.Blocks() {
 		id := blk.ID()
-		t.BlockOpinionsByLayer[genesisLayer.Index()][blk.ID()] = Opinion{
-			BlocksOpinion: make(map[types.BlockID]vec),
-		}
-		t.GoodBlocksIndex[id] = struct{}{}
+		t.BlockOpinionsByLayer[genesisLayer.Index()][id] = Opinion{}
```
lrettig, line 162: don't you still need to initialize the `Opinion` map?

tortoise/verifying_tortoise.go
```diff
@@ -485,7 +466,7 @@ func (t *turtle) HandleIncomingLayer(newlyr *types.Layer) (pbaseOld, pbaseNew ty
 			t.checkBlockAndGetInputVector(b.NeutralDiff, "abstain", abstain, b.ID(), baseBlock.LayerIndex) {
 			// (2) all diffs appear after the base block and are consistent with the input vote vector
 			t.logger.With().Debug("marking good block", b.ID(), b.LayerIndex)
-			t.GoodBlocksIndex[b.ID()] = struct{}{}
+			t.GoodBlocksIndex[b.ID()] = false
```
lrettig: should be `true`

tortoise/state.go
```diff
@@ -0,0 +1,176 @@
...
+				opinions[block2] = opinion
+
+				b.WriteString(namespaceOpinons)
+				b.Write(encodeLayerKey(layer))
+				b.Write(block1.Bytes())
+				b.Write(block2.Bytes())
+
+				buf, err := codec.Encode(opinion)
+				if err != nil {
+					s.log.Panic("can't encode vec", log.Err(err))
```
lrettig:
```suggestion
					s.log.With().Panic("can't encode vec", log.Err(err))
```

lrettig (commented):
I'm not sure how I feel about the custom encoder for tortoise. Let's discuss with @antonlerner

tortoise/state.go
dshulyak: it doesn't matter. actual write will happen in batch.Write

tortoise/verifying_tortoise.go
dshulyak: Opinion is a map now

tortoise/verifying_tortoise.go
dshulyak: i kept it here because beacause HandleIncomingLayer wasn't exposing error

dshulyak:
> I'm not sure how I feel about the custom encoder for tortoise. 

what is custom encoder? writing state the way it is implemented now is not an option tbh

as i mentioned already, encoding 200mb of state requires a lot of memory allocated at once by codec. also every write to a leveldb (or in fact any lsm based storage) is not an inplace update like with sqlite or postgres. it appends data and later it needs to compact this state with lower levels, which will use a lot of memory and will block other writes. so current solution is simply not something that can be used in principal 

one node in the devnet stores state for 1000 layers and is still far from crashing. previously it was crashing once it stored state for 150 layers.

tortoise/verifying_tortoise.go
dshulyak: true is for flushed

tortoise/verifying_tortoise.go
dshulyak: true is for flushed

tortoise/util.go
dshulyak: can it be limited to uint32? will save some inuse memory. given that we will be storing millions of votes in mainnet

dshulyak: based on last observation, opinion layers 18-1057. total 6173753. evict from 28. last 1057. verified 38. reducing it to uint32 will save ~50mb. so maybe won't be that important

antonlerner:
>one node in the devnet stores state for 1000 layers and is still far from crashing. previously it was crashing once it stored state for 150 layers.

why is the state so large? this is probably the root cause for this large memory consumption

dshulyak:
> why is the state so large? this is probably the root cause for this large memory consumption

are you asking about 1000 or 150? 150 we had in the last devnet that crashed, i think @lrettig can explain what size we need. i saw a comment that in mainnet we will want to store for 1000. 

anyway this pattern is still super inefficient

dshulyak:
bors try

dshulyak:
failed with k8s error in latenodes:
```
E           HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"DaemonSet in version \"v1\" cannot be handled as a DaemonSet: v1.DaemonSet.Spec: v1.DaemonSetSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.NodeSelector: ReadString: expects \" or n, but found 5, error found in #10 byte of ...|espace\": 59409}, \"to|..., bigger context ...|eriodSeconds\": 10, \"nodeSelector\": {\"namespace\": 59409}, \"tolerations\": [{\"key\": \"namespace\", \"opera|...","reason":"BadRequest","code":400}
```

bors try

tortoise/util.go
lrettig: no, it needs to encode voting weights, and voting weights are uint64. this would be a very deep change to varpost.

tortoise/verifying_tortoise.go
lrettig: exactly. don't you need to initialize it with `make`?

tortoise/verifying_tortoise.go
lrettig: I don't understand. this line (and the other, similar one) very clearly looks like it's marking the block with id `id` as not good (`false`). am I missing something?

tortoise/verifying_tortoise.go
dshulyak: but Opinion{} and make(Opinion) will do the same thing for map, unless you want to specify length

tortoise/verifying_tortoise.go
dshulyak: so if block is present in this map it means that the block is good. if the value is set to true it means that it was flushed to the disk, and doesn't need to be flushed again

dshulyak:
last stats from a devnet using this branch + self healing: opinion layers 134-338. total 7358437. last evicted 236. last 338. verified 337. so the opinions size seems to be bounded by ~2*windowSize. it consumes around ~700mb to store this data in memory. but the good thing is that network is not even close to crashing due to memory usage.

@lrettig can you clarify please if we need to support windowSize = 1000 in mainnet? it basically will require each node to use 7gb just to store votes, so we will have to figure out something in order to support that.

lrettig:
> @lrettig can you clarify please if we need to support windowSize = 1000 in mainnet? it basically will require each node to use 7gb just to store votes, so we will have to figure out something in order to support that.

This is a question for research, will discuss with @tal-m

tortoise/verifying_tortoise.go
lrettig: okay. that's counterintuitive from the design. please document that clearly :) thanks!

dshulyak:
@lrettig @antonlerner i pushed rebased changes. let me know if you want me to change something

dshulyak:
Weird, i have different hare tests hanging now

dshulyak:
lol, why they all passed if executed separately

dshulyak:
actually not hare i was confused by logs. looks like this TestLayerPatterns/heal_after_bad_layers test hangs on CI

dshulyak:
It is even more convoluted than that.. there are two tests in hare that took 14 minutes to execute
```
2021-08-27T08:57:11.5759012Z === RUN   Test_multipleCPs
2021-08-27T09:01:49.2218036Z === RUN   Test_multipleCPsAndIterations
2021-08-27T09:11:06.1888346Z === RUN   Test_HarePreRoundEmptySet
```

common/types/block.go
```diff
@@ -15,6 +15,14 @@ import (
 	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
+const (
+	// LayerIDSize in bytes.
+	LayerIDSize = 4
+	// BlockIDSize in bytes.
+	// FIXME(dshulyak) why do we cast to hash32 when returning bytes?
```
lrettig, line 22: Could you please open (and link) an issue to track/discuss this?

p2p/service/sim.go
```diff
@@ -78,28 +78,26 @@ func (s *Simulator) SubscribeToPeerEvents(myid p2pcrypto.Key) (chan p2pcrypto.Pu
...
 		log.Debug("publish new peer on chan with len %v, cap %v", len(ch), cap(ch))
-		select {
-		case ch <- peer:
-		default:
-			log.Warning("unable to publish new peer event to subscriber")
-		}
+		// NOTE(dshulyak) services are not expected to be ready to receive all peers
+		// all tests that are using Simulator will become very flaky, especially on slow
+		// environments
+		ch <- peer
```
lrettig: As I commented in #2730, I'm concerned that this will cause issues in app test, let's see

tortoise/state.go
```diff
@@ -0,0 +1,181 @@
...
+	// current time or latest tick). As far as Tortoise is concerned, Last is the current layer. This is a subjective
+	// view of time, but Tortoise receives layers as soon as Hare finishes processing them or when they are received via
+	// gossip, and there's nothing for Tortoise to verify without new data anyway.
+	Last        types.LayerID
+	LastEvicted types.LayerID
+	Verified    types.LayerID
+	// if key exists in the map the block is good. if value is true it was written to the disk.
+	GoodBlocksIndex map[types.BlockID]bool
+	// use 2D array to be able to iterate from latest elements easily
+	BlockOpinionsByLayer map[types.LayerID]map[types.BlockID]Opinion // records hdist, for each block, its votes about every
```
lrettig: You can remove the comment here and on the two following lines, the context has been lost and I think it's not helpful

tortoise/verifying_tortoise.go
```diff
@@ -173,13 +159,10 @@ func (t *turtle) init(ctx context.Context, genesisLayer *types.Layer) {
 	t.BlockOpinionsByLayer[genesisLayer.Index()] = make(map[types.BlockID]Opinion)
 	for _, blk := range genesisLayer.Blocks() {
 		id := blk.ID()
-		t.BlockOpinionsByLayer[genesisLayer.Index()][blk.ID()] = Opinion{
-			BlockOpinions: make(map[types.BlockID]vec),
-		}
-		t.GoodBlocksIndex[id] = struct{}{}
+		t.BlockOpinionsByLayer[genesisLayer.Index()][id] = Opinion{}
+		t.GoodBlocksIndex[id] = false
```
lrettig:
```suggestion
		t.GoodBlocksIndex[id] = false // false means good block, not flushed
```

tortoise/verifying_tortoise.go
```diff
@@ -718,7 +699,7 @@ func (t *turtle) scoreBlocks(ctx context.Context, blocks []*types.Block) {
 		if t.determineBlockGoodness(ctx, b) {
 			// note: we have no way of warning if a block was previously marked as not good
 			logger.With().Debug("marking block good", b.ID(), b.LayerIndex)
-			t.GoodBlocksIndex[b.ID()] = struct{}{}
+			t.GoodBlocksIndex[b.ID()] = false
```
lrettig:
```suggestion
			t.GoodBlocksIndex[b.ID()] = false // false means good block, not flushed
```

lrettig (changes requested):

lrettig:
bors try

tortoise/verifying_tortoise.go
dshulyak: i commented where this data structure is defined. i don't think that it helps to add it everywhere where map is used

p2p/service/sim.go
dshulyak: it definitely makes tests less flaky. i already merged that change

dshulyak:
@lrettig i removed some debug logs from a loop https://github.com/spacemeshos/go-spacemesh/pull/2683/commits/4a63c34b9d0b917f3b35384b053d499585eda3a0 . even on my machine it takes nearly 100s to print those logs in some tests. if they are needed we must find a more efficient way to log this information

common/types/block.go
dshulyak: https://github.com/spacemeshos/go-spacemesh/issues/2731

tortoise/verifying_tortoise.go
lrettig: I disagree. I find this line of code very counterintuitive - that setting the map member to "false" actually indicates that the block is _good_ as opposed to _bad._ I think the explanatory comment helps a lot.

lrettig:
> @lrettig i removed some debug logs from a loop [4a63c34](https://github.com/spacemeshos/go-spacemesh/commit/4a63c34b9d0b917f3b35384b053d499585eda3a0) . even on my machine it takes nearly 100s to print those logs in some tests. if they are needed we must find a more efficient way to log this information

Rather than deleting the debug log lines entirely, could you leave them commented out? It would be helpful for future debugging. Let's continue the discussion about the logging issue in #2732.

common/types/block.go
lrettig: thanks!

lrettig:
Take a look at #2734, which is going to cause conflicts here

dshulyak:
> Take a look at #2734, which is going to cause conflicts here

just one more part of the db key

dshulyak:
@lrettig is anything else needs to be changed?

dshulyak:
bors try

lrettig (approved):
Approving, but just one question... what happens if, theoretically, you were to "restore" a persisted tortoise but with different non-state params, e.g., Hdist, Zdist, ConfidenceInterval, etc.?

dshulyak:
> Approving, but just one question... what happens if, theoretically, you were to "restore" a persisted tortoise but with different non-state params, e.g., Hdist, Zdist, ConfidenceInterval, etc.?

i don't know precisely what is going to happen since it mostly depends on the tortoise logic. but in general, i think we need a general solution to keep consensus critical parameters safe from accidental change. tortoise, tortoise beacon, post, probably hare all require some parameters that can't change arbitrarily after the network was started. i don't think that we want to save them in each module separately.

@lrettig do you know how eth1/eth2 deals with it? i can check in filecoin

dshulyak:
bors merge

antonlerner:
bors try

PR #2698, opened 2021-08-19
[Merged by Bors] - Fix devnet203 tortoise beacon issues

### LINKED ISSUE #2694: `miner has no ATXs in the previous epoch` warning on devnet203
## Description

Node cannot find miner's ATX in the previous epoch when receives a proposal message

### LINKED ISSUE #2696: `received malformed proposal message: vrf is not verified` warning on devnet203
## Description

For some reason, there're proposal message verification errors on devnet.

nkryuchkov:
## Motivation

Closes #2694 
Closes #2696 

## Changes

- Revert changing `GetNodeLastAtxID` signature
- Fix incorrect usage of miner PK when passing to `GetNodeLastAtxID`
- Accept `types.NodeID` instead of ED public key in `tortoisebeacon.New`
- Send `types.NodeID` instead of ED public key in `ProposalMessage`

## Test Plan

Unit tests

## DevOps Notes

### DIFF
--- a/blocks/blockoracle_test.go
+++ b/blocks/blockoracle_test.go
@@ -13,11 +13,13 @@ import (
 	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
-var atxID = types.ATXID([32]byte{1, 3, 3, 7})
-var nodeID, vrfsgn = generateNodeIDAndSigner()
-var validateVRF = signing.VRFVerify
-var edSigner = signing.NewEdSigner()
-var activeSetAtxs = []types.ATXID{atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID} // 10 ATXs
+var (
+	atxID          = types.ATXID([32]byte{1, 3, 3, 7})
+	nodeID, vrfsgn = generateNodeIDAndSigner()
+	validateVRF    = signing.VRFVerify
+	edSigner       = signing.NewEdSigner()
+	activeSetAtxs  = []types.ATXID{atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID} // 10 ATXs
+)
 
 const defaultAtxWeight = 1024
 
@@ -367,5 +369,4 @@ func TestMinerBlockOracle_GetEligibleLayers(t *testing.T) {
 		}
 	}
 	r.Equal(eligibleLayers, len(blockOracle.GetEligibleLayers()))
-
 }
--- a/blocks/blockoracle_test.go
+++ b/blocks/blockoracle_test.go
@@ -13,13 +13,11 @@ import (
 	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
-var (
-	atxID          = types.ATXID([32]byte{1, 3, 3, 7})
-	nodeID, vrfsgn = generateNodeIDAndSigner()
-	validateVRF    = signing.VRFVerify
-	edSigner       = signing.NewEdSigner()
-	activeSetAtxs  = []types.ATXID{atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID} // 10 ATXs
-)
+var atxID = types.ATXID([32]byte{1, 3, 3, 7})
+var nodeID, vrfsgn = generateNodeIDAndSigner()
+var validateVRF = signing.VRFVerify
+var edSigner = signing.NewEdSigner()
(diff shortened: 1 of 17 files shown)


nkryuchkov:
bors try

countvonzero:
bors try

blocks/blockoracle_test.go
```diff
@@ -13,13 +13,11 @@ import (
...
 )
 
-var (
-	atxID          = types.ATXID([32]byte{1, 3, 3, 7})
-	nodeID, vrfsgn = generateNodeIDAndSigner()
-	validateVRF    = signing.VRFVerify
-	edSigner       = signing.NewEdSigner()
-	activeSetAtxs  = []types.ATXID{atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID, atxID} // 10 ATXs
-)
+var atxID = types.ATXID([32]byte{1, 3, 3, 7})
```
dshulyak: i don't see why this block needs to be changed

blocks/blockoracle_test.go
```diff
@@ -369,4 +367,5 @@ func TestMinerBlockOracle_GetEligibleLayers(t *testing.T) {
 		}
 	}
 	r.Equal(eligibleLayers, len(blockOracle.GetEligibleLayers()))
+
```
dshulyak: ?

tortoisebeacon/handlers.go
```diff
@@ -295,7 +295,8 @@ func (tb *TortoiseBeacon) verifyFirstVotingMessage(message FirstVotingMessage, c
 
 	// TODO(nkryuchkov): Ensure that epoch is the same.
 
-	atxID, err := tb.atxDB.GetNodeAtxIDForEpoch(minerPK.String(), currentEpoch-1)
+	nodeID := types.NodeID{Key: minerPK.String()}
```
dshulyak, line 298: i dont think that this is a good solution to use nodeID. someone can change the implementation of the GtNodeAtxIDForEpoch to rely on VRFPublicKey and it will be perfectly valid, but introduce failures in tortoisebeacon

dshulyak (approved):

tortoisebeacon/handlers.go
```diff
@@ -319,7 +320,7 @@ func (tb *TortoiseBeacon) verifyFirstVotingMessage(message FirstVotingMessage, c
 	defer tb.consensusMu.Unlock()
 
 	if tb.hasVoted[firstRound] == nil {
-		tb.hasVoted[firstRound] = make(map[nodeID]struct{})
+		tb.hasVoted[firstRound] = make(map[nodePK]struct{})
```
dshulyak: nodePK alias to string doesn't make much sense to me

dshulyak:
i will try to merge it to test if it helps with unmanaged miners syncing

bors merge

dshulyak:
p2p test failed

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
https://github.com/spacemeshos/go-spacemesh/issues/2700

dshulyak:
latenodes failed again

dshulyak:
bors merge

dshulyak:
bors merge

blocks/blockoracle_test.go
nkryuchkov: I have an enabled `gofumpt` file watcher, so it does auto-formatting when I change files. When I changed the signature of `GetNodeAtxIDForEpoch` the watcher auto-formatted changed files. Now I reverted that commit, so some formatting fixes were reverted as well

blocks/blockoracle_test.go
nkryuchkov: It was auto-formatted in the reverted commit, I'll fix

tortoisebeacon/handlers.go
nkryuchkov: What solution would you suggest?

tortoisebeacon/handlers.go
nkryuchkov: Removed

nkryuchkov:
bors merge

tortoisebeacon/handlers.go
dshulyak: string was better, still ambiguous in other way but better than nodeID. 
in my opinion the best would be to use []byte, but i don't imply that we should switch to []byte in this change.

blocks/blockoracle_test.go
dshulyak: oh thatswhy. i think i am not using gofumpt, so this changes seemed random to me

PR #2709, opened 2021-08-23
[Merged by Bors] - use layer id to query layer content btwn peers

### LINKED ISSUE #2693: devnet: log "requested unknown layer hash"
## Description

this log entry is generated for 2 reasons
- for genesis block. every node has the same genesis block but the hash of the genesis block is not persisted. so when a node asks its peers about the hash of the genesis layer, the peers are able to return the hash. however, when the node asks the same peer about the content of the hash, it is not able to answer it.
- for other blocks. there is a discrepancy between `mesh.GetLayerHash()` and `mesh.GetLayerHashBlocks()`. mesh.GetLayerHash() returns non-persisted hash, which will lead to `mesh.GetLayerHashBlocks()` to log `requested unknown layer hash`

the motivation for not persisting hash as blocks come in is that we don't want to persist hash for late blocks, and only persist layer hash after tortoise validates it. it seems we should still persist the layer hash **before** tortoise validates the layer, and leave it unchanged **after** tortoise validates it, unless tortoise re-validates the layer again.

## Steps To Reproduce
1.  make dockertest-late-nodes namespace=$YOUR_NAMESPACE dump=True
`layer hash not yet persisted`
`responded layer hash request`
`requested unknown layer hash`
for the same node and layer hash.

## Expected Behavior

- layer hash should be persisted as blocks are received before tortoise validates the layer
- after tortoise validates the layer, late blocks should not affected the persisted layer hash and only tortoise itself can change the persisted layer hash as it re-validates a layer

### LINKED ISSUE #2704: devnet: improve debugging by fetching layer by layer ID (instead of hash)
## Description
while debugging devnet failure, it gets pretty gnarly when using layer hash (of block IDs) to look up layer data.

using layer hash to look up layer contents serves 2 purposes:
- the layer id -> layer hash mapping allows verification of the blocks received for this layer
- the layer hash (combined with aggregated layer hash) itself allows efficient syncing between two layers with different mesh hash

currently when syncer queries about a layer content, it 
A. queries a peer about the layer hash (1 RTT)
B. queries the same peer about the contents corresponding to the hash (1RTT)

and when the peer received the request B, it does the following:
1. receive req msg as layer hash
2. look up layer hash -> layer ID
3. look up layer ID -> block IDs
4. return to peer block IDs

the proposal is to remove request A and step 1 and 2 from the responding peer. i.e. use layer ID to query a layer content directly. the pros:
- save 1 RTT
- remove leveldb entries for layerHash->layerID mapping
- ease of debugging

## Expected Behavior
the node fetches layer content correctly and still am able to verify peer's integrity w.r.t layer hash -> layer contents

countvonzero:
## Motivation

Closes #2693
Closes #2704

## Changes

- remove the step that query layer hashes and directly query layer contents by layer ID

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -191,24 +191,23 @@ func (l *Logic) epochATXsReqReceiver(ctx context.Context, msg []byte) ([]byte, e
 // it also returns the validation vector for this data and the latest blocks received in gossip
 func (l *Logic) layerBlocksReqReceiver(ctx context.Context, req []byte) ([]byte, error) {
 	lyrID := types.BytesToLayerID(req)
-	blockIDs, err := l.layerDB.LayerBlockIds(lyrID)
-	if err != nil {
-		l.log.WithContext(ctx).With().Debug("failed to get layer content", lyrID, log.Err(err))
-		return nil, ErrInternal
-	}
-
-	inputVector, err := l.layerDB.GetLayerInputVectorByID(lyrID)
-	if err != nil {
-		// best effort with input vector
-		l.log.WithContext(ctx).With().Debug("failed to get input vector for layer", lyrID, log.Err(err))
-	}
-
 	b := &layerBlocks{
-		Blocks:         blockIDs,
-		InputVector:    inputVector,
 		ProcessedLayer: l.layerDB.ProcessedLayer(),
 		Hash:           l.layerDB.GetLayerHash(lyrID),
 		AggregatedHash: l.layerDB.GetAggregatedLayerHash(lyrID),
+	}
+	var err error
+	b.Blocks, err = l.layerDB.LayerBlockIds(lyrID)
+	if err != nil {
+		if err != database.ErrNotFound {
+			l.log.WithContext(ctx).With().Debug("failed to get layer content", lyrID, log.Err(err))
+			return nil, ErrInternal
+		}
+	} else {
+		if b.InputVector, err = l.layerDB.GetLayerInputVectorByID(lyrID); err != nil {
+			// best effort with input vector
+			l.log.WithContext(ctx).With().Debug("failed to get input vector for layer", lyrID, log.Err(err))
+		}
 	}
 
 	out, err := types.InterfaceToBytes(b)
@@ -242,7 +241,6 @@ func (l *Logic) tortoiseBeaconReqReceiver(ctx context.Context, data []byte) ([]b
 // PollLayerContent polls peers for the content of a given layer ID.
 // it returns a channel for the caller to be notified when responses are received from all peers.
 func (l *Logic) PollLayerContent(ctx context.Context, layerID types.LayerID) chan LayerPromiseResult {
-	l.log.WithContext(ctx).With().Info("polling layer content", layerID)
 	resChannel := make(chan LayerPromiseResult, 1)
 
 	remotePeers := l.net.GetPeers()
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -13,7 +13,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/database"
 	"github.com/spacemeshos/go-spacemesh/fetch"
 	"github.com/spacemeshos/go-spacemesh/log"
-	"github.com/spacemeshos/go-spacemesh/mesh"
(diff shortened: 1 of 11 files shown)


layerfetcher/layers.go
```diff
@@ -495,7 +362,7 @@ func notifyLayerBlocksResult(layerID types.LayerID, channels []chan LayerPromise
 	hasZeroBlockHash := false
 	var firstErr error
 	for _, res := range peerResult {
-		if res.data != nil {
+		if res.err == nil && res.data != nil {
```
antonlerner, line 363: could there be a case where both `err` and `data` are nil?

layerfetcher/layers.go
```diff
@@ -403,74 +264,80 @@ func (l *Logic) PollLayerBlocks(ctx context.Context, layerID types.LayerID, hash
...
 	// send a request to the first peer of the list to get blocks data.
 	// todo: think if we should aggregate or ask from multiple peers to have some redundancy in requests
-	numHashes := len(hashes)
-	for hash, remotePeers := range hashes {
-		peer := remotePeers[0]
-		if hash == mesh.EmptyLayerHash {
-			l.receiveBlockHashes(ctx, layerID, peer, numHashes, nil, ErrZeroLayer)
-			continue
-		}
+	for _, p := range remotePeers {
```
antonlerner, line 265: previous implementation used hashes to avoid requesting all peers for the same data, now as I understand the new flow if all peers will return same blocks than they will be queried again for all the blocks
right?

layerfetcher/layers.go
```diff
@@ -403,74 +264,80 @@ func (l *Logic) PollLayerBlocks(ctx context.Context, layerID types.LayerID, hash
...
-			pRes.err = convertErr
-		} else {
-			pRes.data = data
-			l.fetchLayerBlocks(ctx, layerID, &blocks)
-		}
+func (l *Logic) receiveLayerContent(ctx context.Context, layerID types.LayerID, peer peers.Peer, expectedResults int, data []byte, peerErr error) {
+	l.log.WithContext(ctx).With().Debug("received layer content from peer", log.String("peer", peer.String()))
+	peerRes := extractPeerResult(l.log.WithContext(ctx), layerID, data, peerErr)
+	if peerRes.err == nil {
+		l.fetchLayerBlocks(ctx, layerID, peerRes.data)
```
antonlerner, line 332: will this try to fetch same blocks if they were received from different peers?

layerfetcher/layers.go
countvonzero: indeed. this is the price of this PR. 
previous behavior
1. RTT to N peers to get layer hash, payload: N*32bytes (Hash32)
2. RTT to M peers to grab layer block IDs, payload: 32byte*NumBlocks*M, M is probably 1 or 2
3. RTT to X random peers grab block contents
behavior after this PR
1. RTT to N peers to get layer block IDs, payload: 32bytes*NumBlocks*N
2. RTT to X random peers grab block contents

but because we do have a upper-bound for N (num peers), i think this is acceptable. 

another option is to use half of the peers as a future optimization? for now i feel that simplicity is most important.

layerfetcher/layers.go
countvonzero: no. this happens in fetcher.go, which groups by hash. if there is already a request for the same hash pending, it will wait for the pending request for the same hash.

layerfetcher/layers.go
countvonzero: from the current impl in `layerBlocksReqReceiver`, no. the code would panic before return `nil, nil`
if the peer does not have data for this layer, it will return `nil, ErrInternal` (with the actual error being database.ErrNotFOund)

layerfetcher/layers.go
countvonzero: actually BlockID is 20 bytes and not 32 bytes.

antonlerner (approved):

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

dshulyak:
bors try

antonlerner:
bors try

countvonzero:
bors merge

Issue #2699, opened 2021-08-20
ATXs are delivered after target epoch has started


dshulyak:
## Description

Multiple miners are receiving ATXs after target epoch has already started. It is especially important for tortoise beacon/weak coin, and probably in other areas as well.

## Steps To Reproduce
Late nodes test can reproduce it consistently. It is not only about miners that are added to the network later.

[link to the test](https://kibana-tests.spacemesh.io/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2021-08-18T13:30:00.000Z',to:'2021-08-18T16:30:50.779Z'))&_a=(columns:!(atx_id,atx_node_id,kubernetes.pod_name),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:sm,key:kubernetes.namespace_name,negate:!f,params:(query:kimmy-develop-03),type:phrase),query:(match_phrase:(kubernetes.namespace_name:kimmy-develop-03))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:sm,key:M,negate:!f,params:(query:'processing%20atx'),type:phrase),query:(match_phrase:(M:'processing%20atx'))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:sm,key:epoch_id,negate:!f,params:(query:'4'),type:phrase),query:(match_phrase:(epoch_id:'4')))),index:sm,interval:auto,query:(language:kuery,query:''),sort:!(!('@timestamp',desc))))

Everything after epoch 5 has started can be considered late `Aug 18, 2021 @ 19:06:41.012`.

## Expected Behavior
ATXs are delivered before target epoch has started.

dshulyak:
[...]

PR #2715, opened 2021-08-24
[Merged by Bors] - Execute tortoise beacon round only if node is synced at the start of the epoch

dshulyak:
## Motivation

Tortoise beacon round should be skipped if node is not synced at the start of it.

Tortoise beacon checks that the miner is eligible to send any message during the round (proposal, initial vote, following) using miner weights from the previous epoch. If node doesn't have ATXs at the start of the epoch it is almost guaranteed to produce invalid beacon.  

probably related https://github.com/spacemeshos/go-spacemesh/issues/2699 . will close after verifying

## Changes
- skip tortoise beacon round if we are not synced at the start of it

## Test Plan
it addresses race condition that only appears when new node is added to the network. if this works latenodes test must be stabilized even futher.

## TODO

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -660,10 +660,10 @@ func (app *App) initServices(ctx context.Context,
 		ValidationDelta: time.Duration(app.Config.SyncValidationDelta) * time.Second,
 		AlwaysListen:    app.Config.AlwaysListen,
 	}
-	newSyncer := syncer.NewSyncer(ctx, syncerConf, clock, msh, layerFetch, app.addLogger(SyncLogger, lg))
+	syncer := syncer.NewSyncer(ctx, syncerConf, clock, msh, layerFetch, app.addLogger(SyncLogger, lg))
 	// TODO(dshulyak) this needs to be improved, but dependency graph is a bit complicated
-	tBeacon.SetSyncState(newSyncer)
-	blockOracle := blocks.NewMinerBlockOracle(layerSize, layersPerEpoch, atxDB, tBeacon, vrfSigner, nodeID, newSyncer.ListenToGossip, app.addLogger(BlockOracle, lg))
+	tBeacon.SetSyncState(syncer)
+	blockOracle := blocks.NewMinerBlockOracle(layerSize, layersPerEpoch, atxDB, tBeacon, vrfSigner, nodeID, syncer.ListenToGossip, app.addLogger(BlockOracle, lg))
 
 	// TODO: we should probably decouple the apptest and the node (and duplicate as necessary) (#1926)
 	var hOracle hare.Rolacle
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -660,9 +660,10 @@ func (app *App) initServices(ctx context.Context,
 		ValidationDelta: time.Duration(app.Config.SyncValidationDelta) * time.Second,
 		AlwaysListen:    app.Config.AlwaysListen,
 	}
-	syncer := syncer.NewSyncer(ctx, syncerConf, clock, msh, layerFetch, app.addLogger(SyncLogger, lg))
-
-	blockOracle := blocks.NewMinerBlockOracle(layerSize, layersPerEpoch, atxDB, tBeacon, vrfSigner, nodeID, syncer.ListenToGossip, app.addLogger(BlockOracle, lg))
+	newSyncer := syncer.NewSyncer(ctx, syncerConf, clock, msh, layerFetch, app.addLogger(SyncLogger, lg))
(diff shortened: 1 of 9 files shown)


dshulyak:
bors try

dshulyak:
bors try

cmd/node/node.go
```diff
@@ -661,7 +661,8 @@ func (app *App) initServices(ctx context.Context,
 		AlwaysListen:    app.Config.AlwaysListen,
 	}
 	syncer := syncer.NewSyncer(ctx, syncerConf, clock, msh, layerFetch, app.addLogger(SyncLogger, lg))
-
+	// TODO(dshulyak) this needs to be improved, but dependency graph is a bit complicated
+	tBeacon.SetSyncState(syncer)
```
nkryuchkov, line 665: Why did you decide not to add this to the list of args to `New`?

nkryuchkov (approved):

cmd/node/node.go
dshulyak: beacon needs to be created before syncer. there is a messy dependency order. but i can't resolve it without another layer of indirection

dshulyak:
bors try

tortoisebeacon/tortoise_beacon.go
```diff
@@ -156,13 +162,24 @@ type TortoiseBeacon struct {
 	proposalChans   map[types.EpochID]chan *proposalMessageWithReceiptData
 }
 
+// SetSyncState updates sync state provider. Must be executed only once.
+func (tb *TortoiseBeacon) SetSyncState(sync SyncState) {
```
countvonzero, line 166: rename `SetSyncStateProvide`r? or it sounds like you are changing the state

tortoisebeacon/tortoise_beacon.go
```diff
@@ -359,6 +376,10 @@ func (tb *TortoiseBeacon) handleEpoch(ctx context.Context, epoch types.EpochID)
 
 		return
 	}
+	if !tb.sync.IsSynced(ctx) {
```
countvonzero, line 379: or you can register a channel to be notified when the node turned synced. see `RegisterChForSynced`. 
tho this does not notify you again if the node goes out of sync, i think it serves your purpose here.

cmd/node/node.go
```diff
@@ -661,7 +661,8 @@ func (app *App) initServices(ctx context.Context,
 		AlwaysListen:    app.Config.AlwaysListen,
 	}
 	syncer := syncer.NewSyncer(ctx, syncerConf, clock, msh, layerFetch, app.addLogger(SyncLogger, lg))
-
+	// TODO(dshulyak) this needs to be improved, but dependency graph is a bit complicated
```
countvonzero, line 664: try `RegisterChForSynced` in syncer.go?

countvonzero (approved):

tortoisebeacon/tortoise_beacon.go
dshulyak: that would be better. but if node goes out of sync again i will need to re-register. and for that, i still need a pointer

dshulyak:
bors try

dshulyak:
Anyway this change makes latenodes not less flaky but more, because it misses tortoise beacon in sync consistently now. It can't be merged until we solve that issue.

antonlerner:
>Anyway this change makes latenodes not less flaky but more, because it misses tortoise beacon in sync consistently now. It can't be merged until we solve that issue.

I can propose that for the meantime TB validation will not fail block validation, @nkryuchkov has a new syncing method he's working on that should solve this issue so perhaps it's best to patch it now and wait a couple of days to get full functionality of both TB and TB sync

dshulyak:
> I can propose that for the meantime TB validation will not fail block validation, @nkryuchkov has a new syncing method he's working on that should solve this issue so perhaps it's best to patch it now and wait a couple of days to get full functionality of both TB and TB sync

sounds good to me. @nkryuchkov do you see any issues with the proposal?

nkryuchkov:
@dshulyak no, I don't see any. I did that yesterday in https://github.com/nkryuchkov/go-spacemesh/tree/execute-beacon-only-synced-disable-validation, it increased the values in the test output closer to expected ones, however, there're some still issues in hare (namespace name: `pw37j`). I tried to disable and mock the tortoise beacon, the test passed, so the issue is in it. I'm fixing hare now

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge-

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #2721, opened 2021-08-26
[Merged by Bors] - api: don't query mempool transactions from mesh

### LINKED ISSUE #2714: AccountMeshDataQuery api internal error
## Description
MeshService. AccountMeshDataQuery returns an internal error to clients in some cases.

## Steps To Reproduce
1. Submit a transaction to the mempool from a client via the api to a synced node.
2. Call MeshService.AccountMeshDateQuery() with the flag set to get transactions for the sender's account.

## Behavior 
API response is an internal server error.

Now wait until the transaction gets on the mesh and repeat the repro case. 
Result: no exception and api method returns the new transaction as expected.

## Expected Behavior
API method to not throw an exception in this case.

## Environment
- Full node from https://github.com/spacemeshos/go-spacemesh/tree/healing-state-refactor 
- devnet id 203

```
$ account send-coin
Transfer coins from local account to another account.
$ Enter destination address: 0xdd0c5c77cd8af2745a94aed9944627a4ab3729f3
$ Enter amount to transfer in Smidge: 666
$ Use default transaction fee of 1 Smidge? (y/n) y
New transaction summary:
From:   0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To:     0xdD0C5C77cd8aF2745A94AeD9944627a4AB3729f3
Amount: 666 Smidge
Fee:    1 Smidge
Nonce:  2
$ Confirm transaction (y/n): y
Transaction submitted.
Transaction id: 0xba5e0576bc00d85957945d6687d1a4a433e80ea83fe6165c4b119ca22cf5cef5
Transaction state: Submitted to the network
$ account txs
<APP> ERRO 003 15:56:23.172 repl.printAccountMeshTransactions ▶ failed to print transactions: rpc error: code = Internal desc = error retrieving tx data
$ status node
Node info:
Version: v0.0.0-unreleased
Build: e1e9b1e724d167bb3c6dc811c13692f6fd2a6170
API server: localhost:9092 (GRPC API 1.1). >> Insecure Connection. Use only with a local trusted server <<
Synced: true
Synced layer: 487
Current layer: 487
Verified layer: 484
Peers: 8
$ status net
Network info:
Network id: 203
Max transactions per second: 15
Layers per epoch: 6
Layer duration: 200 seconds
Current layer: 487
Current epoch: 81
Genesis time: 2021-08-23 13:00:03 +0300 IDT
$ account txs
Total mesh transactions: 3
Transaction summary:
Id: 0x67242f56fb175954be2e2040ed0b5aed98bc10316dc14617d5a263b1b070f763
From: 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To (coin account): 0x0000000000000000000000000000000000000001
Nonce: 0
Amount: 1234 Smidge
Fee: 100 Smidge
Layer number: number:466
-----
-----
Transaction summary:
Id: 0xcba381ae43c506ecccde194b0a812a2b88afc41fb6e0b9d42ccf291fb8796644
From: 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To (coin account): 0xdD0C5C77cd8aF2745A94AeD9944627a4AB3729f3
Nonce: 1
Amount: 123456789 Smidge
Fee: 100 Smidge
Layer number: number:467
-----
-----
Transaction summary:
Id: 0xba5e0576bc00d85957945d6687d1a4a433e80ea83fe6165c4b119ca22cf5cef5
From: 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To (coin account): 0xdD0C5C77cd8aF2745A94AeD9944627a4AB3729f3
Nonce: 2
Amount: 666 Smidge
Fee: 100 Smidge
Layer number: number:486
-----
```

dshulyak:
## Motivation
closes: https://github.com/spacemeshos/go-spacemesh/issues/2714
## Changes
- return mempool tx from from mempool, instead of querying mesh for them

## Test Plan
TODO

## TODO

## DevOps Notes


dshulyak:
@avive note that with this change mempool transactions will be without layer (see the last one), e.g.:

```
$ account txs
Total mesh transactions: 4
Transaction summary:
Id: 0x67242f56fb175954be2e2040ed0b5aed98bc10316dc14617d5a263b1b070f763
From: 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To (coin account): 0x0000000000000000000000000000000000000001
Nonce: 0
Amount: 1234 Smidge
Fee: 100 Smidge
Layer number: number:466
-----
-----
Transaction summary:
Id: 0xcba381ae43c506ecccde194b0a812a2b88afc41fb6e0b9d42ccf291fb8796644
From: 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To (coin account): 0xdD0C5C77cd8aF2745A94AeD9944627a4AB3729f3
Nonce: 1
Amount: 123456789 Smidge
Fee: 100 Smidge
Layer number: number:467
-----
-----
Transaction summary:
Id: 0xba5e0576bc00d85957945d6687d1a4a433e80ea83fe6165c4b119ca22cf5cef5
From: 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To (coin account): 0xdD0C5C77cd8aF2745A94AeD9944627a4AB3729f3
Nonce: 2
Amount: 666 Smidge
Fee: 100 Smidge
Layer number: number:486
-----
-----
Transaction summary:
Id: 0x7e1c4be0337db7e867f5bbc87bb0f7c2ab0ed498d9fbe2f2e5063f1a46aec00d
From: 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
To (coin account): 0xfF083C9A22E05D3459B03f3DbeD61c7AD6E0a209
Nonce: 3
Amount: 101 Smidge
Fee: 100 Smidge
Layer number: 
-----

```

avive:
> @avive note that with this change mempool transactions will be without layer (see the last one), 

I think that in order to keep API correctness and high-level design, mempool transactions should only be returned by TransactionService.TransactionsState and TransactionService.TransactionsStateStream and that the MeshService should only return transactions that are on the mesh (with a layer).

@lrettig  - thoughts?

dshulyak:
I think we can just fix an existing bug, and then do redesign separately. For instance i don't see any request that returns all transactions for an account in TransactionService.

countvonzero:
@avive what does it means to return a layer for a mempool tx? projected layer to be mined in the mesh?

from the UX point of view, i'd imagine the smapp should combine TXs mined (in the mesh) and pending TXs (in the mempool) and present it to the user?

avive:
> I think we can just fix an existing bug, and then do redesign separately. For instance i don't see any request that returns all transactions for an account in TransactionService.

@dshulyak @countvonzero here is a product note I've put together on how clients need to use the api to display transactions to users as a guide for what needs to be implemented for Smapp 0.2 - it is not very intuitive. What happened is that we didn't get around to implement the full design which requires transactions receipts via the GlobalStateService for 0.2 so we added late the layer id to the transactions returned by MeshService as this is the only way for clients to have a timestamp for transactions which is obviously important for consistent display to users....

https://github.com/spacemeshos/product/blob/master/api_usage_patterns.md

So, stateful clients like smapp track the ids of txs they submitted to the mempool and query the TransactionService for the status of these. MeshService should only return txs for an account for transactions which are on the mesh. Meaning they have a layer. GlobalService is only supposed to return receipts for transactions that got executed....

Ideally, MeshService should not return Transactions which are not on the mesh yet but it is not a huge deal if it includes mem-pool transactions although it kind of makes the TransactionService queries redundant...

lrettig:
Agree with @avive here: from an API design perspective, MeshService should not return transactions that haven't been added to the mesh yet. TransactionService can do that. We try to keep a strict conceptual separation among the different API services. It's a bit counterintuitive but I see value in the separation, and Aviv did a good job of documenting the API usage pattern in the doc he linked above.

lrettig:
@dshulyak what's the intended downstream use case for this?

dshulyak:
i don't know what was the intended use case. code was clearly trying to return pool transactions, but was doing it incorrectly. so i  assumed that it was by design. if by design stateful client is supposed to poll for each tx separately i am fine with that.

antonlerner (approved):

api/grpcserver/mesh_service.go
```diff
@@ -106,20 +106,24 @@ func (s MeshService) MaxTransactionsPerSecond(context.Context, *pb.MaxTransactio
 
 // QUERIES
 
-func (s MeshService) getFilteredTransactions(startLayer types.LayerID, addr types.Address) (txs []*types.MeshTransaction, err error) {
+func (s MeshService) getFilteredTransactions(startLayer types.LayerID, addr types.Address) ([]*types.MeshTransaction, error) {
 	meshTxIds := s.getTxIdsFromMesh(startLayer, addr)
-	mempoolTxIds := s.Mempool.GetTxIdsByAddress(addr)
+	mempoolTxs := s.Mempool.GetTxsByAddress(addr)
```
antonlerner: according to the conversation on this PR it seems like this should be removed entirely

dshulyak:
bors merge

PR #2722, opened 2021-08-26
[Merged by Bors] - api: start smeshing using background context

### LINKED ISSUE #2712: No smeshing rewards
## Steps To Reproduce
1. Sync a full node
2. Setup post via api
3. Wait ~5 epochs 
4. Check rewards via api - using smesher id or via the rewards address

## Behavior
Post status and smesher status are as expected but no rewards returned by the api 

```
$ post status
👍  Proof of spacetime data was created and is used for smeshing.
Proof of spacetime configuration:
  Data dir (relative to node or absolute): .
  Date files: 1
  Compute provider id: 1
  Throttle when computer is busy: false
  Bits per label: 8
  Units: 2
  Labels: 2048
  Size (GiB): 1.9073486e-06
  Size (Bytes): 2048
$ smesher status
Node is currently smeshing
$ smesher rewards
Smesher id: 0x02b34140cd1d6822afc4015deb69818365590b9815a68528889e8c0229ee319a
Total rewards: 0
$ smesher status
Node is currently smeshing
```

## Expected Behavior
Get smeshing rewards for smesher

## Environment
- macOS/amd64
- node branch healing-state-refactor

dshulyak:
## Motivation

GRPC context should not be used for tasks that are meant to be running after grpc method returns.

closes: https://github.com/spacemeshos/go-spacemesh/issues/2712

## Changes
- use background context for StartSmeshing

## TODO

## DevOps Notes


lrettig:
Rather than background context, I think we should pass a context object into `NewSmesherService`, store this on the object, and pass it down into `smeshingProvider.StartSmeshing` as a cancel context. We should probably do this with all of the GRPC services.

dshulyak:
> Rather than background context, I think we should pass a context object into NewSmesherService, store this on the object, and pass it down into smeshingProvider.StartSmeshing as a cancel context. We should probably do this with all of the GRPC services.

in my opinion this is not necessary. services are stopabble (implement Stop or Close method) and usually, this must be executed in order for services to exit gracefully. So what is the purpose of the passing context to the object in this case?

dshulyak:
I will use cmd context. I already had function options for builder, so it is easy to make it optional :+1:

activation/activation.go
```diff
@@ -177,11 +177,19 @@ func WithPoETClientInitializer(initializer PoETClientInitializer) BuilderOption
...
 
+// WithContext modifies parent context for background job.
+func WithContext(ctx context.Context) BuilderOption {
+	return func(b *Builder) {
+		b.runCtx = ctx
+	}
+}
+
 // NewBuilder returns an atx builder that will start a routine that will attempt to create an atx upon each new layer.
 func NewBuilder(conf Config, nodeID types.NodeID, signer signer, db atxDBProvider, net broadcaster, mesh meshProvider,
```
lrettig, line 188: `layersPerEpoch` is not used here (it uses `conf.LayersPerEpoch`), can we fix this?

lrettig (approved):

activation/activation.go
dshulyak: please open an issue if you think that it is important. i would rather remove global state tbh and pass LayerPerEpoch as config everywhere

dshulyak:
bors merge

Issue #2657, opened 2021-08-11
Refactor log.WithContext to minimize memory allocations


dshulyak:
## Description

15% of all allocations in space and in objects is made by copying logger inside log.WithContext. This pattern is misused, if logger needs to be extended with some fields it must be copied once for the whole scope as in:

```go
clog := log.WithContext(ctx)
clog.Info(...)
clog.Debug(...)
```

And not

```go
clog.WithContext(ctx).Info(...)
clog.WithContext(ctx).Debug(...)
```

This will minimize amount of allocations but still not completely, because ctx is passed through the code and logger will have to be initialized in every separate method. We  need to find a way to avoid allocation but preserve ease of use. One way could is to use interface like:

```go
log.Info(..., log.Context(ctx))
```

log.Context will unpack session (or any other know fields) from context and print them without copying the logger. 

## Affected code

log.WithContext and every place it is used.

dshulyak:
cc @lrettig

dshulyak:
If we will use the approach with log.Context we can get log like:

```
logger.go:130: 2021-08-11T10:20:02.075+0300	INFO	test	{"context": {"session": "yyy", "request": "xxx"}, "name": ""}
```

I think this is good enough, the only question if elastic will be able to parse it correctly (https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html)

lrettig:
Good idea. That works for me, assuming elastic can parse it!

moshababo:
@dshulyak still relevant?

PR #2723, opened 2021-08-26
[Merged by Bors] - only calculate layer aggregated hash after block validity is determined

### LINKED ISSUE #2695: devnet: log "could not get contextual validity for block in list"
## Description

aggregated hash is calculated as part of setting processed layer, because it needs to be calculated in layer order.
as part of mesh.setProcessedLayer(layerN), it attempts to calculate the aggregated hash for layerN.

this is premature because when tortoise validates a layer N, it can only determine block validity for layer N-1. this is because to determine validity for a layer N you need votes from layers >N. hence this log.

## Steps To Reproduce
1. `make dockertest-late-nodes namespace=$YOUR_NAMESPACE dump=True`
`could not get contextual validity for block in list`

## Expected Behavior
`mesh.setProcessedLayer` should only calculate aggregated hash for layer whose block validity is determined.

countvonzero:
## Motivation

Closes #2695

## Changes

- remove processed layer hash. this was used to as a cache of aggregated hash for the latest layer. now with GetAggregatedHash(layerID), this field is no longer needed
- calculate layer hash and aggregated hash after tortoise validate a layer

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -360,49 +360,71 @@ func (vl *validator) ValidateLayer(ctx context.Context, lyr *types.Layer) {
 }
 
 func (msh *Mesh) persistLayerHashes(ctx context.Context, from, to types.LayerID) error {
+	logger := msh.WithContext(ctx)
 	if to.Before(from) {
-		msh.WithContext(ctx).With().Panic("verified layer went backward",
+		logger.With().Panic("verified layer went backward",
 			log.FieldNamed("fromLayer", from),
 			log.FieldNamed("toLayer", to))
 	}
 
-	msh.WithContext(ctx).With().Debug("persisting layer hashes",
+	logger.With().Debug("persisting layer hashes",
 		log.FieldNamed("from_layer", from),
 		log.FieldNamed("to_layer", to))
 	for i := from; !i.After(to); i = i.Add(1) {
-		lyr, err := msh.GetLayer(i)
+		validBlockIDs, err := msh.getValidBlockIDs(ctx, i)
 		if err != nil {
-			msh.WithContext(ctx).With().Error("failed to get verified layer", i, log.Err(err))
+			logger.With().Error("failed to get valid block IDs", i, log.Err(err))
 			return err
 		}
-		validBlocks, _ := msh.BlocksByValidity(lyr.Blocks())
-		sorted := types.SortBlockIDs(types.BlockIDs(validBlocks))
 		hash := types.EmptyLayerHash
-		if len(sorted) > 0 {
-			hash = types.CalcBlocksHash32(sorted, nil)
+		if len(validBlockIDs) > 0 {
+			hash = types.CalcBlocksHash32(validBlockIDs, nil)
 		}
 		if err := msh.persistLayerHash(i, hash); err != nil {
-			msh.WithContext(ctx).With().Error("failed to persist layer hash", i, log.Err(err))
+			logger.With().Error("failed to persist layer hash", i, log.Err(err))
 			return err
 		}
 
 		prevHash, err := msh.getAggregatedLayerHash(i.Sub(1))
 		if err != nil {
-			msh.WithContext(ctx).With().Debug("failed to get previous aggregated hash", i, log.Err(err))
+			logger.With().Debug("failed to get previous aggregated hash", i, log.Err(err))
 			return err
 		}
-		msh.WithContext(ctx).With().Debug("got previous aggregatedHash", i, log.String("prevAggHash", prevHash.ShortString()))
-		newAggHash := types.CalcBlocksHash32(sorted, prevHash.Bytes())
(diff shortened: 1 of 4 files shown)


mesh/mesh.go
```diff
@@ -117,13 +113,18 @@ func NewMesh(db *DB, atxDb AtxDB, rewardConfig Config, trtl tortoise, txPool txM
...
 		latestLayer:         types.GetEffectiveGenesis(),
 		latestLayerInState:  types.GetEffectiveGenesis(),
 	}
 
 	msh.Validator = &validator{Mesh: msh}
+	gLyr := types.GetEffectiveGenesis()
+	for i := types.NewLayerID(1); i.Before(gLyr); i = i.Add(1) {
+		msh.SetZeroBlockLayer(i)
+	}
+	msh.persistLayerHashes(context.Background(), types.NewLayerID(1), gLyr)
```
dshulyak: strange that it doesn't return error

mesh/mesh.go
```diff
@@ -379,6 +336,9 @@ func (vl *validator) ValidateLayer(ctx context.Context, lyr *types.Layer) {
 		logger.With().Error("could not persist tortoise", log.Err(err))
 	}
 	vl.pushLayersToState(ctx, oldPbase, newPbase)
+	if newPbase.After(oldPbase) {
+		vl.persistLayerHashes(ctx, oldPbase.Add(1), newPbase)
```
dshulyak: no error checking?

mesh/mesh.go
```diff
@@ -389,6 +349,46 @@ func (vl *validator) ValidateLayer(ctx context.Context, lyr *types.Layer) {
...
+		log.FieldNamed("from_layer", from),
+		log.FieldNamed("to_layer", to))
+	for i := from; !i.After(to); i = i.Add(1) {
+		lyr, err := msh.GetLayer(i)
+		if err != nil {
+			msh.WithContext(ctx).With().Error("failed to get verified layer", i, log.Err(err))
+			return err
+		}
+		validBlocks, _ := msh.BlocksByValidity(lyr.Blocks())
+		sorted := types.SortBlockIDs(types.BlockIDs(validBlocks))
```
dshulyak: blocks are stored in the sorted order in the database and returned that way during iteration

dshulyak (approved):
- we should return those errors from persist* methods. if writes started to fail we will get an error in some other way most definitely, but i don't understand why we log the error and don't return it in many cases
-  blocks are already returned sorted for the layer, so unless something messes up the order it is possible to rely on that

mesh/mesh.go
countvonzero: how do you mean? do you expect errors here?

mesh/mesh.go
countvonzero: done

mesh/mesh.go
countvonzero: i hesitate. the underlying db impl may change over time. i'd prefer to keep this sort to guarantee its deterministic outcome. 
in the future we will have unified content block so i don't think performance will be the issue.

countvonzero:
i ran a system late-node test and confirmed that these logs disappeared @dshulyak

mesh/mesh.go
dshulyak: we are writing to the database, if smth is wrong with disk errors are guaranteed, no?

mesh/mesh.go
dshulyak: i suppose that it can be ensured with unit tests. db usually good at storing things in sorted order, so i won't worry much that it will change, but up to you.

mesh/mesh.go
```diff
@@ -389,6 +359,52 @@ func (vl *validator) ValidateLayer(ctx context.Context, lyr *types.Layer) {
...
+	if to.Before(from) {
+		msh.WithContext(ctx).With().Panic("verified layer went backward",
+			log.FieldNamed("fromLayer", from),
+			log.FieldNamed("toLayer", to))
+	}
+
+	msh.WithContext(ctx).With().Debug("persisting layer hashes",
+		log.FieldNamed("from_layer", from),
+		log.FieldNamed("to_layer", to))
+	for i := from; !i.After(to); i = i.Add(1) {
```
dshulyak, line 373: i suggest to initialize logger once for this method and use it, there is a certain overhead associated with log.WithContext. see https://github.com/spacemeshos/go-spacemesh/issues/2657

mesh/mesh.go
countvonzero: yes. ok. you mean generic error handling. i thought you saw something specific. thanks

mesh/mesh.go
countvonzero: adding a test is a good idea. done.

mesh/mesh.go
countvonzero: done

countvonzero:
bors merge

Issue #2679, opened 2021-08-16
Optimize tortoise block opinion data structure and vote counting algorithm


lrettig:
## Description
This is done very naively today. It could probably be made much more efficient with better data structures: e.g., using matrix math.

Related issues that this could address: #2369, #2424, #2522, #2672, #2719

Things we need the block opinions data structure to do that it cannot currently:
- efficiently evict opinions on blocks older than a given layer (#2719)
- index into a block's opinions on blocks in a given layer (#2424)
- efficiently perform basic linear algebra/matrix math, e.g., sum all opinions of blocks in a given layer for a given block, or copy all of a block's opinions into another block's opinions vector

## Affected code
tortoise/verifying_tortoise.go

This issue appears in commit hash: 9d3a379994e3652eb0378d2cc555588bba91005e

dshulyak:
@lrettig do i understand it right that 1 and 2 points will be addressed by https://github.com/spacemeshos/go-spacemesh/pull/2734 ?

as for

> efficiently perform basic linear algebra/matrix math, e.g., sum all opinions of blocks in a given layer for a given block, or copy all of a block's opinions into another block's opinions vector

do you have particular ideas on how it can be optimized now? or should we wait for UCB before attempting optimizations?

dshulyak:
It looks like we are already close to a unified content block. The remaining task from this issue to simplify opinions data structure, we can tackle it as a part of https://github.com/spacemeshos/go-spacemesh/issues/2719 .

PR #2734, opened 2021-08-29
Add layerID to tortoise opinion map

### LINKED ISSUE #2424: Add exception votes against blocks the base block votes for
## Description

## Affected code

### LINKED ISSUE #2719: Tortoise vote encoding data structure
Right now tortoise stores/encodes votes in a matrix of size `windowSize**2 * blocksPerLayer**2`: each block includes votes from blocks going back `windowSize` layers, each of which includes votes going back `windowSize` layers.

It seems that this data structure is too big. See @dshulyak comment here: https://github.com/spacemeshos/go-spacemesh/pull/2683#issuecomment-904322215.

Per @tal-m,

> The data structure I had in mind is a 2D matrix, where each row represents a voting block and each column a content block. Each element of the matrix is the opinion of a voting block about the corresponding content block. The window size comes into play because we don't need to store the detailed opinions about content blocks that are more than `windowSize` layers in the past (for those, we only store our own opinion, which should be identical to the opinion of all the voting blocks we see)

> Regarding optimizations:
> I think the unified content block will be a huge improvement to complexity --- since there is almost always 1 block per layer, the matrix needs to be windowSize*votesPerLayer (votesPerLayer=blocksPerLayer in the previous scheme). So that's a blocksPerLayer factor improvement; if we're currently using 50 blocks per layer, it means that even without any additional optimization, the vote storage should be less than 140MB
> I'm sure there's a lot of optimization we can do in the data structure itself. Are we storing it in bit-packed fields? Are we making use of the fact that most honest votes actually have the same opinion? Etc.

See [this slack thread](https://spacemesh.slack.com/archives/CG91F13C7/p1629850014078200)

lrettig:
## Motivation
Closes #2719 
Closes #2424
See also #2679

## Changes
Adds layerID to the inner tortoise block opinion map

This allows efficient eviction of opinions of blocks in older layers,
and it allows us to efficiently add votes against blocks the base block
votes for that are missing from the input vector for a layer

## Test Plan
TBD

## TODO

## DevOps Notes


dshulyak:
it doesn't seem to address the problem in https://github.com/spacemeshos/go-spacemesh/issues/2719, but up to you

lrettig:
> it doesn't seem to address the problem in #2719, but up to you

This is the part that addresses it:

https://github.com/spacemeshos/go-spacemesh/blob/10fee3a5ecae31325e0dd6912e5f0a9ac1a2672d/tortoise/verifying_tortoise.go#L215-L220

dshulyak:
@lrettig what is left to finish in this change? it would be nice to lend it, cause of the memory consumption and it also addresses some of the critical issues from your list. if you don't have time i can finish it

lrettig:
> @lrettig what is left to finish in this change? it would be nice to lend it, cause of the memory consumption and it also addresses some of the critical issues from your list. if you don't have time i can finish it

I think this PR is mostly done. IIRC not all of the existing tortoise tests are passing. So the steps to completion are:

1. resolve merge conflicts
2. get existing tests to pass
3. add new tests - if necessary, not sure it's necessary here since this PR adds no new functionality

Issue #2706, opened 2021-08-23
Reconsider using taskgroup for managing goroutines


dshulyak:
## Description
@noamnelke pointed out that it is safe to use sync.WaitGroup (and errogroup.Group) if all goroutines are spawned before group counter reached 0 (or from within goroutines of the same group in case of errgroup) like:

```go
var wg sync.WaitGroup
wg.Add(1)
go func() {
   defer wg.Done()
   wg.Add(1)
   go func() {
       wg.Done()
   }() 
}()
wg.Wait()
```

But if we call wg.Add(1) after counter reached zero wg.Wait may panic, so the following code is not safe:

```go
var wg sync.WaitGroup
wg.Add(1)
go func() {
   defer wg.Done()
   go func() {
       wg.Add(1)
       wg.Done()
   }() 
}()
wg.Wait()
```

We need to revisit existing code to check if it possible to eliminate the latter pattern and use errgroup module.

## Affected code

taskgroup module and where it is used

PR #2744, opened 2021-08-30
[Merged by Bors] - fix flaky syncer test by using errgroup

### LINKED ISSUE #2742: flaky test: panic: Log in goroutine after TestForceSync has completed
## Description

https://github.com/spacemeshos/go-spacemesh/runs/3443623289
panic: Log in goroutine after TestForceSync has completed

this happened for syncer.TestForceSync but may happen for other tests in syncer_test.go

## Steps to reproduce
harder to produce. even
`go test -race -v ./syncer/ -run TestForceSync -count 1000` does not reproduce.
but when i put a `time.Sleep(time.Second)` after `go.startValidating()` allows to reproduce with
`go test -race -v ./syncer/ -run TestForceSync -count 100`

## Actual Behavior

no test failure

countvonzero:
## Motivation

Closes #2742

## Changes
- add taskgroup to syncer and make sure all goroutines are finished before ending tests
- stop synctimer to prevent ticked synchronize for testing
 
## Test Plan

go test -race -v ./syncer/ -count 1000

## DevOps Notes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -79,6 +79,8 @@ type Syncer struct {
 	fetcher  layerFetcher
 	syncOnce sync.Once
 	// access via atomic.[Load|Store]Uint32
+	started uint32
+	// access via atomic.[Load|Store]Uint32
 	syncState syncState
 	// access via atomic.[Load|Store]Uint32
 	isBusy    uint32
@@ -121,12 +123,14 @@ func NewSyncer(ctx context.Context, conf Configuration, ticker layerTicker, mesh
 func (s *Syncer) Close() {
 	s.syncTimer.Stop()
 	s.cancelFunc()
-	s.logger.Info("waiting for goroutines to finish")
-	if err := s.tg.Wait(); err != nil {
-		s.logger.With().Warning("taskgroup: Wait returned an error",
-			log.Err(err))
+	if s.getStarted() {
+		s.logger.Info("waiting for goroutines to finish")
+		if err := s.tg.Wait(); err != nil {
+			s.logger.With().Warning("taskgroup: Wait returned an error",
+				log.Err(err))
+		}
+		s.logger.Info("all syncer goroutines finished")
 	}
-	s.logger.Info("all syncer goroutines finished")
 }
 
 // RegisterChForSynced registers ch for notification when the node enters synced state
@@ -161,9 +165,12 @@ func (s *Syncer) IsSynced(ctx context.Context) bool {
 func (s *Syncer) Start(ctx context.Context) {
 	s.syncOnce.Do(func() {
 		s.logger.WithContext(ctx).Info("Starting syncer loop")
-		if err := s.tg.Go(s.loop); err != nil {
+		err := s.tg.Go(s.loop)
+		if err != nil {
 			s.logger.With().Warning("taskgroup: Go returned an error",
 				log.Err(err))
+		} else {
+			s.setStarted()
 		}
 	})
 }
@@ -205,6 +212,14 @@ func (s *Syncer) isClosed() bool {
 	default:
 		return false
 	}
+}
+
+func (s *Syncer) setStarted() {
+	atomic.CompareAndSwapUint32(&s.started, 0, 1)
+}
+
+func (s *Syncer) getStarted() bool {
+	return atomic.LoadUint32(&s.started) == 1
 }
 
 func (s *Syncer) getSyncState() syncState {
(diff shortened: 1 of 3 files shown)


countvonzero:
bors try

syncer/syncer.go
```diff
@@ -113,13 +117,22 @@ func NewSyncer(ctx context.Context, conf Configuration, ticker layerTicker, mesh
...
 }
 
 // Close stops the syncing process and the goroutines syncer spawns.
 func (s *Syncer) Close() {
-	// TODO: ensure goroutines are all terminated before shutting down
+	s.syncTimer.Stop()
 	s.cancelFunc()
+	if s.getStarted() {
+		s.logger.Info("waiting for goroutines to finish")
+		if err := s.tg.Wait(); err != nil {
```
dshulyak: i think error can be ignored if it is a context.Canceled, since it means that it was stopped properly

dshulyak (approved):

syncer/syncer.go
```diff
@@ -80,6 +81,8 @@ type Syncer struct {
 	fetcher  layerFetcher
 	syncOnce sync.Once
 	// access via atomic.[Load|Store]Uint32
+	started uint32
```
dshulyak: is it used only to print a log line in case of stop?

syncer/syncer.go
```diff
@@ -153,26 +166,45 @@ func (s *Syncer) IsSynced(ctx context.Context) bool {
...
+		}
+	}
+}
+
 // ForceSync manually starts a sync process outside the main sync loop. If the node is already running a sync process,
 // ForceSync will be ignored.
 func (s *Syncer) ForceSync(ctx context.Context) {
 	s.logger.WithContext(ctx).Debug("executing ForceSync")
-	go s.synchronize(ctx)
+	err := s.tg.Go(func(ctx context.Context) error {
```
dshulyak: i think if it is possible to force a sync using a channel operation (instead of goroutine) then it will be safe to use errgroup in this case (https://github.com/spacemeshos/go-spacemesh/issues/2706)

syncer/syncer.go
countvonzero: what happens is that `TestSpacemeshApp_NodeService` hung because it's trying to call syncer.Close() before syncer.Start() was executed. so syncer.Close() blocks on taskgroup.waitErr. 

do you think it make sense to add another way to terminate (explicitly) rather than waiting for one of its goroutines return error?

syncer/syncer.go
countvonzero: what do you mean by "using a channel operation"? currently it's called directly in cmd/sync/sync.go for testing purposes.

dshulyak: i mean in the main loop we can add another channel, and when token is received from that channel goroutine to sync will be spawned. and in ForceSync we will send that token

syncer/syncer.go
dshulyak: oh thats another use case that i didn't anticipate. i think we can initialize waitErr only when first Go is called, and return Wait immediately if waitErr is nil

syncer/syncer.go
countvonzero: i see. thanks. i recall there was a ForceSync RPC before, but i can't find it anymore.  
do you actually prefer me using errgroup? i had assume that taskgroup is created for similar purpose as errgroup.

dshulyak: i think there is a benefit in using errgroup, since it is simpler and doesn't require to return error from errgroup.Go.

countvonzero: yes. much simpler. no need to create an error for errgroup.Wait() to return.
@dshulyak  please take another look. thanks

syncer/syncer.go
```diff
@@ -153,26 +161,33 @@ func (s *Syncer) IsSynced(ctx context.Context) bool {
...
+		})
 	})
 }
 
 // ForceSync manually starts a sync process outside the main sync loop. If the node is already running a sync process,
 // ForceSync will be ignored.
 func (s *Syncer) ForceSync(ctx context.Context) {
 	s.logger.WithContext(ctx).Debug("executing ForceSync")
-	go s.synchronize(ctx)
+	s.forceSyncCh <- struct{}{}
```
dshulyak, line 196: this needs to either select on shutdown ctx or drop a token if one already exists, otherwise it may hang if loop exited

syncer/syncer.go
dshulyak: it is great that we can use it. i am sort of unhappy with that taskgroup module, i expected it to be a bit simpler and easier to use. going to ditch it if errgroup can be used everywhere

syncer/syncer.go
countvonzero: done. thanks. added test too

countvonzero:
bors merge

PR #2762, opened 2021-09-02
[Merged by Bors] - Include GPU Post and Smrepl in Build

### LINKED ISSUE #2752: CI Release GPU-POST Bins

web3rover:
## Motivation
When tag is created we only build the go-spacemesh binary and don't include gpu-post files
Closes #2752 

## Changes
Now it creates a zip file with smrepl, go-spacemesh and gpu-post files

## DevOps Notes


.github/workflows/release.yml
```diff
@@ -27,6 +27,18 @@ jobs:
         uses: crazy-max/ghaction-chocolatey@v1
         with:
           args: install make
+
+      - if: matrix.os == 'windows-latest'
```
lrettig: I think you can combine these three `if` statements into one

.github/workflows/release.yml
```diff
@@ -51,25 +63,35 @@ jobs:
         run: |
           make install
           make build VERSION=${{ steps.version.outputs.content }} BIN_DIR_WIN=./build
+      - name: Get gpu-post release
+        shell: bash
+        run: |
+          mkdir build_${{ runner.os }}
+          unzip libgpu-setup-*.zip -d build_${{ runner.os }}
```
lrettig, line 58: Where does libgpu-setup come from? Where is it downloaded?

.github/workflows/release.yml
```diff
@@ -51,25 +63,35 @@ jobs:
...
+
+          #Download latest smrepl release
+          wget https://github.com/spacemeshos/smrepl/releases/latest/download/${{ runner.os }}.zip
```
lrettig: I think we should pin the version here rather than using `latest`, it's safer

.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
   push:
     tags:
       - '*'
-
+
```
lrettig: Remove extra spaces

lrettig (changes requested):
Would be good for @noamnelke to take a look as well. Thanks!

.github/workflows/release.yml
web3rover: Fixed.

.github/workflows/release.yml
web3rover: Fixed

.github/workflows/release.yml
web3rover: `make build` downloads it into the go-spacemesh folder.

.github/workflows/release.yml
web3rover: Fixed

.github/workflows/release.yml
avive: Q: What is the best way to test it? Can it be done on a dev-branch? What is the input and what is the expected output?

avive:
Q: What is the best way to test it? Can it be done on a dev-branch? What is the input and what is the expected output? Do you need to run 'make' or do something else in-order to generate release binaries? @moshababo 

Can someone plz trigger this action so we can have a look at the artifacts generated by it?

.github/workflows/release.yml
```diff
@@ -83,18 +93,19 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag_name: ${{ github.ref }}
+          tag_name:  ${{ github.ref }}
           release_name: Release ${{ steps.version.outputs.content }}
           body: |
-            ## Compiled Binaries
-            - Windows: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/windows-latest/go-spacemesh.exe
-            - Linux: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/ubuntu-latest/go-spacemesh
-            - Mac: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/macos-latest/go-spacemesh
```
avive, line 98: It will be nice to change this to `macOS` from `Mac` and order releases by platform popularity: Windows, macOS, Linux .

avive (changes requested):

narayanprusty:
> Q: What is the best way to test it? Can it be done on a dev-branch? What is the input and what is the expected output? Do you need to run 'make' or do something else in-order to generate release binaries? @moshababo
> 
> Can someone plz trigger this action so we can have a look at the artifacts generated by it?

web3rover:
@avive You can create a tag from this branch to see the output. No input is needed. It will create a draft release with the artifacts

avive:
> @avive You can create a tag from this branch to see the output. No input is needed. It will create a draft release with the artifacts

I will do - thanks for the info. btw - i see some errors in the workflow that runs in the pr but I will ignore and do what you recommend.

.github/workflows/release.yml
web3rover: Done

avive (approved):
I wanted 'Mac' to change to 'macOS' but I guess this is fine because this is drafting a release that someone from the team is going to add anyhow so it is not a blocker to merge this. I also tested that the expected artifacts are generated for each platform.

lrettig (approved):

web3rover:
bors merge

PR #2763, opened 2021-09-02
[Merged by Bors] - Decrease GRPC startup time

### LINKED ISSUE #2533: Improve grpc services startup time
## Motivation
Make users wait less time until Smapp 0.2 is interactive. Enable users to execute commands that involve APIs faster and display up-to-date information from the api such as transactions and their statuses faster after app session start.
Currently, it can take up to 30-45 seconds until the api is ready from node startup time.

This is the user headache that this issue causes:

Screen Shot 2021-08-31 at 10 16 10 PM
 

## The Enhancement
Start the node's GRPC service as early as possible in the node startup flow and don't block it on unrelated node startup tasks such as UPNP config.

## Notes
Even if we make Smapp fully interactive before the api is ready (likely not going to happen in release 0.2 due to more basic features being worked on for launch) then all commands that use the api such as submit transaction will need to be disabled until the api services are ready so this node change will improve the UX in a significant way.

@moshababo

nkryuchkov:
## Motivation

Closes #2533

## Changes

- Start API services before P2P initialization

## Test Plan

Tested with unit tests, system tests, manual tests using `smrepl`

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -1230,12 +1230,13 @@ func (app *App) Start() error {
 		return fmt.Errorf("error starting services: %w", err)
 	}
 
+	app.startAPIServices(ctx, app.P2P)
+
 	// P2P must start last to not block when sending messages to protocols
 	if err := app.P2P.Start(ctx); err != nil {
 		return fmt.Errorf("error starting p2p services: %w", err)
 	}
 
-	app.startAPIServices(ctx, app.P2P)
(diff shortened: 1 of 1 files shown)


nkryuchkov:
bors try

lrettig (approved):
that was easy :)

nkryuchkov:
bors merge

PR #2765, opened 2021-09-03
[Merged by Bors] - activation: don't reuse context for different Start/StopSmeshing cycles

dshulyak:
## Motivation

After StopSmeshing was called atleast once we were never able to start Smeshing again.
Reported by @avive 

## Changes
- refactor Start/StopSmeshing flow to support multiple Start/Stop Smeshing cycles

## TODO

## DevOps Notes


nkryuchkov (approved):

activation/activation.go
```diff
@@ -272,8 +259,12 @@ func (b *Builder) StopSmeshing(deleteFiles bool) error {
 	}
 
 	b.stop()
-	b.status = smeshingStatusIdle
-
+	select {
+	case <-b.exited:
+		b.exited = nil
+	case <-ctx.Done():
```
countvonzero: this is a bit weird that we had to check for ctx.Done() here, since this context can be any random one?
if we don't wait for b.exited to be closed here, the next call to StartSmeshing() will not be successful.

activation/activation.go
```diff
@@ -205,65 +197,60 @@ func NewBuilder(conf Config, nodeID types.NodeID, signer signer, db atxDBProvide
...
 	return b
 }
 
 // Smeshing returns true iff atx builder started.
 func (b *Builder) Smeshing() bool {
-	b.mtx.Lock()
-	defer b.mtx.Unlock()
-
-	return b.status == smeshingStatusStarted
+	return b.started.Load()
```
countvonzero, line 209: could this be `return b.exit != nil`

activation/activation.go
dshulyak: thanks, this is clearly a bug

activation/activation.go
dshulyak: was hard to recall why i did it :)

StopSmeshing wasn't blocking previously, so i wanted to guarantee that IsSmeshing doesn't get stuck in a weird way if it will be called concurrently with StopSmeshing.

other reason was to set it to True only after postSetupProvider.StartSession was successful

dshulyak:
bors merge

dshulyak:
bors merge

activation/activation.go
countvonzero: so it's possible that `Smeshing()` returns false, client thought it's not starting and call `StartSmeshing()` again?
why not just hold the lock until after `StartSession` success? is that an expensive call?

dshulyak: i think this is when post file is created. in general i agree that it seems not very clear from ux, maybe it makes sense to change it so that smeshing is true even before file is created and false only on error or if stopped.

countvonzero: i agree. i think it make more sense for smeshing to be true before the post file is created, and set it back to false if post file failed to create. when the post file fails to create, it should behave like StopSmeshing() has been called.

dshulyak:
p2p failed with k8s error during setup.

bors merge

PR #2769, opened 2021-09-04
add test to document the expected behavior when hare output layer out of order

countvonzero:
## Motivation

add a test case that documents tortoise expected behavior when hare output layers out-of-order

## Changes
add test case when `HandleValidateLayer` is called out of layer order

## Test Plan

uint tests

## DevOps Notes

### DIFF
--- a/tortoise/verifying_tortoise_test.go
+++ b/tortoise/verifying_tortoise_test.go
@@ -1261,10 +1261,11 @@ func TestCalculateExceptions(t *testing.T) {
 	// now advance the processed layer
 	alg.trtl.Last = l1ID
 
-	// missing layer data
+	// missing layer data:
+	// expect FOR only for l0, and neutral is 0 since there are no blocks in l1 to add an exception for
 	votes, err = alg.trtl.calculateExceptions(context.TODO(), l1ID, Opinion{})
-	r.Equal(database.ErrNotFound, err)
-	r.Nil(votes)
+	r.NoError(err)
+	expectVotes(votes, 0, 1, 0)
 
 	// layer opinion vector is nil (abstains): recent layer, in mesh, no input vector
 	alg.trtl.Last = l0ID
--- a/tortoise/verifying_tortoise_test.go
+++ b/tortoise/verifying_tortoise_test.go
@@ -8,16 +8,15 @@ import (
 	"testing"
 	"time"
 
-	"github.com/spacemeshos/go-spacemesh/log"
-	"github.com/spacemeshos/go-spacemesh/rand"
-	"github.com/spacemeshos/go-spacemesh/timesync"
-
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/config"
 	"github.com/spacemeshos/go-spacemesh/database"
+	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/log/logtest"
 	"github.com/spacemeshos/go-spacemesh/mesh"
+	"github.com/spacemeshos/go-spacemesh/rand"
 	"github.com/spacemeshos/go-spacemesh/signing"
+	"github.com/spacemeshos/go-spacemesh/timesync"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


tortoise/verifying_tortoise_test.go
```diff
@@ -818,22 +817,54 @@ func TestAddToMesh(t *testing.T) {
...
+	require.NoError(t, addLayerToMesh(mdb, l3))
 	l4 := createTurtleLayer(t, types.GetEffectiveGenesis().Add(4), mdb, atxdb, alg.BaseBlock, getHareResults, defaultTestLayerSize)
 	require.NoError(t, addLayerToMesh(mdb, l4))
+
+	alg.HandleIncomingLayer(context.TODO(), l1.Index())
+	alg.HandleIncomingLayer(context.TODO(), l3.Index())
 	alg.HandleIncomingLayer(context.TODO(), l4.Index())
-	require.Equal(t, int(types.GetEffectiveGenesis().Add(3).Uint32()), int(alg.LatestComplete().Uint32()), "wrong latest complete layer")
+	alg.HandleIncomingLayer(context.TODO(), l2.Index())
+	require.Equal(t, types.GetEffectiveGenesis(), alg.LatestComplete())
```
countvonzero: is this expected? @lrettig 
if not, how should i change the test setup to allow tortoise make progress?

tortoise/verifying_tortoise_test.go
```diff
@@ -818,22 +817,54 @@ func TestAddToMesh(t *testing.T) {
 	require.NoError(t, addLayerToMesh(mdb, l2))
 	alg.HandleIncomingLayer(context.TODO(), l2.Index())
 
-	require.Equal(t, int(types.GetEffectiveGenesis().Add(1).Uint32()), int(alg.LatestComplete().Uint32()))
+	require.Equal(t, types.GetEffectiveGenesis().Add(1), alg.LatestComplete())
```
lrettig, line 820: I add the `int()` here to make the error output more readable. Without this, the error is printed in hex. Would prefer to leave it.

lrettig: it looks like this without `int()`:

```
        	Error Trace:	verifying_tortoise_test.go:902
        	Error:      	Not equal: 
        	            	expected: types.LayerID{Value:0x7}
        	            	actual  : types.LayerID{Value:0x8}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 (types.LayerID) {
        	            	- Value: (uint32) 7
        	            	+ Value: (uint32) 8
        	            	 }
        	Test:       	TestHareOutputLayersOutOfOrder
```

versus:

```
        	Error Trace:	verifying_tortoise_test.go:902
        	Error:      	Not equal: 
        	            	expected: 7
        	            	actual  : 8
        	Test:       	TestHareOutputLayersOutOfOrder
```

lrettig (approved):
I found and fixed one small issue that this test exposed (base block selection would fail if data were missing for a single layer), so that's good. As for the test itself, I updated it along the lines I think you intended. The main difference is that, rather than creating all of the layers upfront before processing any of them, they're created and processed in order. If you don't do this, then blocks in later layers won't vote for blocks in earlier layers, which I think is not what you intended.

The net result here is still that tortoise gets stuck, however. The problem is that, when Hare gets stuck for a layer, then blocks in later layers vote against the missing layer (they do so implicitly: if a layer does not explicitly support a block in an earlier layer, then implicitly that's counted as a vote against). So the global opinion (based on vote counting) will be against the missing layer, but the local opinion (based on what's in the database) will be support.

What happens next depends on how long Hare gets stuck and when blocks in the following layers are created. If Hare is only stuck briefly, e.g., for layer N while layers N+1 and N+2 are created, then once layers N+3...N+6 are created, they will vote _for_ the missing layer, the global vote threshold will be passed, and tortoise will verify all the intervening layers (even without self healing). If Hare fails permanently, or for a sufficiently long period, then I think we end up falling back on self-healing.

To test this, you could try making the test run longer, for a few more (good) layers, and see if tortoise catches up.

If we think we can, or want to, do better than this, let's discuss it with research.

countvonzero:
> I found and fixed one small issue that this test exposed (base block selection would fail if data were missing for a single layer), so that's good. As for the test itself, I updated it along the lines I think you intended. The main difference is that, rather than creating all of the layers upfront before processing any of them, they're created and processed in order. If you don't do this, then blocks in later layers won't vote for blocks in earlier layers, which I think is not what you intended.
> 
> The net result here is still that tortoise gets stuck, however. The problem is that, when Hare gets stuck for a layer, then blocks in later layers vote against the missing layer (they do so implicitly: if a layer does not explicitly support a block in an earlier layer, then implicitly that's counted as a vote against). So the global opinion (based on vote counting) will be against the missing layer, but the local opinion (based on what's in the database) will be support.
> 
> What happens next depends on how long Hare gets stuck and when blocks in the following layers are created. If Hare is only stuck briefly, e.g., for layer N while layers N+1 and N+2 are created, then once layers N+3...N+6 are created, they will vote _for_ the missing layer, the global vote threshold will be passed, and tortoise will verify all the intervening layers (even without self healing). If Hare fails permanently, or for a sufficiently long period, then I think we end up falling back on self-healing.
> 
> To test this, you could try making the test run longer, for a few more (good) layers, and see if tortoise catches up.
> 
> If we think we can, or want to, do better than this, let's discuss it with research.

thanks @lrettig let me try to translate your words into the network params.

i recall that if hare terminates after max iterations, it should output an empty layer. say layer N got stuck for `maxIterations`. and since before `hDist`, tortoise should vote according to the hare output, all layers after N should vote against blocks in layer N. since `hDist` will at least twice as long as `maxIterations/4` layers so we have a chance recover from an attack without resorting to self-healing. 

say `maxIteration` is 40 and `hdist` is 25.  by the time hare terminate with an empty layer for N, hare already finishes layer N+10. tortoise verified layer is stuck at `N-1` (as there are votes supporting layer `N-1` in layer `N+k` where 0 <k<=hdist )
so do you mean that by the time we are at `N+25`, we should be able to advance tortoise's verified layer to N (or a leap from N-1 to N+24)? and if not, we will enter self-healing?

FYI @noamnelke

lrettig:
> say layer N got stuck for `maxIterations`. and since before `hDist`, tortoise should vote according to the hare output, all layers after N should vote against blocks in layer N.

while a node is still waiting for Hare to finish, new blocks/layers explicitly abstain on blocks in layer N. after Hare fails with an empty set, then new blocks/layers created after that implicitly vote against blocks in layer N.

> so do you mean that by the time we are at `N+25`, we should be able to advance tortoise's verified layer to N (or a leap from N-1 to N+24)? and if not, we will enter self-healing?

the exact ratio of how many layers we need relative to `N` to recover from this scenario depends upon other tortoise params (local and global threshold), but we shouldn't need self-healing here. after Hare fails, _all subsequent layers will (implicitly) vote against all the blocks in all the layers for which Hare failed._ eventually there will be enough votes to verify those layers (as empty), and the ones that follow.

self-healing is designed to recover from scenarios where our security assumptions fail to hold, e.g., a prolonged network split, or Hare failing permanently. I don't think any security assumptions are violated in this scenario.

countvonzero:
thanks for the detailed explanation.

> self-healing is designed to recover from scenarios where our security assumptions fail to hold, e.g., a prolonged network split, or Hare failing permanently. I don't think any security assumptions are violated in this scenario.

mostly agreed. but this scenario could happen if more than 1/3 of nodes are adversarial and deliberately caused honest nodes to abstain. i think that's why Tal wanted `hdist` to be at least twice as long as `maxIterations/4` so that the network has a chance to recover, assuming the ratio of adversarial dropped within that time, before engaging in self-healing. i think in this case the security assumption is also broken, albeit for a short period of time.

i will try to change this test that eventually make progress with verified layer. that's the end goal i'm trying to achieve in this PR

lrettig:
One more thing to note: you may need to look at the `Zdist` and `ConfidenceParam` params, in addition to `Hdist`. These should all be pretty clearly documented in the code.

countvonzero:
resolved by https://github.com/spacemeshos/go-spacemesh/pull/2922

Issue #2764, opened 2021-09-03
Flaky systemtest-blocks-remove-node


dshulyak:
Last failed in https://github.com/spacemeshos/go-spacemesh/runs/3494720283

```
>             assert blocks_per_layer == wanted_res, ass_err
E           AssertionError: node 1eaf0dbd349cf2b08dd9ca13f300de5fc3ba4ea6140a1a9eb54f83c6698e257b failed creating the avg block size
E           blocks created per layer 1.0, wanted average block per node 0.6666666666666666, layer_avg_size 50, layers_per_epoch 3, num_miners 51, blocks_sum 3, to_layer 9, from_layer 6, blocks defaultdict(<class 'list'>, {6: [<Hit(fluent-bit-tgocx-2021.09.02/_doc/bAIspnsB25Eb8GgiCdQV): {'@timestamp': '2021-09-02T11:01:48.056Z', 'stream': 'stdout...}>, <Hit(fluent-bit-tgocx-2021.09.02/_doc/cAIspnsB25Eb8GgiCdQV): {'@timestamp': '2021-09-02T11:01:48.056Z', 'stream': 'stdout...}>, <Hit(fluent-bit-tgocx-2021.09.02/_doc/cwIspnsB25Eb8GgiCdQV): {'@timestamp': '2021-09-02T11:01:48.056Z', 'stream': 'stdout...}>], 9: [<Hit(fluent-bit-tgocx-2021.09.02/_doc/9AgupnsB25Eb8Ggixxd3): {'@timestamp': '2021-09-02T11:04:48.022Z', 'stream': 'stdout...}>], 7: [], 8: []})
```

I am unable to get logs from that namespace. Will look more once it fails again. If someone sees a similar failure (blocks are not created, more blocks created by another node, weight is not expected) check the number of atxs published, sometimes one node is stuck in discovery for longer time and misses one or two initial atxs.

dshulyak:
Same issue in mining test https://github.com/spacemeshos/go-spacemesh/runs/3499707195

One node didn't publish atx in epoch 1.

PR #2775, opened 2021-09-10
[Merged by Bors] - Run weak coin only for following vote rounds and after votes were received

dshulyak:
## Motivation

It bubbled up some time ago and also flagged by @countvonzero that there is a mistake in weak coin execution order. It needs to be executed *after* vote was sent to the network. Otherwise, a byzantine actor can negate the weak coin effect by adjusting its own vote.

Also, votes from the last round were ignored fully.

## Changes
- run weak coin after receiving votes for the current round
- change timeouts accordingly
- don't return errors from goroutines spawned in runConsensusPhase, as it will shutdown whole tortoise beacon
- lowercases some logs

## TODO

## DevOps Notes


nkryuchkov (approved):

dshulyak:
bors merge

dshulyak:
weird, i thought that the beacon is disabled, but apparently it is not disabled for block builder
https://kibana-tests.spacemesh.io/goto/aaf4659c6bd032345de6002a0295393e

dshulyak:
bors try

dshulyak:
The test failed because node was expected to produce blocks, but because beacon wasn't found it skipped block production. There are other errors such as atx is not found but those are not the root cause of the failure and doesn't seem to be a new thing in general. Tortoise beacon probably wasn't disabled completely.

There are also a lot of `received malformed first voting message, already received a voting message for this pk and round` which is strange. This should be only if previously received messages were evicted from duplicates cache. But i don't see how that can be related to this change.

tortoisebeacon/tortoise_beacon.go
```diff
@@ -503,84 +503,61 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
-
 	for round := firstRound; round <= tb.lastRound(); round++ {
-		// always use coinflip from the previous round for current round.
-		// round 1 is running without coinflip (e.g. value is false) intentionally
 		round := round
-		previousCoinFlip := coinFlip
+		votes := ownVotes
 		tb.eg.Go(func() error {
 			if round == firstRound {
 				if err := tb.sendProposalVote(ctx, epoch); err != nil {
```
countvonzero, line 498: sendProposalVote() should initialize ownVotes

tortoisebeacon/tortoise_beacon.go
```diff
@@ -503,84 +503,61 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
-
-			return nil
-		})
-
-		tb.eg.Go(func() error {
-			tb.startWeakCoinRound(ctx, epoch, round)
 			return nil
 		})
+		if round != firstRound {
+			tb.eg.Go(func() error {
```
countvonzero: i think there is still a potential race condition here. a malicious peer can still wait till the weakcoin round to finish before it send out the votes, and their late votes will still be counted in this round.

i think we should
- send vote
- wait for the duration for vote to come in
- start weak coin (this is the cutoff time for counting votes in this round)
- count votes without weak coin
- finish weak coin
- count undecided votes with weak coin
- the late vote received after weak coin started will be added to the next round

tortoisebeacon/db_test.go
```diff
@@ -0,0 +1,13 @@
+package tortoisebeacon
+
+import (
+	"testing"
+
+	"github.com/spacemeshos/go-spacemesh/database"
+	"github.com/spacemeshos/go-spacemesh/log/logtest"
+)
+
+func newMemDB(tb testing.TB) *DB {
```
countvonzero: is this used?

dshulyak: it was used in the test which i didn't submit in this pr. can remove

tortoisebeacon/tortoise_beacon.go
dshulyak: why? in current code votes are initialized by calcVotes after received all first round messages. it is not correct?

dshulyak: i think sendProposalVote will just use initial votes for first voting message that were formed from all proposals

tortoisebeacon/tortoise_beacon.go
```diff
@@ -503,84 +503,61 @@ func (tb *TortoiseBeacon) runConsensusPhase(ctx context.Context, epoch types.Epo
...
+				tb.startWeakCoinRound(ctx, epoch, round)
+				return nil
+			})
+		}
 
 		select {
 		case <-timer.C:
 			timer.Reset(tb.config.VotingRoundDuration + tb.config.WeakCoinRoundDuration)
```
countvonzero: this needs to adjust too.

are we sure in the test params that voting round duration is a lot longer than weakcoin duration?

tortoisebeacon/tortoise_beacon.go
dshulyak: that makes sense

tortoisebeacon/tortoise_beacon.go
dshulyak: > this needs to adjust too.

you mean if votes will have to be tallied twice?

> are we sure in the test params that voting round duration is a lot longer than weakcoin duration?

it is longer by 10-15s. why do you think that it should be *a lot* longer?

tortoisebeacon/tortoise_beacon.go
countvonzero: can you point me to the code where this node's own proposal is added to votesMargin? i am lost in the sea of code now...

countvonzero: > why? in current code votes are initialized by calcVotes after received all first round messages. it is not correct?

it is, but that's only counting votes the node received. i cannot find where this node's own initial proposal/vote was added.

dshulyak: it just uses incoming proposals for payload https://github.com/spacemeshos/go-spacemesh/blob/782b11dcdd98ee10c4c43375f82a6236a192715c/tortoisebeacon/tortoise_beacon.go#L629-L636

dshulyak: https://github.com/spacemeshos/go-spacemesh/blob/782b11dcdd98ee10c4c43375f82a6236a192715c/tortoisebeacon/tortoise_beacon.go#L486-L489

countvonzero:
> The test failed because node was expected to produce blocks, but because beacon wasn't found it skipped block production. There are other errors such as atx is not found but those are not the root cause of the failure and doesn't seem to be a new thing in general. Tortoise beacon probably wasn't disabled completely.

i wonder whether this is a timing/duration params issue.
this PR didn't seem to delay the time beacon is generated and just shifts the order of operations around inside the time frame.

dshulyak:
> i wonder whether this is a timing/duration params issue.
this PR didn't seem to delay the time beacon is generated and just shifts the order of operations around inside the time frame.

yes very confusing. if anything it became a bit shorter since the whole tortoise beacon is shorter. previously it was, proposal + first + wc + n-1*(round + wc) and now first wc is gone.

countvonzero:
> yes very confusing. if anything it became a bit shorter since the whole tortoise beacon is shorter. previously it was, proposal + first + wc + n-1*(round + wc) and now first wc is gone.

hmm.. i think it is
before: 
`FirstVotingRoundDuration+WeakCoinRoundDuration + (K-1)(VotingRoundDuration+WeakCoinRoundDuration)`

after: 
`FirstVotingRoundDuration + (K-1)(VotingRoundDuration+WeakCoinRoundDuration+timeSpentCalculatingVotes)`

unless `timeSpentCalculatingVotes` is significant, i agree it should be shorter

countvonzero:
> unless timeSpentCalculatingVotes is significant, i agree it should be shorter

actually i am wrong. the (K-1) multiplier probably did make the beacon generation time longer.

tortoisebeacon/tortoise_beacon.go
countvonzero: oh. i just mean that in this order of operation, VotingRoundDuration includes weakCoinRoundDuration. 
conceptually a voting round includes sending votes, generating weak coin, and counting votes. so i thought VotingRoundDuration covers all those.

tortoisebeacon/tortoise_beacon.go
countvonzero: @dshulyak thanks. but this only serves to send our own proposal to peers, and in later encoding for following votes.
i still don't see where `tb.votesMargin` is updated with own initial votes.
`tb.votesMargin` seems to be only updated when it receives a gossip message from peers.

dshulyak:
> timeSpentCalculatingVotes

it is very low, especially with the amount of nodes in tests. i have inprocess test with 50 nodes, it takes around ~1ms for calcVotes to complete.

tortoisebeacon/tortoise_beacon.go
dshulyak: ah sorry i understood now what we were talking about. everything that is Broadcast'ed will be sent to the local handler for the protocol as well. So, what is sent in `sendProposalVote` will be also received HandleSerializedFirstVotingMessage. that's sort of a feature of p2p framework

dshulyak:
bors try

tortoisebeacon/tortoise_beacon.go
countvonzero: OMG... i'm learning something new about p2p every day. cool! thanks for your patience @dshulyak

countvonzero: but then, does that mean we add our own proposal to the valid list twice? once in `proposalPhaseImpl` and once in the proposal handler? @dshulyak

dshulyak:
previous test failure has the same root cause as in https://github.com/spacemeshos/go-spacemesh/issues/2764 . definitely doesn't have anything to do with tortoise beacon.

bors try

tortoisebeacon/tortoise_beacon.go
dshulyak: i think this is the case. i don't see any code that prevents adding duplicate proposals

tortoisebeacon/tortoise_beacon.go
dshulyak: @countvonzero i updated the code. why do you think that we should:
> the late vote received after weak coin started will be added to the next round

why not to ignore that vote? in the next round we will have a next set of votes

dshulyak:
I guess i will need to address the problem of slow discovery to understand if there is any additional flakiness caused by this PR, previous test failed exactly the same way. One node wasn't able to start sync on time and atx in epoch 1 wasn't published.

dshulyak:
bors try

tortoisebeacon/tortoise_beacon.go
countvonzero: that was basedd on Tal in the R&D meeting. my impression is tgat we want as many vote as possible to cross threshold. as long as a node isnt double voting, we should count it. but if that introduce extra code complexity, we should talk to Tal abt this

dshulyak: i think thats possible in general but we need to revisit how votes are buffered. but what i don't understand is what we should do with the vote that will be sent in the next round in such case? i don't think that we should count votes from the same smesher twice (e.g. first time from late vote in the previous round and second time from timely vote in the current round)

dshulyak:
bors try

tortoisebeacon/tortoise_beacon.go
countvonzero: it's not so much `count votes from the same smesher twice` tho. essentially every smesher gets to vote K times where K is the number of rounds. 

if we consider a honest smesher (not casting multiple votes for the same round), then it's just network delay and there is not reason to _not_ count it.

but if a smesher is malicious, casting late votes (after it learns the weakcoin) can potentially sway the vote to cross the threshold early in that round.

my guess would be it's OK to just drop late votes for code simplicity. but it's worth to double-check with Tal maybe in tuesday R&D meeting

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

tortoisebeacon/tortoise_beacon.go
dshulyak: sure lets discuss it. i can also open a follow up issue to refactor buffering to make counting late votes easier

dshulyak:
bors try

dshulyak:
bors merge

PR #2776, opened 2021-09-10
[Merged by Bors] - Always report metrics if metrics server is started

### LINKED ISSUE #2773: application metrics are always disabled
## Description

Metrics module defines a variable https://github.com/spacemeshos/go-spacemesh/blob/develop/metrics/common.go#L17, which is if set to False (default) will make metrics noop, and nothing will be reported.

Hard to tell when (or if) it worked before, it clearly doesn't work in current devnet and i don't see anything that can overwrite this value. Maybe it was meant to be overwritten during the build using `-X metrics.Enable=true` ? 

## Steps to reproduce

Run your node with metrics enabled or port-forward to any container on 1010 port and request /metrics endpoint. 

## Actual Behavior

Only golang system metrics are reported.

## Expected Behavior

All metrics are reported.

dshulyak:
## Motivation

If Enabled was false we would use noop metrics. The only case when metrics are not needed is for unit tests. But metrics are light weight and shouldn't be a problem, so i suggest to simplify and remove Enabled variable completely.

Closes https://github.com/spacemeshos/go-spacemesh/issues/2773

## Changes
- if metrics server is enabled metrics will be reported always

## TODO

## DevOps Notes


countvonzero (approved):
hahah... i might have just switched it true. but i think your approach is better: fix whatever happens caused by enabling metrics.

dshulyak:
maybe some unit tests will become a bit more flaky, but it shouldn't be overwhelming.

bors merge

PR #2778, opened 2021-09-14
[Merged by Bors] - tortoisebeacon: initialize proposal threshold just once per epoch

### LINKED ISSUE #2786: tortoise beacon: calculate proposal eligibility threshold once per epoch
## Description

currently this calculation is done for every proposal message. when the network scales, this will become very inefficient.
the only difference is that we may be able to capture late ATXs that arrives in between calculation, which doesn't justify the cost.

this is an optimization before https://github.com/spacemeshos/go-spacemesh/issues/2785 is resolved.

## Expected Behavior
proposal eligibility calculation should be as efficient as possible

countvonzero:
## Motivation

Closes #2786
- proposal threshold is calculated once per proposal based on the total weight of previous epoch. this is inefficient.
- the more test the merrier. this is the first PR that add tests to tortoise beacon. more to follow.

## Changes

in commit order
- simplify error handling
- reuse more log.WithContext()
- initialize a proposal eligibility checker and only do the threshold calculation once at the start of epoch
- remove the receiver for some functions to facilitate testing
- add test cases:
  - proposal that doesn't pass eligibility
  - proposal with bad vrf signature

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/tortoisebeacon/handlers.go
+++ b/tortoisebeacon/handlers.go
@@ -77,6 +77,31 @@ func (tb *TortoiseBeacon) HandleSerializedProposalMessage(ctx context.Context, d
 	}
 }
 
+func (tb *TortoiseBeacon) readProposalMessagesLoop(ctx context.Context, ch chan *proposalMessageWithReceiptData) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+
+		case em := <-ch:
+			if em == nil {
+				return
+			}
+
+			if err := tb.handleProposalMessage(ctx, em.message, em.receivedTime); err != nil {
+				tb.Log.WithContext(ctx).With().Error("failed to handle proposal message",
+					log.String("sender", em.gossip.Sender().String()),
+					log.String("message", em.message.String()),
+					log.Err(err))
+
+				return
+			}
+
+			em.gossip.ReportValidation(ctx, TBProposalProtocol)
+		}
+	}
+}
+
 func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalMessage, receivedTime time.Time) error {
 	currentEpoch := tb.currentEpoch()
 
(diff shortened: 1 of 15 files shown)


tortoisebeacon/handlers.go
```diff
@@ -344,27 +332,30 @@ func (tb *TortoiseBeacon) storeFirstVotes(message FirstVotingMessage, minerPK *s
...
+			tb.votesMargin[p].Add(tb.votesMargin[p], voteWeight)
 		}
 	}
 
-	for _, vote := range message.PotentiallyValidProposals {
-		if _, ok := tb.votesMargin[string(vote)]; !ok {
-			tb.votesMargin[string(vote)] = new(big.Int).Neg(voteWeight)
+	for _, proposal := range message.PotentiallyValidProposals {
+		p := string(proposal)
+		if _, ok := tb.votesMargin[p]; !ok {
```
dshulyak, line 346: if string([]byte) is passed as a key to the map (e.g. tb.votesMargin) it is optimized by compiler and memory is not copied. not a big deal probably but still cleaner. so in previous code it was copied only when key didn't exist, and not every time

tortoisebeacon/weakcoin/weak_coin.go
```diff
@@ -290,19 +290,17 @@ func (wc *WeakCoin) FinishRound() {
 	coinflip := wc.smallestProposal[lsbIndex]&1 == 1
 
 	wc.coins[wc.round] = coinflip
-	wc.logger.With().Info("completed round",
-		log.Uint32("epoch_id", uint32(wc.epoch)),
-		log.Uint32("round_id", uint32(wc.round)),
+	logger.With().Info("completed round",
 		log.String("proposal", types.BytesToHash(wc.smallestProposal).ShortString()),
-		log.Bool("coinflip", coinflip))
+		log.Bool("tb_weakcoin", coinflip))
```
dshulyak, line 295: whats the point of tb prefix?

dshulyak (approved):

tortoisebeacon/handlers.go
countvonzero: ah. good to know. thanks!

tortoisebeacon/weakcoin/weak_coin.go
countvonzero: i was trying to differentiate between self-healing weakcoin and tortoise beacon weakcoin

countvonzero:
bors merge

nkryuchkov (approved):

PR #2780, opened 2021-09-14
[Merged by Bors] - Wait for any number of peers not only outbound when sync is started

### LINKED ISSUE #2764: Flaky systemtest-blocks-remove-node

dshulyak:
## Motivation

I had multiple errors due to the slow discovery recently. I also tried other approaches:
1. randcon to 2 resulted in flaky errors due to the problems with message delivery
2. randcon to 5 still got same errors with slow discovery

So i decided to use any peers when sync is started. Note that it uses same parameter (`randcon`) but now we will wait for any peers, not only outbound peers when starting discovery. I didn't want to introduce additional parameter specifically for starting sync.

I will open an issue once this is merged to make sure that we will address the requirement to wait for outbound peers.

Closes https://github.com/spacemeshos/go-spacemesh/issues/2764

## Changes
- wait for both inbound and outbound peers when starting sync  

## Test Plan
app and system tests

## TODO

## DevOps Notes

### DIFF
--- a/cmd/node/multi_node.go
+++ b/cmd/node/multi_node.go
@@ -131,6 +131,8 @@ func getTestDefaultConfig() *config.Config {
 		log.Error("cannot load config from file")
 		return nil
 	}
+	// is set to 0 to make sync start immediatly when node starts
+	cfg.P2P.SwarmConfig.RandomConnections = 0
 
 	cfg.POST = activation.DefaultPostConfig()
 	cfg.POST.LabelsPerUnit = 32
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -45,6 +45,7 @@ import (
 	"github.com/spacemeshos/go-spacemesh/metrics"
 	"github.com/spacemeshos/go-spacemesh/miner"
 	"github.com/spacemeshos/go-spacemesh/p2p"
+	"github.com/spacemeshos/go-spacemesh/p2p/peers"
 	"github.com/spacemeshos/go-spacemesh/p2p/service"
 	"github.com/spacemeshos/go-spacemesh/pendingtxs"
 	"github.com/spacemeshos/go-spacemesh/priorityq"
@@ -312,6 +313,7 @@ type App struct {
 	atxDb          *activation.DB
 	poetListener   *activation.PoetListener
 	edSgn          *signing.EdSigner
+	watchPeers     *peers.Peers
 	tortoiseBeacon TortoiseBeaconService
 	closers        []interface{ Close() }
 	log            log.Log
@@ -772,6 +774,7 @@ func (app *App) initServices(ctx context.Context,
 	app.state = processor
 	app.hare = rabbit
 	app.P2P = swarm
+	app.watchPeers = peers.Start(swarm)
 	app.poetListener = poetListener
 	app.atxBuilder = atxBuilder
 	app.postSetupMgr = postSetupMgr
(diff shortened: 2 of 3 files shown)


cmd/node/multi_node.go
```diff
@@ -131,6 +131,7 @@ func getTestDefaultConfig() *config.Config {
 		log.Error("cannot load config from file")
 		return nil
 	}
+	cfg.P2P.SwarmConfig.RandomConnections = 0
```
dshulyak, line 135: this needs to be set to 0, because simulator returns closed channel for initial discovery and the behaviour needs to be preserved for tests to pass

y0sher:
the only thing is, it means that it'll be possible that we sync initially only from inbound peers which really sounds like an easy eclipse attack vector. though if it opens some bottleneck now, I agree it's better to clear up the way and properly plan this behavior or take any security measures in the future when we start adversarial testing..

dshulyak:
i think we should be able to recover if we got invalid data by using approach described in https://github.com/spacemeshos/SMIPS/issues/51. that's why i was arguing that it is not a big deal if we will get "good" peers after "bad" as long as we will get them. maybe resource-wise it will be more efficient to start with the good.

cmd/node/multi_node.go
countvonzero: can you add comments to the code? this is not intuitive.

countvonzero (approved):

countvonzero:
> the only thing is, it means that it'll be possible that we sync initially only from inbound peers which really sounds like an easy eclipse attack vector

if the adversary wants to pull off an eclipse attack on a node, it's likely it'll redirect the node's outbound connections as well. does it really matter that much?

dshulyak:
bors merge

PR #2790, opened 2021-09-15
[Merged by Bors] - tortoise beacon: disallow double proposals and add more tests

### LINKED ISSUE #2772: tortoise beacon: own FollowingVotingMessage's are marked as a duplicates by a p2p layer
## Description

On devnet from recent version (or system tests) you will see this warning `own message is marked as old` with `protocol` equal to TBFollowingVotingProtocol.

The reason for this is that most of the following voting messages are not unique.

```go
	MinerID        types.NodeID
	RoundID        types.RoundID
	VotesBitVector []uint64
}
``` 

Gossip router holds last 10000 hashes of messages in the local cache to deduplicate them, so even if message is logically different but not unique it will not be sent to the network. And i believe there is also a security issue as well, since someone can rebroadcast previously sent vote in the new epoch and make smesher look malicious.

We should include EpochID into the message to make it unique.

## Steps to reproduce

Check logs for the tortoise beacon in devnet or system tests.

cc @nkryuchkov @countvonzero

### LINKED ISSUE #2787: tortoise beacon: do not accept proposal after phase ends
## Description

the tortoise beacon consists of proposal phase and consensus phase. 
current code continues to accept proposals after the proposal phase ends
- may cause debugging confusion
- can free up memory after beacon is generated for the epoch

## Expected Behavior
tortoise beacon does not accept more proposal messages after proposal phase ends

### LINKED ISSUE #2789: tortoise beacon: smesher can make multiple proposals and be accepted
## Description

current code does not keep track whether a node has made a proposal for the epoch.
code as is in the develop branch, the miner's own proposal will be added twice, once in `proposalPhaseImpl` and once `handleProposalMessage` after the proposal is gossiped.

## Expected Behavior
there should be a generic mechanism to prevent the same smesher to make multiple proposals

countvonzero:
## Motivation

Closes #2772 
Closes #2787
Closes #2789

## Changes

- remove initial epoch delay
- use cached epoch weight everywhere
- stop accepting proposals after proposal round ends
- add epoch ID to first and following vote messages
- drop vote messages from gossip for mismatching epoch ID
- count own proposal once 
- make incoming proposals thread-safe
- prevent double proposals
- add tests for gossip handers

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/tortoisebeacon/handlers.go
+++ b/tortoisebeacon/handlers.go
@@ -32,7 +32,9 @@ var (
 	// ErrMalformedSignature is returned when signature is malformed.
 	ErrMalformedSignature = errors.New("malformed signature")
 
-	// ErrAlreadyVoted is returned when miner has already voted.
+	// ErrAlreadyProposed is returned when miner doubly proposed for the same epoch.
+	ErrAlreadyProposed = errors.New("already proposed")
+	// ErrAlreadyVoted is returned when miner doubly voted for the same epoch/round.
 	ErrAlreadyVoted = errors.New("already voted")
 )
 
@@ -191,6 +193,10 @@ func (tb *TortoiseBeacon) verifyProposalMessage(ctx context.Context, m ProposalM
 		logger.Warning("received malformed proposal message: vrf is not verified")
 
 		return types.ATXID{}, ErrMalformedProposal
+	}
+
+	if err := tb.registerProposed(vrfPK, logger); err != nil {
+		return types.ATXID{}, err
 	}
 
 	proposalShortString := types.BytesToHash(m.VRFSignature).ShortString()
@@ -476,6 +482,21 @@ func (tb *TortoiseBeacon) currentEpoch() types.EpochID {
 	tb.mu.RLock()
 	defer tb.mu.RUnlock()
 	return tb.epochInProgress
+}
+
+func (tb *TortoiseBeacon) registerProposed(minerPK *signing.PublicKey, logger log.Log) error {
+	tb.mu.Lock()
+	defer tb.mu.Unlock()
+
+	minerID := string(minerPK.Bytes())
+	if _, ok := tb.hasProposed[minerID]; ok {
+		// see TODOs for registerVoted()
+		logger.Warning("already received proposal from miner")
+		return ErrAlreadyProposed
+	}
+
+	tb.hasProposed[minerID] = struct{}{}
+	return nil
 }
 
 func (tb *TortoiseBeacon) registerVoted(minerPK *signing.PublicKey, round types.RoundID, logger log.Log) error {
--- a/tortoisebeacon/handlers.go
+++ b/tortoisebeacon/handlers.go
@@ -196,7 +196,8 @@ func (tb *TortoiseBeacon) verifyProposalMessage(ctx context.Context, m ProposalM
 		// the peer may have different total weight from us so that it passes threshold for the peer
 		// but does not pass here
 		logger.With().Warning("rejected proposal that doesn't pass threshold",
-			log.String("proposal", proposalShortString))
+			log.String("proposal", proposalShortString),
+			log.Uint64("total_weight", tb.epochWeight))
 
 		return types.ATXID{}, ErrProposalDoesntPassThreshold
 	}
--- a/tortoisebeacon/handlers.go
+++ b/tortoisebeacon/handlers.go
@@ -53,18 +53,10 @@ func (tb *TortoiseBeacon) HandleSerializedProposalMessage(ctx context.Context, d
 		return
 	}
 
-	currentEpoch := tb.currentEpoch()
-	if message.EpochID < currentEpoch {
-		logger.With().Debug("ignoring proposal message from previous epoch",
-			log.Uint32("message_epoch", uint32(message.EpochID)),
-			log.Uint32("current_epoch", uint32(currentEpoch)))
-
+	ch := tb.getProposalChannel(ctx, message.EpochID)
+	if ch == nil {
 		return
 	}
-
-	tb.mu.Lock()
-	ch := tb.getOrCreateProposalChannel(message.EpochID)
-	tb.mu.Unlock()
 
 	proposalWithReceipt := &proposalMessageWithReceiptData{
 		message:      message,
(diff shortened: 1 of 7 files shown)


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/3611106498)

it failed at the dockerpush stage. all passed after retry.

tortoisebeacon/handlers.go
```diff
@@ -473,19 +469,61 @@ func (tb *TortoiseBeacon) storeFollowingVotes(message FollowingVotingMessage, mi
...
+	// TODO(nkryuchkov): consider having a separate table for an epoch with one bit in it if atx/miner is voted already
+	if _, ok := tb.hasVoted[round][minerID]; ok {
+		logger.Warning("already received vote from miner for this round")
+
+		// TODO(nkryuchkov): report this miner through gossip
+		// TODO(nkryuchkov): store evidence, generate malfeasance proof: union of two whole voting messages
+		// TODO(nkryuchkov): handle malfeasance proof: we have a blacklist, on receiving, add to blacklist
+		// TODO(nkryuchkov): blacklist format: key is epoch when blacklisting started, value is link to proof (union of messages)
+		// TODO(nkryuchkov): ban id forever globally across packages since this epoch
+		// TODO(nkryuchkov): (not tortoise beacon) do the same for ATXs
```
dshulyak, line 522: we shouldn't do anything if we simply received a duplicate message. p2p layer provides only best effort guarantee that the same message won't be delivered many times. so even printing a warning is not necessary.

tortoisebeacon/handlers.go
```diff
@@ -473,19 +469,61 @@ func (tb *TortoiseBeacon) storeFollowingVotes(message FollowingVotingMessage, mi
...
-	tb.hasVoted[message.RoundID][string(minerPK.Bytes())] = struct{}{}
 }
 
 func (tb *TortoiseBeacon) currentEpoch() types.EpochID {
 	tb.mu.RLock()
 	defer tb.mu.RUnlock()
 	return tb.epochInProgress
 }
+
+func (tb *TortoiseBeacon) registerProposed(minerPK *signing.PublicKey, logger log.Log) error {
```
dshulyak, line 489: how do we handle local proposals now? relying exclusively on the self-broadcast to receive them?

tortoisebeacon/handlers.go
```diff
@@ -473,19 +469,61 @@ func (tb *TortoiseBeacon) storeFollowingVotes(message FollowingVotingMessage, mi
...
+	tb.mu.Lock()
+	defer tb.mu.Unlock()
+
+	minerID := string(minerPK.Bytes())
+	if _, ok := tb.hasProposed[minerID]; ok {
+		// see TODOs for registerVoted()
+		logger.Warning("already received proposal from miner")
+		return ErrAlreadyProposed
```
dshulyak, line 497: is it going to be printed with Error if we simply received same proposal twice?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -460,12 +459,39 @@ func (tb *TortoiseBeacon) proposalPhaseImpl(ctx context.Context, epoch types.Epo
...
+			return nil
+		}
+		return tb.getOrCreateProposalChannel(epoch)
+	}
+	// always accept proposals for the next epoch, but not too far in the future
+	if epoch == tb.epochInProgress+1 {
+		ch := tb.getOrCreateProposalChannel(epoch)
+		if len(ch) == proposalChanCapacity {
+			// the reader loop is not started for the next epoch yet. drop old messages if it's already full
+			msg := <-ch
```
dshulyak: i think it is correct to select with default to make in nonblocking. receive is not synchronous with length check, so it may block in some testing for example

tortoisebeacon/handlers.go
countvonzero: the intention is 
1. rely on gossip to catch same round votes with the same contents
2. and here catch same round votes different contents

it looks like step 1 is not reliable. if we choose to blacklist a peer for double voting, we will have to remember the votes in that round. i think this is manageable memory wise as we should be only accepting votes for one round at a time.

tortoisebeacon/handlers.go
countvonzero: it's going to be printed with Warning here, and nowhere else.

tortoisebeacon/handlers.go
countvonzero: yes. is there a possibility that self-broadcast won't work?

tortoisebeacon/tortoise_beacon.go
countvonzero: nice catch. thanks. done

tortoisebeacon/handlers.go
dshulyak: just wanted to clarify. i don't like the pattern much as it seems wasteful to me, but nothing besides that

tortoisebeacon/handlers.go
dshulyak: definitely shouldn't be relying on 1, if we are going to do something about the peer

dshulyak (approved):

tortoisebeacon/tortoise_beacon.go
```diff
@@ -460,12 +459,43 @@ func (tb *TortoiseBeacon) proposalPhaseImpl(ctx context.Context, epoch types.Epo
...
 
-	tb.mu.Lock()
-	defer tb.mu.Unlock()
-
-	tb.incomingProposals.valid = append(tb.incomingProposals.valid, proposedSignature)
+func (tb *TortoiseBeacon) getProposalChannel(ctx context.Context, epoch types.EpochID) chan *proposalMessageWithReceiptData {
+	tb.mu.RLock()
+	defer tb.mu.RUnlock()
 
+	if epoch < tb.epochInProgress {
```
nkryuchkov: What do you think of rewriting this to a `switch`?

nkryuchkov (approved):

tortoisebeacon/tortoise_beacon.go
countvonzero: done. thanks for the suggestion

countvonzero:
bors try

countvonzero:
bors merge

PR #2798, opened 2021-09-18
Local testnet fix

### LINKED ISSUE #43: Update Local net to run Spacemesh 0.2 full nodes
Running a local net with public testnet 128 node (go-sm tag/v0.1.29 node) and default localnet config will result in stuck verified layer as soon as layer 13 or 15. Almost every network i tried running got stuck before epoch 5. We need to find a good set of configs that balance between consensus liveness and having a great dev experience. e.g. short epoch times and layers and short time to wait until a `normal` 3rd epoch.

sudachen:
## Motivation

Starting go-spacemesh in the local-testnet requires to modification:
1. Incoming connection must be counted as well as out coming, because there is to few nodes.
2. Since local-testnet uses docker which can be started over nvidia-docker backend, it's a good idea to have possibility specify CPU compute provider for post GPU backend, otherwise it can fail to initialize post data for some nodes on local-testnet.

Is required to close https://github.com/spacemeshos/local-testnet/issues/43

## Changes
1. Added configuration switch enabling extended connection counting.
2. Added code setting CPU compute provider as active if no provider is specified or 0 is specified as provider id.


dshulyak:
we do count all peers when sync is started https://github.com/spacemeshos/go-spacemesh/commit/61c60597af01eb9cf989859832f9b8bbe5321645 . does it need to be changed in discovery as well?

sudachen:
> does it need to be changed in discovery as well?    

Yes, since gossip is started only if there are enough peers https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/switch.go#L812

dshulyak:
do you mind sharing where is that condition used? 

what i saw in the code is that closing `initial`, will close `gossipC` which will make Switch.GossipReady return. this is what was blocking syncer from starting. i think gossip itself doesn't depend on this condition and will work with any number of peers

p2p/config/config.go
```diff
@@ -56,6 +56,7 @@ type Config struct {
 	SwarmConfig           SwarmConfig   `mapstructure:"swarm"`
 	BufferSize            int           `mapstructure:"buffer-size"`
 	MsgSizeLimit          int           `mapstructure:"msg-size-limit"` // in bytes
+	LocalNetwork          bool          `mapstructure:"local-network"`
```
dshulyak: this needs to be renamed to something less specific. e.g. RandConnIncludeOutbound

sudachen: There is another point of view: The only responsibility of this configuration option is to start local network. It also means you must not use it on normal network. What you offer here does not have any responsibity meaning and just makes configuration more complicated. By the way it does not matter yet, configuration already over-comlicated and useless. So, If someone else agrees to rename I will do.

sudachen:
> do you mind sharing where is that condition used?

It was used as <-app.P2P.GossipReady() before you removed it few days ago... I see now. So, this modification is not required now. I will update PR with GPU provider modification only.

dshulyak (approved):

moshababo:
bors merge

moshababo:
bors merge

Issue #2796, opened 2021-09-16
sync: syncer fetching same block IDs multiple times


countvonzero:
## Description
when starting a new node in an existing network, we see a lot more logs like this than the managed network.

managed nodes:
there are only 476 such messages for a node that runs for 5 days.

![Screenshot from 2021-09-16 16-30-33](https://user-images.githubusercontent.com/30611210/133681661-8e831223-d6c1-4372-ba07-de0a9698c45f.png)

later-joined node: 
for the same block ID `55f63bf363`, we will get 7 such `we already know this block` messages. since our config `rancon` is 8, this strongly suggests that we are asking every peer for the same block ID.

```
2021-09-15T14:16:17.504-0400	INFO	424f8.blockListener	got new block	{"node_id": "424f89cd48733a764660172b893ceaf5412d70e5c17eabc31360346279d7e971", "module": "blockListener", "sessionId": "d12101a6-a38c-4da6-84c0-2a7a90c884a7", "block_id": "55f63bf363", "layer_id": 15, "block_id": "55f63bf363", "layer_id": 15, "epoch_id": 2, "miner_id": "b2a6a", "base_block": "9328940263", "supports": 30, "againsts": 0, "abstains": 0, "atx_id": "b7ab92db2323b29483182e5831359e8abb5057be85feb6ad0ddb685006f9153c", "eligibility_counter": 2, "ref_block": "0f1e60efc1", "active_set_size": 0, "tx_count": 0, "name": "blockListener"}
2021-09-15T14:16:17.507-0400	INFO	424f8.mesh         	added block to database	{"node_id": "424f89cd48733a764660172b893ceaf5412d70e5c17eabc31360346279d7e971", "module": "mesh", "sessionId": "d12101a6-a38c-4da6-84c0-2a7a90c884a7", "block_id": "55f63bf363"}
2021-09-15T14:16:17.523-0400	INFO	424f8.blockListener	we already know this block	{"node_id": "424f89cd48733a764660172b893ceaf5412d70e5c17eabc31360346279d7e971", "module": "blockListener", "sessionId": "d12101a6-a38c-4da6-84c0-2a7a90c884a7", "block_id": "55f63bf363", "layer_id": 15}
[...]

PR #2800, opened 2021-09-20
[Merged by Bors] - make sure layer blocks are fetched

### LINKED ISSUE #2797: sync: layerfetcher should report error fetching block and cause sync to fail
## Description

when adding a new node to an existing network, the node syncs one layer at a time. but layerfetcher doesn't report error when it failed to fetch a block.

https://github.com/spacemeshos/go-spacemesh/blob/398211f24e862300dab8a3414bdcb64ed7a50316/layerfetcher/layers.go#L286

this will cause the node to validate a layer with incomplete data.

## Steps to reproduce
- start a node
- make internet connection unstable 

## Actual Behavior
the node ignores the fetch error and process the layer with incomplete content

## Expected Behavior

the node should fail sync immediately

countvonzero:
## Motivation

Closes #2797

## Changes

double-check at the end of polling layer content and make sure all blocks are fetched

instead of checking individual call to fetchLayerBlocks, we wait until the end of polling layer to make sure all blocks are fetched.
this is because when some fetchLayerBlocks calls fail, others calls may succeed.

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -42,6 +42,7 @@ type layerDB interface {
 	GetLayerInputVectorByID(id types.LayerID) ([]types.BlockID, error)
 	SaveLayerInputVectorByID(ctx context.Context, id types.LayerID, blks []types.BlockID) error
 	ProcessedLayer() types.LayerID
+	GetBlock(types.BlockID) (*types.Block, error)
 }
 
 type atxIDsDB interface {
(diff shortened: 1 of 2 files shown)


countvonzero:
bors try

dshulyak:
what if we send this request (fetchLayerBlocks) to the node that itself is not fully synced and it returns only half of the blocks and empty input vector for example? will we count this layer as synced and move on?

dshulyak:
> what if we send this request (fetchLayerBlocks) to the node that itself is not fully synced and it returns only half of the blocks and empty input vector for example? will we count this layer as synced and move on?

looks like this is a problem only if all peers are in such state

dshulyak (approved):

countvonzero:
> looks like this is a problem only if all peers are in such state

yeah. this is not the most efficient fix. the low level fetcher should take care of retrying hash across the peers (currently 20 times).

let me merge this first, and follow up with an improvement that also fix https://github.com/spacemeshos/go-spacemesh/issues/2796

countvonzero:
@dshulyak also in general we should drop a peer if the peer is less synced than we are when catching up. if you agree, i will open a separate issue

countvonzero:
bors merge

PR #2806, opened 2021-09-21
[Merged by Bors] - Add Prometheus metrics with information about blocks

### LINKED ISSUE #2803: Add metrics
## Description

@dshulyak @countvonzero suggested adding new metrics to make investigating issues like https://github.com/spacemeshos/go-spacemesh/issues/2802 easier.

The list of metrics:
- `layer_distance_to_base_block` - to see how far back a node needs to find a good block
- `avg_layer_block_size` + `avg_num_txs_in_block` - to see abnormal size not caused by num of TXs
- `avg_base_block_exception_length` - to see abnormal diff size
- `layer_num_blocks` - to see how network disagree on layer content
- `block_build_duration`

Also, we need to add a log for `block_build_duration`. If it goes above 2-5s, it should be raised to error to get reports from users.

cc @moshababo

nkryuchkov:
## Motivation

Closes #2803 

## Changes

- Add Prometheus metrics with information about blocks

## Test Plan

Check metrics in Prometheus

## TODO

- [ ] Check metrics in Prometheus

## DevOps Notes


nkryuchkov:
bors try

blocks/metrics/prometheus.go
```diff
@@ -0,0 +1,50 @@
...
+const (
+	subsystem     = "blocks"
+	diffTypeLabel = "diff_type"
+)
+
+// AvgLayerBlockSize checks average block size.
+var AvgLayerBlockSize = metrics.NewHistogram(
+	"avg_layer_block_size",
+	subsystem,
+	"Detect abnormal size not caused by num of TXs",
```
dshulyak: probably a copy paste

blocks/metrics/prometheus.go
```diff
@@ -0,0 +1,50 @@
...
+		"layer_id",
+		"block_id",
+	},
+)
+
+// AvgNumTxsInBlock checks average transaction count in block.
+var AvgNumTxsInBlock = metrics.NewHistogram(
+	"avg_num_txs_in_block",
+	subsystem,
+	"Detect abnormal size not caused by num of TXs",
```
dshulyak: Number of txs in the block?

miner/block_builder.go
```diff
@@ -9,21 +9,27 @@ import (
...
-const defaultFee = 1
+const (
+	defaultGasLimit = 10
+	defaultFee      = 1
+)
 
 // AtxsPerBlockLimit indicates the maximum number of atxs a block can reference
 const AtxsPerBlockLimit = 100
 
+const blockBuildDurationErrorThreshold = 2 * time.Second
```
dshulyak: should be a raised a bit, maybe 10s ?

blocks/blocks.go
```diff
@@ -124,6 +129,43 @@ func (bh *BlockHandler) HandleBlockData(ctx context.Context, data []byte, fetche
...
+		},
+		{
+			hist:  metrics.AvgAgainstDiffLength,
+			value: float64(len(blk.AgainstDiff)),
+		},
+	}
+
+	for _, m := range metricList {
+		m.hist.
+			With("layer_id", blk.LayerIndex.String()).
```
dshulyak: would be better to use constants here and where labels are defined. otherwise this will be prone to typos

miner/block_builder.go
```diff
@@ -280,14 +287,18 @@ func (t *BlockBuilder) createBlockLoop(ctx context.Context) {
 				continue
 			}
 
+			started := time.Now()
+
 			atxID, proofs, atxs, err := t.blockOracle.BlockEligible(layerID)
 			if err != nil {
 				events.ReportDoneCreatingBlock(true, layerID.Uint32(), "failed to check for block eligibility")
+				t.saveBlockBuildDurationMetric(ctx, started, layerID, types.BlockID{})
```
dshulyak: i think we should use this metric only for a successful build. maybe we can time failed builds as well separately, but i don't see why it would be valuable

blocks/metrics/prometheus.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+const (
+	subsystem     = "blocks"
+	diffTypeLabel = "diff_type"
+)
+
+// AvgLayerBlockSize checks average block size.
+var AvgLayerBlockSize = metrics.NewHistogram(
+	"avg_layer_block_size",
```
dshulyak: i think avg_ prefix needs to be removed from metrics. histogram is not about averages, avg prefix will only confuse

blocks/blocks.go
```diff
@@ -124,6 +129,43 @@ func (bh *BlockHandler) HandleBlockData(ctx context.Context, data []byte, fetche
 	return nil
 }
 
+func (bh *BlockHandler) saveMetrics(blk types.Block) {
```
dshulyak: there is no reason to make a method, right?

blocks/metrics/prometheus.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/spacemeshos/go-spacemesh/metrics"
+)
+
+const (
+	subsystem     = "blocks"
+	diffTypeLabel = "diff_type"
+)
+
+// AvgLayerBlockSize checks average block size.
+var AvgLayerBlockSize = metrics.NewHistogram(
```
dshulyak: btw NewHistogram misses buckets configuration. the default buckets are tailored to latency measurements, and they need to be changed for each metric

blocks/metrics/prometheus.go
nkryuchkov: I copied it from a Slack message, which had one description for two fields, but that's not a good idea. Changed

miner/block_builder.go
nkryuchkov: It looks a little bit high but I don't mind. Changed

blocks/metrics/prometheus.go
countvonzero: yep. when i proposed the names, i didn't think to use histogram. i was envisioning having one number (avg) per layer.
histogram of course capture more data is therefore is great. but names need to be changed.

blocks/metrics/prometheus.go
countvonzero: just "block size in bytes"

blocks/blocks.go
nkryuchkov: yes

blocks/metrics/prometheus.go
```diff
@@ -0,0 +1,50 @@
...
+	"avg_layer_block_size",
+	subsystem,
+	"Detect abnormal size not caused by num of TXs",
+	[]string{
+		"layer_id",
+		"block_id",
```
countvonzero: do we need block_id in this and below metrics? i'd imagine one block contribute one data point and we want to see a trend and not necessarily pinpoint individual block. tho this can be nice, i'm concerned about data bloat.

miner/block_builder.go
```diff
@@ -329,8 +345,22 @@ func (t *BlockBuilder) createBlockLoop(ctx context.Context) {
 						logger.WithContext(blockCtx).With().Error("failed to send block", log.Err(err))
 					}
 					events.ReportDoneCreatingBlock(true, layerID.Uint32(), "")
+					t.saveBlockBuildDurationMetric(ctx, started, layerID, blk.ID())
```
countvonzero: i don't understand why we are saving this metric in so many places. i think this should be the only place that we report block build duration.

syncer/syncer.go
```diff
@@ -376,6 +377,11 @@ func (s *Syncer) syncLayer(ctx context.Context, layerID types.LayerID) (*types.L
 		if layer, err = s.getLayerFromPeers(ctx, layerID); err != nil {
 			return nil, err
 		}
+
+		metrics.LayerNumBlocks.
+			With("layer_id", layerID.String()).
+			Observe(float64(len(layer.Blocks())))
```
countvonzero: this may not be the best place to save this metrics. i think mesh/meshdb is better. for nodes that are synced, this code path is not being exercised anymore.

tortoise/metrics/prometheus.go
```diff
@@ -0,0 +1,23 @@
+package metrics
+
+import (
+	"github.com/spacemeshos/go-spacemesh/metrics"
+)
+
+const (
+	// Subsystem is a subsystem shared by all metrics exposed by this package.
+	Subsystem = "tortoise"
```
countvonzero, line 10: i am not really sure i appreciate this Subsystem thing... if i am looking for a metrics, do i need to query by subsystem? i will have no clue/memory of which subsystem a metrics belong to.

tortoise/verifying_tortoise.go
```diff
@@ -362,6 +371,13 @@ func (t *turtle) BaseBlock(ctx context.Context) (types.BlockID, [][]types.BlockI
 				log.Int("support_count", len(exceptionVectorMap[1])),
 				log.Int("neutral_count", len(exceptionVectorMap[2])))
 
+			metrics.LayerDistanceToBaseBlock.
+				With("last_layer", t.Last.String()).
+				With("last_evicted", t.LastEvicted.String()).
+				With("base_block_layer", layerID.String()).
+				With("block_id", blockID.String()).
```
countvonzero, line 376: are these data last_layer/last_evicted/base_block_layer/block_id really going to be useful?

blocks/metrics/prometheus.go
nkryuchkov: I thought it could let us do a further search of block ID in logs to know what could go wrong

miner/block_builder.go
nkryuchkov: Do we need to split the metric into two: with and without network broadcast time?

countvonzero:
aslo, @nkryuchkov can you add tests to verify that these metrics are added correctly? thanks

blocks/metrics/prometheus.go
countvonzero: i understand the advantage, but i worry that this is adding a lot of data for very little benefit. we can quite easily find the block based on the timestamp of the metrics? in any case, let's leave it for now.

miner/block_builder.go
dshulyak: we just need to collect it before sending to the network

dshulyak:
@countvonzero but what tests do you envision? usually, metrics are the same as logs, they use standard API and are just assumed to work. for tests we will need to mock metrics or setup a server for each test. i would rather avoid tests for metrics

tortoise/metrics/prometheus.go
nkryuchkov: What would you suggest to use instead? Having the same subsystem for all metrics?

tortoise/verifying_tortoise.go
nkryuchkov: I think that `last_layer` and `block_id` may be useful to find a source of an issue. The rest may be removed

tortoise/metrics/prometheus.go
countvonzero: do you know the purpose of subsystem? is it used for querying the metrics? 
anyhow, not important. we can fix things if there is UX issues

nkryuchkov:
@countvonzero I tend to agree with @dshulyak. Metrics could be tested but it would be too complex in my opinion

countvonzero:
> @countvonzero but what tests do you envision? usually, metrics are the same as logs, they use standard API and are just assumed to work. for tests we will need to mock metrics or setup a server for each test. i would rather avoid tests for metrics

@dshulyak i don't want to test that they are recorded correctly, but i want to test that metrics is called as expected as part of the app logic. for example:
- after a block is produced, we expect metrics A, B, C to be recorded
- after hare outputs a layer, we expect metrics E, F, G to be recorded

a mock for metrics will be perfect in this case

blocks/metrics/prometheus.go
```diff
@@ -0,0 +1,56 @@
...
+		BlockIDLabel,
+		diffTypeLabel,
+	},
+)
+
+// Blocks diff lengths.
+var (
+	ForDiffLength     = baseBlockExceptionLength.With(diffTypeLabel, "for_diff")
+	AgainstDiffLength = baseBlockExceptionLength.With(diffTypeLabel, "against_diff")
+	NeutralDiffLength = baseBlockExceptionLength.With(diffTypeLabel, "neutral_diff")
```
countvonzero: nit: when searching for labels, it may be easier if they all have the same prefix. e.g. "diff_for", "diff_against", "diff_neutral"

mesh/metrics/prometheus.go
```diff
@@ -0,0 +1,25 @@
...
+// LayerIDLabel is metric label for layer ID.
+const (
+	LayerIDLabel = "layer_id"
+)
+
+// LayerNumBlocks checks how network disagrees on layer content.
+var LayerNumBlocks = metrics.NewIntegerHistogram(
+	"layer_num_blocks",
+	Subsystem,
+	"How network disagrees on layer content",
```
countvonzero: number of blocks in a layer

countvonzero (approved):

metrics/common.go
```diff
@@ -40,6 +40,18 @@ func NewHistogram(name, subsystem, help string, labels []string) Histogram {
 	return prmkit.NewHistogramFrom(prometheus.HistogramOpts{Namespace: Namespace, Subsystem: subsystem, Name: name, Help: help}, labels)
 }
 
+// NewIntegerHistogram creates a Histogram metrics for integer numbers.
+func NewIntegerHistogram(name, subsystem, help string, labels []string) Histogram {
+	buckets := []float64{0, 1, 2, 5, 10, 20, 50, 100, 500}
```
dshulyak: i am not sure that generic integer histogram makes sense. buckets should reflect actual values, sizes of the blocks and number of txs will have very different values, right? i suggest to define buckets per each metric (or metrics type, e.g extensions) and use prmkit.LinearBucket and prmkit.ExponentialBucket (whichever is more suitable).

countvonzero:
> @countvonzero I tend to agree with @dshulyak. Metrics could be tested but it would be too complex in my opinion

you misunderstand me. please see my reply above. i don't mean to test the metrics system itself. i mean to test the timing of what metrics are expected at what events in the app

dshulyak:
> you misunderstand me. please see my reply above. i don't mean to test the metrics system itself. i mean to test the timing of what metrics are expected at what events in the app

i am still confused how implementation will look like. does it imply that we will need to mock metrics, so that we check later if they were called? it just seems too over-complicated to write tests for things like that. i think for the same reason why are not writing tests for logs

countvonzero:
> > you misunderstand me. please see my reply above. i don't mean to test the metrics system itself. i mean to test the timing of what metrics are expected at what events in the app
> 
> i am still confused how implementation will look like. does it imply that we will need to mock metrics, so that we check later if they were called? it just seems too over-complicated to write tests for things like that. i think for the same reason why are not writing tests for logs

in my old projects, we made sure some key metrics are recorded via tests. at that time the metrics system was part of the app framework so it was easy to verify. 

if you think mocking out the underlying metrics in this case is too complicated. i can live with that too.

tortoise/metrics/prometheus.go
nkryuchkov: As I understand, it's needed for optional grouping and may be left blank

metrics/common.go
nkryuchkov: Something like [this](https://github.com/spacemeshos/go-spacemesh/pull/2806/commits/ff06fb007410b521c8de6354f42d3bf717ae86df)?

nkryuchkov:
@countvonzero I think that your suggestion may help with testing, but in this case, we need to refactor all usages of metrics that we have now, which would postpone the devnet launch

countvonzero:
> @countvonzero I think that your suggestion may help with testing, but in this case, we need to refactor all usages of metrics that we have now, which would postpone the devnet launch

to be clear, i wanted to push the evelope a little bit so i asked for adding test. i could be persuaded for the reason that the metrics are not critical enough to put in that extra effort.

bur if we are talking about other more critical app logic, citing delaying devnet launch to forego testing isnt really convincing to me. oftentimes these deadlines are arbitrary and self-imposed. a little more testing as upfront investment save self/others precious time later on.

anyway, you can see i am ranting now. please proceed to merge. thanks

dshulyak (approved):

countvonzero (approved):

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #2815, opened 2021-09-22
[Merged by Bors] - tortoise: compute threshold without float

### LINKED ISSUE #2629: Fix vector and threshold arithmetic in verifying tortoise

dshulyak:
## Motivation

Computes threshold in tortoise without using float.

Closes https://github.com/spacemeshos/go-spacemesh/issues/2629

## TODO

## DevOps Notes


tortoise/util.go
```diff
@@ -81,26 +73,24 @@ func (a vec) String() string {
 	return "abstain"
 }
 
-func calculateOpinionWithThreshold(logger log.Log, v vec, layerSize int, theta uint8, delta float64) vec {
-	threshold := float64(theta) / 100 * delta * float64(layerSize)
-	netVote := float64(v.netVote())
+func calculateOpinionWithThreshold(logger log.Log, v vec, layerSize int, theta uint8, delta uint32) vec {
+	threshold := big.NewFloat(float64(theta))
+	threshold = threshold.Quo(threshold, big.NewFloat(100))
+	threshold = threshold.Mul(threshold, big.NewFloat(float64(delta)*float64(layerSize)))
```
nkryuchkov: Wouldn't this be more precise because of avoiding `float64` multiplication?

```suggestion
	threshold = threshold.Mul(threshold, big.NewFloat(float64(delta)))
	threshold = threshold.Mul(threshold, big.NewFloat(float64(layerSize)))
```

nkryuchkov (approved):

tortoise/util.go
dshulyak: delta and layerSize are integers, in what case there will be issues with precision?

nkryuchkov: Sorry, I looked at the previous signature of `calculateOpinionWithThreshold`. Never mind

nkryuchkov: Although, I think the precision might be lost just a little bit in this case because of converting to `float64`. `float64(delta)` would become some floating-point value around `delta`, then it would be multiplied by something around `layerSize` and the result would be something around `delta*layerSize` (same for `theta` and `100`). `new(big.Float).SetUint64(uint64(delta))` should avoid losing precision.

nkryuchkov: I wrote a test for this, and it shows that I'm wrong
```
func Test_Precision(t *testing.T) {
	var (
		theta     uint8  = 2
		layerSize int    = 5
		delta     uint32 = 10
	)

	t1 := big.NewFloat(float64(theta))
	t1 = t1.Quo(t1, big.NewFloat(100))
	t1 = t1.Mul(t1, big.NewFloat(float64(delta)*float64(layerSize)))

	t2 := new(big.Float).SetUint64(uint64(theta))
	t2 = t2.Quo(t2, new(big.Float).SetUint64(100))
	t2 = t2.Mul(t2, new(big.Float).SetUint64(uint64(delta)))
	t2 = t2.Mul(t2, new(big.Float).SetUint64(uint64(layerSize)))

	t.Log(t1)
	t.Log(t2)
	require.Equal(t, t1.Cmp(t2), 0)
}
```
Output:
```
=== RUN   Test_Precision
    util_test.go:26: 1
    util_test.go:27: 0.99999999999999999995
    util_test.go:28: 
        	Error Trace:	util_test.go:28
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	Test_Precision
```

tortoise/util.go
```diff
@@ -81,26 +75,28 @@ func (a vec) String() string {
 	return "abstain"
 }
 
-func calculateOpinionWithThreshold(logger log.Log, v vec, layerSize int, theta uint8, delta float64) vec {
-	threshold := float64(theta) / 100 * delta * float64(layerSize)
-	netVote := float64(v.netVote())
+func calculateOpinionWithThreshold(logger log.Log, v vec, layerSize int, theta uint8, delta uint32) vec {
+	threshold := big.NewInt(int64(theta))
+	threshold.Mul(threshold, big.NewInt(int64(delta)))
```
countvonzero: why not use `new(big.Int).SetUint64(delta)`? 

i am a bit confused by the choice of uint32/int64/int/uint8 here and in the method signature.... they all should be unsigned numbers...

tortoise/util.go
```diff
@@ -81,26 +75,28 @@ func (a vec) String() string {
...
 		log.Int("layer_size", layerSize),
-		log.String("delta", fmt.Sprint(delta)),
-		log.String("threshold", fmt.Sprint(threshold)),
-		log.String("net_vote", fmt.Sprint(netVote)))
-	if netVote > threshold {
-		// try net positive vote
+		log.Uint32("delta", delta),
+		log.String("threshold", threshold.String()))
+
+	greater := func(val uint64) bool {
```
countvonzero: how about `exceedsThreshold` instead of `greater`?

tortoise/util.go
```diff
@@ -81,26 +75,28 @@ func (a vec) String() string {
 	return "abstain"
 }
 
-func calculateOpinionWithThreshold(logger log.Log, v vec, layerSize int, theta uint8, delta float64) vec {
```
countvonzero, line 84: it will help the readers of this function tremendously if you can add comments about what this method is calculating/doing

tortoise/util.go
```diff
@@ -81,26 +75,28 @@ func (a vec) String() string {
...
-	threshold := float64(theta) / 100 * delta * float64(layerSize)
-	netVote := float64(v.netVote())
+func calculateOpinionWithThreshold(logger log.Log, v vec, layerSize int, theta uint8, delta uint32) vec {
+	threshold := big.NewInt(int64(theta))
+	threshold.Mul(threshold, big.NewInt(int64(delta)))
+	threshold.Mul(threshold, big.NewInt(int64(layerSize)))
```
countvonzero: nit: can you chain this?
`threshold.Mul(threshold, big.NewInt(int64(delta))).Mul(threshold, big.NewInt(int64(layerSize)))`

countvonzero (approved):
in the most recent R&D meeting where Tal requested us to make threshold as a fraction and not as a percentage. 

he also requested to used a fixed64 library instead. tho i am not sure it's still apply now that you do everything in big.Int. 

in any case, the change is better than the current code.

dshulyak:
bors try

dshulyak:
bors try

tortoise/util.go
dshulyak: delta and layerSize are uint32 (thought layerSize is casted to int in the upper). i am casting to int64 because big.NewInt accepts int64 in the initializer, and since the value always fits in 63 bits this is not a problem. SetUint64 is also correct but there is literally not difference

dshulyak:
> he also requested to used a fixed64 library instead. tho i am not sure it's still apply now that you do everything in big.Int.

i think the concern was about the consistency of the representation and lack of precision. e.g. this theta fraction could be 125/1000 or 1/10000 and we can't support that with uint8.

dshulyak:
actually, i think it is good idea to use big.Rat. 

> A Rat represents a quotient a/b of arbitrary precision

it makes sense to store it independently as Rat and then convert to required type that will be used in computation (either fixed float or integer depending on the approach)

nkryuchkov:
@dshulyak I've chosen `big.Rat` for `theta` in tortoise beacon, it's really convenient and works fine

dshulyak:
bors merge

lrettig:
Thanks!

PR #2819, opened 2021-09-23
[Merged by Bors] - tortoise beacon: prevent bad msgs being gossiped and add tests

### LINKED ISSUE #2817: tortoise beacon: some hidden error may cause incorrect gossip
## Description

in tortoisebeacon/handlers.go, code like here
https://github.com/spacemeshos/go-spacemesh/blob/ba3f6d75facc0992932066c9ac15de566ddfbdd8/tortoisebeacon/handlers.go#L104
here
https://github.com/spacemeshos/go-spacemesh/blob/ba3f6d75facc0992932066c9ac15de566ddfbdd8/tortoisebeacon/handlers.go#L269
and here
https://github.com/spacemeshos/go-spacemesh/blob/ba3f6d75facc0992932066c9ac15de566ddfbdd8/tortoisebeacon/handlers.go#L399

will cause the node to gossip bad messages in the network

## Actual Behavior

the code hides verification errors and report an invalid message as valid back to p2p gossip.

## Expected Behavior

should propagate error and not end up gossiping a invalid message

countvonzero:
## Motivation

Closes #2817

## Changes

- for messages that fail verification, propagate errors and don't gossip about them
- add test for various error conditions when handling gossip messages
- rewrite hander_test.go to separate the test cases

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/tortoisebeacon/handlers.go
+++ b/tortoisebeacon/handlers.go
@@ -24,17 +24,20 @@ const TBFirstVotingProtocol = "TBFirstVotingGossip"
 const TBFollowingVotingProtocol = "TBFollowingVotingGossip"
 
 var (
-	// ErrVRFNotVerified is returned if proposal message fails VRF verification.
-	ErrVRFNotVerified = errors.New("proposal failed vrf verification")
+	// errVRFNotVerified is returned if proposal message fails VRF verification.
+	errVRFNotVerified = errors.New("proposal failed vrf verification")
 
-	// ErrProposalDoesntPassThreshold is returned if proposal message doesn't pass threshold.
-	ErrProposalDoesntPassThreshold = errors.New("proposal doesn't pass threshold")
+	// errProposalDoesntPassThreshold is returned if proposal message doesn't pass threshold.
+	errProposalDoesntPassThreshold = errors.New("proposal doesn't pass threshold")
 
-	// ErrAlreadyProposed is returned when miner doubly proposed for the same epoch.
-	ErrAlreadyProposed = errors.New("already proposed")
+	// errAlreadyProposed is returned when miner doubly proposed for the same epoch.
+	errAlreadyProposed = errors.New("already proposed")
 
-	// ErrAlreadyVoted is returned when miner doubly voted for the same epoch/round.
-	ErrAlreadyVoted = errors.New("already voted")
+	// errAlreadyVoted is returned when miner doubly voted for the same epoch/round.
+	errAlreadyVoted = errors.New("already voted")
+
+	// errMinerATXNotFound is returned when miner has no ATXs in the previous epoch.
+	errMinerATXNotFound = errors.New("miner ATX not found in previous epoch")
 )
 
 // HandleSerializedProposalMessage defines method to handle Tortoise Beacon proposal Messages from gossip.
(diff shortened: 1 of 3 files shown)


tortoisebeacon/handlers.go
```diff
@@ -101,37 +103,37 @@ func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalM
...
 }
 
 func (tb *TortoiseBeacon) classifyProposalMessage(ctx context.Context, m ProposalMessage, atxID types.ATXID, currentEpoch types.EpochID, receivedTime time.Time) error {
+	minerID := m.NodeID.ShortString()
+	logger := tb.logger.WithContext(ctx).WithFields(currentEpoch, log.String("miner_id", minerID))
+
 	atxHeader, err := tb.atxDB.GetAtxHeader(atxID)
 	if err != nil {
-		return fmt.Errorf("failed to get ATXID %v header: %w", atxID, err)
+		logger.Error("[proposal] failed to get ATX header", atxID, log.Err(err))
```
dshulyak, line 123: isn't it a warning?

tortoisebeacon/handlers.go
```diff
@@ -101,37 +103,37 @@ func (tb *TortoiseBeacon) handleProposalMessage(ctx context.Context, m ProposalM
...
+		return fmt.Errorf("[proposal] failed to get ATX header (miner ID %v, ATX ID %v): %w", minerID, atxID, err)
 	}
 
 	atxTimestamp, err := tb.atxDB.GetAtxTimestamp(atxID)
 	if err != nil {
-		return fmt.Errorf("failed to get ATXID %v timestamp: %w", atxID, err)
+		logger.Error("[proposal] failed to get ATX timestamp", atxID, log.Err(err))
```
dshulyak, line 129: same

tortoisebeacon/handlers.go
```diff
@@ -175,41 +177,41 @@ func (tb *TortoiseBeacon) addPotentiallyValidProposal(proposal []byte) {
...
-	logger := tb.logger.WithContext(ctx).WithFields(currentEpoch, log.String("miner_id", m.NodeID.ShortString()))
-	currentEpochProposal := buildProposal(currentEpoch, logger)
+	minerID := m.NodeID.ShortString()
+	logger := tb.logger.WithContext(ctx).WithFields(currentEpoch, log.String("miner_id", minerID))
+
 	atxID, err := tb.atxDB.GetNodeAtxIDForEpoch(m.NodeID, currentEpoch-1)
 	if errors.Is(err, database.ErrNotFound) {
-		logger.Warning("miner has no atxs in the previous epoch")
-		return types.ATXID{}, database.ErrNotFound
+		logger.Warning("[proposal] miner has no ATX in previous epoch")
```
dshulyak, line 185: :+1:

dshulyak (approved):
i think there are some Error logs that should be Warnings instead. e.g. when proposal doesn't pass or something externally received is not found in db

tortoisebeacon/handlers.go
countvonzero: but this is supposed to be programming error. we have the ATX ID in database, but not the header?

tortoisebeacon/handlers.go
countvonzero: same here. if we didn't save ATX timestamp, that's a programming error

tortoisebeacon/handlers.go
countvonzero: i just hesitate to panic here as it's not sever enough to stop the server from working

dshulyak: sorry, i missed this. the actual atx is fetched during verification? if so everything makes sense

countvonzero: yes. at this point we already confirmed that the nodeID has ATX in the previous epoch

countvonzero:
bors merge

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/3702720605)

an old issue https://spacemesh.slack.com/archives/C01BBB2U64U/p1630336484017700
quoting @AmitShaul in the slack channel
```
it’s not a timezone issue, it seems that the nodes took too long to start and the preset genesis delta already expired
```

countvonzero:
bors merge

PR #2825, opened 2021-09-24
[Merged by Bors] - Reduce amount of labels in metrics

nkryuchkov:
## Motivation

Queries for some of metrics time out, likely because of an excessive amount of labels. This PR removes less important labels

## Changes

- Reduce amount of labels in metrics

## Test Plan

Unit tests

## DevOps Notes


blocks/blocks.go
```diff
@@ -148,20 +148,11 @@ func saveMetrics(blk types.Block) {
...
-		},
-		{
-			hist:  metrics.AgainstDiffLength,
-			value: float64(len(blk.AgainstDiff)),
-		},
 	}
 
 	for _, m := range metricList {
 		m.hist.
 			With(metrics.LayerIDLabel, blk.LayerIndex.String()).
```
dshulyak: after some sought i am not sure that we need layer. why it would be useful to have a hist per layer? i think metrics should give a high level view into the network and system state. and if we need to find details about a particular blocks we can add those details to logs

blocks/blocks.go
```diff
@@ -148,20 +148,11 @@ func saveMetrics(blk types.Block) {
 			hist:  metrics.ForDiffLength,
 			value: float64(len(blk.ForDiff)),
 		},
-		{
```
dshulyak: why to remove? this is only 2 additional labels and they don't change based on the network state

countvonzero:
i am inclined to remove all the layer labels. we can deduce that from the time when the metrics is recorded.

blocks/blocks.go
nkryuchkov: I agree, removed

blocks/blocks.go
nkryuchkov: I just wanted to remove metrics that are not important at the moment. But we can keep these two. Reverted

nkryuchkov:
@countvonzero I agree. Done

countvonzero (approved):

nkryuchkov:
bors merge

PR #2835, opened 2021-09-26
[Merged by Bors] - Transition `state` package into `svm` package

### LINKED ISSUE #2827: Transition `state` package into `svm` package
- [x] Create `svm` package
- [x] Move `state` package into `svm` package
- [x] Breakout `mempool` into separate package (this is currently the `tx_pool.go` file and matching test)

noamnelke:
## Motivation
Closes #2827

## Changes
Transition `state` package into `svm` package. This is a refactor that should not impact existing functionality.

## Test Plan
Existing unit and system tests.

## DevOps Notes


noamnelke:
bors try

noamnelke:
bors try

dshulyak (approved):
won't it be cleaner to have a `vm` module instead of `svm`? i guess it doesn't matter much, but  `svm` is a vm that implements a certain interface and there could be many different implementations

noamnelke:
@dshulyak I agree in general, but not on the specifics... As I see it, `SVM` is what I call the interface, while `rust-SVM` is a specific implementation of it. We may write a `C-SVM` one day, but it will also be an implementation of `SVM`. If we strip the `s` it will be just a virtual machine, which is too generic here, IMO.

The `svm` package that I create here is just a wrapper around the actual SVM, which will be the old implementation at first and then replaced with `rust-SVM` (inside the wrapper, without changing its exposed methods).

Do you agree?

noamnelke:
bors merge

neysofu (approved):
LGTM!

dshulyak:
> I agree in general, but not on the specifics... As I see it, SVM is what I call the interface, while rust-SVM is a specific implementation of it. We may write a C-SVM one day, but it will also be an implementation of SVM. If we strip the s it will be just a virtual machine, which is too generic here, IMO.

i guess this a valid view. i usually look at the whole module name, in this case `spacemeshos/go-spacemesh/vm` doesn't seem generic and not very easy to misunderstand the intention

noamnelke:
@dshulyak Your last point convinced me. I looked it up and it looks like the Ethereum devs also agree with you - in geth the EVM package is just called `vm`. We can rename it in a following PR.

PR #2836, opened 2021-09-26
Separate tortoise config

### LINKED ISSUE #2707: Separate tortoise config
Right now, tortoise config lives inside the main app config, and is assembled by the App before the verifying tortoise is instantiated:

https://github.com/spacemeshos/go-spacemesh/blob/2d87075870e3057958517f3b3544fa6138b9ea19/cmd/node/node.go#L624-L638

The tortoise should have its own config module. `TortoiseRerunInterval` should be switched from an int to time.Duration so that it can be specified as, e.g., "12h" in the config.

sudachen:
## Motivation
The tortoise should have its own config module

Closes #2707

## Changes
Tortoise config is separated to the own configuration module

## DevOps Notes
It may require to change configuration

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -626,20 +626,14 @@ func (app *App) initServices(ctx context.Context,
 		return err
 	}
 	app.closers = append(app.closers, trtlStateDB)
-	trtlCfg := tortoise.Config{
-		LayerSize:       layerSize,
-		Database:        trtlStateDB,
-		MeshDatabase:    mdb,
-		ATXDB:           atxDB,
-		Clock:           clock,
-		Hdist:           app.Config.Hdist,
-		Zdist:           app.Config.Zdist,
-		ConfidenceParam: app.Config.ConfidenceParam,
-		WindowSize:      app.Config.WindowSize,
-		GlobalThreshold: app.Config.GlobalThreshold,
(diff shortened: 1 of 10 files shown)


cmd/node/node.go
```diff
@@ -626,20 +626,14 @@ func (app *App) initServices(ctx context.Context,
...
-		Clock:           clock,
-		Hdist:           app.Config.Hdist,
-		Zdist:           app.Config.Zdist,
-		ConfidenceParam: app.Config.ConfidenceParam,
-		WindowSize:      app.Config.WindowSize,
-		GlobalThreshold: app.Config.GlobalThreshold,
-		LocalThreshold:  app.Config.LocalThreshold,
-		Log:             app.addLogger(TrtlLogger, lg),
-		RerunInterval:   time.Minute * time.Duration(app.Config.TortoiseRerunInterval),
+	trtlCfg := tortoise.Configuration{
```
dshulyak, line 629: i think that having both Configuration and Config adds more technical debt. would be nicer if tortoise initializer was refactored to accept non-optional dependencies (atxdb, meshdb, etc) as parameters to function and keep only single Config for immutable parameters

dshulyak:
will be obsoleted by https://github.com/spacemeshos/go-spacemesh/pull/2959

PR #2848, opened 2021-09-29
[Merged by Bors] - `cmd/node`: Move `setupGenesis` logic into `svm.SetupGenesis`

### LINKED ISSUE #2830: Implement `SetupGenesis` in SVM package
```go
func (s *SVM) SetupGenesis(...)
```

Should perform all operations required to prepare a new empty instance of SVM/state. In particular, this should create and func the genesis accounts based on the information provided in the config.

### Called From
- `cmd/node/node.go` -> `App.initServices`

### Existing Functions
- `node.go` -> `App.setupGenesis`
- `TransactionProcessor.DB.AddBalance`
- `TransactionProcessor.DB.Commit`
- `TransactionProcessor.GetStateRoot`

### SVM Functions
- `IncreaseBalance`
- `Commit`

neysofu:
Closes #2830. I'm available for iterating based on any feedback you all might have.

## Changes
Moves all genesis setup logic from `cmd/node` into `svm`.

## DevOps Notes

TODO's, for future PRs:
- Hide `state.TransactionProcessor` from `svm.New` public API.

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/spacemeshos/go-spacemesh/svm/state"
 	"io/ioutil"
 	"math/big"
 	"net/http"
@@ -18,6 +17,9 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/spacemeshos/go-spacemesh/svm"
+	"github.com/spacemeshos/go-spacemesh/svm/state"
+
 	"github.com/mitchellh/mapstructure"
 	"github.com/pyroscope-io/pyroscope/pkg/agent/profiler"
 	"github.com/spf13/cobra"
@@ -27,7 +29,6 @@ import (
 
 	"github.com/spacemeshos/go-spacemesh/activation"
 	"github.com/spacemeshos/go-spacemesh/api"
-	apiCfg "github.com/spacemeshos/go-spacemesh/api/config"
 	"github.com/spacemeshos/go-spacemesh/api/grpcserver"
 	"github.com/spacemeshos/go-spacemesh/blocks"
 	cmdp "github.com/spacemeshos/go-spacemesh/cmd"
@@ -98,6 +99,7 @@ const (
 	Fetcher              = "fetcher"
 	LayerFetcher         = "layerFetcher"
 	TimeSyncLogger       = "timesync"
+	SVMLogger            = "SVM"
 )
 
 // Cmd is the cobra wrapper for the node, that allows adding parameters to it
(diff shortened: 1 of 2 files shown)


cmd/node/node.go
```diff
@@ -650,7 +624,9 @@ func (app *App) initServices(ctx context.Context,
 		go msh.CacheWarmUp(app.Config.LayerAvgSize)
 	} else {
 		msh = mesh.NewMesh(mdb, atxDB, app.Config.REWARD, trtl, app.txPool, processor, app.addLogger(MeshLogger, lg))
-		if err := app.setupGenesis(processor, msh); err != nil {
+		svm := svm.New(app.state, app.addLogger(SVMLogger, lg))
```
noamnelke: While the `SVM` instance is still only used to manipulate the `TransactionProcessor` that it receives - I would prefer to already persist it as part of the `App` and therefore instantiate it before this `if`.

Note that you use `app.state` here even though it's not initialized until [later in this method](https://github.com/spacemeshos/go-spacemesh/pull/2848/files#diff-69bfff7b16ff2da1ebcca3ff8dcd296aa906d17d4af33acac1b959329fcd95b0R751). You can use `processor` which is the value eventually set to `app.state`.

noamnelke (changes requested):

cmd/node/node.go
neysofu: Makes sense to persist it, won't me of much use otherwise :) I've pushed some changes also replacing `app.state` with `processor`.

svm/api.go
```diff
@@ -0,0 +1,51 @@
...
+// New creates a new `SVM` instance from the given `state` and `logger`.
+func New(state *state.TransactionProcessor, logger log.Log) *SVM {
+	return &SVM{state, log.NewDefault("svm")}
+}
+
+// SetupGenesis creates new accounts and adds balances as dictated by `conf`.
+func (svm *SVM) SetupGenesis(conf *config.GenesisConfig) error {
+	if conf == nil {
+		conf = config.DefaultGenesisConfig()
+	}
```
noamnelke, line 28: This will replace the pointer, not what it points to - this check needs to be done somewhere else or differently somehow.

noamnelke (changes requested):

svm/api.go
neysofu: I'm not sure I'm following. In case the pointer is null, there's no value to replace, right?

- Receive a pointer to a `config.GenesisConfig`.
- If invalid (`nil`), we point instead to the default configuration options.
- By now, our pointer will point to either the given configuration options or the default ones.
- We're ready to read the value pointed by `config` and use it for initialization.

noamnelke: It's fine for this local function - the value of `conf` will be correct. However, the previous implementation updated the global configuration (`app.Config.Genesis`) to the default if it's `nil`.

I'm not sure this will have any impact, as this may not be read anywhere other than in this function. If that's the case (please make sure) you can leave it the way it is now.

noamnelke: I checked and we don't access `.Genesis` anywhere other than this function - so you can leave this as-is.

noamnelke (approved):

svm/api.go
neysofu: Okay, I didn't realize the old version updated the value as well, I guess that's where my confusion was.

neysofu:
bors try

neysofu:
bors merge

PR #2850, opened 2021-09-29
[Merged by Bors] - beacon sync: wait for N blocks before deciding beacon value from blocks

### LINKED ISSUE #2810: beacon sync: determine beacon value from blocks in the first K layers
## Description

when a node didn't participate in the tortoise beacon protocol in the last epoch (and therefore didn't learn the beacon value for the current epoch), but is eligible to produce a block in the current epoch, it needs to sync the beacon value from peers.

it is important to find the correct beacon value as it affect the rewards from generating blocks and hare participation. we wait for enough blocks from the beginning of the epoch, and use the most weighted beacon value from them.

the number of blocks should give us enough confidence based on the protocol. here we choose K layers where K constitute 2 clusters of 800 blocks.

countvonzero:
## Motivation

Closes #2810

## Changes

- for nodes that did not participate in the tortoise beacon in the previous epoch, wait for N blocks to decide the beacon value for the previous epoch
- node: wait for beacon value to be determined for previous epoch before participating in hare
- miner: wait for beacon value to be determined for previous epoch before producing blocks

## Test Plan

make the late-node test wait one more epoch due to the nature of beacon sync
- before:
late node participates in hare as soon as it is synced with a fake beacon value (epoch ID)
i.e. a late node will join at epoch 4, become synced at epoch 6 and consistently participates in hare during epoch 7.

- after:
late node only participates in hare once it learned of the beacon value from previous epoch. so it will only consistently participate in hare the epoch after it take parts in the tortoise beacon.
i.e. a late node will join at epoch 4, become synced and published an ATX in epoch 6, and participates in beacon protocol in epoch 7. so it only consistently participates in hare in epoch 8. 
after it gets synced, it will participate in hare in the 2nd and 3rd layer of the epoch 6/7 (after it syncs the beacon value during the 1st layer of epoch 6/7)

## DevOps Notes


countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/3750271469)

https://github.com/spacemeshos/go-spacemesh/runs/3750591105

unit tests was canceled. retry the jobs and all the system tests passed

cmd/root.go
```diff
@@ -205,6 +205,8 @@ func AddCommands(cmd *cobra.Command) {
 		"Ratio of votes for reaching consensus")
 	cmd.PersistentFlags().Uint64Var(&config.TortoiseBeacon.VotesLimit, "tortoise-beacon-votes-limit",
 		config.TortoiseBeacon.VotesLimit, "Maximum allowed number of votes to be sent")
+	cmd.PersistentFlags().Uint32Var(&config.TortoiseBeacon.BeaconSyncNumBlocks, "tortoise-beacon-sync-num-blocks",
```
dshulyak, line 208: i am not sure that this is supposed to be configured separately. shouldn't it be based on the layer size and/or number of layers in epoch?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -193,52 +191,121 @@ func (tb *TortoiseBeacon) Close() {
...
+	if _, err := tb.GetBeacon(epoch); err == nil {
+		// already has beacon. i.e. we had participated in tortoise beacon protocol during last epoch
+		return
 	}
 
-	if tb.tortoiseBeaconDB != nil {
-		val, err := tb.tortoiseBeaconDB.GetTortoiseBeacon(epochID - 1)
-		if err == nil {
-			return val.Bytes(), nil
+	if epochBeacon := tb.findMostWeightedBeaconBeaconForEpoch(epoch); epochBeacon != nil {
```
dshulyak: BeaconBeacon?

dshulyak (commented):
great approach, very intuitive.

in case if node skipped participation in tortoise beacon when it wasn't synced at the start of epoch, it just needs to wait for a sufficient number of blocks before enabling all depending protocols and it doesn't matter where it will get those blocks from (sync or gossip). right?

i am bit concerned that db was removed, if node was restarted right after it got a beacon it will have to wait for many blocks again. if the beacon was persisted it will need only to wait for 2 gossip layers, or am i missing something?

hare/eligibility/beacon_test.go
```diff
@@ -18,40 +18,12 @@ func (mbp mockBeaconProvider) GetBeacon(types.EpochID) ([]byte, error) {
...
-
 	beaconValue := []byte{1, 2, 3, 4}
-	b.beaconGetter = &mockBeaconProvider{beaconValue}
-	b.confidenceParam = cfg.ConfidenceParam
-	val, err := b.Value(context.TODO(), 100)
-	r.NoError(err)
-	r.Equal(binary.LittleEndian.Uint32(beaconValue), val)
-	r.Equal(1, c.numGet)
-	r.Equal(1, c.numAdd)
+	encodedValue := uint32(67305985) // binary.LittleEndian.Uint32(beaconValue)
```
nkryuchkov: I think hex number would be more readable here
```suggestion
	encodedValue := uint32(0x4030201) // binary.LittleEndian.Uint32(beaconValue)
```

nkryuchkov (approved):
Great job! The code became much cleaner

countvonzero:
> in case if node skipped participation in tortoise beacon when it wasn't synced at the start of epoch, it just needs to wait for a sufficient number of blocks before enabling all depending protocols and it doesn't matter where it will get those blocks from (sync or gossip). right?

correct. the source of the data (sync/gossip) does not matter. however, each miner only records the beacon in their **first** block in the epoch. so i am thinking the blocks may be concentrated at the beginning of the epoch.

before i only recorded the block that actually has beacon value. it's trivial to record it from every block. i have changed the code to do so (see the last commit)
 
> i am bit concerned that db was removed, if node was restarted right after it got a beacon it will have to wait for many blocks again. if the beacon was persisted it will need only to wait for 2 gossip layers, or am i missing something?

you are right. i simply was willing to let the node wait longer. added a private db inside tortoise beacon to speed up the recovery.

hare/eligibility/beacon_test.go
countvonzero: you are right. done.

tortoisebeacon/tortoise_beacon.go
countvonzero: i was maybe hungry and thinking about bacon when i wrote this. 🥓

cmd/root.go
countvonzero: according to Tal during the R&D meeting, he was not sure whether we should set it to 1, 1.5 or 2 clusters of (800) blocks, as 800 is the security param. it's likely we will have to experiment with this number. i.e. either set it to 1X/1.5X/2X or set the number of blocks.

also, for system testing, i am not sure whether 800 gets reduced to a smaller number. so it's also easier to explicitly set it to a small number, disregarding layer_size and num_layers_per_epoch.

i imagine when we get to a stage where we are sure of the network param, we can actually simplify a lot more.

tortoisebeacon/tortoise_beacon.go
```diff
@@ -193,76 +203,178 @@ func (tb *TortoiseBeacon) Close() {
...
+	defer tb.mu.Unlock()
 	tb.beacons[epoch] = beacon
-	tb.mu.Unlock()
+}
+
+func (tb *TortoiseBeacon) persistBeacon(epoch types.EpochID, beacon types.Hash32) {
+	if tb.db == nil {
+		return
+	}
+	if err := tb.db.Put(epoch.ToBytes(), beacon.Bytes()); err != nil {
```
dshulyak: if persistence is enabled this error should be checked, why to "optimize" this and make persistence best-effort?

tortoisebeacon/tortoise_beacon.go
```diff
@@ -193,76 +203,178 @@ func (tb *TortoiseBeacon) Close() {
...
 func (tb *TortoiseBeacon) setBeacon(epoch types.EpochID, beacon types.Hash32) {
+	tb.persistBeacon(epoch, beacon)
 	tb.mu.Lock()
+	defer tb.mu.Unlock()
 	tb.beacons[epoch] = beacon
-	tb.mu.Unlock()
+}
+
+func (tb *TortoiseBeacon) persistBeacon(epoch types.EpochID, beacon types.Hash32) {
+	if tb.db == nil {
```
dshulyak: but why to make this optional? i think it was optional in previous code as well, but i don't see any reason to do so

tortoisebeacon/tortoise_beacon.go
```diff
@@ -193,76 +203,178 @@ func (tb *TortoiseBeacon) Close() {
...
+	tb.logger.With().Warning("beacon not available",
+		epochID,
+		log.Uint32("beacon_epoch", uint32(beaconEpoch)))
+	return nil, ErrBeaconNotCalculated
+}
+
+func (tb *TortoiseBeacon) getBeacon(epoch types.EpochID) []byte {
+	tb.mu.RLock()
+	defer tb.mu.RUnlock()
+	if beacon, ok := tb.beacons[epoch]; ok {
```
dshulyak, line 338: why not to store data only in db? it is not going to be a bottleneck

tortoisebeacon/tortoise_beacon.go
```diff
@@ -193,76 +203,178 @@ func (tb *TortoiseBeacon) Close() {
...
+			epoch,
+			log.String("beacon", beacon.ShortString()), log.Err(err))
+	}
+}
+
+func (tb *TortoiseBeacon) getPersistedBeacon(epoch types.EpochID) []byte {
+	if tb.db == nil {
+		return nil
+	}
+	if beacon, err := tb.db.Get(epoch.ToBytes()); err == nil {
```
dshulyak: why not to return an error?

cmd/root.go
dshulyak: to me this number seems a bit magical, I've seen it mentioned in tortoise weakcoin as well. do you know if there is any dependency between layer size and layers per epoch? or how it was invented?

dshulyak (approved):
i am not sure that it is good idea to make persistence best-effort, but up to you

tortoisebeacon/tortoise_beacon.go
countvonzero: ok. i didn't think it's worth failing other component if there is error here. but if disk is failing, we have bigger issues. changed as you suggested

tortoisebeacon/tortoise_beacon.go
countvonzero: my reason was for ease of testing. but i think i was tunnel visioned. thanks for pushing me. changed.

tortoisebeacon/tortoise_beacon.go
countvonzero: it is not going to be a bottleneck for sure but i still don't like to access disk every time a beacon is queried.

our protocol is based on synchronous network. and some messages are time sensitive. the hardware assumption for home miners are **not** SSD but HDD. i think we should still avoid disk seeks and the long tail on disk read time as much as we can. the epoch numbers to keep is limited to 10. i'm not concerned abt memory usage.

tortoisebeacon/tortoise_beacon.go
countvonzero: done

countvonzero:
> to me this number seems a bit magical, I've seen it mentioned in tortoise weakcoin as well. do you know if there is any dependency between layer size and layers per epoch? or how it was invented?

@dshulyak i thought the comment in tortoisebeacon/config.go is adequate.

i am not familiar with the tortoise weak coin (for self-healing).
800 is our security parameter in the whitepaper. during the R&D discussion, Tal and Iddo always refer 800 as a cluster. if we see 800 of something, and if more than 2/3 of them agree, then our security assumption holds.

if we just sticks to 800, then i can hard code a value in the code that says beaconSyncNumBlocks := 800.
as far as i know, we only have hare committee size set to 800.

in the R&D meetings, Tal's idea was to wait K layers that satisfy 800 or 1600 blocks. 
but layers are ticked. let's say the average blocks per layer is 200. then we wait minimum 4 layers.
however, i don't want to wait for 4 full layers if we collected 800 blocks earlier than 4 layers. that's why i decided to use numBlocks instead of layers.

there is no relation to layers per epoch that i know of.

cmd/root.go
countvonzero: i am not familiar with the tortoise weak coin (for self-healing).
800 is our security parameter in the whitepaper. during the R&D discussion, Tal and Iddo always refer 800 as a cluster. if we see 800 of something, and if more than 2/3 of them agree, then our security assumption holds.

if we just sticks to 800, then i can hard code a value in the code that says beaconSyncNumBlocks := 800.
as far as i know, we only have hare committee size set to 800.

in the R&D meetings, Tal's idea was to wait K layers that satisfy 800 or 1600 blocks.
but layers are ticked. let's say the average blocks per layer is 200. then we wait minimum 4 layers.
however, i don't want to wait for 4 full layers if we collected 800 blocks earlier than 4 layers. that's why i decided to use numBlocks instead of layers.

there is no relation to layers per epoch that i know of.

countvonzero:
bors merge

PR #2851, opened 2021-09-30
[Merged by Bors] - p2p node: Package improvements

josebalius:
## Motivation
As stated in Discord, I have been reviewing the whole project and hope to open a series of PRs that clean up a few things in the codebase and make things more idiomatic in general. There's a ton of good Go code here and my hope is that these changes can only make it better, more dependable, and idiomatic.

Instead of opening a huge PR that spans several packages, I figured it would be best to open smaller PRs that are package-focused. I have started with the p2p package and have several PRs to come.

## Changes
- Remove unneeded pkg var: `emptyNode`
- Clean up resource leak potentials with open file handlers
- Add more context to errors
- Idiomatic variable naming

## Test Plan
Unit tests all pass

## TODO

## DevOps Notes

### DIFF
--- a/p2p/node/helpers.go
+++ b/p2p/node/helpers.go
@@ -21,7 +21,7 @@ func GenerateTestNode(t *testing.T) (LocalNode, *Info) {
 }
 
 // GenerateTestNodeWithConfig creates a local test node without persisting data to local store.
-func GenerateTestNodeWithConfig(t *testing.T) (LocalNode, *Info) {
+func GenerateTestNodeWithConfig(t *testing.T) (ln LocalNode, info *Info) {
 
 	tcpPort, err := GetUnboundedPort("tcp", 0)
 	if err != nil {
@@ -34,14 +34,12 @@ func GenerateTestNodeWithConfig(t *testing.T) (LocalNode, *Info) {
 
 	addr := net.TCPAddr{IP: net.ParseIP("0.0.0.0"), Port: tcpPort}
 
-	var localNode LocalNode
-
-	localNode, err = NewNodeIdentity()
+	ln, err = NewNodeIdentity()
 	if err != nil {
 		t.Error(ErrFailedToCreate)
-		return emptyNode, nil
+		return ln, nil
 	}
-	return localNode, &Info{localNode.PublicKey().Array(), addr.IP, uint16(tcpPort), uint16(udpPort)}
+	return ln, &Info{ln.PublicKey().Array(), addr.IP, uint16(tcpPort), uint16(udpPort)}
 }
 
 // GenerateRandomNodeData generates a remote random node data for testing.
--- a/p2p/node/localnode.go
+++ b/p2p/node/localnode.go
@@ -1,6 +1,8 @@
 package node
 
 import (
+	"fmt"
+
 	"github.com/spacemeshos/go-spacemesh/p2p/p2pcrypto"
 )
 
@@ -10,7 +12,7 @@ type LocalNode struct {
 	privKey   p2pcrypto.PrivateKey
 }
 
-// PublicKey returns the node's public key
+// PublicKey returns the node's public key.
 func (n LocalNode) PublicKey() p2pcrypto.PublicKey {
 	return n.publicKey
 }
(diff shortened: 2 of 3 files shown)


p2p/node/helpers.go
```diff
@@ -34,14 +34,12 @@ func GenerateTestNodeWithConfig(t *testing.T) (LocalNode, *Info) {
 
 	addr := net.TCPAddr{IP: net.ParseIP("0.0.0.0"), Port: tcpPort}
 
-	var localNode LocalNode
-
-	localNode, err = NewNodeIdentity()
+	ln, err = NewNodeIdentity()
```
josebalius, line 37: Re-use named return which stops the code from needing an "emptyNode" variable in the package.

p2p/node/localnode.go
```diff
@@ -35,15 +35,15 @@ func NewNodeIdentity() (LocalNode, error) {
...
 	priv, err := p2pcrypto.NewPrivateKeyFromBase58(d.PrivKey)
 	if err != nil {
-		return emptyNode, err
+		return ln, fmt.Errorf("failed to parse private key: %w", err)
 	}
 
 	pub, err := p2pcrypto.NewPublicKeyFromBase58(d.PubKey)
 	if err != nil {
-		return emptyNode, err
+		return ln, fmt.Errorf("failed to parse public key: %w", err)
```
josebalius, line 46: Add context to these errors, so that we can tell when a failure happens from parsing each type of key.

p2p/node/localnodestore.go
```diff
@@ -23,114 +21,77 @@ type nodeFileData struct {
...
-	err = f.Sync()
-	if err != nil {
-		return err
-	}
-
-	err = f.Close()
-	if err != nil {
-		return err
+	if err := ioutil.WriteFile(nodefile, finaldata, 0666); err != nil {
+		return fmt.Errorf("failed to write file: %w", err)
```
josebalius, line 43: `ioutil.WriteFile` is a much better encapsulation for what we are trying to do here. The `0666` file mode is also the same that `File.Write` would usually do.

Also, the previous implementation had a potential resource leak by leaving a file handle open if either `f.Write` or `f.Sync` failed. File handle close defers should happen immediately after the file is created so that we can clean up in all cases.

p2p/node/localnodestore.go
```diff
@@ -23,114 +21,77 @@ type nodeFileData struct {
...
 	}
 
+	// TODO(josebalius): should we accept a logger or otherwise not log directly
```
josebalius, line 46: This log call feels like it's best done by callers or at the very least we should accept a `logger`

p2p/node/localnodestore.go
```diff
@@ -23,114 +21,77 @@ type nodeFileData struct {
...
 
-	_, err = io.Copy(data, f)
-
-	if err != nil {
-		return nil, err
-	}
-
-	err = f.Close()
-
+	b, err := ioutil.ReadFile(nodefile)
```
josebalius, line 69: The previous implementation had a potential resource leak. If the `io.Copy` failed, we would fail to close the file handle. `io.ReadFile` can take care of the whole operation for us and we fix two birds with one stone sort of speak.

p2p/node/localnodestore.go
```diff
@@ -141,15 +102,17 @@ func getLocalNodes(path string) ([]string, error) {
 // ReadFirstNodeData reads node data from the data folder.
 // Reads a random node from the data folder if more than one node data file is persisted.
 // To load a specific node on startup - users need to pass the node id using a cli arg.
-func ReadFirstNodeData(path string) (LocalNode, error) {
+//
+// TODO(josebalius): verify that "reads a random node" is actually happening
```
josebalius: The doc here might need updating, I don't see how this function is reading a "random" node from the data folder. We seem to fetch them all and read the first one. Does it need updating?

y0sher: yes you are correct, it is documenting older behavior..

y0sher (approved):

y0sher:
Thanks for doing this! I appreciate the thought and work.

p2p/node/localnodestore.go
josebalius: Thanks, I'll update the comment and remove the TODO.

josebalius:
@y0sher looks like I need a maintainer's approval to have CI run. Do you know if they will automatically take a look or should I ping someone?

y0sher:
Thanks for reminding, approved for running the CI

nkryuchkov:
bors try

nkryuchkov (approved):
Good job! Thanks for your PR

nkryuchkov:
bors merge

Issue #2795, opened 2021-09-16
hare: does not tag layer as empty when it outputs an empty set


countvonzero:
## Description
hare declared output for a given layer, send it over for verifying tortoise to validate, and verifying tortoise failed to look up layer blocks.

this is because hare output an empty set for that layer but didn't tag it as zero-block layer.

![Screenshot from 2021-09-16 16-25-19](https://user-images.githubusercontent.com/30611210/133681046-0dacede8-1979-40d9-989c-163c9205cb7d.png)

## Actual Behavior
verifying tortoise failed to look up content of the given layer when hare output an empty set for that layer

## Expected Behavior
verifying tortoise should know this is a zero-block layer and behave accordingly instead of errorring out.

dshulyak:
This will have a side effect that syncer will not even try to sync the missed layer because processed was advanced by hare, even when node is completely isolated.

countvonzero:
> This will have a side effect that syncer will not even try to sync the missed layer because processed was advanced by hare, even when node is completely isolated.

i think this issue is orthogonal to hare not tagging empty layer.

even when a node is NOT isolated, it can still miss blocks for a given layer due to gossip not propagating to all nodes.
the root of the issue for the problem you pointed out is, like you also stated in the slack channel, that syncer should not sync data based on processed layer but on aggregated layer hash.

this bug may have been caused by https://github.com/spacemeshos/go-spacemesh/issues/2876 tho.
when hare truly terminates with empty set, it invalidates the layer, and that's it. no one will be notifying the mesh that we can try to process the layer. https://github.com/spacemeshos/go-spacemesh/issues/2875

dshulyak:
do we still need this distinction between empty and zero layer in tortoise?

countvonzero:
the original wording was not quite correct.
it really isn't hare's job to tag this layer as an empty layer. (a layer without any blocks).

either one of the conditions below can cause hare output empty set
- there are blocks in the layer, but somehow the hare committee voted down all of them
- hare terminated with failure
- there are no blocks in a layer

so i think this is still an issue that tortoise failed to look up content of the layer when asked to validate a layer
this means either 
- A. tortoise ignore ErrNotFound from mesh when there is empty input vector
- B. some component tag the layer as empty layer in the mesh when there really is no blocks in this layer
i think B is more correct, so that tortoise doesn't have to care about empty layer.

countvonzero:
[...]

PR #2853, opened 2021-10-01
[Merged by Bors] - sync: error if some requested blocks were not fetched or requested layer is not yet processed

### LINKED ISSUE #2847: sync: blocks are not synced when tortoise verifies the layer
## Description

When layer is verfied after the sync tortoise doesn't always receive all blocks that were produced in that layer. For example compare the amount of blocks that were received by the miner-5 that was participating from the genesis and miner-90 that was started after a day - http://34.70.9.43:5601/goto/b9d7302d425f9924600f861996433b4c. The miner got stuck at layer 322, because received 324 and 325 didn't have enough weight for 323 to pass the threshold.

Somehow we managed to sync those blocks later, i downloaded state from miner-90 and all necessary blocks exist in the database. So if i manually trigger re-run starting from 324 layer it works as expected. Note that by manually i mean downloading state and executing HandleIncomingLayer for each new layer.

This problem can't be fixed by self-healing since we do not re-index the internal tortoise data structure. This issue would be fixed by re-run https://github.com/spacemeshos/go-spacemesh/issues/2846 but i don't know if it works as expected.

dshulyak:
## Motivation

These were two edge cases that were causing an observed issue with missing blocks.

If blocks were requested from the first peer then all other requests will automatically be successful. So even if the initial request will complete only partially the layer won't be failed and will be sent for the verification to the tortoise.

Sometimes the only successful response is from the peer that is lagging behind, or may have only a partial state (e.g. requested layer is 101, processed layer in response is 100). In this case, we can treat this response as failed instead of processing it. Also instead of failing when the node was already synced, we will simply interrupt the sync earlier.

I think we can also recover by fetching base blocks (and not only exceptions), but i wasn't able to test base block change on current devnet.

Closes https://github.com/spacemeshos/go-spacemesh/issues/2847

## Changes
- track blocks that were requested but not received and fail layer if we are not able to fetch all blocks
- if the response has a processed layer lower than the requested layer mark it as failed
- fetch base block in handle block data

## TODO

## DevOps Notes


dshulyak:
bors try

layerfetcher/layers.go
```diff
@@ -303,27 +302,32 @@ func (l *Logic) PollLayerContent(ctx context.Context, layerID types.LayerID) cha
...
 
 	logger.With().Debug("fetching new blocks", log.Int("to_fetch", len(toFetch)))
 	if err := l.GetBlocks(ctx, toFetch); err != nil {
 		// fail sync for the entire layer
 		return err
 	}
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	for _, bid := range toFetch {
+		lyrResult.blocks[bid] = true
```
countvonzero: i am not seeing a lot of value with this bool here. does it matter that much that we know how many blocks are missing? we know there are blocks missing as soon as you see peer result's err is ErrBlockNotFetched, assigned in line 358

syncer/syncer.go
```diff
@@ -288,29 +288,23 @@ func (s *Syncer) synchronize(ctx context.Context) bool {
...
+		layer, err := s.syncLayer(ctx, layerID)
+		if err != nil {
 			logger.With().Error("failed to sync to layer", layerID, log.Err(err))
-			return false
+			success = false
+			break
+		}
+		if layer == nil {
+			// if peers don't have the layer don't revert to non-synced if node was synced
+			success = state != notSynced
```
countvonzero: i don't think this make sense. the syncer always sync `currentLayer-1`. i.e. it never works on current layer. so if does get an nil layer here, that means it's not able to get data for `currentLayer-1` or older. that should set the node back to `notSynced`

syncer/syncer.go
```diff
@@ -357,13 +362,11 @@ func (s *Syncer) setStateAfterSync(ctx context.Context, success bool) {
 	}
 }
 
-func (s *Syncer) syncLayer(ctx context.Context, layerID types.LayerID) (*types.Layer, error) {
+func (s *Syncer) syncLayer(ctx context.Context, layerID types.LayerID) (layer *types.Layer, err error) {
```
countvonzero: i am not a fan of named returned values. i've seen quite a bit of programming mistakes that assign the values by accident. it's not very obvious for debugging. i see you return explicitly but it leaves room for ppl who edit the code later. i really think the extra two lines of var declaration is worth it.

syncer/syncer.go
```diff
@@ -373,9 +376,18 @@ func (s *Syncer) syncLayer(ctx context.Context, layerID types.LayerID) (*types.L
 		}
 	} else {
 		s.logger.WithContext(ctx).With().Info("polling layer content", layerID)
-		if layer, err = s.getLayerFromPeers(ctx, layerID); err != nil {
+		var highest types.LayerID
+		layer, highest, err = s.getLayerFromPeers(ctx, layerID)
+		if err != nil {
 			return nil, err
 		}
+		if highest.Before(layerID) {
```
countvonzero: i think this is quite a bit more logic/code than necessary to deal with this issue, here and in layers.go.

what do you think about changing the code in `layerBlocksReqReceiver` to only return blocks when the requested layer is lower than its processed layer?

layerfetcher/layers.go
```diff
@@ -388,35 +387,40 @@ func (l *Logic) receiveLayerContent(ctx context.Context, layerID types.LayerID,
...
 
-	if result == nil { // no block data available
-		result = &LayerPromiseResult{Layer: layerID, Err: nil}
-		if hasZeroBlockHash {
-			// all other non-empty layer hashes returned errors. use the best information we've got
-			result.Err = ErrZeroLayer
-		} else {
-			// no usable result. just return the first error we received
-			result.Err = firstErr
+	if missing > 0 {
```
countvonzero: can you try to simplify the code logic here a little bit? there is a fair amount of switch back and forth on the success and failure cases.

personally i don't see the value of report how many blocks missing, as long as one of the peer reported ErrBlockNotFetched

countvonzero (commented):
my opinion is the code change will be simpler if you simply change `layerBlocksReqReceiver` to return block contents **only** for layer <= its processed layer

syncer/syncer.go
dshulyak: the problem is that when syncer requests currentLayer-1 that layer is not processed usually, and there is no way to know what we are actually receiving

layerfetcher/layers.go
dshulyak: i was using it during debugging to see if this is where blocks are missing, but right i can simply check all errors instead of one

dshulyak (commented):
there is something wrong with sync after whole cluster starts, nodes can't pass layer 4 in tests (5 is effective genesis) because every other node has processed layer 4. it seems that they can only do that if another node responds to sync request

syncer/syncer.go
dshulyak: good suggestion, i considered it before but decided that i don't want to error and switch back to non-synced in some cases. but maybe that was wrong

dshulyak:
@countvonzero do you know why sync doesn't fail when input vector is missing? i noticed that some nodes from devnet doesn't have it, and it is used in tortoise for recent layers, maybe we can make it mandatory?

dshulyak:
bors try

layerfetcher/layers.go
```diff
@@ -190,8 +192,12 @@ func (l *Logic) epochATXsReqReceiver(ctx context.Context, msg []byte) ([]byte, e
 // it also returns the validation vector for this data and the latest blocks received in gossip
 func (l *Logic) layerBlocksReqReceiver(ctx context.Context, req []byte) ([]byte, error) {
 	lyrID := types.BytesToLayerID(req)
+	processed := l.layerDB.ProcessedLayer()
+	if lyrID.After(processed) {
+		return nil, fmt.Errorf("%w: requested layer %v is higher then processed %v", errLayerNotProcessed, lyrID, processed)
```
countvonzero: :s/then/than/

dshulyak:
I think i will have to remove sync system test. In that test processed layer is advanced only by the sync which is not compatible with the change that i am doing. The processed layer needs to be advanced by the tortoise after genesis.

@countvonzero are you using that test (and cmd/sync)?

layerfetcher/layers.go
```diff
@@ -357,35 +363,33 @@ func (l *Logic) receiveLayerContent(ctx context.Context, layerID types.LayerID,
...
-	var firstErr error
+	var (
+		missing, success bool
+		err              error
+	)
 	for _, res := range lyrResult.responses {
 		if res.err == nil && res.data != nil {
-			// at least one layer hash contains blocks. not a zero block layer
-			result = &LayerPromiseResult{Layer: layerID, Err: nil}
+			success = true
```
countvonzero, line 372: is this necessary? i think you just need to check whether missing or err != nil outside the loop?

syncer/syncer.go
```diff
@@ -367,6 +367,8 @@ func (s *Syncer) syncLayer(ctx context.Context, layerID types.LayerID) (*types.L
 		if err = s.mesh.SetZeroBlockLayer(layerID); err != nil {
 			s.logger.WithContext(ctx).With().Panic("failed to set zero-block for genesis layer", layerID, log.Err(err))
 		}
+	}
+	if !layerID.After(types.GetEffectiveGenesis()) {
```
countvonzero, line 371: should be the opposite?

dshulyak:
bors try

countvonzero (commented):
nice!

havent read the unit tests.

i dont understand why the test parameters should change

layerfetcher/layers.go
dshulyak: if i only check the error then even a single failure will mark whole attempt failed

syncer/syncer.go
dshulyak: i think it is correct, why opposite? e.g fetch data from db `if layerID <= genesis`

dshulyak (commented):
app multi node test was hanging, i was blaming layer duration at first, but the it turned out some problem with healing test

@countvonzero but what test parameters are you asking about?

syncer/syncer.go
countvonzero: i misread. sorry

countvonzero:
> I think i will have to remove sync system test. In that test processed layer is advanced only by the sync which is not compatible with the change that i am doing. The processed layer needs to be advanced by the tortoise after genesis.
> 
> @countvonzero are you using that test (and cmd/sync)?

i am not. but @antonlerner should know better

countvonzero:
@dshulyak yes i was talking abt the duration change. 

if self healing test fails due to this change, that should be fixed

dshulyak:
bors try

countvonzero:
> @dshulyak yes i was talking abt the duration change.

and also the tx counts. i dont understand why those need change

countvonzero:
sorry. reviewing with cell phone and something was wrong with browser caching. i see you have reverted those param change

dshulyak:
> if self healing test fails due to this change, that should be fixed

yes, i was going to look at it before merging. but i actually think that if it was relying on that sync response from a node that doesn't have a processed layer then something is wrong with a test

countvonzero (approved):
change LG. approved first (boarding the plane soon)

plz get agreement w Anton on the sync system test and Lane on the self healing test. thanks

dshulyak:
Healing test was relying on syncer to set SetZeroLayer on a layer where no-one was eligible to produce blocks, so that later tortoise won't fail with missing layer. I refactored mesh code to return empty set of blocks even if layer is not set as zero, it seems to me that was a legacy code.

dshulyak:
bors try

dshulyak:
bors merge

turbohare/hare.go
```diff
@@ -64,13 +66,19 @@ func (h *SuperHare) Start(ctx context.Context) error {
...
+					}
+					blocks, err := h.mesh.LayerBlockIds(layerID)
+					if err != nil {
+						logger.With().Warning("error reading block ids for layer, using empty set",
 							layerID,
 							log.Err(err))
-					} else {
-						h.mesh.HandleValidatedLayer(ctx, layerID, layerBlocks)
 					}
+					if len(blocks) == 0 {
```
countvonzero, line 76: you can mark https://github.com/spacemeshos/go-spacemesh/issues/2795 resolved by this PR @dshulyak

countvonzero: oh wait. this change is only for turbo hare...

PR #2855, opened 2021-10-04
[Merged by Bors] - Wrap most errors with additional context

nkryuchkov:
## Motivation

Currently, many error logs are not descriptive enough. E.g., there're many errors like `$ERROR: leveldb: not found`, where it's not clear what's the origin of the error. Also, it's a good practice to wrap errors is Go. 

The PR wraps most of the errors with additional context. Most of them were found with the `wrapcheck` linter, the other part manually. There are still some unwrapped errors inside packages, but I'm not sure we need to wrap them. 

There also will be another PR, which adds the `wrapcheck` linter to the CI, so that we don't add new unwrapped errors.

## Changes

- Wrap most errors with additional context

## Test Plan

- Unit tests
- System tests

## DevOps Notes


nkryuchkov:
bors try

nkryuchkov:
bors try-

nkryuchkov:
bors try

dshulyak (approved):
i haven't used this linter before, but let's try. on a more general note maybe we should use module that adds a caller stack trace to the error message (https://pkg.go.dev/golang.org/x/xerrors#pkg-overview) , it usually very helpful during debugging

countvonzero (commented):
hmmm... i am on the fence about this. when the codebase grows, there will be multiple layers of errors being wrapped, and the error becomes much less readable due to the added text.

i've seen code using https://pkg.go.dev/github.com/juju/errors that simply add file+line# to error.

nkryuchkov:
@countvonzero Yes, the errors would be less readable because of more text, but easier to investigate because we'll see the error trace.

@dshulyak @countvonzero Having a stack trace is a great idea, but I think we can do that iteratively. This PR will make investigations of some errors way easier. We can add something more enhanced later.

nkryuchkov:
bors merge

PR #2859, opened 2021-10-04
Rewrite metrics

nkryuchkov:
## Motivation

Currently, metrics are not mockable, so it's hard to test them. This PR rewrites them so that we could mock and test metrics.

## Changes

- Make metrics mockable

## Test Plan

- Unit tests
- System tests

## TODO

- [ ] Finish the implementation

## DevOps Notes


blocks/blocks.go
```diff
@@ -52,14 +54,15 @@ type Config struct {
 }
 
 // NewBlockHandler creates new BlockHandler
-func NewBlockHandler(cfg Config, m mesh, v blockValidator, lg log.Log) *BlockHandler {
+func NewBlockHandler(cfg Config, m mesh, v blockValidator, lg log.Log, metrics metrics.Metrics) *BlockHandler {
```
dshulyak, line 57: please no need to pass metrics to object. it would mean that we will have to follow this practice and this is just too much complexity. if you want to mock metrics for tests then  it should be done globally, but i in general doubt that this is necessary

dshulyak (changes requested):

metrics/metrics.go
```diff
@@ -0,0 +1,245 @@
...
+const (
+	LayerIDLabel   = "layer_id"
+	BlockIDLabel   = "block_id"
+	LastLayerLabel = "last_layer"
+)
+
+//go:generate mockgen -package=mocks -destination=./mocks/mocks.go -source=./metrics.go
+
+// Metrics defines an interface for metrics.
+type Metrics interface {
```
dshulyak, line 27: don't do this please. if you really want to test metrics - implementation of the metric system should be mocked. e.g. go-kit needs to be replaced during initialization (if it doesn't support testing)

countvonzero:
i tend to agree. there is not an urgent need to test metrics now.

nkryuchkov:
Prometheus metrics should be tested with https://github.com/prometheus/client_golang/prometheus/testutil instead

PR #2862, opened 2021-10-06
[Merged by Bors] - sync: fail on missing input vector

countvonzero:
## Motivation

following up on discussion with @dshulyak on missing input vector during sync

## Changes

- sync fails on missing input vector
- layer fetcher tag layer as empty in mesh instead of syncer. this removes the need to use ErrZeroBlockLayer to signal empty layer from layer fetcher to syncer. syncer simply try to get layer from mesh after fetching is done.
- update syncer tests to be consistent with new logic where we don't fetch blocks for genesis layer anymore

## Test Plan

unit tests

## DevOps Notes


countvonzero:
bors try

layerfetcher/layers.go
```diff
@@ -354,39 +352,48 @@ func (l *Logic) receiveLayerContent(ctx context.Context, layerID types.LayerID,
...
 		if err == nil {
 			err = res.err
 		}
 	}
 	result := LayerPromiseResult{Layer: layerID}
+	// we tolerate errors from peers as long as we fetched all known blocks in this layer.
 	if missing || !success {
 		result.Err = err
 	}
+	if result.Err == nil && !hasBlocks { // none of the successful peers has data
```
dshulyak: we can use len(lyrResult.blocks) instead of one more variable

dshulyak (approved):

layerfetcher/layers.go
countvonzero: done. thanks. i also moved saving input vector to the same block and report error on failure to save input vector in db

countvonzero:
bors merge

Issue #2069, opened 2020-07-13
Standardize rewards type


lrettig:
We use the type `big.Int` to store rewards, but everywhere else, including in the log statement at the bottom of this chunk, and in the API code, we use uint64. Can the reward amount ever be negative? If not, can we harmonize this type and move it all to uint64?

https://github.com/spacemeshos/go-spacemesh/blob/b49a23f5d839f0d2d24b5b4f34f2fd372eeb97e8/mesh/mesh.go#L691-L714

neysofu:
Resurrecting this just to document that we're set on using `uint64`. This decision seems to be final.

https://github.com/spacemeshos/go-spacemesh/issues/2718#issuecomment-941092617

Issue #2718, opened 2021-08-25
Account Balance: proper types to fit all possible amounts


brusherru:
@lrettig @avive
I want to ensure that the huge coin amounts can fit in the `uint64` type, that is used now to represent the account `Balance` type.
To answer this main question, I think we have to answer these questions:
- What is the limit for the total amount of coins?
- If there is no limit, what is the inflation plan, when the emission of coins become such small, that it almost can not meet the maximum value of the type?

So, if the total amount of coins is very much more than `18,446,744,073,709,551,615 Smidge` (`18446744.073709551615 SMH`). I can assume, that someday there might appear a rich guy (especially if the market cost of such amount of Smesh will not strive for infinity), that owns such amount or more and he still should be able to receive payments.

If there exists at least a smallest possibility that it is really possible, I think that we have to use a different type to store `Balance` (`*big.Int`) and, as a consequence, we should change `Amount` type in `types.proto` to something more spacious than just `uint64`.

P.S.
Balance type now: https://github.com/spacemeshos/go-spacemesh/blob/746c085fbf1a85c0226d5e82212a1113efb60320/common/types/account_state.go
Pay attention to the comment "TODO" in the file above. It was a BigInt some time ago: https://github.com/spacemeshos/go-spacemesh/issues/2192

avive:
@noamnelke

lrettig:
The planned supply cap will be around 2.4bn SMH. That's a couple of orders
of magnitude greater than the figure you cited, so yes, it seems like we do
need to use a big int.

>

dshulyak:
We should use https://github.com/holiman/uint256 then. I know there was some overhead in go-etheereum since big.Int uses a lot of dynamic allocations and is inefficient in this way, so they made their own library for uint256 type (unless it was changed lately).

brusherru:
So, we also have to decide how we can pass such amounts via GRPC. Just as a string / as two uint64 / something else.
And then, who will create issues to the relates repos? ;)

lrettig:
We could also just return the numbers from the API as a binary blob, as we already do with addresses

brusherru:
I thought about it a little and came to the opinion, that they should be returned as strings.

Here are the pros:
- the implementation of big numbers on the client-side (Smapp / Smrepl / something else) will not depend on the implementation of big numbers on the "back-end" (is it big-endian, little-endian, what part is an exponent, and so on)
- it is human-readable by default

[...]

PR #2864, opened 2021-10-06
[Merged by Bors] - Refactor `mesh.go` logic into `SVM.ApplyLayer`

### LINKED ISSUE #2831: Implement `ApplyLayer` in SVM package
```go
func (s *SVM) ApplyLayer(layerID types.LayerID, transactions []types.Transaction, rewards []types.Reward)
```
Method called when a layer is applied, after Hare/Tortoise validation. This functionality is currently spread across different places and I'm not sure it's atomic (it should be).

#### Note Regarding `types.Reward`
This may not be a suitable type for this method's parameter, as it holds much more information than is needed (likely only an address and amount are required). I've used it because it was there, but we should definitely consider creating a special type for this.

### Called From
- `mesh/mesh.go` -> `Mesh.applyState`

### Existing Functions
- `TransactionProcessor.ApplyRewards`
- `TransactionProcessor.ApplyTransactions`
- `TransactionProcessor.DB.Commit`
- `TransactionProcessor.GetStateRoot`

### SVM Functions
- `IncreaseBalance`
- `Deploy`
- `Spawn`
- `Call`
- `Commit`

neysofu:
## Motivation

Closes #2831. It also brings forward #2069 by replacing `big.Int` with `uint64` in a few places.

## Changes

This PR completely replaces the `mesh.go:txProcessor` API with `mesh.go:svm`, which adds the `ApplyLayer` method. Some logic is then moved into the `svm` package.

## Test Plan

Relying on the current testing infrastructure as no functional changes are expected.

## DevOps Notes


dshulyak:
it came up some time ago that we can't fit balance into 64 bits - https://github.com/spacemeshos/go-spacemesh/issues/2718#issuecomment-905974522 . if that's the case we should move not to uint64, but to big.Int or uint256/128 using custom lib. can we clarify that before moving forward with a proposed refactoring?

neysofu:
@dshulyak absolutely, I was under the impression that we wanted to move forward with `uint64`, my bad! Most of this refactoring is unrelated to reward types and the work that I've done already is easily reversible. I'll put it on hold. @noamnelke I'll keep this PR focused on `SVM.ApplyLayer` and not deal with reward types.

neysofu:
I see most of the code that I'm touching is still based on `*big.Int` (`uint64` as well, but we've established already that's not okay). I'm not sure if `*big.Int` the final decision, but I guess I should conform to that for all new code. @dshulyak what do you think?

dshulyak:
> I'm not sure if *big.Int the final decision, but I guess I should conform to that for all new code. @dshulyak what do you think?

i think the same. based on the latest observations it is safer to assume that we won't be able to use uint64

noamnelke:
Added this comment: https://github.com/spacemeshos/go-spacemesh/issues/2718#issuecomment-941092617

Let's use `uint64` 😃

activation/activationdb_test.go
```diff
@@ -67,52 +66,56 @@ func (m *MeshValidatorMock) HandleLateBlocks(_ context.Context, blocks []*types.
...
 }
 
-func (MockState) ApplyTransactions(types.LayerID, []*types.Transaction) (int, error) {
-	return 0, nil
+func (MockSVM) ApplyTransactions(types.LayerID, []*types.Transaction) ([]*types.Transaction, error) {
+	return []*types.Transaction{}, nil
 }
 
-func (MockState) ApplyRewards(types.LayerID, []types.Address, *big.Int) {
+func (MockSVM) ApplyLayer(layerID types.LayerID, transactions []*types.Transaction, rewards []types.AmountAndAddress) ([]*types.Transaction, error) {
```
countvonzero: nit: rm variable names

mesh/mesh.go
```diff
@@ -495,21 +494,117 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
 		}
 	}
 }
 
+type layerRewards struct {
+	types.LayerID
+	Total       uint64
+	LayerReward uint64
+	LayerFees   uint64
+	Blocks      []types.AmountAndAddress
```
countvonzero: why is it named Blocks?

common/types/transaction.go
```diff
@@ -134,6 +134,12 @@ type InnerTransaction struct {
 	Amount       uint64
 }
 
+// AmountAndAddress stores an amount together with an account's address.
```
countvonzero: maybe RewardInfo, RewardRecipient?
naming a struct by its contents may not be future-proof :)

mesh/mesh.go
```diff
@@ -495,21 +494,117 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
+	BySmesher   map[string][]int
```
countvonzero: what is this map storing?

mesh/mesh.go
```diff
@@ -495,21 +494,117 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
+
+func (lr *layerRewards) numBlocks() uint64 {
+	return uint64(len(lr.Blocks))
+}
+
+func (msh *Mesh) reportRewards(lr layerRewards) {
+	// Report the rewards for each coinbase and each smesherID within each coinbase.
+	// This can be thought of as a partition of the reward amongst all the smesherIDs
+	// that added the coinbase into the block.
+	for smesherString, idxs := range lr.BySmesher {
```
countvonzero: i am not sure. but it seems that data inside BySmesher and Blocks can be combined somehow to make this code more intuitive? i.e. loop over the coinbase instead of smesher+block-lookups

mesh/mesh.go
```diff
@@ -495,21 +494,117 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
 			// We ignore errors here, since they mean that the tx is no longer valid and we shouldn't re-add it
 			msh.With().Info("transaction from contextually invalid block re-added to mempool", tx.ID())
 		}
 	}
 }
 
+type layerRewards struct {
+	types.LayerID
+	Total       uint64
+	LayerReward uint64
```
countvonzero: what is the diff btwn TotalReward and LayerReward

mesh/mesh.go
```diff
@@ -495,21 +494,117 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
+	LayerFees   uint64
+	Blocks      []types.AmountAndAddress
+	BySmesher   map[string][]int
+}
+
+func (lr *layerRewards) numBlocks() uint64 {
+	return uint64(len(lr.Blocks))
+}
+
+func (msh *Mesh) reportRewards(lr layerRewards) {
```
countvonzero: perhaps use *types.layerRewards here and below to save copying struct. normally i wouldn't care for small/simple struct but here there are 3 copies made for log/report/write.

mesh/mesh.go
```diff
@@ -495,21 +494,117 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
+				Total:       lr.Total,
+				LayerReward: lr.LayerReward,
+				Coinbase:    lr.Blocks[idx].Address,
+				Smesher:     *smesher,
+			})
+		}
+	}
+}
+
+func (msh *Mesh) writeRewards(lr layerRewards) error {
```
countvonzero: any chance report and write can be combined? they seem to do the same loop, just slightly different output

mesh/mesh.go
```diff
@@ -495,21 +494,117 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
+		log.Uint64("total_reward_remainder", lr.Total%lr.numBlocks()),
+		log.Uint64("layer_reward_remainder", lr.LayerReward%lr.numBlocks()),
+	)
+}
+
 func (msh *Mesh) applyState(l *types.Layer) {
-	msh.accumulateRewards(l, msh.config)
-	msh.pushTransactions(l)
+	validBlockTxs := msh.extractUniqueOrderedTransactions(l)
+	fmt.Printf("CALLING APPLYSTATE with %d valid txs, layer %d \n", len(validBlockTxs), l.Index().Value)
```
countvonzero: rm or log?

svm/api.go
```diff
@@ -36,16 +36,31 @@ func (svm *SVM) SetupGenesis(conf *config.GenesisConfig) error {
...
+	_, err := svm.TransactionProcessor.Commit()
 	if err != nil {
 		return fmt.Errorf("cannot commit genesis state: %w", err)
 	}
 	return nil
 }
+
+// ApplyLayer should be called when a layer is applied, i.e. after Hare/Tortoise
+// verification.
+func (svm *SVM) ApplyLayer(layerID types.LayerID, transactions []*types.Transaction, rewards []types.AmountAndAddress) (failed []*types.Transaction, err error) {
```
countvonzero: can we avoid using named returned variables? i have seen too many subtle bugs like assignment by accident.

mesh/mesh.go
neysofu: Removed. This slipped through, it was never intended to make upstream.

mesh/mesh.go
neysofu: `Total` is `LayerReward` plus layer fees. I've removed `Total` and turned it into a method which calculates it on the fly.

mesh/mesh.go
neysofu: It was supposed to indexes into `Blocks`, but it was ugly as hell and confusing. I think it's better now.

mesh/mesh.go
```diff
@@ -48,9 +47,8 @@ type Validator interface {
 	ValidateLayer(context.Context, *types.Layer)
 }
 
-type txProcessor interface {
-	ApplyTransactions(layer types.LayerID, txs []*types.Transaction) (int, error)
-	ApplyRewards(layer types.LayerID, miners []types.Address, reward *big.Int)
+type svm interface {
```
noamnelke: Not sure the interface should be called `svm`. We should call it something based on the role of this object in the local context - possibly `state`. `state.ApplyLayer()` makes sense, all the other functions, too. SVM is just an implementation of the state component.

svm/api.go
```diff
@@ -12,8 +12,8 @@ import (
 
 // SVM is an entry point for all SVM operations.
 type SVM struct {
-	state *state.TransactionProcessor
-	log   log.Logger
+	*state.TransactionProcessor
```
noamnelke: I think this is a bad idea. The purpose of creating the `SVM` struct is to gain precise control over the methods it exposes. I think that this is a bad shortcut that will make the next phases of the SVM transition harder.

With this change, `SVM` exposes the same api as `TransactionProcessor` allowing us to use it in its place and making it harder to see what methods we still need to either implement in `SVM` or eliminate.

If there are methods that need to eventually be implemented in `SVM` - let's create them and call the matching method in `TransactionProcessor` explicitly. If there are methods that **don't** need to be there eventually, but handling them properly is unrelated to this PR - you can give them the same treatment, but add a `TODO` to remove them.

mesh/mesh.go
noamnelke: Also, most of the methods here make sense to be implemented in the `SVM`, with one exception: `ValidateAndAddTxToPool`. Does the transaction processor currently have access to the mempool? I think we should separate it, but not in this PR - for now you can implement this method - but add a `TODO` to remove it.

noamnelke (changes requested):

svm/api.go
neysofu: Ok, so to recap:

- `SVM` is not a good name for the interface. I understand your point, which makes sense. We can use `state` rather than the old `txProcessor`, I think that's okay.
- Exposing `SVM.TransactionProcessor` is simply a step towards full integration. This way, most of the code stays the same - because all previously used methods are also available on `SVM`. This was a plus in my book - incremental changes moving the abstraction layer from `txProcessor` to `SVM`, but of course the downside is that indeed you can lose track of where we are in the integration refactoring process.

neysofu: I'll undo this change and implement all methods on `SVM` directly, which will then call into `TransactionProcessor`. In a future PR, we can decide which methods to keep and which ones to discard; I'll make sure to leave a `TODO` note to help us remember. Sounds good?

mesh/mesh.go
```diff
@@ -944,18 +946,43 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
...
+func (msh *Mesh) getCoinbasesAndSmeshers(l *types.Layer) (coinbasesAndSmeshers map[types.Address]map[string]uint64, coinbases []types.Address) {
+	coinbases = make([]types.Address, 0, len(l.Blocks()))
+	// the reason we are serializing the types.NodeID to a string instead of using it directly as a
+	// key in the map is due to Golang's restriction on only Comparable types used as map keys. Since
+	// the types.NodeID contains a slice, it is not comparable and hence cannot be used as a map key
+	// TODO: fix this when changing the types.NodeID struct, see https://github.com/spacemeshos/go-spacemesh/issues/2269
+	coinbasesAndSmeshers = make(map[types.Address]map[string]uint64)
+	for _, bl := range l.Blocks() {
+		if bl.ATXID == *types.EmptyATXID {
+			msh.With().Info("skipping reward distribution for block with no atx", bl.LayerIndex, bl.ID())
```
countvonzero, line 968: how is this possible?

mesh/mesh.go
```diff
@@ -944,18 +946,43 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
...
+			continue
+		}
+		atx, err := msh.AtxDB.GetAtxHeader(bl.ATXID)
+		if err != nil {
+			msh.With().Warning("atx from block not found in db", log.Err(err), bl.ID(), bl.ATXID)
```
countvonzero, line 973: i think this should be error. these blocks have been previously verified

mesh/mesh.go
```diff
@@ -512,8 +510,50 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
...
+		}
+		failedTxs, svmErr = msh.state.ApplyLayer(l.Index(), validBlockTxs, rewardByMiner)
+		msh.logRewards(&rewards)
+		msh.reportRewards(&rewards, coinbasesAndSmeshers)
+
+		if err := msh.DB.writeTransactionRewards(l.Index(), coinbasesAndSmeshers, rewards.blockTotalReward, rewards.blockLayerReward); err != nil {
+			msh.Log.Error("cannot write reward to db")
+		}
+	} else {
+		msh.With().Info("no valid blocks for layer", l.Index())
```
countvonzero, line 540: is this log message correct? these are all valid blocks...

countvonzero, line 540: also, is this an error? is it possible that there is no coinbase for a layer?

mesh/mesh.go
```diff
@@ -944,18 +946,43 @@ func (msh *Mesh) accumulateRewards(l *types.Layer, params Config) {
...
+		if bl.ATXID == *types.EmptyATXID {
+			msh.With().Info("skipping reward distribution for block with no atx", bl.LayerIndex, bl.ID())
+			continue
+		}
+		atx, err := msh.AtxDB.GetAtxHeader(bl.ATXID)
+		if err != nil {
+			msh.With().Warning("atx from block not found in db", log.Err(err), bl.ID(), bl.ATXID)
+			continue
+		}
+		coinbases = append(coinbases, atx.Coinbase)
```
countvonzero, line 976: this list may contain duplicate elements. is that expected by the consumers?

mesh/mesh.go
neysofu: This whole portion was left untouched by the current PR. Maybe we can investigate in the future in a follow-up PR? I'm not really familiar with this code.

mesh/mesh.go
neysofu: Same as above, this is not super relevant but I think it should be addressed in a separate PR.

mesh/mesh.go
neysofu: Unsure; I left this untouched. I believe some tests were failing without custom handling in this `else` block (triggers division by zero). We can easily remove this if it proves not to be necessary.

neysofu: The log message is inside the `else` block, which deals with the situation where there are no (valid) blocks at all. I believe it's correct, am I missing something?

mesh/mesh.go
neysofu: I think it is, it definitely is for all reward calculation logic and `state` code. It's expected that a miner can submit more than one block per layer.

noamnelke (approved):
I think it looks good to merge. There's obviously lot's of follow up work. Will you create issues for everything and mention them in the epic?

neysofu:
On it!

neysofu:
bors try

neysofu:
bors merge

PR #2868, opened 2021-10-07
[Merged by Bors] - hare: determine good blocks based on beacon value

### LINKED ISSUE #2811: beacon sync: hare uses synced beacon to determine "good" block
## Description

conversation on slack research channel

kimmy
```
@Tal 
regarding tortoise beacon sync, i'd like to double-check with you on the implementation.
for a node that is eligible to propose blocks and participate in hare in epoch N, but didn't 
participate tortoise beacon protocol in epoch N-1:
- we wait for X layers (containing 2 cluster of 800 blocks) and use the beacon that has the 
  most weight, say beacon B1.
- when we receive a block block_A via gossip, we vrf verify the block according to the beacon 
  value in the first block of the epoch, say beacon B2
(below are true for the general case)
- during the hare round, we use B1 to vrf verify and excludes block_A from this layer
- verifying tortoise will use weight 0 for the votes in block_A for K layers
- self-healing will count the votes block_A regardless of the beacon value
does this sound right to you?
```
Tal
```
@kimmy
It all sounds right except for the last bullet. Self-healing uses weight 0 for the votes in block_A 
for K layers. The verifying tortoise actually doesn't need to count the votes at all, ever 
(i.e., we can add to the definition of "good block" that it has the "correct" beacon value) --- 
but using weight 0 for K layers is fine too.
```

countvonzero:
## Motivation

Closes #2811
also follow up on beacon sync PR https://github.com/spacemeshos/go-spacemesh/pull/2850, add test to hare

## Changes

- add test to hare.onTick for not synced and beacon unavailable
- use gomock for mockRoracle
- use epoch beacon value to decide good blocks as the hare's proposal set

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/hare/hare.go
+++ b/hare/hare.go
@@ -257,7 +257,7 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) (bool, error) {
 	c, err := h.broker.Register(ctx, instID)
 	if err != nil {
 		logger.With().Error("could not register consensus process on broker", log.Err(err))
-		return false, fmt.Errorf("broker registerr: %w", err)
+		return false, fmt.Errorf("broker register: %w", err)
 	}
 	cp := h.factory(h.config, instID, set, h.rolacle, h.sign, h.network, h.outputChan)
 	cp.SetInbox(c)
--- a/hare/hare.go
+++ b/hare/hare.go
@@ -1,6 +1,7 @@
 package hare
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -8,8 +9,10 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/spacemeshos/go-spacemesh/blocks"
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/common/util"
+	"github.com/spacemeshos/go-spacemesh/database"
 	"github.com/spacemeshos/go-spacemesh/hare/config"
 	"github.com/spacemeshos/go-spacemesh/hare/metrics"
 	"github.com/spacemeshos/go-spacemesh/log"
@@ -52,6 +55,7 @@ type Hare struct {
 	sign Signer
 
 	mesh    meshProvider
+	beacons blocks.BeaconGetter
 	rolacle Rolacle
 	patrol  layerPatrol
 
@@ -81,6 +85,7 @@ func New(
 	nid types.NodeID,
 	syncState syncStateFunc,
 	mesh meshProvider,
+	beacons blocks.BeaconGetter,
 	rolacle Rolacle,
 	patrol layerPatrol,
 	layersPerEpoch uint16,
@@ -103,6 +108,7 @@ func New(
 	h.sign = sign
 
 	h.mesh = mesh
+	h.beacons = beacons
 	h.rolacle = rolacle
 	h.patrol = patrol
 
@@ -122,9 +128,8 @@ func New(
 
 func (h *Hare) getLastLayer() types.LayerID {
 	h.layerLock.RLock()
-	lyr := h.lastLayer
-	h.layerLock.RUnlock()
-	return lyr
+	defer h.layerLock.RUnlock()
+	return h.lastLayer
 }
 
 // checks if the provided id is too late/old to be requested.
@@ -185,6 +190,11 @@ func (h *Hare) collectOutput(ctx context.Context, output TerminationOutput) erro
 // this function triggers the start of new consensus processes.
 func (h *Hare) onTick(ctx context.Context, id types.LayerID) (bool, error) {
 	logger := h.WithContext(ctx).WithFields(id)
+	if h.IsClosed() {
+		logger.Info("hare exiting")
+		return false, nil
+	}
+
 	h.layerLock.Lock()
 	if id.After(h.lastLayer) {
 		h.lastLayer = id
@@ -207,7 +217,8 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) (bool, error) {
 		}
 	}()
 
-	if !h.rolacle.IsEpochBeaconReady(ctx, id.GetEpoch()) {
+	beacon, err := h.beacons.GetBeacon(id.GetEpoch())
+	if err != nil {
 		logger.Info("not starting hare since beacon is not retrieved")
 		return false, nil
 	}
@@ -238,16 +249,7 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) (bool, error) {
 		return false, nil
 	}
 
-	// retrieve set from orphan blocks
-	blocks, err := h.mesh.LayerBlockIds(h.lastLayer)
-	if err != nil {
-		logger.With().Error("no blocks found for hare, using empty set", log.Err(err))
-		// just fail here, it will end hare with empty set result
-		// return // ?
-		// TODO: there can be a difference between just fail with empty set
-		// TODO:   and achieve consensus on empty set
-	}
-
+	blocks := h.getGoodBlocks(h.lastLayer, beacon, logger)
 	logger.With().Info("starting hare consensus with blocks", log.Int("num_blocks", len(blocks)))
 	set := NewSet(blocks)
 
@@ -269,6 +271,44 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) (bool, error) {
 		log.Int32("count", atomic.AddInt32(&h.totalCPs, 1)))
 	metrics.TotalConsensusProcesses.With("layer", id.String()).Add(1)
 	return true, nil
+}
+
+// getGoodBlocks finds the "good blocks" for the specified layer. a block is considered a good block if
+// it has the same beacon value as the node's beacon value.
+// any error encountered will be ignored and an empty set is returned.
+func (h *Hare) getGoodBlocks(lyrID types.LayerID, epochBeacon []byte, logger log.Log) []types.BlockID {
+	blocks, err := h.mesh.LayerBlocks(lyrID)
+	if err != nil {
+		if err != database.ErrNotFound {
+			logger.With().Error("no blocks found for hare, using empty set", log.Err(err))
+		}
+		return []types.BlockID{}
+	}
+
+	var goodBlocks []types.BlockID
+	for _, b := range blocks {
+		beacon := b.TortoiseBeacon
+		if b.RefBlock != nil {
+			refBlock, err := h.mesh.GetBlock(*b.RefBlock)
+			if err != nil {
+				logger.With().Error("failed to find ref block",
+					b.ID(),
+					log.String("ref_block_id", b.RefBlock.AsHash32().ShortString()))
+				return []types.BlockID{}
+			}
+			beacon = refBlock.TortoiseBeacon
+		}
+
+		if bytes.Equal(beacon, epochBeacon) {
+			goodBlocks = append(goodBlocks, b.ID())
+		} else {
+			logger.With().Warning("block has different beacon value",
+				b.ID(),
+				log.String("block_beacon", types.BytesToHash(beacon).ShortString()),
+				log.String("epoch_beacon", types.BytesToHash(epochBeacon).ShortString()))
+		}
+	}
+	return goodBlocks
 }
 
 var (
--- a/hare/hare.go
+++ b/hare/hare.go
@@ -255,14 +255,14 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) (bool, error) {
 	c, err := h.broker.Register(ctx, instID)
 	if err != nil {
 		logger.With().Error("could not register consensus process on broker", log.Err(err))
-		return false, err
+		return false, fmt.Errorf("broker registerr: %w", err)
 	}
 	cp := h.factory(h.config, instID, set, h.rolacle, h.sign, h.network, h.outputChan)
 	cp.SetInbox(c)
 	if err = cp.Start(ctx); err != nil {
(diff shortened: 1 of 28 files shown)


dshulyak:
@countvonzero does this change effectively enables tortoise beacon? i think it is definitely in a better shape now, but if there are still some issues with reliability it can easily halt the whole network. we need to test it maybe with chaos mesh before pushing it to next devnet

countvonzero:
> @countvonzero does this change effectively enables tortoise beacon? i think it is definitely in a better shape now, but if there are still some issues with reliability it can easily halt the whole network. we need to test it maybe with chaos mesh before pushing it to next devnet

yes, it effectively enables tortoise beacon.

i am not familiar with "test it with chaos mesh". is that a special setup?
my assumption was that it will be tested like the current devnet, running a while with late nodes joining before releasing to the public. and any problems observed we fix them right away.

@moshababo FYI

dshulyak:
> i am not familiar with "test it with chaos mesh". is that a special setup?

it can be used to create various conditions for nodes in the network - https://chaos-mesh.org/

> my assumption was that it will be tested like the current devnet, running a while with late nodes joining before releasing to the public. and any problems observed we fix them right away.

that test by itself is not sufficient to find non-basic issues with joiners. i think that in general, we should plan for advanced testing before merging critical changes, but up to you

countvonzero:
> it can be used to create various conditions for nodes in the network - https://chaos-mesh.org/
>
> > my assumption was that it will be tested like the current devnet, running a while with late nodes joining before releasing to the public. and any problems observed we fix them right away.
> 
> that test by itself is not sufficient to find non-basic issues with joiners. i think that in general, we should plan for advanced testing before merging critical changes, but up to you

thanks for the info @dshulyak this is a great idea. 

specifically for this PR, my thoughts are:
- while i strongly support adding chaos mesh testing to go-spacemesh, it looks like it requires dedicated resource to set up for our nodes. 
- this PR is relatively simple (most code change are adding test) and i'd like to merge it sooner rather than later as the codebase changes rather fast
- this change will expose bugs in current tortoise beacon impl, and failure in that protocol will cause the consensus to stall and cause frustration to our users

so here is my proposal:
- add metrics for different beacon values in the network (done)
- flag-gate this feature. only turn it ON for unit/system tests and OFF for devnet
- there are still a few more PRs to complete the tortosie beacon protocol (in tortoise/self-healing). we can concurrently have resource working on setting up chaos mesh while i finish the rest of the impl. at that time i can do better testing with it.

alternatively, i also feel brave/ignorant enough to run with it and debug as bugs happen :P 
but i don't bear the front of the community's reaction to a failing devnet, it's more fair for @moshababo to make the call

dshulyak:
> while i strongly support adding chaos mesh testing to go-spacemesh, it looks like it requires dedicated resource to set up for our nodes.

it is not complicated. it can be installed with a single command. i was using it to find issues in the last devnets

> alternatively, i also feel brave/ignorant enough to run with it and debug as bugs happen :P

the problem is that if there are unresolved issues they will complicate debugging for other development efforts, in particular for libp2p. in general i don't mind debugging smaller things as we go, but for critical parts that can stall the network it will be very frustrating.

tortoisebeacon/metrics/prometheus.go
```diff
@@ -0,0 +1,37 @@
...
+	LabelBeacon = "beacon"
+)
+
+var (
+	// ObservedBeaconCounter is the total count of each beacon reported by blocks.
+	ObservedBeaconCounter = metrics.NewCounter(
+		"observed_beacon_count",
+		subsystem,
+		"Weight of each beacon collected from blocks for each epoch and value",
+		[]string{"epoch", "beacon"},
```
dshulyak: we can't use labels that grow in time, it is effectively a memory leak. similar to one that we had recently for blocks

dshulyak (changes requested):

tortoisebeacon/metrics/prometheus.go
countvonzero: hmmm.... is it possble to delete old values as we go? this should grow slowly.

or do you have other idea?
i am trying to achieve something like expvar's Map where we can see for a given epoch there are N beacons observed with count/weight.

or if is there a way to dynamically generate a counter at runtime where the counter name contains epoch and beacon value?

countvonzero:
> it is not complicated. it can be installed with a single command. i was using it to find issues in the last devnets

do you think i can try this on the system test. i.e. change the tests to run indefinitely and test the chaos scenario on those 50 nodes?

tortoisebeacon/metrics/prometheus.go
dshulyak: i think [collectors](https://github.com/prometheus/client_golang/blob/v1.11.0/prometheus/collectors/dbstats_collector.go) fits this pattern

dshulyak:
> do you think i can try this on the system test. i.e. change the tests to run indefinitely and test the chaos scenario on those 50 nodes?

should work (i think it is simpler with spacecraft though). maybe disconnecting some fraction of nodes periodically for an extended period of time, so they go out of sync and miss tortoise beacon round, and then checking proposed metrics would be a good test

countvonzero:
bors try

countvonzero:
bors try

tortoisebeacon/metrics/prometheus.go
```diff
@@ -0,0 +1,105 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
```
dshulyak, line 4: it is a bit weird that we are using go-kit everywhere else and prometheus metrics directly here. i don't mind using prometheus library directly, but consistency would be nice

dshulyak (approved):

tortoisebeacon/metrics/prometheus.go
countvonzero: yes. tho go-kit metrics doesn't expose the Exporter/Collector functionality. in the long run it seems easier to use prometheus directly. i'll file an issue for that.

countvonzero:
bors try

countvonzero:
bors try

hare/hare.go
```diff
@@ -240,47 +240,75 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) (err error) {
...
+	blocks := h.getGoodBlocks(h.lastLayer, beacon, logger)
 	logger.With().Info("starting hare consensus with blocks", log.Int("num_blocks", len(blocks)))
 	set := NewSet(blocks)
 
 	instID := id
 	c, err := h.broker.Register(ctx, instID)
 	if err != nil {
 		logger.With().Error("could not register consensus process on broker", log.Err(err))
-		return
+		return false, fmt.Errorf("broker registerr: %w", err)
```
noamnelke:
```suggestion
		return false, fmt.Errorf("broker register: %w", err)
```

hare/hare.go
```diff
@@ -345,8 +373,11 @@ func (h *Hare) tickLoop(ctx context.Context) {
 		select {
 		case layer := <-h.beginLayer:
 			go func() {
-				if err := h.onTick(ctx, layer); err != nil {
+				started, err := h.onTick(ctx, layer)
+				if err != nil {
 					h.WithContext(ctx).With().Error("error processing hare tick", log.Err(err))
+				} else if !started {
+					h.WithContext(ctx).With().Warning("consensus not started for layer", layer)
```
noamnelke, line 380: Why not return this as an error? Seems like we're handling the error and non-error case the same way, so why return an extra boolean?

noamnelke (approved):

countvonzero:
FYI @dshulyak @noamnelke 

here the result of more extensive testing with spacecraft+devnet config (50 nodes)
the network has 20 min epoch. beacon protocol usually takes 8 min 30 seconds

note that i was unable to run chaos-mesh. it seems the chaos-mesh tools couldn't find the runtime container binary. i ended up doing `kubectl get pods -n default --no-headers=true | awk '/miner-4/{print $1}' | xargs kubectl delete pods` to take out 11 pods at a time.

### taking 11 nodes out at the beginning/middle/near the end of tortoise beacon protocol
the network was able to consistently produce the beacon values, and those out-of-sync nodes were able to sync the beacon value via blocks. here's some data
- 50 nodes entered beacon protocol, 39 completed
- in the following rounds 42 nodes enter the protocol, and 42 completed
- took 2 epochs for the network to recover with all 50 nodes participating in the beacon protocol

### taking 22 nodes out at the beginning/middle/near the end of tortoise beacon protocol
very similar to the 11 nodes experiment. 
- 50 nodes entered beacon protocol, 22 completed
- in the following rounds 41 nodes enter the protocol, and 41 completed
- took 4 epochs for the network to recover with all 50 nodes participating in the beacon protocol

### taking 11 nodes out every 2 minutes during the tortoise beacon protocol, total 33 pods got restarted
the network did not recover from this experiment due to sync+hare issues. the network's processed layer stuck at 348.
1 node was able to enter into gossip sync. 32 nodes were not able to sync because none of its peers have processed layer higher than 348. hare produced empty set (`terminating: reached iterations limit`) up to layer 372.

however, the beacon protocol continues to run with the remaining 17 nodes that are still synced. here is the output of each epoch. below is the query result from prometheus `group by(epoch,beacon) (spacemesh_beacons_beacon_calculated_weight)` and `group by(epoch,beacon) (spacemesh_beacons_beacon_observed_total)`


| beacon | epoch |
| ------------- | ------------- |
| aeebad4a79 | 1 |
| aeebad4a79 | 2 |
| 2550f7d0de | 3 |
| 1cc281a72b | 4 |
| 638ec2dd16 | 5 |
| 801398e571 | 6 |
| e4201e7da1 | 7 |
| 29149f6c0a | 8 |
| 3aaecd76da | 9 |
| 170c280455 | 10 |
| 1c1f24fcb8 | 11 |
| 6bda060b4b | 12 |
| cc56e9680f | 13 |
| 9a0bacf88e | 14 |
| 845027bb55 | 15 |
| db235dd4bc | 16 |
| f3562f08ed | 17 |
| 856d7c79b5 | 18 |
| 186ac50f62 | 19 |
| 3da3937d38 | 20 |
| 0a45b1a456 | 21 |
| 075ad1dec4 | 22 |
| e2ab8d2aa9 | 23 |
| 241427139e | 24 |
| 706ac187a9 | 25 |
| c25cef83f7 | 26 |
| a3f6960836 | 27 |
| 1cfe36f972 | 28 |
| bbebce088f | 29 |
| 49ecdfdba0 | 30 |
| acc7809206 | 31 |
| dc1773e02c | 32 |
| 6aac891b53 | 33 |
| f2e42375da | 34 |
| 93f85502f4 | 35 |
| 9ab166183d | 36 |
| 7409e4f848 | 37 |
| 81233bfc96 | 38 |
| d30ba3be9f | 39 |
| 5883d8d668 | 40 |
| c3002d18a8 | 41 |
| 776889d848 | 42 |
| 7f0d138720 | 43 |
| 03cd7bc927 | 44 |
| df0a21aee4 | 45 |
| 76c2196efb | 46 |
| ac65753bad | 47 |
| cb895cb1a3 | 48 |
| 93f7775a27 | 49 |
| 9e4bf683c1 | 50 |
| b8760e36e3 | 51 |
| f63d437fbd | 52 |
| 947594119e | 53 |
| 69a6390bba | 54 |
| c36b90d1aa | 55 |
| 8aa5676026 | 56 |
| 0465d69b61 | 57 |
| 7a845f26f5 | 58 |
| 55ce829d99 | 59 |
| cca8250981 | 60 |
| 8385d30ff7 | 61 |
| dbedfe23f4 | 62 |
| 0daf714a7e | 63 |
| 5e1d436d10 | 64 |
| 01a4a425d7 | 65 |
| e613be85b8 | 66 |
| 7e7698e36a | 67 |
| 83a8827734 | 68 |
| b901ec1356 | 69 |
| e21661fd56 | 70 |
| 6d1370ea64 | 71 |
| 4129555c90 | 72 |
| 5a522eb873 | 73 |
| 8c639223c2 | 74 |
| 744e740af1 | 75 |
| 7aa071a7d2 | 76 |
| ef086080b5 | 77 |
| d05f5da1b8 | 78 |
| 2c89763dab | 79 |
| 99fe5b774b | 80 |
| f90fb46b4c | 81 |
| 651b0e7ec7 | 82 |
| 370e105cd7 | 83 |
| 0eeef9cb6f | 84 |
| 3138da63f2 | 85 |
| 6d5a2850c8 | 86 |
| afd4010d85 | 87 |
| 6e8ba1f8b2 | 88 |
| e32dab4b7e | 89 |
| 90df90bc73 | 90 |
| 34bfc07523 | 91 |
| f0eee015e7 | 92 |
| f6a2d482f3 | 93 |
| c86f94a570 | 94 |
| f86e3e88db | 95 |
| e627a00772 | 96 |
| 6215a2c845 | 97 |
| c7cf62bc9f | 98 |
| b6e63df560 | 99 |
| 38f1d6d8e2 | 100 |
| 8bfa23f38e | 101 |
| 84470f6bff | 102 |
| 91a9e5c4ec | 103 |
| ecfecc9660 | 104 |
| e912ef790d | 105 |
| 788b8aebd0 | 106 |
| e446942efa | 107 |
| 8965acb16a | 108 |
| e7e4aa2967 | 109 |
| 8fe51571f8 | 110 |

hare/hare.go
countvonzero: this is purely for testing and debugging.
it's easier to test individual calls to `onTick` instead of starting `tickerLoop`. and there are legit situations where we don't return an error but didn't participate in the protocol.
- genesis
- node shutting down
- **new behavior and hence needs testing**: node has not collected enough beacon values to participate in the protocol yet.

countvonzero:
bors merge

PR #2870, opened 2021-10-12
Start a SM network with preset data

AmitShaul:
## Motivation
We aim to shorten tests duration by loading a preset data to the nodes, bootstrap and PoET thus we can start a test from a certain point of time instead of starting the SM network and waiting for it to arrive.

## Changes

- Added a new test file under `tests` folder called `state_recorder` with a single test inside called `test_recorder`. Its goal is to start a `Spacemesh` network, wait until a certain point in time (set by the user), stopping all nodes including the poet and saving all relevant data under `Google Cloud Storage`.
- Added a new `mining` test that utilizes `test_recorder` output by loading the data both to the clients, bootstrap and poet before starting them. `setup_clients_with_state` fixture was added for that matter.

related issues:
https://github.com/spacemeshos/pm/issues/87

related PR:
[PR#108](https://github.com/spacemeshos/poet/pull/108)

## TODO
- [ ] Pass arguments to the PoET, at the moment, for debugging purposes the arguments for the PoET are edited in the PoETs `k8s` file (e.g., args: [..., '--data-paths', 'DATA_PATH', '--state-bucket', 'BUCKET_NAME']).
- [ ] Add the capability to save some and not all of the data in the node (following @y0sher suggestion to leave `peers.json` out to overcome `p2p` and `sync` issues).
- [ ] Right now `test_recorder` can only be ran, and finish successfully, by its make command, if we'd like to run it in the future using `pytest` command as well we should add `GOOGLE_APPLICATION_CREDENTIALS` env variable to the environment the tests are ran on. `GOOGLE_APPLICATION_CREDENTIALS` should contain our `gcloud service account` credentials file path.

## DevOps Notes


dshulyak:
@AmitShaul am i correct that if the go-sm internal state changes we will be able to regenerate it easily?

y0sher:
@dshulyak , I believe this is the `state_recorder`. it'll take another network run and uploading to google cloud storage..

Makefile
```diff
@@ -263,6 +264,16 @@ endif
 dockertest-mining: dockerbuild-test dockerrun-mining
 .PHONY: dockertest-mining
 
+dockerrun-mining-preset-state:
+ifndef ES_PASS
```
y0sher, line 268: we should probably take this out to another recipe and run it on any that requires this var

cmd/integration/harness.go
```diff
@@ -2,6 +2,8 @@ package main
 
 import (
 	"fmt"
+	"github.com/spacemeshos/go-spacemesh/common/util"
+	"github.com/spacemeshos/go-spacemesh/log"
```
y0sher, line 6: this is a double import, don't know how you compiled it

PR #2873, opened 2021-10-14
[Merged by Bors] - sync: stayed synced despite sync errors

### LINKED ISSUE #2872: sync: if hare takes longer than a layer to terminate node will go out of sync
## Description

In sync we wait for one additional layer before requesting this layer from peers. If hare takes longer then expected to terminate (or if there are any other performance issues) processed layer won't be updated on time and peer will go out of sync. It appears that we can't go out of sync in such case, since sync status affects every other protocol and the whole network will get stuck and won't be able to recover.

We need a better condition for changing sync status from unsynced to sync, so that it both:
- prevents downloading potentially invalid state
- doesn't prevent network from recovering

## Steps to reproduce
In devnet207 hare for layer 3375 took to long to terminate. 3376 was started at 11:04:50, but hare for 3375 terminated only at 11:06:00.502. Therefore at the time when hare for 3776 was supposed to be started, all nodes went out of sync and didn't participate in hare.

countvonzero:
## Motivation

Closes #2872

## Changes

- stay in synced state despite failure to sync layers from peers, unless the node didn't receive gossip data for 3 layers
- syncer do NOT validate a layer if hare is already running consensus protocol for that layer

## Test Plan

unit test

## DevOps Notes


countvonzero:
@dshulyak the layer patrol struct i added may also be suitable for the coordination component you mentioned between hare and tortoise

countvonzero:
bors try

dshulyak (approved):
cant say that i like layerpatrol name but can't think of any better :)

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/3893797657)

filed https://github.com/spacemeshos/go-spacemesh/issues/2874

countvonzero:
bors try

countvonzero:
update: 

mining system test fails more often with this PR (thank goodness for system tests!)
the issue is:

- short:

hare failed to validate a layer, and syncer does not step in to validate, the node gets stuck.

- long: 

in one of my runs, 3 out of 50 nodes's processed layers get stuck at layer 9 while 47 nodes advanced to layer 24. a block in layer 9 failed syntactic validation during gossip with this error `failed to fetch ref block ebe041945b e: fetch failed after max retries for request`. the ref block is published in the same layer (9) but somehow got propagated over the network later than the subsequent blocks. in any case, when hare terminated, and this block is in the output set, `HandleValidatedLayer` errored out `hare terminated with block that is not present in mesh`. 

with this PR, these 3 nodes stayed synced, but the processed layer stuck at 8 because syncer doesn't try to validate the layer again. whereas before this PR, these 3 nodes will go out-of-sync, syncer takes over and become synced again. (actually the node being synced or not does not matter in this case.)

mining system test randomly selects node to verify transaction and balance at various layers. 1 of these 3 nodes were selected and failed the test.

- fixes
  - added a safeguard in syncer to still validate a layer if it's too old/delayed, despite hare promising to validate it.
  the number of layers to wait 10 was chosen because the max number of hare iteration for mainnet is 40 (~10 layers).
  for simplicity, i didn't add a param for this. however, this number is too large for mining system test because it tests at various layers. therefore, i also
  - increased the number of fetch retries from 20 to 100.

these two fixes combined stabilize the mining system tests. i ran locally 5 times and all passed.

FYI @dshulyak @moshababo

countvonzero:
bors try

dshulyak:
> hare terminated with block that is not present in mesh

this error was popping up a lot in the last devnet. i looked up the origin of one block and it wasn't from cloud nodes. tbh this looks like a severe issue if it can occur even in cloud nodes, as there is almost no latency between nodes

countvonzero:
bors try

countvonzero:
> > hare terminated with block that is not present in mesh
> 
> this error was popping up a lot in the last devnet. i looked up the origin of one block and it wasn't from cloud nodes. tbh this looks like a severe issue if it can occur even in cloud nodes, as there is almost no latency between nodes

filed https://github.com/spacemeshos/go-spacemesh/issues/2876 to formally document this. yes. i agree this is a rather severe issue.

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/3902767117)

same issue as #2874

countvonzero:
bors try

countvonzero:
bors merge

PR #2889, opened 2021-10-26
[Merged by Bors] - p2p: enable logging for libp2p components

dshulyak:
## Motivation

libp2p logging must be enabled.

in test mode libp2p logs will look like:

```
{"L":"DEBUG","T":"2021-10-26T06:43:15.889+0300","N":"basichost","C":"basic/basic_host.go:370","M":"protocol negotiation took 183.635µs","node_id":"1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e","module":"p2p"}
{"L":"DEBUG","T":"2021-10-26T06:43:15.889+0300","N":"basichost","C":"basic/basic_host.go:370","M":"protocol negotiation took 140.061754ms","node_id":"1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e","module":"p2p"}
{"L":"DEBUG","T":"2021-10-26T06:43:15.889+0300","N":"pubsub","C":"go-libp2p-pubsub@v0.5.4/gossipsub.go:1354","M":"HEARTBEAT: Add mesh link to 12D3KooWAsL38r9L1qc3yL6boyWLAAQka9YyfaoHjVi78tFyPEoW in HARE_PROTOCOL","node_id":"1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e","module":"p2p"}
{"L":"DEBUG","T":"2021-10-26T06:43:15.890+0300","N":"pubsub","C":"go-libp2p-pubsub@v0.5.4/gossipsub.go:1354","M":"HEARTBEAT: Add mesh link to 12D3KooWB7oDmnpuXCd1APR4chrkAPkz3BDXyaAtKQmBDdH79FCn in HARE_PROTOCOL","node_id":"1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e","module":"p2p"}
{"L":"DEBUG","T":"2021-10-26T06:43:15.890+0300","N":"pubsub","C":"go-libp2p-pubsub@v0.5.4/gossipsub.go:1354","M":"HEARTBEAT: Add mesh link to 12D3KooW9qnhAVbbLqqPmArtJx2erjD8b85JvaSzow1XHwNLLaUw in HARE_PROTOCOL","node_id":"1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e","module":"p2p"}
```

## Changes
- set log level when host is initialized
- create loggers map by decoding mapstructure tags

## Test Plan
tested format manually

## TODO

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -374,7 +374,7 @@ func (app *App) addLogger(name string, logger log.Log) log.Log {
 	lvl := zap.NewAtomicLevel()
 	loggers, err := decodeLoggers(app.Config.LOGGING)
 	if err != nil {
-		panic("unable to decode loggers into map[string]string")
+		app.log.With().Panic("unable to decode loggers into map[string]string", log.Err(err))
 	}
 	level, ok := loggers[name]
 	if ok {
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -392,12 +392,12 @@ func (app *App) addLogger(name string, logger log.Log) log.Log {
 	return logger.WithName(name).WithFields(log.String("module", name))
 }
 
-func (app *App) getLevel(name string) (log.Level, bool) {
+func (app *App) getLevel(name string) log.Level {
 	alvl, exist := app.loggers[name]
 	if !exist {
-		return 0, false
+		return 0
 	}
-	return alvl.Level(), true
+	return alvl.Level()
 }
 
 // SetLogLevel updates the log level of an existing logger.
@@ -1093,11 +1093,8 @@ func (app *App) Start() error {
 	cfg := app.Config.P2P
 	cfg.DataDir = filepath.Join(app.Config.DataDir(), "p2p")
 	p2plog := app.addLogger(P2PLogger, lg)
-	var exist bool
-	cfg.LogLevel, exist = app.getLevel(P2PLogger)
-	if !exist {
-		lg.Panic("p2p logger level is not set after calling app.addLogger")
-	}
+	// if addLogger won't add a level we will use a default 0 (info).
+	cfg.LogLevel = app.getLevel(P2PLogger)
 	app.host, err = p2p.New(ctx, p2plog, cfg,
 		p2p.WithNodeReporter(events.ReportNodeStatusUpdate),
 	)
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -29,7 +29,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/common/util"
 	"github.com/spacemeshos/go-spacemesh/config"
-	cfg "github.com/spacemeshos/go-spacemesh/config"
 	"github.com/spacemeshos/go-spacemesh/database"
 	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/fetch"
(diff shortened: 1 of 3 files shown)


dshulyak:
@countvonzero @nkryuchkov can you take a look please?

cmd/node/node.go
```diff
@@ -1110,8 +1092,15 @@ func (app *App) Start() error {
 
 	cfg := app.Config.P2P
 	cfg.DataDir = filepath.Join(app.Config.DataDir(), "p2p")
-	app.host, err = p2p.New(ctx, app.addLogger(P2PLogger, lg), cfg,
-		p2p.WithNodeReporter(events.ReportNodeStatusUpdate))
+	p2plog := app.addLogger(P2PLogger, lg)
+	var exist bool
+	cfg.LogLevel, exist = app.getLevel(P2PLogger)
+	if !exist {
```
countvonzero: it's odd to only check p2p logger's existence. why not add to the config instead?

countvonzero (approved):

p2p/host.go
```diff
@@ -58,7 +59,10 @@ func New(ctx context.Context, logger log.Log, cfg Config, opts ...Opt) (*Host, e
 	if err != nil {
 		return nil, err
 	}
+	// what we set in cfg.LogLevel doesn't will not be applied
```
nkryuchkov: `doesn't will` looks like a typo

nkryuchkov (approved):

cmd/node/node.go
dshulyak: it is already a part of logging config, the addLogger takes log level from config, parses it and adds a log level to internal map. the part with internal map seemed like a side effect, so i added this panic to ensure that it is not removed accidentally, but maybe i can just rely on existing behavior and do not overcomplicate

dshulyak:
bors merge

cmd/node/node.go
countvonzero: yeah. also with this change we are decoding to a map each time when `addLogger` is called. do you think it's possible to just do it once?

dshulyak: i am not sure if matters, it is possible though. computational complexity of this decoding operation is not much higher compared to a previous approach, in the worst case it will add milliseconds to app initialization time.

p2p/host.go
```diff
@@ -58,7 +59,10 @@ func New(ctx context.Context, logger log.Log, cfg Config, opts ...Opt) (*Host, e
 	if err != nil {
 		return nil, err
 	}
+	// what we set in cfg.LogLevel will not be used
+	// unless level of the Core is atleast as high
 	lp2plog.SetPrimaryCore(logger.Core())
```
countvonzero, line 64: what happens if some new component call SetPrimaryCore() again after p2p? should we simply SetPrimaryCore to the lowest level in the config once it's parsed?

dshulyak: i added it to p2p initialization so that other components outside p2p module won't touch this. i definitely don't want to move it to the cmd/node.go initialization as it is already a mess. 

also i want libp2p logger to inherit the same logging fields as p2p module, so setting it in p2p module seems natural

dshulyak:
kibana failure https://github.com/spacemeshos/go-spacemesh/runs/4020942620?check_suite_focus=true

dshulyak:
bors merge

PR #2890, opened 2021-10-26
[Merged by Bors] - p2p: select external port only from routable addresses

dshulyak:
## Motivation

When an external port is re-configured we need to prefer a routable address. This was missing in the initial change.

```
2021-10-26T08:58:27.433+0300	INFO	15671.p2p          	host addresses are updated	{"node_id": "1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e", "module": "p2p", "address": "/ip4/109.86.198.208/tcp/45846", "name": "p2p"}
```

## Changes

- select routable address when external port is configured
- negate NatPort configuration so that it is enabled by default, and needs to be disabled with --no-natport

## Test Plan
unit tests and manual tests with router that supports upnp

## TODO

## DevOps Notes


cmd/root.go
```diff
@@ -70,8 +70,8 @@ func AddCommands(cmd *cobra.Command) {
 		config.P2P.Listen, "address for listening")
 	cmd.PersistentFlags().BoolVar(&config.P2P.Flood, "flood",
 		config.P2P.Flood, "flood created messages to all peers (true by default. disable to lower traffic requirements)")
-	cmd.PersistentFlags().BoolVar(&config.P2P.NatPort, "natport",
-		config.P2P.NatPort, "enable port-mapping on the router (upnp must be enabled for this to work)")
+	cmd.PersistentFlags().BoolVar(&config.P2P.NoNatPort, "no-natport",
```
countvonzero: using negative in code is confusing. it further leads to the use of double-negative and make code trickier to read/understand.

p2p/host.go
```diff
@@ -84,7 +84,7 @@ func New(ctx context.Context, logger log.Log, cfg Config, opts ...Opt) (*Host, e
 		libp2p.ConnectionManager(cm),
 		libp2p.Peerstore(pstoremem.NewPeerstore()),
 	}
-	if cfg.NatPort {
+	if !cfg.NoNatPort {
```
countvonzero: this is difficult for me to understand. what's the difference btwn this and previous code? it seems you don't need to change the flag at all...

p2p/peerexchange/discovery.go
```diff
@@ -55,37 +56,30 @@ func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
...
 		return nil, err
 	}
 	d.book.AddAddresses(bootnodes, best)
-	port, err := portFromAddress(h)
-	if err != nil {
-		return nil, err
-	}
-	protocol := newPeerExchange(h, d.book, port, logger)
+
+	protocol := newPeerExchange(h, d.book,
```
countvonzero: nit: either one arg per line or all in one line

p2p/peerexchange/discovery.go
```diff
@@ -105,63 +99,90 @@ func (d *Discovery) Bootstrap(ctx context.Context) error {
...
+}
+
+func portFromHost(logger log.Log, h host.Host) uint16 {
+	addr, err := bestNetAddress(h)
+	if err != nil {
+		logger.With().Warning("failed to find best host address. host won't be dialable", log.Err(err))
+		return 0
 	}
-	return best, nil
+	logger.With().Info("host addresses are updated", log.String("address", addr.String()))
```
countvonzero: i don't understand this. how do you determine that host address is updated?

p2p/peerexchange/discovery.go
```diff
@@ -105,63 +99,90 @@ func (d *Discovery) Bootstrap(ctx context.Context) error {
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to cast addr %s to netaddr: %w", addr, err)
+	}
+	_, portStr, err := net.SplitHostPort(netaddr.String())
+	if err != nil {
+		return 0, fmt.Errorf("parsed netaddr %s is not in expected format: %w", netaddr, err)
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return 0, fmt.Errorf("failed to convert port %s to int", portStr)
```
countvonzero: wrap err?

countvonzero (commented):
apologies for lack of knowledge here. i in general don't understand this change. can you please explain
- how is the external port reconfigured
- and how does the node detect that the port is reconfigured

cmd/root.go
dshulyak: i don't agree with that. this change allows to set `--no-natport` instead of `--natport=false`, and omit the option if you don't need to change default behavior

p2p/host.go
dshulyak: i dont understand the question. the difference is that if you don't set NoNatPort to true we will enable upnp

p2p/peerexchange/discovery.go
dshulyak: by subscribing to events

dshulyak:
>  how is the external port reconfigured

libp2p has several tools to discover that port was updated. we are using upnp port mapping that runs in background. once it completes libp2p host will send an event

> and how does the node detect that the port is reconfigured

in discovery.New i am subscribing to events that are fired when new address is discovered

p2p/host.go
dshulyak: previously you had to set NatPort to true for this to work on the client side, but now the default behavior doesn't require explicitly setting NatPort to true. so user will have to opt-out if for some reason nat port mapping is not desirable. I think the only reason if someone wants to opt-out if they prefer to make the node undialable.

cmd/root.go
countvonzero: when i read `if !cfg.NoNatPort`, i'm sounding out `if not no nat port`. that's what i mean by double-negative and is a mind-twister for me. 

i usually see when a binary has --no-something, it has an arg named `something` and by prepending `no-` or `no` to that arg the custom parser changed the arg's value to false. so in code it still reads `if something {`

how about --disable-natport?

p2p/host.go
countvonzero: understood. thanks

p2p/peerexchange/discovery.go
countvonzero: right. but you also call this in `newPeerExchange` so it's slightly confusing. i we should log this when it receive the event.

countvonzero (approved):

countvonzero:
> > and how does the node detect that the port is reconfigured
> 
> in discovery.New i am subscribing to events that are fired when new address is discovered

that was a stupid question on my part.

cmd/root.go
dshulyak: sure will change, i was considering disable as well

p2p/peerexchange/discovery.go
dshulyak: i will change the log to note that we selected this address as the "best" routable address

dshulyak:
bors merge

PR #2892, opened 2021-10-27
[Merged by Bors] - miner: differentiate database.ErrNotFound when lookup ref block

countvonzero:
## Motivation

make sure we don't create multiple ref blocks on disk error

## Changes

- block builder: differentiate btwn database.ErrNotFound and other disk errors when looking up refblock
- block builder: make failure saving ref block an error

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -16,14 +16,14 @@ const (
 	namespaceEvict    = "e"
 	namespaceVerified = "v"
 	namespaceGood     = "g"
-	namespaceOpinons  = "o"
+	namespaceOpinions = "o"
 )
 
 type state struct {
 	log log.Log
 	db  database.Database
 
-	// if true only non-flushed items will be synced
+	// if true only non-flushed items will be persisted
 	diffMode bool
 
 	// last layer processed: note that tortoise does not have a concept of "current" layer (and it's not aware of the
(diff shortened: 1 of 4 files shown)


tortoise/state.go
```diff
@@ -120,16 +120,20 @@ func (s *state) Recover() error {
 	s.LastEvicted = types.BytesToLayerID(buf)
 
 	it := s.db.Find([]byte(namespaceGood))
-	defer it.Release()
+	defer func(iter database.Iterator) {
```
dshulyak: what is a concern here? i think both Release's will be executed

dshulyak: i made a small snippet

```
type obj struct {
	val string
}

func (o obj) Release() {
	fmt.Println(o.val)
}

func TestRelease(t *testing.T) {
	o := &obj{"dsa"}
	defer o.Release()
	o = &obj{"aaa"}
	defer o.Release()
}
```

will print

```
=== RUN   TestRelease
aaa
dsa
```

i think that pointers to `o` and `it` are copied for defer

dshulyak: in your change they are not not copied, because function accepts `iter`, but the call is executed on `it`. but i think this needs to be reverted

dshulyak (changes requested):

tortoise/state.go
countvonzero: i am embarrassed. you are right. somehow i experimented it yesterday and thought i got a different result. sorry.

dshulyak (approved):

countvonzero:
bors merge

PR #2895, opened 2021-10-27
Implement `HandleGossipTransaction` (#2833)

### LINKED ISSUE #2833: Implement `HandleGossipTransaction` in SVM package
```go
func (s *SVM) HandleGossipTransaction(ctx context.Context, data service.GossipMessage, syncer service.Fetcher)
```

Wrapper around [`HandleTransaction`](https://github.com/spacemeshos/go-spacemesh/issues/2834) that calls `data.ReportValidation(ctx, IncomingTxProtocol)` if successful.

### Called From
- `cmd/node/node.go` -> `App.initServices`
https://github.com/spacemeshos/go-spacemesh/blob/c66afab4f8f3f4886ccb7b577da71fa590e82348/cmd/node/node.go#L759
(the method is called `HandleTxGossipData` today)

kobby-pentangeli:
## Motivation

Closes #2833 

## TODO

## DevOps Notes

### DIFF
--- a/svm/api.go
+++ b/svm/api.go
@@ -1,12 +1,15 @@
 package svm
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/spacemeshos/go-spacemesh/api/config"
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/common/util"
 	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/p2p/pubsub"
 	"github.com/spacemeshos/go-spacemesh/svm/state"
 )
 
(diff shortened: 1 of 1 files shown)


svm/api.go
```diff
@@ -1,12 +1,15 @@
 package svm
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/libp2p/go-libp2p-core/peer"
```
dshulyak, line 7: please try to use p2p.Peer and avoid importing libp2p libs outside of the p2p module, i might be doing it in some places but in general this is something i was trying to avoid

dshulyak (commented):
is this PR meant as a placeholder or it should be functional? if the latter there are some parts missing, it needs actual code that handles tx's and the handler needs to be registered in place of the old one.

svm/api.go
kobby-pentangeli: The follow-up PR (https://github.com/spacemeshos/go-spacemesh/pull/2896) actually handles transactions. So, yes, this PR is a placeholder.

Concerning the use of `libp2p`, I'm making the changes now.

kobby-pentangeli:
Obsoleted by https://github.com/spacemeshos/go-spacemesh/pull/2896

Closing this PR because it does not truly capture the full implementation of `SVM.HandleTransaction`.

Issue #2834, opened 2021-09-26
Implement `HandleTransaction` in SVM package


noamnelke:
```go
func (s *SVM) HandleTransaction(transaction []byte)
```
Called from sync, and should be called from [`HandleGossipTransaction`](https://github.com/spacemeshos/go-spacemesh/issues/2833).

Should be idempotent, so we can also use it to re-insert transactions from invalid blocks to the mempool.

### Called From
The function is currently named `HandleTxData`.
- Currently it's only called from `HandleTxGossipData`
- We should fold `HandleTxSyncData` into this function - figure out why it doesn't validate that the origin exists today.

### Existing Functions
- `TransactionProcessor.HandleTxData`
- `Mempool.Put`

### SVM Functions
- `ValidateDeploy`
- `ValidateSpawn`
- `ValidateCall`
- `Verify`

PR #2896, opened 2021-10-27
[Merged by Bors] - Implement `HandleTransaction` in SVM Package

kobby-pentangeli:
## Motivation

This PR closes both #2833 and #2834 

## Changes

There is a common `HandleTransaction` that receives a `types.Transaction` and a `validateOrigin` boolean. Then there is `HandleTxSyncData` that deserializes a `mesh.DbTransaction` and calls `HandleTransaction(dbTx.Transaction, false)`. There is also `HandleGossipTransaction` that deserializes a `types.Transaction` and calls `HandleTransaction(tx, true)`.

## Test Plan

Unit tests for `HandleTransaction` have been added in `api_test.go`.

## DevOps Notes


countvonzero (commented):
@noamnelke 
svm/api.go has no test. it seems the added code presents a good opportunity to set up test for this file.

it should be very simple to use gomock to mock out svm.state package at this level and test the app logic.

kobby-pentangeli:
bors try

noamnelke:
bors try

kobby-pentangeli:
bors try

svm/api.go
```diff
@@ -134,3 +138,67 @@ func (svm *SVM) ValidateAndAddTxToPool(tx *types.Transaction) error {
...
+	}
+	if err = svm.HandleTransaction(tx.Transaction, false); err != nil {
+		return fmt.Errorf("handle transaction: %w", err)
+	}
+	return nil
+}
+
+// HandleTransaction handles data received on transaction gossip channel.
+// It allows for transactions from both valid and invalid blocks to be added to the mempool.
+func (svm *SVM) HandleTransaction(tx *types.Transaction, validateOrigin bool) error {
```
noamnelke: I don't think we should export this method (in Go, functions that start with capital letters are made available outside the package). This should be private.

noamnelke (changes requested):

noamnelke:
bors try

svm/api.go
kobby-pentangeli: Understandable. I exported it because the function signature (as described in the underlying issue) appears so.

kobby-pentangeli:
bors try

kobby-pentangeli:
bors try

svm/api.go
```diff
@@ -134,3 +138,67 @@ func (svm *SVM) ValidateAndAddTxToPool(tx *types.Transaction) error {
...
+	}
+	if err = tx.CalcAndSetOrigin(); err != nil {
+		return fmt.Errorf("calculate and set origin: %w", err)
+	}
+	svm.state.AddTxToPool(tx.Transaction)
+	return nil
+}
+
+// handleTransaction handles data received on transaction gossip channel.
+func (svm *SVM) handleTransaction(tx *types.Transaction, validateOrigin bool) error {
```
noamnelke: Because this is now only ever called from one place (`HandleGossipTransaction`) we can just inline this code there - I don't see the advantage of extracting this logic anymore.

There's also no longer a need for the `validateOrigin` argument which is always set to `true`.

svm/api.go
```diff
@@ -134,3 +138,67 @@ func (svm *SVM) ValidateAndAddTxToPool(tx *types.Transaction) error {
...
+	if err = svm.handleTransaction(tx, true); err != nil {
+		svm.state.With().Error("handle transaction", log.Err(err))
+		return pubsub.ValidationIgnore
+	}
+
+	return pubsub.ValidationAccept
+}
+
+// HandleTxSyncData handles data received on transaction sync.
+func (svm *SVM) HandleTxSyncData(data []byte) error {
```
noamnelke: Let's switch to more consistent naming: since we have `HandleGossipTransaction`, let's call this `HandleSyncTransaction`.

svm/api.go
```diff
@@ -134,3 +138,67 @@ func (svm *SVM) ValidateAndAddTxToPool(tx *types.Transaction) error {
...
+
+	if err = svm.handleTransaction(tx, true); err != nil {
+		svm.state.With().Error("handle transaction", log.Err(err))
+		return pubsub.ValidationIgnore
+	}
+
+	return pubsub.ValidationAccept
+}
+
+// HandleTxSyncData handles data received on transaction sync.
```
noamnelke:
```suggestion
// HandleTxSyncData handles transactions received via sync. Unlike HandleGossipTransaction, which only stores valid 
// transactions, HandleTxSyncData only deserializes transactions and stores them regardless of validity. This is because 
// transactions received via sync are necessarily referenced somewhere meaning that we must have them stored, even if 
// they're invalid, for the data availability of the referencing block.
```

noamnelke (changes requested):
Good job! Only some minor changes and we can merge this.

kobby-pentangeli:
bors try

svm/api.go
```diff
@@ -139,43 +139,17 @@ func (svm *SVM) ValidateAndAddTxToPool(tx *types.Transaction) error {
 	return nil
 }
 
-// HandleGossipTransaction wraps around HandleTransaction,
-// which handles data received on the transactions gossip channel.
+// HandleGossipTransaction wraps handles data received on the transactions gossip channel.
```
noamnelke:
```suggestion
// HandleGossipTransaction handles data received on the transactions gossip channel.
```

noamnelke (approved):

kobby-pentangeli:
bors merge

Issue #2852, opened 2021-09-30
tortoise: rerun is too slow to be executed synchronously


dshulyak:
## Description

On devnet205 rerun was started when mesh reached ~3000 layers. In 3h rerun made progress up to ~2000 layer. 

It is executed synchronously while holding a ThreadSafeVerifyingTortoise mutex. It means that for the whole rerun duration hare's outputCollectionLoop is blocked. As a result, the network is stuck for the whole rerun duration.

## Expected Behavior
Rerun is not blocking a network.

countvonzero:
the frequency is decided by `--tortoise-rerun-interval` currently set to 7 days in devnet.
i asked to change it from the original value 1 year. the intention was to flush out the issues and decide what the next step should be. 

my view is that this is a good safety net for bugs in the codebase and should be up and running correctly. but if people feel differently, i guess we can reset it back to 1yr. tho i feel that code doesn't get executed ever should simply be removed.

dshulyak:
It is not usable in the current state, if it takes ~5 hours after 7 days, it will take around a day after a month. Clearly can't block network for that long. It needs to be moved into the background, executed without lock and swap with the original tortoise at some point. 

However, another problem is that it doubles memory consumption. The pointer to the original tortoise state is released only after rerun completes, so effectively the heap size got doubled. It shouldn't be a problem by itself but if there is another issue in place (like with metrics in last devnet) the nodes will simply die.

I agree that not executed code can be as well removed, but i don't know what is the best way to proceed here. I guess it just needs to be a high-priority issue to fix.

dshulyak:
Forgot to close it with the GH.

PR #2897, opened 2021-10-28
[Merged by Bors] - tortoise: optimize processing blocks

dshulyak:
## Motivation

Processing blocks takes too much time because of the excessive reads from database and slow decoding.

With this change, it takes ~20 seconds to re-run mesh with ~2600 layers. The same amount on cloud miners was taking 3-4 hours to rerun.

```
20004:2021-10-28T14:00:43.848+0300	INFO	15671.trtl         	triggering tortoise full rerun from genesis	{"node_id": "1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e", "module": "trtl", "sessionId": "acdac760-8913-4c91-adff-0893407e542f", "layer_id": 2654, "name": "trtl"}
38630:2021-10-28T14:01:02.651+0300	INFO	15671.trtl         	full rerun completed	{"node_id": "1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e", "module": "trtl", "sessionId": "acdac760-8913-4c91-adff-0893407e542f", "layer_id": 2654, "name": "trtl"}
38842:2021-10-28T14:02:48.210+0300	INFO	15671.trtl         	triggering tortoise full rerun from genesis	{"node_id": "1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e", "module": "trtl", "sessionId": "acdac760-8913-4c91-adff-0893407e542f", "layer_id": 2655, "name": "trtl"}
57507:2021-10-28T14:03:07.446+0300	INFO	15671.trtl         	full rerun completed	{"node_id": "1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e", "module": "trtl", "sessionId": "acdac760-8913-4c91-adff-0893407e542f", "layer_id": 2655, "name": "trtl"}
57652:2021-10-28T14:04:49.284+0300	INFO	15671.trtl         	triggering tortoise full rerun from genesis	{"node_id": "1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e", "module": "trtl", "sessionId": "acdac760-8913-4c91-adff-0893407e542f", "layer_id": 2656, "name": "trtl"}
76341:2021-10-28T14:05:07.825+0300	INFO	15671.trtl         	full rerun completed	{"node_id": "1567169fa686ce24fa13d01390d1269d7603078fc2c82b74705dd5ba5435ec7e", "module": "trtl", "sessionId": "acdac760-8913-4c91-adff-0893407e542f", "layer_id": 2656, "name": "trtl"}
``` 

There are actually not so many bottlenecks left, all of them related to the number of opinions. I think that once we have a unified block we will be able to drop the persistent state for tortoise completely, and recompute tortoise state on node restart.

This change doesn't have new benchmarks, as there are not good utilities to build good benchmarks, i will add that later. If someone wants i can share profiles. 

related: https://github.com/spacemeshos/go-spacemesh/issues/2852

## Changes
- keep a map to a layer id for blocks in sliding window
- allocate opinions map based on the size of exceptions and opinions in base block
- comment out very slow log 

## Test Plan
existing tests and manual tests

## TODO

## DevOps Notes

### DIFF
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -37,6 +37,7 @@ type state struct {
 	GoodBlocksIndex map[types.BlockID]bool
 	// use 2D array to be able to iterate from latest elements easily
 	BlockOpinionsByLayer map[types.LayerID]map[types.BlockID]Opinion
+	HaveOpinions         map[types.BlockID]struct{}
 }
 
 func (s *state) Persist() error {
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -37,7 +37,8 @@ type state struct {
 	GoodBlocksIndex map[types.BlockID]bool
 	// use 2D array to be able to iterate from latest elements easily
 	BlockOpinionsByLayer map[types.LayerID]map[types.BlockID]Opinion
-	HaveOpinions         map[types.BlockID]types.LayerID
+	// BlockLayer stores reverse mapping from BlockOpinionsByLayer
+	BlockLayer map[types.BlockID]types.LayerID
 }
 
 func (s *state) Persist() error {
@@ -101,6 +102,7 @@ func (s *state) Persist() error {
 func (s *state) Recover() error {
 	s.GoodBlocksIndex = map[types.BlockID]bool{}
 	s.BlockOpinionsByLayer = map[types.LayerID]map[types.BlockID]Opinion{}
+	s.BlockLayer = map[types.BlockID]types.LayerID{}
 
 	buf, err := s.db.Get([]byte(namespaceLast))
 	if err != nil {
@@ -135,6 +137,8 @@ func (s *state) Recover() error {
 		layer := decodeLayerKey(it.Key())
 		offset := 1 + types.LayerIDSize
 		block1 := decodeBlock(it.Key()[offset : offset+types.BlockIDSize])
+
+		s.BlockLayer[block1] = layer
 
 		if _, exist := s.BlockOpinionsByLayer[layer]; !exist {
 			s.BlockOpinionsByLayer[layer] = map[types.BlockID]Opinion{}
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -37,7 +37,7 @@ type state struct {
 	GoodBlocksIndex map[types.BlockID]bool
 	// use 2D array to be able to iterate from latest elements easily
 	BlockOpinionsByLayer map[types.LayerID]map[types.BlockID]Opinion
-	HaveOpinions         map[types.BlockID]struct{}
+	HaveOpinions         map[types.BlockID]types.LayerID
 }
 
 func (s *state) Persist() error {
--- a/tortoise/verifying_tortoise.go
+++ b/tortoise/verifying_tortoise.go
@@ -118,6 +118,7 @@ func newTurtle(
 			log:                  lg,
 			GoodBlocksIndex:      map[types.BlockID]bool{},
 			BlockOpinionsByLayer: map[types.LayerID]map[types.BlockID]Opinion{},
+			HaveOpinions:         map[types.BlockID]struct{}{},
 		},
 		logger:          lg.Named("turtle"),
 		Hdist:           hdist,
@@ -163,6 +164,7 @@ func (t *turtle) init(ctx context.Context, genesisLayer *types.Layer) {
 	for _, blk := range genesisLayer.Blocks() {
 		id := blk.ID()
 		t.BlockOpinionsByLayer[genesisLayer.Index()][id] = Opinion{}
+		t.HaveOpinions[id] = struct{}{}
 		t.GoodBlocksIndex[id] = false // false means good block, not flushed
 	}
 	t.Last = genesisLayer.Index()
@@ -211,6 +213,11 @@ func (t *turtle) evict(ctx context.Context) {
 		for blk := range t.BlockOpinionsByLayer[layerToEvict] {
 			delete(t.GoodBlocksIndex, blk)
 		}
+		for _, bids := range t.BlockOpinionsByLayer[layerToEvict] {
+			for bid := range bids {
+				delete(t.HaveOpinions, bid)
+			}
+		}
 		delete(t.BlockOpinionsByLayer, layerToEvict)
 	}
 	t.LastEvicted = windowStart.Sub(1)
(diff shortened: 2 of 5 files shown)


tortoise/verifying_tortoise.go
```diff
@@ -249,10 +252,10 @@ func (t *turtle) getLocalBlockOpinion(ctx context.Context, layerID types.LayerID
...
 	}
 
-	t.logger.WithContext(ctx).With().Debug("got layer opinion vector",
-		layerID,
-		log.FieldNamed("query_block", blockid),
-		log.String("input", blockIDsToString(input)))
+	// t.logger.WithContext(ctx).With().Debug("got layer opinion vector",
+	// 	layerID,
+	// 	log.FieldNamed("query_block", blockid),
+	// 	log.String("input", blockIDsToString(input)))
```
dshulyak, line 262: expensive logs should be refactored so that data is serialized only if log level is enabled, in case of this log blockIDsToString is expensive

dshulyak:
Screenshot from 2021-10-28 14-17-43

dshulyak:
You can see bottlenecks on the flame graph above. Most of them are about maps access, adding new elements and so on. Also waiting for this change https://github.com/spacemeshos/go-spacemesh/pull/2734 to land, if it will work as i expect this time will be reduced even further, down to ~10s.

dshulyak:
bors try

tortoise/verifying_tortoise.go
```diff
@@ -273,36 +280,46 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
...
-		exceptionBlock, err := t.bdp.GetBlock(exceptionBlockID)
-		if err != nil {
-			logger.With().Error("inconsistent state: can't find block from diff list",
-				log.FieldNamed("exception_block_id", exceptionBlockID))
-			return false
+		lid, exist := t.BlockLayer[exceptionBlockID]
+		if !exist {
+			// NOTE(dshulyak) if exception is out of sliding window it will not be found in t.BlockLayer,
+			// in such case we look it up in db.
+			// i am clarifying with a research if we can use same rule for exceptions as for base blocks
```
dshulyak, line 289: if we do, we can return false right away

nkryuchkov (approved):

dshulyak:
bors try

tortoise/verifying_tortoise.go
```diff
@@ -615,26 +634,19 @@ func (t *turtle) processBlock(ctx context.Context, block *types.Block) error {
 	}
 	for blk, vote := range baseBlockOpinion {
 		// ignore opinions of very old blocks
```
countvonzero: should be "ignore opinions on very old blocks" instead of "of"

tortoise/verifying_tortoise.go
```diff
@@ -234,25 +237,29 @@ func blockIDsToString(input []types.BlockID) string {
...
 	}
-
-	input, err := t.layerOpinionVector(ctx, layerID)
-	// an error here signifies a real database failure
-	if err != nil {
-		return abstain, err
+	input, exist := inputs[layerID]
+	if !exist {
+		var err error
+		input, err = t.layerOpinionVector(ctx, layerID)
```
countvonzero, line 247: should cache this across all 3 calls to `checkBlockAndGetLocalOpinion()`. this info doesn't change until the a new layer is verified.

countvonzero (approved):

tortoise/verifying_tortoise.go
countvonzero: as a matter of fact, this opinions by layers in the sliding windows only change whenever sliding window advances. i wonder if its possible to just calculated once.

tortoise/state.go
```diff
@@ -37,6 +37,8 @@ type state struct {
 	GoodBlocksIndex map[types.BlockID]bool
 	// use 2D array to be able to iterate from latest elements easily
 	BlockOpinionsByLayer map[types.LayerID]map[types.BlockID]Opinion
+	// BlockLayer stores reverse mapping from BlockOpinionsByLayer
+	BlockLayer map[types.BlockID]types.LayerID
```
countvonzero, line 41: i wonder if there is a clever way to encode layerID inside blockID.

dshulyak: what would be the benefit?

countvonzero: that way we can truly have a 2D array of block IDs ordered by layer for the opinions

dshulyak: not sure that i understand, without hash-based index for layer id it will be slower to filter all blocks that belong to a certain layer. are you suggesting some custom map implementation that is using only part of the key? 

i think there is some confusion around this data structure, in practice this is already 2d matrix but indexed by layer for efficient eviction and base block selection. i really don't think that we need to do anything about it. it would be great if we can reduce number of opinions (e.g. block unification)

dshulyak: we can have a separate index such as 
```go
var BlockByLayer map[types.LayerID][]types.BlockID
```
But performance wise there will be no difference

countvonzero: code simplicity is my main motivation.

i found the code in this file overly complicated, especially by the need to operate by layers. ultimately we want to find out opinions on **all** blocks (in the sliding window), as long as the evction logic is correct, we really dont need to care about layers.

second point is that we want to ecode the exception list with just one list of blocks that differ from the base block in terms of opinion, and abstain diff list should be layer numbers. with that encoding, i would imagine the code is more intuitive if we just use pure 2D matrix of blocks without layer to be a intermediate index.

lastly, with unified block, it seems a waste of space to use layer as an intermediate index to blocks as the layer to block ratio is almost always 1:1

dshulyak: > ultimately we want to find out opinions on all blocks (in the sliding window), as long as the evction logic is correct, we really dont need to care about layers.

it is true for opinions, but as i mentioned primary use cases for layer index is efficient eviction and base block selection. maybe for the base block, we can read list of blocks from db, but not sure what we can do for eviction

tortoise/verifying_tortoise.go
dshulyak: i will cache it for all checkBlockAndGetLocalOpinion's for now

dshulyak:
bors merge

tortoise/state.go
countvonzero: but if we get to unified block, and that layer id is encoded as part of block id, or we keep some metadata abt the indexes of a certain layer, then we can do eviction pretty easily too. to me, eviction is a relatively simple part of this file. the opinions, base block selection and good blocks determination are the key parts to keep simple.

dshulyak: i don't disagree with you. i would definitely support the simplification. but in this implementation i don't see an efficient way to evict things without this index, the alternative is to loop through all blocks (that would be ~3000 with 100 window size) and evict each one separately.

countvonzero: if we know that layer k's has 120 blocks, then we just evict the first 120 elements in both axes of the matrix. besides, with unified blocks, if you want to evict 1 layer, most likely just have to iterate one block.

i am not talking abt investing time to change the code now. but as part of the unified block project, i think its worth it

dshulyak: we also need to find a base block opinion efficiently, at least it is gets copied in processBlock subroutine

PR #2899, opened 2021-10-29
[Merged by Bors] - tortoise: evict persisted state without reading it first

dshulyak:
## Motivation

It was an inefficiency in my initial implementation, we already have all required data to evict from the database without touching the disk at all.

related: https://github.com/spacemeshos/go-spacemesh/issues/2852

## Changes
-  unify eviction logic in tortoise/state.go
- persist in multiple batches, to avoid high memory consumption in case of rerun 
- evict from disk based on in-memory data

## Test Plan
existing tests and manual tests on the data from network

## TODO

## DevOps Notes


tortoise/state.go
```diff
@@ -90,12 +82,24 @@ func (s *state) Persist() error {
 				b.Reset()
 			}
 		}
+		if err := batch.Write(); err != nil {
+			return fmt.Errorf("write batch: %w", err)
+		}
+		batch.Reset()
 	}
 
+	if err := batch.Put([]byte(namespaceLast), s.Last.Bytes()); err != nil {
```
countvonzero, line 91: what is the difference doing it here as opposed to earlier?

tortoise/state.go
```diff
@@ -161,37 +165,44 @@ func (s *state) Recover() error {
...
+			}
 		}
-		b.Reset()
-		if err := batch.Delete(it.Key()); err != nil {
-			return fmt.Errorf("delete from batch: %w", err)
+		delete(s.BlockOpinionsByLayer, layerToEvict)
+		if err := batch.Write(); err != nil {
+			return fmt.Errorf("write batch for layer %s to db: %w", layerToEvict, err)
 		}
+		batch.Reset()
```
countvonzero, line 203: what is your guideline for calling `batch.Reset()`? data size?

countvonzero (approved):

tortoise/state.go
dshulyak: when everything was written in a single batch order didn't matter. but if write fails after writing layers tortoise will think that it has more state than it actually has. on the other hand, if votes are updated but layer is not updated it should be ok

tortoise/state.go
dshulyak: pretty much. it should be more efficient to write everything in a single batch, but writing the whole state after rerun was problematic

dshulyak:
bors merge

PR #2900, opened 2021-10-29
[Merged by Bors] - tortoise: move rerun to background goroutine

dshulyak:
## Motivation

In order not to block the network progress rerun is moved to the background goroutine. The majority of the layers are handled in background on timer, and a small number of concurrent layers are applied when HandleIncomingLayer is called by the mesh.

after: https://github.com/spacemeshos/go-spacemesh/pull/2899
closes: https://github.com/spacemeshos/go-spacemesh/pull/2899

## Changes
- move tortoise rerun to a background thread

## Test Plan
existing and manual tests

## TODO

## DevOps Notes


cmd/node/node.go
```diff
@@ -920,6 +922,10 @@ func (app *App) stopServices() {
 		app.ptimesync.Stop()
 		app.log.Debug("peer timesync stopped")
 	}
+	if app.tortoise != nil {
+		app.log.Info("stopping tortoise. if tortoise is in rerun it may take a while")
+		app.tortoise.Stop()
```
countvonzero, line 927: wow.... do you mean we never try to clean up tortoise.... yikes.

tortoise/rerun.go
```diff
@@ -0,0 +1,147 @@
...
+			updated.log = current.log
+			updated.bdp = current.bdp
+			t.trtl = updated
+		}
+		t.update.CAS(value, nil) // Store will miss a concurrent rerun
+	}
+	return reverted, observed
+}
+
+func (t *ThreadSafeVerifyingTortoise) waitRerun(ctx context.Context, period time.Duration) {
```
countvonzero: call this rerunLoop?

tortoise/verifying_algorithm.go
```diff
@@ -39,14 +33,32 @@ type Config struct {
...
+	logger log.Log
+
+	eg     errgroup.Group
+	cancel context.CancelFunc
+
+	mu sync.RWMutex
+	// persistMu is needed to allow concurrent BaseBlock and Persist call
+	// but to prevent multiple concurrent Persist calls
+	// relevant after rerun on nodes with slow disk
+	persistMu sync.Mutex
```
countvonzero, line 48: this lock is only taken out on `Persist()`. doesn't seem to do what the comment says.

tortoise/verifying_algorithm.go
```diff
@@ -39,14 +33,32 @@ type Config struct {
 	RerunInterval   time.Duration // how often to rerun from genesis
 }
 
+// ThreadSafeVerifyingTortoise is a thread safe verifying tortoise wrapper, it just locks all actions.
+type ThreadSafeVerifyingTortoise struct {
+	// update for rerun
+	update atomic.UnsafePointer
```
countvonzero: we can also protect this struct with `mu` right? it seems a simple write lock at the end of the `rerun()`
i don't see why this field should be different from others. code consistency out-weighs performance gain here.

tortoise/rerun.go
```diff
@@ -0,0 +1,147 @@
...
+			updated   = completed.Consensus
+			err       error
+		)
+		if current.Last.After(updated.Last) {
+			start := time.Now()
+			logger.Info("tortoise received more layers while rerun was in progress. running a catchup",
+				log.FieldNamed("last", current.Last),
+				log.FieldNamed("rerun-last", updated.Last))
+
+			err = catchupToCurrent(ctx, current, updated)
```
countvonzero, line 38: is this necessary? 
if rerun decide we want to revert, is catching up to current last layer change that decision? 
if rerun decide no revert, then we wait till next rerun to cover this range.
i think the logic is simpler just exit here.

tortoise/verifying_algorithm.go
dshulyak: how so? it allows to use mu.RLock both in Persist and BaseBlock and use persistMu in Persist

tortoise/rerun.go
dshulyak: it is not about revert. i can't exit here. after rerun completes we need to have votes for the same number of layers as original tortoise, for example:
1. rerun started when last was 20
2. original received 21
3. rerun completed with 20
4. original replaced with rerun
5. original(rerun) receives 22

cmd/node/node.go
dshulyak: but i introduced background worker only in this pr

countvonzero: right. tortoise didn't have anything to clean up.

tortoise/rerun.go
countvonzero: as time goes on, rerun is going to take longer and longer. this means that lag btwn original and rerun will be larger. overtime the original rerun problem will show up again (holding the lock for too long).

when catch-up fails, should we revert to the failed layer?

one thing i didn't understand about the original design was: why do we have to swap out the current tortoise when there is no revert? if we find nothing is wrong with the current tortoise, why not keep using it?

tortoise/verifying_algorithm.go
countvonzero: oh. i guess you just wanted to have one `Persist()` at a time. 

intuitively the locking scheme is complicated by the fact that `Persist()` is called by the mesh. it seems more natural to me to call `state.persist()` internally inside tortoise. that way tortoise have more control of persisting atomic state. as the code stands now, we could potentially process half of the blocks in a layer N and abort, leaving half of the layer N votes in memory, and then they get persisted as a result of layer N+1.

tortoise/rerun.go
dshulyak: > as time goes on, rerun is going to take longer and longer. this means that lag btwn original and rerun will be larger. overtime the original rerun problem will show up again (holding the lock for too long).

this is true, but not a problem. by my estimate rerun after 1 year will take around 30-40 minutes, assuming that we will have layer per minute that's only 40 layers. it will take less than second to do a catchup for 40 layers. even if it will be 60 layers after a year it is still less then a second. 600 layers after 10 years - will take around 5 seconds. this is assuming that we won't optimize tortoise any further. but i am pretty sure that once unified blocks lands rerun will be blazingly fast.

> when catch-up fails, should we revert to the failed layer?

i think we should continue using original state or terminate the application. i think it can only fail due to the bugs or system issues.

> one thing i didn't understand about the original design was: why do we have to swap out the current tortoise when there is no revert? if we find nothing is wrong with the current tortoise, why not keep using it?

i thought that swapping with the fresh state may hide some bugs that we have. but yes it shouldn't matter if we are confident, i'd rather leave it as it is for now though.

tortoise/verifying_algorithm.go
dshulyak: i would rather remove state completely eventually. 

as for the issue you pointed out i don't see how moving persist to tortoise will change anything. if we can abort with half-computed state it will be anyway persisted by the next call.

dshulyak: > oh. i guess you just wanted to have one Persist() at a time.

yes, but it is important to use only mu.RLock so that BaseBlock is never blocked by Persist

countvonzero: > as for the issue you pointed out i don't see how moving persist to tortoise will change anything. if we can abort with half-computed state it will be anyway persisted by the next call.

it won't make a difference now. but in the future, if persist() is called inside of tortoise only, we know exactly the place we aborted and revert the update in memory.

how do you envision removing state completely in the future? how would we store the opinions?

countvonzero (approved):

tortoise/rerun.go
countvonzero: > i thought that swapping with the fresh state may hide some bugs that we have. but yes it shouldn't matter if we are confident, i'd rather leave it as it is for now though.

it definitely hide potential bugs. and we will grow to rely on the rerun too much, which then may cause unnecessary turmoil in the network caused by regular reverts. IMO it's better to leave the state as is. but i agree it can be done later.

tortoise/verifying_algorithm.go
dshulyak: i meant removing the persistent state. we can always recover votes from blocks when application starts.

we may do it even now, currently, we only need votes from the sliding window, and as it was discussed in slack we may even change that window so that we only need votes that start from the verified layer.

we don't even need to do a full rerun when the application starts, just a rerun starting from verified - N layer. i want to double-check if i am not missing anything, but it will be great if the persistent state can be removed.

tortoise/rerun.go
dshulyak: > as time goes on, rerun is going to take longer and longer. this means that lag btwn original and rerun will be larger. overtime the original rerun problem will show up again (holding the lock for too long).

also once we have mesh state sync we probably won't need to do full reruns. if for example we know that aggregated hash is not equal starting from layer 100, we don't need to rerun anything for layers below that, right?

tortoise/verifying_algorithm.go
countvonzero: i agree with everything with a slight modification.

> we don't even need to do a full rerun when the application starts, just a rerun starting from verified - N layer. i want to double-check if i am not missing anything, but it will be great if the persistent state can be removed.

like i said in the slack also, `verified` is a moving target. if we find that our mesh hash differ wither peers further back than `verified`, we need to shift the sliding window back where `verified` is the point of convergence in mesh hash.

dshulyak: i think i understand it differently. if we rerun from scratch it means that our sliding window starts from genesis. if hash differs at some layer we can start sliding window from that layer and rerun from that layer. i think the concept of sliding window was introduced because by design we had to maintain some state in memory for self-healing. if we don't need to maintain it in memory then there is no such thing as a sliding window.

there is another use case for keeping votes in memory, when the verified layer is stuck we want to keep all the votes so that it can unstuck eventually. this is what i was arguing about in the slack.

countvonzero: currently the only way to trigger self-healing is `verified layer is stuck`. so i think these two scenarios are similar.

> if hash differs at some layer we can start sliding window from that layer and rerun from that layer. 

that's effectively the same as what i meant. say our hash differ from peers starting from layer N, current verified layer is N+1000. we need to move verified layer back to N and rerun from there.

tortoise/rerun.go
countvonzero: > also once we have mesh state sync we probably won't need to do full reruns. if for example we know that aggregated hash is not equal starting from layer 100, we don't need to rerun anything for layers below that, right?

yes. agreed.

dshulyak:
bors merge

PR #2901, opened 2021-10-31
Add layer ID to tortoise opinion map (instead of #2734)

### LINKED ISSUE #2424: Add exception votes against blocks the base block votes for

### LINKED ISSUE #2719: Tortoise vote encoding data structure

nkryuchkov:
## Motivation
Based on #2734
Supersedes #2734
Closes #2719 
See also #2679

*Most of the code was copied from #2734 with some fixes. It was hard to fix a couple of tests in #2734 due to some changes that may be excluded from the PR (e.g. https://github.com/spacemeshos/go-spacemesh/pull/2734/files#diff-de1c771222d45ea483c2a6212af7b40473ea6c8c814ecec398deae0ffaddb0d7R1156). Also, uncommenting of https://github.com/spacemeshos/go-spacemesh/pull/2734/files#diff-de1c771222d45ea483c2a6212af7b40473ea6c8c814ecec398deae0ffaddb0d7R491 wasn't copied because it broke `TestLayerPatterns/heal_after_bad_layers`, this likely needs to be done though (#2424).*

## Changes

- Add `layerID` to the inner tortoise block opinion map

This allows efficient eviction of opinions of blocks in older layers,
and it allows us to efficiently add votes against blocks the base block
votes for that are missing from the input vector for a layer

## Test Plan

- Unit tests
- System tests

## DevOps Notes


nkryuchkov:
bors try

tortoise/state.go
```diff
@@ -134,25 +137,31 @@ func (s *state) Recover() error {
 	it = s.db.Find([]byte(namespaceOpinions))
 	defer it.Release()
 	for it.Next() {
-		layer := decodeLayerKey(it.Key())
-		offset := 1 + types.LayerIDSize
-		block1 := decodeBlock(it.Key()[offset : offset+types.BlockIDSize])
+		// first byte is namespace
+		layer1 := decodeLayerKey(it.Key()[1:])
+		offset1 := 1 + types.LayerIDSize
+		offset2 := offset1 + types.BlockIDSize + types.LayerIDSize
```
countvonzero: i think it's more readable to do
layer1Offset := len(namespaceOpinions)
block1Offset := layer1Offset + types.LayerIDSize
layer2Offset := block1OffSet + types.BlockIDSize
block2Offset := layer2Offset + types.LayerIDSize

tortoise/state.go
```diff
@@ -134,25 +137,31 @@ func (s *state) Recover() error {
...
+		block1 := decodeBlock(it.Key()[offset1 : offset1+types.BlockIDSize])
```
countvonzero: use [block1Offset : layer2Offset]

tortoise/state.go
```diff
@@ -134,25 +137,31 @@ func (s *state) Recover() error {
...
-			s.BlockOpinionsByLayer[layer] = map[types.BlockID]Opinion{}
+		if _, exist := s.BlockOpinionsByLayer[layer1]; !exist {
+			s.BlockOpinionsByLayer[layer1] = map[types.BlockID]Opinion{}
+		}
+		if _, exist := s.BlockOpinionsByLayer[layer1][block1]; !exist {
+			s.BlockOpinionsByLayer[layer1][block1] = Opinion{}
 		}
-		if _, exist := s.BlockOpinionsByLayer[layer][block1]; !exist {
-			s.BlockOpinionsByLayer[layer][block1] = Opinion{}
+		layer2 := decodeLayerKey(it.Key()[offset1+types.BlockIDSize : offset2])
```
countvonzero: use [layer2Offset : block2Offset]

tortoise/verifying_tortoise.go
```diff
@@ -498,7 +506,7 @@ func (t *turtle) calculateExceptions(
 		// support a block and we disagree, we need to add a vote against here.
 		// TODO: this is not currently possible since base block opinions aren't indexed by layer. See
 		//   https://github.com/spacemeshos/go-spacemesh/issues/2424
-		//for b, v := range baseBlockOpinion.BlockOpinions {
+		//for b, v := range baseBlockOpinion[layerID] {
```
countvonzero, line 484: are you planning to add against diff in a follow-up PR?

tortoise/verifying_tortoise.go
```diff
@@ -613,34 +621,81 @@ func (t *turtle) processBlock(ctx context.Context, block *types.Block) error {
 		len(block.NeutralDiff) +
 		len(block.NeutralDiff) +
 		len(baseBlockOpinion)
-	opinion := make(map[types.BlockID]vec, lth)
+	opinion := make(Opinion, lth)
 
 	for _, bid := range block.ForDiff {
-		opinion[bid] = support.Multiply(voteWeight)
+		blk, err := t.bdp.GetBlock(bid)
```
countvonzero: should be able to look up layerID from `state.BlockLayer` first to avoid accessing disk

tortoise/verifying_tortoise.go
```diff
@@ -613,34 +621,81 @@ func (t *turtle) processBlock(ctx context.Context, block *types.Block) error {
...
+		}
+
+		if _, ok := opinion[blk.LayerIndex]; !ok {
+			opinion[blk.LayerIndex] = make(map[types.BlockID]vec, t.AvgLayerSize)
+		}
+
+		opinion[blk.LayerIndex][bid] = support.Multiply(voteWeight)
 	}
 	for _, bid := range block.AgainstDiff {
+		blk, err := t.bdp.GetBlock(bid)
```
countvonzero: ditto

tortoise/verifying_tortoise.go
```diff
@@ -613,34 +621,81 @@ func (t *turtle) processBlock(ctx context.Context, block *types.Block) error {
...
-		if _, alreadyVoted := opinion[bid]; alreadyVoted {
+		if _, alreadyVoted := opinion[blk.LayerIndex][bid]; alreadyVoted {
 			return fmt.Errorf("%s %v", errstrConflictingVotes, block.ID())
 		}
-		opinion[bid] = against.Multiply(voteWeight)
+		opinion[blk.LayerIndex][bid] = against.Multiply(voteWeight)
 	}
 	for _, bid := range block.NeutralDiff {
-		if _, alreadyVoted := opinion[bid]; alreadyVoted {
+		blk, err := t.bdp.GetBlock(bid)
```
countvonzero: ditto

dshulyak (changes requested):
wherever bdp.GetBlock is called in a loop, and there are many outer loops as well, it should be replaced with a call BlockLayer to get a layer from memory

tortoise/verifying_tortoise.go
nkryuchkov: Either in a follow-up one or in this one if I fix the broken by this addition tests until merging and if the amount of changes is tiny

nkryuchkov:
bors try

tortoise/verifying_tortoise.go
```diff
@@ -613,34 +621,89 @@ func (t *turtle) processBlock(ctx context.Context, block *types.Block) error {
...
+		if _, alreadyVoted := opinion[blockLayer][blockID]; alreadyVoted {
 			return fmt.Errorf("%s %v", errstrConflictingVotes, block.ID())
 		}
-		opinion[bid] = against.Multiply(voteWeight)
+		opinion[blockLayer][blockID] = against.Multiply(voteWeight)
 	}
-	for _, bid := range block.NeutralDiff {
-		if _, alreadyVoted := opinion[bid]; alreadyVoted {
+	for _, blockID := range block.NeutralDiff {
+		blockLayer, ok := t.BlockLayer[blockID]
```
dshulyak: please deduplicate this code

nkryuchkov:
bors try-

dshulyak:
i merged this change with https://github.com/spacemeshos/go-spacemesh/pull/2900 and launched a new miner in the cloud to see how it affects rerun, memory and some other things. will update later

dshulyak:
my observation so far is that this doesn't reduce memory by a lot. definitely not 50%, maybe slightly. my node that runs locally uses around 1gb of memory as before. 

rerun became two times slower, and i expected it to be faster. i didn't look into profiles, but this is definitely a problem.

dshulyak:
i suggest postponing this change until the unified content block is introduced, as it very likely will be obsolete. we already considered rewriting the core data structures of a tortoise and it is will happen after a unified content block, if not before. 

i don't like that it makes full rerun ~2x slower and introduces more complexity into the data structure. 
also, i don't see that it makes a substantial difference memory-wise. also it enables https://github.com/spacemeshos/go-spacemesh/issues/2424 , which we will have to support, but we can try inefficient variant first and use until better variant will be found.

nkryuchkov:
@dshulyak I agree. It doesn't enable #2424 yet, I will remove mentioning #2424 and will put this PR on hold then

dshulyak:
tortoise is fully rewritten in https://github.com/spacemeshos/go-spacemesh/pull/2994 , this data structure is not relevant

PR #2908, opened 2021-11-01
tortoise: add more cache for getting local input vector

countvonzero:
## Motivation

add more cache for local input vectors for calculating exceptions, and btwn scoring and verifying layers

## Changes

- for miner: add cache for local input vectors across blocks in the same layer
- for tortoise: share cache btwn good blocks determination and verifying blocks

## Test Plan

unit tests

## TODO

## DevOps Notes

### DIFF
--- a/tortoise/verifying_tortoise.go
+++ b/tortoise/verifying_tortoise.go
@@ -824,13 +824,6 @@ func (t *turtle) verifyLayers(ctx context.Context) error {
 candidateLayerLoop:
 	for candidateLayerID := t.Verified.Add(1); candidateLayerID.Before(t.Last); candidateLayerID = candidateLayerID.Add(1) {
 		logger := logger.WithFields(log.FieldNamed("candidate_layer", candidateLayerID))
-
-		// it's possible that self healing already verified a layer
-		if !t.Verified.Before(candidateLayerID) {
-			logger.Info("self healing already verified this layer")
-			continue
-		}
-
 		logger.Info("attempting to verify candidate layer")
 
 		// note: if the following checks fail, we just return rather than trying to verify later layers.
@@ -941,10 +934,6 @@ candidateLayerLoop:
 			// between this unverified candidate layer and the latest layer is greater than this distance, then we trigger
 			// self healing. But there's no point in trying to heal a layer that's not at least Hdist layers old since
 			// we only consider the local opinion for recent layers.
-			if candidateLayerID.After(t.Last) {
-				logger.With().Panic("candidate layer is higher than last layer received",
-					log.FieldNamed("last_layer", t.Last))
-			}
 			logger.With().Debug("considering attempting to heal layer",
 				log.FieldNamed("layer_cutoff", t.layerCutoff()),
 				log.Uint32("zdist", t.Zdist),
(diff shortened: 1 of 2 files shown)


tortoise/verifying_tortoise.go
```diff
@@ -237,19 +237,13 @@ func blockIDsToString(input []types.BlockID) string {
 // returns the local opinion on the validity of a block in a layer (support, against, or abstain)
 // TODO: cache but somehow check for changes (e.g., late-finishing Hare), maybe check hash?
 //    see https://github.com/spacemeshos/go-spacemesh/issues/2672
-func (t *turtle) getLocalBlockOpinion(ctx context.Context, layerID types.LayerID, blockid types.BlockID, inputs map[types.LayerID][]types.BlockID) (vec, error) {
+func (t *turtle) getLocalBlockOpinion(ctx context.Context, layerID types.LayerID, blockid types.BlockID, localOpinionsCache map[types.LayerID][]types.BlockID) (vec, error) {
```
dshulyak: i think that this variable should refer to layer input vectors somehow, and not opinions.
for example this method returns support/abstain/against as an opinion

tortoise/verifying_tortoise.go
```diff
@@ -824,13 +822,6 @@ func (t *turtle) verifyLayers(ctx context.Context) error {
 candidateLayerLoop:
 	for candidateLayerID := t.Verified.Add(1); candidateLayerID.Before(t.Last); candidateLayerID = candidateLayerID.Add(1) {
 		logger := logger.WithFields(log.FieldNamed("candidate_layer", candidateLayerID))
-
```
dshulyak, line 827: why removed?

dshulyak (commented):
i am not sure if it is really needed. i added this cache only where it was a bottleneck. in other places this is not very relevant, and reading from db would keep things simpler

tortoise/verifying_tortoise.go
countvonzero: changed

tortoise/verifying_tortoise.go
countvonzero: i removed it because i cannot figure out how this could happen. self-healing is running in the same thread. no other thread should be able to change t.Verified. do you read differently?

countvonzero:
> i am not sure if it is really needed. i added this cache only where it was a bottleneck. in other places this is not very relevant, and reading from db would keep things simpler

it's not performance critical no. but i think the question we should ask is "why not? why not use cache when we can avoid disk access and be consistent through out the file?"

dshulyak (approved):
> it's not performance critical no. but i think the question we should ask is "why not? why not use cache when we can avoid disk access and be consistent through out the file?"

because now every function has one additional required parameter. on its own it is a minor detail but it all adds up

countvonzero:
> > it's not performance critical no. but i think the question we should ask is "why not? why not use cache when we can avoid disk access and be consistent through out the file?"
> 
> because now every function has one additional required parameter. on its own it is a minor detail but it all adds up

ok. unified content block is going to change a lot of code here. it seems there is little to gain with this PR. i will abandon this one.

PR #2910, opened 2021-11-02
[Merged by Bors] - hare: record empty input vector when terminate with empty set

### LINKED ISSUE #2875: tortoise: layers invalidated by hare are ignored
## Description

`GetLayerInputVectorByID` should check it and return `mesh.ErrInvalidLayer` if the layer is invalid. it is then checked in verifying_tortoise.go https://github.com/spacemeshos/go-spacemesh/blob/4f6a8b2ca56dcaba0d947e2bcec43899299c455b/tortoise/verifying_tortoise.go#L1099

## Actual Behavior
invalidated layer output by hare is ignored.

## Expected Behavior
invalided layer should be processed and voted against

countvonzero:
## Motivation

Closes #2875

## Changes

- make sure that tortoise param zdist (the number of layers its willing to wait for hare before voting against all blocks in the layer) is compatible with the hare params hare_limit_iterations

- hare can only produce two outcomes:
  - when it terminates with success, the output can be empty or non-empty set of BlockIDs
  - when it terminates with failure, either it reaches max iterations or system shutdown, the output is an empty set

  we call `HandleValidatedLayer` in the success case but not in the failure case. we should do so to advance the verified layer since empty set is a legitimate output that will later cause tortoise to vote against all blocks in that layer.

## Test Plan

unit tests

## DevOps Notes


cmd/node/node.go
```diff
@@ -306,6 +297,20 @@ func (app *App) introduction() {
 
 // Initialize sets up an exit signal, logging and checks the clock, returns error if clock is not in sync.
 func (app *App) Initialize() (err error) {
+	// tortoise wait zdist layers for hare to timeout for a layer. once hare timeout, tortoise will
+	// vote against all blocks in that layer. so it's important to make sure zdist takes longer than
+	// hare's max time duration to run consensus for a layer
+	maxHareLayerDurationSec := app.Config.HARE.WakeupDelta + app.Config.HARE.LimitIterations*4*app.Config.HARE.RoundDuration
```
dshulyak: maybe 4 should be defined as a constant in hare module?

hare/hare.go
```diff
@@ -193,18 +193,23 @@ var ErrTooLate = errors.New("consensus process finished too late")
...
-		i++
+	layerID := output.ID()
+	var blocks []types.BlockID
+	if output.Completed() {
+		h.WithContext(ctx).With().Info("hare terminated with success", layerID)
+		set := output.Set()
+		blocks = make([]types.BlockID, len(set.values))
+		i := 0
+		for v := range set.values {
+			blocks[i] = v
```
dshulyak: will be cleaner with capacity instead of len and append

dshulyak (approved):

cmd/node/node.go
dshulyak: don't we have a preround before iterations?

countvonzero:
bors try

cmd/node/node.go
countvonzero: added. thanks

hare/hare.go
countvonzero: done

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

Issue #2893, opened 2021-10-27
tortoise: blocks sequence generator for model based tests


dshulyak:
Tortoise main function is to validate sequence of blocks. By generating sequences randomly, but deterministically we can explore when tortoise fails to make progress and find tricky bugs before they are observed in the running network.

In the happy case each sequence follows best expectations of the protocol, e.g. layers are submitted in order, all hare inputs are present, no exceptions for older layers and so on. The first step would be to implement such generator that is easy to use. After that we can introduce failure scenarios, they will be enabled at random (but deterministically), so in case of failure test can be restarted from a single seed number.

List of failure scenarios:

- out of order layer
- empty hare set
- no good blocks in the layer (for example due to exceptions older then base block)
- missing fraction of blocks

PR #2911, opened 2021-11-02
[Merged by Bors] - tortoise: if block was unknown treat it as contextually invalid instead of failing

dshulyak:
## Motivation

If tortoise never validated a layer or block was unknown at the time when the tortoise validated a layer contextual validity will not be set at all in the database. This was preventing rerun from recovering correct aggregated hash

## Changes
- fail contextual validity check only if error is different from not found
- included testing code related to https://github.com/spacemeshos/go-spacemesh/issues/2893

## Test Plan
TestRerunMissingLayer and tested on devnet

## TODO

## DevOps Notes


nkryuchkov (approved):

tortoise/rerun.go
```diff
@@ -116,7 +118,7 @@ type validityTracer struct {
 func (vt *validityTracer) SaveContextualValidity(bid types.BlockID, lid types.LayerID, validityNew bool) error {
 	if vt.firstUpdatedLayer == nil {
 		validityCur, err := vt.ContextualValidity(bid)
-		if err != nil {
+		if err != nil && !errors.Is(err, database.ErrNotFound) {
```
countvonzero, line 121: this is tricky....
this only make sense if `lid` is later than the verified layer. but if `lid` is older than verified layer, all block contextual validity should be saved already. so database.ErrNotFound should be programming error and should abort rerun as well.

this led me to think, should we only do rerun up to currently verified layer? why should rerun compete with regular tortoise?

countvonzero (changes requested):
should also add test for the scenario you are trying to fix in the devnet

tortoise/rerun.go
dshulyak: > this only make sense if lid is later than the verified layer. but if lid is older than verified layer, all block contextual validity should be saved already. so database.ErrNotFound should be programming error and should abort rerun as well.

blocks can be missing when tortoise verified layer initially. because of this problem the rerun was failing on miner-4

> this led me to think, should we only do rerun up to currently verified layer? why should rerun compete with regular tortoise?

we need to rerun until the last layer anyway to have compatible state, so i am not sure that i can make this change

countvonzero: > blocks can be missing when tortoise verified layer initially. because of this problem the rerun was failing on miner-4

right. but if `database.ErrNotFound` were to happen to an already verified layer, then it's a honest programming error.

> we need to rerun until the last layer anyway to have compatible state, so i am not sure that i can make this change

not asking you to change in this PR. but i am questioning the rerun design. the point of rerun is to double-check if we are still confident of what we've done so far. as i understand it, this need arose when we changed from vote-counting tortoise to verifying tortoise. i think we should take it up to Tal again for the more thorough discussion.

dshulyak:
@countvonzero i wanted to add a test initially but i really don't want to use the existing testing codebase. i wrote some utilities in https://github.com/spacemeshos/go-spacemesh/pull/2912 , but i would like to give them more though because they don't do everything that i want yet.

the case that i am covering is https://github.com/spacemeshos/go-spacemesh/pull/2912/commits/6132e1d97869f0f13e8746029928c0ca93a82b6f

i can merge them with this branch though

countvonzero:
> @countvonzero i wanted to add a test initially but i really don't want to use the existing testing codebase. i wrote some utilities in #2912 , but i would like to give them more though because they don't do everything that i want yet.
> 
> the case that i am covering is [6132e1d](https://github.com/spacemeshos/go-spacemesh/commit/6132e1d97869f0f13e8746029928c0ca93a82b6f)
> 
> i can merge them with this branch though

that's great. merging that commit into this branch will be sufficient. test is the best documentation of code behavior IMO.

tortoise/rerun.go
dshulyak: > right. but if database.ErrNotFound were to happen to an already verified layer, then it's a honest programming error.

i don't think so, in case of self-healing after long lasting partition layer will be verified but every side will be missing some blocks

tortoise/verifying_tortoise.go
```diff
@@ -518,35 +510,12 @@ func (t *turtle) voteWeight(ctx context.Context, votingBlock *types.Block) (uint
 		return 0, fmt.Errorf("get ATX header: %w", err)
 	}
 
-	atxTimestamp, err := t.atxdb.GetAtxTimestamp(votingBlock.ATXID)
```
countvonzero, line 521: do you mean to remove this for now since it's not used anyway?

tortoise/rerun.go
countvonzero: > > right. but if database.ErrNotFound were to happen to an already verified layer, then it's a honest programming error.
> 
> i don't think so, in case of self-healing after long lasting partition layer will be verified but every side will be missing some blocks

i agree. i meant programming error in the sense where we previously find this block contextually valid but didn't save it correctly. in any case, i think i'm splitting hair here. the point of rerun is to fix what was missing previously. so i am agreeing with you now.

countvonzero (approved):

tortoise/verifying_tortoise.go
dshulyak: there was a response from Tal and Lane that it is not needed. i will link you in slack

dshulyak: there was a response from Tal and Lane that it is not needed. i will link you in slack

dshulyak:
bors merge

Issue #2876, opened 2021-10-15
devnet: hare terminated with block that is not present in mesh


countvonzero:
## Description

(labeling with devnet as @dshulyak said this happens a lot in devnet207. in one incident he traced the origin of the block to an external node)

chain of events:
- node received a block `eee46bee2c` via gossip but failed syntactic validation due to error `failed to fetch ref block ebe041945b e: fetch failed after max retries for request`. the ref block `ebe041945b` was produced by another miner 1 min before this failure.
- 500ms later the above failure the node receive block `ebe041945b` via gossip
- later when hare terminates this layer with output that includes `eee46bee2c`, it then failed with `hare terminated with block that is not present in mesh`

it is curious why, in a managed network supposed with good synchrony, that it takes 1 min to propagate. here is a screenshot of the timestamp when the network receive block `ebe041945b`. assuming there is no network delay (very reasonable for managed nodes in the same cluster), it seems the gossip pipeline is clogged somehow. it takes ~400ms for the next miner to receive the block, and another ~400ms to reach the next set of miners. i don't know how much of a outlier this is. in the mining system test, 3 out of 50 nodes suffered this issue.

![Screenshot from 2021-10-15 12-51-22](https://user-images.githubusercontent.com/30611210/137434287-adddee89-5dfe-4363-a788-d45cafc19b4c.png)
![Screenshot from 2021-10-15 12-51-51](https://user-images.githubusercontent.com/30611210/137434292-e9ff42dd-d26d-4283-9bfb-32507c83f4d4.png)
![Screenshot from 2021-10-15 12-53-21](https://user-images.githubusercontent.com/30611210/137434299-d333d861-ea3a-4f76-b681-968f3daa1016.png)

## Steps to reproduce
1. set MaxRetriesForRequest in fetch.go to 20
2. run `make dockertest-mining namespace=$NAMESPACE dump=True`

countvonzero:
discussion btwn @countvonzero @dshulyak @moshababo @noamnelke 

## problem
some miner produced two blocks in the same layer A (ref block) and B. a node received block B via gossip, but failed to fetch its ref block (block A) via peers (after 100 retries from random peers). this can happen when the peer who gossiped block B to us simply dropped out. the node then considers block B syntactically invalid.

later hare terminated with block B in the output. so when tortoise try to verify the hare output, it failed since it didn't save block B in the mesh. this node's verified layer is then stuck.

## discussion
[...]

PR #2913, opened 2021-11-03
[Merged by Bors] - mesh: try fetch blocks from consensus output

countvonzero:
## Motivation

help alleviate impact of #2876

## Changes

hare may not have all blocks in the consensus output due to gossip delay, try to fetch all blocks after we receive hare output.

## Test Plan

unit tests

## DevOps Notes


hare/interfaces.go
```diff
@@ -12,6 +12,10 @@ type layerPatrol interface {
 	SetHareInCharge(types.LayerID)
 }
 
+type blockFetcher interface {
+	GetBlocks(context.Context, []types.BlockID) error
+}
```
dshulyak: maybe reuse interface definition? there is really no reason to define it everywhere.

there is an interface for Fetcher in `system` module, GetBlocks can be decoupled from it and used where necessary

dshulyak (commented):
i guess this is a good change, as we may learn about new blocks from hare, so it makes sense to download them at that point as well. 

but feels bad that fetcher is now a dependency in hare. i would try to keep hare only about consensus and move everything else outside. maybe it is better if mesh will have a dependency on fetcher?

hare/interfaces.go
countvonzero: minor issues (that always make me indecisive)
- i only want one of the interface. so in case i cannot use gomock, i need to create all its empty methods. like tests in cmd/xxx
- i don't know whether we should create mock in the importing pkg or the imported pkg. i feel there is more flexibility creating an interface and mock object at the importing pkg. in this case, instead of creating mock at `system/mocks` i chose to do it in `hare/mocks` with just 1 interface method. of course like you said, there'll be code dupes.

countvonzero:
> i guess this is a good change, as we may learn about new blocks from hare, so it makes sense to download them at that point as well.
> 
> but feels bad that fetcher is now a dependency in hare. i would try to keep hare only about consensus and move everything else outside. maybe it is better if mesh will have a dependency on fetcher?

i was torn too btwn hare and mesh. so since you have a preference i'll go with yours.
my original thought is that mesh receive `HandleValidatedLayer` from syncer and hare, also had assumptions that all the data it needs is in the mesh already. since hare created this output, it's more logical to ensure the blocks are downloaded after the output is generated.

in any case, changed to fetch in mesh.

hare/interfaces.go
dshulyak: > i only want one of the interface. so in case i cannot use gomock, i need to create all its empty methods. like tests in cmd/xxx

i meant that you can decouple 

```go
type BlockFetcher interface {
  FetchBlocks
}
type Fetcher interface {
  BlockFetcher
  AtxFetcher
}
```

and use only a specific subset, like in https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/pubsub/pubsub.go#L55-L69

> i don't know whether we should create mock in the importing pkg or the imported pkg. i feel there is more flexibility creating an interface and mock object at the importing pkg. in this case, instead of creating mock at system/mocks i chose to do it in hare/mocks with just 1 interface method. of course like you said, there'll be code dupes.

i think mocks should be generated where the interface is defined. but i don't see what flexibility is missing

dshulyak (approved):

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/4091001902?check_suite_focus=true)

i canceled this manually

PR #2914, opened 2021-11-03
[Merged by Bors] - tortoise: add test case for out of order layer

dshulyak:
## Motivation

Tortoise can't handle an out-of-order layer without reloading votes from all layer past out-of-order layer.
If we want to support such use case we need to re-process blocks from all layers after the one that is out of order.

Closes #

## TODO

## DevOps Notes


tortoise/sim/generator.go
```diff
@@ -84,6 +85,7 @@ type Generator struct {
 	State State
 
 	nextLayer   types.LayerID
+	reordered   map[types.LayerID]types.LayerID
```
countvonzero, line 89: can you add comment of what key/value supposed to be? thanks

tortoise/verifying_tortoise_test.go
```diff
@@ -3074,12 +3074,39 @@ func TestRerunMissingLayer(t *testing.T) {
...
+	cfg := defaultConfig(t, s.State.MeshDB, s.State.AtxDB)
+	tortoise := NewVerifyingTortoise(ctx, cfg)
+
+	var (
+		inorder  types.LayerID
+		verified types.LayerID
+	)
+	for _, lid := range sim.GenLayers(s,
+		sim.WithSequence(1),
+		sim.WithSequence(1, sim.WithNextReorder(1)),
```
countvonzero, line 3065: can you add comments about what `WithNextReorder()` do or give a concrete example for the layers generated by these sequences? i find this hard to understand quickly.

dshulyak: in this case the layers are 8, 10, 9, 11, 12 . i will try to clarify that in the WithNextReorder, it basically defines the delay for inorder layer

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

Issue #10, opened 2017-12-27
Peer connectivity latency metrics


avive:
Add peers connection latency metrics per peer and use them in routing table ops. e.g. drop from table peers with very high latency.

dgonzalez:
I have the perfect solution for this:
https://github.com/dgonzalez/vishnu

I need to add a couple of examples to it (very rough so far) and it will make perfect sense for using here. What we are looking is some sort of circuit breaker that can inform about telemetry on nodes and open the circuit (stop connecting to that peer) in case of problems.

Will have a look to this next.

avive:
Nice. Can you please elaborate a bit more abut this here?
The solution needs to work in the context of the peer DHT-baesd discovery features - we have a basic partial implementation of DHT but without any notion of peers latency. The basic idea is to have a notion of connection latency based on recent communications with a peer and drop peers with high latency from the DHT routing table. A good reference to this feature is go-lib-p2p.

dgonzalez:
This works similar to (forget about the syntax):
```
vishnu := Vishnu{...}
vishnu.add(peer1)
vishnu.add(peer2)
vishnu.add(peer3)

vishnu.With(func (peerX, stats) {
   before = TimeNow()
   peerX.communicateWithIt(...)
   after = TimeNow()
   latency = after - before
   stats["latency"] = latency
})
```

Then the next time that you execute `With` in order to select another peer, Vishnu is going to select the one with less latency (the healthier one, criteria can be configured with a rating function). Also, if there is a problem connecting to a peer, it will stop using it for 30 seconds and bring it back into the loop after that for a test and see if it has been recovered.

Not sure how clear it is but I intend to have better examples this week.

avive:
@dgonzalez are you working on this? I believe we should start with something basic:
1. Record new latency measurement for the round-trip time of a protocol message and response parts. e.g. Store t when ping message was sent, calc latency as the diff between time where pong was received by sender and update the remote peer metrics.
2. Use exponential smoothing to decay old measurements https://en.wikipedia.org/wiki/Exponential_smoothing
3. Have a notion of peer latency based on the exponentially decayed moving average
4. Have DHT occasionally ping all known peers to update latency.
5. Use the latency in the DHT to make decisions regarding which peers to remove from the routing table.

dgonzalez:
Not so far. I am working on the Merkle.

PR #2915, opened 2021-11-03
[Merged by Bors] - tortoise: persist after rerun to avoid leaking persisted state

dshulyak:
## Motivation

Some on-disk state was leaked because eviction after rerun didn't evict layers that were fetched concurrently with rerun. 

## Changes
- persist rerun state before catchup while holding the mutex

## TODO

## DevOps Notes


dshulyak:
Will add a test tomorrow.

dshulyak:
@countvonzero @nkryuchkov can you review please? i want to create new tag with this hotfix, in case if we will be opening devnet209

tortoise/rerun.go
```diff
@@ -31,7 +31,13 @@ func (t *ThreadSafeVerifyingTortoise) updateFromRerun(ctx context.Context) (bool
 			updated   = completed.Consensus
 			err       error
 		)
-		if current.Last.After(updated.Last) {
+		// we need to persist after rerun is completed so that eviction will not miss any layers
+		// that were received concurrently.
+		err = completed.Consensus.Persist()
```
countvonzero: :s/completed.Consensus/updated/

countvonzero: why `Persist` here instead of after `catchupToCurrent` is completed?

dshulyak: because we are evicting layers based on the window, if state after rerun is not persisted we will evict some layers only from memory while doing a catchup and handling new layer

tortoise/rerun.go
```diff
@@ -31,7 +31,13 @@ func (t *ThreadSafeVerifyingTortoise) updateFromRerun(ctx context.Context) (bool
 			updated   = completed.Consensus
 			err       error
 		)
-		if current.Last.After(updated.Last) {
+		// we need to persist after rerun is completed so that eviction will not miss any layers
```
countvonzero: i feel this method may be more readable we return quickly. for example,
```
if t.update == nil {
  return false, 0
}

if err := update.Persist(); err != nil {
    t.update = nil
    return false, 0
}
```

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
> E           HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"DaemonSet in version \"v1\" cannot be handled as a DaemonSet: v1.DaemonSet.Spec: v1.DaemonSetSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.NodeSelector: ReadString: expects \" or n, but found 6, error found in #10 byte of ...|espace\": 66751}, \"to|..., bigger context ...|eriodSeconds\": 10, \"nodeSelector\": {\"namespace\": 66751}, \"tolerations\": [{\"key\": \"namespace\", \"opera|...","reason":"BadRequest","code":400}

infra failure

dshulyak:
bors merge

PR #2916, opened 2021-11-04
[Merged by Bors] - tortoise: add beacon to good block determination

### LINKED ISSUE #2812: beacon sync: verifying tortoise never counts votes from block with incorrect beacon value
## Description
see https://github.com/spacemeshos/go-spacemesh/issues/2811

verifying tortoise will ignore votes from blocks that have a incorrect tortoise beacon value

### LINKED ISSUE #2813: beacon sync: self-healing ignore votes from block with incorrect beacon for K layers
## Description

see https://github.com/spacemeshos/go-spacemesh/issues/2811

for votes in blocks with incorrect beacon value, self-healing will ignore it for K layers (with 0 weight)
and start counting the vote after K layers.

countvonzero:
## Motivation

Closes #2812
Closes #2813

## Changes

- include beacon value into good block consideration
- delay votes in blocks with the wrong beacon by K layers

## Test Plan

unit test
will try to add a more involved test once https://github.com/spacemeshos/go-spacemesh/pull/2914 is merged

## TODO

## DevOps Notes


countvonzero:
bors try

tortoise/state.go
```diff
@@ -202,7 +209,11 @@ func (s *state) Evict(ctx context.Context, windowStart types.LayerID) error {
 		}
 		batch.Reset()
 	}
+	oldEpoch := s.LastEvicted.GetEpoch()
 	s.LastEvicted = windowStart.Sub(1)
+	if s.LastEvicted.GetEpoch() > oldEpoch {
```
dshulyak: i think we need to evict only if last evicted is the last layer in the epoch, otherwise cache will be invalidated too early

tortoise/verifying_tortoise.go
```diff
@@ -9,34 +10,22 @@ import (
...
 	GetAtxHeader(types.ATXID) (*types.ActivationTxHeader, error)
 }
 
-type layerClock interface {
-	LayerToTime(types.LayerID) time.Time
-}
-
 var (
+	// we will delay counting votes in blocks with different beacon values by this many layers.
+	badBeaconVoteDelays = uint32(10)
```
dshulyak: can it be a constant? or otherwise shouldn't it be defined with config?

tortoise/verifying_tortoise_test.go
```diff
@@ -3059,3 +3065,106 @@ func BenchmarkTortoiseLayerHandling(b *testing.B) {
 		}
 	}
 }
+
+func randomBytes(t *testing.T, size int) []byte {
+	data, err := crypto.GetRandomBytes(size)
```
dshulyak: why is it crypto? not a big deal, but test code should preferably use randomness from math package

tortoise/verifying_tortoise_test.go
```diff
@@ -3059,3 +3065,106 @@ func BenchmarkTortoiseLayerHandling(b *testing.B) {
...
+	require.NoError(t, err)
+	return data
+}
+
+func randomBlock(t *testing.T, lyrID types.LayerID, beacon []byte, refBlockID *types.BlockID) *types.Block {
```
dshulyak: can you please use tb testing.TB and set tb.Helper(), to make it more reusable

dshulyak (approved):

tortoise/state.go
countvonzero: i ended up using a map to collect epoch to evict in case we evicted multiple epochs at a time.

tortoise/verifying_tortoise.go
countvonzero: at this point this value is quite arbitrary. i think we need to do experiment to finalize this value. during R&D discussion Tal had always used arbitrary layers, most often 1 epoch. so i set it in the config with `layersPerEpoch`

tortoise/verifying_tortoise_test.go
countvonzero: changed to math.

tortoise/verifying_tortoise_test.go
countvonzero: done

countvonzero:
bors try

countvonzero:
bors merge

Issue #2782, opened 2021-09-14
tortoise: good block candidate contains exception for block older than its base block


dshulyak:
## Description

Periodically invalid blocks will be created and reported later by tortoise:

```
{"L":"ERROR","T":"2021-09-14T07:06:24.537+0300","N":"d402d.trtl         .turtle","M":"good block candidate contains exception for block older than its base block","node_id":"d402d6a59d870552a7d9c7ca6af2d1fada8e189fea82a712aa5495b6019388c0","module":"trtl","tortoise_rerun":"false","sessionId":"cfda9be0-8afa-4c17-86c4-ba2e333f8c0b","older_block":"02e6cb34f9","older_layer":201,"base_block_layer":202,"name":""}
``` 

@lrettig explained previously that this is implemented according to the specification, but it still seems off to me.

1. why would an honest node create a block if it knows that it will be reported as invalid by tortoise?
2. this definitely should not be happening in the network that has very little asynchrony and runs in cloud environment

lrettig:
Re 1: this does not mean that a block is invalid or that its vote isn't
counted, just that tortoise isn't marking it "good." Not being good means
it won't be used as a base block in future.

dshulyak:
should we lower the log level in this case then? 
it is still confusing why another honest node would create such block

lrettig:
In and of itself, in a production network, this is expected and not an
error, but the log level is set to error because in the testing we're doing
right now it probably does indicate a real error. I agree with you, we
don't expect nodes to create blocks that aren't marked good if the network
is healthy. I'd rather find and address the cause of that rather than
change the log level for now.

dshulyak:
Failure in the mining test https://kibana-tests.spacemesh.io/goto/834a8a45f46214e1994363a0ecd7479e

dshulyak:
I have one idea about this issue, noting here to double check next time when this bug appears.
node is marked synced as soon as we download blocks, but not validate them by the tortoise. 
- we add a layer from hare, for example, 110
- we add a layer from sync, 109
- then we choose a base block from 110, but since blocks for 109 were missing when block for 110 was processed they won't be in votes and we will add them as exceptions.
- every other node marks this blog as not good due to such exceptions

PR #2922, opened 2021-11-08
[Merged by Bors] - tortoise: organize layers in order

### LINKED ISSUE #2909: blocks: verified layer will get stuck if hare terminates with a block that was failed by blockHandler
## Description

This was reproduced several times in recent devnets. In recent devnet the sequence of events is the next one:

- 12:20:40.230 failed to validate block blockid=2f60a4a9e7 err=failed to fetch ref block e97126425a e: protocol not supported
it failed to download one of the exceptions from peers and therefore failed current block
- 12:23:10.067 hare terminated with block that is not present in mesh blockid=2f60a4a9e7 layer=2028
hare terminates with a missing block. which can be expected in theory, but should not be possible in the network that runs in the cloud. as the result we don't push the whole layer to the tortoise. and it will miss the votes from this layer.
- 12:24:00.049 got new block and added block to database blockid=2f60a4a9e7
we get missing block from the network either through the sync or gossip, at this point we can validate 2028 properly.
but the only time we do this is by rerun.
- 12:26:30.062 handling incoming layer layer=2029
we can't count any votes from layer 2029 because we don't have blocks from layer 2028. 

There are several problems:
- fetcher for some reasons fails to download dependency, even though this dependency was added to the database at 12:20:40.081 .
- why sync didn't trigger verify layer for 2028?

## Actual Behavior

The verified layer is stuck and the only way to recover is by self-healing rerun.
Which is even when will be enabled will take time to recover the layer.

## Expected Behavior

In we failed to verify layer using hare - sync will trigger verify layer when all blocks are recovered. 

## Environment

miner-4 from latest unnamed devnet.

dshulyak:
## Motivation

The assumption that tortoise can handle out-of-order layers is not correct, and it will almost always lead to a stuck verified layer. When eventually out-of-order layer is delivered we don't update blocks that were dependant on the blocks from out-of-order layer, this includes "late" blocks.

This change introduces layer organizer that will keep layers ordered within a window, once there is no gaps in the order all of them will be submitted to the tortoise at once.

I kept window for layers large in this change (1000 layers) because i expect it to help with a problem described in https://github.com/spacemeshos/go-spacemesh/issues/2782#issuecomment-952617665.

closes: https://github.com/spacemeshos/go-spacemesh/issues/2909

## Changes
- organizer for layers
- deprecated late blocks handler so that it doesn't mess with ordered layers

## Test Plan
unit tests

## TODO

## DevOps Notes


tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,122 @@
...
+type Organizer struct {
+	logger log.Log
+
+	// last submitted layer
+	submitted types.LayerID
+	// last received layer that is not submitted
+	received types.LayerID
+	// all layers between submitted and received
+	// window is non-empty only if there are gaps after submitted
+	window []types.LayerID
```
countvonzero: can we call this buffer or something that's not `window`? as this can be easily confused with the sliding window in tortoise

tortoise/verifying_algorithm.go
```diff
@@ -115,6 +122,7 @@ func (trtl *ThreadSafeVerifyingTortoise) BaseBlock(ctx context.Context) (types.B
 
 // HandleLateBlocks processes votes and goodness for late blocks (for late block definition see white paper).
 // Returns the old verified layer and new verified layer after taking into account the blocks' votes.
+// DEPRECATED: don't use this method it will be completely removed.
 func (trtl *ThreadSafeVerifyingTortoise) HandleLateBlocks(ctx context.Context, blocks []*types.Block) (types.LayerID, types.LayerID) {
```
countvonzero: why not remove now?

dshulyak:
bors try

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,122 @@
...
+		)
+	}
+}
+
+// submitPending submits all non-nil layers in (o.submitted, o.received].
+func (o *Organizer) submitPending(ctx context.Context, all bool, f LayersIterator) {
+	for lid := o.submitted.Add(1); !lid.After(o.received); lid = lid.Add(1) {
+		pos := int(lid.Uint32()) % len(o.window)
+		if (o.window[pos] == types.LayerID{}) {
+			if !all {
```
countvonzero: i don't understand what the arg `all` mean here. please add comment

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if lid.After(o.submitted.Add(uint32(len(o.window)))) {
+		o.logger.WithContext(ctx).With().Error("tortoise organizer received a layer that does not fit in the current window."+
+			" subbmiting pending with gaps",
+			log.Int("window", len(o.window)),
+			log.Named("submitted", o.submitted),
+			log.Named("received", lid),
+		)
+		o.submitPending(ctx, true, f)
+		o.submitted = lid.Sub(1)
```
countvonzero, line 81: i don't understand the action here. please add comment

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,122 @@
...
+	// window is non-empty only if there are gaps after submitted
+	window []types.LayerID
+}
+
+// Iterate allows to iterate over all ordered layers that are currently within the window including new layer.
+func (o *Organizer) Iterate(ctx context.Context, lid types.LayerID, f LayersIterator) {
+	if !lid.After(o.submitted) {
+		return
+	}
+	if lid.After(o.submitted.Add(uint32(len(o.window)))) {
```
countvonzero: this may be overkill. in hare there are two params to make sure there are limited layers of delay by hare.
`LimitIterations` and `LimitConcurrent`. one can do a startup check to make sure the window size we use here make sense. i think even a buffer of 100 layers in the mainnet is too many.

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,122 @@
...
+			log.Int("window", len(o.window)),
+			log.Named("submitted", o.submitted),
+			log.Named("received", lid),
+		)
+		o.submitPending(ctx, true, f)
+		o.submitted = lid.Sub(1)
+	}
+
+	// at this stage we know that the layer is within the window
+	pos := int(lid.Uint32()) % len(o.window)
```
countvonzero: and the scheme using `% len(o.window)` to decide position is clever but not very intuitive. what do you think of using the buffer scheme in mesh to simply make sure trtl apply layers in order? it's not very space efficient but we are talking about small numbers here. see `nextValidLayers`+`maxValidatedLayer` in mesh.go

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"context"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+// LayersIterator is used to traverse organized layers.
+// For siplicity LayersIterator iterates over all layers unconditionally.
+type LayersIterator func(types.LayerID)
```
countvonzero: the naming suggests that it's an iterator but the code looks like Organize is the actual iterator object.
it seems this is meant to be the function to execute when layer can found in consecutive order.

how about `LayerCallback` or something...

tortoise/organizer/organizer.go
dshulyak: this is also to help with https://github.com/spacemeshos/go-spacemesh/issues/2782 , verified layer may progress much slower compared with synced layer

in general i think that code needs to be robust without worrying about the specifics of the hare or other parts. if small window is sufficient it can be defined at the startup with option. but for this module to be correct it needs to handle window overflow, i thought about three options in such case:
- leak memory
- panic
- apply layers with a gap

tortoise/organizer/organizer.go
dshulyak: the action is to submit whole window and set submitted to the layer that precedes received

tortoise/organizer/organizer.go
dshulyak: circular buffer is a natural fit for this use case. it is in general very common pattern and it also helps with correctness. i know that the code with the array is robust and will never leak, otherwise, there will be panic. with map the surface for bugs is larger. in general, i don't like to replace it with a map. i will encapsulate add and pop operations to make it more intuitive.

tortoise/organizer/organizer.go
dshulyak: this is about submitting whole window in case of window overflow

tortoise/verifying_algorithm.go
dshulyak: it was used in some healing tests, i could probably refactor that in this change, but would rather do it later

countvonzero: but it seems that self-healing just needs the late blocks to be part of the mesh. i'd imagine adding the late blocks to the layer is relatively straightforward and can be done in this PR? if not as simple as i described, does that mean self-healing require some part of this code path?

tortoise/organizer/organizer.go
countvonzero: can you help me understand by describing how a window overflow of 1000 can happen?

tortoise/organizer/organizer.go
countvonzero: my understanding of the code, wrt https://github.com/spacemeshos/go-spacemesh/issues/2782, is that we picked a random base block greedily. we did not try to find a base block that has the least diff (most similar opinions) from our own local opinions, which cause this to happen more than it should.

i am not comparing verified layer with synced/processed layer. but with the hare params, i cannot imagine why we need 1000 buckets for delayed verified layers.

tortoise/organizer/organizer.go
countvonzero: i agree that circular buffer is a very common pattern but mostly in libraries (like resource pooling) where performance is critical. my judgement here is that this is more related to application logic and readability/clarity out-weighs performance here.

encapsulating it sounds like a good idea. making it a library so that it can be reused by the valid layers and processed layers in mesh.go is even better.

tortoise/organizer/organizer.go
dshulyak: there are two cases that i anticipate:
- at the time when the node is synced and runs hare - verified layer may lag behind synced by a lot, i noticed it on cloud nodes, maybe this is due to lack of resources
- due to the bug, like with that block not found after hare terminated issue

tortoise/organizer/organizer.go
dshulyak: as i mentioned this is configurable, it just costs nothing to use 1000 that's why i set it as default.

regarding https://github.com/spacemeshos/go-spacemesh/issues/2782, as i roughly understand it works as:
1. a node with a gap in history created block A in layer 12 with base block B from layer 10
2. another node registered block A as good and used it as a base block for block C in layer 13. because block A misses opinions for block in layer 11, we add them as exceptions
3. every node will not mark block C as good block 

so there are several problems:
1. creating blocks with gaps in history is harmful, and it can either happen due to out-of-order hare or delay between synced and verified layers
2. base block selection needs adjusting

tortoise/verifying_algorithm.go
dshulyak: i don't think that self-healing can depend on HandleLateBlocks in any meaningful way that can work. also the test is weird, it hand-crafted list of blocks that it wants to process and it fails if it re-processes all of them. i would rather rewrite whole test with declarative setup instead of debugging this one

dshulyak: i removed HandleLateBlocks

tortoise/organizer/organizer.go
dshulyak: i moved all buffer code to layerBuffer struct, i think the difference with map is minimal wrt to complexity

tortoise/organizer/organizer.go
countvonzero: > regarding #2782, as i roughly understand it works as:
> 
> 1. a node with a gap in history created block A in layer 12 with base block B from layer 10
> 2. another node registered block A as good and used it as a base block for block C in layer 13. because block A misses opinions for block in layer 11, we add them as exceptions
> 3. every node will not mark block C as good block

in my view, this still comes down to selecting base block with the least diff.
i.e. the node that produce block C should be able to find another base block that doesn't have gap in layer 11 as the minimal diff, instead of using block A since it's the first good block it encounters.

granted if all nodes are stuck, then the network is still left without a good base block.

dshulyak: i agree regarding base blocks and i am going to implement the algorithm described in https://github.com/spacemeshos/research/issues/33#issuecomment-830448138 with some modifications.

but also we should not be even creating such blocks if we can, this is why the buffer is set to 1000 here, because it costs nothing but makes behavior in line with the correct behavior. do you disagree with this or with something else?

countvonzero: i don't disagree with 1000. like you said it costs nothing. i am just not able to wrap my brains around how it could lag till 1000 once layer order is enforced. let's move ahead with what you currently have. thanks.

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,143 @@
...
+func WithLastLayer(lid types.LayerID) Opt {
+	return func(o *Organizer) {
+		o.submitted = lid
+		o.received = lid
+	}
+}
+
+// WithWindowSize configures size of the window.
+// By default window of the size 1000 is used.
+func WithWindowSize(size int) Opt {
```
countvonzero: :s/WithWindowSize/WithBufferSize/

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+// Callback is used to send iterated layers to the caller.
+// For siplicity Callback iterates over all layers unconditionally.
```
countvonzero: nit: simplicity

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,143 @@
...
+	if lid.After(o.submitted.Add(uint32(len(o.buf)))) {
+		o.logger.WithContext(ctx).With().Error("tortoise organizer received a layer that does not fit in the current window."+
+			" subbmiting pending with gaps",
+			log.Int("window", len(o.buf)),
+			log.Named("submitted", o.submitted),
+			log.Named("received", lid),
+		)
+		o.submitAll(f)
+		// after all layers are submitted update submitted to the layer before this one
+		o.submitted = lid.Sub(1)
```
countvonzero, line 81: i don't understand why you have to overwrite this to `lid.Sub(1)`. isn't it updated correctly in `submitAll()`?

say `o.submitted` is 100 and `lid` is 1200, and `o.received` is 1150, don't we now have a gap (after `submitAll()`) btwn 1151 and 1200?

tortoise/organizer/organizer.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		o.submitted = lid
+		f(lid)
+	}
+}
+
+// submitPending submits all non-nil layers in (o.submitted, o.received].
+func (o *Organizer) submitPending(f Callback) {
+	for lid := o.submitted.Add(1); !lid.After(o.received); lid = lid.Add(1) {
+		_, exists := o.buf.pop(lid)
```
countvonzero: should we double-check that lid == the popped layerID? it seems a severe programming error if not equal.

countvonzero (approved):

tortoise/organizer/organizer.go
dshulyak: i have a check that when layer is updated is is "empty", can add a similar one on pop

tortoise/organizer/organizer.go
dshulyak: > i don't understand why you have to overwrite this to lid.Sub(1). isn't it updated correctly in submitAll()?

submitAll updates only to the last layer that was received, and since there is no room for new layer o.received can't be updated beforehand

 > say o.submitted is 100 and lid is 1200, and o.received is 1150, don't we now have a gap (after submitAll()) btwn 1151 and 1200?

yes there will be an arbitrary number of gaps between 100 and 1200, including one between 1151 and 1200, hence there is an ERROR log here

dshulyak:
bors merge

dshulyak:
bors merge

PR #2927, opened 2021-11-10
[Merged by Bors] - add test for healing from vote blocks with different beacon

countvonzero:
## Motivation

follow-up on https://github.com/spacemeshos/go-spacemesh/pull/2916. add test

## Changes

add test for healing counting votes from bad beacon blocks

## Test Plan

unit test

## DevOps Notes

### DIFF
--- a/tortoise/verifying_tortoise_test.go
+++ b/tortoise/verifying_tortoise_test.go
@@ -365,6 +365,10 @@ var (
 )
 
 func makeLayer(t *testing.T, layerID types.LayerID, trtl *turtle, blocksPerLayer int, atxdb atxDataWriter, msh blockDataWriter, inputVectorFn func(id types.LayerID) ([]types.BlockID, error)) *types.Layer {
+	return makeLayerWithBeacon(t, layerID, trtl, nil, blocksPerLayer, atxdb, msh, inputVectorFn)
+}
+
+func makeLayerWithBeacon(t *testing.T, layerID types.LayerID, trtl *turtle, beacon []byte, blocksPerLayer int, atxdb atxDataWriter, msh blockDataWriter, inputVectorFn func(id types.LayerID) ([]types.BlockID, error)) *types.Layer {
 	logger := logtest.New(t)
 	logger.Debug("======================== choosing base block for layer", layerID)
 	if inputVectorFn != nil {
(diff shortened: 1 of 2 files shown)


tortoise/verifying_tortoise_test.go
```diff
@@ -3063,3 +3108,42 @@ func TestVoteBlockFilterForHealing(t *testing.T) {
 	assert.True(t, filter(goodBlock.ID()))
 	assert.True(t, filter(badBlock.ID()))
 }
+
+func TestSelfHealingCountBadBeaconBlocks(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	// increase layer size reduce test flakiness
+	s := sim.New(sim.WithLayerSize(defaultTestLayerSize * 10))
```
dshulyak: generator should be in charge of the state. this commit makes it depend on additional mocking, if beacons need to be supported by sim module then sim module should store it somewhere, preferably in the same database that will be used by tortoise beacon, so that benchmarking is simplified

tortoise/sim/layer.go
```diff
@@ -32,6 +33,12 @@ func WithNextReorder(delay uint32) NextOpt {
 	}
 }
 
+func WithNextBeacon(beacon []byte) NextOpt {
```
dshulyak: i am not sure that this is what i planned for it, this makes a beacon a special case that depends on some external info.
i think that i definitely want to generate good beacon at Setup stage, but idk what is the best way to handle beacon changes. maybe there should be another action Beacon().

for example, i am adding action called [Partition](https://github.com/dshulyak/go-spacemesh/commit/0390a903eccb03b6817684ddcd6dc56b3d33498d#diff-9c8debd08fda2ee55a056f1dd4207fd57f10420f2a9791f7a57e7e9f36afc20c) 

the idea is that some layer on top of generator will select action that leads to state change and construct sequence of such actions randomly

countvonzero: i agree. i wasn't sure how to go about this so also added a test in `TestHealing`.
i think once you mature this simulator package maybe i can tag along the `Partition` setup to have different beacon values. for now, i'll just add the test with mocked beacon

tortoise/verifying_tortoise_test.go
countvonzero: removed.

dshulyak (approved):

countvonzero:
bors merge

PR #2928, opened 2021-11-10
[Merged by Bors] - tortoise: improve base block selection

### LINKED ISSUE #2402: Improve base block selection in verifying tortoise
## Description
Right now, the algorithm for choosing a base block is very naive: it chooses the most recent block marked "good." This can be improved and optimized in several important ways:

- choose a base block by first disagreement layer (which can be outside the sliding window)
- then, by the size of the exception list

See https://github.com/spacemeshos/research/issues/33#issuecomment-830448138 for more

This issue appears in commit hash: (see https://github.com/spacemeshos/research/issues/33#issuecomment-830448138)

Related files (optionally with line numbers): tortoise/verifying_tortoise.go

### LINKED ISSUE #2782: tortoise: good block candidate contains exception for block older than its base block

dshulyak:
## Motivation

Whenever some node misses part of the state it will produce blocks that can be considered good by the protocol, but that will make other blocks non-good if such block will be selected as a base block. To prevent this issue we need to prioritize blocks that are consistent with our local opinions.

For example, if node misses (erroneously) a previous layer when the base block is created it will produce block that will be considered good, but if another node selects this new block as base block it will have to add exceptions for earlier blocks, which is contradicts good block definition. 

closes https://github.com/spacemeshos/go-spacemesh/issues/2782
closes https://github.com/spacemeshos/go-spacemesh/issues/2402

## Changes
- rework base block selection algorithm, it will prioritize blocks based on the following criteria:
  - good blocks (to ensure that verifying tortoise can make progress)
  - blocks that vote consistently with our local opinion
  - blocks from higher layers
- base block selection is now deterministic, two nodes with same local opinion and voting info will choose the same base block
- introduce tcontext struct that is used for caching within a span of a single request
- rewrote TestLayerPatterns using sim module, i wasn't able to make `heal after bad layers` halt because some test setup code was interfering with BaseBlock in a way that prevented network from stalling. with sim module it is easier to see what exactly are the failures and when tortoise recovers

## Test Plan
unit tests for previously observed error condtions

## TODO

## DevOps Notes


dshulyak:
bors try

tortoise/sim/votes.go
```diff
@@ -0,0 +1,25 @@
...
+// Voting contains blocks voting.
+type Voting struct {
+	Base                      types.BlockID
+	Support, Against, Abstain []types.BlockID
+}
+
+// VotesGenerator allows to replace default votes generator.
+type VotesGenerator func(rng *rand.Rand, layers []*types.Layer) Voting
+
+// perfectVoting select base block from previouos layer and supports all blocks from previous layer.
```
nkryuchkov, line 18:
```suggestion
// perfectVoting select base block from previous layer and supports all blocks from previous layer.
```

nkryuchkov (approved):

tortoise/context.go
```diff
@@ -0,0 +1,79 @@
...
+		ValidBlocks:  map[types.LayerID][]types.BlockID{},
+	}
+}
+
+type tcontext struct {
+	context.Context
+	LocalOpinion map[types.LayerID]map[types.BlockID]vec
+	LayerBlocks  map[types.LayerID][]types.BlockID
+	InputVectors map[types.LayerID][]types.BlockID
+	ValidBlocks  map[types.LayerID][]types.BlockID
```
countvonzero: please comment what each field represent.

tortoise/context.go
```diff
@@ -0,0 +1,79 @@
...
+	return &tcontext{
+		Context:      ctx,
+		LocalOpinion: map[types.LayerID]map[types.BlockID]vec{},
+		LayerBlocks:  map[types.LayerID][]types.BlockID{},
+		InputVectors: map[types.LayerID][]types.BlockID{},
+		ValidBlocks:  map[types.LayerID][]types.BlockID{},
+	}
+}
+
+type tcontext struct {
```
countvonzero, line 21: please comment on the purpose of this struct. i am guessing you are using it to carry the cache around for a given context.

i am a bit worried about misusing of this struct. future callers unwittingly use the cached value when s/he doesn't intend to. caching with context  is always controversial...

tortoise/verifying_tortoise.go
```diff
@@ -1208,11 +1203,9 @@ func (t *turtle) heal(ctx context.Context, targetLayerID types.LayerID) {
 		// Note: we look at ALL blocks we've seen for the layer, not just those we've previously marked contextually valid
 		logger.Info("self healing attempting to verify candidate layer")
 
-		layerBlockIds, err := t.bdp.LayerBlockIds(candidateLayerID)
+		layerBlockIds, err := t.getLayerBlocksIDs(ctx, candidateLayerID)
```
countvonzero, line 1187: this is not correct. we want to include late blocks that got added to the mesh too. we don't want cached value here.

for self-healing, we should use the vanilla context.

tortoise/verifying_tortoise_test.go
```diff
@@ -432,175 +433,99 @@ func testLayerPattern(t *testing.T, atxdb atxDataWriter, db blockDataWriter, trt
...
-		trtl.AvgLayerSize = blocksPerLayer
-		trtl.bdp = msh
-		trtl.atxdb = atxdb
-		trtl.init(context.TODO(), mesh.GenesisLayer())
-		numGood := 5
-		pattern := make([]bool, numGood)
-		for i := 0; i < numGood; i++ {
-			pattern[i] = true
-		}
-		testLayerPattern(t, atxdb, msh, trtl, blocksPerLayer, pattern)
```
countvonzero, line 449: delete the func `testLayerPattern()`

tortoise/verifying_tortoise_test.go
```diff
@@ -432,175 +433,99 @@ func testLayerPattern(t *testing.T, atxdb atxDataWriter, db blockDataWriter, trt
...
+			sim.WithSequence(5),
+			sim.WithSequence(2, sim.WithHareFailure()),
+		) {
+			last = lid
+			_, verified, _ = tortoise.HandleIncomingLayer(ctx, lid)
+		}
+		require.Equal(t, genesis.Add(5), verified)
 		vote := abstain
 		numLayersToFullyHeal := 0
 		for i := 0; vote == abstain; i++ {
```
countvonzero, line 466: is it possible to calculate `numLayersToFullyHeal` earlier so you can GentLayers on all 5+2+`numLayersToFullyHeal` at the same time?

tortoise/verifying_tortoise_test.go
```diff
@@ -1060,13 +985,6 @@ func TestBaseBlock(t *testing.T) {
 	alg.HandleIncomingLayer(context.TODO(), l3.Index())
 	require.Equal(t, int(types.GetEffectiveGenesis().Add(1).Uint32()), int(alg.LatestComplete().Uint32()))
 	expectBaseBlockLayer(l2.Index(), 0, defaultTestLayerSize, 0)
-
-	// mark all blocks bad
-	alg.trtl.GoodBlocksIndex = make(map[types.BlockID]bool, 0)
-	baseBlockID, exceptions, err := alg.BaseBlock(context.TODO())
-	r.Equal(errNoBaseBlockFound, err)
```
countvonzero, line 1067: why remove this test? i thought it's testing that it does return `errNoBaseBlockFound`

tortoise/verifying_tortoise_test.go
```diff
@@ -3107,3 +3056,100 @@ func TestVoteBlockFilterForHealing(t *testing.T) {
 	assert.True(t, filter(goodBlock.ID()))
 	assert.True(t, filter(badBlock.ID()))
 }
+
+// gapVote will skip one layer in voting.
+func gapVote(rng *mrand.Rand, layers []*types.Layer) sim.Voting {
+	if len(layers) == 1 {
```
countvonzero: < 2

tortoise/verifying_tortoise_test.go
```diff
@@ -3107,3 +3056,100 @@ func TestVoteBlockFilterForHealing(t *testing.T) {
...
+		panic("need atleast 2 layers")
+	}
+	baseLayer := layers[len(layers)-2]
+	support := layers[len(layers)-2].BlocksIDs()
+	base := baseLayer.Blocks()[rng.Intn(len(baseLayer.Blocks()))]
+	return sim.Voting{Base: base.ID(), Support: support}
```
countvonzero, line 3051: is the vote gap in `layers[len(layers)-1]` because this returned vote is for the new layer?

tortoise/verifying_tortoise_test.go
```diff
@@ -3107,3 +3056,100 @@ func TestVoteBlockFilterForHealing(t *testing.T) {
...
+}
+
+// olderExceptions will vote for block older then base block.
+func olderExceptions(rng *mrand.Rand, layers []*types.Layer) sim.Voting {
+	if len(layers) == 1 {
```
countvonzero: < 2

tortoise/verifying_tortoise_test.go
```diff
@@ -3107,3 +3056,100 @@ func TestVoteBlockFilterForHealing(t *testing.T) {
...
+		panic("need atleast 2 layers")
+	}
+	baseLayer := layers[len(layers)-1]
+	base := baseLayer.Blocks()[rng.Intn(len(baseLayer.Blocks()))]
+	voting := sim.Voting{Base: base.ID()}
+	for _, layer := range layers[len(layers)-2:] {
+		for _, bid := range layer.BlocksIDs() {
+			voting.Support = append(voting.Support, bid)
```
countvonzero, line 3064: so the new block is adding support explicitly for blocks in the last 2 layers, layer N-1 and layer N.
the base block (layer N) should've had explicit support for blocks in layer N-1. so the new block and base block actually agree. i think this artificial exception is confusing.

how about for the new block just add a voting.Neutral or Against for blocks in layer N-1?

tortoise/context.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	opinion, err := t.computeLocalOpinion(ctx, lid)
+	if err != nil {
+		return nil, err
+	}
+	ctx.LocalOpinion[lid] = opinion
+	return opinion, nil
+}
+
+func (t *turtle) getInputVector(ctx *tcontext, lid types.LayerID) ([]types.BlockID, error) {
```
countvonzero, line 46: there is potentially a race here. input vector can be updated by syncer across sync attempts.
i think we need to put a TODO to revisit whether it's ok to cache the input vector here

tortoise/context.go
```diff
@@ -0,0 +1,79 @@
...
+	bidsmap, err := t.bdp.LayerContextuallyValidBlocks(ctx, lid)
+	if err != nil {
+		return nil, fmt.Errorf("read valid blocks for layer %s: %w", lid, err)
+	}
+	bids = blockMapToArray(bidsmap)
+	ctx.ValidBlocks[lid] = bids
+	return bids, nil
+}
+
+func (t *turtle) getLayerBlocksIDs(ctx *tcontext, lid types.LayerID) ([]types.BlockID, error) {
```
countvonzero, line 73: there is a similar caching problem here. sometimes we want to consider late blocks that's added to the mesh after the block IDs are cached

tortoise/sim/generator.go
```diff
@@ -59,15 +66,15 @@ type State struct {
 // New creates Generator instance.
 func New(opts ...GenOpt) *Generator {
 	g := &Generator{
-		rng:       rand.New(rand.NewSource(time.Now().UnixNano())),
+		rng:       rand.New(rand.NewSource(0)),
```
countvonzero, line 69: curious why you made this change? do you want the random sequence to be stable?

tortoise/sim/layer.go
```diff
@@ -88,10 +103,13 @@ func (g *Generator) genLayer() types.LayerID {
 		}
 		layer.AddBlock(block)
 	}
-	// TODO saving layer input should be skipped as one of the error conditions
-	if err := g.State.MeshDB.SaveLayerInputVectorByID(context.Background(), layer.Index(), layer.BlocksIDs()); err != nil {
-		g.logger.With().Panic("failed to save layer input vecotor", log.Err(err))
+	if !cfg.FailHare {
```
countvonzero, line 106: the simulation should set the votes for layers that failed hare to `Against`

tortoise/verifying_tortoise_test.go
countvonzero, line 3051: a gap should be manifested as Abstain votes. i.e. if you want to simulate a gap in `baseLayer+1`, the returned votes should be
```
abstain := layers[len(layers)-1].BlocksIDs()
sim.Voting{Base: base.ID(), Support: support, Abstain: abstain}
```

tortoise/verifying_tortoise.go
```diff
@@ -834,30 +863,16 @@ candidateLayerLoop:
 		// note: if the following checks fail, we just return rather than trying to verify later layers.
 		// we don't presently support verifying layer N+1 when layer N hasn't been verified.
 
-		layerBlockIds, err := t.bdp.LayerBlockIds(candidateLayerID)
-		if err != nil {
-			return fmt.Errorf("%s %v: %w", errstrCantFindLayer, candidateLayerID, err)
```
countvonzero, line 839: `errstrCantFindLayer` now unused

tortoise/verifying_tortoise.go
```diff
@@ -305,89 +265,158 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
...
-		for blockID, opinion := range t.BlockOpinionsByLayer[layerID] {
-			if _, ok := t.GoodBlocksIndex[blockID]; !ok {
-				logger.With().Info("not considering block not marked good as base block candidate", blockID)
-				continue
-			}
+	var (
+		tctx          = wrapContext(ctx)
+		logger        = t.logger.WithContext(ctx)
+		disagreements = map[types.BlockID]types.LayerID{}
+		choices       []types.BlockID // choices from the best to the least bad
```
countvonzero, line 306: i think we should memoize this list of choices as we do `HandleIncomingLayers` instead of making the same computation every time.

disagreement with older blocks won't change as time moves on. their votes stayed the same. the two moving parts that could influence finding a better base-block candidate are
- our local opinions change due to `hdist` cutoff
- new blocks from new layers come in

so in the memoization, for verified layers in the window, we don't need to re-calculate the disagreement. but for layers that are within `hdist` window we do recalculate when the disagreements.

this will also prevent the situation where the base block we choose here goes immediately out of the sliding window.

tortoise/verifying_tortoise.go
```diff
@@ -305,89 +265,158 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
...
+	}
 
-			logger.With().Info("chose base block",
-				blockID,
-				log.Int("against_count", len(exceptionVectorMap[0])),
-				log.Int("support_count", len(exceptionVectorMap[1])),
-				log.Int("neutral_count", len(exceptionVectorMap[2])))
-
-			metrics.LayerDistanceToBaseBlock.WithLabelValues().Observe(float64(t.Last.Value - layerID.Value))
+	sort.Slice(choices, func(i, j int) bool {
```
countvonzero: this sorting should be its own function be tested independently.

tortoise/verifying_tortoise.go
```diff
@@ -305,89 +265,158 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
...
+
+	// if the block is good we know that there are no exceptions that precede base block.
+	// in such case current block will not fix any previous disagreements of the base block.
+	// or if the base block completely agrees with local opinion compute disagreements after base block.
+	if _, exist := t.GoodBlocksIndex[bid]; exist {
+		block, err := t.bdp.GetBlock(bid)
+		if err != nil {
+			return types.LayerID{}, fmt.Errorf("reading block %s: %w", bid, err)
+		}
+		basedis, exist := disagreements[block.BaseBlock]
```
countvonzero, line 368: if block.BaseBlock is also in GoodBlocksIndex, you can return consistent right away

tortoise/verifying_tortoise.go
```diff
@@ -305,89 +265,158 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
...
+		if exist {
+			baselid := t.BlockLayer[block.BaseBlock]
+			if basedis.Before(consistent) {
+				return basedis, nil
+			}
+			from = baselid
+		}
+	}
+	for lid := from; lid.Before(blid); lid = lid.Add(1) {
```
countvonzero, line 377: do you need to check if `blid` is within the sliding window?

tortoise/verifying_tortoise.go
```diff
@@ -305,89 +265,158 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
...
+	// if the block is good we know that there are no exceptions that precede base block.
+	// in such case current block will not fix any previous disagreements of the base block.
+	// or if the base block completely agrees with local opinion compute disagreements after base block.
+	if _, exist := t.GoodBlocksIndex[bid]; exist {
+		block, err := t.bdp.GetBlock(bid)
+		if err != nil {
+			return types.LayerID{}, fmt.Errorf("reading block %s: %w", bid, err)
+		}
+		basedis, exist := disagreements[block.BaseBlock]
+		if exist {
```
countvonzero, line 369: because the loop in `BaseBlock()` starts from `t.LastEvicted.Add(1)`, if `basedis` doesn't exist in `disagreements`, that means the base block is outside the sliding window?

tortoise/verifying_tortoise.go
```diff
@@ -305,89 +265,158 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
...
+		from     = t.LastEvicted.Add(1)
+		// using it as a mark that the votes for block are completely consistent
+		// with a local opinion. so if it two blocks have consistent histories select block
+		// from a higher layer as it is more consistent.
+		consistent = t.Last
+	)
+
+	// if the block is good we know that there are no exceptions that precede base block.
+	// in such case current block will not fix any previous disagreements of the base block.
+	// or if the base block completely agrees with local opinion compute disagreements after base block.
```
countvonzero, line 362: i don't understand this comment.
if we already considered `bid` as a good block, that means
- we once considered base block of `bid` as a good block.
- our local opinion agrees with all exceptions that `bid` has.

so what do you mean by `current block will not fix any previous disagreements of the base block`

countvonzero (commented):
nice switching the test to use sim package

countvonzero:
two more thoughts:
- with this PR we may select a base block that was not considered good 
- i wonder if we should give priority to blocks that has smaller distance to its base block?

tortoise/context.go
dshulyak: what field is not obvious?

tortoise/context.go
dshulyak: i dont think that there is any new race we should worry about, context is used for a single request and then discarded

tortoise/context.go
dshulyak: context is not reused between requests

tortoise/sim/generator.go
dshulyak: yes, seed needs to random only during simulation not regular tests

tortoise/verifying_tortoise.go
dshulyak: good idea, will consider to implement

the problem with choosing base block out of sliding window though is that everyone else will evict it,so base block outside of sliding window is useless

tortoise/verifying_tortoise.go
dshulyak: we don't consider implicit votes against the blocks when we are considering if block is good, so block might be missing some votes, which is exactly the problem i am addressing in this pr

tortoise/verifying_tortoise.go
dshulyak: no see comment above

tortoise/verifying_tortoise.go
dshulyak: yes

tortoise/verifying_tortoise.go
dshulyak: it is within the sliding window by definition, because we iterate from evicted + 1 to last

tortoise/verifying_tortoise.go
dshulyak: i dont understand your concern. context is not reused between requests. if blocks were in when self healing started they will be found.

tortoise/verifying_tortoise_test.go
dshulyak: no, in the gap you don't even know about layer-1, so you vote for them only implicitly

tortoise/verifying_tortoise_test.go
dshulyak: this is not how it works, imagine that you asked tortoise to produce base blocks at layer 12 and it only has blocks from 10, how it can add votes for what it doesn't have?

tortoise/verifying_tortoise.go
dshulyak: i don't like to test private interfaces, it makes refactoring harder. this is tested you can check the coverage

dshulyak:
> with this PR we may select a base block that was not considered good

this is intended, we should be able to continue in self-healing mode if verifying tortoise fails for any reason

> i wonder if we should give priority to blocks that has smaller distance to its base block?

not sure if it matters, we prefer blocks from higher layers if they have consistent opinions

tortoise/verifying_tortoise_test.go
dshulyak: the point is to check that the layer is stuck

tortoise/verifying_tortoise_test.go
dshulyak: i removed this assertion because we are selecting bad block if good is not found, will probably add a test with empty state to check this, otherwise we will always select something

tortoise/context.go
dshulyak: yes, it is to cache things.

it is initialized once per call in HandlingIncomingLayer and BaseBlock

tortoise/sim/layer.go
dshulyak: what if it failed only the node? i will change this condition to NoInpuctVector then

tortoise/verifying_tortoise.go
dshulyak: it may matter if contextually valid blocks will change, so i will re-init context in that case

tortoise/verifying_tortoise.go
dshulyak: i think i will keep it as is for simplicity. as i mentioned we don't want blocks from that are evicted, bad block that is not evicted will allow full tortoise to make progress, evicted blocks will stall the network eventually. so this is only about the performance, and it is sufficient for now

tortoise/verifying_tortoise.go
dshulyak: sorry comment below

tortoise/verifying_tortoise_test.go
dshulyak: not sure i understood the question. gap vote is a votes for block at layer 12 that was requested before we knew about layer 11

tortoise/context.go
dshulyak: it is also not shared between concurrent requests, if such request are allowed by the mutex

tortoise/verifying_tortoise_test.go
dshulyak: sorry, i somehow misplaced my comments.

in this case it doesn't matter if the exception is artificial or not, it is just to mark this block not good

tortoise/verifying_tortoise.go
countvonzero: in self healing, we dont just look at contextual valid blocks, we looked at all blocks, including blocks that arrived late.

so it is possible that we have asked the blocked IDs for this layer and cached it before entering self healing. then a late block arrived after the caching took place. especially during the network rejoin, we want to get as many blocks as possible during self healing. caching here is counter productive

dshulyak: ah thats not exactly the case, for late block to take effect it needs to be processed, and every layer after should be re-processed. otherwise, its votes are mostly useless. so what you are describing is only relevant during rerun.

self-healing during normal operation is only about dealing with stuck verified layer due to the absence of good blocks.

dshulyak: >  then a late block arrived after the caching took place. especially during the network rejoin, we want to get as many blocks as possible during self healing. caching here is counter productive

also regardless of how it works now, relying on such non-determinism is a mistake anyway

tortoise/context.go
countvonzero: i don't mean race within the context or concurrent requests. i mean that syncer can save a different set of input vector during syncing through mesh.go and not controlled by the mutex

tortoise/context.go
countvonzero: yes it's not reused between requests. but within the same request, we may try to do different things that sometimes caching is desired and sometimes not. that's what i mean by it has always been controversial approach. on one hand it's easy. on the other hand it makes it easy for future devs to make coding mistakes. anyway, at least you use a different struct `tcontext` instead of hiding everything in `context`.

tortoise/verifying_tortoise.go
countvonzero: > self-healing during normal operation is only about dealing with stuck verified layer due to the absence of good blocks.

i disagree. in the general terms, self-healing is about being able to heal from temporary malicious majority or network partition. which means we want to count votes that are late or use different beacons.

to that end, we should not rely on the in-memory opinion maps to count votes. i agree the current code is as you described, but i disagree we should keep it that way. we should reload/reprocess all the votes from disk.

in any case, i don't see any harm using just plain context here in the healing case. but then i'm ok if you leave it.

tortoise/verifying_tortoise.go
countvonzero: this is more like a utility function than a private interface. the value of testing here is if in the future i want to change the algorithm, i can run that test and easily know that it's this utility function that i broke, and not other parts that does base block selection.

tortoise/verifying_tortoise_test.go
countvonzero: this was not the comment i remembered posted. i had several windows open for this so there must have been a human error. 

what i meant to say is that if you want to simulate voting for layer 12 before knowing about layer 11, then we should return
```
neutral := layers[len(layers)-1].BlocksIDs()
return sim.Voting{Base: base.ID, Support: support, Abstain: neutral}
```

tortoise/verifying_tortoise.go
dshulyak: > we should reload/reprocess all the votes from disk.

we do that during full rerun periodically. in the future, i assume, we will have an interface for triggering rerun based on sync or other conditions. we can do it even now by counting number of late blocks in block handler and trigger rerun, but is it worth it?

i am not really sure that this is viable to reprocess blocks in self-healing unconditionally.

tortoise/context.go
dshulyak: that an anti pattern in my opinion, i think that HandleIncomingLayer or BaseBlock is as an atomic action that doesn't change its state in the middle of the action. if something is changed from initial condition the action should be repeated

tortoise/verifying_tortoise_test.go
countvonzero: wait. we are creating vote for the next layer in the generator, right?
say in `layers` we have [6,7,8,9,10], and we are creating votes for layer 11.
and we want to pretend that we haven't seen layer 10. correct?

tortoise/verifying_tortoise.go
countvonzero: when i read your code here, i thought the problem is that we currently don't generate AgainstDiff. so even if we find a base block that's good, it's not really good.

your comment only make sense when we don't generate AgainstDiff.

countvonzero (approved):
still have a couple more questions

tortoise/verifying_tortoise_test.go
countvonzero: i think it matters for the reader. and it takes so little effort to make it correct, changing it from support to against, why not?

tortoise/verifying_tortoise.go
countvonzero: > i am not really sure that this is viable to reprocess blocks in self-healing unconditionally.

my understanding is it's by design. but with rerun i am not sure anymore the responsibility of each routine....

tortoise/verifying_tortoise_test.go
dshulyak: the thing is that this is exactly how votes looked when we selected a base block that had a "gap" votes. 

example, 
block A from layer 12 uses base block B from layer 10, and no exceptions for layer 11 - this is block with gap vote, it is still good.
block C from layer 13 uses block A as base block, but because block A had to exceptions for layer 10 we have to add them ourselves, so block C will have exceptions for 12 and 11.

tortoise/verifying_tortoise_test.go
dshulyak: yes, so Last is 9 in tortoise state

tortoise/verifying_tortoise.go
dshulyak: AgainstDiff will be populated with votes that are different from base block votes.

it doesn't change implicit votes, e.g. votes for blocks that we are not aware of, for example a gap vote in layer 12 with base block at layer 10 will implicitly vote against blocks at layer 11. so even if block 10 is good, we still need to check that implicit votes for layer 11 match our local opinion

countvonzero: > AgainstDiff will be populated with votes that are different from base block votes.

we still don't add AgainstDiff to the block. 
https://github.com/spacemeshos/go-spacemesh/blob/6a93e4781ccb5091a0b4dcaad811f4ebb8aa0dc8/tortoise/verifying_tortoise.go#L506

what am i missing?

tortoise/verifying_tortoise_test.go
countvonzero: i totally misunderstood what gap vote is supposed to be. i thought its when hare hasn't terminate for layer 11 when generating blocks for layer 12.

tortoise/verifying_tortoise.go
dshulyak: yes we are not adding, i guess i don't understand what we are discussing.

> your comment only make sense when we don't generate AgainstDiff.

why is that? if there was a disagreement because the base block didn't vote for some layer, and a new block is voting for that layer - such block will not be marked good, but it may fix base block disagreements, so we need to do more work. however if we know that the block is good - then all we need to check if later votes are consistent with local opinion, including implicit votes.

tortoise/context.go
countvonzero: i am thinking as a new team member reading this code. not for myself.

tortoise/verifying_tortoise.go
countvonzero: let me start by saying, the bug is weird in the first place. if we always choose a good block in layer N, and the definition of a good block is that 
- its base block in layer M was a good block according to the local opinions
- the block had no disagreement with our local opinions between layer M and N

then why do we end up with exceptions older than the base block? the only reason i can think of is that we are not adding AgainstDiff to blocks and so the expanded votes based on base block is incorrect. and somehow in the code we end up with disagreements even btwn good blocks in different layers.

so when you are comments are talking about disagreement even with the base block (which we locally considered a good block), i immediately thought about the lack of AgainstDiff in the diff list.

> it doesn't change implicit votes, e.g. votes for blocks that we are not aware of, for example a gap vote in layer 12 with base block at layer 10 will implicitly vote against blocks at layer 11. so even if block 10 is good, we still need to check that implicit votes for layer 11 match our local opinion

i still don't understand what you mean by "votes for blocks that we are not aware of". 

let's for a moment disregard what the code does and discuss what should be in the data.
when we are creating votes for block `12B` in layer 12. we chose block `10A` in layer 10 to be base block. 
two scenarios:
- layer 11 has not terminated yet. then block `12B` should have have 
```
base block: 10A
forDiff: all block IDs in layer 10
againstDiff: empty
neutralDiff: all block IDs in layer 11
```
- layer 11 terminated with failure. then block `12B` should have have 
```
base block: 10A
forDiff: all block IDs in layer 10
againstDiff: all block IDs in layer 11
neutralDiff: empty
```
these are the Block data over the wire.
implicit against votes should only happen when both block `12B` and block `10A` voted against a block in a past layer, say `9Z`. over the wire, block `10A` should still have `9Z` in its AgainstDiff, but `12B` will not, because it agrees with `10A`

so i think what the vote gap you are referring to is in the second case when hare terminated layer 11 with failure, block `12B` actually looked like
```
base block: 10A
forDiff: all block IDs in layer 10
againstDiff: empty
neutralDiff: empty
```
i think this is where the confusion come from during our discussion. the `implicit against` is too confusing and mind-twisting especially with the recursive nature of the base block logic.

dshulyak: thanks for the detailed explanation, it all makes sense except that i am not sure that the conditions that you specified are exhaustive.

1. layer 11 terminated with failure. then block 12B should have have
2.

dshulyak: thank you for the detailed explanation, it all makes sense but i don't think that the specified conditions are exhaustive.

```
base block: 10A
forDiff: all block IDs in layer 10
againstDiff: all block IDs in layer 11
neutralDiff: empty
```

in this example, you assume that we know all block ids that were in the layer 11. f the node doesn't know any or knows only 1 out of 10, how this encoding will change?

this is why i don't think that we can get rid of "implicit" votes against unknown blocks.

so for me, it is a question of when do we consider these "implicit" votes. in this implementation, we consider them only when we count votes for blocks, and with this change, during base block selection. this is probably a mistake because...

> its base block in layer M was a good block according to the local opinions
   the block had no disagreement with our local opinions between layer M and N

of this difference with the implementation, as we don't consider implicit votes when goodness is determined it means that the block can have a disagreement with local opinion but still can be marked good. we only check explicitly that provided exceptions are consistent with local opinion.

dshulyak:
bors merge

dshulyak:
bors merge

tortoise/verifying_tortoise.go
countvonzero: > in this example, you assume that we know all block ids that were in the layer 11. f the node doesn't know any or knows only 1 out of 10, how this encoding will change?

@dshulyak 
the encoding won't change. you still add that 1 block you know about in the `againstDiff`. this means when the node is voting about layer 11, either the hare terminate layer 11 with 1 block (if we are voting within hdist) or the votes you see from the blocks only have 1 block for layer 11. the rest of 9 blocks must have appeared late in the network.

the implicit against is worse. you have no idea what the node saw for layer 11. it unequivocally votes against all blocks in a layer. doesn't matter whether the node actually saw 1 or 2 or all 10 blocks. i don't see this as an advantage for implicit votes. i see this as an disadvantage because we lose information.

countvonzero: but of course when you know there are 10 blocks in layer 11, and the vote has only one block in its `againstDiff`, you can still explicitly vote against the rest of the 9 blocks when we expand the voting vector in memory. the point is, at the time of voting, we know exactly what the node saw.

countvonzero: > of this difference with the implementation, as we don't consider implicit votes when goodness is determined it means that the block can have a disagreement with local opinion but still can be marked good. we only check explicitly that provided exceptions are consistent with local opinion.

are you sure this is not simply because we don't encode add `againstDiff` now? if we add `againstDiff` to the current implementation, a good block should satisfy this property

> its base block in layer M was a good block according to the local opinions
the block had no disagreement with our local opinions between layer M and N

and in this case, if a block B had used a good block as base block, and block B is also determined as a good block, there should not be any disagreement btwn block B and local opinions.

dshulyak: > the encoding won't change.

it will change as you noted above

> but of course when you know there are 10 blocks in layer 11, and the vote has only one block in its againstDiff, you can still explicitly vote against the rest of the 9 blocks when we expand the voting vector in memory. the point is, at the time of voting, we know exactly what the node saw.

this is was my point in a previous reply https://github.com/spacemeshos/go-spacemesh/pull/2928#discussion_r748793531 . but this is not explicit voting, since you don't add those blocks as exceptions. 

and as i mentioned this is just a question of when do you count those votes, in the current implementation we count them later, which is probably a mistake that follows from the imprecise implementation of good block.

> are you sure this is not simply because we don't encode add againstDiff now? if we add againstDiff to the current implementation, a good block should satisfy this property

i am sure, we still need to consider implicit votes as you noted above

PR #2930, opened 2021-11-10
[Merged by Bors] - Enable `-race` flag in unit tests

### LINKED ISSUE #2226: Add -race flag to tests and fix data races

nkryuchkov:
## Motivation
Closes #2226
There're some data races left after merging https://github.com/spacemeshos/go-spacemesh/pull/2608. I cannot reproduce them locally but in CI only. Therefore, I'm creating this PR to be able to check what's left to be fixed.

## Changes

- Enable `-race` flag in unit tests

## Test Plan
- Unit tests

## DevOps Notes


nkryuchkov:
On hold until https://github.com/spacemeshos/go-spacemesh/issues/3066, https://github.com/spacemeshos/go-spacemesh/issues/3067, https://github.com/spacemeshos/go-spacemesh/issues/3097, https://github.com/spacemeshos/go-spacemesh/issues/3099, https://github.com/spacemeshos/go-spacemesh/issues/3100 are closed.

nkryuchkov:
bors try

dshulyak:
do we need to merge any other changes?

nkryuchkov:
@dshulyak we need to fix https://github.com/spacemeshos/go-spacemesh/issues/3100, to check if https://github.com/spacemeshos/go-spacemesh/issues/3066 is reproducible and then to check if any other data races appear

nkryuchkov:
@dshulyak https://github.com/spacemeshos/go-spacemesh/pull/3095 should fix a mesh data race and I think it's still not merged here

poszu:
bors try

cmd/node/node.go
```diff
@@ -95,62 +95,74 @@ const (
...
-// Cmd is the cobra wrapper for the node, that allows adding parameters to it.
-var Cmd = &cobra.Command{
-	Use:   "node",
-	Short: "start node",
-	Run: func(cmd *cobra.Command, args []string) {
-		conf, err := loadConfig(cmd)
-		if err != nil {
-			log.With().Fatal("failed to initialize config", log.Err(err))
-		}
+func GetCommand() *cobra.Command {
```
fasmat, line 98: I think if one calls `GetCommand()` mulitple times this might cause issues, since there is a side effect on line 138 with

```golang
	cmdp.AddCommands(cmd)
```

Can you wrap this into a `sync.Once` to make sure that multiple calls to `GetCommand()` always return the same instance (and still be concurrency safe?)

hare/hare_rounds_test.go
```diff
@@ -124,7 +124,9 @@ func Test_HarePreRoundEmptySet(t *testing.T) {
 	var mu sync.RWMutex
 	m := [layers][nodes]int{}
 
-	w := runNodesFor(t, nodes, 2, layers, 2, false,
+	ctx, cancel := context.WithCancel(context.Background())
```
fasmat, line 127: You can add `defer cancel()` here. Calling cancel after the first time is a no-op. Without the defer, if the test fails between here and line 147 we have a dead go routine.

fasmat (changes requested):
We should take a look at the side-effect of `GetCommand()` besides that the PR lgtm

cmd/node/node.go
poszu: What is the side effect? This call adds flags to local `cmd`.

fasmat: You're right; I was confused by the name `AddCommands` and assumed that `cmd` would be added to something. A better name for the function would probably be `ParseCmdFlags` or similar. But yes `sync.Once` isn't needed here.

fasmat (approved):

countvonzero (approved):
yay!

fasmat:
I hate being a buzz-kill, but there are still some races left 🙁. Check the Windows unittests. At least some of them fail because of data races. The macOS tests seem to fail because of timing issues at least.

Also `libgpu-setup.dynlib` seems to have been built on a newer version of macOS than we are currently using in our `go-spacemesh` CI. I should double check our CI for `go-spacemesh` and `gpu-post`, so that we can support at all macOS versions since at least 11.0 (Big Sur).

poszu:
> I hate being a buzz-kill, but there are still some races left 🙁. Check the Windows unittests. At least some of them fail because of data races. The macOS tests seem to fail because of timing issues at least.
> 
> Also `libgpu-setup.dynlib` seems to have been built on a newer version of macOS than we are currently using in our `go-spacemesh` CI. I should double check our CI for `go-spacemesh` and `gpu-post`, so that we can support at all macOS versions since at least 11.0 (Big Sur).

I'm aware there are still data races in the code. The goal of this PR is to only fix failing tests and to enable `-race` so that no new data races are merged. 

Regarding windows unit tests - I'm surprised as I was pretty sure I removed the `allow-failure: true` for windows some time ago 🤔 .

poszu:
bors merge

PR #2935, opened 2021-11-12
[Merged by Bors] - tortoise: preserve block even if base block was evicted

### LINKED ISSUE #2936: tortoise: handle blocks with base block outside of the sliding window gracefully
## Description

If the network receives any block with a base block that is outside of the sliding window it will switch to self-healing mode, and eventually halt. Obviously, there are issues that allow the creation of such blocks, one of them is https://github.com/spacemeshos/go-spacemesh/issues/2934, maybe there are more.

It was also a primary reason for devnet 209 failure. This is how the failure looked in a general terms:
- block B from layer 100 used base block A from layer 50
- block B is syntactically valid and included into the layer 100 input vector.
- tortoise for layer 100 **fails to process block B because block A is already evicted**
- block C from layer 101 adds support for all blocks in layer 100, including block B, since it is included in the input vector. this doesn't contradict the good block rule yet.
- **opinion for Block B is added from support exceptions**
- block D from layer 102 selects block C as a base block, still good, not contradicting good block rule.
- handling block D, **when opinions from block C are copied we filtered out opinion on block B since it is unknown to tortoise**
- block E from 103 uses block D as base block, it now had to add opinion on block B since it is not in block D opinions list.
- at this point, we are entering self-healing mode, and after SlidingWindow+self-healing lag tortoise won't have any good blocks. 

Also see a repro in https://github.com/spacemeshos/go-spacemesh/pull/2935/commits/884219fef7bfb14d12e00b0b9b84d56814fcd4f0 with a workaround for this issue.

## Actual Behavior
tortoise forced into self-healing when adding base block outside of the sliding window.

## Expected Behavior
tortoise handles block that uses old base block gracefully, potentially by not marking it as good

dshulyak:
## Motivation

If the base block was evicted we will not update the internal tortoise state and lose votes from the new block.

With this change, we always preserve the block and relevant exceptions, but since the base block's votes are not in the memory we will not copy them, and this new block will not be marked good. 

closes https://github.com/spacemeshos/go-spacemesh/issues/2936

## Changes
- preserve votes from the block that has a base block outside of the sliding window

## Test Plan
TestBaseBlockEvictedBlock

## TODO

## DevOps Notes


dshulyak:
bors try

dshulyak:
bors try

tortoise/sim/votes.go
```diff
@@ -13,11 +13,12 @@ type Voting struct {
 }
 
 // VotesGenerator allows to replace default votes generator.
-type VotesGenerator func(rng *rand.Rand, layers []*types.Layer) Voting
+// TODO(dshulyak) what is the best way to encapsulate all configuration that is required to generate votes?
+type VotesGenerator func(rng *rand.Rand, layers []*types.Layer, i int) Voting
 
-// perfectVoting select base block from previouos layer and supports all blocks from previous layer.
+// PerfectVoting select base block from previouos layer and supports all blocks from previous layer.
```
nkryuchkov: nit: 
```suggestion
// PerfectVoting selects base block from previous layer and supports all blocks from previous layer.
```

nkryuchkov (approved):

dshulyak:
bors merge

PR #2939, opened 2021-11-15
[Merged by Bors] - sync: allow sync to resume when validation takes too long

### LINKED ISSUE #2934: blocks, sync: do not publish blocks until the the most recent layer is verified
## Description

After sync fetches layers from peers it submits them for verification to a separate goroutine. Verification always lags behind sync, especially on slow machines. So at the time when we set synced state, we still might be lagging behind arbitrarily. 

The synced state enables blocks builder to publish blocks. So if our current verified layer is behind cluster verified layers by more than sliding window difference - all blocks that will be produced by use base block that is not valid from the rest of the network point of view.

It doesn't mean that we shouldn't handle such blocks gracefully, but we definitely shouldn't increase the likelihood of creating such blocks. 

## Steps to reproduce
1. Have a network with at least 2000-3000 layers
2. Launch sync
3. Compare by how many layers verified is lagging behind synced

## Actual Behavior
Block from new joiner is very likely to use base block from evicted layer.

## Expected Behavior
Block from new joiner is not likely to use base block from evicted layer.

countvonzero:
## Motivation

Closes #2934

## Changes

allow for the code to resume the sync loop if the current layer has advanced after validation thread exited

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -50,6 +50,9 @@ const (
 
 	// the max amount of layer delays syncer can tolerate before it tries to validate a layer.
 	maxHareDelayLayers uint32 = 10
+
+	// the max number of attempts syncer tries to advance the processed layer within a sync run.
+	maxAttemptWithinRun = 3
 )
 
 type syncState uint32
@@ -307,28 +310,25 @@ func (s *Syncer) synchronize(ctx context.Context) bool {
 	var (
 		vQueue chan *types.Layer
 		vDone  chan struct{}
-
-		target = s.mesh.ProcessedLayer()
 		// number of attempts to start a validation goroutine
-		attempt = 1
+		attempt = 0
 		// whether the data sync succeed. validation failure is not checked by design.
 		success = true
 	)
-	attemptFunc := func() (bool, types.LayerID) {
+
+	attemptFunc := func() bool {
 		var (
-			layer       *types.Layer
-			syncedLayer types.LayerID
-			err         error
+			layer *types.Layer
+			err   error
 		)
 		// using ProcessedLayer() instead of LatestLayer() so we can validate layers on a best-efforts basis
 		// and retry in the next sync run if validation fails.
-		// our clock starts ticking from 1 so it is safe to skip layer 0
+		// our clock starts ticking from 1, so it is safe to skip layer 0
 		// always sync to currentLayer-1 to reduce race with gossip and hare/tortoise
 		for layerID := s.mesh.ProcessedLayer().Add(1); layerID.Before(s.ticker.GetCurrentLayer()); layerID = layerID.Add(1) {
 			if layer, err = s.syncLayer(ctx, layerID); err != nil {
-				return false, layerID
+				return false
 			}
-			syncedLayer = layerID
 			vQueue <- layer
 			logger.With().Debug("finished data sync", layerID)
 		}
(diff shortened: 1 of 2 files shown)


syncer/syncer.go
```diff
@@ -297,49 +300,78 @@ func (s *Syncer) synchronize(ctx context.Context) bool {
...
-			log.FieldNamed("processed", s.mesh.ProcessedLayer()))
-		s.setSyncerIdle()
-	}()
 
-	// using ProcessedLayer() instead of LatestLayer() so we can validate layers on a best-efforts basis.
+	var (
+		vQueue chan *types.Layer
+		vDone  chan struct{}
+		// number of attempts to start a validation goroutine
+		attempt = uint64(0)
```
dshulyak: can be regular int

countvonzero:
bors try

syncer/syncer.go
```diff
@@ -297,49 +300,78 @@ func (s *Syncer) synchronize(ctx context.Context) bool {
...
+		// whether the data sync succeed. validation failure is not checked by design.
+		success = true
+	)
+
+	// using ProcessedLayer() instead of LatestLayer() so we can validate layers on a best-efforts basis
+	// and retry in the next sync run if validation fails.
 	// our clock starts ticking from 1 so it is safe to skip layer 0
 	// always sync to currentLayer-1 to reduce race with gossip and hare/tortoise
+syncLoop:
```
dshulyak: i would rearrange this a bit to decouple attempts and queue control from the validation loop, and also remove some duplication

```go
attemptFunc := func() (bool, types.LayerID) {
		var (
			layerID types.LayerID
			layer   *types.Layer
			err     error
		)
		for layerID = s.mesh.ProcessedLayer().Add(1); layerID.Before(s.ticker.GetCurrentLayer()); layerID = layerID.Add(1) {
			if layer, err = s.syncLayer(ctx, layerID); err != nil {
				logger.With().Error("failed to sync to layer", layerID, log.Err(err))
				return false, layerID
			}
			vQueue <- layer
			logger.With().Debug("finished data sync", layerID)
		}
		logger.With().Debug("data is synced, waiting for validation",
			log.Uint64("attempt", attempt),
			log.FieldNamed("current", s.ticker.GetCurrentLayer()),
			log.FieldNamed("latest", s.mesh.LatestLayer()),
			log.FieldNamed("processed", s.mesh.ProcessedLayer()))
		return true, layerID
	}
	// check if we are still on target. if not, go back to the sync loop
	for ; success && target != s.ticker.GetCurrentLayer().Sub(1); attempt++ {
		vQueue, vDone = s.startValidating(ctx, s.run, attempt)
		success, target = attemptFunc()
		close(vQueue)
		select {
		case <-vDone:
		case <-s.shutdownCtx.Done():
			return false
		}
	}
```

dshulyak (approved):

countvonzero:
bors try

syncer/syncer.go
countvonzero: done

syncer/syncer.go
countvonzero: done

countvonzero: thanks for the suggestion. much better now

countvonzero:
bors try

countvonzero:
bors merge

PR #2943, opened 2021-11-16
[Merged by Bors] - tortoise: make tortoise layers pipeline easier to read

dshulyak:
## Motivation

tortoise code had excessive nested levels, for example for self-healing they didn't made any sense at all. this change makes it easier to grok all conditions that are applied in the main pipeline of the tortoise. 

## Changes
- reduce nesting level and split up some methods

## Test Plan
ut

## TODO

## DevOps Notes

### DIFF
--- a/tortoise/verifying_tortoise.go
+++ b/tortoise/verifying_tortoise.go
@@ -183,12 +183,12 @@ func (t *turtle) evict(ctx context.Context) {
 		return
 	}
 	logger.With().Info("attempting eviction",
-		log.FieldNamed("effective_genesis", types.GetEffectiveGenesis()),
+		log.Named("effective_genesis", types.GetEffectiveGenesis()),
 		log.Uint32("hdist", t.Hdist),
-		log.FieldNamed("verified", t.Verified),
+		log.Named("verified", t.Verified),
 		log.Uint32("window_size", t.WindowSize),
-		log.FieldNamed("last_evicted", t.LastEvicted),
-		log.FieldNamed("window_start", windowStart))
+		log.Named("last_evicted", t.LastEvicted),
+		log.Named("window_start", windowStart))
 
 	// evict from last evicted to the beginning of our window
 	if err := t.state.Evict(ctx, windowStart); err != nil {
@@ -225,7 +225,7 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
 			exceptionBlock, err := t.bdp.GetBlock(exceptionBlockID)
 			if err != nil {
 				logger.With().Error("inconsistent state: can't find block from diff list",
-					log.FieldNamed("exception_block_id", exceptionBlockID),
+					log.Named("exception_block_id", exceptionBlockID),
 					log.Err(err),
 				)
 				return false
@@ -235,27 +235,27 @@ func (t *turtle) checkBlockAndGetLocalOpinion(
 
 		if lid.Before(baseBlockLayer) {
 			logger.With().Error("good block candidate contains exception for block older than its base block",
-				log.FieldNamed("older_block", exceptionBlockID),
-				log.FieldNamed("older_layer", lid),
-				log.FieldNamed("base_block_layer", baseBlockLayer))
+				log.Named("older_block", exceptionBlockID),
+				log.Named("older_layer", lid),
+				log.Named("base_block_layer", baseBlockLayer))
 			return false
 		}
 
 		v, err := t.getLocalBlockOpinion(ctx, lid, exceptionBlockID)
 		if err != nil {
 			logger.With().Error("unable to get single block opinion for block in exception list",
-				log.FieldNamed("older_block", exceptionBlockID),
-				log.FieldNamed("older_layer", lid),
-				log.FieldNamed("base_block_layer", baseBlockLayer),
+				log.Named("older_block", exceptionBlockID),
+				log.Named("older_layer", lid),
+				log.Named("base_block_layer", baseBlockLayer),
 				log.Err(err))
 			return false
 		}
 
 		if v != voteVector {
 			logger.With().Debug("not adding block to good blocks because its vote differs from local opinion",
-				log.FieldNamed("older_block", exceptionBlockID),
-				log.FieldNamed("older_layer", lid),
-				log.FieldNamed("local_opinion", v),
+				log.Named("older_block", exceptionBlockID),
+				log.Named("older_layer", lid),
+				log.Named("local_opinion", v),
 				log.String("block_exception_vote", className))
 			return false
 		}
@@ -396,7 +396,7 @@ func (t *turtle) calculateExceptions(
 	baseBlockLayerID types.LayerID,
 	baseBlockOpinion Opinion, // candidate base block's opinion vector
 ) ([]map[types.BlockID]struct{}, error) {
-	logger := t.logger.WithContext(ctx).WithFields(log.FieldNamed("base_block_layer_id", baseBlockLayerID))
+	logger := t.logger.WithContext(ctx).WithFields(log.Named("base_block_layer_id", baseBlockLayerID))
 
 	// using maps prevents duplicates
 	againstDiff := make(map[types.BlockID]struct{})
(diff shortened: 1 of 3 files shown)


tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
+		// 1. record our opinion on this block and go on evaluating the rest of the blocks in this layer to see if
+		//    we can verify the layer (if local and global consensus match, and global consensus is decided)
+		// 2. keep waiting to verify the layer (if not, and the layer is relatively recent)
+		// 3. trigger self healing (if not, and the layer is sufficiently old)
+		consensusMatches := globalOpinionOnBlock == localOpinionOnBlock
+		globalOpinionDecided := globalOpinionOnBlock != abstain
+
+		if consensusMatches && globalOpinionDecided {
+			// Opinion on this block is decided, save and keep going
+			contextualValidity[bid] = globalOpinionOnBlock == support
```
nkryuchkov: It looks to me that we can move these two lines to the end of the function before `return`, which would let us remove the `if` statement above. Is that correct?

tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
+	// all positions up to k: in other words, we can verify a layer k if the total weight of the global opinion
+	// exceeds the confidence threshold, and agrees with local opinion.
+	for bid, localOpinionOnBlock := range localOpinion {
+		// count the votes of the input vote vector by summing the voting weight of good blocks
+		logger.With().Debug("summing votes for candidate layer block",
+			bid,
+			log.FieldNamed("layer_start", lid.Add(1)),
+			log.FieldNamed("layer_end", t.Last),
+		)
+		sum, err := t.sumVotesForBlock(ctx, bid, lid.Add(1), func(votingBlockID types.BlockID) bool {
```
nkryuchkov: I think extracting the anonymous function would make the function a bit more readable

nkryuchkov (approved):

tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
 	}
 	if len(layerBlocks) == 0 {
-		// nothing to do
 		t.logger.WithContext(ctx).Warning("cannot process empty layer block list")
 		return nil
 	}
 	return t.processBlocks(ctx, layerBlocks)
 }
 
+func (t *turtle) verifyingTortoise(ctx *tcontext, logger log.Log, lid types.LayerID) (map[types.BlockID]bool, error) {
```
countvonzero, line 818: this function name doesn't say much. can you either change the name or comment on what it does?

tortoise/verifying_tortoise.go
countvonzero: i second this.

tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
+
+	// Count the votes of good blocks. localOpinionOnBlock is our local opinion on this block.
+	// Declare the vote vector "verified" up to position k if the total weight exceeds the confidence threshold in
+	// all positions up to k: in other words, we can verify a layer k if the total weight of the global opinion
+	// exceeds the confidence threshold, and agrees with local opinion.
+	for bid, localOpinionOnBlock := range localOpinion {
+		// count the votes of the input vote vector by summing the voting weight of good blocks
+		logger.With().Debug("summing votes for candidate layer block",
+			bid,
+			log.FieldNamed("layer_start", lid.Add(1)),
```
countvonzero: maybe use the shorthand you created `Named` through out?

tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
+				bid,
+				log.FieldNamed("global_opinion", globalOpinionOnBlock),
+				log.FieldNamed("local_opinion", localOpinionOnBlock))
+			return nil, nil
+		}
+
+		// There are only two scenarios that could result in a global opinion of abstain: if everyone is still
+		// waiting for Hare to finish for a layer (i.e., it has not yet succeeded or failed), or a balancing attack.
+		// The former is temporary and will go away after `zdist' layers. And it should be true of an entire layer,
+		// not just of a single block. The latter could cause the global opinion of a single block to permanently
```
countvonzero, line 891: not that you wrote this. but i don't understand why should we consider all blocks in a layer as a whole. if there is only one block in the layer that both global and local opinions is abstain, and for a sufficiently long period of time, why can't we simply disregard that block and accept only TXs in other blocks?

tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
+				bid,
+				log.FieldNamed("global_opinion", globalOpinionOnBlock),
+				log.FieldNamed("local_opinion", localOpinionOnBlock))
+			return nil, nil
+		}
+	}
+	return contextualValidity, nil
+}
+
+func (t *turtle) healingTortoise(ctx *tcontext, logger log.Log, lid types.LayerID) (bool, error) {
```
countvonzero, line 910: again, using a name as a function name isn't very self-documenting

tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
+		// give up trying to verify layers and keep waiting
+		logger.With().Info("failed to verify candidate layer, will reattempt later")
+		return false, nil
+	}
+
+	// reinitialize context because contextually valid blocks were changed
+	ctx = wrapContext(ctx)
+	// rescore goodness of blocks in all intervening layers on the basis of new information
+	for layerID := lastVerified.Add(1); !layerID.After(t.Last); layerID = layerID.Add(1) {
+		if err := t.scoreBlocksByLayerID(ctx, layerID); err != nil {
```
countvonzero, line 946: i think this operation is reversed.
if we really want to heal, we should reload the blocks from disk (including blocks with different beacons and late blocks, i.e. blocks that didn't make it to hare output) and process them before calling `heal()`

tortoise/verifying_tortoise.go
```diff
@@ -809,27 +809,164 @@ func (t *turtle) handleLayerBlocks(ctx *tcontext, layerID types.LayerID) error {
...
+	}
+	if len(localOpinion) == 0 {
+		// warn about this to be safe, but we do allow empty layers and must be able to verify them
+		logger.With().Warning("empty vote vector for layer", lid)
+	}
+
+	contextualValidity := make(map[types.BlockID]bool, len(localOpinion))
+
+	// Count the votes of good blocks. localOpinionOnBlock is our local opinion on this block.
+	// Declare the vote vector "verified" up to position k if the total weight exceeds the confidence threshold in
```
countvonzero, line 840: the comment about vote vector seems out of place and should be in `verifyLayers` instead

countvonzero (approved):

tortoise/verifying_tortoise.go
dshulyak: for me it is self-documenting, as it says that this function performs verifying tortoise, unlike other function that performs full tortoise

tortoise/verifying_tortoise.go
dshulyak: this comment is about comparing local opinion with global and it is done only in verifying tortoise

tortoise/verifying_tortoise.go
dshulyak: i just assumed that we must reach by design consensus on a block before verifying layer, why exactly i don't know

tortoise/verifying_tortoise.go
dshulyak: but this is self-explanatory, we have verifying tortoise and healing (full) tortoise. people who want to understand this code will need to learn about what each of them is doing

tortoise/verifying_tortoise.go
dshulyak: no, we rescore after healing because local opinion changed and goodness may have changed as well. doing this before will not change goodness. 

as for loading blocks i think it was already discussed, we can't do it just any time when healing is called so we will need to have external signal (state sync) or just do it periodically.

dshulyak:
bors merge

Issue #2921, opened 2021-11-08
tortoise should use current layer as its `Last` layer


countvonzero:
## Description

tortoise now uses the last layer output by hare as its `Last` layer. this is different from the rest of the system that consider the current layer the latest layer. and the following two issues make this notion obsolete.

https://github.com/spacemeshos/go-spacemesh/issues/2919
https://github.com/spacemeshos/go-spacemesh/issues/2920

tortoise should simply uses current layer as its last layer for simplicity and correctness.

dshulyak:
Just adding a note that we shouldn't use layer clock in turtle structure, otherwise, last layer might be changed while we are running same method, also there are several computations that are based on the last layer changes (e.g. update thresholds). Probably the easiest would to spawn a goroutine that will listen to layer clock, once trigger it will obtain tortoise mutex and change the last layer.

dshulyak:
Also realized recently that this issue is very important for voting properly on undecided layers (e.g. within zdist). Currently, we will vote against layers within zdist that weren't submitted yet, which is a mistake obviously.

dshulyak:
> Also realized recently that this issue is very important for voting properly on undecided layers (e.g. within zdist). Currently, we will vote against layers within zdist that weren't submitted yet, which is a mistake obviously.

this was solved by tallying votes before voting.

dshulyak:
this should not be implemented. caller always have to tally votes, before casting votes or fetching results from the tortoise. tally votes accepts last layer as a parameter

countvonzero:
@dshulyak how do you propose to resolve the 2nd part of https://github.com/spacemeshos/go-spacemesh/issues/3003

if tortoise always use the "last layer" given to it, not the actual current layer, that means we can never remove data that's only needed within hdist layers (certificate).
syncer will always have to sync a certificate from genesis.

dshulyak:
downloading certificate is an orthogonal problem. certificate in tortoise is needed for a valid opinion vector for a verifying tortoise. if instead of certificate it can just assume that peer gave an honest opinion - that can be used in place of certificate.

alternatively full tortoise can be used for syncing nodes, and then certificate is irrelevant as well. there is a small problem with that, but right now it is better to use full tortoise for new nodes instead of downloading every certificate.

PR #2944, opened 2021-11-16
[Merged by Bors] - tortoise: use weak coin for voting on undecided blocks

### LINKED ISSUE #2688: Score individual blocks separately using weak coin in verifying tortoise
According to @tal-m:

> We vote for each block separately.
> Example: blocks A, B and C are all in layer 100, blocks D and E are in layer 200. We're creating a block V in layer 300
> Block A has a vote margin of 0, block B a vote margin > \theta_L, block C a vote margin < -\theta_L and blocks D and E both have vote margin 0.
> Then block V will vote according to the layer-300 weak coin  (i.e., the one that's generated after layer 299 votes are "final", but before layer 300 blocks are created) regarding blocks A, D and E, and will vote  for block B and against C

Right now, tortoise scores an entire layer according to the weak coin when it finds a single block in that layer that does not pass the threshold.

Related: #2678 

## Affected code
tortoise/verifying_tortoise.go:`layerOpinionVector`

This issue appears in commit hash: d08b55ce28dfd59bd04639c7b3db2d4397d709f9

dshulyak:
## Motivation

Vote on undecided blocks from layers that are not in range of [last-hdist, last] according to weak coin from last completed hare.

Closes #2688

## Changes
- refactored computeLocalOpinion to work without weak  coin in general case, also refactored tests
- refactored calculateExceptions to vote on each undecided block using weakcoin. currently we are using weakcoin from the layer where hare has completed, it is generally correct unless hare takes longer then one layer to complete. 

## Test Plan
TestWeakCoinVoting and TestHealBalanceAttack

## TODO

## DevOps Notes


tortoise/verifying_tortoise.go
```diff
@@ -419,75 +398,65 @@ func (t *turtle) calculateExceptions(
 	if startLayer.Before(types.GetEffectiveGenesis()) {
 		startLayer = types.GetEffectiveGenesis()
 	}
-	for layerID := startLayer; !layerID.After(t.Last); layerID = layerID.Add(1) {
-		logger := logger.WithFields(log.Named("diff_layer_id", layerID))
+	for lid := startLayer; !lid.After(t.Last); lid = lid.Add(1) {
```
nkryuchkov, line 401: Why do you want to rename `layerID` to `lid`?

nkryuchkov (approved):

tortoise/verifying_tortoise.go
```diff
@@ -419,75 +398,65 @@ func (t *turtle) calculateExceptions(
...
-		// otherwise, nil means we should abstain
-		if layerInputVector == nil {
-			// still waiting for Hare results, vote neutral and move on
-			logger.With().Debug("input vector is empty, adding neutral diffs", log.Err(err))
-			for _, b := range layerBlockIds {
-				addDiffs(b, "neutral", abstain, neutralDiff)
+		for bid, opinion := range local {
+			usecoinflip := opinion == abstain && lid.Before(t.layerCutoff())
+			if usecoinflip {
+				coin, exist := t.bdp.GetCoinflip(ctx, votinglid)
```
countvonzero, line 427: are you getting votinglid.Sub(1)? the log below says so

tortoise/verifying_tortoise.go
```diff
@@ -419,75 +398,65 @@ func (t *turtle) calculateExceptions(
...
+				if !exist {
+					return nil, fmt.Errorf("coinflip is not recorded in %s", votinglid.Sub(1))
```
countvonzero: this would happen when the node didn't participate in the hare process for that votinglib.Sub(1) layer.
i think not-participating in hare is common. so it come downs to how often do we need to use coin flip. i think we should at least open an issue for the missing coin flip. do you think we can record it in the hare certificate? or did we already?

tortoise/verifying_tortoise.go
```diff
@@ -1009,103 +966,78 @@ func (t *turtle) layerCutoff() types.LayerID {
...
-	}
+	// and see if they pass the local threshold.
 	logger.With().Debug("counting votes for and against blocks in old, unverified layer",
-		log.Int("num_blocks", len(layerBids)))
-	supported := make([]types.BlockID, 0, len(layerBids))
-	for _, bid := range layerBids {
-		logger := logger.WithFields(log.Named("candidate_block_id", bid))
+		log.Int("num_blocks", len(bids)))
+	for _, bid := range bids {
+		logger := logger.WithFields(log.FieldNamed("candidate_block_id", bid))
```
countvonzero: Named

tortoise/verifying_tortoise.go
dshulyak: no, sorry the log should be adjusted. in current implementation the value should be last layer that tortoise received

tortoise/verifying_tortoise.go
dshulyak: > i think not-participating in hare is common. so it come downs to how often do we need to use coin flip. 

there is some misunderstanding. i think that participating in hare voting is not a prerequisite for receiving a coin, i will doublecheck.

tortoise/verifying_tortoise.go
dshulyak: i tend to use bid and lid consistently, and was going to change everything to this format at least in tortoise. they are both short and very clear (in my opinion). if there is something specific like base block id - it is basebid, or base block layer - baselid.

tortoise/verifying_tortoise.go
countvonzero: what's your consideration for not changing it to Sub(1) here?

dshulyak: for creating a block we need to use weakcoin from last completed hare `prepare` round. we update Last in tortoise when hare sends a notification, therefore votinglid should be t.Last and not t.Last - 1. 

this doesn't change what is written in the issue that you created, if we will use layer clock in tortoise then we need to use t.Last - 1 here.

countvonzero: > therefore votinglid should be t.Last and not t.Last - 1

my confusion is, we should be generating block for the current ticked layer (say `N`). 
so in the happy flow, votinglid should be `t.Last+1` on line 280 and here we use `votinglid.Sub(1)`. i think that's more consistent with the design discussion.

but in the unhappy flow, `t.Last` is the most recent layer hare terminated, but not necessarily `N-1`. `t.Last` can be layer `N-2` while `N-1` hasn't terminated. in this case, i think it's prolly good enough to use coinflip in `t.Last` (`N-2`) as the situation is the same for all nodes.

otherwise we need to change hare to record coinflip when hare started instead of terminated.

countvonzero (approved):

tortoise/verifying_tortoise.go
dshulyak: yes, this is my understanding as well. right now we use latest available that in happy case is CURRENT - 1, sometimes CURRENT - 2. while working on https://github.com/spacemeshos/go-spacemesh/issues/2921 weakcoin needs to be saved right after preround and then we can always use CURRENT - 1 based on layer clock.

dshulyak:
bors merge

PR #2945, opened 2021-11-16
[Merged by Bors] - api: include layer ID in `AccountMeshDataStream`

nkryuchkov:
## Motivation

This PR includes layer ID in transaction streamed by `AccountMeshDataStream`

## Changes

- Add layer ID to `AccountMeshDataStream`

## Test Plan

UT

## DevOps Notes

### DIFF
--- a/api/grpcserver/grpcserver_test.go
+++ b/api/grpcserver/grpcserver_test.go
@@ -494,7 +494,7 @@ func (m *MempoolMock) Put(id types.TransactionID, tx *types.Transaction) {
 	m.poolByTxid[id] = tx
 	m.poolByAddress[tx.Recipient] = id
 	m.poolByAddress[tx.Origin()] = id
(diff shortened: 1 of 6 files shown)


dshulyak (approved):
as i remember it was a temporary solution until receipts are implemented, not sure where we are with receipts at this point

nkryuchkov:
@dshulyak As far as I know, with receipts, we are at planning to create a SMIP for them. I'm not sure what you meant by mentioning receipts. Did you mean that we should pass a receipt instead of layer ID, and then extract layer ID, so this PR is a temporary solution?

nkryuchkov:
bors merge

dshulyak:
i meant that once receipts are implemented we won't need to have this workaround (and one that i added before for queries)

nkryuchkov:
@dshulyak I see, thanks

PR #2948, opened 2021-11-16
[Merged by Bors] - events: use pub-sub event delivery

### LINKED ISSUE #2384: Streaming apis implementation can miss sending data to clients

nkryuchkov:
## Motivation

Closes #2384
Supersedes #2389

## Changes

- Use pub-sub event delivery in reporter

## Test Plan

Unit and system tests

## DevOps Notes


nkryuchkov:
bors try

dshulyak (changes requested):
the implementation needs to handle the case when the receiver is too slow or disconnected but the stream remains open. we can't allow publishers to block consensus code.

i suggest when stream is opened create another channel with sufficiently large buffer, spawn goroutine that will consume from eventbus and send events to that new channel, if the buffer of this first channel is full - stream must be closed with an error.

nkryuchkov:
@dshulyak Does https://github.com/spacemeshos/go-spacemesh/pull/2948/commits/fc0e92eca06d08e8e3cd3de16615d9540aa347c2 contain what you meant?

nkryuchkov:
bors try

api/grpcserver/events.go
```diff
@@ -0,0 +1,17 @@
+package grpcserver
+
+const subscriptionChanBufSize = 1 << 16
+
+func consumeEvents(inputCh <-chan interface{}) <-chan interface{} {
+	outputCh := make(chan interface{}, subscriptionChanBufSize)
+
+	go func(inputCh <-chan interface{}) {
+		for event := range inputCh {
+			outputCh <- event
```
dshulyak: exit the loop if send is blocking, otherwise it doesn't change anything. eventually when buffer is full publisher will block

dshulyak (commented):
you can test this btw, create a stream but don't read from it. eventually, network buffer, subscription buffer and new channel buffer will fill up and publishing must block. it may need some experimentation to find when it will block, but once it does just check that stream exits.

api/grpcserver/globalstate_service.go
```diff
@@ -271,23 +271,45 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
-		channelReceipt chan events.TxReceipt
+		channelAccount <-chan interface{}
+		channelReward  <-chan interface{}
+		channelReceipt <-chan interface{}
 	)
 	if filterAccount {
-		channelAccount = events.GetAccountChannel()
+		if accountSubscription := events.SubscribeAccount(); accountSubscription != nil {
+			defer func() {
+				if err := accountSubscription.Close(); err != nil {
```
dshulyak: this a bit repetitive, smth like `defer closeSub(sub)` will be better

api/grpcserver/events.go
nkryuchkov: @dshulyak If send is blocking, should we drop the peer after exiting the loop? If what you suggested is implemented, a peer wouldn't receive new events anymore after exiting the loop

dshulyak: yes the stream must be terminated, preferably with an error that will explain that the subscriber was too slow (let's not use peer outside of the p2p context)

api/grpcserver/globalstate_service.go
```diff
@@ -271,23 +271,45 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
 	var (
-		channelAccount chan types.Address
-		channelReward  chan events.Reward
-		channelReceipt chan events.TxReceipt
+		accountCh       <-chan interface{}
+		rewardsCh       <-chan interface{}
+		receiptsCh      <-chan interface{}
+		accountBufFull  <-chan struct{}
+		rewardsBufFull  <-chan struct{}
+		receiptsBufFull <-chan struct{}
```
dshulyak: why not to reuse same channel and return error on `!ok` ? if stream is terminated by the caller it will exit on `stream.Context().Done()`

api/grpcserver/globalstate_service.go
```diff
@@ -271,23 +271,45 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
-		case updatedAccount, ok := <-channelAccount:
+		case <-accountBufFull:
+			log.Info("account buffer is full, shutting down")
+			return fmt.Errorf("account buffer is full")
+		case <-rewardsBufFull:
+			log.Info("rewards buffer is full, shutting down")
+			return fmt.Errorf("rewards buffer is full")
+		case <-receiptsBufFull:
+			log.Info("receipts buffer is full, shutting down")
+			return fmt.Errorf("receipts buffer is full")
```
dshulyak: errors should probably use status.Errorf(status.Canceled, "stream buffer is full")

dshulyak (approved):
i think the second channel is redundant. also tests for new behavior are missing:
- for multiple subscribers to the same events (previously only one of them will receive events)
- for subscriber that stopped listening for events

api/grpcserver/globalstate_service.go
nkryuchkov: I started implementing this, but then thought that if `(event.Bus).Subscribe` closes the channel (although the only case should be closing the subscription which means that the stream handler is finished), theoretically, some data may be lost, so the second channel is safer. But I think we can get rid of it as you said

api/grpcserver/globalstate_service.go
nkryuchkov: good idea

nkryuchkov:
@dshulyak the tests are being implemented

api/grpcserver/globalstate_service.go
dshulyak: but why would it close the channel? i don't think that there is any logic to do that

nkryuchkov: it shouldn't, I just thought it's safer not to rely on it

nkryuchkov: but I agree that we can assume that it wouldn't close the channel and get rid of the second channel

nkryuchkov:
bors merge

Issue #2743, opened 2021-08-30
flaky test: app test TestShutdown unexpected number of goroutines


countvonzero:
## Description

https://github.com/spacemeshos/go-spacemesh/runs/3443623393
```
app_test.go:775: 
        	Error Trace:	app_test.go:775
        	Error:      	Not equal: 
        	            	expected: 4
        	            	actual  : 5
        	Test:       	TestShutdown
```

countvonzero:
this turned out to be quite difficult to reproduce

$ go test -timeout 0 -v ./cmd/node -run TestShutdown -count 100

3 times now and still could not reproduce it

nkryuchkov:
https://github.com/spacemeshos/go-spacemesh/runs/3815964636

nkryuchkov:
https://github.com/spacemeshos/go-spacemesh/runs/4275261405

moshababo:
Obsolete.

PR #2949, opened 2021-11-16
[Merged by Bors] - api: hide `layerId` field when it isn't passed to `AccountMeshDataStream`

nkryuchkov:
## Motivation

#2945 added layer ID to `AccountMeshDataStream`, however, if `layerId` isn't passed, `AccountMeshDataStream` returns an empty `layerId` field now:
```
{
  "datum": {
    "meshTransaction": {
      "transaction": {
        "id": {
          "id": "nkYYCMJv7vCrxY+9AlbUaMe7OnKBets6zOeBE0Nr/do="
        },
        "coinTransfer": {
          "receiver": {
            "address": "AAAAAAAAAAAAAAAAAAAAAAAAAAI="
          }
        },
        "sender": {
          "address": "/wg8miLgXTRZsD89vtYcetbgogk="
        },
        "gasOffered": {
          "gasProvided": "100"
        },
        "amount": {
          "value": "3"
        },
        "counter": "39",
        "signature": {
          "scheme": "SCHEME_ED25519_PLUS_PLUS",
          "signature": "dHeM70gCBfuUd3+SZptN7W/eMKLkWCcdpkbBAkxkT5VK60wUynHxhWbECWEL2DlvlISLuwlxvDXMg3FHI/5NCg==",
          "publicKey": "/wg8miLgXTRZsD89vtYcetbgogk="
        }
      },
      "layerId": {
        
      }
    }
  }
}
``` 

The PR disables showing `layerId` field if it isn't available.

## Changes

- hide layer ID in `AccountMeshDataStream` if it isn't passed

## Test Plan

UT

## DevOps Notes


nkryuchkov:
bors try

nkryuchkov:
bors try

blocks/blocks.go
```diff
@@ -199,7 +199,7 @@ func (bh BlockHandler) blockSyntacticValidation(ctx context.Context, block *type
 
 	// get the TXs
 	if len(block.TxIDs) > 0 {
-		err := bh.fetcher.GetTxs(ctx, block.TxIDs)
+		err := bh.fetcher.GetTxs(ctx, block.TxIDs, block.LayerIndex)
```
dshulyak: this doesn't make sense

mempool/tx_pool_test.go
```diff
@@ -32,23 +32,23 @@ func TestNewTxPoolWithAccounts(t *testing.T) {
 	r.Equal(prevBalance, balance)
 
 	tx1 := newTx(t, 4, 50, signer)
-	pool.Put(tx1.ID(), tx1)
+	pool.Put(tx1.ID(), tx1, types.LayerID{})
```
dshulyak: this also doesn't make sense

dshulyak (changes requested):
the proposed solution is too hackish, i don't think that we can use it as is. 

is it viable to change AccountMeshDataStream to return tx only when it is committed as a part of the block?

mempool/tx_pool.go
```diff
@@ -108,9 +108,9 @@ func (t *TxMempool) Put(id types.TransactionID, tx *types.Transaction) {
 }
 
 // Invalidate removes transaction from pool.
-func (t *TxMempool) Invalidate(id types.TransactionID) {
+func (t *TxMempool) Invalidate(txID types.TransactionID, layerID types.LayerID) {
```
dshulyak: what layer even means in the context of the txpool.Invalidate ?

mesh/mesh.go
```diff
@@ -55,16 +55,16 @@ type state interface {
...
 	GetNonce(addr types.Address) uint64
 	GetAllAccounts() (*types.MultipleAccountsState, error)
 }
 
 type txMemPool interface {
-	Invalidate(id types.TransactionID)
+	Invalidate(txID types.TransactionID, layerID types.LayerID)
 	Get(id types.TransactionID) (*types.Transaction, error)
-	Put(id types.TransactionID, tx *types.Transaction)
+	Put(id types.TransactionID, tx *types.Transaction, layerID types.LayerID)
```
dshulyak: the same for put

dshulyak (changes requested):
we should use layer in that event only when transaction is applied to the state (e.g. the way receipt would work, if we had it implemented). in other cases either ignore layer, or maybe don't consume those events in AccountMeshDataStream, but that should be considered case by case.

countvonzero:
i am not very familiar with the API. but it doesn't make sense to attach an layer ID to a TX when it's still in the pending state.

> however, in some cases, it's either not passed properly or is unavailable, so in these cases AccountMeshDataStream returns an empty layerId field now

in which case is it not passed properly? if there is, that should be fixed. but should be part of the event reporting flow when the TX is applied to the state.

when it's not available, that means it's still pending. i don't see missing a layerID in AccountMeshDataStream as a problem at all as long as the document is clear. tho, i'd prefer it to return layerID == 0 explicitly so the client knows that it's not applied to the mesh state yet.

nkryuchkov:
bors try

nkryuchkov:
bors try

nkryuchkov:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/4275294876?check_suite_focus=true)

https://github.com/spacemeshos/go-spacemesh/issues/2743#issuecomment-974781932

moshababo:
@nkryuchkov following https://github.com/spacemeshos/go-spacemesh/pull/2949#issuecomment-972289990, can you elaborate on the case(s) where it's not passed?

nkryuchkov:
@countvonzero @dshulyak 
Thanks. I reviewed the PR and agree that layer ID should be passed not in all cases. This has been fixed in the latest commits. Could you please review the PR again?

@countvonzero
> in which case is it not passed properly?

I meant the cases when a tx is still not applied to the state — in those cases the layer ID was not passed, but I agree that we don't need to pass layer ID then.

> tho, i'd prefer it to return layerID == 0 explicitly so the client knows that it's not applied to the mesh state yet.

Why do you think that lack of layerID doesn't look better than `layerID == 0`? I'd prefer to just omit it in this case.

@moshababo
Sure, thanks for reminding

blocks/blocks.go
nkryuchkov: reverted

mempool/tx_pool_test.go
nkryuchkov: reverted

mempool/tx_pool.go
nkryuchkov: actually, it doesn't make sense to add it here, I reverted it

mesh/mesh.go
nkryuchkov: reverted

countvonzero:
> > in which case is it not passed properly?
> 
> I meant the cases when a tx is still not applied to the state — in those cases the layer ID was not passed, but I agree that we don't need to pass layer ID then.

this PR currently is just fixing data presentation. but the goal of https://github.com/spacemeshos/go-spacemesh/pull/2945 is to provide layer data in account stream. 

in terms of TX life cycle below, i think we may need to push more events to make the account stream more complete.
TX life cycle:
- enters mempool: `ReportNewTx` is called with empty layer ID, in pending state
- packed in a block received in mesh: `AddBlockWithTxs` is called but no TX related event were triggered. 
  at this time we should be able to update that this TX were packed in a specific layer, but still in pending state
- applied to a state: `ReportNewTx` is called with valid layer ID, in applied state.

so i'd prefer this PR to add a explicit TX state and determine the state via pushing more events, instead of simply fixing the syntax.
 
> > tho, i'd prefer it to return layerID == 0 explicitly so the client knows that it's not applied to the mesh state yet.
> 
> Why do you think that lack of layerID doesn't look better than `layerID == 0`? I'd prefer to just omit it in this case.

this indicate that the system deliberately set it to 0 to indicate the pending state. in any case it's a minor point. i'd much prefer it explicitly shows that it's in pending state.

system/mocks/mocks.go
```diff
@@ -6,290 +6,289 @@ package mocks
...
 	context "context"
-	reflect "reflect"
-
 	gomock "github.com/golang/mock/gomock"
 	types "github.com/spacemeshos/go-spacemesh/common/types"
+	reflect "reflect"
 )
 
-// MockFetcher is a mock of Fetcher interface.
+// MockFetcher is a mock of Fetcher interface
```
dshulyak, line 14: do you know why this generated code got changed?

nkryuchkov: I changed a signature of one of the mocked methods, re-generated mocks, then reverted the change and re-generated again

api/grpcserver/mesh_service.go
```diff
@@ -264,6 +264,14 @@ func (s MeshService) getTxIdsFromMesh(minLayer types.LayerID, addr types.Address
 	return txIDs, nil
 }
 
+func convertLayerID(l types.LayerID) *pb.LayerNumber {
```
dshulyak, line 267: other changes seem to be unrelated, can we keep only this one and revert other changes

dshulyak (approved):

nkryuchkov:
@countvonzero 
> this PR currently is just fixing data presentation. but the goal of #2945 is to provide layer data in account stream.

Yes, #2945 provides it, but there's a small issue with data representation, which should be fixed by this PR.

> in terms of TX life cycle below, i think we may need to push more events to make the account stream more complete.
TX life cycle:
>
> enters mempool: ReportNewTx is called with empty layer ID, in pending state
packed in a block received in mesh: AddBlockWithTxs is called but no TX related event were triggered.
at this time we should be able to update that this TX were packed in a specific layer, but still in pending state
applied to a state: ReportNewTx is called with valid layer ID, in applied state.
so i'd prefer this PR to add a explicit TX state and determine the state via pushing more events, instead of simply fixing the syntax.

I think we may consider doing that, but I think we need a separate issue/PR for that

api/grpcserver/mesh_service.go
nkryuchkov: done, the mocks aren't reverted though, I hope that's no problem

nkryuchkov:
bors merge

nkryuchkov:
@countvonzero here's the issue for your proposal: https://github.com/spacemeshos/go-spacemesh/issues/2961

PR #2954, opened 2021-11-18
[Merged by Bors] - tortoise: vote against blocks that are supported by base block, but not by local opinion

### LINKED ISSUE #2424: Add exception votes against blocks the base block votes for

dshulyak:
## Motivation

closes: https://github.com/spacemeshos/go-spacemesh/issues/2424

## Changes
- vote against blocks that are supported by base block, but not by local opinion

## Test Plan
TestVoteAgainstSupportedByBaseBlock

## TODO

## DevOps Notes


tortoise/sim/layer.go
```diff
@@ -49,7 +54,16 @@ func WithoutInputVector() NextOpt {
...
 func WithEmptyInputVector() NextOpt {
 	return func(c *nextConf) {
-		c.EmptyHare = true
+		c.HareFraction.Nominator = 0
+	}
+}
+
+// WithPartialHare will set input vector only to a fraction of all known blocks.
+// Input vector will be limited to nominator*size/denominator.
+func WithPartialHare(nominator, denominator int) NextOpt {
```
countvonzero, line 63: check denominator != 0

countvonzero (approved):

nkryuchkov (approved):

dshulyak:
bors merge

Issue #2902, opened 2021-10-31
Refactor: `Block`→`Proposal`


noamnelke:
Change the structure of the existing blocks to actually be `Proposal`s. They contain all the same information, only have two signatures instead of one. At this point, we'll just do everything with the `Proposal` like today, as if it was a `Block` (including storing it in the blocks database).

Validation should be updated to check both signatures (both the `Proposal` and `Ballot`).

The blocks database should be renamed to `BallotDB` (even though at this point it actually stores proposals). In the next phase we'll create a `ProposalStore` and later on a `BlockDB` so we can't keep using this name.

At first the proposals stored in the `BallotDB` should be indexed by the `ProposalID` so all references to proposals, ballots and blocks should be of type `ProposalID`.

PR #2955, opened 2021-11-18
[Merged by Bors] - tortoise: use ballot to represent votes

countvonzero:
## Motivation

1st PR for #2902

## Changes

- add struct Ballot and Proposal
- change tortoise to use Ballot to represent votes
## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/common/types/ballot.go
+++ b/common/types/ballot.go
@@ -1,0 +1,166 @@
+package types
+
+import (
+	"bytes"
+
+	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/signing"
+)
+
+const (
+	// BallotIDSize in bytes.
+	// FIXME(dshulyak) why do we cast to hash32 when returning bytes?
+	BallotIDSize = Hash32Length
+)
+
+// BallotID is a 20-byte sha256 sum of the serialized ballot used to identify a Ballot.
+type BallotID Hash20
+
+// EmptyBallotID is a canonical empty BallotID.
+var EmptyBallotID = &BallotID{}
+
+// Ballot contains the smesher's signed vote on the mesh history.
+type Ballot struct {
+	// the actual votes on the mesh history
+	InnerBallot
+	// smesher's signature on InnerBallot
+	Signature []byte
+
+	// the following fields are kept private and from being serialized
+	ballotID BallotID
+	// the public key of the smesher used
+	smesherID *signing.PublicKey
+}
+
+// InnerBallot contains all info about a smesher's votes on the mesh history. this structure is
+// serialized and signed to produce the signature in Ballot.
+type InnerBallot struct {
+	// the smesher's ATX in the epoch this ballot is cast.
+	AtxID ATXID
+	// the proof of the smesher's eligibility to vote and propose block content in this epoch.
+	EligibilityProof VotingEligibilityProof
+
+	// a smesher creates votes in the following steps:
+	// - select a Ballot in the past as a base Ballot
+	// - calculate the opinion difference on history between the smesher and the base Ballot
+	// - encode the opinion difference in 3 list:
+	//	 - ForDiff
+	//	   contains blocks we support while the base block did not support (i.e. voted against)
+	//	   for blocks we support in layers later than the base block, we also add them to this list
+	//   - AgainstDiff
+	//     contains blocks we vote against while the base block explicitly supported
+	//	 - NeutralDiff
+	//	   contains blocks we vote neutral while the base block explicitly supported or voted against
+	//
+	// example:
+	// layer | unified content block
+	// -----------------------------------------------------------------------------------------------
+	//   N   | UCB_A (genesis)
+	// -----------------------------------------------------------------------------------------------
+	//  N+1  | UCB_B base:UCB_A, for:[UCB_A], against:[], neutral:[]
+	// -----------------------------------------------------------------------------------------------
+	//  N+2  | UCB_C base:UCB_B, for:[UCB_B], against:[], neutral:[]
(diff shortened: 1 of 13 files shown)


countvonzero:
bors try

dshulyak:
@countvonzero are we going to have both ballot and block when implementation is completed? can we have bid for ballot id please where bid was used previously

countvonzero:
> @countvonzero are we going to have both ballot and block when implementation is completed? can we have bid for ballot id please where bid was used previously

@dshulyak 
we will have both ballot and block eventually yes.
my plan is to typedef BlockID to ProposalID when i swap out Block with Proposal next. 
but will eventually reintroduce BlockID for real again for unified content block.

when all is said and done. there will be Ballot and Block in tortoise.

common/types/ballot.go
```diff
@@ -0,0 +1,166 @@
...
+	// BallotIDSize in bytes.
+	// FIXME(dshulyak) why do we cast to hash32 when returning bytes?
+	BallotIDSize = Hash32Length
+)
+
+// BallotID is a 20-byte sha256 sum of the serialized ballot used to identify a Ballot.
+type BallotID Hash20
+
+// EmptyBallotID is a canonical empty BallotID.
+var EmptyBallotID = &BallotID{}
```
dshulyak: why is it a pointer? can we define it as a value and get a pointer when we need a pointer?

common/types/ballot.go
```diff
@@ -0,0 +1,166 @@
...
+	// -----------------------------------------------------------------------------------------------
+	//  N+1  | UCB_B base:UCB_A, for:[UCB_A], against:[], neutral:[]
+	// -----------------------------------------------------------------------------------------------
+	//  N+2  | UCB_C base:UCB_B, for:[UCB_B], against:[], neutral:[]
+	// -----------------------------------------------------------------------------------------------
+	//  (hare hasn't terminated for N+2)
+	//  N+3  | UCB_D base:UCB_B, for:[UCB_B], against:[], neutral:[UCB_C]
+	// -----------------------------------------------------------------------------------------------
+	//  (hare succeeded for N+2 but failed for N+3)
+	//  N+4  | UCB_E base:UCB_C, for:[UCB_C], against:[UCB_D], neutral:[]
```
dshulyak: against will not be included in such case, at least in current implementation. also as i understand the current approach is a general agreement with research per https://github.com/spacemeshos/go-spacemesh/issues/2369

i understood that you prefer to set it explicit, but it is not consistent with every other place

common/types/ballot.go
```diff
@@ -0,0 +1,166 @@
...
+	ForDiff     []BlockID
+	NeutralDiff []BlockID
+
+	// the first Ballot the smesher cast in the epoch. this Ballot is a special Ballot that contains information
+	// that cannot be changed mid-epoch.
+	// - ActiveSet: from the smesher's view, the set of ATXs eligible to vote and propose block content in this epoch
+	// - Beacon: the beacon value the smesher recorded for this epoch
+	RefBallot BallotID
+	ActiveSet *[]ATXID
+	Beacon    []byte
```
dshulyak: it will be cleaner if ActiveSet and Beacon were a part of the same struct and we should have only a pointer to that struct. also why do we define `ActiveSet *[]ATXID` as a pointer to a pointer? it makes sense only in very specific cases

tortoise/verifying_tortoise.go
```diff
@@ -21,12 +21,11 @@ type atxDataProvider interface {
 }
 
 var (
-	errNoBaseBlockFound                 = errors.New("no good base block within exception vector limit")
-	errNotSorted                        = errors.New("input blocks are not sorted by layerID")
-	errstrNoCoinflip                    = "no weak coin value for layer"
-	errstrTooManyExceptions             = "too many exceptions to base block vote"
-	errstrConflictingVotes              = "conflicting votes found in block"
-	errstrUnableToCalculateLocalOpinion = "unable to calculate local opinion for layer"
+	errNoBaseBlockFound     = errors.New("no good base block within exception vector limit")
```
dshulyak: ballot?

tortoise/verifying_tortoise.go
```diff
@@ -566,42 +571,43 @@ func (t *turtle) processBlock(ctx context.Context, block *types.Block) error {
 			opinion[blk] = nvote
 		}
 	}
-	t.BlockLayer[block.ID()] = block.LayerIndex
-	logger.With().Debug("adding or updating block opinion")
-	t.BlockOpinionsByLayer[block.LayerIndex][block.ID()] = opinion
+	t.BallotLayer[ballot.ID()] = ballot.Layer()
```
dshulyak: it should be LayerIndex, there is also a Layer object that stores collection of blocks

dshulyak (approved):

common/types/ballot.go
countvonzero: oh i was mentally thinking when hare timeout in tortoise's view (no result past zdist+confidence param). you are right. in this case it is an implicit against.

common/types/ballot.go
countvonzero: done. was copying `EmptyATXID`

common/types/ballot.go
countvonzero: done and done

tortoise/verifying_tortoise.go
countvonzero: done

tortoise/verifying_tortoise.go
countvonzero: done

countvonzero:
bors merge

Issue #2946, opened 2021-11-16
tortoise: for local/global opinion threshold tortoise should use space weight, not layer size


dshulyak:
## Description

This is defined in protocol paper. Tortoise should use the expected weight of votes per layer. In the implementation we are using average layer size, which is constant.

My interpretation is that this is critical for recovering from partitions that lasted multiple epochs, so that every partition agreed on its own tortoise beacon and created average number of blocks per layer. However after rejoining each part of the network should not be able to make progress without blocks from another part, this is why threshold should depend on the actual weight.

PR #2956, opened 2021-11-19
[Merged by Bors] - tortoise: add generator for partition and a negative test case

dshulyak:
## Motivation

Partition simulates splitting states as if the network was separated for longer than an epoch. After split - each par will have its own atx's, beacon and share of miners.

Merge consumes state from another partition, as would happen after state sync.

There is a test TestRecoverFromFullPartition, which confirms that we do not recover from such partition by rerun. I found several issues that prevent from healing:
- rerun should not even try to run verifying tortoise for layers that are before hdist. it always counts votes from good blocks, which prematurely confirms the layer
- we are using layer size and not weight for computing threshold - https://github.com/spacemeshos/go-spacemesh/issues/2946 , because threshold didn't change when we received more atx's we are deciding on validity before we can count votes with bad beacon. if weight would have been doubled we will need more votes and it will make votes from blocks with bad beacon count eventually.

## TODO

## DevOps Notes

### DIFF
--- a/tortoise/sim/partition.go
+++ b/tortoise/sim/partition.go
@@ -1,0 +1,132 @@
+package sim
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+// Frac is a shortcut for creating Fraction object.
+func Frac(nominator, denominator int) Fraction {
+	return Fraction{Nominator: nominator, Denominator: denominator}
+}
+
+// Fraction of something.
+type Fraction struct {
+	Nominator, Denominator int
+}
+
+func (f Fraction) String() string {
+	return fmt.Sprintf("%d/%d", f.Nominator, f.Denominator)
+}
+
+// SplitOpt is for configuring partition.
+type SplitOpt func(*splitConf)
+
+// WithPartitions configures number of miners in each partition
+// relative to original Generator.
+func WithPartitions(parts ...Fraction) SplitOpt {
+	return func(c *splitConf) {
+		c.Partitions = parts
+	}
+}
+
+func splitConfDefaults() splitConf {
+	return splitConf{
+		Partitions: []Fraction{Frac(1, 2), Frac(1, 2)},
+	}
+}
+
+type splitConf struct {
+	Partitions []Fraction
+	// TODO(dshulyak) support both transient and full partitions
+	// in transient partition miners still have the same atxs and beacons
+	//
+	// transient partitions can be configured without splitting state, in transient partitions
+	// - votes are splitted
+	// - hare will not reach consensus (or will reach only in one of the partition)
+	// all of this can be done by option to Next
+}
+
+// Split generator into multiple partitions.
+// First generator will use original tortoise state (mesh, activations).
+func (g *Generator) Split(opts ...SplitOpt) []*Generator {
+	conf := splitConfDefaults()
+	for _, opt := range opts {
+		opt(&conf)
+	}
+	if len(conf.Partitions) < 2 {
+		panic("should be atlest two parititions")
+	}
+	gens := make([]*Generator, len(conf.Partitions))
+	total := len(g.activations)
+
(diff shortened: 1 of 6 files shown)


tortoise/sim/partition.go
```diff
@@ -0,0 +1,186 @@
...
+	return splitConf{
+		Partitions: []Fraction{Frac(1, 2), Frac(1, 2)},
+	}
+}
+
+type splitConf struct {
+	Partitions []Fraction
+	// TODO(dshulyak) support both transient and full partitions
+	//
+	// transient partitions can be configured without splitting state, in transient partitions
```
countvonzero, line 46: is a transient partition like a malicious attack in the network?

tortoise/sim/partition.go
```diff
@@ -0,0 +1,186 @@
...
+// therefore rerun after merge will work only in the first generator.
+func (g *Generator) Split(opts ...SplitOpt) []*Generator {
+	// Split everything that would be changed after long network partition
+	// - activations
+	// - beacons
+	// - number of miners in each partition
+	//
+	//  things that should remain the same:
+	// - blocks, contextually valid blocks and input vectors before partition
+	// - activations before partition
```
countvonzero, line 67: aren't beacons before partition also unchanged?

tortoise/sim/partition.go
```diff
@@ -0,0 +1,186 @@
...
+
+	conf := splitConfDefaults()
+	for _, opt := range opts {
+		opt(&conf)
+	}
+	if len(conf.Partitions) < 2 {
+		panic("should be atlest two parititions")
```
countvonzero: :s/atlest/at least

tortoise/verifying_tortoise_test.go
```diff
@@ -3357,3 +3363,87 @@ func TestComputeLocalOpinion(t *testing.T) {
...
+		_, verified1, _ = tortoise1.HandleIncomingLayer(ctx, last)
+	}
+	require.NoError(t, tortoise1.rerun(ctx))
+	last = s1.Next()
+	_, verified1, _ = tortoise1.HandleIncomingLayer(ctx, last)
+	require.Equal(t, last.Sub(1), verified1)
+
+	// verify that all blocks that were created in s2 during partition are
+	// contextually valid in s1 state.
+	// this is not the case right, this is a negative test case
```
countvonzero: rephrase?

tortoise/verifying_tortoise_test.go
```diff
@@ -3357,3 +3363,87 @@ func TestComputeLocalOpinion(t *testing.T) {
 		})
 	}
 }
+
+func TestRecoverFromFullPartition(t *testing.T) {
```
countvonzero: how about `TestNetowrkDoesNotRecoverFromFullPartition`?

countvonzero (approved):

tortoise/sim/partition.go
```diff
@@ -0,0 +1,186 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+// Frac is a shortcut for creating Fraction object.
+func Frac(nominator, denominator int) Fraction {
```
nkryuchkov, line 12: Why not `big.NewRat`?

nkryuchkov (approved):

tortoise/sim/partition.go
dshulyak: 1. simpler function signature
2. no need to cast big int to int when i will use fraction

tortoise/sim/partition.go
dshulyak: i was thinking more about just a regular network partition, but within single epoch. so there will be no new atx's and beacon remains the same

tortoise/sim/partition.go
dshulyak: yes they should, i will mention them

dshulyak:
bors merge

PR #2959, opened 2021-11-22
[Merged by Bors] - tortoise: refactor initialization and configuration for tortoise

### LINKED ISSUE #2707: Separate tortoise config

dshulyak:
## Motivation

Config for a tortoise is a bit messy, it includes both mandatory deps (like mesh db, beacons, atx db) and immutable protocol parameters as a part of the same struct. 

closes: https://github.com/spacemeshos/go-spacemesh/issues/2707

## Changes
- refactor tortoise module to have cleaner initialization (New function, and options for non-mandatory parameters)
- refactor tortoise config to avoid duplication and use it when initializing node

## Test Plan
ut

## TODO

## DevOps Notes


tortoise/algorithm.go
```diff
@@ -0,0 +1,212 @@
...
+		t.cfg,
+	)
+
+	if err := t.trtl.Recover(); err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			t.trtl.init(t.ctx, mesh.GenesisLayer())
+		} else {
+			t.logger.With().Panic("can't recover turtle state", log.Err(err))
+		}
+	}
```
nkryuchkov: What do you think about removing the `else` statement?
```suggestion
	if err := t.trtl.Recover(); err != nil {
		if !errors.Is(err, database.ErrNotFound) {
			t.logger.With().Panic("can't recover turtle state", log.Err(err))
		}
		t.trtl.init(t.ctx, mesh.GenesisLayer())
	}
```

nkryuchkov (approved):

dshulyak:
bors merge

dshulyak:
looks like setup issue `Exception: genesis_delta time=75sec, is too short for this deployment. delta_from_genesis=-47.795942`

bors merge

dshulyak:
bors merge

PR #2960, opened 2021-11-22
[Merged by Bors] - blocks: add validation for conflicting exceptions and max exceptions limit

### LINKED ISSUE #2673: In Tortoise save and vote against blocks that exceed the max exception list size for DoS prevention
## Description
Self-explanatory

## Affected code
https://github.com/spacemeshos/go-spacemesh/blob/9d3a379994e3652eb0378d2cc555588bba91005e/tortoise/verifying_tortoise.go#L595

This issue appears in commit hash: 9d3a379994e3652eb0378d2cc555588bba91005e

dshulyak:
## Motivation

blocks must be validated before they are used as an input vector in the hare, we can't drop them in tortoise once they are in input vector as it will force tortoise to switch to self-healing mode.

closes https://github.com/spacemeshos/go-spacemesh/issues/2673

## Changes
- validates that the block doesn't have conflicting votes and that the votes are within limit
- don't skip blocks in tortoise on error, instead fail (there should be no skippable errors in consensus code)

## Test Plan
ut

## TODO

## DevOps Notes


blocks/blocks.go
```diff
@@ -41,46 +39,70 @@ type blockValidator interface {
...
 }
 
 // Config defines configuration for block handler.
 type Config struct {
-	Depth       uint32
-	GoldenATXID types.ATXID
+	MaxExceptions int
+}
+
+// defaultConfig fro BlockHandler.
```
countvonzero: :s/fro/for

blocks/blocks.go
```diff
@@ -41,46 +39,70 @@ type blockValidator interface {
...
+func NewBlockHandler(fetcher system.Fetcher, m mesh, v blockValidator, opts ...Opt) *BlockHandler {
+	b := &BlockHandler{
+		logger:    log.NewNop(),
+		cfg:       defaultConfig(),
+		fetcher:   fetcher,
+		mesh:      m,
+		validator: v,
+	}
+	for _, opt := range opts {
+		opt(b)
```
countvonzero, line 89: i always wonder why some args are configured and some now when most of them will be manipulated during testing.

what's your rationale not make all arguments to BlockHandler as options?

blocks/blocks.go
```diff
@@ -96,28 +118,19 @@ func (bh *BlockHandler) HandleBlockData(ctx context.Context, data []byte) error
...
-		logger.With().Error("failed to add block to database", log.Err(err))
-		// we return nil here so that the block will still be propagated
-		return nil
-	}
-
-	if !blk.Layer().After(bh.mesh.ProcessedLayer()) { //|| blk.Layer() == bh.mesh.getValidatingLayer() {
-		logger.With().Warning("block is late",
-			log.FieldNamed("processed_layer", bh.mesh.ProcessedLayer()),
-			log.FieldNamed("miner_id", blk.MinerID()))
+		return fmt.Errorf("persisting block %s: %w", blk.ID(), err)
```
countvonzero: adding block to mesh is not exactly the same as persisting...

blocks/blocks.go
```diff
@@ -166,13 +179,36 @@ func blockDependencies(blk *types.Block) []types.BlockID {
 	return combined
 }
 
+func validateExceptions(block *types.Block, max int) error {
```
countvonzero, line 182: maybe it's worth returning all the BlockIDs as list for fetching later in the code.

tortoise/tortoise.go
```diff
@@ -540,31 +529,15 @@ func (t *turtle) processBallots(ctx *tcontext, ballots []*types.Ballot) error {
 				log.Named("last_evicted", t.LastEvicted))
 			continue
 		}
-		if b.LayerIndex().Before(lastLayerID) {
-			return errNotSorted
-		} else if b.LayerIndex().After(lastLayerID) {
-			lastLayerID = b.LayerIndex()
-		}
-
```
countvonzero, line 548: double-checking. this check is removed because we don't call processBallots on late blocks anymore?

countvonzero (approved):

blocks/blocks.go
dshulyak: i think there should be no "required" options, so i am passing all dependencies explicitly. if one of those deps is not provided block handler will crash with nil-pointer error.

blocks/blocks.go
dshulyak: why? i think it is simpler and has almost no overhead to use them from the block again

tortoise/tortoise.go
dshulyak: pretty much, since we always sort layers with organizer i don't see in what case this check is useful

blocks/blocks.go
dshulyak: right, need to change, i checked why it can fail, and wrote a message based on that

dshulyak:
bors merge

nkryuchkov (approved):

PR #2963, opened 2021-11-23
[Merged by Bors] - tortoise: use expected layer weight when computing thresholds

### LINKED ISSUE #2946: tortoise: for local/global opinion threshold tortoise should use space weight, not layer size

dshulyak:
## Motivation

This is basic for security of the tortoise, otherwise voting is disproportionate to the actual weight and block can be verified even when they supported by true majority (see TestVerifyLayerByWeightNotSize) . 

Also important for recovery from the partition where we learn about new atxs. During rerun we will see that the weight increases but we can't make progress with votes from a single beacon, so the verified layer will be stuck until votes from ballots with "bad" beacon can be counted. Will send a followup change for this use case. 

closes https://github.com/spacemeshos/go-spacemesh/issues/2946

## Changes
- expected weight of votes for a block in layer N is derived from atxs of the ballots from all layers after N up to last known

## Test Plan
TestVerifyLayerByWeightNotSize, TestComputeExpectedWeight and other uts

## TODO

## DevOps Notes


tortoise/util.go
```diff
@@ -83,21 +83,19 @@ var thetaMu sync.Mutex
 
 // calculateOpinionWithThreshold computes opinion vector (support, against, abstain) based on the vote weight
 // and theta, layer size and delta.
-func calculateOpinionWithThreshold(logger log.Log, v vec, theta *big.Rat, layerSize uint32, delta uint32) vec {
+func calculateOpinionWithThreshold(logger log.Log, v vec, theta *big.Rat, weight uint64, delta uint32) vec {
```
countvonzero: delta unused?

tortoise/util.go
```diff
@@ -83,21 +83,19 @@ var thetaMu sync.Mutex
 
 // calculateOpinionWithThreshold computes opinion vector (support, against, abstain) based on the vote weight
 // and theta, layer size and delta.
```
countvonzero: plz update comment

countvonzero (approved):
question: is it more accurate to use `GetEpochWeight()/layersPerEpoch`?

dshulyak:
this is not finished yet, i learned new details and switched it to draft

> question: is it more accurate to use GetEpochWeight()/layersPerEpoch?

i thought about this, but then i found the details in the paper and it suggested to use 
 
`wB: total weight of blocks in Bt that are eligible to vote about B’s validity`

it was computed by summing up weight of blocks when votes were counted, but Tal described another approach which is closer to what you mentioned but a bit more involved. so i will be changing it

dshulyak:
@countvonzero i made final changes after latest discussion, ready for review

tortoise/tortoise.go
```diff
@@ -1110,3 +1119,35 @@ func (t *turtle) ballotFilterForHealing(candidateLayerID types.LayerID, logger l
...
+		layers = uint64(types.GetLayersPerEpoch())
+	)
+	// TODO(dshulyak) this can be improved by computing weight for epoch at once
+	for lid := target.Add(1); !lid.After(last); lid = lid.Add(1) {
+		weight, err := getEpochWeight(atxdb, epochWeight, lid.GetEpoch())
+		if err != nil {
+			return nil, err
+		}
+		layerWeight := new(big.Float).SetUint64(weight)
+		layerWeight.Quo(layerWeight, new(big.Float).SetUint64(layers))
```
countvonzero, line 1149: i read Tal's comment after i commented. i think he was looking for max weight in that epoch to calculate the max possible weight here. but i think this is also good enough.

countvonzero (approved):

tortoise/tortoise.go
dshulyak: can you clarify? in my understanding this is exactly what he meant, there is even similar test:

> To give a concrete example, suppose epochs are exactly 100 layers each, block X is in layer 80 and we're currently in layer 210. Also, suppose the total ATX weight for the first epoch is 30, for the second epoch 40 and for the third epoch 50.  Then in the first epoch, there are only 20 (out of 100) layers that can vote about block X, so we count the maximum weight in that epoch as 30*20/100 = 6. For the second epoch, all layers are eligible to vote, so the total voting weight is 40. In the third epoch, there are only 10 layers so far, so the expected voting weight is 50*10/100 = 5. Thus, for the purpose of the global threshold,  the total voting weight is 6+40+5=51.

the only difference is that i am not computing full layer weight at once, instead it will added incrementally

countvonzero: you are right. i misread Tal's comment. i thought he meant to find the max weight in that epoch and extrapolate. my bad.

dshulyak:
bors merge

PR #2965, opened 2021-11-24
[Merged by Bors] - UCB: add gossip handlers for Proposal and Ballot

countvonzero:
## Motivation

3rd PR for #2902

## Changes

- add gossip handler for Proposal and Ballot

PRs to follow:
- miner produce Proposal
- layer fetch and fetcher fetch Proposal
- save Proposal and Ballot in mesh

## Test Plan

unit tests

## DevOps Notes


proposals/eligibility_validator_test.go
```diff
@@ -0,0 +1,464 @@
...
+	activeSet   = types.ATXIDList{types.RandomATXID(), types.RandomATXID()}
+	defaultUnit = uint(5)
+	ballotUnit  = uint(1)
+	// eligibleSlots is calculated based on layerAveSize, layersPerEpoch, epoch ATX weight and smesher's own weight.
+	eligibleSlots = uint32(3)
+	epoch         = types.EpochID(3)
+	beacon        = types.RandomBytes(types.BeaconSize)
+)
+
+type testedValidator struct {
```
dshulyak: testValidator ?

proposals/eligibility_validator_test.go
```diff
@@ -0,0 +1,464 @@
...
+	"github.com/spacemeshos/go-spacemesh/log/logtest"
+	"github.com/spacemeshos/go-spacemesh/signing"
+)
+
+var (
+	activeSet   = types.ATXIDList{types.RandomATXID(), types.RandomATXID()}
+	defaultUnit = uint(5)
+	ballotUnit  = uint(1)
+	// eligibleSlots is calculated based on layerAveSize, layersPerEpoch, epoch ATX weight and smesher's own weight.
+	eligibleSlots = uint32(3)
```
dshulyak, line 20: uints can be constant, i generally don't see any reason to use global variables in tests and would be better to create initializers for common objects, e.g. genActiveSet(), genBeacon

proposals/handler.go
```diff
@@ -0,0 +1,380 @@
...
+
+// WithLayerPerEpoch defines the number of layers per epoch.
+func WithLayerPerEpoch(layers uint32) Opt {
+	return func(h *Handler) {
+		h.cfg.LayersPerEpoch = layers
+	}
+}
+
+// WithValidator defines eligibility validator.
+func WithValidator(v eligibilityValidator) Opt {
```
dshulyak: is it only for testing? can use private option if so

proposals/handler.go
```diff
@@ -0,0 +1,380 @@
...
+	if err := h.processProposal(ctx, msg); err != nil {
+		h.logger.WithContext(ctx).With().Error("failed to process proposal gossip", log.Err(err))
+		return pubsub.ValidationIgnore
+	}
+	return pubsub.ValidationAccept
+}
+
+// HandleBallot is the gossip receiver for Ballot.
+// TODO: register this gossip handler.
+func (h *Handler) HandleBallot(ctx context.Context, _ peer.ID, msg []byte) pubsub.ValidationResult {
```
dshulyak: why do we gossip both but proposal also has a ballot included?

proposals/handler.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+
+	if err := h.checkExceptions(b, h.cfg.MaxExceptions); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (h *Handler) checkExceptions(ballot *types.Ballot, max int) error {
```
dshulyak: it doesn't have to be a method, it deliberately passes all dependencies using arguments, so that there is no implicit stuff going on

proposals/handler.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+	if len(exceptions) > max {
+		return fmt.Errorf("%w: %d exceptions with max allowed %d in ballot %s",
+			errExceptionsOverflow, len(exceptions), max, ballot.ID())
+	}
+	return nil
+}
+
+func ballotBlockView(b *types.Ballot) []types.BlockID {
+	combined := make([]types.BlockID, 0, len(b.AgainstDiff)+len(b.ForDiff)+len(b.NeutralDiff))
```
dshulyak, line 302: why base ballot was removed from this list and now fetched separately? can we join everything then into a single list and fetch in a single call

proposals/handler.go
```diff
@@ -0,0 +1,380 @@
...
+
+	reportBallotMetrics(b)
+	return nil
+}
+
+func (h *Handler) checkBallotSyntacticValidity(ctx context.Context, b *types.Ballot) error {
+	h.logger.WithContext(ctx).With().Debug("checking proposal syntactic validity")
+
+	if err := h.checkBallotDataIntegrity(b); err != nil {
+		h.logger.WithContext(ctx).With().Error("ballot integrity check failed", log.Err(err))
```
dshulyak: this and following logs should be wanings. receiving invalid data doesn't indicate any problems with protocol or node

dshulyak (commented):
what is the idea behind publishing both ballots and proposals separately, but including ballots into proposals? i guess validity of the proposal is based on the ballot validity, that's why it is included.  but in what case ballot can be published separately?

what do you think about two different modules for proposal handler and ballot handler, it is generally easier to see dependencies and interactions between modules

proposals/eligibility_validator_test.go
countvonzero: done

proposals/handler.go
countvonzero: done

proposals/handler.go
countvonzero: sorry. this was a mistake. we don't gossip Ballot separately. we will only fetch Ballot directly from peers.

proposals/handler.go
countvonzero: done

proposals/handler.go
countvonzero: base ballot is now fetched together with ref ballot if any. the separation is done because we have different API for fetching Ballot (`GetBallots()`) and Block (`GetBlocks()`).

proposals/handler.go
countvonzero: done

countvonzero:
> what is the idea behind publishing both ballots and proposals separately, but including ballots into proposals? i guess validity of the proposal is based on the ballot validity, that's why it is included. but in what case ballot can be published separately?

i was wrong. we only need to fetch `Ballot` independently from peers, but not gossip about `Ballot`.
 
> what do you think about two different modules for proposal handler and ballot handler, it is generally easier to see dependencies and interactions between modules

without the Ballot gossip handler, do you still prefer to break them to separate modules? i find them quite closely connected and logically easier to see all code in the same module.

dshulyak (approved):
> without the Ballot gossip handler, do you still prefer to break them to separate modules? i find them quite closely connected and logically easier to see all code in the same module.

one module works for me as well

countvonzero:
bors merge

PR #2968, opened 2021-11-25
[Merged by Bors] - tortoise: recover from multi-epoch partition (partial changes)

dshulyak:
## Changes
- introduce Processed layer, Last should be the last layer according to the clock and used for computing Hdist distance, beacon eligibility, and some other time-sensitive parameters . During rerun Last needs to be preserved so that we can make correct decision which protocol to use 
- change beacon eligibility condition, from `vote layer - candidate layer > bad beacon distance` to `current layer - vote layer > bad beacon distance`, otherwise votes with bad beacon are not counted when they suppose to be counted. 

## Test Plan
TestNetworkRecoversFromFullPartition

## TODO

## DevOps Notes


tortoise/tortoise.go
```diff
@@ -943,7 +955,7 @@ func (t *turtle) computeLocalOpinion(ctx *tcontext, lid types.LayerID) (map[type
 		return opinion, nil
 	}
 
-	weight, err := computeExpectedVoteWeight(t.atxdb, t.epochWeight, lid, t.Last)
+	weight, err := computeExpectedVoteWeight(t.atxdb, t.epochWeight, t.Processed.Sub(1), t.Processed)
```
dshulyak, line 950: submit separately

tortoise/sim/state.go
```diff
@@ -0,0 +1,64 @@
...
+// State of the node.
+type State struct {
+	logger log.Log
+
+	MeshDB  *mesh.DB
+	AtxDB   *activation.DB
+	Beacons *beaconStore
+}
+
+// OnBeacon ...
```
nkryuchkov: Missing comment (same below)

nkryuchkov (approved):

tortoise/sim/generator.go
```diff
@@ -106,16 +107,16 @@ type Generator struct {
...
 	conf   config
 
-	State State
+	states []State
 
 	nextLayer types.LayerID
 	// key is when to return => value is the layer to return
 	reordered   map[types.LayerID]types.LayerID
 	layers      []*types.Layer
+	units       [2]int
```
countvonzero, line 116: what is the 1st and what is the 2nd element?

tortoise/tortoise.go
```diff
@@ -842,44 +846,52 @@ func (t *turtle) healingTortoise(ctx *tcontext, logger log.Log, lid types.LayerI
...
-		contextualValidity, err := t.verifyingTortoise(ctx, logger, candidateLayerID)
-		if err != nil {
-			return err
+		var (
+			contextualValidity map[types.BlockID]bool
+			err                error
+		)
+		// during rerun verifying tortoise prematurely verifies layers, which full tortoise won't be a
+		// able to verify because of additional eligible ballots with wrong beacon
+		if !(candlid.Before(t.layerCutoff()) && t.Last.Difference(candlid) > t.Zdist+t.ConfidenceParam) {
```
countvonzero: i am not able to connect the comments and the code. why do we only run verifying tortoise for recent layers?

> don't use verifying tortoise for old layers, otherwise, it prematurely verifies layers because blocks with bad beacon are ignored. this problem will be resolved when ballot weight will be calculated properly.

isn't this by design? if there are two partitions with two different beacons, each partition should be happily verifying old layers until the rejoin. no?

tortoise/tortoise.go
countvonzero: i don't understand this change. it should be the number of layers that could've cast votes on lid, no?

tortoise/tortoise.go
```diff
@@ -1116,7 +1129,10 @@ func (t *turtle) ballotFilterForHealing(candidateLayerID types.LayerID, logger l
 			logger.With().Error("inconsistent state: ballot not found", ballotID)
 			return false
 		}
-		return lid.Uint32() > t.BadBeaconVoteDelayLayers && lid.Sub(t.BadBeaconVoteDelayLayers).After(candidateLayerID)
+		// if candidate layer is 11 and vote with bad beacon is in 12, but current is 100
+		// we should obviously count such vote.
+		// is there any case
```
countvonzero: maybe this is a misinterpretation on my part? by delaying votes for beacon for N layers, do you think Tal means `current-ballotLayer`?

tortoise/tortoise.go
```diff
@@ -652,16 +653,19 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+func (t *turtle) HandleIncomingLayer(ctx context.Context, lid types.LayerID) error {
 	defer t.evict(ctx)
 
 	tctx := wrapContext(ctx)
 	// unconditionally set the layer to the last one that we have seen once tortoise or sync
 	// submits this layer. it doesn't matter if we fail to process it.
-	if t.Last.Before(layerID) {
-		t.Last = layerID
+	if t.Last.Before(lid) {
+		t.Last = lid
```
countvonzero, line 663: why shouldn't t.Last get updated by the clock?

tortoise/tortoise.go
countvonzero: the comment is incomplete

tortoise/tortoise.go
countvonzero: ah. this is the local opinion. ok. i see in the other PR

tortoise/sim/generator.go
dshulyak: lower and upper bound

tortoise/tortoise.go
dshulyak: my interpretation is that we shouldn't count votes with bad beacon in current epoch

> the tortoise does count the ballots from previous epochs regardless of the beacon, but because there was a partition, each side sees less total weight and thus generates double the number of ballots

tortoise/tortoise.go
dshulyak: > isn't this by design? if there are two partitions with two different beacons, each partition should be happily verifying old layers until the rejoin. no?

yes, but i don't see any contradiction. it will work as you said. verifying tortoise should have the same output as full, but because bugs with weight it currently doesn't

we  cannot run verifying tortoise during rerun because atx weight in the implementation is meaningless (i am fixing it in next pr). so if after rejoin verifying tortoise will ignore bad blocks and erroneously verify layers which wouldn't be verified by full tortoise. because full tortoise will also count blocks with bad beacons.

tortoise/tortoise.go
dshulyak: i don't get the question. Last should be updated by the clock (but currently isn't), Processed shouldn't

tortoise/tortoise.go
countvonzero: i think adding the if condition here is confusing. if this is a bug due to ATX weight, then maybe we should revisit this PR after the ATX weight is fixed.

in my mind, after rejoin, rerun will find out that there are a lot more weight ignored, and a different beacon. that should cause a revert in state because the old verified layers are no longer crossing the global threshold. self-healing/full tortoise should now kick in

tortoise/tortoise.go
countvonzero: that's not my understanding. Tal explicitly said delay by X layers. there was no mention about current epoch. maybe this is worth clarifying with him.

tortoise/tortoise.go
countvonzero: ok. the PR description made me think you were going to update Last by the clock. maybe leave a TODO?

dshulyak: why? there is already an issue for that, and this change affects Last only minimally

tortoise/tortoise.go
dshulyak: > Tal explicitly said delay by X layers

i would interpret that as delay by x layers from current layer, so that it can be less than single epoch. but i asked in the channel

tortoise/tortoise.go
dshulyak: why is it confusing? in my opinion running full tortoise explicitly if we can is less confusing than using it as a fallback. but sure i can revert this change from this pr.

countvonzero: > running full tortoise explicitly if we can is less confusing than using it as a fallback.

the confusing part for me is the special case with the old layers. my understanding is that the protocol, by design, avoid running full tortoise as much as possible. a fallback is exactly what it should be. self-healing should be triggered as rare as we possibly can. i think we should clarify with Tal about this too.

countvonzero: i would prefer your ATX weight PR fix the bug first and re-evaluate the conditions that trigger self-healing/full tortoise.

dshulyak: > the confusing part for me is the special case with the old layers

this is because there is a condition that doesn't allow full tortoise to run for new layers. 

> my understanding is that the protocol, by design, avoid running full tortoise as much as possible. a fallback is exactly what it should be.

there is no reason to do so, the consensus output of verifying and full should be the same by definition, and it is not much different in terms of computation. i will clarify but i don't see why it needs to be fallback

dshulyak:
bors merge

tortoise/tortoise.go
countvonzero: @dshulyak this is because currently verifying tortoise is doing exactly the same thing as full tortoise, which is not the intention. verifying tortoise is supposed to only count margins only, not ALL the votes.

PR #2970, opened 2021-11-26
[Merged by Bors] - tortoise: vote weight is atx weight adjusted by the number of eligible ballots based on active set 

### LINKED ISSUE #2966: tortoise: recover from partition with separate tortoise beacons and atxs
It was blocked by https://github.com/spacemeshos/go-spacemesh/issues/2946

Now it should be possible to find the last culprits and adjust TestNetworkDoesNotRecoverFromFullPartition to work as intended.

### LINKED ISSUE #2967: tortoise: weight of the ballot should be equal to weight of the atx divided by the number of eligible ballots in the epoch
TODO details should be added from conversation on slack, leaving it is a reminder for now

dshulyak:
## Motivation

Per protocol the weight of the ballot should be the weight of the atx divided by the number of ballots in the epoch. It is critical both for the crash and byzantine failures.

closes: https://github.com/spacemeshos/go-spacemesh/issues/2967
closes: https://github.com/spacemeshos/go-spacemesh/issues/2966

## Changes
- compute proper weight for the ballot
- ballot weight is stored in a map BallotWeight, opinions map stores only sign of the vote (0, -1, 1)
- refactor some unit tests for better tests

## Test Plan
system tests and TestSumVotes, TestComputeBallotWeight, TestNetworkRecoversFromFullPartition

## TODO

## DevOps Notes


nkryuchkov (approved):

tortoise/tortoise.go
```diff
@@ -498,12 +483,15 @@ func (t *turtle) processBallot(ctx context.Context, ballot *types.Ballot) error
...
-			nvote := simplifyVote(vote).Multiply(voteWeight)
-			opinion[blk] = nvote
+			opinion[blk] = vote.copy()
 		}
 	}
+
+	logger.With().Debug("adding or updating ballot opinion", log.Stringer("ballot_weight", weight))
+
+	t.BallotWeight[ballot.ID()] = weight
+	t.BlockLayer[types.BlockID(ballot.ID())] = ballot.LayerIndex
```
countvonzero: why is this needed? we should only look up t.BlockLayer when we look to trim blocks in the opinion map. 
we should only do this when processing blocks (not ballots)

tortoise/tortoise.go
```diff
@@ -1156,3 +1147,121 @@ func computeExpectedVoteWeight(atxdb atxDataProvider, epochWeight map[types.Epoc
...
+				foundWeight = true
+			}
+		}
+		// TODO is it valid not to include your own atxid into active set?
+		if !foundWeight {
+			atx, err := atxdb.GetAtxHeader(ballot.AtxID)
+			if err != nil {
+				return nil, fmt.Errorf("atx %s in active set of %s is unknown", ballot.AtxID, ballot.ID())
+			}
+			weight = atx.GetWeight()
```
countvonzero: if the ballot's ATXID not in the active set, then miner is not eligible to vote. we should return zero weight instead, even if it is found in the atx db

tortoise/tortoise.go
```diff
@@ -1156,3 +1147,121 @@ func computeExpectedVoteWeight(atxdb atxDataProvider, epochWeight map[types.Epoc
...
+	ballot *types.Ballot,
+	layerSize,
+	layersPerEpoch uint32,
+) (*big.Float, error) {
+	if ballot.EpochData != nil {
+		var (
+			total, weight uint64
+			foundWeight   bool
+		)
+		for _, atxid := range ballot.EpochData.ActiveSet {
```
countvonzero, line 1175: what's the rationale that we use the active set from the ref ballot instead of from our local DB?

can you provide links to the slack discussion for more context?

tortoise/tortoise.go
```diff
@@ -692,13 +680,22 @@ func (t *turtle) handleLayer(ctx *tcontext, layerID types.LayerID) error {
 		t.logger.WithContext(ctx).Warning("cannot process empty layer block list")
 		return nil
 	}
-	ballots := make([]*types.Ballot, 0, len(layerBlocks))
-	for _, b := range layerBlocks {
-		// cache block's layer so we can check to ignore opinions on very old blocks
-		t.BlockLayer[b.ID()] = b.LayerIndex
```
countvonzero, line 698: why remove it from here? this will be the real blocks we vote on.

tortoise/tortoise.go
```diff
@@ -1156,3 +1147,121 @@ func computeExpectedVoteWeight(atxdb atxDataProvider, epochWeight map[types.Epoc
...
+			}
+		}
+	}
+	weight, err := computeBallotWeight(atxdb, weights, ballot, layerSize, layersPerEpoch)
+	if err != nil {
+		return err
+	}
+	// reference ballot can be evicted, in order to guarantee that we won't
+	// add evicted ref ballot - we always delete reference weight if it didn't
+	// exist when we computed weight for current ballot
```
countvonzero, line 1264: the logic here seems more complicated than needed. do you think it's worth evicting weight by epoch?

tortoise/tortoise.go
dshulyak: i don't understand the question. we get weight from active set and store it in memory. what do you mean by local db?

tortoise/tortoise.go
dshulyak: i think the current version is simpler overall, indexing weight by epoch will simplify some things and make complex other. this is only for recovery, it is not used in the regular eviction code path

tortoise/tortoise.go
dshulyak: it is needed for unit tests that are not using db. it was previously done in handleLayer that reads blocks from database, if later there will be separate code path for blocks there should be processBlocks then

tortoise/tortoise.go
dshulyak: should be probably validated in proposals handler

tortoise/tortoise.go
dshulyak: i added processBlocks function

tortoise/tortoise.go
countvonzero: i meant `atxdb.GetEpochWeight()`
the slack thread you pointed out didn't help me tho. why should we assign weight according to their view and not the local view?

say our local atxdb has ATX of  weight 1000 for epoch 10. in the ballot, there is active set that weigh 2000 for epoch 10. 
we calculate the global threshold based on our local atxdb. in that case, we will assign too much weight for this ballot?

tortoise/tortoise.go
countvonzero: it's handled in `handleLayer` because that's the only place we will potentially see Block. this is the only time we know about what blocks belong to which layer. it still doesn't make sense to do it here. BallotID will be totally different from BlockID/ProposalID in a few more PRs. 

can you change the unit test instead?

tortoise/tortoise_test.go
```diff
@@ -1351,32 +1353,33 @@ func TestSumVotes(t *testing.T) {
...
-					layerBallots = append(layerBallots, ballot)
+					block.Signature = signer.Sign(block.Bytes())
+					block.Initialize()
+					layerBlocks = append(layerBlocks, block)
+					layerBallots = append(layerBallots, block.ToBallot())
 				}
-				layers = append(layers, layerBallots)
+				layers = append(layers, layerBlocks)
 
+				consensus.processBlocks(ctx, layerBlocks)
```
countvonzero, line 1407: i think this change made the test more confusing.

we are testing `sumVotesForBlock()` here.
we should call `processBlocks()` on the blocks in the diff list, not on the ballots.

tortoise/tortoise.go
dshulyak: GetEpochWeight returns weight of all atxs that are recorded for epoch. so if we have one block that has active set of 2000, and this block is valid, GetEpochWeight will always return at least 2000.

Specifically how it works during partition:
1. there are two separate activesets, each 1000
2. blocks in each partition specify activeset at the start of epoch and because we compute weight based on activeset, weight will remain stable even after partition is removed.
3. let's say partition lasted for an epoch, expected weight will be 1000 before rejoin and each partition will make progress independently
4. after rejoin expected weight will grow to 2000 (GetEpochWeight), but because we are using activeset neither of partitions will be able to make progress without each other weight in votes

countvonzero: ok. thanks for the detailed explanation.

countvonzero (approved):
i am still confused by the test. we should only call processBlocks on blocks that are listed in the 3 diff lists. not on the ballots

tortoise/tortoise_test.go
dshulyak: getDiff is just a helper to compute votes from indexes, what is confusing?

dshulyak: ok i see what you mean, i think because blocks and ballots are aliases in current codebase i made it this way

dshulyak:
bors try

tortoise/tortoise_test.go
countvonzero: but something is wrong with the test if you need to change the code for it to pass.
theoretically the setup should expect `t.BlockLayer` to only contains BlockIDs in the diff lists

dshulyak: i will split up ballots and blocks in test

dshulyak:
bors merge

PR #2971, opened 2021-11-26
[Merged by Bors] - tortoise: local threshold should be based on expected weight from last layer

dshulyak:
## Motivation

It fixes local threshold to use weight from single layer, as per protocol. I messed it up slightly when was refactoring threshold weights.

## Changes
- for local threshold count expected weight only from last processed layer

## Test Plan
unit tests, specifically TestMultiTortoise/unequal_partition_and_rejoin doesn't require weak coin

## TODO

## DevOps Notes


countvonzero (approved):

dshulyak:
bors merge

PR #2972, opened 2021-11-26
[Merged by Bors] - api: stream account update on tx creation

nkryuchkov:
## Motivation

Currently, account update is streamed only when tx is processed (when `stateCurrent` becomes equal to `stateProjected`). However, API also needs to stream information about balance change right after tx creation

## Changes

- stream account update on tx creation

## Test Plan

Unit tests

## TODO

- [X] Explain motivation or link existing issue(s)

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #2976, opened 2021-12-01
[Merged by Bors] - Add API tests missing in #2948 and #2972

nkryuchkov:
## Motivation

#2948 and #2972 were merged without tests, this PR adds them

## Changes

- Add missing tests

## Test Plan

UT

## DevOps Notes


nkryuchkov:
@dshulyak 
I'm going to implement tests from https://github.com/spacemeshos/go-spacemesh/pull/2948#pullrequestreview-813795933 here. I have an issue with implementing the second test. Could you please suggest how to cause blocking of https://github.com/nkryuchkov/go-spacemesh/blob/feature/add-api-tests/api/grpcserver/transaction_service.go#L226? Internally, it calls `SendMsg` which is not blocked until a message is received (https://github.com/grpc/grpc-go/blob/v1.42.0/stream.go#L1398)

api/grpcserver/events.go
```diff
@@ -10,25 +10,43 @@ import (
...
+func consumeEvents(ctx context.Context, in <-chan interface{}) (out <-chan interface{}, bufFull <-chan struct{}) {
+	outCh := make(chan interface{}, subscriptionChanBufSize)
+	bufFullCh := make(chan struct{})
 
 	go func() {
-		defer close(out)
+		defer func() {
+			close(outCh)
+			for range in {
+			}
```
countvonzero: rm?

api/grpcserver/events.go
```diff
@@ -10,25 +10,43 @@ import (
...
+	errRewardsBufferFull     = "rewards buffer is full"
+	errReceiptsBufferFull    = "receipts buffer is full"
+	errActivationsBufferFull = "activations buffer is full"
+	errStatusBufferFull      = "status buffer is full"
+	errErrorsBufferFull      = "errors buffer is full"
+)
+
+func consumeEvents(ctx context.Context, in <-chan interface{}) (out <-chan interface{}, bufFull <-chan struct{}) {
+	outCh := make(chan interface{}, subscriptionChanBufSize)
+	bufFullCh := make(chan struct{})
```
countvonzero, line 27: i'm wary of unbuffered channels. they block on send until there is a receiver. maybe of size 1?

api/grpcserver/globalstate_service.go
```diff
@@ -270,33 +270,53 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
 		select {
+		case <-accountBufFull:
+			log.Info("account buffer is full, shutting down")
+			return status.Error(codes.Canceled, errAccountBufferFull)
+		case <-rewardsBufFull:
+			log.Info("rewards buffer is full, shutting down")
+			return status.Error(codes.Canceled, errRewardsBufferFull)
+		case <-receiptsBufFull:
+			log.Info("receipts buffer is full, shutting down")
+			return status.Error(codes.Canceled, errReceiptsBufferFull)
```
countvonzero, line 307: if all the result is the same when ANY buffer is full, can we simply pass in one channel for any of the go routine to close?
that way, other goroutines can check if that channel is closed and exit?

api/grpcserver/grpcserver_test.go
```diff
@@ -2910,3 +2992,138 @@ func TestGatewayService(t *testing.T) {
...
+type ProjectorMock struct {
+	nonceDiff   uint64
+	balanceDiff uint64
+}
+
+func (p *ProjectorMock) GetProjection(addr types.Address, prevNonce, prevBalance uint64) (nonce, balance uint64, err error) {
+	return prevNonce + p.nonceDiff, prevBalance - p.balanceDiff, nil
+}
+
+func TestEventsCalled(t *testing.T) {
```
countvonzero: TestEventsReceived?

countvonzero (commented):
thanks for adding tests.

api/grpcserver/events.go
dshulyak: i assume this is added to drain the subscription channel until we close the subscription. otherwise we will block the emitter, which is the exact thing we want to prevent. can we pass the subscription and close it when the buffer is full?

api/grpcserver/events.go
dshulyak: it is used as a notification that we terminated subscription and stream needs to exit, e.g there are not sends over this channel

api/grpcserver/events.go
```diff
@@ -10,25 +10,43 @@ import (
...
 
 	go func() {
-		defer close(out)
+		defer func() {
+			close(outCh)
```
dshulyak: if we are using separate buffer this doesn't have to be closed

api/grpcserver/globalstate_service.go
```diff
@@ -270,33 +270,53 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
 		case updatedAccountEvent, ok := <-accountCh:
 			if !ok {
-				log.Info("account stream buffer is full, shutting down")
-				return status.Errorf(codes.Canceled, "stream buffer is full")
+				// we could handle this more gracefully, by no longer listening
```
dshulyak: i think this comment is pretty much a terrible idea, can we remove it everywhere please?
if one subscription gets terminated, we never should continue sending events from other subscriptions

also, if we are not going to close accountCh (and other channels) we don't need to handle `!ok` case here. otherwise you still need to return same error here

dshulyak (changes requested):

api/grpcserver/globalstate_service.go
nkryuchkov: sure, this comment originally existed, so it appeared again after I reverted a commit

api/grpcserver/globalstate_service.go
nkryuchkov: We could, but in this case, logs won't show what buffer became full, which can make debugging harder

dshulyak (approved):

nkryuchkov:
bors merge

api/grpcserver/globalstate_service.go
countvonzero: the one that closes the channel when its buffer is full can log it. i dont see how that's an issue.

PR #2981, opened 2021-12-02
[Merged by Bors] - UCB: change tortoise to use BaseBallot instead of BaseBlock

countvonzero:
## Motivation

6th PR for #2902, change tortoise and test setup to use BaseBallot instead of BaseBlock

## Changes

- simplify layer/fetcher setup
- change tortoise BaseBallot to return types.BallotID
- prepare tortoise test setup to use Proposal

## Test Plan

unittest

## DevOps Notes


dshulyak (approved):
> prepare tortoise test setup to use Proposal

why would tortoise use Proposal? i thought tortoise will use Unified Proposals

nkryuchkov (approved):

countvonzero:
> i think it needs to use array encoder for keys, otherwise there will be only one "layer" and "slots" key in log

the current log looks like this. it essentially is a map of layer to slots. if i use array encoder, i need to create an object encoder for the pair of (layer, slots) which seems more complicated.

```
logger.go:130: 2021-12-03T09:38:54.981+0800	INFO	proposal eligibility calculated	{"epoch_id": 3, "beacon": "51a66b4559", "weight": 1024, "total_weight": 10240, "total_num_slots": 20, "num_layers_eligible": 11, "layers_to_num_proposals": {"layer": 40, "slots": 2, "layer": 42, "slots": 1, "layer": 43, "slots": 1, "layer": 44, "slots": 1, "layer": 46, "slots": 3, "layer": 47, "slots": 1, "layer": 48, "slots": 1, "layer": 49, "slots": 1, "layer": 51, "slots": 1, "layer": 53, "slots": 4, "layer": 57, "slots": 1, "layer": 58, "slots": 1, "layer": 59, "slots": 2}, "name": ""}
```

countvonzero:
> > prepare tortoise test setup to use Proposal
> 
> why would tortoise use Proposal? i thought tortoise will use Unified Proposals

tortoise would not. only the tests needs to create Proposal and convert it to balltos and blocks for the 1st phase
my next PR will convert Block to Proposal for the whole codebase and close #2902. in this PR, Proposal will be typedef'ed to Block inside of tortoise pkg to minimize future code change. Block will be removed from other packages.

after that, i will re-create Block struct for the unified content block that contains only TXs and tortoise will just use that and Ballot. it will not care about Proposal, in code or test.

countvonzero:
bors merge

Issue #2871, opened 2021-10-12
activation: failed contextual validation due to inconsistent state between atx receiver and atx producer


dshulyak:
## Description

Multiple smeshers are producing atx's that are discarded with the next error on receiver side: `last atx is not the one referenced`.
ATX's are indexed using last epoch id, last atx is the one with largest epoch id. It must mean that the receiver has a newer atx than the producer.

## Steps to reproduce
Reproduced on devnet207.
Look for error with M: `atx failed contextual validation` or errmsg: `last atx is not the one referenced`

dshulyak:
It is possible if we sent an ATX, but terminated the node before ATX was stored in the database locally. After restart smesher might be never able to produce ATX that will be accepted by the network again. This is exactly what i am seeing with `39f5d5fd0b32f9f67a727e68d07cad0bdb29d583534e03acd133313fd2da34d1` and `5c83fcbd0878435eb698c451d05e18e9de7973ba4c9cd4ff41752290718bad80` on devnet207. As a solution we should always persist ATX before sending it.

nkryuchkov:
https://github.com/spacemeshos/go-spacemesh/pull/2982 changes the `atx failed contextual validation` log level to WARN, so it won't be found by ERROR level anymore

moshababo:
@dshulyak outdated?

PR #2982, opened 2021-12-02
[Merged by Bors] - Change some error logs to warnings

### LINKED ISSUE #2633: Only critical logs should use ERROR logging level
## Description

Currently, some logged errors are actually not errors, but debugging messages and warnings. Therefore, there're too many errors in Kibana which makes finding real issues more difficult. Also, it's not clear what cases need to be investigated and what don't. 

We should use the ERROR logging level only for critical errors which require immediate investigation and which mean that something went wrong. Other logs should be classified as WARN/INFO/DEBUG.

After only critical errors are classified to the ERROR level, we may consider using notifications for such logs.

nkryuchkov:
## Motivation

Closes #2633

Some error logs are actually not critical and distract when logs are being watched. This PR changes their levels to WARN.

## Changes

- Change some error logs to warnings

## DevOps Notes

### DIFF
--- a/activation/atxdb.go
+++ b/activation/atxdb.go
@@ -186,11 +186,14 @@ func (db *DB) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error {
 		log.FieldNamed("atx_node_id", atx.NodeID),
 		atx.PubLayerID)
 	if err := db.ContextuallyValidateAtx(atx.ActivationTxHeader); err != nil {
-		db.log.WithContext(ctx).With().Error("atx failed contextual validation",
+		db.log.WithContext(ctx).With().Warning("atx failed contextual validation",
 			atx.ID(),
 			log.FieldNamed("atx_node_id", atx.NodeID),
 			log.Err(err))
 		// TODO: Blacklist this miner
+		// NOTE(nkryuchkov): Sometimes, this case may be seen in devnet logs
+		// with error: "last atx is not the one referenced",
+		// perhaps we need to add a threshold for blacklisting.
 	} else {
 		db.log.WithContext(ctx).With().Info("atx is valid", atx.ID())
 	}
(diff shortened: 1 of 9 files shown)


activation/atxdb.go
```diff
@@ -186,11 +186,14 @@ func (db *DB) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error {
...
 	if err := db.ContextuallyValidateAtx(atx.ActivationTxHeader); err != nil {
-		db.log.WithContext(ctx).With().Error("atx failed contextual validation",
+		db.log.WithContext(ctx).With().Warning("atx failed contextual validation",
 			atx.ID(),
 			log.FieldNamed("atx_node_id", atx.NodeID),
 			log.Err(err))
 		// TODO: Blacklist this miner
+		// NOTE(nkryuchkov): Sometimes, this case may be seen in devnet logs
+		// with error: "last atx is not the one referenced",
+		// perhaps we need to add a threshold for blacklisting.
```
dshulyak: it is very likely due to the bug in the codebase https://github.com/spacemeshos/go-spacemesh/issues/2871 . i don't think that these todo's about blacklisting should be implemented until we know that software always works as intended.

right now blacklisting will do more harm than good. i would just remove todo and note, i think it is more confusing then useful

dshulyak (approved):

activation/atxdb.go
nkryuchkov: I think removing only this to-do won't give any benefits as we have similar to-do's in the codebase. Maybe something like `TODO: Consider blacklisting after we ensure that this case always works as intended (see #2871)`?

dshulyak: i don't think that this todo make any sense. you can't just go and fix this todo, it needs to be designed properly for the whole project. in my opinion you just added more unnecessary comments to an already confusing todo, that's why i suggested to remove it altogether

nkryuchkov: > you can't just go and fix this todo, it needs to be designed properly for the whole project

Well, that's true. I'll remove it then

nkryuchkov:
bors merge

PR #2990, opened 2021-12-07
[Merged by Bors] - UCB: hare eligibility oracle uses ballots for hare active set

countvonzero:
## Motivation

7th PR for #2902, hare eligibility oracle uses ballots for hare active set

## Changes

- hare eligibility uses ballots to calculate hare active set
this removes code for BFS on the ATXs on blocks and its diff list and further remove the deps from activaction db->mesh db
 
- change hare oracle test to use gomock
 
## Test Plan

unit test

## DevOps Notes


dshulyak:
@countvonzero it is not completely obvious what was the reason to find weights starting from the set of valid blocks, but new change looks rather different. do you know why it was necessary to do that thing the way it was done and why new solution works?

countvonzero:
> @countvonzero it is not completely obvious what was the reason to find weights starting from the set of valid blocks, but new change looks rather different. do you know why it was necessary to do that thing the way it was done and why new solution works?

there is nothing new here. in UCB, there is no contextual validity to ballots. every ballot is valid. 

the old code look for contextual valid blocks in those layers in the safe range, and add all the blocks in every diff list in those blocks. and find all the ATXs referenced in those blocks that fall in the safe range.

the new code simply have a flat structure (just ballot). and collect the active set from ballots in the safe range layers.

the overall goal is the find the hare active set, which is defined as the set of ATXIDs in the safe layer range. i didn't change the safe layer range algo.

countvonzero:
bors try

dshulyak:
> there is nothing new here. in UCB, there is no contextual validity to ballots. every ballot is valid.

so for instance ballots with wrong beacons are considered adversarial for a certain number of layers. i don't know if that attack is relevant in the context of hare, but this is an obvious difference if compared with the previous solution.

dshulyak (approved):

countvonzero:
> > there is nothing new here. in UCB, there is no contextual validity to ballots. every ballot is valid.
> 
> so for instance ballots with wrong beacons are considered adversarial for a certain number of layers. i don't know if that attack is relevant in the context of hare, but this is an obvious difference if compared with the previous solution.

i think hare active set affects hare eligibility. i will double check with research to make sure.

btw, i think there may not be a need to get the ref ballot if the ballot has no active set. the idea is the safe layer range should cover enough ref ballots.

countvonzero:
> > there is nothing new here. in UCB, there is no contextual validity to ballots. every ballot is valid.
> 
> so for instance ballots with wrong beacons are considered adversarial for a certain number of layers. i don't know if that attack is relevant in the context of hare, but this is an obvious difference if compared with the previous solution.

as discussed in research channel, i'll filter out ATXs that published ballots with bad beacon. thanks for catching this. there is bad ballots after all :D

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors try

countvonzero:
bors merge

Issue #2985, opened 2021-12-03
tortoise: switch from healing to verifying


dshulyak:
After the healing period, we want to switch to verifying tortoise for efficiency reasons. If we can find a good ballot from a ballots that we keep in memory, and the amount of exceptions is within the limit then we don't need to do anything. We just select this ballot as a base and encode our opinion vector as exceptions.

If there are no good ballots that fit specified criteria we will have to manually mark some ballots as good. Those should be ballots that do not contradict our own opinion.

So if this is our definition of `isGood`:

```go
func isGood(s VerifyingTortoise, ballot Ballot) bool {
    if _, exist := s.GoodBallots[ballot.Base]; !exist {
        return false
    }
    for vote, sign := range ballot.Votes {
        if s.Validity[vote] != sign {
            return false
        } 
    }
    return true
}
```

`canBeGood` will be a simplified version of `isGood`.

```go
func canBeGood(s VerifyingTortoise, ballot Ballot) bool {
    for vote, sign := range ballot.Votes {
        if s.Validity[vote] != sign {
            return false
        } 
    }
    return true
}
```

So tentatively if we receive a layer after a period of self-healing where all ballots can be good, and the base ballot of those ballots can be good we mark all of them as good.

Issue #2980, opened 2021-12-02
tortoise: expected weight during rerun should be based on current layer


dshulyak:
Let's assume that the current layer is 100, when we process layer 10 during rerun expected weight should be for the whole difference between 100 - 10.

This will enable rerun to count votes that weren't counted when layer was verified initially, and fix layers that were verified erroneously, for example, due to the dishonest majority.

PR #2991, opened 2021-12-07
[Merged by Bors] - tortoise: verifying tortoise implementation

### LINKED ISSUE #2979: verifying tortoise implementation
## Overview

In this table we store marging differences for all blocks and validity. For layers within hdist this is based on input vector, outside of hdist it is based on contextual validity.

Start with two arbitrary blocks Q, W. Q is valid, W is invalid. Q and W are both in layer 1. Threshold is 60%.

| blocks             | Q   | W   |     |     |     |
| ------------------ | --- | --- | --- | --- | --- |
| validity           | 1   | -1  |     |     |     |
| margin differences | 0   | 0   |     |     |     |

At next layer we receive ballots A, S, D that vote for Q and against W. Total expected vote weight is 30, each ballot weight is 10. And one new block E, which is also confirmed by input vector. Q and W are now confirmed and the layer 1 is verified.

| blocks             | Q     | W     | E   |     |     |
| ------------------ | ----- | ----- | --- | --- | --- |
| validity           | 1     | -1    | 1   |     |     |
| margin differences | 30/30 | 30/30 | 0   |     |     |

Next layer we receive ballots F and G. Both are voting for E. Only F is good, it provides 10 weight, it is not enough to verify layer since threshold is 18 (30 * 0.6). We also receive block R, confirmed by the input vector.

| blocks             | E     | R   |     |
| ------------------ | ----- | --- | --- |
| validity           | 1     | 1   |     |
| margin differences | 10/30 | 0   |     |

Receive ballots H, J, K. All are good. Add 30 weight in votes.

| blocks             | E     | R     |     |
| ------------------ | ----- | ----- | --- |
| validity           | 1     | 1     |     |
| margin differences | 40/60 | 30/30 |     |

Once E is verified, we decrease total expected weight by weight of layer 2, and then verify R.

## Implementation

Code is simplified and provided for clarity, actual implementation may differ.

```go
type VerifyingTortoiseState struct {
    // optimized margin differences stores total expected weight - voting weight
    MarginDifferences map[BlockID]*big.Float
    // Validity stores opinion according to input vector within hdist
    // before verified layer it stores contextual validity.
    // if outside hdist opinion is abstain - verifying tortoise can't make progress
    Validity          map[BlockID]int8
    LayerBlocks       map[LayerID][]BlockID
    GoodBallots       map[BallotID]bool
}

func addBlocks(s VerifyingTortoiseState, lid LayerID, bids []BlockID) {
    s.LayerBlocks[lid] = bids
    for _, bid := range bids {
        s.Validity[bid] = -1
    }
}

func addInputVector(s VerifyTortoiseState, vector []BlockID) {
    for _, bid := range vector {
        s.Validity[bid] = 1
    }
}

// Ballot simplified.
type Ballot struct {
    ID, Base   BallotID
    // Votes store both explicit and implicit votes.
    Votes  map[BlockID]int8
    // ATX weight / number of ballots
    Weight *big.Float
}

// get expected weight based on activations
func getExpectedWeight(LayerID) *big.Float

func isGood(s VerifyingTortoise, ballot Ballot) bool {
    if _, exist := s.GoodBallots[ballot.Base]; !exist {
        return false
    }
    for vote, sign := range ballot.Votes {
        if s.Validity[vote] != sign {
            return false
        } 
    }
    return true
}

func addBallots(s VerifyingTortoise, ballots []Ballot) (*big.Float) {
    sum := 0
        if !isGood(s, ballot) {
            continue
        }
        s.GoodBallots[ballot.ID] = true
        sum += ballot.Weight
    } 
    return sum
}

func verifyLayers(s VerifyingTortoiseState, verified, last LayerID, sum *big.Float) LayerID {
    for lid := verified + 1; lid < last; lid++ {
        for _, bid := range s.LayerBlocks[lid] {
            s.MarginDifferences[bid] += sum
            if s.MarginDifferences[bid] <= getExpectedWeight(lid) * GLOBAL_TRESHOLD {
                return lid - 1
            }
        }
    }
    return last - 1
}

func processLayer(s VerifyingTortoiseState, lid, verified LayerID, blocks, vector []BlockID, ballots []Ballot) LayerID {
    sum := addBallots(s, ballots)
    verified := verifyLayers(s, verified, lid, sum)

    addBlocks(s, lid, blocks)
    addVector(s, vector)
    return verified
}
```

## Questions

- if block is neutral outside of hdist, does it mean that verifying tortoise failed and we suppose to enter healing?

### LINKED ISSUE #2987: tortoise: base block selection compatible with verifying tortoise
As we don't need to keep full voting matrix for verifying tortoise we will need to simplify base block selection in case if there are good ballot in early layers.

- find earliest good ballot, add exceptions based on local opinion
- if there is no good ballot or there are only very old good ballots, order blocks by the difference with local opinion and then by layer

```go
type State struct {
    Validity map[BlockID]int8
    Votes    map[BallotID]map[BlockID]int8
    Good     map[BallotID]struct{}
    Ballots  map[LayerID][]BallotID
    Blocks   map[LayerID][]BlockID

    Last, Processed, Evicted LayerID
}

func getSign(opinion map[BlockID]int8, bid BlockID) int8 {
    if opinion == nil {
        return -1
    }
    sign, exist := opinion[bid]
    if !exist {
        return -1
    }
    return sign
}

func encodeVotes(s State, opinion map[BlockID]int8, lid LayerID) map[BlockID]int8 {
    rst := map[BlockID]int8{}
    for ; lid <= s.Last; lid++ {
        for _, bid := range s.Blocks[lid] {
            local := s.Validity[bid]
            // TODO use weakcoin for old layers here
            if local != getSign(opinion, bid) {
                rst[bid] = local
            }
        }
    }
    return rst
}

func chooseGoodBase(s State) (BallotID, map[BlockID]int8) {
    for lid := s.Processed; lid > s.Evicted; lid-- {
        for _, bid := range s.Ballots[lid] {
            if _, exist := s.Good[bid] {
                return bid, encodeVotes(s, nil, lid)
            }
        }
    }
    return Empty, nil
}

// get least bad returns ballot with the largest disagreement layer
func getLeastBad(s State) (BallotID, LayerID)

func chooseBase(s State) (BallotID, map[blockID]int8) {
    base, votes := chooseGoodBase(s)
    if base != Empty {
        return base, votes
    }
    // this should be a list, first choice may have more then limit exceptions
    bid, lid := getLeastBad(s)
    return base, encodeVotes(s, s.Votes[bid], lid)
}

### PR DESCRIPTION
**Pull request by dshulyak**
## Motivation

closes https://github.com/spacemeshos/go-spacemesh/issues/2979 
closes https://github.com/spacemeshos/go-spacemesh/issues/2987 

## Changes
- verifying tortoise implemented in a way that doesn't require vote counting for every block. when new set of ballots is received we check if they vote consistently with local opinion (including implicit votes) and if so we count weight from good ballots towards previous layers. if previous layer crosses threshold of expected weight (defined as local + global thresholds) - the blocks in the layer are verified according to the local opinion.
- if outside of hdist local opinion remains undecided - tortoise will switch into full mode and will remain in full mode. the previous solution for switching into verifying from full mode was relying on vote counting in verifying tortoise, and it was removed. the switch will be implemented in https://github.com/spacemeshos/go-spacemesh/issues/2985 according to the discussion with research. note that full tortoise is not a bottleneck anyway, it works very fast. the only reason why we may need verifying tortoise is for rerun that considers weight for long distances.
- specifically for rerun we need to know the latest historically verified layer, so that we can form an opinion on blocks that are long ago verified. otherwise, nodes will switch into full mode. previously it was not needed because local opinion for verifying tortoise was based on vote counting.
- also removed redundant tests as they are not needed, and refactoring them wastes time
- confidence interval is repurposed and used as a way to prevent switching into full mode prematurely. for example if the confidence interval is 1 and tortoise can't verify the layer using good ballots from the next layer - tortoise will switch into full mode immediately. instead we will set confidence to 10, which will make it less likely that we switch before actual failure. we will have to find better heuristic later on when rerun will consider large layer distance for weight, but for now this is good enough. 

## Test Plan
new ut and existing tests.
also replayed state from devnet210, it is recomputed in 20s and stays in verifying mode.

## TODO

## DevOps Notes

### CODE REVIEW

**Comment by lrettig** [comment]
Is this ready for review?

**Comment by dshulyak** [comment]
> Is this ready for review?

@lrettig i was going to create a separate pr that removes persistent state from tortoise, and explain motivation for some changes. but beside that the change is ready

File: tortoise/state.go [hunk 1]
```diff
@@ -18,55 +44,28 @@ type state struct {
...
-	}
-	s.LastEvicted = windowStart.Sub(1)
-	return nil
+// fullState contains state necessary to run full (self-healing) tortoise.
+//
+// as long as the verifying tortoise makes progress this state can remain empty.
+// if verifying tortoise get stuck we need to have data for two use cases:
+// - select a base ballot that is the least bad, requires votes since genesis,
+//   for now we keep votes only in a small window and only after verifying tortoise is stuck
+//   relevant only if we can't find a good ballot that can fit all votes int the exception limit
```
countvonzero: :s/int/in

tortoise/state.go
```diff
@@ -18,55 +44,28 @@ type state struct {
 	badBeaconBallots map[types.BallotID]struct{}
 
 	// epochWeight average weight of atx's that target keyed epoch
```
countvonzero: :s/average weight/average weight per layer/

tortoise/tortoise.go
```diff
@@ -1108,151 +890,95 @@ func (t *turtle) ballotFilterForHealing(logger log.Log) func(types.BallotID) boo
...
-	ballotID types.BallotID,
-) error {
-	block, err := mdb.GetBlock(types.BlockID(ballotID))
+	epochWeight, _, err := atxdb.GetEpochWeight(eid)
 	if err != nil {
-		return fmt.Errorf("failed to load %s from db: %w", ballotID, err)
-	}
-	if block.ATXID == *types.EmptyATXID {
-		return nil
+		return layerWeight, fmt.Errorf("epoch weight %s: %w", eid, err)
```
countvonzero: return weightFromUint64(0)?

tortoise/tortoise.go
```diff
@@ -1108,151 +890,95 @@ func (t *turtle) ballotFilterForHealing(logger log.Log) func(types.BallotID) boo
...
+		expected, err := proposals.GetNumEligibleSlots(targetWeight, total, layerSize, layersPerEpoch)
 		if err != nil {
-			return nil, fmt.Errorf("unable to compute number of eligibile ballots for atx %s", ballot.AtxID)
+			return weightFromUint64(0), fmt.Errorf("unable to compute number of eligibile ballots for atx %s", ballot.AtxID)
 		}
-		rst := new(big.Float).SetUint64(weight)
-		rst.Quo(rst, new(big.Float).SetUint64(uint64(expected)))
-		return rst, nil
+		rst := weightFromUint64(targetWeight)
+		return rst.div(weightFromUint64(uint64(expected))), nil
```
countvonzero, line 933: check expected > 0?

tortoise/tortoise.go
```diff
@@ -1108,151 +890,95 @@ func (t *turtle) ballotFilterForHealing(logger log.Log) func(types.BallotID) boo
...
 	}
 	if ballot.RefBallot == types.EmptyBallotID {
-		return nil, fmt.Errorf("empty ref ballot and no epoch data on ballot %s", ballot.ID())
+		return weightFromUint64(0), fmt.Errorf("empty ref ballot and no epoch data on ballot %s", ballot.ID())
 	}
-	weight, exist := weights[ballot.RefBallot]
+	rst, exist := weights[ballot.RefBallot]
 	if !exist {
-		return nil, fmt.Errorf("ref ballot %s for %s is unknown", ballot.ID(), ballot.RefBallot)
+		return rst, fmt.Errorf("ref ballot %s for %s is unknown", ballot.ID(), ballot.RefBallot)
```
countvonzero: return weightFromUint64(0)

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+		for _, ballot := range part {
+			ballotWeight, err := computeBallotWeight(t.atxdb, t.ballotWeight, ballot, t.LayerSize, types.GetLayersPerEpoch())
+			if err != nil {
+				return nil, err
+			}
+			t.ballotWeight[ballot.ID()] = ballotWeight
+			t.ballotLayer[ballot.ID()] = ballot.LayerIndex
+
+			// TODO(dshulyak) this should not fail without terminating tortoise
+			t.ballotHasGoodBeacon(ballot, t.logger)
```
countvonzero: i don't understand the comment. do you mean we should fail tortoise if we fail to get beacon?

in any case, i think this method should change to change to something like
`t.markBeaconWithBadBallot` and not return anything.

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+		processed   = t.processed
+		min         = t.last // just max value
+		refs, other []*types.Ballot
+	)
+
+	for _, ballot := range original {
+		if ballot.EpochData != nil {
+			refs = append(refs, ballot)
+		} else {
+			other = append(other, ballot)
```
countvonzero, line 580: just to understand. is this separation for the case where ref ballot and its following ballots to be in the same layer?

tortoise/state.go
```diff
@@ -1,15 +1,41 @@
...
+	// historicallyVerified matters only for local opinion for verifying tortoise
+	// during rerun. for live tortoise it is identical to the verified layer.
+	historicallyVerified types.LayerID
+	// last processed layer
+	processed types.LayerID
+	// last evicted layer
+	evicted types.LayerID
+
+	blocks      map[types.LayerID][]types.BlockID
+	ballots     map[types.LayerID][]types.BallotID
```
countvonzero, line 36: are these mapping necessary? is there a performance downgrade if you get them from mesh?

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+			}
+			for _, bid := range ballot.ForDiff {
+				votes[bid] = support
+			}
+			for _, bid := range ballot.NeutralDiff {
+				votes[bid] = abstain
+			}
+			for _, bid := range ballot.AgainstDiff {
+				votes[bid] = against
+			}
```
countvonzero, line 616: is this necessary since every block after baselid starts with against?

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
 			}
 		}
-		logger.With().Info("verified candidate layer", log.Named("new_verified", t.Verified))
+		t.full.votes[ballot.id] = votes
 	}
-	return nil
+}
+
+// the idea here is to give enough room for verifying tortoise to complete. if verifying tortoise fails to verify
+// a layer within last hdist layers or within zdist + confidence param (whichever is larger) we switch to full vote counting.
```
countvonzero: i thought we need to make sure hdist > zdist+confidence param.

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+//
+// verifying tortoise is a fastpath for full vote counting, that works if everyone is honest and have good synchrony.
+// as for the safety and livenesss both protocols are identical.
```
countvonzero: liveliness

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+func (t *turtle) isHealingEnabled() bool {
```
countvonzero: i think the name should be `shouldTriggerSelfHealing` or something similar. healing should always be enabled.

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+	lid := t.verified.Add(1)
+	return lid.Before(t.layerCutoff()) &&
+		t.last.Difference(lid) > t.Zdist &&
+		// unlike previous two parameters we count confidence interval from processed layer
+		// processed layer is significantly lower then the last layer during rerun.
+		// we need to wait some distance before switching from verifying to full during rerun
+		// as previous two parameters will be always true even if single layer is not verified.
```
countvonzero, line 670: in terms of protocol, i don't understand why we should ever consider `t.processed` when we want to consider enter healing mode.
why not simply forbid rerun to enter healing mode?

dshulyak:
bors try

tortoise/verifying.go
```diff
@@ -0,0 +1,143 @@
...
+
+	goodBallots map[types.BallotID]struct{}
+	// weight of good ballots in the layer
+	layerWeights map[types.LayerID]weight
+	// total weight of good ballots from verified + 1 up to last
+	totalWeight weight
+}
+
+func (v *verifying) processLayer(logger log.Log, lid types.LayerID, ballots []tortoiseBallot) {
+	counted, count := v.sumGoodBallots(logger, ballots)
```
countvonzero: counted and count are a bit too close. can you rename counted to goodWeight or something?

tortoise/verifying.go
```diff
@@ -0,0 +1,143 @@
...
+	localThreshold := weightFromUint64(0)
+	// TODO(dshulyak) expected weight for local threshold should be based on last layer.
+	localThreshold = localThreshold.add(v.epochWeight[v.processed.GetEpoch()])
+	localThreshold = localThreshold.fraction(v.LocalThreshold)
+
+	for lid := v.verified.Add(1); lid.Before(v.processed); lid = lid.Add(1) {
+		// TODO(dshulyak) expected weight must be based on the last layer.
+		threshold := computeExpectedWeight(v.epochWeight, lid, v.processed)
+		threshold = threshold.fraction(v.GlobalThreshold)
+		threshold = threshold.add(localThreshold)
```
countvonzero, line 56: why does it need to add local threshold?

tortoise/verifying.go
```diff
@@ -0,0 +1,143 @@
...
+		// 0 - there is not enough weight to cross threshold.
+		// 1 - layer is verified and contextual validity is according to our local opinion.
+		if votingWeight.cmp(threshold) == abstain {
+			layerLogger.With().Warning("candidate layer is not verified. voting weight is lower than the threshold")
+			return lid.Sub(1)
+		}
+
+		// if there is any block with neutral local opinion - we can't verify the layer
+		// if it happens outside of hdist - protocol will switch to full tortoise
+		for _, bid := range v.blocks[lid] {
```
countvonzero, line 78: why is this check necessary when you have the check on line 70 `if votingWeight.cmp(threshold) == abstain {`

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
-					return fmt.Errorf("saving contextual validity for %s: %w", blk, err)
-				}
+func (t *turtle) keepVotes(ballots []tortoiseBallot) {
+	for _, ballot := range ballots {
+		votes := ballot.votes
+		basevotes := t.full.votes[ballot.base]
+		if len(basevotes) == 0 {
+			t.logger.With().Warning("base votes are empty", ballot.id, log.Stringer("base_ballot", ballot.base))
+		}
+		for vote, sign := range basevotes {
```
countvonzero: rename vote to block? vote and sign for me is the same thing

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+	return nil
+}
 
-		if contextualValidity != nil {
-			for blk, v := range contextualValidity {
-				if err := t.bdp.SaveContextualValidity(blk, candlid, v); err != nil {
-					logger.With().Error("error saving contextual validity on block", blk, log.Err(err))
-					return fmt.Errorf("saving contextual validity for %s: %w", blk, err)
-				}
+func (t *turtle) keepVotes(ballots []tortoiseBallot) {
```
countvonzero, line 638: i wonder whether we can keep a separate eviction policy for base ballots. most ballot in a layer should have the same base ballot. and we only evict base ballots when there is no references to it anymore.

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
-		// TODO: abstain only for entire layer at a time, not for individual blocks (optimization)
-		//   see https://github.com/spacemeshos/go-spacemesh/issues/2674
-		if !globalOpinionDecided {
-			logger.With().Warning("global opinion on block is abstain, cannot verify layer",
-				bid,
-				log.Stringer("global_opinion", globalOpinionOnBlock),
-				log.Stringer("local_opinion", localOpinionOnBlock))
-			return nil, nil
+		if t.verified.After(previous) {
+			if err := persistContextualValidity(logger,
```
countvonzero, line 499: why isn't this done inside t.verifying.verifyLayers()?

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+			}
+		}
+		t.updateHistoricallyVerified()
+		if t.verified == t.processed.Sub(1) {
+			return nil
+		}
+	}
+	// keep votes starting from the layer when verifying tortoise failed to make progress
+	// for example, if verifying failed while counting ballots from layer 13 we will
+	// most likely have to use those ballots later in full mode.
```
countvonzero, line 515: the comment here doesn't seem to match code in `keepVotes()`

say verified is 10 and processed is 12. we are only keeping votes for layer 12 but not 11. is that ok?

tortoise/tortoise.go
```diff
@@ -636,268 +449,264 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) ([]byte,
...
+			t.mode = fullMode
+		}
+		previous := t.verified
+		healed := t.healingTortoise(ctx, logger)
+		if !healed {
+			return nil
+		}
+		// TODO(dshulyak) refactor full tortoise to be consistent with verifying.
+		// update local opinion in memory and persist afterwards.
+		t.updateHistoricallyVerified()
```
countvonzero, line 537: i don't grasp the idea of historically verified. it's updated in both verifying and full mode, how is this different from verified then?

countvonzero (commented):
mostly questions and nits for readability.

tortoise/state.go
dshulyak: i think it is critical for correctness and verifiability to implement tortoise as a state machine. so i am minimizing implicit inputs to the tortoise, in fact the only implicit input that remains is hare input vector. so performance here wasn't a concern

tortoise/verifying.go
dshulyak: we referred to this weight as counted in all discussions. so i wanted to keep it as is, because there also will be a variable with uncounted weight. i will probably rename count to n or smth

tortoise/verifying.go
dshulyak: it was described by Tal in https://community.spacemesh.io/t/uncounted-weight-and-the-global-threshold/228

tortoise/verifying.go
dshulyak: on line 70 i check if i have enough weight to verify the layer, but the blocks in a layer can be undecided if the layer is still within zdist

tortoise/tortoise.go
dshulyak: i think it matches, maybe i need to improve wording though
layer 12 verifies layer 11, so if verifying failed we will use votes from layer 12 later in full mode.

tortoise/tortoise.go
dshulyak: my intention is to separate code that is required for consensus correctness from the code that loads/saves state on to the disk.

also, not related to this pr, but i think that it will be safer not to persist contextual validity in tortoise, but instead return it as in memory object to the caller, and let the caller persist at the right time.

tortoise/tortoise.go
dshulyak: it is different during rerun. we are using verified to track progress of the rerun, but local opinion must be based on historically verified.

tortoise/tortoise.go
dshulyak: yes. it was already in the codebase, i just refactored it a bit

tortoise/tortoise.go
dshulyak: yes this is what i meant. will change the name

tortoise/tortoise.go
dshulyak: this is still necessary if we have exceptions before baselid

tortoise/tortoise.go
dshulyak: i intend to change the way we store votes in https://github.com/spacemeshos/go-spacemesh/pull/2994

tortoise/tortoise.go
dshulyak: i think the conclusion from that discussion is that we set hdist to some large value, zdist to the value when hare is expected to terminate (e.g. that computation with max iterations and max instances). and it leaves confidence param unnecessary and i will use it for different purpose

tortoise/tortoise.go
dshulyak: i don't think that works. the rerun is more likely to enter self-healing because we may learn new state (e.g. different votes and atxs)

tortoise/tortoise.go
dshulyak: i can return just weight{}

tortoise/tortoise.go
dshulyak: the intuition is that we need to have some heuristic to delay when we switch into full mode, during live tortoise this delay is hdist. during rerun, hdist is irrelevant as we are long past hdist. i want to discuss it on the next review with research. confidence param works for now, but maybe there are better solutions

tortoise/tortoise.go
countvonzero: before a rerun starts, `consensus.historicallyVerified` is set to the normal run's `historicallyVerified`. does that mean that rerun will form local opinions based on the normal run's historicallyVerified? i don't understand why rerun cannot form local opinions based on its own value of `verified`

tortoise/state.go
countvonzero: how do you plan to handle late ballots in the future?

tortoise/tortoise.go
dshulyak: what is meant is usually referred as liveness, e.g. it is about making progress

tortoise/tortoise.go
dshulyak: i renamed to canUseFullMode

tortoise/tortoise.go
countvonzero: in node.go, i added a check for zdist+conf_param to be at least > max_hare_iteration (in terms of time).

your comments (whichever is larger) implies zdist can be larger than hdist. i don't think that's the conclusion.

tortoise/tortoise.go
countvonzero: > i don't think that works. the rerun is more likely to enter self-healing because we may learn new state (e.g. different votes and atxs)

but it's just as easy for rerun to signal stuck-ness to tortoise and enter self-healing mode from there.
is it possible that verifying tortoise enter healing while rerun also enter healing?

self-healing however should get the full state from mesh for correctness?

tortoise/tortoise.go
```diff
@@ -1017,83 +812,70 @@ func (t *turtle) sumVotesForBlock(
...
+		logger.With().Info("healing: expected voting weight at the layer",
+			log.Stringer("expected_weight", expected),
+			log.Stringer("threshold", threshold),
+			candidateLayerID,
+		)
 
 		// record the contextual validity for all blocks in this layer
-		for _, blockID := range layerBlockIds {
-			logger := logger.WithFields(log.Named("candidate_block_id", blockID))
+		for _, blockID := range t.blocks[candidateLayerID] {
```
countvonzero, line 845: i am still not understanding why we don't reload all blocks from mesh

tortoise/tortoise.go
dshulyak: it is handled in GetNumEligibleSlots, so that if expected is 0 and total weight is not 0 - it is always at least 1 ballot. i guess i will just error for safety

tortoise/tortoise.go
dshulyak: why do we need to load them from mesh? we already loaded them in processLayer

it is a big mistake to have implicit inputs in the code that has a high bar for correctness. if we load more blocks here it means that the rest of the tortoise state won't know about them (e.g. blockLayer, votes and son on). so by seemingly innocent change, it introduces inconsistencies that make it hard to reason about tortoise correctness.

dshulyak: by processing them explicitly i know that the state is consistent and there will no weird stuff when block is added in one place, but not added in another

tortoise/state.go
dshulyak: exactly the same way they are handled now. the only difference is that tortoise rerun will be triggered explicitly instead of trigger it on time.

tortoise/tortoise.go
dshulyak: where it follows from my comment that zdist can be larger? Tal mentioned in the discussion on slack that hdist > zdist, and we have this check during tortoise initialization

tortoise/tortoise.go
dshulyak: > but it's just as easy for rerun to signal stuck-ness to tortoise and enter self-healing mode from there.
is it possible that verifying tortoise enter healing while rerun also enter healing?

i am confused by this question. the comparison between verifying tortoise and rerun doesn't make sense. there is live tortoise (just a normal workflow) and rerun, and there is verifying tortoise and full tortoise. live tortoise can be in both modes and rerun can be in both modes.

> self-healing however should get the full state from mesh for correctness?

why? it has all the state it needs to make progress

tortoise/tortoise.go
dshulyak: so how rerun should work: 
last layer is kept according to the clock
rerun start from genesis and processes all layers up to the last
important: expected weight is set based on the last layer (we don't do it currently https://github.com/spacemeshos/go-spacemesh/issues/2980). it is set this way because we want to count votes that we didn't count during normal flow. it prevents concentration attacks, or layers that were verified with dishonest majority. it was discussed on last call.
because of the previous heuristic - verified layer will be stuck long in the past. but we also want to use verifying tortoise, this is why we base our local opinion on what we initially verified.

note that the condition i described does happen even now (on data from devnet210, but layer is not stuck far in the past, but may get stuck for 1-2 layers)

tortoise/tortoise.go
dshulyak: > is it possible that verifying tortoise enter healing while rerun also enter healing?

you meant live tortoise enters full mode and later rerun enters full mode when it processes same layers? yes, in fact it is very likely to happen. for example if there were many ballots with incorrect beacon during live tortoise, and later during rerun verifying tortoise won't make progress as well

tortoise/tortoise.go
dshulyak: that check that you mentioned uses only zdist `app.Config.LayerDurationSec*int(app.Config.Tortoise.Zdist) <= maxHareLayerDurationSec` , which i think is correct

countvonzero: > where it follows from my comment that zdist can be larger? Tal mentioned in the discussion on slack that hdist > zdist, and we have this check during tortoise initialization

the comment below to me hints that (zdist+confidence param) can be larger than hdist
> a layer within last hdist layers or within zdist + confidence param (whichever is larger)

tortoise/tortoise.go
countvonzero: yes. i meant that live tortoise entering full mode, and just a little bit later rerun enters full mode.

since full-mode is expensive and should be avoided, why not follow the heuristic.
when live tortoise gets stuck
- first wait for the latest rerun to finish (without entering full mode). if it made progress, swap rerun state.
- if the latest rerun also get stuck, then stop future reruns and enter full mode.
- resume live tortoise and rerun after exiting full mode

tortoise/tortoise.go
countvonzero: ok. i think i understand better now. whatever missing data in the normal run would be reloaded during rerun.

countvonzero (approved):

tortoise/tortoise.go
dshulyak: ah sorry, thats about this comment, i reworded after this discussion has started

tortoise/tortoise.go
dshulyak: > first wait for the latest rerun to finish (without entering full mode). if it made progress, swap rerun state.

i don't understand what is expected to get fixed by this. it is expected that full mode will fix our local opinion and eventually we will be able to switch into verifying mode, either naturally or forcefully (as was discussed on last call https://github.com/spacemeshos/go-spacemesh/issues/2985)

> if the latest rerun also get stuck, then stop future reruns and enter full mode.

it is not full mode OR verifying tortoise & rerun. if we can't do rerun in verifying mode - we will have to do it full mode, because of the same concerns, dishonest majority at the time of live-run and more uncounted weight from missing atxs.

dshulyak:
bors merge

PR #2993, opened 2021-12-09
[Merged by Bors] - tortoise: reload state in background when node is started

dshulyak:
## Motivation

Removing persistent state from tortoise greatly simplifies code and further changes. It is also fast enough to recover it from original databases (mesh and atx) when we start the node.

on cloud nodes, it takes 2minutes to load 10 000 layers. on a node with a faster hardware, specifically disk, it takes 50s. verifying tortoise, from mentioned pr, loads same amount of layers in 20s.

prereq for https://github.com/spacemeshos/go-spacemesh/pull/2991

## Changes
- load state in background when node is started
- wait for tortoise to become ready before starting sync

## Test Plan
unit and manual tests

## TODO

## DevOps Notes


tortoise/algorithm.go
```diff
@@ -110,37 +113,43 @@ func New(db database.Database, mdb blockDataProvider, atxdb atxDataProvider, bea
...
+	if needsRecovery {
+		t.trtl.Processed = t.cfg.MeshProcessed
+		// TODO(dshulyak) last should be set according to the clock.
+		t.trtl.Last = t.cfg.MeshProcessed
+
+		t.logger.Info("loading state from disk. make sure to wait until tortoise is ready",
+			log.Stringer("last_layer", t.cfg.MeshProcessed),
+		)
+		t.eg.Go(func() error {
+			t.ready <- t.rerun(ctx)
```
countvonzero, line 139: can you document in the PR how long does this take with mesh growth? thanks

mesh/mesh.go
```diff
@@ -159,21 +157,10 @@ func NewRecoveredMesh(ctx context.Context, db *DB, atxDb AtxDB, rewardConfig Con
 	if err != nil {
 		logger.With().Panic("failed to load state for layer", msh.LatestLayerInState(), log.Err(err))
 	}
-	// in case we load a state that was not fully played
-	nextLayer := msh.LatestLayerInState().Add(1)
-	lastComplete := msh.trtl.LatestComplete()
-	if nextLayer.Before(msh.trtl.LatestComplete()) {
```
countvonzero, line 165: say for whatever reason, the rerun on startup yields a different verified layer before the restart. and the verified layer is different from the layer applied layer in state. i think we need to double check that too.

countvonzero (approved):
so much nicer.

mesh/mesh.go
dshulyak: that would be a standard revert flow. this can be expected because we may load more data when node is re-started

dshulyak:
bors merge

mesh/mesh.go
countvonzero: but how about the verified layer is higher than before the restart? say before the restart the state is applied up to N. but after restart, rerun verified up to N+10. how will the state catch up and apply TXs in the extra 10 layers? when the next hare call `HandleIncomingLayer`?

dshulyak: for example processed = 20, verified = 10. both processed and verified are loaded from meshdb, so all the data is already there. when recover starts it turns out that we can verify up to 19. so we return old = 10, new = 19 to mesh. and it applies everything in between.

sync will call HandleIncomingLayer from 21 and onward. Hare is more likely from current and onward, and current maybe 21 but unlikely.

PR #2994, opened 2021-12-09
[Merged by Bors] - tortoise: refactor and optimize full tortoise

### LINKED ISSUE #2988: full tortoise (self-healing) rewrite
full tortoise code needs to be improved so that it is easier to reason that it is compatible with the designed protocol. The idea here is to implement verifying and full tortoise as decoupled as possible (they will share some common functions, but otherwise it will be possible to reason about each one separately).

```go
type State struct {
    // specific for full tortoise
    Votes    map[BallotID]map[BlockID]int8
    Base     map[BallotID]BallotID
    Weight   map[BallotID]uint
    
    // common. also used for base block
    Blocks   map[LayerID][]BlockID
    Ballots  map[LayerID][]BallotID
    
    // common. local opinion
    Validity map[BlockID]int8
}

func addBlocks(s State, lid LayerID, blocks []BlockID) {
    s.Blocks[lid] = blocks
}

type Ballot struct {
    ID     BallotID
    Base   BallotID
    Votes  map[BlockID]int8
    Weight uint
}

func addBallots(s State, lid LayerID, ballots []Ballot) {
    s.Ballots[lid] = ballots // ids
        s.Votes[ballot.ID] = ballot.Votes
        s.Base[ballot.ID] = ballot.Base
        s.Weight[ballot.ID] = ballot.Weight
    }
}

func getSign(s State, ballot BallotID, block BlockID) int8 {
    sign, exist := s.Votes[ballot][block]
    if !exist {
        return getSign(s, s.Base[ballot], block)
    }
    return sign
}

func sumVotes(s State, bid BlockID, blid, last LayerID) (uint, uint) {
    support, against := 0
    for lid := blid + 1; lid <= last; lid++ {
        for _, ballot := range s.Ballots[lid] {
            // for global threshold delay ballot by beacon distance
            // TODO ask Tal if we need to delay by beacon distance when we count votes for local threshold
            sign := getSign(s, ballot, bid)
            switch sign {
                case 1:
                    support += s.Weight[ballot]
                case -1:
                    against += s.Weight[ballot]
            }
        }
    }
    return support, against
}

func verifyLayers(s State, verified, last LayerID) LayerID {
    for lid := verified + 1; lid < last; lid++ {
        for _, bid := range s.Blocks[bid] {
            support, against := sumVotes(s, bid, lid, last)            
            if support - against > GLOBAL_THRESHOLD {
                s.Validity[bid] = 1
            } else if against - support > GLOBAL_THRESHOLD {
                s.Validity[bid] = -1
            } else {
                return lid - 1
            }
        }
    }
    return last - 1
}
```

### LINKED ISSUE #2989: tortoise: votes from ballots with bad beacon should be ignored when computing margin for local threshold
In the implementation votes from ballots with incorrect beacon are delayed by a configurable distance when votes are summed in full tortoise. The same logic should be used when votes are summed up for comparing with local threshold.

Per conversation with Tal:

> The margin is computed using all eligible ballots; ballots with a bad beacon are only eligible if they are old enough (e.g., at least 1 epoch old). The local threshold is computed using only weight of the last epoch's ATXs, regardless of which block we're considering.So, for example, suppose an epoch is 100 layers,  we have block X in layer 1, block Y in layer 120, and the current layer is 150.  Suppose the total weight of ATXs for epoch 1 is 30 and for epoch 2 is 70, but of that weight, in epoch each epoch 20 units use a bad beacon.Then the local threshold for both blocks X and Y uses the weight 70/100 for its calculation (the expected weight of a layer in epoch 2, ignoring beacon values).The margin for block X will take into account all ballots up to layer 49 (this means 48 layers with expected weight 30/100 each, and ballots with a good beacon from layers 50-150 (50 layers with expected weight 10/100, and 50 with expected weight 50/100). The margin for block Y will take into account ballots with a good beacon from layers 120-150 (30 layers with expected weight 50/100 each).

dshulyak:
## Motivation

This is a complete rewrite of the full tortoise, two big differences:
- votes are not copied eagerly, instead, they are fetched recursively when they are needed. it significantly reduces the amount of memory for full votes. From layer_size * window_size ** 2 to layer_size * window_size. for example it takes only ~530 mb to keep a window for 2000 layers.
- weight from ballots is computed in a dynamic programming way. full tortoise state keeps weight map, this is a weight margin from all ballots that we've seen so far. to verify a block we compare this weight margin with threshold (local + global) and if all blocks are decided the layer is verified.

benchmark after rewrite:

```
BenchmarkTortoiseLayerHandlingVerifying-16    	      10	 160678904 ns/op	74620717 B/op	 1025136 allocs/op
BenchmarkTortoiseLayerHandlingFull-16         	      10	 197655215 ns/op	76322932 B/op	 1112789 allocs/op
```
before 

```
BenchmarkTortoiseLayerHandlingVerifying-16    	      10	 161618693 ns/op	72998507 B/op	 1020338 allocs/op
BenchmarkTortoiseLayerHandlingFull-16         	      10	1143632575 ns/op	696198644 B/op	 1534722 allocs/op
```

Note that this benchmarks live-tortoise, for the rerun with full weight the situation is very different.

closes https://github.com/spacemeshos/go-spacemesh/issues/2988
closes https://github.com/spacemeshos/go-spacemesh/issues/2989

## Changes
- isolate code of the full tortoise
- look up the vote recursively instead of copying eagerly
- maintain weight margin for each block

## Test Plan
existing uts

## TODO

## DevOps Notes


tortoise/verifying.go
```diff
@@ -43,16 +43,12 @@ func (v *verifying) processLayer(logger log.Log, lid types.LayerID, ballots []to
...
-	localThreshold = localThreshold.add(v.epochWeight[v.processed.GetEpoch()])
-	localThreshold = localThreshold.fraction(v.LocalThreshold)
+func (v *verifying) verify(logger log.Log) types.LayerID {
+	localThreshold := computeLocalThreshold(v.Config, v.epochWeight, v.processed)
 
 	for lid := v.verified.Add(1); lid.Before(v.processed); lid = lid.Add(1) {
 		// TODO(dshulyak) expected weight must be based on the last layer.
-		threshold := computeExpectedWeight(v.epochWeight, lid, v.processed)
-		threshold = threshold.fraction(v.GlobalThreshold)
+		threshold := computeGlobalThreshold(v.Config, v.epochWeight, lid, v.processed)
```
countvonzero, line 51: since you name it `computeGlobalThreshold`, i think it's better to add local threshold inside that method per definition.

tortoise/tortoise.go
```diff
@@ -256,41 +260,48 @@ func (t *turtle) getGoodBallot(logger log.Log) (types.BallotID, types.LayerID) {
 
 // firstDisagreement returns first layer where local opinion is different from ballot's opinion within sliding window.
 func (t *turtle) firstDisagreement(ctx *tcontext, blid types.LayerID, ballotID types.BallotID, disagreements map[types.BallotID]types.LayerID) (types.LayerID, error) {
```
countvonzero, line 262: doesn't look like disagreement is used...

countvonzero (approved):

tortoise/tortoise.go
dshulyak: can you elaborate? this method is used in BaseBallot method

tortoise/verifying.go
dshulyak: i thought that for clarity there will be global threshold, local threshold and the threshold (which is a sum of both). you think it will be more accurate to have only first and second, and the first is a sum of both?

countvonzero: i thought that was the definition of `TG` in https://community.spacemesh.io/t/uncounted-weight-and-the-global-threshold/228 in all cases. TG now always include TL all the time.

tortoise/tortoise.go
countvonzero: i mean it's not used in this particular method `firstDisagreement`

dshulyak: thanks, i accidentally dropped an optimization that if base ballot is not consistent then i can just reuse disagreement layer. will need to think of the benchmark for this case

tortoise/verifying.go
dshulyak: you are right, it will be accurate to have only global and local. i will update it in next changeset, as i am optimizing it to work for a rerun with full weight

dshulyak:
bors merge

PR #2995, opened 2021-12-10
[Merged by Bors] - simplify beacon representation

countvonzero:
## Motivation

simplify beacon usage a little by using types.Beacon

## Changes

- simplify beacon usage a little by using types.Beacon

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -1,7 +1,6 @@
 package eligibility
 
 import (
-	"bytes"
 	"context"
 	"encoding/binary"
 	"errors"
@@ -148,8 +147,8 @@ func buildKey(l types.LayerID, r uint32) [2]uint64 {
 	return [2]uint64{uint64(l.Uint32()), uint64(r)}
 }
 
-func encodeBeacon(beacon []byte) uint32 {
-	return binary.LittleEndian.Uint32(beacon)
+func encodeBeacon(beacon types.Beacon) uint32 {
+	return binary.LittleEndian.Uint32(beacon.Bytes())
 }
 
 func (o *Oracle) getBeaconValue(ctx context.Context, epochID types.EpochID) (uint32, error) {
@@ -161,7 +160,7 @@ func (o *Oracle) getBeaconValue(ctx context.Context, epochID types.EpochID) (uin
 	value := encodeBeacon(beacon)
 	o.WithContext(ctx).With().Debug("hare eligibility beacon value for epoch",
 		epochID,
-		log.String("beacon", types.BytesToHash(beacon).ShortString()),
+		log.String("beacon", beacon.ShortString()),
 		log.Uint32("beacon_val", value))
 	return value, nil
 }
@@ -450,10 +449,10 @@ func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID) (map[st
 	}
 
 	var (
-		beacons       = make(map[types.EpochID][]byte)
+		beacons       = make(map[types.EpochID]types.Beacon)
 		activeBallots = make(map[types.BallotID]*types.Ballot)
 		epoch         types.EpochID
-		beacon        []byte
+		beacon        types.Beacon
 		err           error
 	)
 	for layerID := safeLayerStart; !layerID.After(safeLayerEnd); layerID = layerID.Add(1) {
@@ -464,7 +463,7 @@ func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID) (map[st
 				return nil, fmt.Errorf("error getting beacon for epoch %v: %w", epoch, err)
 			}
 			logger.With().Warning("found beacon for epoch", epoch,
-				log.String("epoch_beacon", types.BytesToHash(beacon).ShortString()))
+				log.String("epoch_beacon", beacon.ShortString()))
 			beacons[epoch] = beacon
 		}
 		ballots, err := o.meshdb.LayerBallots(layerID)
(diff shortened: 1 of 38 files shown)


hare/eligibility/oracle.go
```diff
@@ -190,7 +189,11 @@ func (o *Oracle) buildVRFMessage(ctx context.Context, layer types.LayerID, round
 	}
 
 	// marshal message
-	msg := vrfMessage{Beacon: v, Round: round, Layer: layer}
+	roundUint16 := uint16(round)
```
dshulyak: we have this in hare

```
preRound     uint32 = math.MaxUint32
certifyRound uint32 = math.MaxUint32 >> 1
```

i think we need to change round signature everywhere, weird that tests didn't crash

hare/eligibility/oracle_test.go
```diff
@@ -82,6 +81,12 @@ func mockLayerBallots(tb testing.TB, to *testOracle, layer types.LayerID, beacon
 	}
 }
 
+func beaconWithValOne() types.Beacon {
+	beacon := types.RandomBytes(types.BeaconSize)
+	copy(beacon, []byte{1, 0, 0, 0})
+	return types.BytesToBeacon(beacon)
```
dshulyak: is it the same as types.Beacon{1} ?

hare/eligibility/oracle.go
dshulyak: maybe we don't send vrf for preround and certify round

dshulyak: but i still think that that round signature needs to be changed everywhere for safety

dshulyak (approved):
i guess go.sum changed because of go version mismatch. have you tried go mod tidy? maybe it will cleanup smth

hare/eligibility/oracle.go
countvonzero: i am not sure either. i decided to drop this part of the change. too risky to be included in a no-op PR

hare/eligibility/oracle_test.go
countvonzero: yes. duh. thanks!

countvonzero:
@dshulyak 

need your help. i cannot figure out these test failure in tortoise. 
https://github.com/spacemeshos/go-spacemesh/runs/4482276204?check_suite_focus=true

these two tests fail consistently. but the odd thing is if i tweak the number of layers being tested, they passed. for example
- TestRecoverState
if i change from `sim.WithSequence(2)` to `sim.WithSequence(3)`, test passed consistently.
- TestBaseBallotEvictedBlock
if i change from `for i := 0; i < 50; i++` to `for i := 0; i < 51; i++`, test passed consistently.

thanks in advance.

tortoise/sim/layer.go
```diff
@@ -108,8 +108,7 @@ func (g *Generator) Next(opts ...NextOpt) types.LayerID {
 
 func (g *Generator) genBeacon() {
 	eid := g.nextLayer.Sub(1).GetEpoch()
-	beacon := make([]byte, 32)
-	g.rng.Read(beacon)
+	beacon := types.RandomBeacon()
```
dshulyak: btw please use internal g.rng here. the idea is that if seed is the same - all generated objects are the same

dshulyak:
atx weight is randomly generated in some range (from 1 to 10) and it is generated deterministically using seed (which is 0 for all test)

so in develop branch weight was generated in a way that allowed those tests to cross the threshold:

```
logger.go:130: 2021-12-10T14:26:41.327+0200	INFO	computed weight from good ballots	{"rerun": true, "layer_id": 57, "weight": "40.5", "total": "83.5", "expected": "37.25", "ballots_count": 10, "good_ballots_count": 10, "name": ""}
    logger.go:130: 2021-12-10T14:26:41.327+0200	INFO	candidate layer is verified	{"rerun": true, "layer_id": 57, "candidate_layer": "56", "voting_weight": "40.5", "threshold": "29.8", "local_threshold": "7.45", "name": ""}
```

in your branch because of that change for the beacon randomness weights were generated differently.

```
logger.go:130: 2021-12-10T14:33:22.977+0200	INFO	computed weight from good ballots	{"layer_id": 57, "weight": "28", "total": "68", "expected": "36.75", "ballots_count": 10, "good_ballots_count": 10, "name": ""}
    logger.go:130: 2021-12-10T14:33:22.977+0200	WARN	candidate layer is not verified. voting weight is lower than the threshold	{"layer_id": 57, "candidate_layer": "56", "voting_weight": "28", "threshold": "29.4", "local_threshold": "7.35", "name": ""}
```

i think it uncovered two issues with the generator:
- for unit tests we don't need to randomize weight, except when it is expected
- if weight is randomized generator should not be generating blocks evenly

but with the next change tests are passing 
```
diff --git a/tortoise/sim/layer.go b/tortoise/sim/layer.go
index 51fa761d..abd64540 100644
--- a/tortoise/sim/layer.go
+++ b/tortoise/sim/layer.go
@@ -108,7 +108,8 @@ func (g *Generator) Next(opts ...NextOpt) types.LayerID {
 
 func (g *Generator) genBeacon() {
        eid := g.nextLayer.Sub(1).GetEpoch()
-       beacon := types.RandomBeacon()
+       beacon := types.Beacon{}
+       g.rng.Read(beacon[:])
        for _, state := range g.states {
                state.OnBeacon(eid, beacon)
        }
****
```

tortoise/sim/layer.go
countvonzero: thanks for the explanation. make sense. changed as advised

countvonzero:
bors merge

PR #2996, opened 2021-12-10
[Merged by Bors] - tortoise: rational numbers for weight

dshulyak:
## Motivation

I intended to replace big.Float with fixed float library. But realized that rational number's make more sense here, we are using two distinct weights:
- counted weight from ballots, each unit is atx_weight / eligibility_count
- expected weight from epochs in a single, each unit is total_atx_weight / layers_in_epoch  
Later expected weight is multiplied by a fraction, that is also represented as a rational number. And after multiplication two weights are compared. 

So it seems just more straightforward to use rational numbers in this case. They are slightly less performant for full tortoise, but it doesn't seem to matter.

## Changes
- use big.Rat instead of big.Float for weight

## Test Plan
ut's

## TODO

## DevOps Notes


countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #2997, opened 2021-12-13
[Merged by Bors] - UCB: use Proposal wherever make sense

### LINKED ISSUE #2902: Refactor: `Block`→`Proposal`

countvonzero:
## Motivation

Closes #2902

## Changes

- typedef Block/BlockID to Proposal/ProposalID
- based on UCB logic, changed Block->Proposal wherever it fits in the codebase
- Genesis Block/Ballot now generated with a hard-coded private key. the idea is to be able to retrieve from DB and initialize without special handling.
- replace blocks with proposals package
- in meshdb, a new ballot DB is created for ballots for independent fetching from Proposals
- in meshdb, `DBBlock` struct is removed 

## Test Plan

unittest /system tests

## TODO

- [ ] add tests in mesh for ballots

## DevOps Notes


countvonzero:
@dshulyak sorry for the large PR. it's nevertheless rote and monolithic. 
i've highlighted the important stuff in the PR description. it's prolly easier to review commit-by-commit.

countvonzero:
bors try

dshulyak:
> in meshdb, DBBlock struct is removed

this can't be done. Initilize is very expensive and needs to be done only when block is received from the network, extracting signature every time is pretty much unacceptable. it was affecting hare in the past https://github.com/spacemeshos/go-spacemesh/pull/2562 and will affect tortoise

common/types/proposal.go
```diff
@@ -21,6 +24,25 @@ const (
 	ProposalIDSize = Hash32Length
 )
 
+// GenesisBlock returns the genesis block.
+func GenesisBlock() *Block {
```
dshulyak: in my opinion, this gets us further from how it should be not closer. genesis block should be just an id. is there even any reason to store in database? in tortoise it is added to the state by re-generating it, which is also not good. i think it should be just a parameter in the genesis config.

mesh/meshdb.go
```diff
@@ -145,109 +159,188 @@ func (m *DB) Close() {
...
+	data, err := m.getBallotBytes(id)
+	if err != nil {
+		return nil, err
+	}
+	b := &types.Ballot{}
+	if err := types.BytesToInterface(data, b); err != nil {
+		return nil, fmt.Errorf("parse ballot: %w", err)
 	}
-	return nil, err
+	if err := b.Initialize(); err != nil {
```
dshulyak: should not be called when loading from db

mesh/meshdb.go
```diff
@@ -145,109 +159,188 @@ func (m *DB) Close() {
...
+func (m *DB) HasProposal(id types.ProposalID) bool {
+	has, err := m.proposals.Has(id.Bytes())
+	return err == nil && has
 }
 
-// LayerBlocks retrieves all blocks from a layer by layer index.
-func (m *DB) LayerBlocks(index types.LayerID) ([]*types.Block, error) {
-	ids, err := m.LayerBlockIds(index)
+// AddProposal adds a proposal to the database.
+func (m *DB) AddProposal(p *types.Proposal) error {
```
dshulyak, line 276: proposal stores mostly the same data as the ballot. i don't think that duplication is good solution, instead when loading proposal should fetch data from ballots if needed

mesh/meshdb.go
```diff
@@ -145,109 +159,188 @@ func (m *DB) Close() {
...
+func (m *DB) getProposal(id types.ProposalID) (*types.Proposal, error) {
+	data, err := m.getProposalBytes(id)
+	if err != nil {
+		return nil, err
+	}
+	p := &types.Proposal{}
+	if err := types.BytesToInterface(data, p); err != nil {
+		return nil, fmt.Errorf("parse proposal: %w", err)
+	}
+	if err := p.Initialize(); err != nil {
```
dshulyak: should not call initialize in db

mesh/meshdb.go
```diff
@@ -1076,3 +1165,29 @@ func (db *BlockFetcherDB) Get(hash []byte) ([]byte, error) {
...
+
+// Get types.Block encoded in byte using hash.
+func (db *BallotFetcherDB) Get(hash []byte) ([]byte, error) {
+	id := types.BallotID(types.BytesToHash(hash).ToHash20())
+	blk, err := db.mdb.GetBallot(id)
+	if err != nil {
+		return nil, fmt.Errorf("get ballot: %w", err)
+	}
+
+	data, err := types.InterfaceToBytes(blk)
```
dshulyak: can we use codec.Encode / codec.Decode in the new code please

common/types/proposal.go
noamnelke: I intend to write a SMIP for the genesis flow in the next few days. I wasn't even aware that we have a `GenesisBlock`...

Does it have any functional purpose at the moment? What breaks if we remove it?

As far as I know we should only have a `GoldenATX` in epoch 0 which is used as the `PositioningATX` for the ATXs in epoch 1. That should be only an ID as well - eventually, it won't be a just a random number, but a hash of something we define in advance. I'll elaborate on that in the SMIP, but just want to know if I have to solve any problems if I want to remove the genesis **block**.

mesh/meshdb.go
noamnelke: I think we should be smart about when to do this costly initialization.

When receiving a ballot over the wire, we obviously have to calculate the ballot ID and extract the public key from the signature for validation.

When fetching from the DB, since the key includes the ID we should inject that into the object before returning it.

The author smesher ID should be calculated lazily and then cached. In most cases I don't expect to need it, but if we do need it in a performance critical code path, we should store it in the DB.

common/types/proposal.go
dshulyak: tortoise depends on it so that it can start in verifying mode, otherwise, it will switch into full mode immediately. once this is implemented https://github.com/spacemeshos/go-spacemesh/issues/2985 that dependency will be irrelevant

countvonzero:
> > in meshdb, DBBlock struct is removed
> 
> this can't be done. Initilize is very expensive and needs to be done only when block is received from the network, extracting signature every time is pretty much unacceptable. it was affecting hare in the past #2562 and will affect tortoise

technically this is a separate issue from calling Initialize(). the only value i see is to preserve the original ID. but i guess there is always a possibility that ID can change later on. so i am bringing back the DB specific structure. thanks.

dshulyak:
> technically this is a separate issue from calling Initialize(). 

Initialize computes ID and extracts public key. both are expensive. if they won't do either then it is not about Initialize 

> the only value i see is to preserve the original ID. but i guess there is always a possibility that ID can change later on.

but id is a hash of the data, it never changes unless there is a bug, which is a different story

countvonzero:
> this can't be done. Initilize is very expensive and needs to be done only when block is received from the network, extracting signature every time is pretty much unacceptable. it was affecting hare in the past #2562 and will affect tortoise

restored.

mesh/meshdb.go
countvonzero: done for the files in this PR

mesh/meshdb.go
countvonzero: removed

mesh/meshdb.go
countvonzero: i ended up trimming proposal to the minimum info to store in db.

mesh/meshdb.go
countvonzero: changed to such that SmesherID() is calculated eagerly upon receiving the data, but lazily when retrieved from DB.

common/types/proposal.go
countvonzero: ok. now changed to use hardcoded hex strings for genesis ballot and block IDs.

> is there even any reason to store in database?

i tried not storing it in the database but that caused several tortoise test failures. i didn't want to spend time fixing them so leave them as is for now.

countvonzero:
bors try

countvonzero:
@dshulyak please take another look. thank you for catching the performance issue.

my next PR will be changing hare to generate the final content block. that will officially separate Proposal from Block.

mesh/meshdb.go
countvonzero: there were two paths to restore a Proposal:
- DBProposal.ToBlock(): used by meshdb clients that are interested in getting Blocks (tortoise/mesh)
- DBProposal.ToProposal(): used by serving peer's request from fetch.GetBlocks(). this is currently how the node sync Proposals/Ballots.

mesh/meshdb.go
noamnelke: Didn't look at the code, but for simplicity it can always be lazy. When receiving we will immediately try to validate the ballot, so this will trigger the ID calculation.

dshulyak:
now smesher id will be computed when we log the ballot/proposal fields, can we please restore previous approach or remove smesher id from the structure all together. e.g. recover it only when ballot is received, do something with it if necessary and log there.

countvonzero:
> now smesher id will be computed when we log the ballot/proposal fields, can we please restore previous approach or remove smesher id from the structure all together. e.g. recover it only when ballot is received, do something with it if necessary and log there.

done. restored the previous behavior

mesh/meshdb.go
countvonzero: changed the code as @dshulyak suggested (restored the previous behavior).

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

PR #3000, opened 2021-12-14
[Merged by Bors] - tortoise: rerun counting votes from longer distances

### LINKED ISSUE #2980: tortoise: expected weight during rerun should be based on current layer

dshulyak:
## Motivation

Count votes from later layers during rerun to un-verify erroneously verified layers. I know only about one case - when malicious majority accumulated enough weight to verify a layer, but later layers disagreed with it. This is probably enough to justify this feature, but this is also something that will be discussed on the next tortoise research call.

closes: https://github.com/spacemeshos/go-spacemesh/issues/2980

## Changes
- recompute thresholds only when necessary
- dont load the same hare input vectors multiple times, load only new once and undecided
- expected weight in both modes is computed according to the window, for example, if last layer is 200, target layer is 10, and window is 100 - expected weight will be based on weight from layers 11 - 110 inclusive. the window is different for full and verifying tortoise, as the complexity of the protocols is different. even with recent optimizations full takes an unreasonably long time when window is larger, so for now it is limited to 100. Also, we do not switch into a full mode during rerun until we count votes from the same layers that are considered for expected weight.

## Test Plan
new and existing uts

## TODO

## DevOps Notes


tortoise/algorithm.go
```diff
@@ -20,12 +20,13 @@ type Config struct {
 	Hdist uint32 `mapstructure:"tortoise-hdist"` // hare/input vector lookback distance
 	Zdist uint32 `mapstructure:"tortoise-zdist"` // hare result wait distance
 	// how long we are waiting for a switch from verifying to full. relevant during rerun.
-	ConfidenceParam uint32        `mapstructure:"tortoise-confidence-param"`
```
countvonzero, line 23: rm'ed

tortoise/rerun_test.go
```diff
@@ -50,3 +50,99 @@ func TestRecoverState(t *testing.T) {
...
+	require.Equal(t, last.Sub(1), verified)
+
+	blocks, err := s.GetState(0).MeshDB.LayerBlockIds(misverified)
+	require.NoError(t, err)
+	for _, block := range blocks {
+		validity, err := s.GetState(0).MeshDB.ContextualValidity(block)
+		require.NoError(t, err)
+		require.False(t, validity, "validity for block %s", block)
+	}
+	require.NoError(t, tortoise.rerun(ctx))
```
countvonzero, line 90: can you test here that rerun did change our opinions on the mis-verified layer? 

as the code is now, i am not sure it's the call tortose.rerun() or the extra layer votes later that made the difference.

tortoise/threshold.go
```diff
@@ -84,16 +85,52 @@ func computeExpectedWeight(weights map[types.EpochID]weight, from, to types.Laye
...
+// updateThresholds recomputes local and global thresholds:
+// - when last layer is updated
+// - when verified layer is updated
+// - when switching from one mode into the other.
+func updateThresholds(logger log.Log, config Config, state *commonState, tmode mode) {
+	state.localThreshold = computeLocalThreshold(config, state.epochWeight, state.last)
+
+	window := getConfidenceWindow(config, state, tmode)
+	if window.Before(state.processed) {
+		window = state.processed
```
countvonzero, line 121: why should window be reset to state.processed?

and if this is necessary, why not do it inside `getConfidenceWindow`

tortoise/util.go
```diff
@@ -115,16 +119,84 @@ func persistContextualValidity(logger log.Log,
...
+	reasonCoinflip       voteReason = "coinflip"
+)
+
+// lsb is a mode, second bit is rerun
+// examples (lsb is on the right, prefix with 6 bits is droppped)
+// 10 - rerun in verifying
+// 00 - live tortoise in verifying
+// 11 - rerun in full
+// 01 - live tortoise in full.
+type mode uint8
```
countvonzero: is this performance critical that requires bit manipulation like this? what's wrong with the previous version with iota?

tortoise/threshold.go
```diff
@@ -84,16 +85,52 @@ func computeExpectedWeight(weights map[types.EpochID]weight, from, to types.Laye
...
 }
+
+func getConfidenceWindow(config Config, state *commonState, tmode mode) types.LayerID {
+	target := state.verified.Add(1)
+	window := state.last
+	if tmode.isFull() && state.last.Difference(target) > config.FullModeRerunWindow {
+		window = target.Add(config.FullModeRerunWindow)
+	} else if tmode.isVerifying() && state.last.Difference(target) > config.VerifyingModeRerunWindow {
+		window = target.Add(config.VerifyingModeRerunWindow)
+	}
```
countvonzero, line 108: do you plan for this to have a 3rd kind of window for sync?

dshulyak: i think sync should use same window, i will rename it. maybe FullModeVerificationWindow

tortoise/threshold.go
dshulyak: so let's say window is 100, we want to verify layer 1 and the processed layer is 400. it means that we counted votes from layers 2 up to 400, so expected weight should be no less than that

tortoise/util.go
dshulyak: the reason was that i can change rerun state without changing mode, and mode without changing rerun. using 1 and 2 bit was convenient, so i can toggle them in toggleRerun and toggleMode. i can use [2]bool instead if that's seems odd

tortoise/threshold.go
dshulyak: basically it should be applied any time when last layer - verified > this window

tortoise/rerun_test.go
dshulyak: sure, will reorder checks

tortoise/threshold.go
dshulyak: it will the case if we can't decide on the layer

dshulyak: i am also using getConfidenceWindow(config, state, tmode) to decide if i need to recompute thresholds

countvonzero (approved):

dshulyak:
bors merge

PR #3001, opened 2021-12-15
[Merged by Bors] - tortoise: restart verifying by using output from full tortoise

dshulyak:
## Motivation

first part of https://github.com/spacemeshos/go-spacemesh/issues/2985

## Changes

after switching into full mode run verifying tortoise after full tortoise makes progress. switch back once verifying tortoise can verify same layer

## TODO

## DevOps Notes


tortoise/tortoise.go
```diff
@@ -508,37 +525,48 @@ func (t *turtle) processLayer(ctx context.Context, logger log.Log, lid types.Lay
...
+		if !success && (t.canUseFullMode() || t.mode.isFull()) {
+			if t.mode.isVerifying() {
+				t.switchModes(logger)
+				// reset accumulated weight as verifying tortoise will re-counting votes
+				// with input from full tortoise.
+				t.verifying.resetWeights()
 			}
-			t.verified = lid
+			t.full.countVotes(logger)
+			success = t.full.verify(logger, target)
```
countvonzero, line 542: would this be easier if you make t.mode an tortoise object and so something similar to the psudocode below
```
verify:
t.mode.countVotes()
for target := t.verified.Add(1); target.Before(t.processed); target = target.Add(1) {
   success = t.mode.verify(target)
   if !success {
        if t.mode == full {
           // fail
        }
        t.swithcModes()
        goto verify  
   }
}
```

countvonzero (approved):

tortoise/tortoise.go
dshulyak: but the pseudocode dropped almost everything, it is hard for me to tell what it will actually simplify.

restart will not be simplified, because it needs to recount votes for full distance between target + 1 and processed.
also i can't use single-mode object as i always need to have access to both.
beside that i think that goto in golang is a mistake in the language and there is never a good reason to use it

dshulyak:
bors merge

tortoise/tortoise.go
countvonzero: now that that verifying and full have the same interface (countVotes and verify), maybe it's possible to remove the frequent `if t.mode.isVerifying()` checks. the underlying mode should know what to do respectively.

just an idea. no biggie.

PR #3002, opened 2021-12-16
[Merged by Bors] - tortoise: restart verifying by force

### LINKED ISSUE #2678: Prevent base block conflicts after rescoring local opinion using vote counting
## Description
Related: #2402

If a node does not have an opinion on a layer that's sufficiently old, tortoise will calculate its opinion by counting votes (i.e., the global opinion). If the global opinion isn't distinct, it will fall back on the tortoise weak coin. This is a core element of self-healing.

When this happens, it's likely that many otherwise-good blocks won't be marked good because they'll contain exceptions that disagree with the (new) local opinion on blocks that have been rescored in this fashion. We may need to change the way we pick base blocks in this scenario.

## Affected code
https://github.com/spacemeshos/go-spacemesh/blob/9d3a379994e3652eb0378d2cc555588bba91005e/tortoise/verifying_tortoise.go#L1049-L1060

This issue appears in commit hash: 9d3a379994e3652eb0378d2cc555588bba91005e

### LINKED ISSUE #2985: tortoise: switch from healing to verifying

dshulyak:
## Motivation

when tortoise sees that some ballots started to vote consistently with local opinion it tries to bootstrap a verifying mode. in order to bootstrap a verifying mode it needs to find a cut of ballots that vote consistently with local opinion

| block/ballot | local opinion | 0xaa | 0xbb | 0xcc | 0xdd | 0xee | 0xff |
| ------------ | ------------- | ---- | ---- | ---- | ---- | ---- | ---- |
| base         | -             | -    | -    | 0xaa | 0xaa | 0xcc | 0xcc |
| layer        | -             | 10   | 10   | 11   | 11   | 12   | 12   |
| 0x11         | 1             | -1   | -1   | 1    | -1   | 1    | 1    |
| 0x22         | 1             | -1   | -1   | 1    | -1   | 1    | 1    |
| 0x33         | -1            | -    | -    | -1   | -1   | -1   | -1   |
| 0x44         | -1            | -    | -    | -1   | -1   | -1   | -1   |
| 0x55         | 1             | -    | -    | -    | -    | 1    | 1    |
| 0x66         | -1            | -    | -    | -    | -    | -1   | -1   |

In this example ballots starting from layer 10 are voting for some old layer (for example 5). Ballots from layer 10 disagree with local opinion and will be marked bad. Ballot 0xcc from layer 10 and ballots 0xee and 0xff will be marked "can be good" because votes from those ballots are consistent with local opinion. 0xee -> 0xcc and 0xff -> 0xcc is sufficient to bootstrap verifying tortoise.

closes: https://github.com/spacemeshos/go-spacemesh/issues/2985
closes: https://github.com/spacemeshos/go-spacemesh/issues/2678

## Changes
- try to bootstrap tortoise into verifying mode even if there are no good in the memory

## Test Plan
uts

## TODO

## DevOps Notes

### DIFF
--- a/tortoise/threshold.go
+++ b/tortoise/threshold.go
@@ -98,7 +98,7 @@ func computeThresholdForLayers(config Config, epochWeight map[types.EpochID]weig
 	return threshold
 }
 
-func getConfidenceWindow(config Config, state *commonState, tmode mode) types.LayerID {
+func getVerificationWindow(config Config, state *commonState, tmode mode) types.LayerID {
 	target := state.verified.Add(1)
 	window := state.last
 	if tmode.isFull() && state.last.Difference(target) > config.FullModeVerificationWindow {
@@ -116,7 +116,7 @@ func getConfidenceWindow(config Config, state *commonState, tmode mode) types.La
 func updateThresholds(logger log.Log, config Config, state *commonState, tmode mode) {
 	state.localThreshold = computeLocalThreshold(config, state.epochWeight, state.last)
 
-	window := getConfidenceWindow(config, state, tmode)
+	window := getVerificationWindow(config, state, tmode)
 	if window.Before(state.processed) {
 		window = state.processed
 	}
--- a/tortoise/verifying.go
+++ b/tortoise/verifying.go
@@ -24,9 +24,13 @@ func (g goodness) String() string {
 
 const (
 	bad goodness = iota
+	// good ballot must:
+	// - agree on a beacon value
+	// - don't vote on blocks before base ballot
+	// - have consistent votes with local opinion
+	// - have a good base ballot.
 	good
-	// canBeGood is for ballots that are voting consistently with local opinion
-	// within base ballot layer up to ballot layer.
+	// canBeGood is the same as good, but doesn't require good base ballot.
 	canBeGood
 )
 
@@ -54,10 +58,6 @@ type verifying struct {
 func (v *verifying) resetWeights() {
 	v.totalWeight = weightFromUint64(0)
 	v.layerWeights = map[types.LayerID]weight{}
-}
-
-func (v *verifying) checkIsGood(ballot types.BallotID) bool {
-	return v.goodBallots[ballot] == good
 }
 
 func (v *verifying) checkCanBeGood(ballot types.BallotID) bool {
@@ -144,7 +144,7 @@ func (v *verifying) sumGoodBallots(logger log.Log, ballots []tortoiseBallot) (we
 	sum := weightFromUint64(0)
 	n := 0
 	for _, ballot := range ballots {
-		rst := v.isGood(logger, ballot)
+		rst := v.determineGoodness(logger, ballot)
 		if rst == bad {
 			continue
 		}
@@ -157,7 +157,7 @@ func (v *verifying) sumGoodBallots(logger log.Log, ballots []tortoiseBallot) (we
 	return sum, n
 }
 
-func (v *verifying) isGood(logger log.Log, ballot tortoiseBallot) goodness {
+func (v *verifying) determineGoodness(logger log.Log, ballot tortoiseBallot) goodness {
 	logger = logger.WithFields(ballot.id, log.Stringer("base", ballot.base))
 
 	if _, exists := v.badBeaconBallots[ballot.id]; exists {
@@ -192,7 +192,7 @@ func (v *verifying) isGood(logger log.Log, ballot tortoiseBallot) goodness {
 	}
 
 	if rst := v.goodBallots[ballot.base]; rst != good {
-		logger.With().Debug("ballot is not good. base is not good")
+		logger.With().Debug("ballot can be good. only base is not good")
 		return canBeGood
 	}
 
--- a/tortoise/verifying.go
+++ b/tortoise/verifying.go
@@ -1,15 +1,40 @@
 package tortoise
 
 import (
+	"fmt"
+
 	"github.com/spacemeshos/go-spacemesh/common/types"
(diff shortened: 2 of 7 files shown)


dshulyak:
i think this is also allows to get rid from "genesis" block thing, but i will try to do it in a separate pr.

countvonzero:
> Ballot 0xcc from layer 10 and ballots 0xee and 0xff will be marked "can be good" because exceptions from those ballots are consistent with local opinion. 0xee -> 0xcc and 0xff -> 0xcc is sufficient to bootstrap verifying tortoise.

in your example, why would oxee and oxff have exceptions from oxcc? they agree on those blocks.

do you think you can give "can be good" a definition?

tortoise/threshold.go
```diff
@@ -98,7 +98,7 @@ func computeThresholdForLayers(config Config, epochWeight map[types.EpochID]weig
 	return threshold
 }
 
-func getConfidenceWindow(config Config, state *commonState, tmode mode) types.LayerID {
+func getVerificationWindow(config Config, state *commonState, tmode mode) types.LayerID {
```
countvonzero, line 101: i think it'll be beneficial to write comments about why this window is important, as discussed in the research call.

tortoise/verifying.go
```diff
@@ -31,6 +56,31 @@ func (v *verifying) resetWeights() {
 	v.layerWeights = map[types.LayerID]weight{}
 }
 
+func (v *verifying) checkIsGood(ballot types.BallotID) bool {
```
countvonzero: unused

tortoise/verifying.go
```diff
@@ -94,27 +144,25 @@ func (v *verifying) sumGoodBallots(logger log.Log, ballots []tortoiseBallot) (we
...
 	for _, ballot := range ballots {
-		if !v.isGood(logger, ballot) {
+		rst := v.isGood(logger, ballot)
+		if rst == bad {
 			continue
 		}
-		v.goodBallots[ballot.id] = struct{}{}
-		sum = sum.add(ballot.weight)
-		n++
+		v.goodBallots[ballot.id] = rst
```
countvonzero, line 151: use checkIsGood?

tortoise/verifying.go
```diff
@@ -139,10 +187,15 @@ func (v *verifying) isGood(logger log.Log, ballot tortoiseBallot) bool {
 				log.Stringer("base_layer", baselid),
 				log.Stringer("block", block),
 			)
-			return false
+			return bad
 		}
 	}
 
+	if rst := v.goodBallots[ballot.base]; rst != good {
```
countvonzero, line 194: checkIsGood?

tortoise/verifying.go
```diff
@@ -94,27 +144,25 @@ func (v *verifying) sumGoodBallots(logger log.Log, ballots []tortoiseBallot) (we
...
+		if rst == good {
+			sum = sum.add(ballot.weight)
+			n++
+		}
 	}
 	return sum, n
 }
 
-func (v *verifying) isGood(logger log.Log, ballot tortoiseBallot) bool {
+func (v *verifying) isGood(logger log.Log, ballot tortoiseBallot) goodness {
```
countvonzero: can you add comment here that compare "can be good" and "good"? thanks

countvonzero (approved):

tortoise/verifying.go
dshulyak: Can you clarify the request please

tortoise/threshold.go
dshulyak: Not related to this pr though

dshulyak:
Please check the definition for can be good in verifyong.go. or you are looking for smth else?

countvonzero:
> Please check the definition for can be good in verifyong.go. or you are looking for smth else?

do you mean by reading the code? i am looking for a comment that gives definition of 
- good
- can be good
- bad

nkryuchkov (approved):

dshulyak:
@countvonzero 

> do you mean by reading the code? i am looking for a comment that gives definition of

yes i meant the comment above constants. i also tried to improve it in last commit

> in your example, why would oxee and oxff have exceptions from oxcc? they agree on those blocks.

i included them in the table so that they can be easily compared with local opinion. i updated the example with more blocks, so it should make more sense now

tortoise/verifying.go
dshulyak: i will drop checkIsGood i guess, it makes more sense to check that rst is god

dshulyak:
bors merge

PR #3011, opened 2021-12-20
[Merged by Bors] - tortoise: depreoritize ballots with bad beacons

dshulyak:
## Motivation

if tortoise is in verifying mode it is using last good ballot as a base ballot, and encodes only votes starting from the base ballot layer. once switched into full mode we can't rely on this optimization anymore and need to recompute the least bad ballot, and encode votes for every layer that we have in memory. in such case we shouldn't select base ballot with bad beacon, as it complicates procedure for switching tortoise into verifying mode.

reproduced this during testing.

## Changes
- use ballots with a bad beacon only if there are no ballots with good beacon

## Test Plan
unit tests

## TODO

## DevOps Notes


nkryuchkov (approved):

dshulyak:
bors merge

PR #3016, opened 2021-12-21
[Merged by Bors] - mesh: maintain a separate list of ballots and proposals

### LINKED ISSUE #3015: sync, tortoise: after node was disconnected for long time it may miss significant part of state
For testing recent changes i created a network with a multiple chaos conditions. One of them disconnects 5 nodes every 2h for 1h. As this time is longer than a single epoch all of them individually made progress and verified mesh that consisted only of blocks generated in isolation.

After they reconnect with the network i expected them to fetch missing data, because the fetcher is supposed to be recursive. So if we see a ballot from gossip that references some unknown base ballot and votes for unknown blocks we will try to fetch every unknown ballot and block. However after comparing data on those nodes it turns out not to be the case, and because of missing data the recovery process is very slow, not even sure that they will be able to recover completely.

So specifically miner46 received 32 ballots on layer 1117, while there was 53. However, some of the ballots on layer 1118 are referencing ballots from layer 1117 and they are processed without error. It means that ballot receiver for some reason saved ballots from 1118 before saving ballots from 1117, which should not be possible as voting is recursive and it screws up the tortoise.

dshulyak:
## Motivation

HandleBallotData wasn't (and can't) update list of proposals, but we still need to keep database state consistent.

closes: https://github.com/spacemeshos/go-spacemesh/issues/3015

## Changes
- keep a list of ballots that gets updated every time AddBallot is called

## Test Plan
uts

## TODO

## DevOps Notes

### DIFF
--- a/mesh/meshdb.go
+++ b/mesh/meshdb.go
@@ -18,6 +18,9 @@ import (
 
 const (
 	layerSize = 200
+
+	layerProposals = "p"
+	layerBallots   = "b"
 )
 
 // DB represents a mesh database instance.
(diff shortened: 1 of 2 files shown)


countvonzero (approved):
was in my next PR that moves proposals away from mesh :) 
thanks.

dshulyak:
bors merge

nkryuchkov (approved):

PR #3017, opened 2021-12-21
[Merged by Bors] - start syncing after genesis layer

countvonzero:
## Motivation

simplify mesh's layer processing logic

## Changes

- set processed layer to genesis layer when initializing mesh. this means syncing will start from genesis+1 layer
- remove logic for out-of-order validated layer in mesh as this cannot happen with tortoise logic
- use gomock for state/tortoise in mesh tests
- in mesh, change rewards calculations to use utility functions

## Test Plan

unit test

## DevOps Notes


countvonzero:
bors try

mesh/mesh.go
```diff
@@ -641,7 +594,7 @@ func (msh *Mesh) getAggregatedLayerHash(layerID types.LayerID) (types.Hash32, er
 	return hash, fmt.Errorf("get from DB: %w", err)
 }
 
-func (msh *Mesh) extractUniqueOrderedTransactions(l *types.Layer) (validBlockTxs []*types.Transaction) {
+func extractUniqueOrderedTransactions(l *types.Layer, mdb *DB, logger log.Log) (validBlockTxs []*types.Transaction) {
```
dshulyak: why do you prefer to pass logger as a last value? i think we need some consistent approach

i am usually using it as a first argument as to me the idea is the same as with ctx. so logger logically can be a part of context

dshulyak (approved):

mesh/mesh.go
countvonzero: i don't actually have a preference. changed.

countvonzero:
bors merge

countvonzero:
bors merge

nkryuchkov (approved):

Issue #2906, opened 2021-10-31
Derive `Block` from Hare Output Set


noamnelke:
When the Hare completes we'll now want to generate a `Block` and store it in a separate database.

We'll use the same mechanism we use today to create a list of ordered transactions (`mesh.extractUniqueOrderedTransactions()`), to generate a `Block`. In addition to an ordered list of transactions, a block should contain a list of rewards. This should be derived from the list of smeshers that authored the proposals that the Hare agreed on. For each smesher we'll find the coinbase address in the relevant ATX and add a reward proportional to the number of proposals/ballots authored by them and included in the Hare output set.

Blocks should be persisted to a `BlockDB`. When the node calculates a block from Hare results, it should also store a value in a `layer -> block` index in the same database. To be explicit - we should NOT store a value in this index if we add a block to the database because we've synced it or fetched it from a peer, only when we've created it ourselves based on Hare results.

PR #3020, opened 2021-12-23
[Merged by Bors] - UCB: add block handler and generator

countvonzero:
## Motivation

1st PR for #2906

## Changes

- add a temporary struct UCBlock. this will be renamed to Block in a later PR
- add block data handler when fetching from peers
- add block generator that generate a block from a list of proposals

## Test Plan

unit tests

## DevOps Notes


blocks/reward.go
```diff
@@ -0,0 +1,53 @@
...
+	numProposals uint64
+	feesReward   uint64
+	layerReward  uint64
+	// smesher reward per proposal
+	totalRewardPer uint64
+	// the layer reward part of the smesher reward per proposal
+	layerRewardPer uint64
+}
+
+func (info *layerRewardsInfo) Fields() []log.LoggableField {
```
dshulyak: is it a new code? should be using ObjectMarshaller interface if so

we should stop using Fields interface for logging, need to add a note that somewhere that it is deprecated

common/types/ucb.go
```diff
@@ -0,0 +1,74 @@
...
+	return bytes
+}
+
+// ID returns the BlockID.
+func (b *UCBlock) ID() BlockID {
+	return b.blockID
+}
+
+// Fields returns an array of LoggableFields for logging.
+func (b *UCBlock) Fields() []log.LoggableField {
```
dshulyak: should use ObjectMarshaler interface instead

dshulyak (approved):

blocks/reward.go
countvonzero: thanks for the reminder. i've dropped Fields here and for Ballot and Proposal too.

common/types/ucb.go
countvonzero: done

countvonzero:
bors merge

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/4617008764?check_suite_focus=true)

turned out the system test relied on one of the proposal log field.

countvonzero:
bors merge

PR #3031, opened 2021-12-28
[Merged by Bors] - tortoise: describe tortoise implementation in README

dshulyak:
Meant as an overview for tortoise protocol, including implementation details.

[rendered](https://github.com/spacemeshos/go-spacemesh/blob/83d0e9fff1253c28dcbf1a1ddde0ad646c0a9d86/tortoise/README.md)

### DIFF
--- a/tortoise/README.md
+++ b/tortoise/README.md
@@ -14,17 +14,19 @@ Tortoise counts votes from the syntactically valid ballots.
 
 For tortoise purposes every ballot must have a unique ID, reference to the base ballot ID, explicit votes (exceptions), weight and a beacon.
 
-###### Base ballot and exceptions
+##### Base ballot and exceptions
 
 Discussed in [Voting section](#voting).
 
-###### Weight
+##### Weight
 
-Weight of the ballot is computed from the weight of the activation transaction divided by the total number of eligible ballots. Total number of eligible ballots is computed based on the fraction of actiovation transaction compared to the weight from the active set that was recorded in reference ballot.
+Weight of the ballot is computed from the weight of the activation transaction divided by the total number of eligible ballots. Total number of eligible ballots is computed based on the fraction of activation transaction compared to the weight from the active set that was recorded in reference ballot.
 
-###### Beacon
+##### Beacon
 
-Ballot with an inconsistent beacon will not be marked good and will not be counted immediatly. Discussed in [full](#votes-from-ballots-with-wrong-beacon) and [verifying](#good-ballots) sections.
+Beacon is a source of randomness and is calculated by the all smeshers per epoch. Ballots with a different beacon is either malicious or a result of a network split.
+
+Ballot with an inconsistent beacon will not be marked good and will not be counted immediately. Discussed in [full](#votes-from-ballots-with-wrong-beacon) and [verifying](#good-ballots) sections.
 
 #### Expected weight and thresholds
 
@@ -39,7 +41,7 @@ In order to minimize amount of data that needs to be sent and stored, Tortoise s
 Types of votes:
 - support
 
-    Support all blocks according to local opinion that are not supported by base ballot. Usually support is added for layers starting from base ballot layer up to last layer. In the event of healing tortoise may add support for blocks before base ballot layer.
+    Support all blocks according to the local opinion that are not supported by base ballot. Usually support is added for layers starting from base ballot layer up to the last processed layer. In the event of healing tortoise may add support for blocks before base ballot layer.
 
 - against
   
@@ -49,16 +51,16 @@ Types of votes:
     
     Undecided is a valid vote only until hare is terminated for the layer (it may take more than one layer for hare to terminate, this is expressed as zdist).
 
-If ballot doesn't specify explicit vote for a block then it votes agains the block. This prevents malicious smeshers from retroactively reorganizing mesh by intentionally creating late ballots.
+If ballot doesn't specify explicit vote for a block then it votes against the block. This prevents malicious smeshers from retroactively reorganizing mesh by intentionally creating late ballots.
 
 There are 4 distinct reasons to cast a vote:
 - hare output
     
     tortoise uses output from the hare consensus for voting on blocks within hdist.
 
-- validity/decisions
+- validity
 
-    for layer before hdist the vote is casted according to the decision. 
+    for layer before hdist the vote is cast according to the decision. 
 
 - local threshold
 
@@ -66,7 +68,7 @@ There are 4 distinct reasons to cast a vote:
 
 - weak coin
 
-    otherwise tortoise votes according to the coin recorded in the last layer. if sufficient number of honest votes were casted this way then tortoise will either reach decision or atleast vote according to the local threshold in future layers. if honest nodes disagreed on the coin value then the voting will continue until they agree.
+    otherwise tortoise votes according to the coin recorded in the last layer. if sufficient number of honest votes were cast this way then tortoise will either reach decision or atleast vote according to the local threshold in future layers. if honest nodes disagreed on the coin value then the voting will continue until they agree.
 
 ## Full tortoise
 
@@ -84,7 +86,7 @@ In the above example there are two ballots (0xaa and 0xbb). They vote for 4 bloc
 
 #### Votes from ballots with wrong beacon
 
-Because a smesher can select arbitrary beacon, there is an attack that will allow malicious smeshers to concentrate weight in a specific layer. If tortoise counts malicious weight immediatly it will make erroneous decisions (decide that a block is valid when in fact it is not). In order to prevent such attack tortoise will delay counting ballots with a wrong beacon until enough weight from honest smeshers was counted. 
+Because a smesher can select arbitrary beacon, there is an attack that will allow malicious smeshers to concentrate weight in a specific layer. If tortoise counts malicious weight immediately it will make erroneous decisions (decide that a block is valid when in fact it is not). In order to prevent such attack tortoise will delay counting ballots with a wrong beacon until enough weight from honest smeshers was counted. 
 
 Note that we can't discard ballots with wrong beacon completely as recovery from partition requires counting ballots from both sides. And they will likely have different beacons if it was a long partition, or occured at the time when tortoise beacon was computed.
 
@@ -94,13 +96,13 @@ Full tortoise complexity grows with the number of layers that are not finalized.
 
 During live-tortoise this can be a problem only in extreme cases when recent layer fails to be finalized. Due to the excessive malicious voting (> 2/3 fraction of total weight) or insufficient voting from honest smeshers (due to the network problems).
 
-However, during so called rerun (or sync) tortoise needs to count all votes. Otherwise there is always a risk that votes from latest layers changed decision that was made without counting them. With optimized full tortoise it might be practical to count votes when there are 2000 undecided layers, but not higher.
+However, during so called rerun (or sync) tortoise needs to count votes from all layers. Otherwise there is always a risk that votes from latest layers changed decision that was made without counting them. With optimized full tortoise it might be practical to count votes when there are 2000 undecided layers, but not higher.
 
 In practice during rerun layer is expected to be finalized after tortoise counted votes from some limited number of layers, this is referred to as verification window in the codebase.  
 
 ## Verifying tortoise
 
-Verifying tortoise is meant as a solution for (scaling-issues)[#scaling-issues]of the full tortoise.
+Verifying tortoise is meant as a solution for [scaling-issues](#scaling-issues) of the full tortoise.
 
 Intuitively if there is enough weight from ballots that vote consistently with each other to finalize a layer then tortoise doesn't full vote counting and can use more efficient batched method.
 
@@ -131,15 +133,15 @@ In this example only 0xbb ballot is consistent with local opinion. Consistency w
 - ballot votes are consistent with local opinion 
 - base ballot is good
 
-In the implementation we have a state named "can be good", the role of it's state will be clarified in [modes interaction](#modes-interaction) section. Ballot is marked "can be good" if first 3 condtions are satisfied, but not necessarily the last one.
+In the implementation we have a state named "can be good", the role of it's state will be clarified in [modes interaction](#modes-interaction) section. Ballot is marked "can be good" if first 3 conditions are satisfied, but not necessarily the last one.
 
 #### Vote counting in verifying mode
 
-Votes in verifying mode are counted in a batched form, unlike full tortoise. After "goodness" of the ballot is determinted tortoise filters out good ballots and sums up their weight. The result is a counted weight that is consistent with our local opinion. To get an accurate margin that can be compared with global threshold we assume that all uncounted ballots votes against our local opinion.
+Votes in verifying mode are counted in a batched form, unlike full tortoise. After "goodness" of the ballot is determinted tortoise filters good ballots and sums up their weight. The result is a counted weight that is consistent with our local opinion. To get an accurate margin that can be compared with global threshold we assume that all uncounted ballots votes against our local opinion.
 
 ## Modes interaction
 
-Verifying mode is safer as tortoise doesn't have an imposed limit on the amount of counted votes (or atleast imposed limit can be signicantly higher). Naturally we want to stay in verifying mode as long as tortoise can make progress.
+Verifying mode is safer as tortoise doesn't have an imposed limit on the amount of counted votes (or at least imposed limit can be signicantly higher). Naturally we want to stay in verifying mode as long as tortoise can make progress.
 
 #### When to switch into full mode?
 
@@ -151,7 +153,7 @@ During rerun we expect verifying tortoise to make progress after counting `verif
 
 It is safe to make an attempt at any point. However we need a sufficient number of ballots that vote consistently with each other in order for verifying tortoise to work. 
 
-In the implementation, we make an attempt after full tortoise made progress and changed local opinion on layers that were undecided according to the verifying tortoise. In the layer that was just finalized by full tortoise we filter ballots that ["can be good"](#good-ballots) with base ballots in the same state. After they are marked good we restart verifying tortoise and if same layer it made the same progress as full tortoise we switch the mode.
+In the implementation, we make an attempt after full tortoise made progress and changed local opinion on layers that were undecided according to the verifying tortoise. In the layer that was just finalized by full tortoise we find ballots that ["can be good"](#good-ballots) with base ballots in the same state. After they are marked good we restart verifying tortoise and if same layer it made the same progress as full tortoise we switch the mode.
 
 | block/ballot | local opinion | 0xaa | 0xbb | 0xcc | 0xdd | 0xee | 0xff |
 | ------------ | ------------- | ---- | ---- | ---- | ---- | ---- | ---- |
@@ -164,4 +166,4 @@ In the implementation, we make an attempt after full tortoise made progress and 
 | 0x55         | 1             | -    | -    | -    | -    | 1    | 1    |
 | 0x66         | -1            | -    | -    | -    | -    | -1   | -1   |
 
-In this example ballots starting from layer 10 are voting for some old layer (for example 9). Ballots from layer 10 disagree with out local opinion and will be marked bad. Ballot 0xcc from layer 10 and ballots 0xee and 0xff will be marked "can be good" because exceptions from those ballots are consistent with local opinion.
+In this example ballots starting from layer 10 are voting for some old layer (for example 9). Ballots from layer 10 disagree with our local opinion and will be marked bad. Ballot 0xcc from layer 10 and ballots 0xee and 0xff will be marked "can be good" because exceptions from those ballots are consistent with local opinion.
--- a/tortoise/README.md
+++ b/tortoise/README.md
@@ -1,0 +1,167 @@
+Tortoise is used to make decisions on the blocks that will be applied to the ledger. The protocol is asynchronous and makes progress based on the received data, without timing assumptions.
+
+## Core concepts
+
+#### Block
+
+From the tortoise perspective it is important that the block has a unique ID and belongs to some layer.
+
+In the rest of the system block is a wrapper for transactions and rewards. So once tortoise decides on a block - related transactions and rewards will be applied to the ledger.
+
+#### Ballot
+
+Tortoise counts votes from the syntactically valid ballots.
+
+For tortoise purposes every ballot must have a unique ID, reference to the base ballot ID, explicit votes (exceptions), weight and a beacon.
+
+###### Base ballot and exceptions
+
+Discussed in [Voting section](#voting).
+
+###### Weight
+
+Weight of the ballot is computed from the weight of the activation transaction divided by the total number of eligible ballots. Total number of eligible ballots is computed based on the fraction of actiovation transaction compared to the weight from the active set that was recorded in reference ballot.
+
+###### Beacon
+
+Ballot with an inconsistent beacon will not be marked good and will not be counted immediatly. Discussed in [full](#votes-from-ballots-with-wrong-beacon) and [verifying](#good-ballots) sections.
+
+#### Expected weight and thresholds
+
+Expected weight is a total weight from the activation transactions. Unlike weight for the ballot we are using a sum of all activation transactions that target a specific epoch. Expected weight for the layer is fraction of the expected weight for the epoch.
+
+## Voting
+
+When voting protocol chooses base ballot and encodes the difference between local opinion and base ballot votes. Since we don't keep votes from genesis in memory the implementation has a shortcoming that it encodes difference only within some configurable window (we can refer to it as voting window). 
+
+In order to minimize amount of data that needs to be sent and stored, Tortoise selects ballot that has the least difference with a local opinion. 
+
+Types of votes:
+- support
+
+    Support all blocks according to local opinion that are not supported by base ballot. Usually support is added for layers starting from base ballot layer up to last layer. In the event of healing tortoise may add support for blocks before base ballot layer.
+
+- against
+  
+    Against is added only if the base ballot supports the block and local opinion is against that block. 
+
+- abstain (undecided) 
+    
+    Undecided is a valid vote only until hare is terminated for the layer (it may take more than one layer for hare to terminate, this is expressed as zdist).
+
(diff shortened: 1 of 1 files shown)


dshulyak:
Please let me know if something is not sufficiently explained or poorly written.

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+In the rest of the system block is a wrapper for transactions and rewards. So once tortoise decides on a block - related transactions and rewards will be applied to the ledger.
+
+#### Ballot
+
+Tortoise counts votes from the syntactically valid ballots.
+
+For tortoise purposes every ballot must have a unique ID, reference to the base ballot ID, explicit votes (exceptions), weight and a beacon.
+
+###### Base ballot and exceptions
```
nkryuchkov: Although it's a header, it looks smaller than the text. Perhaps level 5 would be better?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Discussed in [Voting section](#voting).
+
+###### Weight
+
+Weight of the ballot is computed from the weight of the activation transaction divided by the total number of eligible ballots. Total number of eligible ballots is computed based on the fraction of actiovation transaction compared to the weight from the active set that was recorded in reference ballot.
```
nkryuchkov:
```suggestion
Weight of the ballot is computed from the weight of the activation transaction divided by the total number of eligible ballots. Total number of eligible ballots is computed based on the fraction of activation transaction compared to the weight from the active set that was recorded in reference ballot.
```

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+During live-tortoise this can be a problem only in extreme cases when recent layer fails to be finalized. Due to the excessive malicious voting (> 2/3 fraction of total weight) or insufficient voting from honest smeshers (due to the network problems).
+
+However, during so called rerun (or sync) tortoise needs to count all votes. Otherwise there is always a risk that votes from latest layers changed decision that was made without counting them. With optimized full tortoise it might be practical to count votes when there are 2000 undecided layers, but not higher.
+
+In practice during rerun layer is expected to be finalized after tortoise counted votes from some limited number of layers, this is referred to as verification window in the codebase.  
+
+## Verifying tortoise
+
+Verifying tortoise is meant as a solution for (scaling-issues)[#scaling-issues]of the full tortoise.
```
nkryuchkov: Not rendered as a link

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+| base         | -             | -    | -    | 0xaa | 0xaa | 0xcc | 0xcc |
+| layer        | -             | 10   | 10   | 11   | 11   | 12   | 12   |
+| 0x11         | 1             | -1   | -1   | 1    | -1   | 1    | 1    |
+| 0x22         | 1             | -1   | -1   | 1    | -1   | 1    | 1    |
+| 0x33         | -1            | -    | -    | -1   | -1   | -1   | -1   |
+| 0x44         | -1            | -    | -    | -1   | -1   | -1   | -1   |
+| 0x55         | 1             | -    | -    | -    | -    | 1    | 1    |
+| 0x66         | -1            | -    | -    | -    | -    | -1   | -1   |
+
+In this example ballots starting from layer 10 are voting for some old layer (for example 9). Ballots from layer 10 disagree with out local opinion and will be marked bad. Ballot 0xcc from layer 10 and ballots 0xee and 0xff will be marked "can be good" because exceptions from those ballots are consistent with local opinion.
```
nkryuchkov:
```suggestion
In this example ballots starting from layer 10 are voting for some old layer (for example 9). Ballots from layer 10 disagree with our local opinion and will be marked bad. Ballot 0xcc from layer 10 and ballots 0xee and 0xff will be marked "can be good" because exceptions from those ballots are consistent with local opinion.
```

nkryuchkov (approved):
Great docs, thanks

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Weight of the ballot is computed from the weight of the activation transaction divided by the total number of eligible ballots. Total number of eligible ballots is computed based on the fraction of actiovation transaction compared to the weight from the active set that was recorded in reference ballot.
+
+###### Beacon
+
+Ballot with an inconsistent beacon will not be marked good and will not be counted immediatly. Discussed in [full](#votes-from-ballots-with-wrong-beacon) and [verifying](#good-ballots) sections.
+
+#### Expected weight and thresholds
+
+Expected weight is a total weight from the activation transactions. Unlike weight for the ballot we are using a sum of all activation transactions that target a specific epoch. Expected weight for the layer is fraction of the expected weight for the epoch.
```
countvonzero, line 33: > Expected weight for the layer is fraction of the expected weight for the epoch.

it's exactly total epoch weight divided by layers per epoch. 
also, the epoch weight is precise. it should not be termed as "expected weight for the epoch"

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Discussed in [Voting section](#voting).
+
+###### Weight
+
+Weight of the ballot is computed from the weight of the activation transaction divided by the total number of eligible ballots. Total number of eligible ballots is computed based on the fraction of actiovation transaction compared to the weight from the active set that was recorded in reference ballot.
+
+###### Beacon
+
+Ballot with an inconsistent beacon will not be marked good and will not be counted immediatly. Discussed in [full](#votes-from-ballots-with-wrong-beacon) and [verifying](#good-ballots) sections.
```
countvonzero: maybe a short intro to beacon? 

`Beacon is a source of randomness and is calculated by the all smeshers per epoch. Ballots with a different beacon is either malicious or a result of a network split.`

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Expected weight and thresholds
+
+Expected weight is a total weight from the activation transactions. Unlike weight for the ballot we are using a sum of all activation transactions that target a specific epoch. Expected weight for the layer is fraction of the expected weight for the epoch.
+
```
countvonzero, line 34: thresholds are not mentioned. maybe a separate section for thresholds?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+From the tortoise perspective it is important that the block has a unique ID and belongs to some layer.
+
+In the rest of the system block is a wrapper for transactions and rewards. So once tortoise decides on a block - related transactions and rewards will be applied to the ledger.
+
+#### Ballot
+
+Tortoise counts votes from the syntactically valid ballots.
+
+For tortoise purposes every ballot must have a unique ID, reference to the base ballot ID, explicit votes (exceptions), weight and a beacon.
```
countvonzero, line 15: briefly describe what are in the exceptions?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Ballot with an inconsistent beacon will not be marked good and will not be counted immediatly. Discussed in [full](#votes-from-ballots-with-wrong-beacon) and [verifying](#good-ballots) sections.
+
+#### Expected weight and thresholds
+
+Expected weight is a total weight from the activation transactions. Unlike weight for the ballot we are using a sum of all activation transactions that target a specific epoch. Expected weight for the layer is fraction of the expected weight for the epoch.
+
+## Voting
+
+When voting protocol chooses base ballot and encodes the difference between local opinion and base ballot votes. Since we don't keep votes from genesis in memory the implementation has a shortcoming that it encodes difference only within some configurable window (we can refer to it as voting window).
```
countvonzero, line 37: what is local opinions? i think there should be a separate section in core concepts about `local opinions` and `global opinions`

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+## Voting
+
+When voting protocol chooses base ballot and encodes the difference between local opinion and base ballot votes. Since we don't keep votes from genesis in memory the implementation has a shortcoming that it encodes difference only within some configurable window (we can refer to it as voting window). 
+
+In order to minimize amount of data that needs to be sent and stored, Tortoise selects ballot that has the least difference with a local opinion. 
+
+Types of votes:
+- support
+
+    Support all blocks according to local opinion that are not supported by base ballot. Usually support is added for layers starting from base ballot layer up to last layer. In the event of healing tortoise may add support for blocks before base ballot layer.
```
countvonzero: what is last layer?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+- against
+  
+    Against is added only if the base ballot supports the block and local opinion is against that block. 
+
+- abstain (undecided) 
+    
+    Undecided is a valid vote only until hare is terminated for the layer (it may take more than one layer for hare to terminate, this is expressed as zdist).
+
+If ballot doesn't specify explicit vote for a block then it votes agains the block. This prevents malicious smeshers from retroactively reorganizing mesh by intentionally creating late ballots.
```
countvonzero: how does implicit against help prevent malicious reorg with late ballots? i thought this is about late blocks.

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
+Tortoise is used to make decisions on the blocks that will be applied to the ledger. The protocol is asynchronous and makes progress based on the received data, without timing assumptions.
+
+## Core concepts
```
countvonzero, line 3: i think it's worth adding 
- hdist
- zdist
- local opinions
- global opinions
- local threshold
- global threshold 
- live tortoise
- rerun tortoise
- sync tortoise
- verifying tortoise
- full tortoise
- self healing
in core concepts

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+If ballot doesn't specify explicit vote for a block then it votes agains the block. This prevents malicious smeshers from retroactively reorganizing mesh by intentionally creating late ballots.
+
+There are 4 distinct reasons to cast a vote:
+- hare output
+    
+    tortoise uses output from the hare consensus for voting on blocks within hdist.
+
+- validity/decisions
+
+    for layer before hdist the vote is casted according to the decision.
```
countvonzero: why should be have two terms for the same thing? i'd prefer to say validity

countvonzero: :s/casted/cast

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+- validity/decisions
+
+    for layer before hdist the vote is casted according to the decision. 
+
+- local threshold
+
+    in case when decision wasn't reached but margin from counted weight crosses local threshold tortoise will cast a vote according to the local threshold.
+
+- weak coin
```
countvonzero, line 69: how is weak coin generated?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Note that we can't discard ballots with wrong beacon completely as recovery from partition requires counting ballots from both sides. And they will likely have different beacons if it was a long partition, or occured at the time when tortoise beacon was computed.
+
+#### Scaling issues
+
+Full tortoise complexity grows with the number of layers that are not finalized. 
+
+During live-tortoise this can be a problem only in extreme cases when recent layer fails to be finalized. Due to the excessive malicious voting (> 2/3 fraction of total weight) or insufficient voting from honest smeshers (due to the network problems).
+
+However, during so called rerun (or sync) tortoise needs to count all votes. Otherwise there is always a risk that votes from latest layers changed decision that was made without counting them. With optimized full tortoise it might be practical to count votes when there are 2000 undecided layers, but not higher.
```
countvonzero: what does "count all votes" mean here? do you mean during rerun/sync we need to run full mode?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+In the implementation we have a state named "can be good", the role of it's state will be clarified in [modes interaction](#modes-interaction) section. Ballot is marked "can be good" if first 3 condtions are satisfied, but not necessarily the last one.
+
+#### Vote counting in verifying mode
+
+Votes in verifying mode are counted in a batched form, unlike full tortoise. After "goodness" of the ballot is determinted tortoise filters out good ballots and sums up their weight. The result is a counted weight that is consistent with our local opinion. To get an accurate margin that can be compared with global threshold we assume that all uncounted ballots votes against our local opinion.
+
+## Modes interaction
+
+Verifying mode is safer as tortoise doesn't have an imposed limit on the amount of counted votes (or atleast imposed limit can be signicantly higher). Naturally we want to stay in verifying mode as long as tortoise can make progress.
```
countvonzero: :s/atleast/at least

countvonzero: what is this `imposed limit on the amount of counted votes`?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+- ballot has a correct beacon
+- ballot doesn't vote for block before base ballot layer
+- ballot votes are consistent with local opinion 
+- base ballot is good
+
+In the implementation we have a state named "can be good", the role of it's state will be clarified in [modes interaction](#modes-interaction) section. Ballot is marked "can be good" if first 3 condtions are satisfied, but not necessarily the last one.
+
+#### Vote counting in verifying mode
+
+Votes in verifying mode are counted in a batched form, unlike full tortoise. After "goodness" of the ballot is determinted tortoise filters out good ballots and sums up their weight. The result is a counted weight that is consistent with our local opinion. To get an accurate margin that can be compared with global threshold we assume that all uncounted ballots votes against our local opinion.
```
countvonzero: :s/determinted/determined

countvonzero: filter out good ballots means to remove good ballots. how about just `After "goodness" of the ballot is determined tortoise sums up the weight of good ballots`

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Modes interaction
+
+Verifying mode is safer as tortoise doesn't have an imposed limit on the amount of counted votes (or atleast imposed limit can be signicantly higher). Naturally we want to stay in verifying mode as long as tortoise can make progress.
+
+#### When to switch into full mode?
+
+Tortoise certainly has to switch into full mode if there is a layer that wasn't finalized past hdist layers. In such case it needs to count previously uncounted ballots so that real margin can be compared with local threshold, see [voting](#voting). Without counting all votes we can get only pessimistic margin, which works for the purposes of the verifying tortoise, but may not work for self-healing purposes (e.g. instead of voting according to local threshold sign tortoise will vote according to weak coin).
```
countvonzero, line 148: > real margin can be compared with local threshold

do you mean global threshold?

countvonzero, line 148: > but may not work for self-healing purposes

this is the first time self-healing is mentioned in this doc. i think it needs a brief intro in core concept too.

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+During rerun we expect verifying tortoise to make progress after counting `verifying-mode-verification-window` layers. This window can be as large as memory will allow us, computational complexity doesn't increase with the window. However if tortoise didn't make progress after counting votes within this window it will have to switch into full mode during rerun. 
+
+#### When to switch into verifying mode?
+
+It is safe to make an attempt at any point. However we need a sufficient number of ballots that vote consistently with each other in order for verifying tortoise to work. 
+
+In the implementation, we make an attempt after full tortoise made progress and changed local opinion on layers that were undecided according to the verifying tortoise. In the layer that was just finalized by full tortoise we filter ballots that ["can be good"](#good-ballots) with base ballots in the same state. After they are marked good we restart verifying tortoise and if same layer it made the same progress as full tortoise we switch the mode.
```
countvonzero: :s/we filter ballots/we find ballots

tortoise/README.md
dshulyak: yes thanks it is about creating late blocks

tortoise/README.md
dshulyak: in this case i meant count votes from all layers, not necessarily all votes from all layers, that depends on how we count. will try to rephrase

tortoise/README.md
dshulyak: thanks, i guess i wanted to write filter good ballots, but your variant is better

tortoise/README.md
dshulyak: that's about verification window. need to link it somehow

tortoise/README.md
dshulyak: no, i meant that we if we have "real" margin it may cross local threshold, when "conservative" margin may not. so whenever we vote on undecided layers past hdist we want to compute "real" margin.

"real" is for margin computed using full tortoise. "conservative" is for - verifying.

tortoise/README.md
lrettig: Seconded - this document assumes a lot of prior knowledge and having way more context here would be helpful

tortoise/README.md
lrettig:
```suggestion
Ballot with an inconsistent beacon will not be marked good and will not be counted immediately. Discussed in [full](#votes-from-ballots-with-wrong-beacon) and [verifying](#good-ballots) sections.
```

tortoise/README.md
lrettig: Probably worth taking a step back here and explaining the concept of weight and why it matters here

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Expected weight and thresholds
+
+Expected weight is a total weight from the activation transactions. Unlike weight for the ballot we are using a sum of all activation transactions that target a specific epoch. Expected weight for the layer is fraction of the expected weight for the epoch.
+
+## Voting
+
+When voting protocol chooses base ballot and encodes the difference between local opinion and base ballot votes. Since we don't keep votes from genesis in memory the implementation has a shortcoming that it encodes difference only within some configurable window (we can refer to it as voting window). 
+
+In order to minimize amount of data that needs to be sent and stored, Tortoise selects ballot that has the least difference with a local opinion.
```
lrettig, line 39: How is this selection performed?

tortoise/README.md
lrettig: Why would a ballot choose to support a block not supported by the base ballot?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+In order to minimize amount of data that needs to be sent and stored, Tortoise selects ballot that has the least difference with a local opinion. 
+
+Types of votes:
+- support
+
+    Support all blocks according to local opinion that are not supported by base ballot. Usually support is added for layers starting from base ballot layer up to last layer. In the event of healing tortoise may add support for blocks before base ballot layer.
+
+- against
+  
+    Against is added only if the base ballot supports the block and local opinion is against that block.
```
lrettig, line 48: Why/under what circumstances might this happen?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+    Support all blocks according to local opinion that are not supported by base ballot. Usually support is added for layers starting from base ballot layer up to last layer. In the event of healing tortoise may add support for blocks before base ballot layer.
+
+- against
+  
+    Against is added only if the base ballot supports the block and local opinion is against that block. 
+
+- abstain (undecided) 
+    
+    Undecided is a valid vote only until hare is terminated for the layer (it may take more than one layer for hare to terminate, this is expressed as zdist).
```
lrettig, line 52: Why is this an invalid vote in other circumstances?

tortoise/README.md
lrettig:
```suggestion
If ballot doesn't specify explicit vote for a block then it votes against the block. This prevents malicious smeshers from retroactively reorganizing mesh by intentionally creating late ballots.
```

lrettig: I think this explanation needs to be fleshed out further. How does "implicit against" prevent this attack?

tortoise/README.md
lrettig:
```suggestion
    for layer before hdist the vote is cast according to the decision. 
```

lrettig: What decision? Why?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+    for layer before hdist the vote is casted according to the decision. 
+
+- local threshold
+
+    in case when decision wasn't reached but margin from counted weight crosses local threshold tortoise will cast a vote according to the local threshold.
+
+- weak coin
+
+    otherwise tortoise votes according to the coin recorded in the last layer. if sufficient number of honest votes were casted this way then tortoise will either reach decision or atleast vote according to the local threshold in future layers. if honest nodes disagreed on the coin value then the voting will continue until they agree.
```
lrettig:
```suggestion
    otherwise tortoise votes according to the coin recorded in the last layer. if sufficient number of honest votes were cast this way then tortoise will either reach decision or at least vote according to the local threshold in future layers. if honest nodes disagreed on the coin value then the voting will continue until they agree.
```

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Votes from ballots with wrong beacon
+
+Because a smesher can select arbitrary beacon, there is an attack that will allow malicious smeshers to concentrate weight in a specific layer. If tortoise counts malicious weight immediatly it will make erroneous decisions (decide that a block is valid when in fact it is not). In order to prevent such attack tortoise will delay counting ballots with a wrong beacon until enough weight from honest smeshers was counted. 
+
+Note that we can't discard ballots with wrong beacon completely as recovery from partition requires counting ballots from both sides. And they will likely have different beacons if it was a long partition, or occured at the time when tortoise beacon was computed.
+
+#### Scaling issues
+
+Full tortoise complexity grows with the number of layers that are not finalized.
```
lrettig, line 95: Why/how? Computational, memory, messages, or some other form of complexity?

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Because a smesher can select arbitrary beacon, there is an attack that will allow malicious smeshers to concentrate weight in a specific layer. If tortoise counts malicious weight immediatly it will make erroneous decisions (decide that a block is valid when in fact it is not). In order to prevent such attack tortoise will delay counting ballots with a wrong beacon until enough weight from honest smeshers was counted. 
+
+Note that we can't discard ballots with wrong beacon completely as recovery from partition requires counting ballots from both sides. And they will likely have different beacons if it was a long partition, or occured at the time when tortoise beacon was computed.
+
+#### Scaling issues
+
+Full tortoise complexity grows with the number of layers that are not finalized. 
+
+During live-tortoise this can be a problem only in extreme cases when recent layer fails to be finalized. Due to the excessive malicious voting (> 2/3 fraction of total weight) or insufficient voting from honest smeshers (due to the network problems).
```
lrettig, line 97: What is meant by live-tortoise?

tortoise/README.md
lrettig: How is the number 2000 derived here?

tortoise/README.md
lrettig:
```suggestion
Verifying tortoise is meant as a solution for (scaling-issues)[#scaling-issues] of the full tortoise.
```

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
...
+| 0x33         | 12    | 1       | 1    | 1    |
+| 0x44         | 13    | 1       | -1   | 1    |
+
+In this example only 0xbb ballot is consistent with local opinion. Consistency with a local opinion is a part of the definition of the good ballot, full definition is:
+- ballot has a correct beacon
+- ballot doesn't vote for block before base ballot layer
+- ballot votes are consistent with local opinion 
+- base ballot is good
+
+In the implementation we have a state named "can be good", the role of it's state will be clarified in [modes interaction](#modes-interaction) section. Ballot is marked "can be good" if first 3 condtions are satisfied, but not necessarily the last one.
```
lrettig:
```suggestion
In the implementation we have a state named "can be good", the role of its state will be clarified in [modes interaction](#modes-interaction) section. Ballot is marked "can be good" if first 3 conditions are satisfied, but not necessarily the last one.
```

tortoise/README.md
lrettig: What is meant by "safer" here? Why is it safer?

lrettig (dismissed):

tortoise/README.md
```diff
@@ -0,0 +1,167 @@
+Tortoise is used to make decisions on the blocks that will be applied to the ledger. The protocol is asynchronous and makes progress based on the received data, without timing assumptions.
+
+## Core concepts
+
+#### Block
+
+From the tortoise perspective it is important that the block has a unique ID and belongs to some layer.
+
+In the rest of the system block is a wrapper for transactions and rewards. So once tortoise decides on a block - related transactions and rewards will be applied to the ledger.
```
avive, line 9: q: are we 100% sure that rewards must be part of a block's content? We have discussed this many times in the past but I'm not sure what's the latest consensus on this question.

tortoise/README.md
dshulyak: because we don't have a limit on number of layers that we will count before the target layer is verified.

tortoise/README.md
dshulyak: live-tortoise is what we doing when it is not a rerun or sync

tortoise/README.md
dshulyak: mostly computational, for every new ballot we have to update all blocks in all undecided layers

tortoise/README.md
noamnelke: a: Yes, we're 100% sure we need explicit rewards under UCB. The reason is that blocks under this model have no reference to the smeshers that provided input for them. This information is known at the time of creating the block but it's ephemeral, so can't be used for making local decisions (crediting some account).

The rewards in the block will not specify an explicit amount: when constructing the block we can't know the exact fees collected - the transactions must be executed to know for sure. Instead, the rewards will specify a weight per coinbase account and the full layer reward (incl. fees) will be distributed between the accounts by weight.

This approach will also make it easier to provide rewards for other roles than layer committee member (the ones that create ballots and proposals), like Hare participants and one day validators. For now, we'll only reward layer committee members, though.

tortoise/README.md
dshulyak: it is actually described in the rest of the sentance

tortoise/README.md
dshulyak: this depends on the strategy for choosing base ballot. it might be more efficient to choose a ballot from later layer, and vote against blocks from previous layers. we are not doing this kind of optimizations

tortoise/README.md
dshulyak: i don't remember details. would prefer not to look them up

dshulyak:
wanted to work on it more, but honestly, i am getting bored to death describing these details. so will proceed to merge it as is and come back sometime in the future. fixed grammar and minor issues for now

dshulyak:
bors merge

PR #3033, opened 2021-12-28
[Merged by Bors] - log, cmd/node: add error code and args to fatal errors

### LINKED ISSUE #3024: Propagate an error to smapp on clock error
## Description

When a clock is out of sync, the node exits with exit code `1` and prints the following message to stdout: `Failed to initialize node.	{"errmsg": "check system clock drift: System clock is -3h0m3.648804859s away from NTP servers, please synchronize your OS time", "name": ""}`. Since the process has exited, the app can’t connect to GRPC API to retrieve such error there (or at least it almost doesn't have time to do that).

Therefore, on a crash, all the app can do is to show that some error has occurred. We need to have a reliable way to deliver what error occurred without making the app do complex log parsing.

@brusherru suggested a solution:
> Here is my vision of the ideal solution:
> 1. Determine in Node specification exit codes for all known errors and implement it. So “consumers” can show up the proper error message as they wish easily, without straight depending on messages from the Node. (E.G. it will be very useful if we will support multiple languages in Smapp)
> 2. To provide some details about the error to consumers, the Node should send it to the `stderr` (since it is the only way to tell something while the process is exiting) in  the JSON with specific shape and extracted details to the separate properties. For this particular case it may look like: `{"message": "System clock is -3h0m3.648804859s away from NTP servers", "solution": "Please synchronize your OS time" "details": {"clock": "-3h0m3.648804859s" }, "name": "NTP_UNSYNC", "exitCode": 31}` . Pay attention that I’ve added `details` property which may contain different set of data (for different errors), and it contains `clock` property, so the consumer doesn’t need to extract such value using RegExp or something similar.
>     - Btw, probably it will be better to use more “computer-friendly” format like timestamp. E.G. `nanos: -10803648804859` , but I’m not sure about it.
>     - Also I’m not sure that the `exitCode` should be here, but probably if some errors can occur sending a bunch of errors to the `stderr` before closing and then it closes with some exitCode — it will be easier to find out which one to show. But, again, I’m not sure about it.
> 3. To make easier for consumers to parse the JSON with error details, we can wrap it in special characters, like `STX` (`\x02`) and `ETX`  (`\x03`). So consumers won’t need to do a tricky parsing, just start reading everything from `\x02` and stop on `\x03` and parse everything that was between as a JSON. However, I’m not sure about this particular special characters, probably will be better to use something else. Research needed.
>
> All these stuff sounds a bit complicated, so probably we should make a temporary solution with just extracting an `errmsg` from `stderr` in case if the process exits with code >0. However, I’m sure that the solution proposed above will work much better 🙂

cc @brusherru @moshababo @avive

### LINKED ISSUE #3027: log: add a field with error text code
## Description

Introduce a new field in `FATAL` logs, the field has a common name and contains an error text code like `ERR_CLOCK_AWAY`. With it, the app knows what error occurred and, for example, shows a possible solution. The error text code can also be used for localization (e.g. in the future we can have a web service that accepts a text code (e.g. `ERR_CLOCK_AWAY`) and a language code (e.g. `en-us`) and returns a text in that language (e.g. `Clock is away from NTP servers`). The field is urgent.

https://github.com/spacemeshos/go-spacemesh/issues/3024#issuecomment-1001620954

## Affected code

Logging

`9972b2c9c7ee79a2ef913eeb0d8ead05ee74769c`

### LINKED ISSUE #3028: log: arguments for error text codes
## Description

Implement arguments for error text codes sent to the app to provide additional context. E.g. we can pass what time node is away from NTP servers for the `ERR_CLOCK_AWAY` error. This is urgent but depends on [implementing the error text code](https://github.com/spacemeshos/go-spacemesh/issues/3027).

https://github.com/spacemeshos/go-spacemesh/issues/3024#issuecomment-1001620954

## Affected code

Logging

`9972b2c9c7ee79a2ef913eeb0d8ead05ee74769c`

nkryuchkov:
## Motivation

Closes #3027 
Closes #3028
Closes #3024

## Changes

- add error code to fatal errors
- add args to fatal errors
- create error definitions for some common errors

## Test Plan

Unit tests

## DevOps Notes


dshulyak:
i think there is a more generic way to structure errors, there is no need for fatalError type. each error can have its own type with different details, also this mechanism can be reused for any errors that want to provide more details instead of simple message.

for example:

```go
type clockErrorDetails struct {
	Drift time.Duration
}

func (c *clockErrorDetails) MarshalLogObject(encoder log.ObjectEncoder) error {
	encoder.AddDuration("drift", c.Drift)
	return nil
}

type clockError struct {
	err     error
	details clockErrorDetails
}

func (c *clockError) MarshalLogObject(encoder log.ObjectEncoder) error {
	encoder.AddString("code", "CLOCK")
	encoder.AddString("error", c.err.Error())
	encoder.AddObject("details", &c.details)
	return nil
}

func Err(err error) {
	var loggable log.ObjectMarshaller
	if errors.As(err, &loggable) {
		return zap.Inline(loggable)
	}
	return zap.NamedError("errmsg", err)
}
```

dshulyak (changes requested):
in prev comment

log/fatal_test.go
```diff
@@ -0,0 +1,58 @@
...
+	got := Entry{}
+
+	defer func() {
+		if r := recover(); r != nil {
+			require.NoError(t, json.Unmarshal(buf.Bytes(), &got))
+			require.EqualValues(t, expect, got)
+		}
+	}()
+
+	logger.With().Fatal(testStr, Err(ErrClockAway(testDur)))
```
dshulyak: but it exits with os.Exit , how recover in defer will help with that?

log/fatal.go
```diff
@@ -0,0 +1,64 @@
+package log
+
+import (
+	"fmt"
+)
+
+// Common errors that can happen on node startup.
+var (
+	ErrMalformedConfig    = newFatalError("ERR_MALFORMED_CONFIG", "config file is malformed: %v")
```
dshulyak: let's use a code only for errors that are meaningful for smapp. for example ERR_START_PROFILING doesn't make sense

log/fatal_test.go
nkryuchkov: `WithOptions(zap.OnFatal(zapcore.WriteThenPanic))` panics instead

brusherru:
For cross-links and clearness, this PR is related to #3024 as well.

nkryuchkov:
@brusherru it actually closes #3024, I wrote that in the description

log/fatal.go
brusherru: @dshulyak 
If we're talking only about solving the current issue — that's right.
However, I think that the homogeneous shape of fatal errors is a good idea. And if we took this is for the standard for fatal errors, then developers do not have to think anymore "Do somebody need an error code for custom handling of such errors?".

dshulyak: no, fatal error is not something that should be added in the first place. any error can lead to fatal exit. code is needed only if smapp has a special handling for some of the codes.

cmd/node/node.go
```diff
@@ -294,16 +295,18 @@ func (app *App) Initialize() (err error) {
 			log.Int("hare_wakeup_delta", app.Config.HARE.WakeupDelta),
 			log.Int("hare_limit_iterations", app.Config.HARE.LimitIterations),
 			log.Int("hare_round_duration", app.Config.HARE.RoundDuration))
-		return errors.New("incompatible tortoise hare params")
+
+		desc := "layer-duration-sec * tortoise-zdist <= hare-wakeup-delta + (1 + hare-limit-iterations * 4) * hare-round-duration-sec"
+		return log.ErrTortoiseHareParams(desc)
```
dshulyak: please no need to invent something weird with these wrappers. 

there are errors that have custom handling in smapp and they should have code, for every other error code is not needed and smapp should provide only information available in the error. we definitely won't be adding code to every possible error that is returned across the stack.

another thing to consider that some errors will want to log their details in a structured way instead of flat string. those errors should implement interface the way it was described above.

log/fatal.go
nkryuchkov: @dshulyak What are the disadvantages of having errors like `ERR_START_PROFILING`? With the homogeneous implementation, they are cheap to add, so smapp can decide whether or not it needs them

dshulyak: disadvantage is that now every error must be have a code. which is a nonsense

nkryuchkov: Why? It must not. If an error doesn't have a code, it works as before this PR

dshulyak: this is pretty insignificant problem that will impose requirements for writing errors across whole application. we can cover couple of errors that smapp wants to present in a nice way (like clock issue), but in the general case it will have to present information that it is available in the log.

honestly i think that even clock thing doesn't require custom presentation. just show what is in the log

dshulyak: > Why? It must not. If an error doesn't have a code, it works as before this PR

then what was the reason to add those weird wrappers in the first place? let them work as before

dshulyak: to clarify. you did that because you went over the source and added code to errors that could lead to fatal exit. however this list is not exhaustive. so some errors that will lead to exit will have code and some not. tomorrow i will add another error in some package that could lead to an exit and i will need a code. this is clearly nonsense.

log/fatal.go
```diff
@@ -0,0 +1,64 @@
...
+	ErrBadFlags           = newFatalError("ERR_BAD_FLAGS", "bad CLI flags: %v")
+	ErrTortoiseHareParams = newFatalError("ERR_TORTOISE_HARE_PARAMS", "incompatible tortoise hare params: %v")
+	ErrEnsureDataDir      = newFatalError("ERR_ENSURE_DATA_DIR", "could not open/create data dir %v: %v")
```
dshulyak: templates like this `could not open/create data dir %v: %v` should be defined where the error is created, not in some random module

log/fatal.go
nkryuchkov: > this is pretty insignificant problem that will impose requirements for writing errors across whole application

I don't understand why you think that those requirements will be imposed. 

> honestly i think that even clock thing doesn't require custom presentation

Do you mean the wrappers or anything else by "custom presentation"?

> then what was the reason to add those weird wrappers in the first place? let them work as before

What do you disagree with? The way the wrappers are implemented or adding more context to errors other that clock error?

> tomorrow i will add another error in some package that could lead to an exit and i will need a code

You will be able to create a usual error like we used to do, the error code just won't be shown in logs and the smapp wouldn't know how to classify the error. What are you suggesting instead? Not to add exit codes? But the result would be the same as in case you described: code is not passed, smapp handles it like a usual non-classified error

log/fatal.go
nkryuchkov: I agree that they can be moved. I'll do that. I was in doubt when implementing it but thought it would be convenient to have all error codes in one place

log/fatal.go
dshulyak: i suggest adding a code for a specific subset of errors that will be known to smapp. currently it will be only clock error. otherwise smapp should just present an error that is specified in fatal log.

as for implementation please see example above, it allows to encode descriptive object without args

nkryuchkov:
@dshulyak Could you please review the latest changes? Is that what you meant?

cmd/node/node.go
```diff
@@ -280,6 +280,34 @@ func (app *App) introduction() {
...
+}
+
+type clockError struct {
+	err     error
+	details clockErrorDetails
+}
+
+func (c *clockError) MarshalLogObject(encoder log.ObjectEncoder) error {
+	encoder.AddString("code", "ERR_CLOCK_DRIFT")
+	encoder.AddString("error", c.err.Error())
```
dshulyak: maybe it needs to be `errmsg` so that it remains compatible with other errors

cmd/node/node.go
```diff
@@ -280,6 +280,34 @@ func (app *App) introduction() {
...
+	return nil
+}
+
+type clockError struct {
+	err     error
+	details clockErrorDetails
+}
+
+func (c *clockError) MarshalLogObject(encoder log.ObjectEncoder) error {
+	encoder.AddString("code", "ERR_CLOCK_DRIFT")
```
dshulyak: maybe codes can be defined in a separate module, but this is not important for now i guess

dshulyak (approved):

cmd/node/node.go
nkryuchkov: @dshulyak Do you think such a module should be inside the `log` package?

nkryuchkov:
bors merge

PR #3038, opened 2021-12-29
tortoise: count abstain votes as consistent with any opinion

### LINKED ISSUE #3032: tortoise: abstain votes are consistent with any local opinion
If ballot has an abstain vote it will definitely won't be consistent during rerun, and it will break verifying mode. 

determineGoodness method needs to be adjusted, so that abstain vote is consistent with anything.

dshulyak:
The problem is that goodness definition as it is implemented right now is not compatible with abstain votes and zdist. So if tortoise votes for a layer that is not yet terminated - during rerun node will be forced into full mode. It could be also forced into full mode during live-run but it is less critical because full mode is not expensive at that point  

closes: https://github.com/spacemeshos/go-spacemesh/issues/3032

changes:
- abstain vote is consistent with any opinion

### DIFF
--- a/tortoise/verifying.go
+++ b/tortoise/verifying.go
@@ -178,6 +178,11 @@ func (v *verifying) determineGoodness(logger log.Log, ballot tortoiseBallot) goo
 			return bad
 		}
 
+		// by protocol abstain vote is consistent with any opinion
+		if vote == abstain {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


dshulyak:
bors try

dshulyak:
superseded by #3045

PR #3041, opened 2021-12-30
[Merged by Bors] - tortoise: recompute verification window when switching into full mode

### LINKED ISSUE #3009: tortoise: properly recompute verification window when switching from verifying to full
The purpose of this window is to limit number of votes that we need to count 
before deciding on the layer. In live-tortoise we decide on the layer immediatly once
it crosses global threshold. However our assumption is that majority is honest is valid 
only on long distance. Single layer can be verified by weight from dishonest majority, 
or simply because part of the ballots were not received.

Let's name this as a verification window, for verifying tortoise it can be large because of
the reduced complexity. For full tortoise it is much smaller, target is 2000 but implementation 
is not efficient enough to support that yet.

Verification window must be applied when last - target > verification window. Target is the next layer
that we want to verify. If target is consisntely verified then the window will be equal to configured param.
If target gets stuck, because of conflicting votes for example, the window may grow beyond configured.

When we switch from verifying to full this window needs to be recomputed. But if the target was stuck
then we will force full tortoise to count votes for the whole distance when it was stuck. 
For example if verifying tortoise window is 1000, we may force full tortoise to count votes for
the 1000 layers, which currently will take very long time.

For the next network, i will reduce verifying tortoise window to a 100.

dshulyak:
## Motivation

verifying has a large verification window (think 1_000_000) and if it failed to verify layer the threshold will be computed according to that window. if we won't reset the threshold full tortoise will have to count votes for 1_000_000 layers before
any layer can be expected to get verified. this is infeasible given current performance of the full tortoise and may take weeks to finish.

closes https://github.com/spacemeshos/go-spacemesh/issues/3009

## Changes
- compute thresholds according to the full tortoise configuration when in full mode

## Test Plan
TestRerunDistanceVoteCounting

### DIFF
--- a/tortoise/full.go
+++ b/tortoise/full.go
@@ -93,7 +93,7 @@ func (f *full) countVotesFromBallots(logger log.Log, ballotlid types.LayerID, ba
 	}
 }
 
-func (f *full) countVotes(logger log.Log) {
+func (f *full) countLayerVotes(logger log.Log, lid types.LayerID) {
 	for front := f.delayedQueue.Front(); front != nil; {
 		delayed := front.Value.(delayedBallots)
 		if f.last.Difference(delayed.lid) <= f.BadBeaconVoteDelayLayers {
@@ -110,8 +110,12 @@ func (f *full) countVotes(logger log.Log) {
 		f.delayedQueue.Remove(front)
 		front = next
 	}
+	f.countVotesFromBallots(logger, lid, f.ballots[lid])
+}
+
+func (f *full) countVotes(logger log.Log) {
 	for lid := f.counted.Add(1); !lid.After(f.processed); lid = lid.Add(1) {
-		f.countVotesFromBallots(logger, lid, f.ballots[lid])
+		f.countLayerVotes(logger, lid)
 	}
 	f.counted = f.processed
 }
--- a/tortoise/full.go
+++ b/tortoise/full.go
@@ -128,7 +128,7 @@ func (f *full) countVotes(logger log.Log) {
 func (f *full) verify(logger log.Log, lid types.LayerID) bool {
 	logger = logger.WithFields(
 		log.String("verifier", fullTortoise),
-		log.Stringer("processed_layer", f.processed),
+		log.Stringer("counted_layer", f.counted),
 		log.Stringer("candidate_layer", lid),
 		log.Stringer("local_threshold", f.localThreshold),
 		log.Stringer("global_threshold", f.globalThreshold),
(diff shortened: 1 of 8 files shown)


tortoise/full.go
```diff
@@ -93,7 +93,12 @@ func (f *full) countVotesFromBallots(logger log.Log, ballotlid types.LayerID, ba
 	}
 }
 
-func (f *full) countVotes(logger log.Log) {
+func (f *full) countLayerVotes(logger log.Log, lid types.LayerID) {
+	if lid.After(f.counted) {
+		f.counted = lid
+	} else {
+		return
+	}
```
nkryuchkov:
```suggestion
	if !lid.After(f.counted) {
		return
	}
	f.counted = lid
```

tortoise/threshold.go
```diff
@@ -98,39 +98,20 @@ func computeThresholdForLayers(config Config, epochWeight map[types.EpochID]weig
...
+	return last
 }
 
-// updateThresholds recomputes local and global thresholds:
-// - when last layer is updated
-// - when verified layer is updated
-// - when switching from one mode into the other.
-func updateThresholds(logger log.Log, config Config, state *commonState, tmode mode) {
-	state.localThreshold = computeLocalThreshold(config, state.epochWeight, state.last)
+func computeThresholds(logger log.Log, config Config, tmode mode, target, last, processed types.LayerID, epochWeight map[types.EpochID]weight) (weight, weight) {
```
nkryuchkov: Looking at the signature, it's not clear what weights are returned. What do you think of something like the below?
```suggestion
func computeThresholds(logger log.Log, config Config, tmode mode, target, last, processed types.LayerID, epochWeight map[types.EpochID]weight) (local, global weight) {
```

tortoise/tortoise_test.go
```diff
@@ -1366,15 +1366,22 @@ func TestGetBallotBeacon(t *testing.T) {
...
-		panic("need atleast 2 layers")
+func gapVote(rng *mrand.Rand, layers []*types.Layer, i int) sim.Voting {
+	return skipLayers(1)(rng, layers, i)
+}
+
+func skipLayers(n int) sim.VotesGenerator {
+	return func(rng *mrand.Rand, layers []*types.Layer, _ int) sim.Voting {
+		position := n + 1
+		if len(layers) < position {
+			panic(fmt.Sprintf("need atleast %d layers", position))
```
nkryuchkov:
```suggestion
			panic(fmt.Sprintf("need at least %d layers", position))
```

nkryuchkov (approved):

tortoise/threshold.go
dshulyak: makes sense :+1:

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors merge

PR #3042, opened 2022-01-03
[Merged by Bors] - mesh, hare: fallback to sync if we don't have data to process the layer

### LINKED ISSUE #3039: mesh: panic with missing tx's when block applied to the state
## Description
Node unexpectedly exited yesterday, while it syncing and smeshing.

## Steps to reproduce
1. Run the Smapp (or probably just run Node is enough)
2. Setup smeshing. I'm not sure that this is a required step to reproduce, but my Node does it when it happened.
3. Wait...

## Actual Behavior
Panic

## Expected Behavior
Everything works fine without any panics :)

## Environment
 - OS: macOS 11.6 (20G165)
 - Node Version: 0.2.10-beta.0
 - Smapp version: 0.2.0-beta.2

## Additional Resources 
Here is a part of the log:
```
2021-12-29T02:15:05.093+0530	INFO	8be19.trtl         	attempting eviction	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "trtl", "rerun": true, "last-layer-id": 212, "protocol": "HARE_PROTOCOL_outputloop", "sessionId": "a7994ab7-629e-4f9e-96e8-4aceb61c330c", "effective_genesis": 11, "hdist": 10, "verified": 224, "window_size": 30, "last_evicted": 193, "window_start": 194, "name": "trtl"}
2021-12-29T02:15:05.093+0530	INFO	8be19.trtl         	finished handling incoming layer	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "trtl", "protocol": "HARE_PROTOCOL_outputloop", "sessionId": "a7994ab7-629e-4f9e-96e8-4aceb61c330c", "old_pbase": 223, "new_pbase": 224, "incoming_layer": 230, "name": "trtl"}
2021-12-29T02:15:05.093+0530	INFO	8be19.mesh         	tortoise results	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "mesh", "protocol": "HARE_PROTOCOL_outputloop", "sessionId": "a7994ab7-629e-4f9e-96e8-4aceb61c330c", "layer_id": 225, "reverted": false, "old_pbase": 223, "new_pbase": 224, "name": "mesh"}
2021-12-29T02:15:05.422+0530	INFO	8be19.trtl         	persist tortoise	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "trtl", "protocol": "HARE_PROTOCOL_outputloop", "sessionId": "a7994ab7-629e-4f9e-96e8-4aceb61c330c", "duration": "328.949667ms", "name": "trtl"}
2021-12-29T02:15:05.422+0530	INFO	8be19.mesh         	pushing layers to state	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "mesh", "protocol": "HARE_PROTOCOL_outputloop", "sessionId": "a7994ab7-629e-4f9e-96e8-4aceb61c330c", "old_pbase": 223, "new_pbase": 224}
2021-12-29T02:15:05.423+0530	WARN	8be19.meshDb       	could not fetch tx	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "meshDb", "tx_id": "91643dce03d0e0f5937ef6d997b45514333ac24f0f119dc9b6edc62e1993f058", "errmsg": "could not find transaction in database 91643dce03d0e0f5937ef6d997b45514333ac24f0f119dc9b6edc62e1993f058 err=get value: leveldb: not found", "name": "meshDb"}
2021-12-29T02:15:05.423+0530	ERROR	8be19.mesh         	Fatal: goroutine panicked. Stacktrace: goroutine 368 [running]:
runtime/debug.Stack(0xc99d110f4fc23a33, 0x58f093192ec6edb6, 0xf5e0d003ce3d6491)
	/Users/runner/hostedtoolcache/go/1.15.13/x64/src/runtime/debug/stack.go:24 +0x9f
github.com/spacemeshos/go-spacemesh/log.Log.Panic(0xc0001e5e00, 0x4fd6ebd, 0x4, 0x500ecf6, 0x2c, 0xc0025b8a18, 0x2, 0x2)
	/Users/runner/work/go-spacemesh/go-spacemesh/log/zap.go:45 +0x9e
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).getTxs(0xc000408620, 0xc00a320b80, 0x1, 0x1, 0xc0000000e0, 0x1, 0x1, 0xba50354f8b569681)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:727 +0x152
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).extractUniqueOrderedTransactions(0xc000408620, 0xc00d4aef00, 0x49ea48f, 0xc001d7a73e, 0x20)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:698 +0x385
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).applyState(0xc000408620, 0xc00d4aef00)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:502 +0x50
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).updateStateWithLayer(0xc000408620, 0x52af380, 0xc00add3ad0, 0xc00d4aef00)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:632 +0x445
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).pushLayersToState(0xc000408620, 0x52af380, 0xc00add3ad0, 0xe0000000df)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:465 +0x407
github.com/spacemeshos/go-spacemesh/mesh.(*validator).ValidateLayer(0xc008781440, 0x52af380, 0xc00add3ad0, 0xc00fc12640)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:348 +0x889
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).HandleValidatedLayer(0xc000408620, 0x52af380, 0xc00add3ad0, 0xe1, 0x0, 0x0, 0x0)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:587 +0x5b6
github.com/spacemeshos/go-spacemesh/hare.(*Hare).collectOutput(0xc0002d3200, 0x52af380, 0xc00add3ad0, 0x52b0280, 0xc00fc12200, 0x0, 0x0)
	/Users/runner/work/go-spacemesh/go-spacemesh/hare/hare.go:212 +0x2ed
github.com/spacemeshos/go-spacemesh/hare.(*Hare).outputCollectionLoop(0xc0002d3200, 0x52af380, 0xc00b6b06f0)
	/Users/runner/work/go-spacemesh/go-spacemesh/hare/hare.go:404 +0x67b
created by github.com/spacemeshos/go-spacemesh/hare.(*Hare).Start
	/Users/runner/work/go-spacemesh/go-spacemesh/hare/hare.go:468 +0xa25
	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "mesh"}
2021-12-29T02:15:05.423+0530	PANIC	8be19.mesh         	could not find transactions map[0x91643dce03d0e0f5937ef6d997b45514333ac24f0f119dc9b6edc62e1993f058:{}] from layer 224	{"node_id": "8be196e35895af351f21bf950b27d8a45f13d0d17b28c7e01f740de4d50dc392", "module": "mesh"}
panic: could not find transactions map[0x91643dce03d0e0f5937ef6d997b45514333ac24f0f119dc9b6edc62e1993f058:{}] from layer 224

goroutine 368 [running]:
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc0060e4c00, 0x0, 0x0, 0x0)
	/Users/runner/go/pkg/mod/go.uber.org/zap@v1.19.0/zapcore/entry.go:232 +0x51f
go.uber.org/zap.(*SugaredLogger).log(0xc0025b88c0, 0x804, 0x500ecf6, 0x2c, 0xc0025b8a18, 0x2, 0x2, 0x0, 0x0, 0x0)
	/Users/runner/go/pkg/mod/go.uber.org/zap@v1.19.0/sugar.go:227 +0x111
go.uber.org/zap.(*SugaredLogger).Panicf(...)
	/Users/runner/go/pkg/mod/go.uber.org/zap@v1.19.0/sugar.go:159
github.com/spacemeshos/go-spacemesh/log.Log.Panic(0xc0001e5e00, 0x4fd6ebd, 0x4, 0x500ecf6, 0x2c, 0xc0025b8a18, 0x2, 0x2)
	/Users/runner/work/go-spacemesh/go-spacemesh/log/zap.go:46 +0x21a
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).getTxs(0xc000408620, 0xc00a320b80, 0x1, 0x1, 0xc0000000e0, 0x1, 0x1, 0xba50354f8b569681)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:727 +0x152
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).extractUniqueOrderedTransactions(0xc000408620, 0xc00d4aef00, 0x49ea48f, 0xc001d7a73e, 0x20)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:698 +0x385
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).applyState(0xc000408620, 0xc00d4aef00)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:502 +0x50
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).updateStateWithLayer(0xc000408620, 0x52af380, 0xc00add3ad0, 0xc00d4aef00)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:632 +0x445
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).pushLayersToState(0xc000408620, 0x52af380, 0xc00add3ad0, 0xe0000000df)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:465 +0x407
github.com/spacemeshos/go-spacemesh/mesh.(*validator).ValidateLayer(0xc008781440, 0x52af380, 0xc00add3ad0, 0xc00fc12640)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:348 +0x889
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).HandleValidatedLayer(0xc000408620, 0x52af380, 0xc00add3ad0, 0xe1, 0x0, 0x0, 0x0)
	/Users/runner/work/go-spacemesh/go-spacemesh/mesh/mesh.go:587 +0x5b6
github.com/spacemeshos/go-spacemesh/hare.(*Hare).collectOutput(0xc0002d3200, 0x52af380, 0xc00add3ad0, 0x52b0280, 0xc00fc12200, 0x0, 0x0)
	/Users/runner/work/go-spacemesh/go-spacemesh/hare/hare.go:212 +0x2ed
github.com/spacemeshos/go-spacemesh/hare.(*Hare).outputCollectionLoop(0xc0002d3200, 0x52af380, 0xc00b6b06f0)
	/Users/runner/work/go-spacemesh/go-spacemesh/hare/hare.go:404 +0x67b
created by github.com/spacemeshos/go-spacemesh/hare.(*Hare).Start
	/Users/runner/work/go-spacemesh/go-spacemesh/hare/hare.go:468 +0xa25
```
The full log for the Smapp session: [panic-log.txt.zip](https://github.com/spacemeshos/go-spacemesh/files/7788083/panic-log.txt.zip)

dshulyak:
## Motivation

when blocks are applied to the state node needs to have all transactions that are referenced by those blocks. blocks are applied to the state only after tortoise validates them. so if at the time of validation we are missing transactions from blocks that are valid according to the tortoise node needs to fall back to sync and loop until it fetches all transactions referenced by valid blocks.

why not to enforce data availability when blocks are received from the network (e.g. in handler)?
--

any node can propose a block and reference non-existing transactions. because of that, we can't loop in sync indefinitely until we download transactions for all blocks that are sent to us from peers. at some point, sync has to give up on fetching transactions from any random block, and if the node misses some transactions it should vote against the block. however if according to the tortoise block is valid then it means that majority of the smeshers have transactions and node should continue trying to download transactions from that block, as it doesn't have any other choice. 

closes: https://github.com/spacemeshos/go-spacemesh/issues/3039

## Changes
- if valid blocks or transactions from valid blocks are missing terminate layer verification
- allow sync to pick up layer immediately without waiting additional time
- add a variable to the mesh called MissingLayer, it is updated if layer is failed to be applied to the state and reset to null if this layer eventually succeeds
- sync watches MissingLayer variable and will try to download it when the layer is not null value

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -10,7 +10,6 @@ import (
 	"sync"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
-	"github.com/spacemeshos/go-spacemesh/common/util"
 	"github.com/spacemeshos/go-spacemesh/database"
 	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/log"
@@ -287,25 +286,36 @@ func (vl *validator) ProcessLayer(ctx context.Context, layerID types.LayerID) er
 		log.FieldNamed("old_verified", oldVerified),
 		log.FieldNamed("new_verified", newVerified))
 
+	// set processed layer even if later code will fail, as that failure is not related
+	// to the layer that is being processed
+	vl.setProcessedLayer(layerID)
+
 	// check for a state reversion: if tortoise reran and detected changes to historical data, it will request that
 	// state be reverted and reapplied. pushLayersToState, below, will handle the reapplication.
 	if reverted {
+		vl.setLatestLayerInState(oldVerified)
 		if err := vl.revertState(ctx, oldVerified); err != nil {
 			logger.With().Error("failed to revert state, unable to process layer", log.Err(err))
 			return err
 		}
 	}
-
-	if newVerified.After(oldVerified) {
-		if err := vl.pushLayersToState(ctx, oldVerified, newVerified); err != nil {
+	// mesh can't skip layer that failed to complete
+	from := minLayer(oldVerified, vl.latestLayerInState)
+	to := newVerified
+	logger.With().Debug("pushing layers to the state",
+		log.Stringer("from_layer", from),
+		log.Stringer("to_layer", to),
+	)
+	if to.After(from) {
+		if err := vl.pushLayersToState(ctx, from, to); err != nil {
 			logger.With().Error("failed to push layers to state", log.Err(err))
 			return err
 		}
-		if err := vl.persistLayerHashes(ctx, oldVerified.Add(1), newVerified); err != nil {
+		if err := vl.persistLayerHashes(ctx, from.Add(1), to); err != nil {
 			logger.With().Error("failed to persist layer hashes", log.Err(err))
 			return err
 		}
-		for lid := oldVerified.Add(1); !lid.After(newVerified); lid = lid.Add(1) {
+		for lid := from.Add(1); !lid.After(to); lid = lid.Add(1) {
 			events.ReportLayerUpdate(events.LayerUpdate{
 				LayerID: lid,
 				Status:  events.LayerStatusTypeConfirmed,
@@ -313,7 +323,6 @@ func (vl *validator) ProcessLayer(ctx context.Context, layerID types.LayerID) er
 		}
 	}
 
-	vl.setProcessedLayer(layerID)
 	logger.Info("done processing layer")
 	return nil
 }
@@ -365,13 +374,12 @@ func (msh *Mesh) persistLayerHashes(ctx context.Context, from, to types.LayerID)
 
 func (msh *Mesh) getValidBlockIDs(ctx context.Context, layerID types.LayerID) ([]types.BlockID, error) {
 	logger := msh.WithContext(ctx)
-	lyr, err := msh.GetLayer(layerID)
+	blocks, err := msh.LayerBlockIds(layerID)
 	if err != nil {
-		logger.With().Error("failed to get layer", layerID, log.Err(err))
 		return nil, err
 	}
 	var validBlockIDs []types.BlockID
-	for _, bID := range lyr.BlocksIDs() {
+	for _, bID := range blocks {
 		valid, err := msh.ContextualValidity(bID)
 		if err != nil {
 			// block contextual validity is determined by layer. if one block in the layer is not determined,
@@ -387,24 +395,23 @@ func (msh *Mesh) getValidBlockIDs(ctx context.Context, layerID types.LayerID) ([
 }
 
 // apply the state of a range of layers, including re-adding transactions from invalid blocks to the mempool.
-func (msh *Mesh) pushLayersToState(ctx context.Context, oldVerified, newVerified types.LayerID) error {
+func (msh *Mesh) pushLayersToState(ctx context.Context, from, to types.LayerID) error {
 	logger := msh.WithContext(ctx).WithFields(
-		log.FieldNamed("old_verified", oldVerified),
-		log.FieldNamed("new_verified", newVerified))
+		log.Stringer("from_layer", from),
+		log.Stringer("to_layer", to))
 	logger.Info("pushing layers to state")
-	if oldVerified.Before(types.GetEffectiveGenesis()) || newVerified.Before(types.GetEffectiveGenesis()) {
+	if from.Before(types.GetEffectiveGenesis()) || to.Before(types.GetEffectiveGenesis()) {
 		logger.Panic("tried to push genesis layers")
 	}
 
 	// we never reapply the state of oldVerified. note that state reversions must be handled separately.
-	for layerID := oldVerified.Add(1); !layerID.After(newVerified); layerID = layerID.Add(1) {
-		l, err := msh.GetLayer(layerID)
+	for layerID := from.Add(1); !layerID.After(to); layerID = layerID.Add(1) {
+		layerBlocks, err := msh.LayerBlocks(layerID)
 		if err != nil {
-			logger.With().Error("failed to get layer", layerID, log.Err(err))
-			return err
+			return fmt.Errorf("failed to get layer %s: %w", layerID, err)
 		}
 
-		validBlocks, invalidBlocks := msh.BlocksByValidity(l.Blocks())
+		validBlocks, invalidBlocks := msh.BlocksByValidity(layerBlocks)
 		var (
 			applied    *types.Block
 			notApplied = invalidBlocks
@@ -427,10 +434,10 @@ func (msh *Mesh) pushLayersToState(ctx context.Context, oldVerified, newVerified
 
 		msh.Event().Info("end of layer state root",
 			layerID,
-			log.String("state_root", util.Bytes2Hex(msh.state.GetStateRoot().Bytes())),
+			log.Stringer("state_root", msh.state.GetStateRoot()),
 		)
 
-		if err = msh.reInsertTxsToPool(applied, notApplied, l.Index()); err != nil {
+		if err = msh.reInsertTxsToPool(applied, notApplied, layerID); err != nil {
 			logger.With().Error("failed to reinsert TXs to pool", layerID, log.Err(err))
 			return err
 		}
@@ -455,7 +462,7 @@ func (msh *Mesh) reInsertTxsToPool(applied *types.Block, notApplied []*types.Blo
 	}
 	returnedTxs, missing := msh.GetTransactions(uniqueTxIds(notApplied, seenTxIds))
 	if len(missing) > 0 {
-		msh.Panic("could not find transactions %v from layer %v", missing, l)
+		msh.With().Error("could not reinsert transactions", log.Int("missing", len(missing)), l)
 	}
 
 	grouped, accounts, err := msh.removeFromUnappliedTxs(returnedTxs)
@@ -489,7 +496,7 @@ func (msh *Mesh) applyState(block *types.Block) error {
 	}
 	txs, missing := msh.GetTransactions(block.TxIDs)
 	if len(missing) > 0 {
-		msh.Panic("could not find transactions %v from layer %v", missing, block.LayerIndex)
+		return fmt.Errorf("could not find transactions %v from layer %v", missing, block.LayerIndex)
 	}
 	// TODO: should miner IDs be sorted in a deterministic order prior to applying rewards?
 	failedTxs, svmErr = msh.state.ApplyLayer(block.LayerIndex, txs, rewardByMiner)
@@ -784,4 +791,11 @@ func getAggregatedLayerHashKey(layerID types.LayerID) []byte {
 	b.WriteString("ag")
 	b.Write(layerID.Bytes())
 	return b.Bytes()
+}
+
+func minLayer(i, j types.LayerID) types.LayerID {
+	if i.Before(j) {
+		return i
+	}
+	return j
 }
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -319,22 +319,19 @@ func (vl *validator) ProcessLayer(ctx context.Context, layerID types.LayerID) er
 		}
 	}
 	// mesh can't skip layer that failed to complete
-	from := minLayer(oldVerified, vl.latestLayerInState)
+	from := minLayer(oldVerified, vl.latestLayerInState).Add(1)
 	to := newVerified
-	logger.With().Debug("pushing layers to the state",
-		log.Stringer("from_layer", from),
-		log.Stringer("to_layer", to),
-	)
-	if to.After(from) {
+
+	if !to.Before(from) {
 		if err := vl.pushLayersToState(ctx, from, to); err != nil {
 			logger.With().Error("failed to push layers to state", log.Err(err))
 			return err
 		}
-		if err := vl.persistLayerHashes(ctx, from.Add(1), to); err != nil {
+		if err := vl.persistLayerHashes(ctx, from, to); err != nil {
 			logger.With().Error("failed to persist layer hashes", log.Err(err))
 			return err
 		}
-		for lid := from.Add(1); !lid.After(to); lid = lid.Add(1) {
+		for lid := from; !lid.After(to); lid = lid.Add(1) {
 			events.ReportLayerUpdate(events.LayerUpdate{
 				LayerID: lid,
 				Status:  events.LayerStatusTypeConfirmed,
@@ -421,11 +418,12 @@ func (msh *Mesh) pushLayersToState(ctx context.Context, from, to types.LayerID) 
 	logger.Info("pushing layers to state")
 	if from.Before(types.GetEffectiveGenesis()) || to.Before(types.GetEffectiveGenesis()) {
 		logger.Panic("tried to push genesis layers")
+		return nil
 	}
 
 	missing := msh.MissingLayer()
 	// we never reapply the state of oldVerified. note that state reversions must be handled separately.
-	for layerID := from.Add(1); !layerID.After(to); layerID = layerID.Add(1) {
+	for layerID := from; !layerID.After(to); layerID = layerID.Add(1) {
 		if err := msh.pushLayer(ctx, layerID); err != nil {
 			msh.setMissingLayer(layerID)
 			return err
@@ -824,6 +822,13 @@ func getAggregatedLayerHashKey(layerID types.LayerID) []byte {
 
 func minLayer(i, j types.LayerID) types.LayerID {
 	if i.Before(j) {
+		return i
+	}
+	return j
+}
+
+func maxLayer(i, j types.LayerID) types.LayerID {
+	if i.After(j) {
 		return i
 	}
 	return j
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -57,7 +57,9 @@ type Mesh struct {
 	// latestLayerInState is the latest layer whose contents have been applied to the state
 	latestLayerInState types.LayerID
 	// processedLayer is the latest layer whose votes have been processed
-	processedLayer      types.LayerID
+	processedLayer types.LayerID
+	// see doc for MissingLayer()
+	missingLayer        types.LayerID
 	nextProcessedLayers map[types.LayerID]struct{}
 	maxProcessedLayer   types.LayerID
 	mutex               sync.RWMutex
@@ -152,16 +154,33 @@ func (msh *Mesh) CacheWarmUp(layerSize int) {
 
 // LatestLayerInState returns the latest layer we applied to state.
 func (msh *Mesh) LatestLayerInState() types.LayerID {
-	defer msh.mutex.RUnlock()
 	msh.mutex.RLock()
+	defer msh.mutex.RUnlock()
 	return msh.latestLayerInState
 }
 
 // LatestLayer - returns the latest layer we saw from the network.
 func (msh *Mesh) LatestLayer() types.LayerID {
-	defer msh.mutex.RUnlock()
 	msh.mutex.RLock()
+	defer msh.mutex.RUnlock()
 	return msh.latestLayer
+}
+
+// MissingLayer is a layer in (latestLayerInState, processLayer].
+// this layer is missing critical data (valid blocks or transactions)
+// and can't be applied to the state.
+//
+// First valid layer starts with 1. 0 is empty layer and can be ignored.
+func (msh *Mesh) MissingLayer() types.LayerID {
+	msh.mutex.RLock()
+	defer msh.mutex.RUnlock()
+	return msh.missingLayer
+}
+
+func (msh *Mesh) setMissingLayer(lid types.LayerID) {
+	msh.mutex.Lock()
+	defer msh.mutex.Unlock()
+	msh.missingLayer = lid
 }
 
 // setLatestLayer sets the latest layer we saw from the network.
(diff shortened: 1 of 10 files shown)


dshulyak:
bors try

mesh/mesh.go
```diff
@@ -436,7 +463,7 @@ func (msh *Mesh) reInsertTxsToPool(validBlocks, invalidBlocks []*types.Block, l
 	uniqueTxIds(validBlocks, seenTxIds) // run for the side effect, updating seenTxIds
 	returnedTxs, missing := msh.GetTransactions(uniqueTxIds(invalidBlocks, seenTxIds))
 	if len(missing) > 0 {
-		msh.Panic("could not find transactions %v from layer %v", missing, l)
```
noamnelke, line 458: The reason this was originally a `Panic` is that there was a data availability phase when syntactically validating blocks, so if a block ended up not being applied (not contextually valid) and we try to re-insert the transactions to the pool it's an impossibility that we don't already have them locally. I chose to panic because this is unexpected behavior that indicates a big in the software.

At a minimum, this should be an error.

IMO, errors should indicate internal inconsistencies (bugs) or IO failures, while any condition that could be caused by behavior of other nodes should be limited to info/warning. In this case it's an internal inconsistency.

noamnelke (approved):
Elegant solution 👍🏻

mesh/mesh.go
```diff
@@ -394,31 +409,43 @@ func (msh *Mesh) getValidBlockIDs(ctx context.Context, layerID types.LayerID) ([
...
-		if err != nil || l == nil {
-			logger.With().Error("failed to get layer", layerID, log.Err(err))
-			return
+	for lid := from.Add(1); !lid.After(to); lid = lid.Add(1) {
+		if err := msh.pushLayer(ctx, logger, lid); err != nil {
+			logger.With().Warning("failed to push layer. reverting processed layer",
+				log.Err(err),
+				lid,
+			)
+			msh.overwriteProcessed(lid.Sub(1))
```
countvonzero: i find this hacky and confusing. 
in a steady-state, processedLayer (verified+1) is in lockstep with tortoise verified layer.
now for this rare occasion, processed is the older than tortoise verified layer.

i think we should change syncer instead of the logic in mesh. maybe syncer should be based off applied layer in SVM rather than processed layer?

countvonzero (changes requested):

mesh/mesh.go
dshulyak: i think that doesn't work. sync needs to make progress otherwise we may not have enough votes to verify the layer in tortoise and therefore applied layer in svm will be stuck

mesh/mesh.go
dshulyak: but those are from the invalid blocks, how do you know that the invalid block actually referenced existing transactions and not some random ids?

mesh/mesh.go
countvonzero: > we may not have enough votes to verify the layer in tortoise and therefore applied layer in svm will be stuck

applied layer in svm will be stuck anyway if any TX is not available. i don't see advancing tortoise verified layer help the situation. what's the point of tortoise making progress if applied layer is stuck?

in any case, i think you can make syncer starting to sync from appliedLayer+1, and still advance processed layer.
appliedLayer(svm) can be completely separate from processedLayer(mesh) and verifiedLayer(tortoise). you can advance the latter two with the appliedLayer stuck.

countvonzero: i also don't think we should rush this change. what can't we remove the panic stmt for now and tolerate inconsistent state in the network?

dshulyak: > i also don't think we should rush this change. what can't we remove the panic stmt for now and tolerate inconsistent state in the network?

i think this change is perfectly fine. also i don't understand what you are suggesting

dshulyak: i think what you suggested leads to redundant work. the verified layer may be stuck but all the data is available, but sync will loop through layers (because it will start from layer in state) and download data from disk.

it is hard for me to say how impactful that work will be, but given that we are using only 2 vcpus on cloud nodes, it can consume 1 of them fully.

dshulyak: > i also don't think we should rush this change. what can't we remove the panic stmt for now and tolerate inconsistent state in the network?

i thought about it and i guess what you meant is that it is ok to submit a layer without some transactions. i don't think that this is ok, basically it means that after several days every node in the network will have different state. at this point we may even disable transactions and run only consensus because state machine is meaningless if no one can agree on what the state looks like.

so to summarize:

>  i think you can make syncer starting to sync from appliedLayer+1, and still advance processed layer.
appliedLayer(svm) can be completely separate from processedLayer(mesh) and verifiedLayer(tortoise). you can advance the latter two with the appliedLayer stuck.

this doesn't work because of the redundant work. solution in this change does more work only when there is a lack of data

> what can't we remove the panic stmt for now and tolerate inconsistent state in the network?

this doesn't work because this is not a rare issue. after a day every node hit it at least once, and according to panics not one transaction was missing but many. and anyway this is a temporary hack, i don't know what would be the reason to implement it when the solution to this problem is very clear

> in a steady-state, processedLayer (verified+1) is in lockstep with tortoise verified layer.
now for this rare occasion, processed is the older than tortoise verified layer.

i don't really understand what it has to do with anything. it was in "almost" lockstep, now it is not because we need to trigger sync based on a feedback from tortoise... what it changes in practice?

dshulyak:
btw behavior in this pr is also consistent with a behavior that will be required for pruned blocks. since pruned block might be unavailable we shouldn't try to download every block that we know about. only after the tortoise decided that the block is valid we can be sure that the block is not pruned.

mesh/mesh.go
countvonzero: > this doesn't work because of the redundant work. solution in this change does more work only when there is a lack of data

it is not too complicated to add code to syncer that it directly try to fetch contents in valid blocks fro appliedLayer+1 to processedLayer if applied layer is stuck. and with fetcher its a noop if the block data is available locally. syncer just need to be exposed to more APIs from mesh and fetcher. if you want, i can work on that part in this PR.

my point about the applied layer being independent of processedLayer/verifiedLayer is that the node should be able to make consensus progress (advancing verifiedLayer) without ever applying the TXs in any block to SVM. i.e. all TXs stay in pending state. 

i like this PR in that it leverage syncer as a backup when data is missing. the part i dont like is to revert processedLayer only because some TXs are missing and therefore cannot be applied to SVM state. the logic doesnt follow because like i stated above, they should be independent. consensus outcome should not be impacted by whether we are able to apply to the state.

additionally, it makes the mesh code more complex. processed layer can be reverted randomly. i strongly prefer the processed layer only progress in one monolithic direction and in order. it will make future debugging a lot easier.

countvonzero:
> 



> btw behavior in this pr is also consistent with a behavior that will be required for pruned blocks. since pruned block might be unavailable we shouldn't try to download every block that we know about. only after the tortoise decided that the block is valid we can be sure that the block is not pruned.

i agree. but again, we can do that in this PR without having to revert processedLayer

mesh/mesh.go
dshulyak: > and with fetcher its a noop if the block data is available locally

it is not noop. it will read data from disk in a loop even when applied layer is legitimately stuck (not due to the missing data)

dshulyak: >  i strongly prefer the processed layer only to progress in one monolithic direction and in order. it will make future debugging a lot easier.

we can then have a queue of layers that failed to be applied to the state, new layers will be enqueued where i am overwriting processed layer. and dequeued when sync round starts, does it work?

it will do even less redundant work

countvonzero: we just need to add HasTX API in the db and read the keys. i think this overhead is worth it if processedLayer can be kept simple.

countvonzero: besides, i think we also need to discuss data availability in general. not just missing TXs.

dshulyak: > we just need to add HasTX API in the db and read the keys. i think this overhead is worth it if processedLayer can be kept simple.

i realized that we don't need it to be in db. we can have a map in mesh that will keep layers with missing data. layer will be added to the map only if mesh failed to process the layer. so it is kind of a queue.

syncer will read that map and for layer between applied + 1 ... processed it will fetch data if the key is in the map.

> besides, i think we also need to discuss data availability in general. not just missing TXs.

sure, we can discuss it. in my understanding, it is always the same. syncer should be the best effort until we know that the data is absolutely required (after tortoise verified layer). and it will work generically both if the ballot was pruned or data is being maliciously withhold

countvonzero: > i realized that we don't need it to be in db. we can have a map in mesh that will keep layers with missing data. layer will be added to the map only if mesh failed to process the layer. so it is kind of a queue.
syncer will read that map and for layer between applied + 1 ... processed it will fetch data if the key is in the map.

i am happy w this solution. this means the logic around processedLayer stay the same.

> sure, we can discuss it. in my understanding, it is always the same. syncer should be the best effort until we know that the data is absolutely required (after tortoise verified layer). and it will work generically both if the ballot was pruned or data is being maliciously withhold

yes. id like to discuss the general strategy in the dev meeting on Wednesday first and see if this PR fit in the picture

mesh/mesh.go
noamnelke: I might be wrong about this, but I thought that blocks go through syntactic validation when they're received. Syntactic validation should include data availability.

This specific code is not where we handle an incoming block (be it via gossip or sync). It's not where we handle a syntactically invalid block (that might include garbage) - rather it's where we handle a syntactically valid block that's been rejected by the consensus mechanism. At this point we should have already obtained the dependencies. If we couldn't obtain them (because, as you said someone may have created a block with garbage, non-existent transactions) then it wouldn't be syntactically valid and any object referencing it (e.g. a ballot) shouldn't be considered syntactically valid, either.

If I got it wrong then we should discuss this to make sure that the behavior makes sense. We eventually want to support pruning transactions, but this is not as simple as just not having them when they're referenced in a block (even a contextually invalid one - those are precisely the transactions we would want to prune).

dshulyak: in general yes. but right now there is no strict check for transactions availability (i don't know for how long this was the case). fetcher just tries to download transactions, if it can't it moves on. and i don't think that we should change that.

in my understanding that in the long term when we receive a block we won't be able to determine if it was pruned or simply malicious being withheld by a peer. so we can't simply ignore it. we also can't try to download it indefinitely.  instead node should count votes for that block and if according to the tortoise such block is valid - we can assume that the block is available and can be downloaded.

this is why i think that we should relax data availability for blocks during gossip and enforce it only when we know which blocks are valid.

noamnelke: This code deals with proposals, not blocks (despite the outdated naming) and a proposal is only ever stored by gossiping:

The only production code path that leads to storing a proposal goes through `HandleProposalData` and there's this check there that prevents storing the proposal if a transaction is unavailable:
https://github.com/spacemeshos/go-spacemesh/blob/c75fc000224636e881f48999245d7dc3a5447615/proposals/handler.go#L207-L210

dshulyak: you should look into fetcher.GetTXs

nkryuchkov (approved):

dshulyak:
@countvonzero please take a look, i reworked change according to the last discussion. 

@noamnelke if we want to fix fetcher.GetTXs let's do it in a separate change. i am not confident if that's would be a future proof solution after a reply from Iddo and Tal

mesh/mesh.go
```diff
@@ -287,33 +305,43 @@ func (vl *validator) ProcessLayer(ctx context.Context, layerID types.LayerID) er
...
-		if err := vl.pushLayersToState(ctx, oldVerified, newVerified); err != nil {
+	// mesh can't skip layer that failed to complete
+	from := minLayer(oldVerified, vl.latestLayerInState)
+	to := newVerified
+	logger.With().Debug("pushing layers to the state",
+		log.Stringer("from_layer", from),
+		log.Stringer("to_layer", to),
+	)
+	if to.After(from) {
+		if err := vl.pushLayersToState(ctx, from, to); err != nil {
```
countvonzero, line 326: maybe pass from.Add(1) here too.

mesh/mesh_test.go
```diff
@@ -712,3 +711,108 @@ func TestMesh_AddBlockWithTXs(t *testing.T) {
...
+	defer tm.ctrl.Finish()
+	defer tm.Close()
+
+	ctx := context.TODO()
+	genesis := types.GetEffectiveGenesis()
+	last := genesis.Add(10)
+
+	for lid := genesis.Add(1); !lid.After(last); lid = lid.Add(1) {
+		tm.mockTortoise.EXPECT().HandleIncomingLayer(gomock.Any(), gomock.Any()).
+			Return(lid.Sub(1), lid, false)
```
countvonzero: actually this should return `lid.sub(2), lid.sub(1)`

syncer/syncer_test.go
```diff
@@ -1112,3 +1117,59 @@ func TestGetATXsOldAndCurrentEpoch(t *testing.T) {
...
+	goldenATXID := types.ATXID(types.HexToHash32("77777"))
+	atxdb := activation.NewDB(atxStore, nil,
+		activation.NewIdentityStore(database.NewMemDatabase()),
+		layersPerEpoch, goldenATXID, nil, lg.WithName("atxDB"))
+	svmstate := mmocks.NewMockstate(gomock.NewController(t))
+	svmstate.EXPECT().GetStateRoot().AnyTimes()
+
+	tortoise := mmocks.NewMocktortoise(ctrl)
+
+	m := mesh.NewMesh(memdb, atxdb, tortoise, nil, svmstate, lg.WithName("mesh"))
```
countvonzero, line 1145: maybe a mock mesh is better in syncer. the test setup was mostly copied from the old syncer code. i didn't do a good job updating it.

countvonzero (approved):
thanks for doing this. i really like this approach.

countvonzero:
> @noamnelke if we want to fix fetcher.GetTXs let's do it in a separate change. i am not confident if that's would be a future proof solution after a reply from Iddo and Tal

i agree with doing this in a separate PR. but i also don't think it's necessary to make it future proof. 

research seems to think there is no difference in terms of security.
the trade-off i saw was to download extra data when the network tells us that some previously invalid data is now valid.
the code will be simpler, and easier to reason about if the data availability is guaranteed at data reception time.
if, in the future, some major inefficiency or ddos attack vector arise, it's also pretty straightforward to relax it.

i.e. the cost of making the wrong decision is low, but the immediate benefit is that it's easier to reason about the code logic

dshulyak:
thats not how i interpreted the reply 

> ballots should not be invalidated if they vote for an invalid block, only if they use an invalid base ballot (since we want to allow block pruning). 

so even if we cannot fetch referenced block (vote) - ballot should be saved to the database

countvonzero:
> thats not how i interpreted the reply
> 
> > ballots should not be invalidated if they vote for an invalid block, only if they use an invalid base ballot (since we want to allow block pruning).
> 
> so even if we cannot fetch referenced block (vote) - ballot should be saved to the database

oh! i thought Tal was talking abt contextual validity. when we receive a ballot, we can determine the syntactic validity of the blocks it references to, not contextual validity. so we should only reject a ballot if it references to syntactic ally invalid blocks

dshulyak:
i thought that `(since we want to allow block pruning)` implies that blocks won't be available, so it won't possible to determine syntactic validity of the block, e.g block will be pruned

syncer/syncer_test.go
dshulyak: you meant that it will be better to replace mesh with interface and use mock for it? i guess it will be cleaner, on the other hand the interaction between syncer and mesh got tested.

dshulyak:
bors merge

PR #3043, opened 2022-01-04
[Merged by Bors] - tortoise: encode abstain votes as an array with layers

### LINKED ISSUE #2369: blocks, tortoise: merge for/against exceptions into a single list, vote neutral on layers instead of blocks

dshulyak:
## Motivation

In verifying tortoise we should not count weight from the ballot if the ballot abstained for that layer. This will be easier to implement if abstain votes were encoded as an array of layer. 

Also, this change is a must for UBC, as we don't know the actual block value before hare termination. So there is no way to vote abstain.

Support/Against are not collapsed into a single list as it leads to less efficient implementation. Maybe we should revise it in the future.

Closes: https://github.com/spacemeshos/go-spacemesh/issues/2369

## Changes
- encode abstain votes as an array of layers

## Test Plan
existing unit tests, logically nothing has changed


countvonzero (approved):

tortoise/tortoise.go
```diff
@@ -356,6 +356,10 @@ func (t *turtle) encodeVotes(
 				votes.Support = append(votes.Support, bid)
 			case against:
 				votes.Against = append(votes.Against, bid)
+			case abstain:
```
countvonzero, line 360: maybe instead comment that tortoise cannot vote abstain past hdist.

common/types/votes.go
```diff
@@ -0,0 +1 @@
+package types
```
nkryuchkov: Why is this file needed?

tortoise/tortoise.go
```diff
@@ -744,6 +743,19 @@ func (t *turtle) addLocalVotes(ctx context.Context, logger log.Log, lid types.La
...
+	genesis := types.GetEffectiveGenesis()
+	limit := genesis
+	if state.last.After(genesis.Add(config.Zdist)) {
+		limit = state.last.Sub(config.Zdist)
+	}
+	_, isUndecided := state.undecided[lid]
+	if !lid.Before(limit) && isUndecided {
+		return true
+	}
+	return false
```
nkryuchkov:
```suggestion
	return !lid.Before(limit) && isUndecided
```

nkryuchkov (approved):

common/types/votes.go
dshulyak: my bad, forgot to remove it

tortoise/tortoise.go
dshulyak: will change to an error, just in case

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3044, opened 2022-01-04
[Merged by Bors] - UCB: hare generate unified content block

### LINKED ISSUE #2903: Create `ProposalStore` under Hare
Today, when a block is processed, it's stored in the database. Every time the clock ticks, the Hare waits `WakeupDelta` seconds and then calls `mesh.LayerBlocks(lyrID)` to get the list of on-time blocks from the database.

This mechanism should be replaced by a `ProposalStore` managed by the Hare.

The `ProposalStore` should support:
- Storing a `Proposal`
- Getting a list of `Proposal`s for a given layer
- Fetching a specific `Proposal` (given `LayerID` and `ProposalID`)
- Pruning all `Proposal`s from before a specific layer

The `ProposalStore` should be a LevelDB. The key should be the concatenation of `LayerID+ProposalID`. Layers more than `hdist` back should be pruned regularly.

Once the `ProposalStore` is implemented we should replace any usage of the `BallotDB` to the `ProposalStore` if the proposal (list of transactions) is actually needed. The Hare should agree on a set of `ProposalID`s and executing layers will be done with proposals, but the Tortoise should use ballots.

## Why not keep it in memory?
I originally thought of just keeping it in memory as `map[LayerID]map[ProposalID]Proposal`. There should be `blocks_per_layer * hdist` proposals in the store. Not sure about the final values, but I think reasonable numbers for scale are `50 * 120 = 6000`. The size of a `Proposal` will mainly depend on the number of transactions that we allow referencing in one (assuming we strip the embedded `Ballot`), each `TransactionID` is currently 32-bytes, but we may decide to reduce it to 20-bytes. If we allow 100 transactions per proposal, this comes out to <20mb which is not negligible to keep in memory, but still viable.

I'm assuming 20% "wasted" transactions (duplicates, eventually invalid, etc.) which means `100 transactions per proposal * 50 blocks per layer * 80% = 4000` then, dividing by 300 seconds per layer brings us to ~13 transactions per second. I think we're aiming for higher TPS, so we should probably store the proposals on disk.

### LINKED ISSUE #2904: Ability to Fetch `Proposal`s from Peers
Eventually we'll want all existing mechanisms for blocks to work for `Ballot`s. However, in addition to the ability to sync ballots (which is already implemented for blocks today) we also need to the ability to fetch a specific `Proposal` from peers if we see it referenced in a Hare message and we don't have it.

There should never be a need to get a list of proposals. We only need the ability to fetch a specific proposal from peers, given `LayerID` and `ProposalID`. Proposals will be fetched from the `ProposalStore`.

The fetching mechanism will be used when implementing `Block` derivation.

### LINKED ISSUE #2905: Store Only Ballots in `BallotDB` and Use `BallotID` for Referring to Ballots
At this point we can start saving only the `Ballot` that's embedded in the `Proposal` in the `BallotDB`, index the database using `BallotID`s and change references to ballots so they use `BallotID`s.

### LINKED ISSUE #2906: Derive `Block` from Hare Output Set

### LINKED ISSUE #2907: Update Voting to Blocks
When publishing a `Ballot` we should compare the index of blocks we've derived ourselves with the base block. We should have derived one block for every layer in `hdist` and the lists should match. If they don't we should add appropriate diffs. If we don't have a derived block for a layer within `hdist` - we should vote neutral for it.

Now that votes are for blocks, the state transition function (`mesh.applyState`) should no longer derive the list of transactions itself, and instead it should pick the block with sufficient votes and execute the rewards and transactions from it in order.

countvonzero:
## Motivation

Closes #2903
Closes #2904
Closes #2905
Closes #2906
Closes #2907

## Changes

- hare: 
  generate unified content block from proposals in the consensus output
- mesh
  when apply a layer, it picks the lowest BlockID among all the valid blocks verified by tortoise
- tortoise
  within hdist, vote Support for hare output block only, and vote Against for any other blocks in the layer
  outside hdist, no change.

## Test Plan

unit tests, system tests

## DevOps Notes


countvonzero:
bors try

tortoise/tortoise.go
```diff
@@ -761,8 +761,12 @@ func (t *turtle) addLocalVotes(ctx context.Context, logger log.Log, lid types.La
 			t.undecided[lid] = struct{}{}
 			return nil
 		}
-		for _, bid := range hareOutput {
-			t.hareOutput[bid] = support
+		bids := t.blocks[lid]
+		for _, bid := range bids {
```
dshulyak: please remove this loop. if hare doesn't support a block it should not be added to the t.hareOutput

tortoise/tortoise.go
```diff
@@ -746,8 +746,8 @@ func (t *turtle) addLocalVotes(ctx context.Context, logger log.Log, lid types.La
 	)
 
 	if !lid.Before(t.layerCutoff()) {
-		// for newer layers, we vote according to the local opinion (hare output, from hare or sync)
-		hareOutput, err := t.bdp.GetHareConsensusOutput(lid)
+		// for newer layers, we vote according to the local opinion (hare output, from live participation or sync)
+		theOne, err := t.bdp.GetHareConsensusOutput(lid)
```
dshulyak: hareOutput is better variable

tortoise/tortoise.go
```diff
@@ -761,8 +761,12 @@ func (t *turtle) addLocalVotes(ctx context.Context, logger log.Log, lid types.La
 			t.undecided[lid] = struct{}{}
 			return nil
 		}
-		for _, bid := range hareOutput {
-			t.hareOutput[bid] = support
+		bids := t.blocks[lid]
+		for _, bid := range bids {
+			t.hareOutput[bid] = against
+		}
+		if theOne != types.EmptyBlockID {
```
dshulyak: when block will be empty and error is nil?

tortoise/sim/layer.go
```diff
@@ -142,39 +132,44 @@ func (g *Generator) genLayer(cfg nextConf) types.LayerID {
...
+		if err = ballot.Initialize(); err != nil {
+			g.logger.With().Panic("failed to init ballot", log.Err(err))
+		}
+		for _, state := range g.states {
+			state.OnBallot(ballot)
+		}
+		layer.AddBallot(ballot)
+	}
+	for i := 0; i < numBlocks; i++ {
+		block := types.GenLayerBlock(g.nextLayer, types.RandomTXSet(100))
```
dshulyak: please use seeded randomness, this helpers cannot be used

tortoise/rerun_test.go
```diff
@@ -101,10 +102,13 @@ func testWindowCounting(tb testing.TB, maliciousLayers, verifyingWindow, fullWin
 	}
 	require.NoError(tb, tortoise.rerun(ctx))
 
-	for _, block := range blocks {
-		validity, err := s.GetState(0).MeshDB.ContextualValidity(block)
-		require.NoError(tb, err)
-		require.Equal(tb, expectedValidity, validity, "validity for block %s", block)
+	validBlocks, err := s.GetState(0).MeshDB.LayerContextuallyValidBlocks(context.TODO(), misverified)
```
dshulyak: what was the reason for this change? i think it needs to be reverted

this is actually completely incorrect now. all blocks must be verified

dshulyak (changes requested):

tortoise/tortoise_test.go
```diff
@@ -1377,17 +1408,18 @@ func skipLayers(n int) sim.VotesGenerator {
 			panic(fmt.Sprintf("need at least %d layers", position))
 		}
 		baseLayer := layers[len(layers)-position]
-		support := layers[len(layers)-position].BlocksIDs()
+		support := layers[len(layers)-position].BlocksIDs()[0:1]
```
dshulyak: please revert this change. it doesn't have anything to do with hare

tortoise/tortoise_test.go
```diff
@@ -1698,11 +1730,13 @@ func TestVoteAgainstSupportedByBaseBallot(t *testing.T) {
 
 	unsupported := map[types.BlockID]struct{}{}
 	for lid := genesis; lid.Before(last); lid = lid.Add(1) {
-		bids, err := s.GetState(0).MeshDB.LayerBlockIds(lid)
+		theOne, err := s.GetState(0).MeshDB.GetHareConsensusOutput(lid)
```
dshulyak: please rename to hareOutput

tortoise/tortoise_test.go
```diff
@@ -1698,11 +1730,13 @@ func TestVoteAgainstSupportedByBaseBallot(t *testing.T) {
...
 		require.NoError(t, err)
-		tortoise.trtl.validity[bids[0]] = against
-		tortoise.trtl.hareOutput[bids[0]] = against
-		unsupported[bids[0]] = struct{}{}
+		if theOne != types.EmptyBlockID {
```
dshulyak: this conditiona should be removed

tortoise/tortoise_test.go
```diff
@@ -1804,7 +1838,18 @@ func TestComputeLocalOpinion(t *testing.T) {
 			require.NoError(t, err)
 			for _, bid := range blocks {
 				vote, _ := getLocalVote(&tortoise.trtl.commonState, cfg, tc.lid, bid)
-				require.Equal(t, tc.expected, vote, "block id %s", bid)
+				if tc.expected == support {
+					theOne, err := s.GetState(0).MeshDB.GetHareConsensusOutput(tc.lid)
```
dshulyak: pleased rename to hare output

tortoise/rerun_test.go
dshulyak: this are the changes to fix it https://github.com/dshulyak/go-spacemesh/commit/8d7067db9f640eb270ebb987c104e71a039dfb90

tortoise/sim/layer.go
countvonzero: i am wary of disturbing the random sequence in sim package.
when i tried using `g.rng` some tests randomly failed. for example, if i use `g.genTXIDs(5)` TestBaseBallotEvictedBlock passes but `g.genTXIDs(100)` it fails. this is just one example.

tortoise/tortoise.go
countvonzero: when hare output empty set

tortoise/rerun_test.go
countvonzero: thanks for the patch. applied

tortoise/sim/layer.go
dshulyak: i guess it has something to do with random weights, i thought they were not used by default anymore. nvm then i will take a look after

tortoise/tortoise_test.go
countvonzero: this is due to the explicit `sim.WithEmptyHareOutput()`. so one of the valid block will be types.EmptyBlockID which won't be in the support list.

collector/collector.go
```diff
@@ -113,6 +120,17 @@ func (c *EventsCollector) collectEvents(url string) {
 loop:
 	for {
 		select {
+		case data := <-proposals:
```
dshulyak, line 123: what this collector does? maybe this is some old artifact and we can ditch it instead of maintaining?

common/types/block.go
```diff
@@ -1,41 +1,100 @@
...
+type Block struct {
+	InnerBlock
+	// the following fields are kept private and from being serialized
+	blockID BlockID
+}
+
+// InnerBlock contains the transactions and rewards of a block.
+type InnerBlock struct {
+	LayerIndex LayerID
+	Rewards    []AnyReward
```
dshulyak, line 34: type Reward is reserved for something else?

mesh/mesh.go
```diff
@@ -406,19 +399,43 @@ func (msh *Mesh) pushLayersToState(ctx context.Context, oldVerified, newVerified
...
+		if len(blocks) > 0 {
+			// when tortoise verify multiple blocks in the same layer, we only apply one with the lowest
+			// lexicographical sort order
+			applied = blocks[0]
+			if len(blocks) > 1 {
+				notApplied = append(notApplied, blocks[1:]...)
+			}
+		}
+
+		if err = msh.updateStateWithLayer(ctx, layerID, applied); err != nil {
```
dshulyak, line 423: is it expected that applied will be nil is layer is empty?

mesh/mesh.go
```diff
@@ -431,133 +448,128 @@ func (msh *Mesh) revertState(ctx context.Context, layerID types.LayerID) error {
...
-	lyr := types.NewExistingLayer(validatedLayer, blocks)
-	msh.ProcessLayer(ctx, lyr)
+	return msh.ProcessLayer(ctx, layerID)
 }
 
-// apply the state for a single layer. stores layer results for early layers, and applies previously stored results for
-// now-older layers.
-func (msh *Mesh) updateStateWithLayer(ctx context.Context, layer *types.Layer) {
+// apply the state for a single layer.
+func (msh *Mesh) updateStateWithLayer(ctx context.Context, layerID types.LayerID, hareOutput *types.Block) error {
```
dshulyak: i think it doesn't make sense to call it hare output after tortoise verified a block

dshulyak (approved):

mesh/mesh.go
countvonzero: yes. if the hare output an empty set, then it's likely that the tortoise will not verify any block. i.e. no blocks in that layer has contextual validity bit set to true.

common/types/block.go
countvonzero: yes. it's used in other place. i meant for whoever that works on TX selection/rewards to re-arrange the naming.

collector/collector.go
countvonzero: the only thing is the memory collector used in cmd/multi_node_sim.go. i am leaving it for now. not sure what it's for tho.

mesh/mesh.go
countvonzero: changed to `block`

countvonzero:
bors merge

PR #3045, opened 2022-01-05
[Merged by Bors] - tortoise: support abstain votes in verifying tortoise

### LINKED ISSUE #3032: tortoise: abstain votes are consistent with any local opinion

dshulyak:
## Motivation

ballots may vote abstain for a layer if the hare instance didn't terminate
for that layer. verifying tortoise counts weight from those ballots
but it doesn't count weight for abstained layers.
	
for voting tortoise still prioritizes good ballots, ballots that selects base ballot
that abstained will be marked as a bad ballot. this condition is necessary
to simplify implementation. otherwise verifying tortoise needs to copy abstained
votes from the base ballot if they weren't fixed.
	
Example:
Block: 0x01 Layer: 9
Ballot: 0xaa Layer: 11 Votes: {Base: 0x01, Support: [0x01], Abstain: [10]}
Assuming that Support is consistent with local opinion, verifying tortoise
will count the weight from this ballot for layer 9, but not for layer 10.

closes: https://github.com/spacemeshos/go-spacemesh/issues/3032

## Changes
- introduce abstained state in verifying tortoise
- reduce margin by abstained weight when layer is being verified

## Test Plan
additional unit tests

### DIFF
--- a/tortoise/tortoise_test.go
+++ b/tortoise/tortoise_test.go
@@ -2106,6 +2106,49 @@ func TestSwitchVerifyingByChangingGoodness(t *testing.T) {
 	require.True(t, tortoise.trtl.mode.isVerifying(), "verifying mode")
 }
 
+func perfectVotingFirstBaseBallot(rng *mrand.Rand, layers []*types.Layer, _ int) sim.Voting {
+	baseLayer := layers[len(layers)-1]
+	support := layers[len(layers)-1].BlocksIDs()[0:1]
+	against := layers[len(layers)-1].BlocksIDs()[1:]
+	ballots := baseLayer.Ballots()
+	base := ballots[0]
+	return sim.Voting{Base: base.ID(), Against: against, Support: support}
+}
+
+func abstainVoting(rng *mrand.Rand, layers []*types.Layer, _ int) sim.Voting {
+	baseLayer := layers[len(layers)-1]
+	ballots := baseLayer.Ballots()
+	return sim.Voting{Base: ballots[0].ID(), Abstain: []types.LayerID{baseLayer.Index()}}
+}
+
+func TestAbstainVotingVerifyingMode(t *testing.T) {
+	ctx := context.Background()
+	const size = 10
+	s := sim.New(sim.WithLayerSize(size))
+	s.Setup(sim.WithSetupUnitsRange(2, 2))
+
+	cfg := defaultTestConfig()
+	cfg.LayerSize = size
+
+	tortoise := tortoiseFromSimState(s.GetState(0), WithLogger(logtest.New(t)), WithConfig(cfg))
+	var last, verified types.LayerID
+	for _, last = range sim.GenLayers(s,
+		sim.WithSequence(1),
+		sim.WithSequence(10, sim.WithVoteGenerator(
+			sim.VaryingVoting(1, perfectVotingFirstBaseBallot, abstainVoting),
+		)),
+	) {
+		_, verified, _ = tortoise.HandleIncomingLayer(ctx, last)
+	}
+	require.Equal(t, last.Sub(2), verified)
+	for _, last = range sim.GenLayers(s,
+		sim.WithSequence(1, sim.WithVoteGenerator(perfectVotingFirstBaseBallot)),
+	) {
+		_, verified, _ = tortoise.HandleIncomingLayer(ctx, last)
+	}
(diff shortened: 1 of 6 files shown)


tortoise/verifying.go
```diff
@@ -84,38 +106,52 @@ func (v *verifying) markGoodCut(logger log.Log, lid types.LayerID, ballots []tor
...
+	if w, exist := v.abstainedWeight[lid]; exist {
+		margin.sub(w)
+	}
 
 	logger = logger.WithFields(
 		log.String("verifier", verifyingTortoise),
 		log.Stringer("candidate_layer", lid),
-		log.Stringer("voting_weight", votingWeight),
+		log.Stringer("margin", margin),
+		log.Stringer("asbtained_weight", v.abstainedWeight[lid]),
```
nkryuchkov:
```suggestion
		log.Stringer("abstained_weight", v.abstainedWeight[lid]),
```

nkryuchkov (approved):

tortoise/tortoise_test.go
```diff
@@ -2106,6 +2106,49 @@ func TestSwitchVerifyingByChangingGoodness(t *testing.T) {
 	require.True(t, tortoise.trtl.mode.isVerifying(), "verifying mode")
 }
 
+func perfectVotingFirstBaseBallot(rng *mrand.Rand, layers []*types.Layer, _ int) sim.Voting {
+	baseLayer := layers[len(layers)-1]
+	support := layers[len(layers)-1].BlocksIDs()[0:1]
+	against := layers[len(layers)-1].BlocksIDs()[1:]
```
countvonzero, line 2112: i recall you said this is not necessary?

tortoise/verifying.go
```diff
@@ -24,6 +26,21 @@ func (g goodness) String() string {
...
 const (
 	bad goodness = iota
+	// ballots may vote abstain for a layer if the hare instance didn't terminate
+	// for that layer. verifying tortoise counts weight from those ballots
+	// but it doesn't count weight for abstained layers.
+	//
+	// for voting tortoise still prioritizes good ballots, ballots that selects base ballot
+	// that abstained will not be marked as a bad ballot. this condition is necessary
+	// to simplify implementation. otherwise verifying tortoise needs to copy abstained
+	// votes from base ballot if they weren't fixed.
```
countvonzero: i am not sure i fully understand this paragraph.

> this condition is necessary to simplify implementation. otherwise verifying tortoise needs to copy abstained votes from base ballot if they weren't fixed.

i don't understand why ` ballots that selects base ballot that abstained will not be marked as a bad ballot.` helps simplify the implementation.
and i don't see why `copy abstained votes from base ballot` is very complicated.

tortoise/verifying.go
```diff
@@ -195,7 +234,19 @@ func (v *verifying) determineGoodness(logger log.Log, ballot tortoiseBallot) goo
...
-	if rst := v.goodBallots[ballot.base]; rst != good {
+	base := v.goodBallots[ballot.base]
+
+	if len(ballot.abstain) > 0 {
+		logger.With().Debug("ballot has abstained on some layers", log.Stringer("base_goodness", base))
+		switch base {
+		case good:
+			return abstained
+		default:
+			return bad
```
countvonzero, line 245: earlier your comments says:

> ballots that selects base ballot that abstained will not be marked as a bad ballot.

but it seems in the impl if base == abstained, the ballot will be marked as bad.

tortoise/tortoise_test.go
dshulyak: i don't see how is that related. it is unnecessary in different test case and vote generator

tortoise/verifying.go
dshulyak: yes sorry it should have been - will be marked as a bad ballot.

> and i don't see why copy abstained votes from base ballot is very complicated.

it is recursive and adds unnecessary complexity to verifying tortoise. also you need to track if new ballot has terminated for the layer that is previously was in abstain list. i am saying that all of that is unnecessary

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3046, opened 2022-01-05
[Merged by Bors] - Move generating TXs in tests into `svm`; add & use spawn TX in tests

### LINKED ISSUE #2829: Move code for generating transactions and addresses in tests into SVM package
Currently, addresses are the the 20-byte suffix of a public key. When we transition to SVM with account abstraction, they will become more complex - a hash of the template address, constructor arguments and some nonce. For the sake of tests at the time we transition to SVM, all accounts will use the same template and nonce, and they'll only differ in their public key (which the only argument to the constructor that changes). Creating an address for tests can be implemented as a function in the SVM package that only receives the public key as an argument.

Since we'll only use a single template for now and we should have a way to predeploy templates - tests will never need to explicitly call `deploy`. However, the first time we use an account, it will have to be spawned.

Create a function for generating a simple spend transaction. Other than the sender's private key, amount and recipient, this function should receive a `bool` called `isFirst` - when true, it returns a "spawn + spend" transaction.

## Tasks
- [ ] `func GenerateAddress(publicKey []byte) Address`
- [ ] `func GenerateSpawnTransaction(signer *signing.EdSigner, target Address) *types.Transaction`
- [ ] `func GenerateCallTransaction(signer *signing.EdSigner, target Address, nonce, amount, gas, fee uint64) *types.Transaction`
- [ ] Use the new functions instead of `types.NewSignedTx()` and similar functions (e.g. the 7 implementations of `NewTx()` that we have in different tests).

nkryuchkov:
## Motivation

Closes #2829

## Changes

- Move code for generating transactions into `svm` package
- Add spawn transactions and use them in tests

## TODO

## DevOps Notes


common/types/address.go
```diff
@@ -12,10 +17,11 @@ import (
 
 const (
 	// AddressLength is the expected length of the address.
+	// TODO(nkryuchkov): change to 32.
 	AddressLength = 20
 )
 
-// Address represents the 20 byte address of an spacemesh account.
+// Address represents the address of an spacemesh account with AddressLength length.
```
noamnelke:
```suggestion
// Address represents the address of a spacemesh account with AddressLength length.
```

activation/atxdb_test.go
```diff
@@ -739,12 +739,12 @@ func TestActivationDb_ContextuallyValidateAtx(t *testing.T) {
 	idStore := NewIdentityStore(database.NewMemDatabase())
 	atxdb := NewDB(database.NewMemDatabase(), nil, idStore, layersPerEpochBig, goldenATXID, &ValidatorMock{}, lg.WithName("atxDB"))
 
-	validAtx := types.NewActivationTx(newChallenge(nodeID, 0, *types.EmptyATXID, goldenATXID, types.LayerID{}), [20]byte{}, nil, 0, nil)
+	validAtx := types.NewActivationTx(newChallenge(nodeID, 0, *types.EmptyATXID, goldenATXID, types.LayerID{}), [types.AddressLength]byte{}, nil, 0, nil)
```
countvonzero: just `types.Address{}`, here and below

common/types/address.go
```diff
@@ -12,10 +17,11 @@ import (
 
 const (
 	// AddressLength is the expected length of the address.
+	// TODO(nkryuchkov): change to 32.
```
countvonzero: why? can you put the issue number here? thanks

common/types/address.go
```diff
@@ -112,3 +118,43 @@ func (a *Address) SetBytes(b []byte) {
 	}
 	copy(a[AddressLength-len(b):], b)
 }
+
+type nonce128 struct {
+	mu          sync.Mutex
+	little, big uint64
```
countvonzero: is it possible to encode this with xdr in the codec package?

common/types/address.go
```diff
@@ -112,3 +118,43 @@ func (a *Address) SetBytes(b []byte) {
...
+	// TODO(nkryuchkov): For now nonce is always 0. When we need to fix that, the code below needs to be uncommented.
+	//n.little++
+	//if n.little == 0 {
+	//	n.big++
+	//}
+
+	return a
+}
+
+var nonce nonce128
```
countvonzero: this implies all addresses will share the same nonce?

common/types/address.go
```diff
@@ -112,3 +118,43 @@ func (a *Address) SetBytes(b []byte) {
 	}
 	copy(a[AddressLength-len(b):], b)
 }
+
+type nonce128 struct {
```
countvonzero: i think we need some comment on the nonce here. (what is little? big?)

common/types/address.go
```diff
@@ -112,3 +118,43 @@ func (a *Address) SetBytes(b []byte) {
...
+	mu          sync.Mutex
+	little, big uint64
+}
+
+func (n *nonce128) next() []byte {
+	n.mu.Lock()
+	defer n.mu.Unlock()
```
countvonzero: why not instantiate one nonce per address? i don't understand the need for mutex

common/types/transaction.go
```diff
@@ -192,8 +189,13 @@ type Reward struct {
 	Coinbase            Address
 }
 
-// NewSignedTx is used in TESTS ONLY to generate signed txs.
-func NewSignedTx(nonce uint64, rec Address, amount, gas, fee uint64, signer *signing.EdSigner) (*Transaction, error) {
+// GenerateSpawnTransaction generates a spawn transaction.
+func GenerateSpawnTransaction(signer *signing.EdSigner, target Address) *Transaction {
```
countvonzero: unused?

common/types/transaction.go
```diff
@@ -192,8 +189,13 @@ type Reward struct {
...
+	return &Transaction{}
+}
+
+// GenerateCallTransaction generates a call transaction.
+func GenerateCallTransaction(signer *signing.EdSigner, rec Address, nonce, amount, gas, fee uint64) (*Transaction, error) {
```
countvonzero: if this is intended for testing only, can you move it to common/type/testutil.go?

common/types/address.go
```diff
@@ -112,3 +118,43 @@ func (a *Address) SetBytes(b []byte) {
...
+
+	return a
+}
+
+var nonce nonce128
+
+// GenerateAddress generates an address from a public key.
+func GenerateAddress(publicKey []byte) Address {
+	input := bytes.Repeat([]byte{0x00}, 20)
+	input = append(input, nonce.next()...)
```
noamnelke: Not sure I understand this... Address generation should be deterministic, so why are we hashing a nonce into it?

Also not sure why we're hashing 20 zero-bytes into the address...

common/types/address.go
nkryuchkov: I assumed it should be changed to 32, but I just thought about it and now I'm not sure it should be. @noamnelke @neysofu will it be 20 or 32?

common/types/address.go
nkryuchkov: Could you explain your thought, please? I didn't understand

common/types/address.go
nkryuchkov: No, for now, all nonces are just zeros of size 128 bit. Later, we will need to have a nonce implementation. Perhaps I need to simplify the code to make it more clear

common/types/address.go
noamnelke: it should be 20 :)

common/types/address.go
noamnelke: I don't think this whole struct is not needed. See my comment on where it's used.

common/types/address.go
noamnelke: See my comment below, I think this should be removed.

common/types/transaction.go
noamnelke: See my review comment - we should start using it in this PR before merging.

common/types/transaction.go
noamnelke: While this will only be used in tests - it should actually live inside the SVM package (see my review comment).

noamnelke (changes requested):
What's missing:
- Some cleanup, e.g. `StringToAddress` from `common/types/address.go`.
- Replacing the various implementations of `NewTx` in tests with the new function.
- Actually use `GenerateSpawnTransaction` in tests. The idea is that when we switch to SVM a transaction from a new account that hasn't been spawned yet will fail. We want to update the tests now so that we won't need to touch them as part of the transition to SVM. In simple tests we can just execute a spawn transaction before the first transaction from an account. In tests where transactions are generated more dynamically (if we have something like that) we can check if the account nonce is zero and then generate a spawn transaction.
- Last, but not least: the code for generating an address and a transaction should live inside the SVM package. This is critical so that when we replace the package those get replaced as well. The `types` package should not need to import those methods, if I'm not mistaken - let me know if I missed something and we'll figure it out.

common/types/address.go
nkryuchkov: My bad. Removed

common/types/address.go
nkryuchkov: As I understood, the generated address should be `SHA256(20 null-bytes || 128-bit-nonce || PubKey)`. If not, how should it be generated?

nkryuchkov: @noamnelke and I had a discussion about this — `GenerateAddress` should return the last 20 bytes of the public key. So I'll change this

common/types/address.go
nkryuchkov: I'll remove it according to https://github.com/spacemeshos/go-spacemesh/pull/3046#discussion_r781388327

common/types/address.go
nkryuchkov: https://github.com/spacemeshos/go-spacemesh/pull/3046#discussion_r781388828

common/types/address.go
nkryuchkov: https://github.com/spacemeshos/go-spacemesh/pull/3046#discussion_r781388828

common/types/address.go
nkryuchkov: https://github.com/spacemeshos/go-spacemesh/pull/3046#discussion_r781388828
@countvonzero I'm taking my last words back 🙂

neysofu (changes requested):
I agree with @noamnelke 's assessment on switching to `GenerateSpawnTransaction` in tests already and defining `GenerateCallTransaction` in the `svm` package. Otherwise LGTM.

activation/atxdb_test.go
```diff
@@ -741,12 +741,12 @@ func TestActivationDb_ContextuallyValidateAtx(t *testing.T) {
...
 	atxdb := NewDB(database.NewMemDatabase(), nil, idStore, layersPerEpochBig, goldenATXID, &ValidatorMock{}, lg.WithName("atxDB"))
 
-	validAtx := types.NewActivationTx(newChallenge(nodeID, 0, *types.EmptyATXID, goldenATXID, types.LayerID{}), [20]byte{}, nil, 0, nil)
+	validAtx := types.NewActivationTx(newChallenge(nodeID, 0, *types.EmptyATXID, goldenATXID, types.LayerID{}), types.Address{}, nil, 0, nil)
 	err := atxdb.ContextuallyValidateAtx(validAtx.ActivationTxHeader)
 	r.NoError(err)
 
 	arbitraryAtxID := types.ATXID(types.HexToHash32("11111"))
-	malformedAtx := types.NewActivationTx(newChallenge(nodeID, 0, arbitraryAtxID, goldenATXID, types.LayerID{}), [20]byte{}, nil, 0, nil)
+	malformedAtx := types.NewActivationTx(newChallenge(nodeID, 0, arbitraryAtxID, goldenATXID, types.LayerID{}), [types.AddressLength]byte{}, nil, 0, nil)
```
countvonzero: :s/[types.AddressLength]byte{}/types.Address{}

activation/atxdb_test.go
```diff
@@ -793,7 +793,7 @@ func BenchmarkGetAtxHeaderWithConcurrentStoreAtx(b *testing.B) {
 			pub, _, _ := ed25519.GenerateKey(nil)
 			id := types.NodeID{Key: util.Bytes2Hex(pub), VRFPublicKey: []byte("22222")}
 			for i := 0; ; i++ {
-				atx := types.NewActivationTx(newChallenge(id, uint64(i), *types.EmptyATXID, goldenATXID, types.NewLayerID(0)), [20]byte{}, nil, 0, nil)
+				atx := types.NewActivationTx(newChallenge(id, uint64(i), *types.EmptyATXID, goldenATXID, types.NewLayerID(0)), [types.AddressLength]byte{}, nil, 0, nil)
```
countvonzero: :s/[types.AddressLength]byte{}/types.Address{}

pendingtxs/pending_txs.go
```diff
@@ -136,7 +136,7 @@ func validTxWithHighestFee(txs map[types.TransactionID]nanoTx, balance uint64) t
 	bestID := types.EmptyTransactionID
 	var maxFee uint64
 	for id, tx := range txs {
-		if (tx.Fee > maxFee || (tx.Fee == maxFee && bytes.Compare(id[:], bestID[:]) < 0)) &&
+		if (tx.Fee > maxFee || (tx.Fee == maxFee && bytes.Compare(id[:], bestID[:]) < 0) || (tx.Fee == 0 && maxFee == 0)) &&
```
countvonzero: i think this will be clearer to rewrite to
```
if balance >= (tx.Amount+tx.Fee) && 
(bestID == types.EmptyTransactionID || // first tx with enough balance
 tx.Fee > maxFee || 
 (tx.Fee == maxFee && bytes.Compare(id[:], bestID[:]) < 0))
```

svm/transaction/transaction.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("serialize: %w", err)
+	}
+
+	sst := &types.Transaction{
+		InnerTransaction: inner,
+		Signature:        [64]byte{},
+	}
+
+	copy(sst.Signature[:], signer.Sign(buf))
```
countvonzero, line 52: why do this extra copy and not directly do
```
sst := &types.Transaction{
	InnerTransaction: inner,
	Signature:        signer.Sign(buf),
}
```

trie/trie_test.go
```diff
@@ -569,7 +569,7 @@ func BenchmarkHash(b *testing.B) {
 	random := rand.New(rand.NewSource(0))
 
 	// Create a realistic account trie to hash
-	addresses := make([][20]byte, b.N)
+	addresses := make([][types.AddressLength]byte, b.N)
```
countvonzero: addresses := make([]types.Address, b.N)

svm/transaction/transaction.go
nkryuchkov: @countvonzero I copied it from the existing code. If we assume that `signer.Sign(buf)` always returns a 64-byte result, your suggestion would work the same way.

pendingtxs/pending_txs.go
nkryuchkov: thanks

svm/transaction/transaction.go
nkryuchkov: @countvonzero I looked at it and actually `signer.Sign(buf)` returns a slice, but we need an array, so we need to convert it using `copy`

dshulyak:
@nkryuchkov is this change about moving code for generating transactions to svm module, any other important changes? can it be reflected in the pr title, in the current state the title is somewhat obscure

countvonzero (commented):
code LGTM

please update the PR content (currently empty) to detail what was done.

will leave the approval to @neysofu and @noamnelke

nkryuchkov:
@dshulyak @countvonzero thanks, updated

neysofu (approved):

nkryuchkov:
bors merge

neysofu:
Needs approval from @noamnelke

nkryuchkov:
bors merge

cmd/node/app_test.go
```diff
@@ -267,7 +268,7 @@ func txWithUnorderedNonceGenerator(dependencies []int) TestScenario {
 	txsSent := 25
 	setup := func(suite *AppTestSuite, t *testing.T) {
 		for i := 0; i < txsSent; i++ {
-			tx, err := types.GenerateCallTransaction(acc1Signer, dst, uint64(txsSent-i), 10, 1, 1)
+			tx, err := transaction.GenerateCallTransaction(acc1Signer, dst, uint64(txsSent-i), 10, 1, 1)
```
noamnelke, line 271: I haven't followed the entire test here, but if we didn't add a spawn transaction at nonce == 0, won't this test fail when those are needed?

We should probably make sure that tests without a spawn transaction fail.

cmd/node/app_test.go
```diff
@@ -329,8 +330,14 @@ func txWithRunningNonceGenerator(dependencies []int) TestScenario {
 				time.Sleep(250 * time.Millisecond)
 				actNonce = getNonce()
 			}
-			tx, err := types.GenerateCallTransaction(acc1Signer, dst, uint64(i), 10, 1, 1)
-			suite.NoError(err, "failed to create signed tx: %s", err)
+			var tx *types.Transaction
+			if i == 0 {
+				tx = transaction.GenerateSpawnTransaction(acc1Signer, dst)
```
noamnelke, line 335: I'm wondering how this can work... If the spawn transaction is basically empty and it is the only transaction that gets the nonce 0 and accounts are still initialized with nonce == 0 and skipping is not allowed. How can any transaction make it in?

mempool/tx_pool_test.go
```diff
@@ -173,7 +174,7 @@ func createBatch(t testing.TB, signer *signing.EdSigner) ([]*types.Transaction,
 	var txBatch []*types.Transaction
 	var txIDBatch []types.TransactionID
 	for i := uint64(0); i < 10000; i++ {
-		tx, err := types.GenerateCallTransaction(signer, types.Address{}, 5+1, 50, 100, 1)
+		tx, err := transaction.GenerateCallTransaction(signer, types.Address{}, 5+1, 50, 100, 1)
```
noamnelke, line 177: No spawn transactions in this file?

mesh/mesh_test.go
```diff
@@ -76,7 +77,7 @@ func addTransactions(t testing.TB, mesh *DB, layerID types.LayerID) []types.Tran
 	txs := make([]*types.Transaction, 0, numTXs)
 	txIDs := make([]types.TransactionID, 0, numTXs)
 	for i := 0; i < numTXs; i++ {
-		tx, err := types.GenerateCallTransaction(signing.NewEdSigner(), types.HexToAddress("1"), 1, 10, 100, rand.Uint64())
+		tx, err := transaction.GenerateCallTransaction(signing.NewEdSigner(), types.HexToAddress("1"), 1, 10, 100, rand.Uint64())
```
noamnelke, line 80: Spawn transactions?

mesh/mesh_test.go
```diff
@@ -545,7 +546,7 @@ func TestMesh_persistLayerHash(t *testing.T) {
 
 func addTxToMempool(t *testing.T, msh *Mesh, signer *signing.EdSigner, nonce uint64) *types.Transaction {
 	t.Helper()
-	tx := newTx(t, signer, nonce, 111)
+	tx := newCallTx(t, signer, nonce, 111)
```
noamnelke, line 549: Spawn?

mesh/meshdb_test.go
```diff
@@ -227,8 +234,8 @@ func TestMeshDB_GetStateProjection_WrongNonce(t *testing.T) {
 
 	signer, origin := newSignerAndAddress(t, "123")
 	err := mdb.writeTransactions(types.NewLayerID(1), types.BlockID{1},
-		newTx(t, signer, 1, 10),
-		newTx(t, signer, 2, 20),
+		newCallTx(t, signer, 1, 10),
```
noamnelke, line 237: Spawn?

miner/proposal_builder_test.go
```diff
@@ -74,7 +75,7 @@ func createBuilder(tb testing.TB) *testBuilder {
 
 func genTX(tb testing.TB, nonce uint64, recipient types.Address, signer *signing.EdSigner) *types.Transaction {
 	tb.Helper()
-	tx, err := types.GenerateCallTransaction(signer, recipient, nonce, 1, defaultGasLimit, defaultFee)
+	tx, err := transaction.GenerateCallTransaction(signer, recipient, nonce, 1, defaultGasLimit, defaultFee)
```
noamnelke, line 78: Spawn?

sql/transactions/transactions_test.go
```diff
@@ -28,9 +29,9 @@ func TestGetHas(t *testing.T) {
 	lid := types.NewLayerID(10)
 	bid := types.BlockID{1, 1}
 	txs := []*types.Transaction{
-		mustTx(types.NewSignedTx(1, types.Address{1}, 191, 1, 1, signer1)),
-		mustTx(types.NewSignedTx(1, types.Address{2}, 191, 1, 1, signer2)),
-		mustTx(types.NewSignedTx(1, types.Address{3}, 191, 1, 1, signer1)),
+		mustTx(transaction.GenerateCallTransaction(signer1, types.Address{1}, 1, 191, 1, 1)),
```
noamnelke, line 32: Aren't spawn transactions needed in this entire file?

svm/api_test.go
```diff
@@ -37,7 +38,7 @@ func (appliedTxsMock) NewBatch() database.Batch           { panic("implement me"
 func (appliedTxsMock) Find(key []byte) database.Iterator  { panic("implement me") }
 
 func createTransaction(t *testing.T, nonce uint64, destination types.Address, amount, fee uint64, signer *signing.EdSigner) *types.Transaction {
-	tx, err := types.GenerateCallTransaction(signer, destination, nonce, amount, 100, fee)
+	tx, err := transaction.GenerateCallTransaction(signer, destination, nonce, amount, 100, fee)
```
noamnelke, line 41: Spawn?

noamnelke (commented):
@nkryuchkov it seems that the addition of spawn transactions is inconsistent (sometimes you haven't done it). This made me realize that nothing checks for this. I think that tests should fail when a spawn transaction isn't added.

Additionally, I don't fully understand how it currently works WITH spawn transactions. Do they get processed and increment the account nonce? I couldn't see the code that does that. Otherwise, assuming the account nonce is initialized to zero, and we replace transactions with that nonce to spawn transactions - how does the first call transaction (with nonce == 1) get applied?

nkryuchkov:
@noamnelke 

I added spawn TXs everywhere where nonce was 0 in tests. In other cases, TXs were not submitted, so I assumed that nonce was not checked, therefore there was no need for spawn TXs, so I didn't add spawn TXs there.

> Do they get processed and increment the account nonce? I couldn't see the code that does that. 

Yes, when `SubmitTransaction` is called, the nonce is then incremented when a transaction is being processed here:
https://github.com/spacemeshos/go-spacemesh/blob/develop/pendingtxs/pending_txs.go#L123

In other words, I followed this rule when adding spawn TXs:

1) If a test calls `SubmitTransaction`, then a transaction will be processed and the nonce will be incremented. When `SubmitTransaction` is called, tests start with account nonce == 0. For this case, I added spawn TXs.

2) If a test doesn't call `SubmitTransaction`, a transaction is not processed, so there's no nonce increment. The tests start with nonce != 0 in this case. For this case, I didn't add spawn TXs.

Please let me know if my explanation is not clear.

cmd/node/app_test.go
nkryuchkov: This `txWithUnorderedNonceGenerator` test is intended to error at https://github.com/spacemeshos/go-spacemesh/blob/feature/move-tests-tx-gen-into-svm/cmd/node/app_test.go#L278 because nonces are out of order: from 25 to 0. Perhaps, it would look clearer if the last TX with nonce 0 would be spawn TX, but that wouldn't change the behavior, so IMO it doesn't matter here

cmd/node/app_test.go
nkryuchkov: What do you mean by "skipping is not allowed"? If the nonce is 0, a spawn TX is generated here, then nonce gets increased inside https://github.com/spacemeshos/go-spacemesh/blob/feature/move-tests-tx-gen-into-svm/cmd/node/app_test.go#L344, so next TXs become call TXs

mempool/tx_pool_test.go
nkryuchkov: No, because they are not submitted. There's no nonce == 0 in this file, so as I understand, there's no need for spawn TXs here. Please correct me if I'm wrong

mesh/mesh_test.go
nkryuchkov: https://github.com/spacemeshos/go-spacemesh/pull/3046#discussion_r790871476

mesh/mesh_test.go
nkryuchkov: https://github.com/spacemeshos/go-spacemesh/pull/3046#discussion_r790871476

noamnelke (approved):

nkryuchkov:
bors merge

Issue #2918, opened 2021-11-05
Database hardening proposal


dshulyak:
The proposal focuses on eliminating hard to debug bugs that will occur once spacemesh software will be running in non-managed environments, and besides bugs make database code more robust. The main issues with database code:

- all writes are not synchronous

If OS or hardware crashes, due to the power loss, for example, the node may enter into a state that is not visibly corrupted but invalid. One example could be https://github.com/spacemeshos/go-spacemesh/issues/2871, in case if ATX wasn't persisted but was broadcasted to another node - any future ATX that is produced by the crashed node is discarded by the network, but considered valid by the crashed node.

- in many places writes that must be atomic are not atomic

Examples of this are https://github.com/spacemeshos/go-spacemesh/issues/2516 and https://github.com/spacemeshos/go-spacemesh/issues/2547. Even though the software can be "designed" to handle non-atomic writes this is usually a bad idea and will lead to many unexpected bugs. In the case of ATXs and blocks none of them will be processed again if the main body of the object was written but the write for the secondary index failed. For example, the node will think that block is fully synced but we will not add it to the layer.

- non-atomic state transitions between go-sm and svm

Because of this design decision, there will be certain problems that we will need to handle. One example: marking a block during rerun contextually valid before rewinding state in svm. If we crash before rewinding the state we will never discover that block again in tortoise and the state will never again be reapplied.

We won't be able to solve such issues just by writing the correct database code, and they will have to be handled in a special way.

## Schema and requirements

- schema is simple, no complex queries
- handle large binary blobs efficiently (ATX is around ~3kb, block size depends on the tortoise state and can vary between ~1-10kb)
- database is mostly random reads, with low writes, so better to optimize for reads

we are also using databases in some other modules (tortoise, tortoise beacon), but they are not critical for correctness since we can always rebuild them from the main data.

#### mesh module

transactions and rewards are omitted, as i wasn't following what is getting moved to svm.

###### blocks

- main blob storage, indexed by block id
- layer index, indexed by layer_id || block_id
- contextual validity index, indexed by block_id, separate database
- input vector, indexed by layer id, separate database 

###### layers

separate database

- indexes for latest, processed, instate layers. using constant as an index
- indexes for hash and aggregated hash, using layer_id in each bucket

###### activations

all in a single database, but no atomic and synced writes

- blob storage for headers and bodies
- index from epoch_id || node_id to ATX 
- ATX timestamp, indexed by ATX id
[...]

PR #3047, opened 2022-01-07
[Merged by Bors] - sql: store ballots, blocks and layers in sqlite 

dshulyak:
## Motivation

The existing approach lacks atomicity, causal durability (e.g. an atx may not be on disk when a ballot is saved), and the durability can't be enforced in general without running every leveldb operation in sync mode. All this problems will result in subtle bugs that are hard to diagnose.

For those 3 requirements, we want to maintain all state in a single db. Moving state to a single leveldb will require us to enforce isolation ourselves (by maintaining separate namespace and manually concatenating keys like we do in some modules), beside that we have to create every single index manually while with sqlite we can just do `CREATE INDEX` and sqlite will do it for us and probably do a better job. 

Another significant benefit is that we can duplicate some state in sql table to avoid loading the whole structure into memory. For instance it will be relevant for atx, which is a large (10kb) and usually, after it was validated, we want to know only the associated smesher and weight of the atx. This would be problematic with leveldb, and would require adding custom index.

related: https://github.com/spacemeshos/go-spacemesh/issues/2918

## Changes
- general plumbing for core database stuff (db, transaction, migrations) using https://github.com/crawshaw/sqlite that is a relatively simple wrapper around C sqlite 
- tables and for layers, blocks and ballots
- reworked ZeroLayer, it is relevant only for hare_output, which can be in 3 states - nil, empty, non-empty. 
  SetZeroLayer update hare_output to empty state, at which tortoise will vote against all blocks within hdist. 
- updated to golang 1.16 for `embed` module. note that there is a bug with go mode tidy in 1.16 so i had to manually fix go.sum and disable go mod tidy on ci - https://github.com/golang/go/issues/44129

## Test Plan
existing and new uts


dshulyak:
bors try

countvonzero:
wow! super happy to move to relational database for data integrity. a couple high level questions

- do you observe any performance degradation?
in one of my previous project we ported geth to use sqlite because we need multi-threaded read and leveldb doesn't support that. my recollection was that there was a performance penalty due to the sheer amount of data.

- wrt to go version, why not go straight to 17? the tidy bug seems to be fixed at that version.

dshulyak:
> do you observe any performance degradation?

i observed an improvement with more data, and small degradation when there is very little data.
for example for loading 10 000 layers

```
benchmark                             old ns/op       new ns/op       delta
BenchmarkRerun/Verifying/10000-16     15456916985     13337943478     -13.71%
```

but relative improvement will be much larger than -13.71% when encoding will be replaced, as xdr wastes a lot of time.
with 100, 1000 layers leveldb is slightly faster, but i think this is less relevant.

i actually expected sqlite to be faster, as leveldb data structure is write-optimized. @countvonzero do you remember if `database/sql` was used in the previous project or a thin wrapper? there are some explanations why `database/sql` is not a good choice for sqlite, as sqlite is very different from other databases (https://turriate.com/articles/making-sqlite-faster-in-go, https://crawshaw.io/blog/go-and-sqlite).

 > wrt to go version, why not go straight to 17? the tidy bug seems to be fixed at that version.

@noamnelke was against updating versions if we are not using new features, so i decided not to rush with 1.17.  anyway i think we will want to use 1.18 in near future, as it introduces significant improvements to the language (fuzzing support in stdlib, and generics)

countvonzero:
> do you remember if database/sql was used in the previous project or a thin wrapper?

yes. we were. we didn't explore other libraries further because core geth devs did not support the change.

sql/ballots/ballots.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+func decodeBallot(id types.BallotID, sig, pubkey, body *bytes.Reader) (*types.Ballot, error) {
+	sigBytes := make([]byte, sig.Len())
+	if _, err := sig.Read(sigBytes); err != nil {
+		if err != io.EOF {
+			return nil, fmt.Errorf("copy sig %w", err)
+		}
+		sigBytes = nil
```
countvonzero, line 19: shouldn't this be an error too? same for pubkeyBytes
a ballot has to have signature and public key

sql/layers/layers.go
```diff
@@ -0,0 +1,95 @@
...
+	default:
+		panic(fmt.Sprintf("unknown status %d", s))
+	}
+}
+
+const (
+	// Latest layer that was added to db.
+	Latest Status = iota
+	// Processed layer is either synced from peers or updated by hare.
+	Processed = iota
```
countvonzero: this iota is not necessary

sql/layers/layers.go
```diff
@@ -0,0 +1,95 @@
...
+		return rst, fmt.Errorf("is empty %s: %w", lid, err)
+	} else if rows == 0 {
+		return rst, fmt.Errorf("%w hare output is not set for %s", sql.ErrNotFound, lid)
+	}
+	return rst, err
+}
+
+// SetStatus updates status of the layer.
+func SetStatus(db sql.Executor, lid types.LayerID, status Status) error {
+	if _, err := db.Exec(`insert into layers (id, status) values (?1, ?2)
```
countvonzero: use upsert?

sql/layers/layers.go
```diff
@@ -0,0 +1,95 @@
...
+			stmt.BindInt64(2, int64(status))
+		}, nil); err != nil {
+		return fmt.Errorf("insert %s %s: %w", lid, status, err)
+	}
+	return nil
+}
+
+// GetByStatus return latest layer with the status.
+func GetByStatus(db sql.Executor, status Status) (rst types.LayerID, err error) {
+	if _, err := db.Exec("select max(id) from layers where status >= ?1;",
```
countvonzero: what is your consideration for not using a separate status table to capture different values for latests/processed/applied? the cost of this query will grows as layers grow.

sql/migrations.go
```diff
@@ -0,0 +1,87 @@
...
+type Migrations func(Executor) error
+
+func embeddedMigrations(db Executor) error {
+	files, err := embedded.ReadDir("migrations")
+	if err != nil {
+		return fmt.Errorf("readdir migrations: %w", err)
+	}
+	var migrations []migration
+	for _, file := range files {
+		parts := strings.Split(file.Name(), "_")
```
countvonzero, line 33: what do you think about using regex to capture order?

sql/migrations/1_initial.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE blocks (
```
countvonzero: i think the filename should just be 0001.sql. the extra digits for file system ordering. the "initial" doesn't mean much.

sql/migrations/1_initial.sql
```diff
@@ -0,0 +1,27 @@
...
+);
+
+CREATE INDEX blocks_by_layer ON blocks(layer);
+CREATE INDEX blocks_by_validity ON blocks(layer, validity) where validity = 1;
+
+CREATE TABLE ballots ( 
+	id CHAR(20) PRIMARY KEY,
+	layer INT,
+    signature VARCHAR,
+    pubkey VARCHAR,
```
countvonzero: indentation?

sql/migrations/1_initial.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE blocks ( 
+	id CHAR(20) PRIMARY KEY,
+	layer INT,
```
countvonzero: why not use `not null` constraints in the schema?

sql/migrations/1_initial.sql
```diff
@@ -0,0 +1,27 @@
...
+    pubkey VARCHAR,
+	ballot BLOB
+);
+
+CREATE INDEX ballots_by_layer ON ballots(layer);
+
+CREATE TABLE layers (
+    id INT PRIMARY KEY,
+    hare_output VARCHAR,
+    status SMALL INT
```
countvonzero: the app never asks the status of a layer. the app only cares about the snapshot of each status.
i think it's more natural to keep a separate table for the states (latest/processed/applied)

sql/migrations.go
```diff
@@ -0,0 +1,87 @@
...
+type migration struct {
+	order   int
+	name    string
+	content *bufio.Scanner
+}
+
+// Migrations is interface for migrations provider.
+type Migrations func(Executor) error
+
+func embeddedMigrations(db Executor) error {
```
countvonzero, line 26: what is to be expected if migration fails to execute? does it require manual fixing database schema?

sql/migrations.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			return fmt.Errorf("invalid migration %s: %w", file.Name(), err)
+		}
+		path := filepath.Join("migrations", file.Name())
+		content, err := embedded.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("readfile %s: %w", path, err)
+		}
+		scanner := bufio.NewScanner(bytes.NewBuffer(content))
+		scanner.Split(func(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
countvonzero, line 47: i am surprised that you need to write your own parser

sql/migrations.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		for m.content.Scan() {
+			if _, err := db.Exec(m.content.Text(), nil, nil); err != nil {
+				return fmt.Errorf("exec %s: %w", m.content.Text(), err)
+			}
+		}
+		// binding values in pragma statement is not allowed
+		if _, err := db.Exec(fmt.Sprintf("PRAGMA user_version = %d;", m.order), nil, nil); err != nil {
+			return fmt.Errorf("update user_version to %d: %w", m.order, err)
+		}
```
countvonzero, line 84: roll back to the previous version if the migration failed? can these migration stmts wrapped in a tx?

sql/layers/layers.go
```diff
@@ -0,0 +1,95 @@
...
+			stmt.BindInt64(1, int64(lid.Value))
+			stmt.BindBytes(2, output[:])
+		}, nil); err != nil {
+		return fmt.Errorf("set hare output %s: %w", lid, err)
+	}
+	return nil
+}
+
+// GetHareOutput for layer.
+func GetHareOutput(db sql.Executor, lid types.LayerID) (rst types.BlockID, err error) {
```
countvonzero, line 49: i don't get the combo. "name returned value + returning explicitly"
i do prefer to not have named returned values.

sql/database.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	return db, nil
+}
+
+// Database is an instance of sqlite database.
+type Database struct {
+	pool *sqlitex.Pool
+}
+
+// Tx creates deferred sqlite transaction.
```
countvonzero, line 119: i am surprised that you need to write this wrappers in this file. what's the reason that the library doesn't provide them?

sql/blocks/blocks.go
```diff
@@ -0,0 +1,122 @@
...
+	}); err != nil {
+		return false, fmt.Errorf("select verified %s: %w", id, err)
+	} else if rows == 0 {
+		return false, fmt.Errorf("%w block %s is not in the database", sql.ErrNotFound, err)
+	}
+	return rst, err
+}
+
+// LayerIDs returns list of block ids in the layer.
+func LayerIDs(db sql.Executor, lid types.LayerID) ([]types.BlockID, error) {
```
countvonzero: the name implies returning a list of LayerID. call this LayerBlockIDs?

sql/blocks/blocks.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// SetValid updates verified status for a block.
+func SetValid(db sql.Executor, id types.BlockID) error {
+	return setValidity(db, id, 1)
+}
+
+// SetInvalid updates blocks to an invalid status.
+func SetInvalid(db sql.Executor, id types.BlockID) error {
+	return setValidity(db, id, -1)
```
countvonzero: put 1 and -1 as named constants? 
btw, why -1 instead of 0?

sql/migrations/1_initial.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE blocks ( 
+	id CHAR(20) PRIMARY KEY,
+	layer INT,
+	validity SMALL INT,
+	block    BLOB
+);
+
+CREATE INDEX blocks_by_layer ON blocks(layer);
+CREATE INDEX blocks_by_validity ON blocks(layer, validity) where validity = 1;
```
countvonzero: why only index on value 1? it seems unused. i thought you meant to select blocks where validity == 1 but i don't see that API in sql/blocks/blocks.go.

countvonzero (commented):
initial pass only on sql package

sql/blocks/blocks.go
dshulyak: 0 is a default value and means undecided

sql/blocks/blocks.go
dshulyak: blocks is a name of the module, maybe IDsInLayer

sql/layers/layers.go
dshulyak: err is assigned inside the closure. 

i think this is normal and shorter instead of defining rst and err as

```
var (
     rst ...
     err error
)
```

sql/layers/layers.go
dshulyak: this is an upsert. https://www.sqlite.org/lang_UPSERT.html

sql/layers/layers.go
dshulyak: this query will be super fast even after years of layers. we will store row for each layer, because of the hash, aggregated_hash, hare_output. and layers progress from latest -> processed -> applied.

so in my opinion performance is non-issue and it is better to avoid creating additional tables if we can reuse one table.

sql/migrations.go
dshulyak: if migration fails to execute it means that developer made a bug or database was manually altered. i don't see any case when manually fixing database schema will be needed

sql/migrations.go
dshulyak: why? split does what i need

sql/migrations.go
dshulyak: statement can be multiline and follows specific structure

sql/migrations.go
dshulyak: migrations are applied using transaction

sql/migrations/1_initial.sql
dshulyak: initial means exactly that this is initial schema, later there could be 2_add_some_field.sql . i can change it to 1.sql, but requiring a name for migration makes sense to me.

sql/migrations/1_initial.sql
dshulyak: yes i intended to use it, but then i removed methods that used this index. i will drop it

sql/migrations/1_initial.sql
dshulyak: what do you mean? indentation seems consistent

sql/migrations/1_initial.sql
dshulyak: i guess i could use them where appropriate, just didn't seem to matter

sql/database.go
dshulyak: i want db and tx to share common interface (Exec method), that's why it is implemented this way. otherwise, i don't find it surprising, as i mentioned this is a thin wrapper, it has some opinionated utilities for executing multiple statements atomically but they don't match what i need.

sql/migrations/1_initial.sql
dshulyak: what you described is covered using a specific API, i don't understand why it is natural to create more tables if same functionality can be provided with less bloat

sql/ballots/ballots.go
dshulyak: i made it this way because for testing i don't need signature and pubkey, can change it to a stricter validation

dshulyak: actually, it is very inconvenient to use valid signature and pubkey for testing, so i would rather leave it as it is, and enforce validation in other places (like it works now)

.github/workflows/ci.yml
```diff
@@ -60,10 +60,9 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: ${{ env.go-version }}
-      - name: fmt, tidy, lint
+      - name: fmt, lint
```
nkryuchkov: Why is `tidy` to be removed?

.golangci.yml
```diff
@@ -208,7 +208,7 @@ linters:
     - misspell
     - staticcheck
     - whitespace
-    - wrapcheck
+#    - wrapcheck
```
nkryuchkov, line 211: Why?

sql/ballots/ballots.go
countvonzero: hmmm... for me, the biggest benefit of using sql-like database is the guarantee on data integrity. if we cannot trust that whatever in the database is correct, then it lost a lot of its value to me.

we should have NOT NULL constraints for these two fields for ballot.

sql/migrations/1_initial.sql
countvonzero: on github, line 14-15 is indented less than line 12-13

.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
 
 env:
-  go-version: '1.15.13'
+  go-version: '1.16.13'
```
nkryuchkov: It also needs to be changed in:
- `.golangci.yml` (2 occurrences)
- `release.yml`
- `Makefile` (`install` target)

sql/migrations/1_initial.sql
countvonzero: i'm fine with the initial. but can we please have `0001` instead? i'd like them to be sorted in filesystem

sql/migrations.go
countvonzero: for me the benefit of regex allows me to see what's expected in the filename. it's self-documenting. but not important.

sql/migrations/1_initial.sql
countvonzero: we really should. like i stated earlier, this aspect of sql-like db is important. it guarantees that whatever we save in the DB is correct data.

sql/migrations/1_initial.sql
countvonzero: i don't think the number of tables is that important. i don't understand why layer should have a status attached to it.
they are not attribute of the layer but attributes of the mesh. a given layer can be `latest` and `processed` at the same time.

sql/migrations.go
countvonzero: can you elaborate a little more. where is the transaction enforced? 
say if i apply version 10-15 for a release, and 12 failed, will my sqlite remain in version 9 after the failure?

sql/layers/layers.go
countvonzero: i understand the syntax, i just prefer the longer syntax because i've seen quite a few subtle bugs caused by named return variables in long methods.

mesh/meshdb.go
```diff
@@ -116,29 +101,25 @@ func NewPersistentMeshDB(path string, blockCacheSize int, logger log.Log) (*DB,
 
 // PersistentData checks to see if db is empty.
 func (m *DB) PersistentData() bool {
-	if _, err := m.general.Get(constLATEST); err == nil {
-		m.Info("found data to recover on disk")
-		return true
+	lid, err := layers.GetByStatus(m.db, layers.Latest)
+	if err != nil || (lid == types.LayerID{}) {
```
nkryuchkov: What if this would be e.g. a connection error while DB is not empty? Perhaps this error could be compared to a not-found error?

sql/ballots/ballots.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/spacemeshos/go-spacemesh/codec"
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+func decodeBallot(id types.BallotID, sig, pubkey, body *bytes.Reader) (*types.Ballot, error) {
+	sigBytes := make([]byte, sig.Len())
+	if _, err := sig.Read(sigBytes); err != nil {
+		if err != io.EOF {
+			return nil, fmt.Errorf("copy sig %w", err)
```
nkryuchkov:
```suggestion
			return nil, fmt.Errorf("copy sig: %w", err)
```

sql/ballots/ballots.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		sigBytes = nil
+	}
+	pubkeyBytes := make([]byte, pubkey.Len())
+	if _, err := pubkey.Read(pubkeyBytes); err != nil {
+		if err != io.EOF {
+			return nil, fmt.Errorf("copy pubkey %w", err)
```
nkryuchkov:
```suggestion
			return nil, fmt.Errorf("copy pubkey: %w", err)
```

sql/ballots/ballots.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		pubkeyBytes = nil
+	}
+	inner := types.InnerBallot{}
+	if _, err := codec.DecodeFrom(body, &inner); err != nil {
+		if err != io.EOF {
+			return nil, fmt.Errorf("decoder body of the %s: %w", id, err)
```
nkryuchkov: looks like a typo
```suggestion
			return nil, fmt.Errorf("decode body of the %s: %w", id, err)
```

sql/migrations.go
dshulyak: enforced where migrations are applied. 

yes it will remain in version 9

sql/ballots/ballots.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Add ballot to the database.
+func Add(db sql.Executor, ballot *types.Ballot) error {
+	bytes, err := codec.Encode(ballot.InnerBallot)
+	if err != nil {
+		return fmt.Errorf("encode ballot %s: %w", ballot.ID(), err)
+	}
+	if _, err := db.Exec("insert into ballots (id, layer, signature, pubkey, ballot) values (?1, ?2, ?3, ?4, ?5);",
+		func(stmt *sql.Statement) {
```
nkryuchkov, line 45: Have you considered using something like https://github.com/jmoiron/sqlx for easier usage?

sql/ballots/ballots.go
dshulyak: it is a benefit for sure but not the most significant. i think that it doesn't matter much because we didn't enforce it with leveldb, so for me it is irrelevant if it will be enforced with sqlite.

what i see as a main benefit is that it is much harder to make mistakes writing custom keys, when i joined almost all mesh dbcode was subtly (and some not so much subtly) broken

sql/migrations/1_initial.sql
dshulyak: weird, it is the same for me

sql/blocks/blocks.go
nkryuchkov: I'd also prefer having named constants

sql/migrations/1_initial.sql
dshulyak: > i don't think the number of tables is that important

i think status reflects a stage in the life-cycle of the layer, latest -> processed -> applied. it always moves from left to right, so if the layer is processed layer it also implies that it can be latest.

> they are not attribute of the layer but attributes of the mesh.

i don't see why the are not attribute of the layer. for instance only applied layers has a hash and aggregated hash. processed layer has a hare_output. we actually could determine status just from looking at fields, but that would be more complex

.github/workflows/ci.yml
dshulyak: explained in pr description

sql/migrations/0001_initial.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE blocks (
```
nkryuchkov, line 1: Could you please fix the indentation in this file? The content of `CREATE TABLE blocks` is aligned by 1 tab, `ballots` by mixed 4 spaces/1 tab, `layers` by 4 spaces

.golangci.yml
dshulyak: i think this linter doesn't make sense. errors became very ugly, and in particular for my change i need to double-wrap errors almost with the same info

.github/workflows/ci.yml
dshulyak: thanks, didn't catch it with search

sql/ballots/ballots.go
dshulyak: sqlx relies on database/sql

sql/migrations/0001_initial.sql
dshulyak: didn't notice that, will fix

sql/ballots/ballots.go
nkryuchkov: Yes, but it has some wrappers that allow not to write things like `func(stmt *sql.Statement) {...}`

dshulyak: this is irrelevant, database/sql should not be used with sqlite

nkryuchkov: Why? `crawshaw.io/sqlite` has features that drivers for `database/sql` lack?

dshulyak: database/sql made design choices that do not work well for sqlite, as it is the only embedded sql database. from  the discussion in this change https://github.com/spacemeshos/go-spacemesh/pull/3047#issuecomment-1010667111

nkryuchkov: @dshulyak Thanks! I see

nkryuchkov (approved):

sql/migrations/1_initial.sql
countvonzero: so if layer 100 is set to `latest` first, and then set to `processed` later. when the app asked for latest layer, do you return 99 or 100?

countvonzero: > i think status reflects a stage in the life-cycle of the layer, latest -> processed -> applied. it always moves from left to right, so if the layer is processed layer it also implies that it can be latest.

except you don't guarantee that in code. and there is no where in the app we asked "what is the status of this layer?"
in our app. that's why this seem unnatural to me.

the only query in the app is about the mesh itself. what is the mesh's processed/applied/latest layer. for the purpose of spacemesh app, i don't think it's a property of the layer. it is the property of the mesh.

countvonzero: what does it mean when you say the layer is "latest"? not every layer get set to the latest. 
for example, if there is unstable network, we can set the latest from 1 to 3 if we receive data from layer 3 before 2.

countvonzero: and again, for a given layer, the status may not be set to anything. if there were out of order hare, then we could skip processed layer from N to N+5. same for verified layer.

sql/ballots/ballots.go
countvonzero: but what is the cost of adding them? why not add them? isn't it nice to be able to reap more benefit from sqlite?

sql/migrations/1_initial.sql
dshulyak: > so if layer 100 is set to latest first, and then set to processed later. when the app asked for latest layer, do you return 99 or 100?

100. because applied > processed > latest

> and again, for a given layer, the status may not be set to anything. if there were out of order hare, then we could skip processed layer from N to N+5. same for verified layer.

is it the case for verified? i thought it is always applied in order
but yes i can see how it can be confusing for latest and processed, i would probably fix that as well
ok i will add one more table

mesh/mesh.go
```diff
@@ -577,7 +577,7 @@ func (msh *Mesh) ProcessLayerPerHareOutput(ctx context.Context, layerID types.La
 		Status:  events.LayerStatusTypeApproved,
 	})
 
-	logger.Info("saving hare output for layer")
+	logger.With().Info("saving hare output for layer")
```
countvonzero: no need for `With()`

mesh/mesh.go
```diff
@@ -577,7 +577,7 @@ func (msh *Mesh) ProcessLayerPerHareOutput(ctx context.Context, layerID types.La
...
 	if err := msh.SaveHareConsensusOutput(ctx, layerID, blockID); err != nil {
 		logger.Error("saving layer hare output failed")
```
countvonzero: this needs `With()`

sql/migrations/0001_initial.sql
```diff
@@ -0,0 +1,26 @@
...
+    ballot BLOB
+);
+CREATE INDEX ballots_by_layer ON ballots(layer);
+
+CREATE TABLE layers (
+    id INT PRIMARY KEY,
+    hare_output VARCHAR
+) WITHOUT ROWID;
+
+CREATE TABLE layers_status (
```
countvonzero: how about mesh_status?

syncer/syncer_test.go
```diff
@@ -437,15 +437,16 @@ func TestSynchronize_BeaconDelay(t *testing.T) {
 		syncer.Close()
 	})
 
+	// allow synchronize to finish
+	feedLayerResult(gLayer.Add(1), lyr, mf, mm)
+
```
countvonzero, line 446: i cannot figure out why changes in this test is needed.

appveyor.yml
```diff
@@ -5,7 +5,7 @@ clone_folder: c:\gopath\src\github.com\spacemeshos\go-spacemesh\
 environment:
   GOPATH: c:\gopath
 
-stack: go 1.15
+stack: go 1.16
```
countvonzero: 17 as discussed. thanks

hare/hare.go
```diff
@@ -373,8 +373,6 @@ func (h *Hare) getGoodProposal(lyrID types.LayerID, epochBeacon types.Beacon, lo
 	if err != nil {
 		if err != database.ErrNotFound {
 			logger.With().Error("no proposals found for hare, using empty set", log.Err(err))
-		} else if err = h.mesh.SetZeroBallotLayer(lyrID); err != nil {
```
countvonzero, line 376: do you know why we don't need this now? i recall i added this due to tortoise test failures.

countvonzero (approved):

hare/hare.go
dshulyak: can't tell anything about tortoise failures. in general we only need to know if hare_output is empty or nil, other than that tortoise doesn't care if layer has any ballots/blocks

sql/ballots/ballots.go
dshulyak: added for layers, but can't add them for signature and pubkey because genesis ballot doesn't have them and it messes up some tests.

dshulyak:
bors merge

dshulyak:
bors merge

PR #3050, opened 2022-01-10
[Merged by Bors] - tortoise: load validity for all blocks

dshulyak:
## Motivation

During rerun tortoise should reload contextual validity for all blocks, otherwise tortoise will think that opinion on some old block is abstain, which will prevent verifying tortoise from making progress.

## Changes
- load contextual validity for all blocks

## Test Plan
TestRerunStaysInVerifyingMode and BenchmarkRerun/Verifying/*


countvonzero (approved):

dshulyak:
bors merge

PR #3051, opened 2022-01-10
[Merged by Bors] - mesh: prefixes in the same space should not overlap

dshulyak:
## Motivation

prefixes in the same space should not overlap. as it leads to corrupted database state.
for example keys `bl || abc` and `b || lzxd` overlap, so when we query for `bl` we could unexpectedly load object with key zxd.

## Changes
- use nonoverlapping prefixes for blocks and ballots

## Test Plan
i don't know how to add a failing test if prefixes are correct, so i added an example that would fail if prefixes are incorrect in a specific way.

### DIFF
--- a/common/types/block.go
+++ b/common/types/block.go
@@ -21,6 +21,11 @@ type BlockID Hash20
 // EmptyBlockID is a canonical empty BlockID.
 var EmptyBlockID = BlockID{}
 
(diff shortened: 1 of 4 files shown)


countvonzero (approved):
thanks for fixing this. sorry

dshulyak:
bors merge

PR #3053, opened 2022-01-12
[Merged by Bors] - tortoise: add OnBlock and OnBallot to tortoise interface

### LINKED ISSUE #3004: tortoise: if ballot votes for block that is not in the local opinion it should not be good
Tortoise check's that all votes for a block in our local opinion are consistent with ballot votes. If there is vote for a block that is not loaded into local opinion, because that block was late, we will mark ballot as good when it may be not.

there are several ways to fix it:
- ballot should have the same number of votes that we have in local opinion (e.g. if it votes for layer 9 - 10, and we have 8 ballots in those layers then ballot should have the same number, counting both implicit and explicit)
- add missing ballot to the state, since it is guaranteed to be in the database
- change tortoise interface to receive all ballots from ballots handler. we can actually split layer verification from receiving data

I tend to implement the last one, as for me splitting data from verification is a correct solution.

dshulyak:
## Motivation

Several bugs can occur because tortoise may miss state that wasn't available at the time when the layer was passed to the tortoise.

A ballot that uses a base ballot that wasn't available when layer was created will not be marked good and will implicitly vote for many blocks in the past. This problem will be fixed during rerun, but we can avoid it.

If block was received after layer was processed we won't count votes for it and can verify layer without deciding on that particular block. This again will be fixed by rerun, but it can be avoided as well.

Besides that OnBallot will be required for work on this issue https://github.com/spacemeshos/go-spacemesh/issues/3054 .

OnBlock will be removed after pruning is integrated, as votes will have all information that we need in tortoise (layer id and block id). 

closes: https://github.com/spacemeshos/go-spacemesh/issues/3004

## Changes
- introduce OnBlock and OnBallot methods for tortoise
- call them from mesh

## Test Plan
unit tests

### DIFF
--- a/mesh/interface.go
+++ b/mesh/interface.go
@@ -26,5 +26,7 @@ type state interface {
 }
 
 type tortoise interface {
+	OnBallot(*types.Ballot)
+	OnBlock(*types.Block)
 	HandleIncomingLayer(context.Context, types.LayerID) (oldPbase, newPbase types.LayerID, reverted bool)
 }
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -725,10 +725,18 @@ func (msh *Mesh) AddTXsFromProposal(ctx context.Context, layerID types.LayerID, 
 	return nil
 }
 
+// AddBallot to the mesh.
+func (msh *Mesh) AddBallot(ballot *types.Ballot) error {
+	msh.trtl.OnBallot(ballot)
+	return msh.DB.AddBallot(ballot)
+}
+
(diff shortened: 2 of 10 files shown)


nkryuchkov (approved):

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #3055, opened 2022-01-14
[Merged by Bors] - config: introduce preset with options for testnets

dshulyak:
## Motivation

Our current configuration approach is error prone and wastes a lot of time. For instance even for devnet217 config got messed up again, layer size was reverted to 30 while it was set specifically to 50, and some other parameters went missing.

This change introduces presets that override default values. For example, this is how i can launch devnet client using --preset option:

```
/home/dd/go-spacemesh/build/go-spacemesh --preset=testnet --network-id=217 --smeshing-start=true --smeshing-opts-datadir=./post/ -d=./data/ --poet-server=34.71.84.204:32695 --bootnodes=/ip4/34.121.232.230/tcp/5002/p2p/12D3KooWLpPxwu2tFJeaYsMi8Apa2USqmvRQP4AKzARWnwutK7wq --genesis-time=2022-01-12T16:07:59+05:30
```

## Changes
- add `--preset / -p` option that can use one of the presets registered in the module `config/presets`
- add preset for testnet/devnets

## TODO

- [ ] tests


nkryuchkov (approved):

config/presets/testnet.go
```diff
@@ -0,0 +1,75 @@
...
+		Accounts: map[string]uint64{
+			"0x8868600d2835019249a0f756f754d5a5ab17ebf0": 100000000000000000,
+			"0x9f120e40e330affe6aa27d5bda49ac0960df7d6f": 100000000000000000,
+			"0xafed9a1c17ca7eaa7a6795dbc7bee1b1d992c7ba": 100000000000000000,
+			"0xb20c3a31f973231e01bf2d6b5c00a22cc13b5c63": 100000000000000000,
+			"0xff083c9a22e05d3459b03f3dbed61c7ad6e0a209": 100000000000000000,
+		},
+	}
+
+	conf.AtxsPerBlock = 25
```
noamnelke: I think this one's no longer used, maybe let's clean it up?

noamnelke (approved):
Didn't we want to have a separate preset for devnets (faster than testnets)?

Otherwise LGTM.

cmd/node/node.go
```diff
@@ -192,19 +199,24 @@ func LoadConfigFromFile() (*config.Config, error) {
 	}
 
 	conf := config.DefaultConfig()
+	if name := viper.GetString("preset"); len(name) > 0 {
+		preset, err := presets.Get(name)
+		if err != nil {
+			return nil, err
+		}
+		conf = preset
+	}
```
countvonzero, line 208: i find it a little confusing that we do LoadConfig from the config file, pass the args to viper, and then load another config which may overwrite the previous config file. is it possible to do one or the other.

i.e. if i want to load config for testnet i do, `go-spacemesh --config=./configs/testnet.toml`, and if i want to load devnet i do `--config=./configs/devnet.toml`

dshulyak: can you clarify please? the whole point of this change is to have different sets of defaults (out of the last 6-7 networks only 2 were deployed with the correct config). config that is passed using --config has a highest priority, and will overwrite any state when loaded on L217

countvonzero: i guess i don't really like the idea of default config. i'd much prefer to have *every* cmdline flags explicitly specified in a config file, one per network type. (testnet/devent/mainet)

the fact that we have a baked-in default config, and then overwriting with another config file, is error prone. 
if there's any default values, then they are defined in the flag declaration.

but that's my preference (having every flags value set in one place).

countvonzero (approved):

dshulyak:
> Didn't we want to have a separate preset for devnets (faster than testnets)?

i think the only difference right now is that testnets are open and may use older code, and devnets are closed and deployed from the tip of develop. but in general yes we will want shorter layers for testing and local networks, can be added when they are needed

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

cmd/node/node.go
noamnelke: I think the whole purpose of this PR was to allow putting only network-specific parameters in the testnet/devnet config, like the network id, bootnodes, poet address, etc. and leave everything else at default levels. This has two advantages: 1. it's **less** error prone, because you only change settings you actually intend to, and 2. it allows seamlessly moving between versions on the node, even if the config options changed between them, as long as everything is taken from the presets (which will almost always be the case).

countvonzero: agreed that this PR make it less error-prone than before. but i was pushing a little bit more.

default options allow developers to be lazy. we are not forced to think whether the param make sense in different setups, as long as the unit tests pass. the SREs that i previously worked with basically reject them. they want explicit params set for different environments so they don't get surprised in production. when team grows, things like default options fall through cracks easily. so i was pushing for one catch-all config file for each setup.

noamnelke: I'm not sure I follow.

If we need to test that a configuration parameter can accept different values and the node will exhibit expected behavior - this needs to be tested explicitly. If we move the parameter from the preset to the config file that's hardcoded in Spacecraft - this doesn't change anything.

It's also worth noting that these parameters don't change between environments - they're global to a network and remain constant for all execution environments the node runs in. The only reason these parameters are configurable at all is that we want to make it easy to change the value in a single place if/when research decides, before genesis. After genesis, these parameters will never change anymore (with the exception of new testnets where you _could_ change parameters, but it's not clear _why_ you would).

PR #3058, opened 2022-01-17
[Merged by Bors] - sql: layers hashes, and rewards

dshulyak:
## Motivation

part of https://github.com/spacemeshos/go-spacemesh/issues/2918

## Changes
- move layers hashes (blocks hare and aggregated hash) to layers table
- reward table
- sql function add_uint64 to add rewards in sqlite (sqlite doesn't support uint64 natively)


nkryuchkov (approved):

mesh/mesh.go
```diff
@@ -369,11 +369,8 @@ func (msh *Mesh) persistLayerHashes(ctx context.Context, from, to types.LayerID)
 			return err
 		}
 
-		prevHash, err := msh.getAggregatedLayerHash(i.Sub(1))
-		if err != nil {
-			logger.With().Debug("failed to get previous aggregated hash", i, log.Err(err))
-			return err
-		}
+		prevHash := msh.GetAggregatedLayerHash(i.Sub(1))
```
countvonzero: this change cause the code to ignore error. this is not safe. we need to error out if we failed to get previous layer hash instead of using EmptyLayerHash

sql/functions.go
```diff
@@ -0,0 +1,16 @@
+package sql
+
+import (
+	"fmt"
+
+	"crawshaw.io/sqlite"
+)
+
+func registerFunctions(conn *sqlite.Conn) error {
```
countvonzero, line 9: can you comment on/explain what this function is for and why is it needed? i don't really understand

sql/layers/layers.go
```diff
@@ -93,3 +93,51 @@ func GetByStatus(db sql.Executor, status Status) (rst types.LayerID, err error)
...
+			stmt.BindBytes(2, hash[:])
+		}, nil); err != nil {
+		return fmt.Errorf("update %s to %s: %w", field, hash, err)
+	}
+	return nil
+}
+
+// SetHash updates hash for layer.
+func SetHash(db sql.Executor, lid types.LayerID, hash types.Hash32) error {
+	return setHash(db, "hash", lid, hash)
```
countvonzero: please use named constants for `hash` and `aggregated_hash`

sql/migrations/0001_initial.sql
```diff
@@ -17,10 +17,23 @@ CREATE INDEX ballots_by_layer ON ballots(layer);
...
 ) WITHOUT ROWID;
 
 CREATE TABLE mesh_status (
     status SMALL INT PRIMARY KEY,
     layer INT NOT NULL
-) WITHOUT ROWID;
\ No newline at end of file
+) WITHOUT ROWID;
+
+CREATE TABLE rewards (
```
countvonzero, line 30: why not create 0002 file? wouldn't sqlite check that you already have version 1?

sql/migrations/0001_initial.sql
```diff
@@ -17,10 +17,23 @@ CREATE INDEX ballots_by_layer ON ballots(layer);
...
+    smesher CHAR(64),
+    coinbase CHAR(20),
+    layer INT NOT NULL,
+    total_reward UNSIGNED LONG INT,
+    layer_reward UNSIGNED LONG INT,
+    PRIMARY KEY (smesher, layer)
```
countvonzero, line 36: i am weary of this primary key design. this constrains how we can write into database in the app. are we sure we don't ever want to write multiple rewards for the same smesher and layer? i think this should use ROWID instead

sql/rewards/rewards.go
```diff
@@ -0,0 +1,92 @@
...
+		LayerRewardEstimate: uint64(stmt.ColumnInt64(4)),
+		SmesherID:           *nodeid,
+	}
+	stmt.ColumnBytes(1, reward.Coinbase[:])
+	return reward, nil
+}
+
+// FilterByCoinbase filters rewards from all layers by coinbase address.
+func FilterByCoinbase(db sql.Executor, address types.Address) (rst []types.Reward, err error) {
+	if _, err := db.Exec("select smesher, coinbase, layer, total_reward, layer_reward from rewards where coinbase = ?1;",
```
countvonzero, line 56: do you want to do `order by layer desc`? seems most relevant are recent rewards. and then in the future we can perhaps limit the number of row retrieves if the data is too old.

sql/rewards/rewards.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+	); err != nil {
+		return nil, err
+	}
+	return rst, err
+}
+
+// FilterBySmesher filters rewards from all layers by smesher address.
+func FilterBySmesher(db sql.Executor, address []byte) (rst []types.Reward, err error) {
+	if _, err := db.Exec("select smesher, coinbase, layer, total_reward, layer_reward from rewards where smesher = ?1",
```
countvonzero, line 76: ditto

countvonzero (changes requested):

sql/migrations/0001_initial.sql
dshulyak: smesher rewards are aggregated per layer. coinbase can't be changed in the same layer. i think this design is correct

sql/migrations/0001_initial.sql
dshulyak: it should not be 0002. everything before mainnet launch goes into 0001, because migrations are for backward compatibility

sql/functions.go
dshulyak: sqlite doesn't have native support for uint64, so you can't add them or sort by uint64 . this function cast sqlite int64 type to uint64 and adds them

sql/rewards/rewards.go
dshulyak: i think the previous implementation was returning them in ascending order, which is also insert order. so i would preserve that behavior for now

mesh/mesh.go
dshulyak: done

sql/layers/layers.go
dshulyak: done

sql/migrations/0001_initial.sql
dshulyak: note that previously rewards were aggregated in the application, so there is definitely no reason to store multiple rows for the same smesher in the layer

countvonzero: it's easy enough to change this before mainnet. but i think for db schema it's better not to assume how the app will save it.

sql/functions.go
countvonzero: can you add comment that explain this. it's not obvious for readers.

countvonzero (approved):

sql/layers/layers.go
```diff
@@ -93,3 +98,51 @@ func GetByStatus(db sql.Executor, status Status) (rst types.LayerID, err error)
...
+		return rst, fmt.Errorf("select %s for %s: %w", field, lid, err)
+	} else if rows == 0 {
+		return rst, fmt.Errorf("%w layer %s", sql.ErrNotFound, lid)
+	}
+	return rst, err
+}
+
+// GetHash for layer.
+func GetHash(db sql.Executor, lid types.LayerID) (types.Hash32, error) {
+	return getHash(db, "hash", lid)
```
countvonzero: here too.

sql/layers/layers.go
```diff
@@ -93,3 +98,51 @@ func GetByStatus(db sql.Executor, status Status) (rst types.LayerID, err error)
...
+}
+
+// GetAggregatedHash for layer.
+func GetAggregatedHash(db sql.Executor, lid types.LayerID) (types.Hash32, error) {
+	return getHash(db, "aggregated_hash", lid)
```
countvonzero: ditto

sql/layers/layers.go
dshulyak: thanks, forgot about that

dshulyak:
bors merge

dshulyak:
flake in TestBuilder_RestartSmeshing

dshulyak:
bors merge

dshulyak:
bors merge

PR #3059, opened 2022-01-18
[Merged by Bors] - p2p, proposals, atx: don't broadcast already known data

dshulyak:
## Motivation

p2p layer uses bounded time-restricted cache for seen messages, as long as the message is in the cache it will not re-broadcast it and will not be passed to the handler. if node receives a message after it was evicted from the cache - it will be passed to the handler and broadcasted again. additionally, p2p layer uses lazy push, that periodically sends `ihave` message to peers. and if the peer doesn't have the message advertised in `ihave` - it will ask for it. i suspect that the could be some issues where one peer A evicts message earlier than the other peer B, peer B advertises this message to peer A and the message gets downloaded again. 

logs from testnet23 indicate that some old messages were received over and over again, there is not enough metrics to be sure that it is caused by behavior that i described above, but ignoring old message is safer than what we do now.

## Changes
- ignore previously received proposals and activations

### DIFF
--- a/activation/atxdb.go
+++ b/activation/atxdb.go
@@ -79,6 +79,7 @@ func getAtxTimestampKey(atxID types.ATXID) []byte {
 var (
 	errInvalidSig   = errors.New("identity not found when validating signature, invalid atx")
 	errGenesisEpoch = errors.New("tried to retrieve miner weights for target epoch 0")
+	errKnownAtx     = errors.New("known atx")
 )
 
 type atxChan struct {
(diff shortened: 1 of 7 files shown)


activation/atxdb.go
```diff
@@ -696,9 +706,7 @@ func (db *DB) HandleAtxData(ctx context.Context, data []byte) error {
 	existing, _ := db.GetAtxHeader(atx.ID())
```
countvonzero, line 706: what do you think about adding a HasATX() interface and just do a key lookup instead of loading the header

countvonzero (approved):

activation/atxdb.go
dshulyak: makes sense, probably simpler to add that after (or together with) sql changes

nkryuchkov (approved):

countvonzero:
actually ATX, proposals, ballots, blocks and TXs can all be fetched. @dshulyak

dshulyak:
> actually ATX, proposals, ballots, blocks and TXs can all be fetched. @dshulyak

for the rest i was aware, but i thought that we don't sync proposals. i thought that if hare terminates with some proposals that are not available locally we download block from peers and don't care about proposals

countvonzero:
> 

yeah. sorry my bad. the code should've returned nil on known proposals. 

at the app level, there is incentive to try harder to push the consensus along. when hare fails, that means tortoise will generate ballots that vote against blocks in that layer for hdist layers.

dshulyak:
@countvonzero can you check please if the last commit works for you? i need to remove that AddProposal so that it can be added in validator before publishing

countvonzero:
> @countvonzero can you check please if the last commit works for you? i need to remove that AddProposal so that it can be added in validator before publishing

yes. but can you add some comment there to say the proposal will be saved before publishing for future readers? thanks.

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

Issue #3060, opened 2022-01-18
tortoise, mesh: handle the case when verified layer didn't reach historically verified layer during rerun


dshulyak:
during rerun we track if opinion on blocks has changed, and if it did we will notify mesh to revert the state. this doesn't cover the case when we failed to decide on blocks that were previously decided. this could happen because during rerun we count votes from all known layers. 

this caused miner-11 to crash with the following exception, in short applied state was layer 2517, but according to the tortoise latest verified was 2516. which effectively went backward, and this is only expected by mesh if the revert is true.

```
{"L":"PANIC","T":"2022-01-18T01:40:29.803-0500","N":"26605.mesh         ","M":"update state out-of-order","node_id":"26605f4fb2c89bef0d1fcc6165b54234736eeca8e4ba947643a953332c5f8e42","module":"mesh","sessionId":"62989d20-a9a1-4b6c-b663-25d037b242d1","verified":2516,"latest":2517,"name":"mesh"}
panic: update state out-of-order

goroutine 2660 [running]:
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc007d980c0, 0xc005f70300, 0x3, 0x3)
	/go/pkg/mod/go.uber.org/zap@v1.19.0/zapcore/entry.go:232 +0x51f
go.uber.org/zap.(*Logger).Panic(0xc00c7150e0, 0x13e7a34, 0x19, 0xc005f70300, 0x3, 0x3)
	/go/pkg/mod/go.uber.org/zap@v1.19.0/logger.go:230 +0x85
github.com/spacemeshos/go-spacemesh/log.FieldLogger.Panic(0xc00c7150e0, 0x13ccac7, 0x4, 0x13e7a34, 0x19, 0xc00f4476c8, 0x2, 0x2)
	/go/src/github.com/spacemeshos/go-spacemesh/log/zap.go:288 +0x1d0
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).updateStateWithLayer(0xc0005bab40, 0x16a0d20, 0xc00c79ad50, 0xc0000009d4, 0xc005d473b0, 0x0, 0x0)
	/go/src/github.com/spacemeshos/go-spacemesh/mesh/mesh.go:593 +0x3e5
github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).pushLayer(0xc0005bab40, 0x16a0d20, 0xc00c79ad50, 0x9d4, 0x17, 0x0)
[...]

PR #3061, opened 2022-01-18
[Merged by Bors] - mesh: error if tortoise reverts layer without notification

dshulyak:
## Motivation

This is a stop-gap to prevent crashing node, so that we can release a network. Will work on a proper fix in the meantime.

related: https://github.com/spacemeshos/go-spacemesh/issues/3060 

## Changes
- error if tortoise reverts layer without notifying mesh

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -434,7 +434,7 @@ func (msh *Mesh) pushLayersToState(ctx context.Context, from, to types.LayerID) 
 	// we never reapply the state of oldVerified. note that state reversions must be handled separately.
 	for layerID := from; !layerID.After(to); layerID = layerID.Add(1) {
 		if !layerID.After(msh.latestLayerInState) {
-			logger.With().Warning("trying to apply layer before currently applied layer",
+			logger.With().Error("trying to apply layer before currently applied layer",
 				log.Stringer("applied_layer", msh.latestLayerInState),
 				layerID,
 			)
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -317,13 +317,6 @@ func (vl *validator) ProcessLayer(ctx context.Context, layerID types.LayerID) er
 			return err
 		}
 	}
-	if !reverted && newVerified.Before(vl.latestLayerInState) {
-		logger.With().Error("tortoise reverted state without notifying mesh",
-			log.Stringer("applied_layer", vl.latestLayerInState),
-			log.Stringer("verified_layer", newVerified),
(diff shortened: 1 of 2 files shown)


mesh/mesh.go
```diff
@@ -317,6 +317,14 @@ func (vl *validator) ProcessLayer(ctx context.Context, layerID types.LayerID) er
 			return err
 		}
 	}
+	if !reverted && newVerified.Before(vl.latestLayerInState) {
```
nkryuchkov:
```suggestion
	} else if newVerified.Before(vl.latestLayerInState) {
```
looks a bit less redundant, but up to you

nkryuchkov (approved):

mesh/mesh.go
dshulyak: i changed to a more robust variant

mesh/mesh.go
```diff
@@ -432,6 +433,13 @@ func (msh *Mesh) pushLayersToState(ctx context.Context, from, to types.LayerID)
 	missing := msh.MissingLayer()
 	// we never reapply the state of oldVerified. note that state reversions must be handled separately.
 	for layerID := from; !layerID.After(to); layerID = layerID.Add(1) {
+		if !layerID.After(msh.latestLayerInState) {
+			logger.With().Warning("trying to apply layer before currently applied layer",
```
countvonzero: should be an error log

mesh/mesh.go
```diff
@@ -432,6 +433,13 @@ func (msh *Mesh) pushLayersToState(ctx context.Context, from, to types.LayerID)
...
+				log.Stringer("applied_layer", msh.latestLayerInState),
+				layerID,
+			)
+			continue
```
countvonzero, line 441: why not immediately return an error? do you mean to skip layers that are <= applied and then apply those higher? this will result in state root difference. right?

countvonzero (approved):
i guess fork caused by different state root is acceptable

mesh/mesh.go
dshulyak: > do you mean to skip layers that are <= applied and then apply those higher? 

yes

> this will result in state root difference. right?

it may, but not necessarily.
i think it is not as bad as it seems, because after next rerun (+1 more hour) state diff will be resolved anyway

mesh/mesh.go
dshulyak: changed

dshulyak:
bors merge

PR #3064, opened 2022-01-19
[Merged by Bors] - sql, mesh, api: refactor transactions

dshulyak:
## Motivation

part of: https://github.com/spacemeshos/go-spacemesh/issues/2918

## Changes
- table for transactions to replace transactions and unappliedTxs databases
- add API to filter transactions for multiple layers. instead of querying layer by layer
- simplify logic around pending transactions (transactions now added as pending to the table, and either switch to applied or marked as deleted)


mesh/mesh.go
```diff
@@ -499,21 +500,14 @@ func (msh *Mesh) reInsertTxsToPool(applied *types.Block, notApplied []*types.Blo
 	if len(missing) > 0 {
 		msh.With().Error("could not reinsert transactions", log.Int("missing", len(missing)), l)
 	}
-
-	grouped, accounts, err := msh.removeFromUnappliedTxs(returnedTxs)
-	if err != nil {
+	if err := msh.deleteTransactions(returnedTxs...); err != nil {
```
countvonzero: this doesn't means delete the entry from DB. this means set the TX to applied. we should never delete TXs from db.

sql/transactions/transactions.go
```diff
@@ -0,0 +1,185 @@
...
+	}, nil); err != nil {
+		return fmt.Errorf("applied %s: %w", id, err)
+	} else if rows == 0 {
+		return fmt.Errorf("%w: tx %s", sql.ErrNotFound, id)
+	}
+	return nil
+}
+
+// Delete transaction from database.
+func Delete(db sql.Executor, id types.TransactionID) error {
```
countvonzero: we don't need this interface. not until we decided how to deal with old data.

sql/transactions/transactions.go
```diff
@@ -0,0 +1,185 @@
...
+		func(stmt *sql.Statement) {
+			stmt.BindBytes(1, id.Bytes())
+		}, nil)
+	if err != nil {
+		return false, fmt.Errorf("has %s: %w", id, err)
+	}
+	return rows > 0, nil
+}
+
+// query MUST ensure that this order of fields tx, layer, block, origin, id.
```
countvonzero, line 133: prolly safer to define the select stmt as a string. `select tx, layer, block, origin, id`

countvonzero (changes requested):

mesh/mesh.go
dshulyak: > this doesn't means delete the entry from DB. this means set the TX to applied. 

i think this is false. applied is set for transactions that were *applied* to the state.

> we should never delete TXs from db.

i think we should. if tx is not applied we are adding it back to mempool. later it will be included with another block or dropped from the mempool

countvonzero: right. this is different from current behavior but i agree to delete here.

countvonzero (approved):

mesh/mesh.go
dshulyak: i misunderstood current behavior. we were removing from unnaplied db but not from transactions db. maybe i should keep them and rename applied to something that makes more sense

dshulyak: i think maybe the node can enter some unexpected state if a block wasn't applied and later tortoise changed opinion and block became applied. so i guess it is safer to preserve original behavior for now

countvonzero: > i think maybe the node can enter some unexpected state if a block wasn't applied and later tortoise changed opinion and block became applied. so i guess it is safer to preserve original behavior for now

yes. a TX is added to db only when a proposal is received via gossip or when a block is fetched. 
if you delete here, you will have to add it back from the block when it become contextually valid from invalid. that will make sense too.

dshulyak: i also thought about that. the reason why i wasn't sure that this is safe is because nodes during self-healing will vote for a block even if transactions are not stored locally. this may lead to a condition where the block becomes valid during self-healing, but data is not available to honest nodes.

so until we implement voting against the blocks with missing data we probably can't do what i wanted

dshulyak:
bors merge

PR #3075, opened 2022-01-23
[Merged by Bors] - beacon: always return beacon for epoch 2 from memory

### LINKED ISSUE #3073: devnet: failed to get beacon for epoch 2 in tortoise rerun
## Description

tortoise rerun check votes from genesis and will keep trying to get beacon for epoch 2 to verify ballots in epoch 2.
however, genesis beacon is set in memory only and not persisted. therefore when beacon's memory evicted the old epoch beacons (currently it keeps 10 epochs), it produce this log.

this also means that ballots in epoch 2 will not be counted correctly during rerun once the beacon value is evicted.

countvonzero:
## Motivation

Closes #3073

## Changes

- move all mocks to one interface.go file
- make beacons store calculated beacon by target epoch
- always return bootstrap beacon for epoch <= 2

## Test Plan

unit tests

## DevOps Notes


countvonzero:
bors try

countvonzero:
bors try

common/types/layer.go
```diff
@@ -14,8 +14,8 @@ const (
 	// LayerIDSize in bytes.
 	LayerIDSize = 4
 
-	// GenesisBeacon is the hex value of the beacon used during genesis.
-	GenesisBeacon = "0xaeebad4a796fcc2e15dc4c6061b45ed9b373f26adfc798ca7d2d8cc58182718e" // sha256("genesis")
+	// BootstrapBeacon is the hex value of the beacon used during genesis.
+	BootstrapBeacon = "0x333c04dd151a2a6831c039cb9a651df29198be8a04e16ce861d4b6a34a11c954" // sha256("bootstrap")
```
dshulyak, line 18: i would rather leave it as 32 zeros

dshulyak (approved):

nkryuchkov (approved):

common/types/layer.go
countvonzero: all zeros (types.EmptyBeacon) is used in some places as error condition. at this point i don't want to spend more time on this. i am assuming when we are getting ready for mainnet, this will change again.

countvonzero:
bors merge

PR #3076, opened 2022-01-23
[Merged by Bors] - cmd: add flag for genesis accounts

dshulyak:
For system tests, we want to have additional keys with funds in order to create transactions without waiting for rewards to accumulate. It is convenient to create those keys when the cluster is created and pass the addresses using command line.

## Changes
- add -a/-accounts map[string]uint64 flag


countvonzero (approved):

dshulyak:
bors merge

PR #3082, opened 2022-01-24
[Merged by Bors] - tortoise: setup for model based tests

dshulyak:
this pr implements initial infrastructure for writing model based tests, including one example.
the end goal is to implement sophisticated infra for writing long running tortoise simulations (including failure conditions), with sophisticated properties.

code organization heavily borrows from https://p-org.github.io/P/. every independent model (subsystem) is implemented as a separate object. core, beacon, hare are all independent models. they interact by exchanging messages. it enables simulation of failures for independent components. messages may get dropped or reordered. additionally every model may send events, unlike messages events are not dropped or reordered and guaranteed to be received by monitors.

TestBasicModel implements a simulation where part of smeshers are failing at every layer, but the tortoise is expected to continue verifying layers.


dshulyak:
bors try

tortoise/model/beacon.go
```diff
@@ -0,0 +1,33 @@
...
+// beacon outputs beacon at the last layer in epoch.
+type beacon struct {
+	rng *rand.Rand
+}
+
+// OnMessage ...
+func (b *beacon) OnMessage(m Messenger, event Message) {
+	switch ev := event.(type) {
+	case MessageLayerStart:
+		if ev.LayerID.GetEpoch() == ev.LayerID.Sub(1).GetEpoch() {
```
countvonzero: consider doing `if ev.LayerID == ev.LayerID.GetEpoch().FirstLayer()`

tortoise/model/runner.go
```diff
@@ -0,0 +1,119 @@
...
+		rng:         rng,
+		probability: probability,
+	}
+}
+
+type failingRunner struct {
+	cluster     *cluster
+	messenger   Messenger
+	monitors    []Monitor
+	probability [2]int
```
countvonzero, line 74: i don't understand `probability`, can you add comment to explain? thanks

tortoise/model/runner_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/spacemeshos/go-spacemesh/log/logtest"
+)
+
+func TestBasicModel(t *testing.T) {
+	types.SetLayersPerEpoch(4)
+	const total = 20
+
+	rng := rand.New(rand.NewSource(1001))
+	c := newCluster(logtest.New(t), rng)
+	for i := 0; i < 50; i++ {
```
countvonzero: make 50 a named constant

tortoise/model/runner_test.go
```diff
@@ -0,0 +1,31 @@
...
+		c.addCore()
+	}
+	c.addHare().addBeacon()
+
+	msgr := reliableMessenger{}
+	monitor := newVerifiedMonitor(t, types.GetEffectiveGenesis())
+
+	r := newFailingRunner(c, &msgr, []Monitor{monitor}, rng, [2]int{5, 100}).
```
countvonzero, line 28: make 5 and 100 named constants

tortoise/model/runner_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"math/rand"
+	"testing"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/log/logtest"
+)
+
+func TestBasicModel(t *testing.T) {
+	types.SetLayersPerEpoch(4)
+	const total = 20
```
countvonzero: :s/total/numLayers ?

tortoise/model/core.go
```diff
@@ -0,0 +1,200 @@
...
+		// TODO(dshulyak) produce ballot according to eligibilities
+		if !ev.LayerID.After(types.GetEffectiveGenesis()) {
+			return
+		}
+		if c.refBallot == nil {
+			total, _, err := c.atxdb.GetEpochWeight(ev.LayerID.GetEpoch())
+			if err != nil {
+				panic(err)
+			}
+			c.eligibilities = max(uint32(c.weight*50/total), 1)
```
countvonzero: what's 50?

tortoise/model/core.go
```diff
@@ -0,0 +1,200 @@
...
+			NodeID:     types.NodeID{Key: c.signer.PublicKey().String()},
+			StartTick:  1,
+			EndTick:    2,
+			PubLayerID: ev.LayerID,
+		}
+		atx := types.NewActivationTx(nipost, types.BytesToAddress(c.signer.PublicKey().Bytes()), nil, uint(c.units), nil)
+
+		c.refBallot = nil
+		c.atx = atx.ID()
+		c.weight = atx.GetWeight()
```
countvonzero, line 145: i'd imagine these (refBallot/atx/weight) should be reset in MessageLayerStart event?

tortoise/model/runner.go
```diff
@@ -0,0 +1,119 @@
...
+	if r.failables == nil {
+		r.failables = map[reflect.Type]struct{}{}
+	}
+	for _, ev := range events {
+		r.failables[reflect.TypeOf(ev)] = struct{}{}
+	}
+	return r
+}
+
+func (r *failingRunner) isFailable(ev Message) bool {
```
countvonzero, line 98: the naming is not straightforward to me. what does it mean for a message to be failable?

tortoise/model/runner.go
```diff
@@ -0,0 +1,119 @@
...
+	if r.failables == nil {
+		return true
+	}
+	_, exist := r.failables[reflect.TypeOf(ev)]
+	return exist
+}
+
+func (r *failingRunner) consume() {
+	for msg := r.messenger.PopMessage(); msg != nil; msg = r.messenger.PopMessage() {
+		if r.isFailable(msg) && r.probability[0] > rand.Intn(r.probability[1]) {
```
countvonzero, line 108: i don't understand this line. can you explain why if this true we don't broadcast the message?

countvonzero (commented):
what do you think of adding anther MessageEpochStart to separate those per-epoch data and per-layer data?

tortoise/model/runner_test.go
dshulyak: in this case it doesn't make sense, this is just a parameters, can be read as 5/100

tortoise/model/runner.go
dshulyak: if the message is "failable" and a dice roll is below threshold. so for example if i am using 5/100 then only 5% of all messages will be dropped, if am using 1/2 then i can roll only 0 or 1 and 50% will be dropped

tortoise/model/runner.go
dshulyak: message capable of failing?

tortoise/model/runner.go
dshulyak: "probability that message will fail" is good enough?

tortoise/model/core.go
dshulyak: hardcode for the number of ballots, will change it to the consntant
i left a todo on top to enable eligibility computation in the model

tortoise/model/runner.go
countvonzero: oh! you are making it fail automatically. ok. got it. thanks

tortoise/model/runner.go
countvonzero: how about "probability that messages in `failables` will fail by the model"

tortoise/model/core.go
countvonzero: make it a named constant like `numBallots` bellow the TODO?

tortoise/model/core.go
dshulyak: it is more convenient to reset  atx and weight in the last layer of the previous epoch, otherwise i will have to keep a reference to the atx

dshulyak:
> what do you think of adding anther MessageEpochStart to separate those per-epoch data and per-layer data?

actually even LayerEnd event needs to be removed. LayerStart is like a regular clock event, and layer end needs to be replaced. i don't see a good reason to add more events

countvonzero (approved):
i think the code will be clearer to have MessageEpochStart/MessageEpochEnd also.

countvonzero:
> > what do you think of adding anther MessageEpochStart to separate those per-epoch data and per-layer data?
> 
> actually even LayerEnd event needs to be removed. LayerStart is like a regular clock event, and layer end needs to be replaced. i don't see a good reason to add more events

yeah. i think LayerEnd doesn't really make sense. for me MessageEpochStart/MessageLayerStart make more sense in the modeling.

dshulyak:
> i think the code will be clearer to have MessageEpochStart/MessageEpochEnd also.

i will experiment with them once i will get back to this

dshulyak:
bors merge

Issue #2999, opened 2021-12-13
Missing check for duplicate use of voting eligibility


noamnelke:
When processing a ballot or block, we never check that a valid eligibility isn't used more than once.

A malicious smesher could create an unlimited number of blocks/ballots in a layer all using an identical eligibility proof.

When we detect this - the expected behaviour is to blacklist the publishing smesher - a mechanism we don't have yet.

Handling this is covered under https://github.com/spacemeshos/go-spacemesh/issues/2998 (by validating that we have no more than a single ballot from a smesher in a layer and that eligibility proofs are all distinct).

dshulyak:
I also discovered this in https://github.com/spacemeshos/go-spacemesh/issues/844 and after my question on slack.

dshulyak:
copying my message from that issue:

From Tal on slack:

> The translation to our case is if someone publishes two ATXs with the same id and the same epoch, or two ballots with the same ATX in the same slot. This is verifiably malicious behavior, and should cause the ID to be canceled (all of its ballots have weight 0, and the node should vote against any type of reward being given to this ID) .

I don't know if we are going to implement proofs of malicious behaviour for mainnet. But the solution for this particular issue should be:
- if there are two or more ballots using the same atxid and same slot they should not be counted by the tortoise (e.g. as the weight is 0)
- we can verify it when the ballot is received and add an additional bit to the ballot stored on disk
- tortoise should load this bit and act accordingly
- at least two malicious ballots can't be pruned, as then we won't be able to identify that ballots were malicious

dshulyak:
This is actually tricky to get right. Storing ballots with non-unique eligibility is a no-go as it makes network vulnerable to spam, and allows to consume storage space without bound. And the fact that this attack is very cheap makes it important to handle it correctly. But there are several problems:

1. Let's assume that 2 nodes received 2 ballots from the same smesher in the same layer but in a different order. If they will learn too late about the 2nd ballot they may use 1st as a base ballot. This will create issues for the tortoise, as each node won't be able to decode some part of the votes.

A simple solution is to let fetcher do its job, and it will invalidate 1st ballot if 2nd ballot is selected as a base ballot. This still will make some votes useless, as tortoise won't be able to decode votes for blocks before base ballot layer.

A more complex solution is to let every node broadcast 1st ballot that has a conflict, every node will *very likely* receive two malicious ballots before it needs to cast votes.

[...]

PR #3085, opened 2022-01-25
[Merged by Bors] - proposals, tortoise: consolidate prosals from the same layer

### LINKED ISSUE #2998: Consolidate ballots and proposals with same smesher and layer to one
1. Update the `EligibilityProof` in `InnerBallot` to `EligibilityProofs`—a slice of `VotingEligibilityProof`s. https://github.com/spacemeshos/go-spacemesh/blob/a22d1eeee6f593b54fbb5f89fb32573f965a6528/common/types/ballot.go#L38-L44
2. Update `proposal_builder.go`→`createProposal()` https://github.com/spacemeshos/go-spacemesh/blob/a22d1eeee6f593b54fbb5f89fb32573f965a6528/miner/proposal_builder.go#L245-L253 to accept a slice of eligibility proofs.
3. Unwrap the code in this loop: https://github.com/spacemeshos/go-spacemesh/blob/8951b6fa049c8f91e774e69f751973eebc34d690/miner/proposal_builder.go#L362-L412 in `proposal_builder.go`→`handleLayer()` - instead of looping, pass `proofs` into `createProposal` directly.
4. In `handler.go`→`checkBallotDataIntegrity()` https://github.com/spacemeshos/go-spacemesh/blob/a22d1eeee6f593b54fbb5f89fb32573f965a6528/proposals/handler.go#L242-L280 add a check that counter values (`J`) of all eligibility proofs are distinct.
5. In `handler.go`→`processBallot()` https://github.com/spacemeshos/go-spacemesh/blob/a22d1eeee6f593b54fbb5f89fb32573f965a6528/proposals/handler.go#L204-L218 add a check for an existing ballot by the same smesher in the same layer. If such a different ballot already exists in our database the expected behaviour is to blacklist the smesher and publish a fraud-proof consisting of the two ballots. Since we don't have this mechanism implemented (and it's non-trivial to implement) we should create a stub method as follows and call it:

```golang
func Blacklist(types.SmesherID) {
    // TODO: Implement me!
}
```

6. Change the following method to account for multiple eligibility proofs in a ballot. This will probably require changing the `weights` data structure or augmenting it with another one for the weight of a single ballot of a smesher in an epoch. https://github.com/spacemeshos/go-spacemesh/blob/a22d1eeee6f593b54fbb5f89fb32573f965a6528/tortoise/threshold.go#L10-L59

We currently don't limit the number of transactions in a proposal. Sizing proposals based on ballot weight will be part of the task of limiting the size (part of the transaction selection epic that's being written these days).

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/2998

## Changes
- consolidate proposals by extending ballot with an array of eligibilities
- request txs based on the number of eligibilities, also it seems we don't enforce max number of txs on the receiver side
- invalidate proposals that have eligibility counter out of order
- in tortoise introduce referenceWeight map which will store weight per single eligibility
- multiply rewards by the number of eligibilities


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

blocks/generator.go
```diff
@@ -106,7 +106,11 @@ func (g *Generator) extractOrderedUniqueTXs(layerID types.LayerID, proposals []*
 }
 
 func (g *Generator) calculateSmesherRewards(logger log.Log, layerID types.LayerID, proposals []*types.Proposal, txs []*types.Transaction) ([]types.AnyReward, error) {
-	rInfo := calculateRewardPerProposal(layerID, g.cfg, txs, len(proposals))
+	eligibilities := 0
+	for _, proposal := range proposals {
+		eligibilities += len(proposal.EligibilityProofs)
+	}
+	rInfo := calculateRewardPerProposal(layerID, g.cfg, txs, eligibilities)
```
countvonzero: rename to `calculateRewardPerEligibility`

common/types/ballot.go
```diff
@@ -42,8 +43,8 @@ type InnerBallot struct {
 	// the smesher's ATX in the epoch this ballot is cast.
 	AtxID ATXID
 	// the proof of the smesher's eligibility to vote and propose block content in this epoch.
```
countvonzero, line 45: please update comment regarding the order expected

cmd/node/app_test.go
```diff
@@ -621,9 +624,9 @@ func (suite *AppTestSuite) validateBlocksAndATXs(untilLayer types.LayerID) {
 	suite.log.Info("node %v", suite.apps[0].nodeID.ShortString())
 
 	totalBallots := 0
-	for id, l := range nodedata.layertoballots {
-		totalBallots += len(l)
-		suite.log.Info("node %v: layer %v, ballots %v", suite.apps[0].nodeID.ShortString(), id, len(l))
+	for id, l := range nodedata.eligibilities {
+		totalBallots += l
```
countvonzero: this doesn't make sense to me. number of eligibility is not equal to number of ballots.
in this case, it should always be numBallots <= numEligibities

miner/proposal_builder.go
```diff
@@ -299,8 +299,11 @@ func (pb *ProposalBuilder) createProposal(
 	if err := p.Initialize(); err != nil {
 		logger.Panic("proposal failed to initialize", log.Err(err))
 	}
-
-	logger.Event().Info("proposal created", p.Fields()...)
+	// NOTE(dshulyak) python tests read this log. alternative is to read number of eligibilities
+	// in the python tests, which seems like a bad idea.
+	for range p.EligibilityProofs {
+		logger.Event().Info("proposal created", log.Inline(p))
```
countvonzero: this is confusing for debugging when looking at logs.

i think leave the log as is, add a field to `Ballot.MarshalLogObject()` as `num_eligible`.
in python, capture `num_eligible` and add those up to match the number of `proposal created` it expected.

miner/proposal_builder.go
```diff
@@ -352,60 +355,50 @@ func (pb *ProposalBuilder) handleLayer(ctx context.Context, layerID types.LayerI
...
+		ballotID := p.Ballot.ID()
+		logger.With().Debug("storing ref ballot", epoch, ballotID)
+		if err := pb.storeRefBallot(epoch, ballotID); err != nil {
+			logger.With().Error("failed to store ref ballot", ballotID, log.Err(err))
+			return err
+		}
 	}
+
+	pb.eg.Go(func() error {
+		// generate a new requestID for the new block message
```
countvonzero: :s/new block/new proposal/

proposals/eligibility_validator.go
```diff
@@ -92,36 +93,48 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot)
...
+		}
+		vrfSig := proof.Sig
 
-	beaconStr := beacon.ShortString()
-	if !signing.VRFVerify(vrfPubkey, message, vrfSig) {
-		return false, fmt.Errorf("%w: beacon: %v, epoch: %v, counter: %v, vrfSig: %v",
-			errIncorrectVRFSig, beaconStr, epoch, counter, types.BytesToHash(vrfSig).ShortString())
-	}
+		beaconStr := beacon.ShortString()
+		if !signing.VRFVerify(vrfPubkey, message, vrfSig) {
```
countvonzero, line 120: i think it makes more sense to produce one vrf signature per Ballot. can make that `vrfMessage` contain `[]uint32` as Counters

proposals/eligibility_validator.go
```diff
@@ -92,36 +93,48 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot)
...
-		return false, fmt.Errorf("%w: proof counter (%d) numEligibleBallots (%d), totalWeight (%v)",
-			errIncorrectCounter, counter, numEligibleBallots, totalWeight)
-	}
+	var (
+		last    uint32
+		isFirst = true
+	)
+	for _, proof := range ballot.EligibilityProofs {
+		counter := proof.J
+		if counter >= numEligibleBallots {
```
countvonzero: rename the variable to `numEligibleSlots` since we only produce one ballot now.

proposals/eligibility_validator.go
```diff
@@ -92,36 +93,48 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot)
 		return false, err
 	}
 
-	counter := ballot.EligibilityProof.J
-	if counter >= numEligibleBallots {
-		return false, fmt.Errorf("%w: proof counter (%d) numEligibleBallots (%d), totalWeight (%v)",
-			errIncorrectCounter, counter, numEligibleBallots, totalWeight)
-	}
```
countvonzero, line 99: also check `len(ballot.EligibilityProofs) == numEligibleBallots`

tortoise/threshold.go
```diff
@@ -31,32 +32,35 @@ func computeBallotWeight(
...
-	if ballot.RefBallot == types.EmptyBallotID {
-		return weight{}, fmt.Errorf("empty ref ballot and no epoch data on ballot %s", ballot.ID())
-	}
-	rst, exist := weights[ballot.RefBallot]
-	if !exist {
-		refballot, err := bdp.GetBallot(ballot.RefBallot)
-		if err != nil {
-			return weight{}, fmt.Errorf("ref ballot %s for %s is unknown", ballot.ID(), ballot.RefBallot)
+		reference = weightFromUint64(targetWeight)
+		reference = reference.div(weightFromUint64(uint64(expected)))
```
countvonzero, line 40: don't understand the need for `referenceWeights`. 

this PR makes a miner produce one ballot per layer. this means this ballot takes up all the weight for this layer.
there is no need to call `GetNumEligibleSlots` anymore. can assign `weights[ballot.ID()] = targetWeight` directly.

countvonzero (commented):
while you are here, do you think you can also do something about https://github.com/spacemeshos/go-spacemesh/issues/2999 or leave it as separate issue?

i think there's still a missing check where we don't check two different ballots are submitted by the same miner with the same eligibility proofs

miner/proposal_builder.go
dshulyak: honestly i don't want to do anything about those tests, i will rewrite them, i understand that i can add num_eligible and parse it but this is a waste of time

cmd/node/app_test.go
dshulyak: not sure that i follow, previously we were creating proposal (and ballot) per eligibility

dshulyak:
> while you are here, do you think you can also do something about #2999 or leave it as separate issue?

it is far more complex than what is assumed, it is not as simple as a db check. i will submit it separately

tortoise/threshold.go
dshulyak: but weight of the ballot is not constant (e.g. it is not atx weight / layers). it depends on the number of eligibilities (previously number of ballots)

cmd/node/app_test.go
countvonzero: the code is
```
datamap[ap.nodeID.Key].eligibilities[lyr.Index()] += len(b.EligibilityProofs)
....
for id, l := range nodedata.eligibilities {
		totalBallots += l
}
```
here totalBallots is the number of all eligibilities. total number of ballots should be less than this if a ballot has multiple eligibility.

proposals/eligibility_validator.go
dshulyak: vrfsig is an input to vrf function

miner/proposal_builder.go
countvonzero: fair enough. how about this.
```
logger.Info("proposal created", log.Inline(p))
for range p.EligibilityProofs {
    logger.Event().Info("eligibility claimed by proposal", log.Inline(p))
}
```
and change the python to look for "eligibility claimed by proposal".

proposals/eligibility_validator.go
dshulyak: i don't think that there is any guarantee for that

proposals/eligibility_validator.go
countvonzero: i mean to change
```
type VotingEligibilityProof struct {
	// the counter value used to generate this eligibility proof. if the value of J is 3, this is the smesher's
	// eligibility proof of the 3rd ballot/proposal in the epoch.
	Counters []uint32
	// the VRF signature of some epoch specific data and J. one can derive a Ballot's layerID from this signature.
	Sig []byte
}

type vrfMessage struct {
	Beacon  types.Beacon
	Epoch   types.EpochID
	Counters []uint32
}

```
and this Sig is generated by doing `Sig := o.vrfSigner.Sign(vrfMessage)`
this means we don't need to have an array of proofs, just an array of counters inside the proof.

countvonzero: i.e we VRF verify once per Ballot, not once per eligibility

proposals/eligibility_validator.go
countvonzero: for honest parties? why not?

proposals/eligibility_validator.go
dshulyak: i don't think that your proposal works because we need to have vrfsig to decide if J is eligible in layer N before it can be appended to the list

cmd/node/app_test.go
dshulyak: ok, i will rename the variable. because we are testing number of eligibilities in this code

tortoise/threshold.go
countvonzero: before you did `targetWeight/numEligibleSlots` for that ballot. now that this ballot has `numEligibleSlots` combined in one, why isn't it `targetWeight` now for that ballot? or are you saying the previous code is incorrect?

proposals/eligibility_validator.go
dshulyak: i just think this is not the way it works. maybe i misunderstood things, but numEligibleBallots is a boundary for the counter, not the number of ballots  in the layer. e.g. this is a number of ballots in the epoch

miner/proposal_builder.go
dshulyak: done

proposals/eligibility_validator.go
countvonzero: > vrfsig is an input to vrf function

ah... now i understood what you mean by this. the vrf output per counter determine which layer it is.

proposals/eligibility_validator.go
countvonzero: i stand corrected. apologies.

tortoise/threshold.go
dshulyak: targetWeight is for the epoch. numEligibleSlots is for the epoch. by dividing them up we get weight for a single eligibility (single ballot). previously we had eligibility = ballot. therefore we could store it in the same map. now every ballot has a variable number of eligibilities, so same map doesn't work. at least this was my underestanding.

countvonzero: my confusion came from mixing up numSlots is for the whole epoch, not just a layer. what you said made sense.

countvonzero (approved):

nkryuchkov (approved):

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
those failures in add node test doesn't seem to be related to this change. new node fails to create a proposal with missing beacon error. i will open an issue for that

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
added https://github.com/spacemeshos/go-spacemesh/issues/3096
really strange that it reproduces so often in my branch, i don't see any changes in the code path that leads to this flake

dshulyak:
bors try

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/4979199501?check_suite_focus=true)

this failure looks reasonable tho. it's expecting twice the number of blocks in the test. looks like the test needs changing?

```
2022-01-28T10:17:42.5324472Z E           AssertionError: node 0184352753e33a413e9cb0d81c88ba48b3c71bdeea9203c7a219a0cbcfe8181c failed creating the avg block size
2022-01-28T10:17:42.5325811Z E           blocks created per layer 0.3333333333333333, wanted average block per node 0.6666666666666666, layer_avg_size 50, layers_per_epoch 3, num_miners 51, blocks_sum 1, to_layer 18, from_layer 15, blocks defaultdict(<class 'list'>, {16: [<Hit(fluent-bit-vzl6c-2022.01.28/_doc/DvwuoH4B6oHHSXvj2oe2): {'@timestamp': '2022-01-28T10:15:32.063Z', 'stream': 'stdout...}>], 15: [], 17: []})
```

dshulyak:
here is a link http://34.105.32.157:5601/goto/6f9daa3e1268dfdb6a3e55e28105bffc

dshulyak:
> this failure looks reasonable tho. it's expecting twice the number of blocks in the test. looks like the test needs changing?

why does it look reasonable? i am sure that the trace that i provided is accurate. node fails to produce a proposal because it is not synced in time

dshulyak:
bors try

countvonzero:
> > this failure looks reasonable tho. it's expecting twice the number of blocks in the test. looks like the test needs changing?
> 
> why does it look reasonable? i am sure that the trace that i provided is accurate. node fails to produce a proposal because it is not synced in time

my guess is that it *failed* to produce two proposals instead of one in the same layer, which is the intended effect of your PR.

dshulyak:
> my guess is that it failed to produce two proposals instead of one in the same layer, which is the intended effect of your PR.

there are no logs to prove that though. and in my pr i will send 2 logs so they should be counted correctly

countvonzero:
> blocks created per layer

i saw one for layer 17 (same epoch as layer 15). it makes sense to me.

http://34.105.32.157:5601/goto/ac66ab6f568ba84c93ca33075e8877d5

countvonzero:
> > blocks created per layer
> 
> i saw one for layer 17 (same epoch as layer 15). it makes sense to me.
> 
> http://34.105.32.157:5601/goto/ac66ab6f568ba84c93ca33075e8877d5

oh wait. there is only one log, meaning 1 eligibility.

dshulyak:
there are examples when 2 eligibilities were counted correctly http://34.105.32.157:5601/goto/01686c23f403e4e06c839c54125b0ea1

in layer 16 and in layer 8. so i think that works as expected

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

Issue #3072, opened 2022-01-21
beacon: reduce vote message size


countvonzero:
## Description

the first round of voting in the beacon protocol, miners need to list which proposals they are voting for. 
since all miners vote, reducing the size of this message could have significant impact on the total communication

## Actual Behavior

currently in the node uses the full VRF output (64 byte) to identify each vote.
the final beacon is Hash(sorted proposal strings)

## Expected Behavior

the first 10 byte (or less) should be enough.
the final beacon will be Hash(sorted and deduplicated proposal strings)

## Research Discussion
https://spacemesh.slack.com/archives/CG91F13C7/p1642514200091200

PR #3086, opened 2022-01-25
[Merged by Bors] - beacon: a little smaller vote bit vector

countvonzero:
## Motivation

Part of #3072

## Changes

- collapse valid and potentially valid votes for codec. we don't need this distinction anymore once we cast votes for the first voting round.

- use big.Int to encode bit vector instead of using extra lib bits-and-blooms/bitset
this also reduces the resolution of bit vector from 8 bytes to 1 byte.
  - before
it takes 16 bytes (2 uint64) to encode votes for 100 proposals and 4 bytes for length => 20 bytes total
it takes 8 bytes (1 uint64) to encode votes for 42 proposals and 4 bytes for length => 12 bytes total

  - after
this will encode the votes to as many bytes as needed. 
it takes 13 bytes (CEIL(100 / 8)) to encode votes for 100 proposals and 4 bytes for length => 17 bytes total
it takes 6 bytes (CEIL(42 / 8)) to encode votes for 42 proposals and 4 bytes for length => 10 bytes total

100 is the vote limit
42 is the expected number of votes

## Test Plan

unit test

## DevOps Notes


dshulyak (approved):
btw there will be additional 4 bytes for the length, so it will be 10 and 17. idk if that matters, but maybe it makes sense to always use 13 and make the field to be a fixed size.

countvonzero:
> btw there will be additional 4 bytes for the length, so it will be 10 and 17. idk if that matters, but maybe it makes sense to always use 13 and make the field to be a fixed size.

thanks for the reminder. update the PR desc.
interesting idea. tho since the expected number of proposal is 42, it's probably better to use 10 bytes on average instead of 13 bytes all the time....i think...

beacon/handlers_test.go
```diff
@@ -794,7 +791,7 @@ func Test_HandleSerializedFollowingVotingMessage_Success(t *testing.T) {
 	}, nil).Times(1)
 
 	// this msg will contain a bit vector that set bit 0 and 2
-	msg := createFollowingVote(t, signer, epoch, round, []uint64{0b101}, false)
+	msg := createFollowingVote(t, signer, epoch, round, []byte{5}, false)
```
nkryuchkov: Why would `5` be clearer than `0b101` here? IMO, `0b101` looks self-descriptive.

nkryuchkov (approved):
looks great

beacon/handlers_test.go
countvonzero: it's not. was trying something else and forgot to change it back. thanks

countvonzero:
bors merge

PR #3091, opened 2022-01-27
[Merged by Bors] - beacon: parameterize proposal size in the first vote msg

### LINKED ISSUE #2794: tortoise beacon: should not accept votes in future rounds
## Description
conversation from slack research channel:
kimmy
```for simplicity, current implementation accepts votes from all rounds regardless 
of what the current round is.
i.e. when we are at round N, we accept and count votes from round N-i or N+j
Q: apart from the obvious attack where round >= total_num_rounds, do you have 
concern that we accept votes from N+j prematurely?
```
Tal
```
This could allow a concentration attack. We rely on the fact that there's a large enough 
honest majority in every round. If the adversary can "borrow" votes from future rounds, 
this no longer holds.
```
kimmy
```
followup: how about votes that arrive x seconds before the round start? 
(x is the network latency)
```
Tal
```
You mean just one round "into the future"? That might not kill security, but it's not 
covered in the security analysis, so it's better to be on the safe side...
```

## Actual Behavior
accepting votes from all rounds during the tortoise beacon and count them as we go.

## Expected Behavior
- refuse votes where round > max round number
- only counts votes with the matching round number for that round

### LINKED ISSUE #3072: beacon: reduce vote message size

countvonzero:
## Motivation

Closes #3072 
Closes #2794 

## Changes

- reject proposals that arrive too late
- reject votes that arrive too early
- refactor epoch-specific data to state to separate protocol logic and epoch proposals/votes manipulation
- parameterize the size of proposal in the first vote message
- match beacon size to be the same as the proposal size to signal the same level of entropy

## Test Plan

unit tests / system tests

## DevOps Notes

### DIFF
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -85,7 +85,7 @@ func New(
 		db:                 db,
 		clock:              clock,
 		beacons:            make(map[types.EpochID]types.Beacon),
-		beaconsFromBallots: make(map[types.EpochID]map[string]*ballotWeight),
+		beaconsFromBallots: make(map[types.EpochID]map[types.Beacon]*ballotWeight),
 		current:            newState(conf),
 		next:               newState(conf),
 	}
@@ -130,7 +130,7 @@ type ProtocolDriver struct {
 	// beaconsFromBallots store beacons collected from ballots.
 	// the map key is the epoch when the ballot is published. the beacon value is calculated in the
 	// previous epoch and used in the current epoch.
-	beaconsFromBallots map[types.EpochID]map[string]*ballotWeight
+	beaconsFromBallots map[types.EpochID]map[types.Beacon]*ballotWeight
 
 	// states for the current epoch and the next epoch. we accept early proposals for the next epoch.
 	current, next *state
@@ -210,7 +210,7 @@ func (pd *ProtocolDriver) ReportBeaconFromBallot(epoch types.EpochID, bid types.
 		return
 	}
 
-	if eBeacon := pd.findMostWeightedBeaconForEpoch(epoch); !eBeacon.Equal(types.EmptyBeacon) {
+	if eBeacon := pd.findMostWeightedBeaconForEpoch(epoch); eBeacon != types.EmptyBeacon {
 		pd.setBeacon(epoch, eBeacon)
 	}
 }
@@ -220,11 +220,11 @@ func (pd *ProtocolDriver) recordBeacon(epochID types.EpochID, bid types.BallotID
 	defer pd.mu.Unlock()
 
 	if _, ok := pd.beaconsFromBallots[epochID]; !ok {
-		pd.beaconsFromBallots[epochID] = make(map[string]*ballotWeight)
+		pd.beaconsFromBallots[epochID] = make(map[types.Beacon]*ballotWeight)
 	}
-	entry, ok := pd.beaconsFromBallots[epochID][string(beacon)]
+	entry, ok := pd.beaconsFromBallots[epochID][beacon]
 	if !ok {
-		pd.beaconsFromBallots[epochID][string(beacon)] = &ballotWeight{
+		pd.beaconsFromBallots[epochID][beacon] = &ballotWeight{
 			weight:  weight,
 			ballots: map[types.BallotID]struct{}{bid: {}},
 		}
@@ -293,7 +293,7 @@ func (pd *ProtocolDriver) GetBeacon(targetEpoch types.EpochID) (types.Beacon, er
 	}
 
 	beacon := pd.getBeacon(targetEpoch)
-	if !beacon.Equal(types.EmptyBeacon) {
+	if beacon != types.EmptyBeacon {
 		return beacon, nil
 	}
 
@@ -344,7 +344,7 @@ func (pd *ProtocolDriver) getPersistedBeacon(epoch types.EpochID) (types.Beacon,
 		return types.EmptyBeacon, fmt.Errorf("get from DB: %w", err)
 	}
 
-	return data, nil
+	return types.BytesToBeacon(data), nil
 }
 
 func (pd *ProtocolDriver) setBeginProtocol(ctx context.Context) {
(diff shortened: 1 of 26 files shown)


tortoise/tortoise_test.go
```diff
@@ -1480,7 +1480,7 @@ func ensureBaseAndExceptionsFromLayer(tb testing.TB, lid types.LayerID, votes *t
 }
 
 func TestBaseBallotEvictedBlock(t *testing.T) {
-	const size = 10
+	const size = 12
```
countvonzero, line 1483: i don't know why test fails. changing this to 12 stabilized the test. maybe the random-ness is altered again.

cmd/root.go
```diff
@@ -183,6 +183,8 @@ func AddCommands(cmd *cobra.Command) {
 		config.Beacon.WeakCoinRoundDuration, "Weak coin round duration in milliseconds")
 	cmd.PersistentFlags().Var((*types.RatVar)(config.Beacon.Theta), "beacon-theta",
 		"Ratio of votes for reaching consensus")
+	cmd.PersistentFlags().IntVar(&config.Beacon.ProposalNumBytes, "beacon-proposal-num-bytes",
```
dshulyak: in what case we want to change this? can we define size as a constant?

countvonzero: this is per request by Iddo in slack https://spacemesh.slack.com/archives/CG91F13C7/p1642773410106600?thread_ts=1642514200.091200&cid=CG91F13C7

it is possible that we have so many smeshers that we change it to 2 bytes?

countvonzero: but maybe it's better to name it `beacon-num-bytes` instead

dshulyak: i don't think that he meant it to be a parameter for a node binary, that wouldn't make any sense

dshulyak: i think if we don't intend to change something after binary is compiled it shouldn't be configurable. what matters from research pow is that this can be changed in general, not for each network. we already have a bunch of useless parameters

countvonzero: music to my ears. happy to oblige. changed it to using a constant. thanks

countvonzero:
bors try

tortoise/tortoise_test.go
dshulyak: i will address random weights in one of the open prs

beacon/beacon.go
```diff
@@ -813,11 +770,19 @@ func (pd *ProtocolDriver) sendFirstRoundVote(ctx context.Context, epoch types.Ep
...
-	bitVector := encodeVotes(ownCurrentRoundVotes, pd.firstRoundIncomingVotes[string(pd.edSigner.PublicKey().Bytes())])
-	pd.mu.RUnlock()
+	defer pd.mu.RUnlock()
+	return pd.current.getMinerFirstRoundVote(minerPK)
+}
+
+func (pd *ProtocolDriver) sendFollowingVote(ctx context.Context, epoch types.EpochID, round types.RoundID, ownCurrentRoundVotes allVotes) error {
+	firstRoundVotes, err := pd.getFirstRoundVote(pd.edSigner.PublicKey())
+	if err != nil {
+		return fmt.Errorf("failed to get own first round votes %v: %w", pd.edSigner.PublicKey().String(), err)
```
nkryuchkov: Maybe just
```suggestion
		return fmt.Errorf("get own first round votes %v: %w", pd.edSigner.PublicKey().String(), err)
```
?

With the current text, the outer code 
```go
if err := pd.sendFollowingVote(ctx, epoch, round, votes); err != nil {
	rLogger.With().Error("failed to send following vote", log.Err(err))
}
```
would output `failed to send following vote errmsg=failed to get...` which is a bit redundant

beacon/handlers.go
```diff
@@ -185,7 +197,11 @@ func (pd *ProtocolDriver) addPotentiallyValidProposal(proposal []byte) {
 
 	pd.mu.Lock()
 	defer pd.mu.Unlock()
-	pd.incomingProposals.potentiallyValid = append(pd.incomingProposals.potentiallyValid, proposal)
+	if pd.current.incomingProposals.valid == nil {
```
nkryuchkov: Did you mean `potentiallyValid`?
```suggestion
	if pd.current.incomingProposals.potentiallyValid == nil {
```

beacon/handlers.go
```diff
@@ -501,54 +513,45 @@ func (pd *ProtocolDriver) currentEpoch() types.EpochID {
...
+
+func (pd *ProtocolDriver) checkProposalEligibility(logger log.Log, vrfSig []byte) bool {
+	pd.mu.RLock()
+	defer pd.mu.RUnlock()
+	eligible := pd.current.proposalChecker.IsProposalEligible(vrfSig)
+	if !eligible {
+		// the peer may have different total weight from us so that it passes threshold for the peer
+		// but does not pass here
+		proposalShortString := types.BytesToHash(vrfSig).ShortString()
+		logger.With().Warning("rejected proposal that doesn't pass threshold",
```
nkryuchkov: This method doesn't reject such a proposal. What do you think about moving this log where the rejection happens?

beacon/proposal_list_test.go
```diff
@@ -20,13 +20,13 @@ func Test_proposalList_sort(t *testing.T) {
 	}{
 		{
 			name:     "Sorted order remains not changed",
-			hashes:   proposalList{"0x1", "0x2", "0x3"},
-			expected: proposalList{"0x1", "0x2", "0x3"},
+			hashes:   proposalList{[]byte("0x1"), []byte("0x2"), []byte("0x3")},
```
nkryuchkov: Perhaps `0x` prefixes could be removed? They aren't needed now

beacon/beacon.go
countvonzero: done

beacon/handlers.go
countvonzero: oops. changed and added test

beacon/handlers.go
countvonzero: since we want to log weight here, i didn't want to take out read lock on the callsite. changed log wording.

beacon/proposal_list_test.go
countvonzero: done

countvonzero:
bors try

beacon/handlers.go
countvonzero: https://github.com/spacemeshos/go-spacemesh/actions/runs/1756359387

i am surprised all the system tests passes. but then every node is run in a controlled environment.

nkryuchkov (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #3093, opened 2022-01-27
[Merged by Bors] - ballots, tortoise: track malicious ballots and ignore them in tortoise

### LINKED ISSUE #2999: Missing check for duplicate use of voting eligibility

dshulyak:
this is basic security validation. ballots with duplicate eligibility must be ignored.
we still store all of them in db and gossip to the network, so it leaves network vulnerable to trivial ddos.

Closes https://github.com/spacemeshos/go-spacemesh/issues/2999

## Changes
- extend ballot type to track if it is malicious
- set ballot to malicious if identity is malicious
- identity is malicious if there are at least 2 ballots from it in the same layer
- tortoise doesn't count malicious ballots
- tortoise doesn't use malicious ballots as a base ballot


common/types/ballot.go
```diff
@@ -35,6 +35,8 @@ type Ballot struct {
 	ballotID BallotID
 	// the public key of the smesher used
 	smesherID *signing.PublicKey
+	// malicious is set to true if smesher that produced this ballot is known to be malicious.
```
countvonzero, line 38: will we ever change our mind about this field?

mesh/meshdb.go
```diff
@@ -27,6 +29,15 @@ const (
 	layerSize = 200
 )
 
+var (
+	// ErrMalicious is a common error type.
+	ErrMalicious = errors.New("mesh: malicious")
+	// ErrMaliciousBallot is raised if ballot is malicious.
+	ErrMaliciousBallot = fmt.Errorf("%w: ballot", ErrMalicious)
+	// ErrMaliciousSmesher is raised if smesher produced more than 2 ballots in the same layer.
+	ErrMaliciousSmesher = fmt.Errorf("%w: smesher", ErrMalicious)
```
countvonzero: are these errors used?

tortoise/full.go
```diff
@@ -82,6 +82,9 @@ func (f *full) countVotesFromBallots(logger log.Log, ballotlid types.LayerID, ba
 			continue
 		}
 		ballotWeight := f.ballotWeight[ballot]
+		if ballotWeight.isNil() {
```
countvonzero, line 85: what changed in this PR that cause this to be true?

mesh/meshdb.go
```diff
@@ -137,7 +148,46 @@ func (m *DB) Transactions() database.Getter {
...
-	return m.addBallot(b)
+	mal, err := identities.IsMalicious(m.db, b.SmesherID().Bytes())
+	if err != nil {
+		return err
+	}
+	if mal {
+		b.SetMalicious()
+	}
+
+	// it is important to run add ballot and set identity to malicious atomically
```
countvonzero, line 150: why? it seems whether a ballot is malicious is entirely dependent on whether a identity is malicious at that time.
it seems safe to do `identities.SetMalicious` first and only save the ballot when that succeeds.

rather `ballots.CountByPubkeyLayer()` and `ballots.Add()` should be in a DBTX instead.

mesh/meshdb.go
```diff
@@ -137,7 +148,46 @@ func (m *DB) Transactions() database.Getter {
...
+	if err := ballots.Add(tx, b); err != nil && !errors.Is(err, sql.ErrObjectExists) {
+		return err
+	}
+	if !mal {
+		count, err := ballots.CountByPubkeyLayer(tx, b.LayerIndex, b.SmesherID().Bytes())
+		if err != nil {
+			return err
+		}
+		if count > 1 {
+			if err := identities.SetMalicious(tx, b.SmesherID().Bytes()); err != nil {
```
countvonzero, line 166: it feels slightly out of place to set an identity malicious at the DB layer.
thinking forward, how should this information gets propagated up to proceed with blacklisting.

maybe it make more sense to add a malicious field in the ballot table? we can mark that this ballot is determined to use eligibility more than once and propagate that error up. let the upstream caller to decide whether to mark an identity malicious and further blacklist? (and perhaps retroactively mark other ballots of the same identity malicious)

mesh/meshdb.go
dshulyak: ~~what is a DBTX?~~


> why? it seems whether a ballot is malicious is entirely dependent on whether a identity is malicious at that time.
it seems safe to do identities.SetMalicious first and only save the ballot when that succeeds.

ballot is saved regardless of identity status.

if SetMalicious succeeds and ballots.Add fails then identity will be marked as malicious but we won't have two ballots from that identity in the database. 

> it seems safe to do identities.SetMalicious first and only save the ballot when that succeeds.

but you can do that only if there are more than 1 ballot in the same layer. how it can be done before ballots.Add/Count ?

common/types/ballot.go
dshulyak: i don't know. what do you mean?

mesh/meshdb.go
dshulyak: thanks, forgot to remove

mesh/meshdb.go
dshulyak: > thinking forward, how should this information gets propagated up to proceed with blacklisting.

it is blacklisted by tracking that identity is malicious. what exactly is meant by blacklisting?

> maybe it make more sense to add a malicious field in the ballot table?

it is hard for me to tell if it works or not. there is an important concurrency issue that i addressed by executing statements in transaction. we need to guarantee that if there are 2 ballots in the datatabse then identity is set to malicious. 

for example if Count statement is executed not within transaction then when we are adding two distinct ballots they both may return 0 and identity won't be malicious.

other case, if SetMalicious is not within transaction then ballots.Add may succeed and SetMalicious fail - identity again not tracked as malicious.

dshulyak: > maybe it make more sense to add a malicious field in the ballot table?

recalled another reason. if identity was malicious at some layer in the past we don't want to use new ballots from it. scanning whole ballots table to decide if identity is malicious is not an option

tortoise/full.go
dshulyak: weight is nil if ballot is malicious

mesh/meshdb.go
dshulyak: > it feels slightly out of place to set an identity malicious at the DB layer.

also I think that we don't need MeshDB at all now, low level db stuff is moved to sql module.
and higher level composition should be moved to the Mesh

common/types/ballot.go
countvonzero: just asking if you can think of any cases where we will set an identity's malicious field from true to false.

actually, do we really consider an identity to be malicious if it has multiple proposals in the same layer?
say they customize the binary such that they still make one proposal per eligibility. technically that's not wrong.

mesh/meshdb.go
countvonzero: > what is a DBTX?

DB transaction

> if SetMalicious succeeds and ballots.Add fails then identity will be marked as malicious but we won't have two ballots from that identity in the database.

i think we need to decouple this in general. when we set an identity to be malicious, we need to have proof that it is in the database as well. for example. if we figure action A (double-vote in beacon protocol) and B (double ballots for same eligibility) are both malicious, how can you tell, from the identity table, which action it is?

i think we need to have enum to define the maleficence (Tal's word) type and use bytes as proof, and save both in the DB. separate that logic from saving ballots

mesh/meshdb.go
countvonzero: > it is blacklisted by tracking that identity is malicious. what exactly is meant by blacklisting?

from one of the R&D recording, Tal said we need to gossip maleficence proof on a bad actor. i'd imagine once an identity is blacklisted, we don't consider data from them anymore. but i don't think there is an official definition.

> for example if Count statement is executed not within transaction then when we are adding two distinct ballots they both may return 0 and identity won't be malicious.

right. that's why i said earlier that Count and Add needs to be in a DBTX (db transaction)

> recalled another reason. if identity was malicious at some layer in the past we don't want to use new ballots from it. scanning whole ballots table to decide if identity is malicious is not an option

i am not saying to remove malicious field from identities, i am saying to add another malicious field to ballots. that way we know at the time of add this ballot is determined malicious.

as your implementation now, whether a ballot is malicious actually changes. for example, ballot A is added, then an identity is determined malicious when ballot B is added, both ballot A and ballot B are queried as malicious later. this is confusing for debugging, especially if ballot A is chosen as base ballot, but is actually malicious.

tortoise/full.go
countvonzero: this heavily depend on timing. it's possible a ballot isn't considered malicious the first time tortoise sees it, but then is considered malicious the second time tortoise sees it.

dshulyak: ofcourse it is dependent on timing, you don't know if ballot is malicious until you received second with same eligibilities, i don't see how it changes anything

common/types/ballot.go
dshulyak: > actually, do we really consider an identity to be malicious if it has multiple proposals in the same layer?

 there is no reason for honest node to produce multiple. and enforcing one simplifies validation

mesh/meshdb.go
dshulyak: i don't think that this is important for this change

mesh/meshdb.go
dshulyak: > i am not saying to remove malicious field from identities, i am saying to add another malicious field to ballots. that way we know at the time of add this ballot is determined malicious.

ballot can be determined as malicious only after we saw two ballots with the same eligibility. once we know that identity is malicious - all ballots from this identity are malicious. it should be expected that some ballots, due to the concurrency or malicious intent, can be used as base ballot before they are discovered as malicious. we can't remove malicious ballots from state but we also don't count weight from them...

mesh/meshdb.go
countvonzero: > i don't think that this is important for this change

why not? i feel like we are rushing into this implementation without knowing the bigger implication of a malicious identity. 
i mean why is knowing a ballot malicious more important than other malicious behavior like referencing a non-existent block ID?

i am just not sure why we are doing this change now without the bigger picture of malicious behavior and blacklisting.

countvonzero:
i think i have more questions than answers. 

the big thing from me is that without the higher level picture, it's difficult for me to agree that mesh DB is the right place to mark an identity is malicious. can we have a more complete discussion with research before doing this change?

mesh/meshdb.go
dshulyak: because without this verification the protocol doesn't work

dshulyak:
> the big thing from me is that without the higher level picture, it's difficult for me to agree that mesh DB is the right place to mark an identity is malicious. can we have a more complete discussion with research before doing this change?

mesh db is an obsolete component. i don't think that it makes sense to postpone this change. so if someone will unblock it i will merge. we can have discussion afterwards

mesh/meshdb.go
countvonzero: does this mean that tortoise rerun may keep changing its opinions when we continue to discover malicious ballots?

dshulyak: yes

countvonzero:
> mesh db is an obsolete component. i don't think that it makes sense to postpone this change. so if someone will unblock it i will merge. we can have discussion afterwards

the logic is encapsulated inside a database transaction. i don't think it's obsolete. i just think it's the wrong component to handle it (too low level).

dshulyak:
> the logic is encapsulated inside a database transaction. i don't think it's obsolete. i just think it's the wrong component to handle it (too low level).

meshdb became a simple proxy to sql functions. pretty clear to me that it is obsolete. my intention for transactions was exactly cases like this, i don't understand how is that low level...

from the protocol point of view, it is important not to count ballots with the same eligibilities (https://github.com/spacemeshos/go-spacemesh/issues/2999#issuecomment-993219954). using transactions allows implementing that in a safe and fast way. maybe there are other approaches, but saying that using transactions is not ok doesn't make much sense to me

countvonzero:
> > the logic is encapsulated inside a database transaction. i don't think it's obsolete. i just think it's the wrong component to handle it (too low level).
> 
> meshdb became a simple proxy to sql functions. pretty clear to me that it is obsolete. my intention for transactions was exactly cases like this, i don't understand how is that low level...

fine. remove meshdb. my point is that the logic of marking an identity malicious is encapsulated inside a sql db transaction and is too low level for me.

my imagination is that there is a component in spacemesh, say `blacklister` that takes care of this logic for determining malicious behavior and saving proof. specifically for this PR, i'd prefer

- gossip handler provides notify blacklister of this suspicious behavior
- blasklister gather evidence and mark an identity as malicious and save the evidence, gossip the maleficence proof
- other components, tortoise/hare/beacon, then can check whether an identity is malicious and act accordingly in each round

of course the blacklister component can mostly be left empty for now.

dshulyak:
> my imagination is that there is a component in spacemesh, say blacklister that takes care of this logic for determining malicious behavior and saving proof. specifically for this PR, i'd prefer

atomicity of ballot/identity status write is important for crash safety. what you described is not crash-safe. also, i disagree with coordination that e.g. tortoise should go to that component.. tortoise should act just based on stored or passed data. 

and in general i don't see why such component is necessary. every sub-protocol will have its own rules for malicious behavior, and they can operate on data using sql functions. why would i want to create a component that will handle that instead of handling each malicious behavior separately?

just for reference, even without sql db the write needs to be atomic and would have to be made with write batch. so moving logic to some component is not an option in my opinion. maybe there is some complex way to make it robust, but i don't see it.

countvonzero:
> and in general i don't see why such component is necessary. every sub-protocol will have its own rules for malicious behavior, and they can operate on data using sql functions. why would i want to create a component that will handle that instead of handling each malicious behavior separately?

i disagree each component handle its own malicious behavior separately. are you saying that we will have different `identities` table for it to remember its own malicious public keys? and when asked to provide proofs, we exhaustively ask each component to provide? i find that design tedious.

> just for reference, even without sql db the write needs to be atomic and would have to be made with write batch. so moving logic to some component is not an option in my opinion. maybe there is some complex way to make it robust, but i don't see it.

the writes can still be atomic, if you pass the database transaction object btwn components. each component will have DB/sql logic anyway. this doesn't seem too far-fetched.

dshulyak:
> i disagree each component handle its own malicious behavior separately. are you saying that we will have different identities table for it to remember its own malicious public keys?

they will use the same identities table. 

>  and when asked to provide proofs, we exhaustively ask each component to provide?

the proof is very vague definition. i don't think that there will be a generic proof for anything. for instance in this case the proof is any two ballots, there is no need for additional gossip message type or anything.

actually, after some thought, the proof will be always some existing type that preserves the original signature - so there will be no need to generalize it.

> the writes can still be atomic, if you pass the database transaction object btwn components. each component will have DB/sql logic anyway. this doesn't seem too far-fetched.

this won't work. if sqlite transaction is in effect it will lock the database, transactions must be short living and it will be impossible to enforce within multiple modules.

countvonzero:
let's imagine now we need to handle malicious ballot (A) and malicious node that gossip about no-existing data (B).
how do you plan to differentiate the two?

> they will use the same identities table.

how do i track down why an identity is marked malicious? 

> actually, after some thought, the proof will be always some existing type that preserves the original signature - so there will be no need to generalize it.

right. but same question applies. how do we know why an identity is marked malicious?

> this won't work. if sqlite transaction is in effect it will lock the database, transactions must be short living and it will be impossible to enforce within multiple modules.

i don't understand the module boundary. will it work with multiple methods in the same file? how about a callback provided by the blacklsiter?

in any case, i don't intend to argue about implementation. i am more interested in how to investigate why an identity is marked malicious and how to provide proof from a high level design.

dshulyak:
> let's imagine now we need to handle malicious ballot (A) and malicious node that gossip about no-existing data (B).
how do you plan to differentiate the two?

but those two are completely separate use cases, identities for them are not shared at all. one is p2p identity and another smesher. they are should not be mixed anywhere

> how do i track down why an identity is marked malicious?

when do you think it will be needed? for auditing? we can just attach additional info when identity is marked malicious, but i don't see why it is needed. also for auditing you can use logs

> i don't understand the module boundary. will it work with multiple methods in the same file? how about a callback provided by the blacklsiter?

it should be obvious that when transactions are initiated (the first operation occurred) there should be no mutexes or io in the code path, otherwise it will be too error-prone and lead to multiple unexpected failures.

>  i am more interested in how to investigate why an identity is marked malicious and how to provide proof from a high level design.

what is a proof from high level design? there is no generic proof, proof is something specific to the particular protocol. as i said in this case proof is at least two ballots. you don't need to encapsulate it in any high level and provide separately to any other node, any node can establish that identity is malicious by downloading these two ballots. 

let's say another "proof" is voting differently on the same block within hdist, again no need to encapsulate it in any special object. i can't speak in broader terms because this is not something that will be covered for mainnet. because proofs of malicious behavior are not critical for safety. but it is very likely that proofs of malicious behavior will not require any additional objects or high-level design, as you need to preserve original object with the signature that can prove malicious behavior.

dshulyak:
had some time to think about other cases:
1. in hare or beacon if the same identity votes differently for a proposal it can be considered as malicious
2. if identity outputs more than one reference ballot and changes beacon

in 1st case we will need to have a place to store those proofs, and the proofs will probably given to a peer. in cases where ballots are itself is a proof we don't need to change anything in sync or gossip.

nkryuchkov (approved):

countvonzero:
note that i am not objecting to this PR as the first step, i am requesting that the devs collectively have a plan regarding how the code should deal with malicious behavior before any specific fixes are made. 

my reading from your response is that you want to do a localize thing that node itself knows about this malicious smesher but not share. my preference is to have a more complete plan on how to deal with malicious behavior in general before we add code for it. i don't see why this is more urgent than, say, preventing DDOS by sending us malformed ballots so don't understand why we should rush this change in without a general discussion.

>> let's imagine now we need to handle malicious ballot (A) and malicious node that gossip about no-existing data (B).
how do you plan to differentiate the two?

> but those two are completely separate use cases, identities for them are not shared at all. one is p2p identity and another smesher. they are should not be mixed anywhere

to me they are related. whoever p2p identity gossiped this ballot to us, that p2p identity is just as malicious as the smesher that produced this ballot. maybe less likely since there is an element of timing involved. in the example you gave on voting multiple blocks within hdist, the p2p identity and the smesher that produced the ballot are both malicious.

> let's say another "proof" is voting differently on the same block within hdist, again no need to encapsulate it in any special object. 

right. but there is need to differentiate different kinds of malicious behavior. my thoughts are we need to define various types of bad acts and define the proof for each of them. in this PR, bad act will be reusing the same eligibility. the proof will be the two ballots IDs. in the case of "voting on multiple blocks within hdist", the proof will be the ballot ID that captures this bad act. the outcome of this could be whether we want to ban the p2p identity, the smesher, or both.

> i can't speak in broader terms because this is not something that will be covered for mainnet. 

i am not aware that we are NOT covering malicious behavior before mainnet.

a note on the atomicity. we don't need DB transaction to achieve this in this case.
let's say we do have component `blacklister`. when gossip handler realized that a ballot is suspicious, it calls a method on `blacklister` to scrutinize this ballot. `blacklister` then generates proofs and saves it to the DB atomically, then handler the proof to the gossip handler so it can act accordingly.

if you insist to merge this PR, i don't want to stand in the way either. the code itself LGTM.

dshulyak:
> my reading from your response is that you want to do a localize thing that node itself knows about this malicious smesher but not share. 

i don't understand how it follows. the node will share ballots that it knows, if it knows about 2 ballots with the same eligibilities - the identity will be decided as malicious just by following the encoded protocol. 

> my preference is to have a more complete plan on how to deal with malicious behavior in general before we add code for it. i don't see why this is more urgent than, say, preventing DDOS by sending us malformed ballots so don't understand why we should rush this change in without a general discussion.

this is urgent in the sense that it is one of the few validations that we are missing for mainnet. because this is a basic security assumption, the other one is that the same nipost can't be reused for multiple atxs. if they are not implemented whole protocol is meaningless. we still have to deal with other things, by properly invalidating double voting in hare or beacon, but there is no requirement to implement any proofs for them.

> DDOS by sending us malformed ballots

this is not proof of malicious behavior. we can blacklist p2p identity that shares irrelevant stuff, but it is not cryptographically verifiable, and nobody should believe us if we send a message saying that this p2p identity is malicious.

> in this PR, bad act will be reusing the same eligibility. the proof will be the two ballots IDs.

but this is not the proof. the proof is the ballots themself. unless you mean something different by the proof. we can't just gossip ids and expect everyone to believe.

> to me they are related. whoever p2p identity gossiped this ballot to us, that p2p identity is just as malicious as the smesher that produced this ballot. maybe less likely since there is an element of timing involved. in the example you gave on voting multiple blocks within hdist, the p2p identity and the smesher that produced the ballot are both malicious.

there is a significant difference. we can't proof that p2p identity is malicious to any other node in the network. the proof needs to be cryptographically verifiable. i don't think that it will *always* work the way you said. for instance some malicious ballots will have to be stored and propagated into the network. as you mentioned because of the timing/concurrency node can't just mark p2p identity malicious if it sends us a missing piece of information (like 2nd ballot with same eligibility). 

in case if p2p identity sends syntactically invalid messages then we can determine with certainty that p2p identity is malicious, but we won't store any proofs.

> let's say we do have component blacklister. when gossip handler realized that a ballot is suspicious, it calls a method on blacklister to scrutinize this ballot. blacklister then generates proofs and saves it to the DB atomically, then handler the proof to the gossip handler so it can act accordingly.

but if there are two concurrent ballots with the same eligibilities how exactly blacklister will do its job. i am not sure that i understand how it will work and why it will be safe. you definitely can do it without transactions, it just becomes much more complex problem.

also i don't understand what is meant by generating proof means. in this case there is nothing to generate 

---

i don't understand why we should wait for a discussion with a research, what i said is not something new, and in my opinion doesn't require research involvement at this stage at all. things like separation of p2p malicious behavior and protocol malicious behavior were also discussed some time ago and solutions for them are different. protocol proofs will be implemented when research will describe them. but most of them are not mandatory for safety.

dshulyak:
bors merge

countvonzero:
> but this is not the proof. the proof is the ballots themself. unless you mean something different by the proof. we can't just gossip ids and expect everyone to believe.

sure. i mean to store the IDs in local db, but when asked, compiled the ballots body as proof.

> also i don't understand what is meant by generating proof means. in this case there is nothing to generate

by generating i mean creating the struct that will represent proof of a given bad act and store it locally in db. i don't mean to generate something cryptographically. my hope is to define the bad acts we will deal with instead of leaving all this discussion to the next person tasked to this, since i think we still need to do this before mainnet.

> i don't understand why we should wait for a discussion with a research, what i said is not something new, and in my opinion doesn't require research involvement at this stage at all.

i didn't mention research. discussing with devs teams for me means involving Noam and Lane in the design and implementation.

dshulyak:
> i think we still need to do this before mainnet.

i don't understand why though. proofs, such as double voting in hare or beacon, are not critical for safety or correctness. it is just important to recognize such behavior during the execution of the protocol and prevent it from affecting result

countvonzero:
> > i think we still need to do this before mainnet.
> 
> i don't understand why though. proofs, such as double voting in hare or beacon, are not critical for safety or correctness. it is just important to recognize such behavior during the execution of the protocol and prevent it from affecting result

i don't specifically mean proof. just like you said, to recognize such behavior, handle it cohesively, and record it for debugging purposes.

PR #3095, opened 2022-01-28
[Merged by Bors] - tortoise: revert to the last verified layer

### LINKED ISSUE #3060: tortoise, mesh: handle the case when verified layer didn't reach historically verified layer during rerun

### LINKED ISSUE #3083: logging: creating a logger.WithFields is not lazily initialized and will lead to significant allocations in the hot loop
In particular, https://github.com/spacemeshos/go-spacemesh/blob/develop/tortoise/verifying.go#L203-L210 leads to 60Gb of memory allocated on the heap in an hour. The fix is simple, fields should be duplicated for every logline.

Adding this issue as a reminder to be wary of this pattern.

### LINKED ISSUE #3097: mesh: data race
## Description

I was running `make test` with enabled `-race` flag locally and got this data race inside `TestAppTestSuite`:
```
==================
WARNING: DATA RACE
Write at 0x00c0001be4f4 by goroutine 510:
  github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).setLatestLayerInState()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/mesh/mesh.go:630 +0x25c
  github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).updateStateWithLayer()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/mesh/mesh.go:614 +0x284
  github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).pushLayer()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/mesh/mesh.go:477 +0x314
  github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).pushLayersToState()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/mesh/mesh.go:444 +0x4b0
  github.com/spacemeshos/go-spacemesh/mesh.(*validator).ProcessLayer()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/mesh/mesh.go:327 +0x554
  github.com/spacemeshos/go-spacemesh/mesh.(*Mesh).ProcessLayerPerHareOutput()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/mesh/mesh.go:595 +0x36c
  github.com/spacemeshos/go-spacemesh/turbohare.(*SuperHare).Start.func1.1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/turbohare/hare.go:102 +0x7a8

Previous read at 0x00c0001be4f4 by goroutine 291:
  [failed to restore the stack]

Goroutine 510 (running) created at:
  github.com/spacemeshos/go-spacemesh/turbohare.(*SuperHare).Start.func1()
      /Users/nkryuchkov/GolandProjects/go-spacemesh/turbohare/hare.go:59 +0xb0
==================
--- FAIL: TestAppTestSuite (772.45s)
    --- FAIL: TestAppTestSuite/TestMultipleNodes (772.44s)

```

## Actual Behavior

Data race

## Expected Behavior

No data race

## Environment
 - OS: 
 
 darwin/arm64
 - Node Version: 
 
 `2828c18eef1b30e56a2e4b0fcee4e10a50d3afb9`
 - Smapp version (if applicable): 
 
 n/a

dshulyak:
if during rerun tortoise can't reach the previously verified layer we have no other option but to revert to the newly verified layer.

the mesh component may optimistically wait additional time for the tortoise to reach the same layer so that it doesn't have to rewind state prematurely. but that requires additional handling in the mesh, so i will skip that for now.

closes https://github.com/spacemeshos/go-spacemesh/issues/3060 https://github.com/spacemeshos/go-spacemesh/issues/3083 #3097 

#### Changes
- serialize layers processing in mesh to prevent concurrency issues if sync and hare call ProcessLayer at the same time with different layers
- allow reading layer fields (latest, latest in state, processed, missing) without locking, as they are used in multiple components and in API
- if the tortoise fails to verify up to the historically verified layer during rerun it reverts to the last verified layer
- logging improvements in tortoise


mesh/mesh.go
```diff
@@ -258,77 +232,75 @@ func (msh *Mesh) setProcessedLayer(layerID types.LayerID) {
...
-}
-
 // ProcessLayer performs fairly heavy lifting: it triggers tortoise to process the full contents of the layer (i.e.,
 // all of its blocks), then to attempt to validate all unvalidated layers up to this layer. It also applies state for
 // newly-validated layers.
-func (vl *validator) ProcessLayer(ctx context.Context, layerID types.LayerID) error {
-	logger := vl.WithContext(ctx).WithFields(layerID)
+func (msh *Mesh) ProcessLayer(ctx context.Context, layerID types.LayerID) error {
+	msh.mu.Lock()
+	defer msh.mu.Unlock()
```
countvonzero, line 268: it looks like you only need to acquire this lock inside `msh.setProcessedLayer()`?
or i am missing some logic here.

mesh/mesh.go
```diff
@@ -172,31 +152,30 @@ func (msh *Mesh) LatestLayer() types.LayerID {
...
 	})
-	defer msh.mutex.Unlock()
-	msh.mutex.Lock()
-	if idx.After(msh.latestLayer) {
-		events.ReportNodeStatusUpdate()
-		msh.With().Info("set latest known layer", idx)
-		msh.latestLayer = idx
-		if err := layers.SetStatus(msh.db, idx, layers.Latest); err != nil {
-			msh.Error("could not persist latest layer index")
+	for {
```
countvonzero, line 168: what is the purpose of the for block?

countvonzero (approved):

mesh/mesh.go
countvonzero: yikes. i don't think i understand this. how does the code exit the for loop?

dshulyak: if there are two concurrent updates with layer 9 and layer 10, it is important not execute set=10 and then set=9.  in order to guarantee it i am using CAS on the previous value. for example two threads:

- T1. lid = 9
- T2. lid = 10
- T1. load. current = 8
- T2. load. current = 8
- T2. update. CAS 8 10. SUCCESS => exit
- T1. update CAS 8 9. FAILED
- T1. load. current = 10
- T1.  lid <= current => exit
-

mesh/mesh.go
dshulyak: the issue that i want to prevent is concurrent calls by sync and hare, with a different layers. whole method needs to be locked otherwise we can read one value for msh.LatestLayerInState() on L295 and the other one on L572. 

also previously we had a lock in updateStateWithLayer, so i assume we need it there as well, i just moved it up so that it is easier to reason about ProcessLayer and prevent unexpected concurrency issues

mesh/mesh.go
countvonzero: i see. thanks for the explanation. 
tho is this extra logic still necessary given that you always hold a write lock when executing this?

mesh/mesh.go
```diff
@@ -3,12 +3,13 @@
 package mesh
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"sync"
 
+	"go.uber.org/atomic"
```
nkryuchkov, line 11: Why not `sync/atomic`? It has `atomic.Value` too

nkryuchkov (approved):

mesh/mesh.go
dshulyak: uber/atomic is just a wrapper around sync/atomic , the motivation is pretty much summarized in https://github.com/uber-go/atomic#usage

mesh/mesh.go
dshulyak: setLatestLayer is called when block or proposal are added

dshulyak:
bors merge

PR #3104, opened 2022-01-30
[Merged by Bors] - presets: network with 15s layer for cloud or local environments

dshulyak:
preset with short layer time for fast tests, usable in environments where network is mostly stable and has very low latency. locally it is obviously has low latency, and in the same zone of the cloud.

## Changes
- add fastnet preset with short layer time
- add option for tortoise rerun, as it can be used to speed up tests that require recovery from partition


cmd/root.go
```diff
@@ -188,7 +186,11 @@ func AddCommands(cmd *cobra.Command) {
 	cmd.PersistentFlags().Uint32Var(&config.Beacon.BeaconSyncNumBallots, "beacon-sync-num-blocks",
 		config.Beacon.BeaconSyncNumBallots, "Numbers of blocks to wait before determining beacon values from them.")
 
-	/**======================== Post Flags ========================== **/
+	/**======================== Tortoise Flags ========================== **/
+	cmd.PersistentFlags().Uint32Var(&config.Tortoise.Hdist, "tortoise-hdist",
```
countvonzero, line 190: i found in spacecraft configs use just "hdist"

config/presets/fastnet.go
```diff
@@ -0,0 +1,66 @@
...
+	conf.POST.MinNumUnits = 2
+
+	conf.SMESHING.CoinbaseAccount = "1"
+	conf.SMESHING.Start = false
+	conf.SMESHING.Opts.ComputeProviderID = 1
+	conf.SMESHING.Opts.NumFiles = 1
+	conf.SMESHING.Opts.NumUnits = 2
+	conf.SMESHING.Opts.Throttle = true
+
+	conf.Beacon.FirstVotingRoundDuration = 4 * time.Second
```
countvonzero: first voting round should be much longer. maybe 10 sec in this case

countvonzero (approved):

cmd/root.go
dshulyak: i forgot to update it, we are using default 10 anyways

dshulyak:
bors merge

nkryuchkov (approved):

PR #3106, opened 2022-01-31
[Merged by Bors] - sql: beacons

nkryuchkov:
## Motivation

Part of #2918 

## Changes

- move beacons storage to SQLite
- format SQL code
- extract `*sql.Database` from `mesh`

## Test Plan

UT, ST

## DevOps Notes

### DIFF
--- a/mesh/meshdb.go
+++ b/mesh/meshdb.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/spacemeshos/go-spacemesh/common/types"
-	"github.com/spacemeshos/go-spacemesh/common/util"
 	"github.com/spacemeshos/go-spacemesh/database"
 	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/log"
@@ -17,7 +16,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/signing"
 	"github.com/spacemeshos/go-spacemesh/sql"
 	"github.com/spacemeshos/go-spacemesh/sql/ballots"
-	"github.com/spacemeshos/go-spacemesh/sql/beacons"
 	"github.com/spacemeshos/go-spacemesh/sql/blocks"
 	"github.com/spacemeshos/go-spacemesh/sql/identities"
 	"github.com/spacemeshos/go-spacemesh/sql/layers"
@@ -128,11 +126,6 @@ func (m *DB) Blocks() database.Getter {
 // Transactions exports the transactions DB.
 func (m *DB) Transactions() database.Getter {
 	return &txFetcher{mdb: m}
-}
-
-// Beacons exports the beacon database.
-func (m *DB) Beacons() database.Store {
-	return &beaconDB{mdb: m}
 }
 
 // AddBallot adds a ballot to the database.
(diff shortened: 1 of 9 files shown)


nkryuchkov:
bors try

dshulyak:
i meant to extract db from mesh, and reuse it for different modules. instead of creating accessor, could you initialize db in cmd/node instead and pass the pointer to mesh and other modules?

nkryuchkov:
@dshulyak Do you mean passing a `*mesh.DB` to `beacon` and other packages?

dshulyak:
@nkryuchkov no, i was talking about *sql.Database. we can open it in the cmd/node and then pass it to all modules (mesh, beacon, activations, proposals, etc)

sql/migrations/0001_initial.sql
```diff
@@ -1,57 +1,72 @@
...
+CREATE INDEX transaction_by_origin ON transactions (origin, layer);
+CREATE INDEX transaction_by_destination ON transactions (destination, layer);
+
+CREATE TABLE beacons
+(
+    epoch  INT NOT NULL PRIMARY KEY,
+    beacon CHAR(4)
+);
+
+CREATE INDEX beacons_by_epoch ON beacons (epoch);
```
dshulyak: primary key is effectively is an index with unique a constraint, there is no need to add one more

sql/migrations/0001_initial.sql
```diff
@@ -1,57 +1,72 @@
...
-CREATE INDEX transaction_by_origin ON transactions(origin, layer);
-CREATE INDEX transaction_by_destination ON transactions(destination, layer);
+CREATE INDEX transaction_by_origin ON transactions (origin, layer);
+CREATE INDEX transaction_by_destination ON transactions (destination, layer);
+
+CREATE TABLE beacons
+(
+    epoch  INT NOT NULL PRIMARY KEY,
+    beacon CHAR(4)
+);
```
dshulyak: you can add WITHOUT ROWID. in this case it makes sense

sql/migrations/0001_initial.sql
nkryuchkov: Indeed, thanks. Removed

sql/migrations/0001_initial.sql
nkryuchkov: done

mesh/meshdb.go
```diff
@@ -41,15 +41,8 @@ type DB struct {
...
-	if err := os.MkdirAll(path, os.ModePerm); err != nil {
-		return nil, fmt.Errorf("failed to create %s: %w", path, err)
-	}
-	db, err := sql.Open("file:" + filepath.Join(path, "state.sql"))
-	if err != nil {
-		return nil, fmt.Errorf("open sqlite db %w", err)
-	}
-
+func NewPersistentMeshDB(db *sql.Database, blockCacheSize int, logger log.Log) (*DB, error) {
+	var err error
```
dshulyak: can we remove it and use err below in if statement? seemed like it hangs in some weird way

mesh/meshdb.go
```diff
@@ -137,6 +130,11 @@ func (m *DB) Transactions() database.Getter {
 	return &txFetcher{mdb: m}
 }
 
+// Beacons exports the beacon database.
+func (m *DB) Beacons() database.Store {
```
dshulyak: is it a leftover? i think it needs to be removed

mesh/meshdb.go
```diff
@@ -658,3 +656,22 @@ func LayerIDs(db database.Database, namespace string, lid types.LayerID, f func(
 	}
 	return database.ErrNotFound
 }
+
+type beaconDB struct {
```
dshulyak: and this too

dshulyak (approved):

mesh/meshdb.go
nkryuchkov: Sure, good idea

mesh/meshdb.go
nkryuchkov: I've already removed it, but after you started reviewing

mesh/meshdb.go
nkryuchkov: already removed, but after I requested your review

nkryuchkov:
bors merge

nkryuchkov:
bors merge

nkryuchkov:
bors merge

PR #3110, opened 2022-02-04
[Merged by Bors] - transactions: publish synchronously and allow to fetch from pool

dshulyak:
this change addresses several issues found in the implementation of tests using systest framework.
- transactions were published in separate goroutines, which sometimes resulted in an immediate failure due to the verification and sometimes out-of-order network delivery. i am not 100% sure that my change guarantees in-order delivery, but the test is consistently passing. 
- when the proposal is received node needs to verify that all transactions are available. however it doesn't look them up in the mempool, and almost always transactions will be available only in mempool. so it starts asking peers for transactions, and peers don't check mempool as well.. so only after it exhausts retries does it move on and find transactions in the mempool, saves them and makes progress this way. this is woefully inefficient and obviously a mistake.

## Changes
- execute validation synchronously in tx handler
- allow fetcher to look into the mempool for transactions
- misc logging changes for improved debugging


mesh/mesh.go
```diff
@@ -653,14 +653,7 @@ func (msh *Mesh) SetZeroBlockLayer(lyr types.LayerID) error {
 func (msh *Mesh) AddTXsFromProposal(ctx context.Context, layerID types.LayerID, proposalID types.ProposalID, txIDs []types.TransactionID) error {
 	logger := msh.WithContext(ctx).WithFields(layerID, proposalID, log.Int("num_txs", len(txIDs)))
 	logger.Debug("adding proposal txs to mesh")
-
 	msh.addTransactionsForBlock(logger, layerID, types.EmptyBlockID, txIDs)
-
-	if err := msh.storeTransactionsFromPool(layerID, types.EmptyBlockID, txIDs); err != nil {
```
countvonzero, line 659: why remove this? this should be the first chance of a TX making into the database.

mesh/mesh.go
```diff
@@ -653,14 +653,7 @@ func (msh *Mesh) SetZeroBlockLayer(lyr types.LayerID) error {
...
-		logger.With().Error("not all txs were processed", log.Err(err))
-	}
-
-	msh.setLatestLayer(layerID)
```
countvonzero, line 663: why remove this?

api/grpcserver/transaction_service.go
```diff
@@ -89,11 +89,10 @@ func (s TransactionService) SubmitTransaction(ctx context.Context, in *pb.Submit
 	log.Info("GRPC TransactionService.SubmitTransaction BROADCAST tx address: %x (len: %v), amount: %v, gas limit: %v, fee: %v, id: %v, nonce: %v",
 		tx.GetRecipient(), len(tx.GetRecipient()), tx.Amount, tx.GasLimit, tx.GetFee(), tx.ID().ShortString(), tx.AccountNonce)
 
-	go func() {
-		if err := s.publisher.Publish(ctx, svm.IncomingTxProtocol, in.Transaction); err != nil {
-			log.Error("error broadcasting incoming tx: %v", err)
-		}
-	}()
+	if err := s.publisher.Publish(ctx, svm.IncomingTxProtocol, in.Transaction); err != nil {
```
countvonzero, line 92: > transactions were published in separate goroutines, which sometimes resulted in an immediate failure due to the verification and sometimes out-of-order network delivery

what exactly is the verification failure? TXs should be mostly independent from each other. what kind of ordering is expected?

dshulyak: projected nonce verification. for example if bot publishes 100 transactions most of them will be dropped

mesh/mesh.go
dshulyak: because this code is redundant, please check in addTransactionsForBlock

mesh/mesh.go
dshulyak: same

api/grpcserver/transaction_service.go
dshulyak: they can't be totally independent. i know that we are not going to enforce strict order, but even then there will be a limit on  transactions that can be submitted concurrently.

mesh/mesh.go
countvonzero: we did this twice, yes. once when adding proposals and once when block is generated/synced. but we want those TXs in DB and the latest layer (most recent gossiped data) to be set as early as possible. i think it should stay as is.

api/grpcserver/transaction_service.go
countvonzero: but it sounds more like the way bot generate txs are not matching reality. if most TXs have different principals, then the async-ness make more sense.

countvonzero: > they can't be totally independent. i know that we are not going to enforce strict order, but even then there will be a limit on transactions that can be submitted concurrently.

maybe, but the way to honor limit should be ratelimiting instead of serializing the processing.

mesh/mesh.go
dshulyak: @countvonzero please take a look into the code. this code that i removed was redundant specifically in the code path for proposals

api/grpcserver/transaction_service.go
dshulyak: there is literally no reason to create a goroutine for publishing a transaction.. it doesn't enable anything but creates pretty bad artificail issues if account wants to apply transactions serially. i don't understand your point at all

mesh/mesh.go
dshulyak: you can see that transactions are saved on the line above removed line

countvonzero: oh! what?! i must have forgot to delete these 3 lines after creating `addTransactionsForBlock()`. sorry. my bad

api/grpcserver/transaction_service.go
countvonzero: hmmm... the ratelimiting should prolly done at the RPC level.

> if account wants to apply transactions serially.

why should we even honor `apply transactions serially` at the RPC level? that should be done by arranging nonce.

i recall you changed beacon to publish proposal to a goroutine because it does synchronous self-broadcast (and cause some issue that i don't know).  in any case, i am fine with changing to sync.

countvonzero (approved):

api/grpcserver/transaction_service.go
dshulyak: > why should we even honor apply transactions serially at the RPC level? that should be done by arranging nonce.

because then we can't add more than one transaction from the same account to the mempool reliably. so for example here is a test that sends 5 transactions from the same account https://github.com/dshulyak/systest/blob/master/tests/transactions_test.go#L74-L90 . without synchronous API i will have to wait until a transaction is added to the mempool using different means, for example listening to events. so it creates artificial usability issues.

> i recall you changed beacon to publish proposal to a goroutine because it does synchronous self-broadcast (and cause some issue that i don't know). in any case, i am fine with changing to sync.

that's a problem only if we are sending a message to ourselves and can accidentally deadlock, in beacon we are receiving a message in the same component and accidental deadlock is possible

dshulyak:
bors merge

PR #3111, opened 2022-02-04
[Merged by Bors] - api, events, proposal: stream created and included proposals

dshulyak:
proposals api is used in system tests to assert common functionality (proposals are equally created, hare creates proposals, beacon is the same on all proposals)

depends on https://github.com/spacemeshos/api/pull/162

## Changes
- fire events when proposals are created or included in the block
- add proposals stream to debug service, potentially it can be moved into "public" service
- remove old unused code from events module, i don't think that we have any plan on using it and it is a bit messy


api/grpcserver/debug_service.go
```diff
@@ -65,3 +69,61 @@ func (d DebugService) Accounts(_ context.Context, in *empty.Empty) (*pb.Accounts
...
+	var (
+		eventch <-chan interface{}
+		fullch  <-chan struct{}
+	)
+	if sub := events.SubcribeProposals(); sub != nil {
+		eventch, fullch = consumeEvents(stream.Context(), sub)
+	}
+	if eventch == nil || fullch == nil {
+		return status.Errorf(codes.FailedPrecondition, "event reporting is not enabled")
+	}
```
nkryuchkov: `consumeEvents` always returns non-nil channels, so `codes.FailedPrecondition` would mean that `events.SubcribeProposals() == nil`, maybe we can simplify this?
```suggestion
	sub := events.SubcribeProposals()
	if sub == nil {
		return status.Errorf(codes.FailedPrecondition, "event reporting is not enabled")
	}
	eventch, fullch := consumeEvents(stream.Context(), sub)
```

nkryuchkov (approved):

events/proposals.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// SubcribeProposals subscribes to the proposals.
+func SubcribeProposals() Subscription {
+	mu.RLock()
+	defer mu.RUnlock()
+	if reporter != nil {
+		sub, err := reporter.bus.Subscribe(new(EventProposal))
+		if err != nil {
+			log.With().Panic("Failed to subscribe to proposals")
```
countvonzero, line 54: how is this called? i am wondering whether panic is the right thing to do here.

api/grpcserver/debug_service.go
```diff
@@ -65,3 +69,61 @@ func (d DebugService) Accounts(_ context.Context, in *empty.Empty) (*pb.Accounts
 func (d DebugService) NetworkInfo(ctx context.Context, _ *empty.Empty) (*pb.NetworkInfoResponse, error) {
 	return &pb.NetworkInfoResponse{Id: d.identity.ID().String()}, nil
 }
+
+// ProposalsStream streams all proposals confirmed by hare.
+func (d DebugService) ProposalsStream(_ *emptypb.Empty, stream pb.DebugService_ProposalsStreamServer) error {
```
countvonzero, line 74: where can i find definition and pb files for DebugService_ProposalsStreamServer?

dshulyak: https://github.com/spacemeshos/api/pull/162

events/proposals.go
dshulyak: yes this is a right thing todo, all the errors condition are based on the input from the programmer https://github.com/libp2p/go-eventbus/blob/master/basic.go#L164-L225

api/grpcserver/debug_service.go
dshulyak: thanks, applied

countvonzero (approved):

dshulyak:
bors merge

PR #3112, opened 2022-02-05
[Merged by Bors] - testnet: new epoch every 2h with layer duration 2m

dshulyak:
the current testnet configuration is too unrealistic. verifying atxs is expensive, we shouldn't create them very often in the public testnets, as it leads to unnecessarily long sync times on resource-restricted machines.

## Changes
- change the layer to 120s. hare round 10s, 2 iterations of 4 rounds + pre round + wakeup + commitment round should fit in
- epoch size is 60 layers, so that atx is generated once in 2h

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -190,9 +190,9 @@ func loadConfig(cmd *cobra.Command) (*config.Config, error) {
 // LoadConfigFromFile tries to load configuration file if the config parameter was specified.
 func LoadConfigFromFile() (*config.Config, error) {
 	fileLocation := viper.GetString("config")
-	vip := viper.New()
+
 	// read in default config if passed as param using viper
-	if err := config.LoadConfig(fileLocation, vip); err != nil {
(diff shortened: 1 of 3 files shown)


cmd/node/node.go
```diff
@@ -190,9 +190,9 @@ func loadConfig(cmd *cobra.Command) (*config.Config, error) {
 // LoadConfigFromFile tries to load configuration file if the config parameter was specified.
 func LoadConfigFromFile() (*config.Config, error) {
 	fileLocation := viper.GetString("config")
-	vip := viper.New()
```
countvonzero, line 193: curious why the change to use global viper?

countvonzero (approved):

cmd/node/node.go
dshulyak: because using both requires checking variables in both, global is used when flags are parsed and local for reading config. i don't understand what was the motivation in the first place.

dshulyak:
bors merge

PR #3103, opened 2022-01-29
[Merged by Bors] - beacon: ignore votes on proposals unknown to the node

countvonzero:
## Motivation

Closes #3102

## Changes

ignore votes on proposals that are not in the node's initial proposals received during the proposal phase

this change can potentially cause disagreement on beacon value if network delay proves to be more significant than the grace period. we need empirical data to tune the grace period param to make sure miners agree on beacon value

## Test Plan

unittest, system tests

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #3113, opened 2022-02-05
[Merged by Bors] - beacon: allow early proposals and votes within grace period

countvonzero:
## Motivation

while debugging https://github.com/spacemeshos/go-spacemesh/pull/3103, i found that early proposals are rejected due to the `isInProtocol` check. 

early votes are also rejected due to the round number check. as nodes are not perfectly time-synced, some go into the next round earlier than others. 

this is not an issue in the current system tests, but may cause participants seeing different set of proposals in a larger network. we should accept early proposals and votes within grace period, accounting for time sync difference.

## Changes

- use context to decide whether the protocol is shutting down
- don't use channels to buffer proposals. instead, process proposals as they arrive. this simplify the processing logic
- allow proposals that arrive within grace period before its intended epoch begins.
- allow following votes that arrive within grace period before its intended round begins.

## Test Plan

unit tests / system tests

## DevOps Notes

### DIFF
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -30,7 +30,6 @@ import (
 )
 
 const (
-	protoName      = "BEACON_PROTOCOL"
 	proposalPrefix = "BP"
 
 	proposalChanCapacity = 1024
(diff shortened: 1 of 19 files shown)


countvonzero:
bors try

countvonzero:
bors try

beacon/interface.go
```diff
@@ -4,8 +4,12 @@ import (
 	"context"
 	"time"
 
+	"github.com/libp2p/go-libp2p-core/peer"
```
dshulyak: please use p2p.Peer, so that libp2p is not imported anywhere outside p2p module. if i forgot to update it somewhere that is a mistake

beacon/weakcoin/weak_coin_test.go
```diff
@@ -1,4 +1,4 @@
-package weakcoin_test
+package weakcoin
```
dshulyak: why? using _test package is a better practice

beacon/beacon.go
```diff
@@ -399,109 +484,95 @@ func (pd *ProtocolDriver) listenLayers(ctx context.Context) {
...
-
 func (pd *ProtocolDriver) setRoundInProgress(round types.RoundID) {
+	now := time.Now()
+	var nextRoundStartTime time.Time
+	if round == types.FirstRound {
+		nextRoundStartTime = now.Add(pd.config.FirstVotingRoundDuration)
+	} else {
+		nextRoundStartTime = now.Add(pd.config.VotingRoundDuration + pd.config.WeakCoinRoundDuration)
+	}
+	earliestVoteTime := nextRoundStartTime.Add(-1 * pd.config.GracePeriodDuration)
```
dshulyak: can be -pd.Config.GracePeriodDuration

dshulyak (approved):

beacon/interface.go
countvonzero: done, for other packages too.

beacon/weakcoin/weak_coin_test.go
countvonzero: changed back. i am aware this is better practice, but for the most part i find impractical and not consistent with the rest of the codebase.

beacon/beacon.go
countvonzero: done

countvonzero:
bors merge

PR #3115, opened 2022-02-09
[Merged by Bors] - TX service: only do checks in TX data handler

### LINKED ISSUE #2162: Deduplicate logic in API TransactionService and TransactionProcessor
This logic is perfectly duplicated:

https://github.com/spacemeshos/go-spacemesh/blob/bc91d5bf2bb95b742eff53be6281dd63022bd347/api/grpcserver/transaction_service.go#L60-L82

https://github.com/spacemeshos/go-spacemesh/blob/bc91d5bf2bb95b742eff53be6281dd63022bd347/state/processor.go#L314-L331

We don't need to run this logic twice. If we want the API to reject invalid transactions outright, then we need to make this logic modular so the API can run it directly, or call into the transaction processor (not currently exposed via the Mesh). Alternatively we may decide to let the API accept all incoming transactions and let the tx processor drop them later (before they're gossiped).

countvonzero:
## Motivation

Closes #2162

## Changes

remove duplicate TX check at RPC level. the gossip data handler does exactly the same check

## Test Plan

unit tests

## DevOps Notes

### DIFF
--- a/api/grpcserver/grpcserver_test.go
+++ b/api/grpcserver/grpcserver_test.go
@@ -1904,19 +1904,6 @@ func TestTransactionService(t *testing.T) {
 			require.Equal(t, globalTx.ID().Bytes(), res.Txstate.Id.Id)
 			require.Equal(t, pb.TransactionState_TRANSACTION_STATE_MEMPOOL, res.Txstate.State)
 		}},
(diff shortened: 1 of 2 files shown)


dshulyak (approved):

countvonzero:
bors merge

nkryuchkov (approved):

PR #3117, opened 2022-02-09
[Merged by Bors] - sql/blocks: ensure that blocks for layer are returned in order 

dshulyak:
we expect blocks to be lexicographically sorted when layer hash is computed

### DIFF
--- a/common/types/layer.go
+++ b/common/types/layer.go
@@ -76,7 +76,7 @@ func InitGenesisData() {
 		},
 		blockID: GenesisBlockID,
 	}
-	genesisLayer = NewExistingLayer(GetEffectiveGenesis(), []*Ballot{ballot}, []*Block{block})
+	genesisLayer = NewExistingLayer(GetEffectiveGenesis(), Hash32{}, []*Ballot{ballot}, []*Block{block})
 }
 
 // GetEffectiveGenesis returns when actual proposals would be created.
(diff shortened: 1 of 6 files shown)


nkryuchkov (approved):

countvonzero (approved):

dshulyak:
bors merge

common/types/layer.go
```diff
@@ -76,7 +76,7 @@ func InitGenesisData() {
 		},
 		blockID: GenesisBlockID,
 	}
-	genesisLayer = NewExistingLayer(GetEffectiveGenesis(), []*Ballot{ballot}, []*Block{block})
+	genesisLayer = NewExistingLayer(GetEffectiveGenesis(), Hash32{}, []*Ballot{ballot}, []*Block{block})
```
countvonzero, line 79: shouldn't this be hash of the actual genesis block?

common/types/layer.go
```diff
@@ -284,10 +285,7 @@ func (l *Layer) BallotIDs() []BallotID {
 
 // Hash returns the 32-byte sha256 sum of the block IDs in this layer, sorted in lexicographic order.
 func (l Layer) Hash() Hash32 {
```
countvonzero, line 287: at some point i was tempted to simply delete this method. it's too confusing with `mesh.GetLayerHash`

common/types/layer.go
dshulyak: i think there should be no genesis block or ballot, so i am not sure what the point to hash genesis block... this is a tech debt that needs to be removed.
i know that ballot is used in tortoise to bootstrap verifying mode. but i am not even sure why block was added

countvonzero: for the sake of consistency. agree that it's otherwise useless.
i am all for removing genesis block and ballot. i wasn't able to do that because i wasn't familiar enough with tortoise test setup.

dshulyak:
bors merge

PR #3120, opened 2022-02-11
[Merged by Bors] - systest: add framework for writing system tests

dshulyak:
See [systest README](https://github.com/dshulyak/go-spacemesh/blob/systest/systest/README.md#testing-approach) for details on motivation.

module systest is integrated as a separate golang module to not mess with go-spacemesh dependencies, as they are relatively separate and systest adds a lot of k8s related dependencies that are not necessary if you are not running system tests.

Makefile from systest will probably migrate to root level Makefile once we finalize integration with ci. 

### TestSmeshing

Showcases how multiple tests can be run in parallel on the same cluster. It creates two parallel tests, one of them generates transactions and the other one keeps tracks of the proposals. It is expected that both of them should succeed and therefore can share the same cluster.

### TestAddNodes

Tests that new nodes are able to join the cluster. It syncs only a few layers.

### TestFailedNodes

Tests that failed containers are successfully restarted.

### TestPartition

Partitions 7 nodes in two unequal partitions (5 and 2) and verify that they will re-join successfully. Note that equal partition will not rejoin, as nodes don't download ballots from older epochs and self-healing is never triggered.

## Systest usage

```
Usage of ./tests.test:
  -bootstrap duration
    	bootstrap time is added to the genesis time. it may take longer on cloud environmens due to the additional resource management (default 30s)
  -clusters int
    	controls how many clusters are deployed on k8s (default 1)
  -image string
    	go-spacemesh image (default "spacemeshos/go-spacemesh-dev:proposal-events")
  -keep
    	if true cluster will not be removed after test is finished
  -kubeconfig string
    	Paths to a kubeconfig. Only required if out-of-cluster.
  -labels value
    	test will be executed only if it matches all labels
  -level value
    	verbosity of the logger
  -namespace string
    	namespace for the cluster. if empty every test will use random namespace
  -node-selector value
    	select where test pods will be scheduled
  -poet-image string
    	poet server image (default "spacemeshos/poet:ef8f28a")
  -size int
    	size of the cluster. all test must use at most this number of smeshers (default 10)
  -test-timeout duration
    	timeout for a single test (default **30m0s)**
```

## Local setup

For local setup refer to [README](https://github.com/dshulyak/go-spacemesh/blob/systest/systest/README.md#installation).


systest/tests/smeshing_test.go
```diff
@@ -0,0 +1,120 @@
...
+		testSmeshing(t, tctx, cl)
+	})
+	t.Run("Transactions", func(t *testing.T) {
+		t.Parallel()
+		testTransactions(t, tctx, cl)
+	})
+}
+
+func testSmeshing(t *testing.T, tctx *testcontext.Context, cl *cluster.Cluster) {
+	const limit = 15
```
countvonzero, line 32: what is the concept of limit? below you check for proposal layer id to be less than this number. how is this controled?

systest/tests/smeshing_test.go
```diff
@@ -0,0 +1,120 @@
...
+					"layer=%d client=%s", layer, cl.Client(i).Name)
+			}
+		}
+	}
+	for epoch := range beacons {
+		require.Len(t, beacons[epoch], 1, "epoch=%d", epoch)
+	}
+}
+
+func requireEqualEligibilities(tb testing.TB, proposals map[uint32][]*spacemeshv1.Proposal) {
```
countvonzero, line 107: just want to clarify, this is because every node in the cluster have equal weight. right?

systest/tests/smeshing_test.go
```diff
@@ -0,0 +1,120 @@
...
+		created[proposal.Layer.Number] = append(created[proposal.Layer.Number], proposal)
+		if edata := proposal.GetData(); edata != nil {
+			if _, exist := beacons[proposal.Epoch.Value]; !exist {
+				beacons[proposal.Epoch.Value] = map[string]struct{}{}
+			}
+			beacons[proposal.Epoch.Value][prettyHex(edata.Beacon)] = struct{}{}
+		}
+	}
+	requireEqualEligibilities(t, created)
+	for layer := range created {
```
countvonzero: 79-97: perhaps create a method (much like requireEqualEligibilities) to explain what this is verifying. i dont fully understand.

systest/tests/transactions_test.go
```diff
@@ -0,0 +1,92 @@
...
+		batch       = 1
+		amount      = 100
+	)
+	receiver := [20]byte{11, 1, 1}
+
+	eg, ctx := errgroup.WithContext(tctx)
+	for i := 0; i < keys; i++ {
+		client := cl.Client(i % cl.Total())
+		submitter := newTransactionSubmitter(cl.Private(i), receiver, amount, client)
+		collectLayers(ctx, eg, client, func(layer *spacemeshv1.LayerStreamResponse) (bool, error) {
```
countvonzero: maybe don't call this collectLayers, as you are submitting TXs.

countvonzero (approved):

systest/tests/smeshing_test.go
dshulyak: limit is just a number of the layer. when watcher sees this number it will exit

systest/tests/smeshing_test.go
dshulyak: yes

systest/tests/transactions_test.go
dshulyak: right, the idea is to have a generic function that watches layers. probably `watchLayers` will be a better name

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

PR #3124, opened 2022-02-14
[Merged by Bors] - sql: PoETs

nkryuchkov:
## Motivation

## Changes

- move PoETs storage to SQLite

## Test Plan

UT, ST

## DevOps Notes


nkryuchkov:
bors try

sql/migrations/0001_initial.sql
```diff
@@ -88,3 +88,18 @@ CREATE TABLE atx_top
...
+    poet BLOB
+) WITHOUT ROWID;
+
+CREATE TABLE poet_subscriptions
+(
+    service_id VARCHAR,
+    round_id   VARCHAR,
+    ref        VARCHAR,
+    PRIMARY KEY (service_id, round_id)
+);
```
dshulyak, line 98: please merge these two tables. to keep things simple ref can remain primary key, and round_id, service_id can be an index if they are queried somewhere.

nkryuchkov:
bors try

activation/poetdb.go
```diff
@@ -83,31 +86,45 @@ func (db *PoetDb) storeProof(proofMessage *types.PoetProofMessage) error {
...
+	if err != nil {
+		return err
 	}
-	key := makeKey(proofMessage.PoetServiceID, proofMessage.RoundID)
-	if err := batch.Put(key[:], ref); err != nil {
-		return fmt.Errorf("failed to store poet proof index entry for poetId %x round %s: %v",
+
+	defer dbtx.Release()
+
+	if err := poets.Add(dbtx, ref, messageBytes, proofMessage.PoetServiceID, proofMessage.RoundID); err != nil {
```
countvonzero: why does this need a dbtx if there is only one stmt to execute?

nkryuchkov: Forgot to remove it after resolving a CR request. Thanks. Removed

sql/poets/poets.go
```diff
@@ -0,0 +1,77 @@
+package poets
+
+import (
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+// GetPoET gets a PoET for a given ref.
+func GetPoET(db sql.Executor, ref []byte) (poet []byte, err error) {
```
dshulyak: maybe just Get instead of GetPoeT and GetBlob ?

activation/poetdb.go
```diff
@@ -216,3 +213,23 @@ func validatePoet(membershipRoot []byte, merkleProof shared.MerkleProof, leafCou
...
+func (db *PoetDb) PoETs() database.Getter {
+	return newPoETFetcherDB(db)
+}
+
+func newPoETFetcherDB(db *PoetDb) *PoETFetcher {
+	return &PoETFetcher{DB: db}
+}
+
+// PoETFetcher is an adapter of SQLite implementation to legacy LevelDB interfaces.
+// TODO(nkryuchkov): Remove when transition to SQLite is finished.
```
dshulyak: it is not about legacy leveldb, this is an interface for the fetcher.
we probably won't get rid of it in near future

sql/migrations/0001_initial.sql
```diff
@@ -88,3 +88,13 @@ CREATE TABLE atx_top
...
     layer  INT NOT NULL
 );
+
+CREATE TABLE poets
+(
+    ref        VARCHAR PRIMARY KEY,
+    poet       BLOB,
+    service_id VARCHAR,
+    round_id   VARCHAR
+) WITHOUT ROWID;
```
dshulyak: WITHOUT ROWID is not good with big blobs, it reduces tree fanout that makes traversal slower
checkout [part 4](https://www.sqlite.org/withoutrowid.html)

dshulyak (approved):

sql/poets/poets.go
nkryuchkov: I don't mind

activation/poetdb.go
nkryuchkov: I reviewed the fetcher code — yes, you're right. I'll remove the same comment in ATXs.

sql/migrations/0001_initial.sql
nkryuchkov: thanks for the link

nkryuchkov:
bors merge

PR #3128, opened 2022-02-17
[Merged by Bors] - tortoise: if ballots misses some data don't add it to the state

### LINKED ISSUE #3065: tortoise, beacon: OnBallot handler is called before beacon is available
Noticed these errors during sync:

```
2140:2022-01-19T13:56:10.534+0200	ERROR	198a3.trtl         	failed to get beacon for epoch	{"node_id": "198a33ff64764343494a202e62a77536abf11640b3767eb505b272c4325ce25b", "module": "trtl", "epoch_id": 3, "errmsg": "beacon is not calculated for this epoch", "name": "trtl"}
2146:2022-01-19T13:56:10.536+0200	ERROR	198a3.trtl         	failed to get beacon for epoch	{"node_id": "198a33ff64764343494a202e62a77536abf11640b3767eb505b272c4325ce25b", "module": "trtl", "epoch_id": 3, "errmsg": "beacon is not calculated for this epoch", "name": "trtl"}
```

Beacon got eventually determined, as expected after some number of ballots:

```
2324:2022-01-19T13:56:11.631+0200	INFO	198a3.tBeacon      	beacon determined for epoch	{"node_id": "198a33ff64764343494a202e62a77536abf11640b3767eb505b272c4325ce25b", "module": "tBeacon", "epoch_id": 3, "num_ballots": 30, "beacon": "e717411682", "weight": 60, "name": "tBeacon"}
```

Two problems:
- if beacon is not found tortoise should delay votes from ballots, not count them immediatly
- OnBallot handler should no modify any internal data if beacon is not available

dshulyak:
Tortoise won't add the ballot to the state If the state doesn't have a beacon from the ballots epoch, base ballot, or is unable to compute ballot weight due to the missing atx or reference ballot. This ensures that the state won't be corrupted by an improperly computed state, for example, because some of the votes will be missing, or ballot won't be marked bad when it actually supposed to be marked bad due to the difference with a beacon.

closes: https://github.com/spacemeshos/go-spacemesh/issues/3065

## Changes
- return an error from onBallot if a base ballot is not found in the state
- return an error if beacon from the ballot epoch is not available locally
- don't fail layer processing if some of the ballots are missing data, generally, it shouldn't happen, but sometimes there could be ballots that are using base ballot outside of the window that is kept in memory


countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #3129, opened 2022-02-17
[Merged by Bors] - sql: proposals

nkryuchkov:
## Motivation

## Changes

- move proposals storage to SQLite

## Test Plan

Unit and system tests

## DevOps Notes


common/types/proposal.go
```diff
@@ -86,6 +86,11 @@ func (p *Proposal) ID() ProposalID {
 	return p.proposalID
 }
 
+// SetID set the ProposalID.
```
countvonzero, line 89: can delete DBProposal

proposals/proposaldb.go
```diff
@@ -62,28 +50,18 @@ func newDB(opts ...dbOpt) *DB {
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize proposal layers db: %w", err)
-	}
-	return newDB(withProposalDB(pdb), withLayerDB(ldb), withMeshDB(msh), withLogger(logger)), nil
+func NewProposalDB(sqlDB *sql.Database, msh meshDB, logger log.Log) (*DB, error) {
+	return newDB(withSQLDB(sqlDB), withMeshDB(msh), withLogger(logger)), nil
 }
 
 // Close closes all resources.
 func (db *DB) Close() {
```
countvonzero: delete this?

proposals/proposaldb_test.go
```diff
@@ -2,176 +2,155 @@ package proposals
...
 	*DB
-	mockMesh *mocks.MockmeshDB
+	mesh *mesh.Mesh
 }
 
 func createTestDB(t *testing.T) *testDB {
 	types.SetLayersPerEpoch(layersPerEpoch)
+
+	db := sql.InMemory()
+	mdb, err := mesh.NewPersistentMeshDB(db, 1, logtest.New(t))
```
countvonzero: i don't understand why you are testing with a real mesh.
please use mocks.MockmeshDB. that way you don't need to care about mock tortoise and conservative state.

sql/proposals/proposals.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+// Get gets a proposal by a given ID.
+func Get(db sql.Executor, id types.ProposalID) (proposal *types.Proposal, err error) {
+	if rows, err := db.Exec(`
+		select 
+			ballots.signature, 
+			ballots.pubkey, 
+			ballots.ballot,
```
countvonzero, line 19: this is not necessary. in the app, after a proposal is saved, we almost never get the ballot again from a proposal.
only hare cares about proposals again.
i.e. you don't need to query ballot when getting proposal. if the odd case that we need to you can always separate to Get and GetWithBallot or something like that.

proposals/proposaldb_test.go
nkryuchkov: I wanted to test querying proposals with ballots, so I needed to add ballots to SQL DB too. Another option was to add them to the actual DB using `gomock`, but I thought that testing with a real mesh would be more reliable. Although, I can change this

countvonzero: why do you want to test querying proposals with ballots?
once proposals come in via gossip or sync, proposals and ballots are totally separate entities.
right now in tests Proposal is stitched back together with a separate Ballot obj. i am not sure if thats even necessary.
so yes, please dont use a real mesh in test

nkryuchkov: I thought it would be better to just copy the original behavior. But assuming what you said, a real mesh is not needed, I've already removed it and will update the PR soon

sql/migrations/0001_initial.sql
```diff
@@ -89,6 +89,17 @@ CREATE TABLE atx_top
     layer  INT NOT NULL
 );
 
+CREATE TABLE proposals
+(
+    id        CHAR(20) PRIMARY KEY,
+    ballot_id CHAR(20),
+    layer     INT NOT NULL,
+    tx_ids    BLOB,
```
countvonzero, line 97: proposal itself should also be a BLOB for servicing fetch by hash

proposals/proposaldb_test.go
countvonzero: i think you are talking about the fetch, right? you should follow the other tables (ballots/atx) that serve hash requests with a blob

sql/proposals/proposals.go
countvonzero: actually i am not fully correct here. hare wants to know the beacon value recorded in a proposal. note that proposal data is ephemeral and will be deleted after a while.

maybe two options
- add a field beacon
- i think its better to add a blob for the whole proposal for serving fetch. and if that's the case, might as well decode the proposal blob instead of the ballot blob.

countvonzero: thinking more about it, it's fine to leave your query code as is.
but still add a BLOB for the proposal for serving fetch.

proposals/proposaldb_test.go
countvonzero: in this cause you should do 

td.mockMesh.EXPECT().AddBallot(&p.Ballot).DoAndReturn(func() {
  // save ballots to db
}).Times(1)

countvonzero:
i am tempted to use two structs for proposals. one is the original but renamed to GossipProposal. it is saved as a BLOB in db, and used to serve fetch requests. one struct is Proposal, which contains BallotID,  Beacon, TXs and Signature. both will be saved under one db table with ID as the primary key. i think its cleaner and makes more sense.

sorry for the back and forth. if you dont feel like doing this, i can do it instead

nkryuchkov:
In `turbohare`, when we generate a block, we pass a proposals list taken from the DB (https://github.com/spacemeshos/go-spacemesh/blob/develop/turbohare/hare.go#L92). Then, in blocks generator, in `calculateSmesherRewards`, we use those proposals to access `EligibilityProofs` (https://github.com/spacemeshos/go-spacemesh/blob/develop/blocks/generator.go#L114) and `AtxID` (https://github.com/spacemeshos/go-spacemesh/blob/develop/blocks/generator.go#L120). So let's leave it as it is.

proposals/proposaldb_test.go
```diff
@@ -23,44 +23,25 @@ type testDB struct {
...
-func TestDB_AddProposal_FailToAddTXFromProposal(t *testing.T) {
-	td := createTestDB(t)
-	layerID := types.GetEffectiveGenesis().Add(10)
-	p := types.GenLayerProposal(layerID, types.RandomTXSet(100))
-	td.mockMesh.EXPECT().AddBallot(&p.Ballot).Return(nil).Times(1)
-	errUnknown := errors.New("unknown")
-	td.mockMesh.EXPECT().AddTXsFromProposal(gomock.Any(), p.LayerIndex, p.ID(), p.TxIDs).Return(errUnknown).Times(1)
-	assert.ErrorIs(t, td.AddProposal(context.TODO(), p), errUnknown)
-}
-
```
countvonzero, line 56: can you restore these tests?

countvonzero (approved):

nkryuchkov:
bors merge

Issue #3035, opened 2021-12-28
Proposal Construction


noamnelke:
A proposal is a set of transactions randomly selected from the predicted block. The block is predicted by assuming all smeshers see the same mempool and selecting the highest gas-price transactions that fit into a block out of the mempool.

Given pre-configured values for `block_gas_limit` and `proposal_num_transactions` perform the following:
1. Predict the contents of the next block by doing the following:
    ```python
    def predict_block():
        remaining_gas = block_gas_limit
        txs = []
        for tx in mempool.txs_by_gas_price_and_seniority():
            if tx.max_gas > remaining_gas:
                continue
            remaining_gas -= tx.max_gas
            txs.append(tx)
            if remaining_gas < MIN_TX_GAS:  # gas required for the most basic transaction
                break
        return txs
    ```
    _Note: Because transactions currently don't have a `max_gas` property (more precisely, it's not used) we should add a getter to transactions that currently returns a fixed value of 1. This should work until we finish SVM integration, when the getter will return the correct value._

2. Randomly select `proposal_num_transactions` from the predicted block (using local randomness).
_Note: The randomness of a proposal's transactions cannot be verified and doesn't need to be. Since layer committee members share rewards, they are naturally incentivized to collaboratively produce a block with all highest `gas_price` transactions. Since there will be some redundancy in proposed transaction, a small number of deviating participants won't significantly change the outcome._
3. Check if all layers prior to the proposal layer are verified. If they are, include the `state_root` in the proposal, otherwise use a sentinel value (e.g. all-zeros).
4. Sign the proposal and publish it (along with the ballot, which includes the layer committee eligibility proof).

countvonzero:
@noamnelke 

> 3. Check if all layers prior to the proposal layer are verified. If they are, include the state_root in the proposal, otherwise use a sentinel value (e.g. all-zeros).

if by "verified" you mean by tortoise, this is NEVER the case.
in the most optimistic case, the latest hare committee is running on layer N, the node only has ballots up to layer N-1, this means the highest layer tortoise can verify is layer N-2. 

cc @dshulyak

dshulyak:
I realized some time ago that this a shortcoming of the current implementation (https://github.com/spacemeshos/go-spacemesh/issues/3054). 

We don't need to wait for a hare from layer N to complete in order to verify layer N - 1. we just need to wait for ballots from layer N and they are expected to be received much sooner, like 5-10s after layer N has started. It does require some additional work.

[...]

Issue #3003, opened 2021-12-16
tortoise, sync: required data for tortoise when syncing a new node


dshulyak:
In the current tortoise version expected weight is computed based on the difference from the last layer with the layer that we want to verify (target layer). Currently last layer is just a last layer that was ever received by the tortoise. But the last layer must be set according to the clock as defined in https://github.com/spacemeshos/go-spacemesh/issues/2921.

It brings up two additional requirements on sync:
- [x]  all atxs must be added to the database before tortoise will start verification, otherwise, it won't be able to compute the expected weight
- [ ] in order to use verifying tortoise we need to get input vectors from peers. previously it was mistakenly assumed that input vector means hare output, in fact it means actual local peer opinion from layers before hdist, and hare output for layers in the hdist distance.

Basically we will have to implement this issue before https://github.com/spacemeshos/go-spacemesh/issues/2921 is implemented.

cc @countvonzero

countvonzero:
the second part of this issue is currently blocked by #3283

PR #3131, opened 2022-02-18
[Merged by Bors] - TX Selection: optimistically apply a layer after hare terminates

countvonzero:
## Motivation

1st PR of #3035

## Changes

- apply a layer as soon as hare terminates
- calculate layer hash/agg. hash with just the applied block, not all contextually valid blocks
- mesh self-determines the layer to revert to based on change on contextual validity of blocks. 
  tortoise simply updates contextual validity and no longer needs to care which layer the state should revert to. this is a cleaner separation between consensus and state.

## TODO

- [ ] Update tests

## DevOps Notes

### DIFF
--- a/activation/poetdb.go
+++ b/activation/poetdb.go
@@ -1,7 +1,6 @@
 package activation
 
 import (
-	"errors"
 	"fmt"
 	"sync"
 
@@ -90,7 +89,7 @@ func (db *PoetDb) StoreProof(proofMessage *types.PoetProofMessage) error {
 		return fmt.Errorf("could not marshal proof message: %v", err)
 	}
 
-	if err := poets.Add(db.sqlDB, ref, messageBytes, proofMessage.PoetServiceID, proofMessage.RoundID); err != nil && !errors.Is(err, sql.ErrObjectExists) {
+	if err := poets.Add(db.sqlDB, ref, messageBytes, proofMessage.PoetServiceID, proofMessage.RoundID); err != nil {
 		return fmt.Errorf("failed to store poet proof for poetId %x round %s: %v",
 			proofMessage.PoetServiceID[:5], proofMessage.RoundID, err)
 	}
--- a/activation/poetdb.go
+++ b/activation/poetdb.go
@@ -1,6 +1,7 @@
 package activation
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
@@ -89,7 +90,7 @@ func (db *PoetDb) StoreProof(proofMessage *types.PoetProofMessage) error {
 		return fmt.Errorf("could not marshal proof message: %v", err)
 	}
 
-	if err := poets.Add(db.sqlDB, ref, messageBytes, proofMessage.PoetServiceID, proofMessage.RoundID); err != nil {
+	if err := poets.Add(db.sqlDB, ref, messageBytes, proofMessage.PoetServiceID, proofMessage.RoundID); err != nil && !errors.Is(err, sql.ErrObjectExists) {
 		return fmt.Errorf("failed to store poet proof for poetId %x round %s: %v",
 			proofMessage.PoetServiceID[:5], proofMessage.RoundID, err)
 	}
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -77,7 +77,7 @@ func NewMesh(db *DB, atxDb AtxDB, trtl tortoise, state conservativeState, logger
 				msh.With().Panic("failed to set zero-block for genesis layer", i, log.Err(err))
 			}
 		}
-		if err := msh.persistLayerHashes(context.Background(), i, types.EmptyBlockID); err != nil {
+		if err := msh.persistLayerHashes(context.Background(), i, []types.BlockID{types.EmptyBlockID}); err != nil {
 			msh.With().Panic("failed to persist hashes for layer", i, log.Err(err))
 		}
 		msh.setProcessedLayer(i)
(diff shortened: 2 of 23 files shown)


countvonzero:
@dshulyak 
i want to check if this approach is ok with you before i start updating tests for the `HandleIncomingLayer` API change.

dshulyak:
> i want to check if this approach is ok with you before i start updating tests for the HandleIncomingLayer API change.

i think it is fine. i had in mind that `reverted` is basically useless, and also i wanted to remove saving contextual validity from the tortoise, but i am not sure if i will get to it any time soon.

but since you are also adding another variable to mesh, i don't understand what the point of keeping `oldVerified` variable? is it possible to keep only one?

dshulyak:
also, there is this https://github.com/spacemeshos/go-spacemesh/issues/3003 . we will need to download contextual validity from peers. just to ensure that this solution doesn't conflict with that somehow

countvonzero:
> i don't understand what the point of keeping oldVerified variable? is it possible to keep only one?

yes. i was keeping for logging purpose but i guess tortoise already logged

> also, there is this #3003 . we will need to download contextual validity from peers. just to ensure that this solution doesn't conflict with that somehow

i dont think it will conflict. but in stead of `hareoutput before hdist and contextual validity after hdist` a node can just return the block ID it applied for a given layer.

dshulyak:
> i dont think it will conflict. but in stead of hareoutput before hdist and contextual validity after hdist a node can just return the block ID it applied for a given layer.

the difference here is that we need to download a certificate for layers within hdist

countvonzero:
@dshulyak the PR is ready for view. thanks

countvonzero:
bors try

activation/poetdb.go
```diff
@@ -90,7 +91,7 @@ func (db *PoetDb) storeProof(proofMessage *types.PoetProofMessage) error {
 		return fmt.Errorf("could not marshal proof message: %v", err)
 	}
 
-	if err := poets.Add(db.sqlDB, ref, messageBytes, proofMessage.PoetServiceID, proofMessage.RoundID); err != nil {
+	if err := poets.Add(db.sqlDB, ref, messageBytes, proofMessage.PoetServiceID, proofMessage.RoundID); err != nil && !errors.Is(err, sql.ErrObjectExists) {
```
dshulyak: if the object already exists we should return Ignore (not Accept) in p2p handler, so maybe this error needs to be handled later

mesh/mesh.go
```diff
@@ -127,6 +136,46 @@ func (msh *Mesh) CacheWarmUp(layerSize int) {
...
+	}
+
+	if oldValid != newValid {
+		for {
+			minUpdated := msh.getMinUpdatedLayer()
+			if minUpdated != (types.LayerID{}) && !lid.Before(minUpdated) {
+				break
+			}
+			if msh.minUpdatedLayer.CompareAndSwap(minUpdated, lid) {
+				msh.With().Debug("min updated layer set for block", lid, bid)
```
dshulyak, line 154: break

mesh/mesh.go
```diff
@@ -255,6 +304,48 @@ func (msh *Mesh) setProcessedLayer(layerID types.LayerID) {
...
+
+	var revertTo types.LayerID
+	for lid := minUpdated; !lid.After(newVerified); lid = lid.Add(1) {
+		block, _, err := msh.getBlockToApply(ctx, lid, newVerified)
+		if err != nil {
+			return err
+		}
+		hash := msh.GetLayerHash(lid)
+		if hash != types.EmptyLayerHash &&
+			hash != types.CalcBlocksHash32([]types.BlockID{block.ID()}, nil) {
```
dshulyak: what if when we get hare output we set contextual validity of that block to true as well. and once minUpdated is not nil we know exactly which layer to apply. will it work?

mesh/mesh.go
```diff
@@ -408,40 +442,69 @@ func (msh *Mesh) pushLayersToState(ctx context.Context, from, to types.LayerID)
...
 	var (
-		applied    *types.Block
-		notApplied = invalidBlocks
-		blocks     = types.SortBlocks(validBlocks)
+		toApply *types.Block
+		others  []*types.Block
 	)
 
-	if len(blocks) > 0 {
+	if layerID.After(latestVerified) {
```
dshulyak, line 474: i think this routine will be simplified as well if we save hare output as a contextually valid block. in such case there will be no need for special case for hare

dshulyak (approved):
the error handling in poet needs to be changed, otherwise, there will be a gossip spam

mesh/mesh.go
countvonzero: there are some concerns with this approach:
- minUpdated can be not nil but we don't necessarily have to revert. if block_A was applied, and then block_B later was also valid, we don't need to revert the state. just need to update the hash.
- in `GetHareConsensusOutput()` is mesh's only way to know if hare has terminated for that layer. we currently is relying on that to return an error to skip applying a layer when hare terminated out of order. if we solely rely on contextual validity, then we may end up applying an empty block for a late hare. for example, if hare terminates in this order:N->N+2->N+1. at the time N+2 terminated, we will apply an empty block for layer N+1
- in tortoise, there is also an assumption about the existence of contextual validity implies that tortoise has verified that layer. i am hesitant to break that assumption at this point.

mesh/mesh.go
countvonzero: done

activation/poetdb.go
countvonzero: done separately in https://github.com/spacemeshos/go-spacemesh/pull/3142

countvonzero:
bors try

countvonzero:
bors merge

PR #3141, opened 2022-02-21
[Merged by Bors] - sql: ref ballots

nkryuchkov:
## Motivation

## Changes

- move ref ballots storage to SQLite

## Test Plan

Unit and system tests

## DevOps Notes


nkryuchkov:
bors try

sql/migrations/0001_initial.sql
```diff
@@ -17,6 +17,12 @@ CREATE TABLE ballots
 );
 CREATE INDEX ballots_by_layer_by_pubkey ON ballots (layer, pubkey);
 
+CREATE TABLE ref_ballots
```
countvonzero: why does ref ballot need to be a separate table and not a field of ballots table?

btw, the primary key should be (epoch, smesher public key)

dshulyak:
@nkryuchkov this change is not crash safe. 

1. save ref ballot
2. crash
3. save ballot

when app recovers we will have a pointer, but not the ballot itself. note that we can't simply reverse the order, as it will result in a different problem.

as i mentioned this approach is ok only if pointer is saved atomically with the ballot itself. after reviewing code we can actually reuse existing index to find ref ballot (`ballots_by_layer_by_pubkey`) . the ref ballot would be the first ballot in the layer from epoch range. i think we should use that approach since it is both crash-safe and simplifies db access (we actually need only one additional read)

nkryuchkov:
bors try

sql/ballots/ballots.go
```diff
@@ -148,3 +148,28 @@ func CountByPubkeyLayer(db sql.Executor, lid types.LayerID, pubkey []byte) (int,
...
 	return rows, nil
 }
+
+// GetRefBallot gets a ref ballot for a layer and a pubkey.
+func GetRefBallot(db sql.Executor, epochID types.EpochID, pubkey []byte) (ballotID types.BallotID, err error) {
+	firstLayer := epochID.FirstLayer()
+	lastLayer := firstLayer.Add(types.GetLayersPerEpoch())
+	rows, err := db.Exec(`
+		select * from ballots 
+		where layer between ?1 and ?2 and pubkey = ?3
```
dshulyak, line 158: `between` keyword inclusive or exclusive? would be nice to have a test for this case as well

dshulyak:
please add a test for new function, also i don't remember if index is sorted in asc order by default, it is better to make this change 

```
-CREATE INDEX ballots_by_layer_by_pubkey ON ballots (layer, pubkey);
+CREATE INDEX ballots_by_layer_by_pubkey ON ballots (layer asc, pubkey);
```

miner/proposal_builder_test.go
```diff
@@ -297,55 +282,13 @@ func TestBuilder_HandleLayer_BaseBlockError(t *testing.T) {
 	assert.ErrorIs(t, b.handleLayer(context.TODO(), layerID), errUnknown)
 }
 
-func TestBuilder_HandleLayer_GetRefBallotError(t *testing.T) {
```
countvonzero, line 300: can you change this test to use sqlite?

sql/ballots/ballots.go
```diff
@@ -148,3 +148,28 @@ func CountByPubkeyLayer(db sql.Executor, lid types.LayerID, pubkey []byte) (int,
 	}
 	return rows, nil
 }
+
+// GetRefBallot gets a ref ballot for a layer and a pubkey.
+func GetRefBallot(db sql.Executor, epochID types.EpochID, pubkey []byte) (ballotID types.BallotID, err error) {
+	firstLayer := epochID.FirstLayer()
+	lastLayer := firstLayer.Add(types.GetLayersPerEpoch()).Sub(1)
+	rows, err := db.Exec(`
+		select * from ballots
```
countvonzero: in code you are consistent about getting ballot including the malicious field. i think you should do it here too.
i think the query should be the same as `Get()` with the extra where clause

sql/ballots/ballots.go
```diff
@@ -148,3 +148,28 @@ func CountByPubkeyLayer(db sql.Executor, lid types.LayerID, pubkey []byte) (int,
...
+			stmt.BindInt64(2, int64(lastLayer.Value))
+			stmt.BindBytes(3, pubkey)
+		}, func(stmt *sql.Statement) bool {
+			stmt.ColumnBytes(0, ballotID[:])
+			return true
+		})
+	if err != nil {
+		return types.BallotID{}, fmt.Errorf("ref ballot epoch %v: %w", epochID, err)
+	} else if rows == 0 {
+		return types.BallotID{}, fmt.Errorf("%w ballot epoch %s", sql.ErrNotFound, epochID)
```
countvonzero: ref ballot epoch

miner/proposal_builder_test.go
countvonzero: done

sql/ballots/ballots.go
countvonzero: changed to `select id` instead

sql/ballots/ballots.go
countvonzero: done

countvonzero (approved):

countvonzero:
bors merge

nkryuchkov:
@countvonzero thanks for fixing

PR #3142, opened 2022-02-22
[Merged by Bors] - POET: don't gossip on repeat poet proof

countvonzero:
## Motivation

repeat poet proof cause log spam and network spam.

## Changes

- ignore repeat poet proof msg at p2p layer
- changed test to use gmock

## Test Plan

unit tests

## DevOps Notes


activation/poetlistener.go
```diff
@@ -41,7 +37,11 @@ func (l *PoetListener) HandlePoetProofMessage(ctx context.Context, _ p2p.Peer, m
 		return pubsub.ValidationIgnore
 	}
 
-	if err := l.poetDb.storeProof(&proofMessage); err != nil {
+	if err := l.poetDb.StoreProof(&proofMessage); err != nil {
+		if err == sql.ErrObjectExists {
+			// don't spam the network
+			return pubsub.ValidationIgnore
+		}
```
dshulyak, line 45: i thought that we are already not sending it to the network in cases when validation failed :/.

what about layerfetcher? will it handle ErrObjectExists correctly?

activation/poetlistener.go
```diff
@@ -41,7 +37,11 @@ func (l *PoetListener) HandlePoetProofMessage(ctx context.Context, _ p2p.Peer, m
 		return pubsub.ValidationIgnore
 	}
 
-	if err := l.poetDb.storeProof(&proofMessage); err != nil {
+	if err := l.poetDb.StoreProof(&proofMessage); err != nil {
+		if err == sql.ErrObjectExists {
```
dshulyak: errors.Is ?

dshulyak (approved):

activation/poetlistener.go
countvonzero: it treated it as a validation error which i think incorrect. fixed

activation/poetlistener.go
countvonzero: chagned

noamnelke:
Checking the database on-disk for the object is inefficient, since getting the same gossip message from multiple peers close to the same time shouldn't be a rare occasion (depending on how long it takes to transmit and validate, but a PoET message isn't very small or fast to validate).

I think there should be some fixed-length in-memory LRU cache for the IDs of received objects of each type.

What this PR does should probably exist **in addition** to the cache.

dshulyak:
> Checking the database on-disk for the object is inefficient, since getting the same gossip message from multiple peers close to the same time shouldn't be a rare occasion (depending on how long it takes to transmit and validate, but a PoET message isn't very small or fast to validate).

p2p layer has a cache with time based eviction policy. this particular check is needed because every peer evicts messages based on its own clock. because of that we may end up in the situation when same message gets gossiped multiple times

countvonzero:
yes. right now i am operating on the theory that p2p layer will block the first duplicate messages. at he message-specific handlers, we do use database to check whether we have the messages (block/proposals). in the future we can create msg-specific cache as inefficiency is observed.

countvonzero:
bors merge

countvonzero:
bors merge

PR #3145, opened 2022-03-13
logging: add `log-filepath` flag, rotate logs if it's set

### LINKED ISSUE #3092: logging: add log rotation with lumberjack
Logging is pretty verbose even with INFO level.
Setup rotation with some sensible limits, for example 1day of logging or 2gb of data

https://github.com/natefinch/lumberjack

nkryuchkov:
## Motivation

Closes https://github.com/spacemeshos/go-spacemesh/issues/3092

## Changes

- add `log-filepath` flag
- rotate logs if `log-filepath` flag is set

## Test Plan

unit and system tests

## DevOps Notes

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -10,6 +10,7 @@ import (
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"gopkg.in/natefinch/lumberjack.v2"
 )
 
 // mainLoggerName is a name of the global logger.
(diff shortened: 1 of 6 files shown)


log/log.go
```diff
@@ -96,6 +97,22 @@ func JSONLog(b bool) {
 	initLogging()
 }
 
+// FileLog turns on logging to file
+func FileLog(filepath string) {
```
dshulyak, line 101: there is probably no need to print output to stdout if it is logged to a file. and we should expose config options for file rotations

countvonzero: unrelated to this PR: welcome back @dshulyak 🏆

PR #3147, opened 2022-03-18
[Merged by Bors] - tortoise: enable neutral votes

### LINKED ISSUE #3137: tortoise: enable neutral votes
Neutral votes are effectively disabled by the tortoise because of some dependencies. In order to enable them we need to:
- pass hare output to the tortoise out of order (currently we do it only in order, and if there is a gap tortoise won't know about newer hare output)
- use the latest layer according to the clock for voting (basically https://github.com/spacemeshos/go-spacemesh/issues/2919), this issue by itself is trivial but we can't implement it in a straightforward way due to https://github.com/spacemeshos/go-spacemesh/issues/3003

With neutral votes network is less likely to enter full mode.

dshulyak:
## Motivation

If some smeshers fail to reach hare consensus in time they will vote for an empty layer. It causes long-running recovery and stalls the network for hours (it takes 6-10h every single time). Neutral votes will prevent such delays.

As we cannot use layer clock in tortoise (because of this dep https://github.com/spacemeshos/go-spacemesh/issues/3003) I pass last layer as a parameter to EncodeVotes method.

closes: https://github.com/spacemeshos/go-spacemesh/issues/3137

## Changes
- remove layer organizer, as tortoise should be able to accept hare results out of order
- do not count votes from layers out of order by keeping track of decided layers
- pass current layer as an option to EncodeVotes (previously BaseBallot) method. this way it doesn't affect vote counting and the global threshold. in later iteration we will remove this option and use last layer according to the clock


dshulyak:
bors try

dshulyak:
bors try

sql/blocks/blocks.go
```diff
@@ -124,3 +124,20 @@ func IDsInLayer(db sql.Executor, lid types.LayerID) ([]types.BlockID, error) {
...
+
+// ContextualValidity returns tuples with block id and contextual validity for all blocks in the layer.
+func ContextualValidity(db sql.Executor, lid types.LayerID) ([]types.BlockContextualValidity, error) {
+	var rst []types.BlockContextualValidity
+	if _, err := db.Exec("select id, validity from blocks where layer = ?1;", func(stmt *sql.Statement) {
+		stmt.BindInt64(1, int64(lid.Uint32()))
+	}, func(stmt *sql.Statement) bool {
+		validity := types.BlockContextualValidity{}
+		stmt.ColumnBytes(0, validity.ID[:])
+		validity.Validity = stmt.ColumnInt(1) == 1
```
countvonzero: `stmt.ColumnInt(1) == valid`

tortoise/algorithm.go
```diff
@@ -177,16 +170,39 @@ func (t *Tortoise) LatestComplete() types.LayerID {
...
+// EncodeVotesOpts is for configuring EncodeVotes options.
+type EncodeVotesOpts func(*encodeConf)
+
+// EncodeVotesWithLast changes last known layer that will be used for encoding votes.
+//
+// NOTE(dshulyak) why do we need this?
+// tortoise computes threshold from last non-verified till the last known layer,
+// since we dont download atxs before starting tortoise we won't be able to compute threshold
+// based on the last clock layer (see https://github.com/spacemeshos/go-spacemesh/issues/3003)
+func EncodeVotesWithLast(last types.LayerID) EncodeVotesOpts {
```
countvonzero: so this a way to sneak in `current layer` to tortoise?

tortoise/algorithm.go
```diff
@@ -177,16 +170,39 @@ func (t *Tortoise) LatestComplete() types.LayerID {
 	return t.trtl.verified
 }
 
-// BaseBallot chooses a base ballot and creates a differences list. needs the hare results for latest layers.
-func (t *Tortoise) BaseBallot(ctx context.Context) (*types.Votes, error) {
+type encodeConf struct {
+	last *types.LayerID
```
countvonzero: i thinking `votingLayer` or `ballotLayer` is a better name than `last`

tortoise/tortoise.go
```diff
@@ -493,21 +513,21 @@ func (t *turtle) switchModes(logger log.Log) {
 	)
 }
 
-func (t *turtle) processLayer(ctx context.Context, logger log.Log, lid types.LayerID) error {
-	logger.With().Info("adding layer to the state")
+func (t *turtle) processLayer(logger log.Log, lid types.LayerID) error {
+	logger.With().Info("count layer")
```
countvonzero: rm or elaborate log msg?

tortoise/tortoise.go
```diff
@@ -476,10 +477,29 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) (types.Be
 	return beacon, nil
 }
 
-// HandleIncomingLayer processes all layer ballot votes.
-func (t *turtle) HandleIncomingLayer(ctx context.Context, lid types.LayerID) error {
+// onLayerTerminated is expected to be called when ballots for layer were received by mesh.
```
countvonzero: when hare terminated for a layer?

tortoise/tortoise.go
```diff
@@ -476,10 +477,29 @@ func (t *turtle) getBallotBeacon(ballot *types.Ballot, logger log.Log) (types.Be
...
+	t.logger.With().Info("on layer terminated", lid)
 	defer t.evict(ctx)
-	return t.processLayer(ctx, t.logger.WithContext(ctx).WithFields(lid), lid)
+	if err := t.updateLayer(t.logger, lid); err != nil {
+		return err
+	}
+	if err := t.loadConsesusData(lid); err != nil {
+		return err
+	}
+	for process := t.minprocessed.Add(1); !process.After(t.processed); process = process.Add(1) {
```
countvonzero, line 492: question: should we check for `t.minprocessed` in `evict()` just in case tortoise got stuck for a long time?
is it possible even for the t.minprocessed to fall outside the sliding window?

countvonzero (approved):
question: 
before this PR if hare terminate out of order, we will not encode those layer as `abstain`, but with this PR we will?. 

i.e. if current layer is 100, and layer 97 have not terminated, before we will only encode up to layer 96. with this PR, we will be able to encode abstain for layer 97, 98, and 99?

tortoise/tortoise.go
dshulyak: i think we still evict based on verified layer, so both processed and minprocessed won't affect sliding window

tortoise/algorithm.go
dshulyak: yes

dshulyak:
>  if current layer is 100, and layer 97 have not terminated, before we will only encode up to layer 96. with this PR, we will be able to encode abstain for layer 97, 98, and 99?

yes, exactly like that. but 98 and 99 are independent from 97, so we may abstain or support some block from those layers

tortoise/algorithm.go
dshulyak: renamed to current

dshulyak:
bors try

dshulyak:
bors merge

Issue #3094, opened 2022-01-27
Conservative State Tracking


noamnelke:
*A.K.A. Pending Transactions.*

# Motivation
The node must maintain an index of pending transactions (known, applicable transactions that have not been executed yet) in order to calculate the conservative nonce and balance of accounts that might be principle accounts. This information will be used for processing gossip transactions: if the conservative balance of the principle account cannot cover the transaction's gas cost, or the conservative account nonce indicates the transaction's nonce is invalid - the node should refrain from relaying the transaction or storing it in its mempool. Clearly, layer committee members should only include in their proposals transactions that pass nonce and balance checks against their principle account's conservative state.

## Principles of Conservative State
1. **Transactions are treated asymmetrically.** We only consider the money going out of each account and not money coming in. This makes the state "conservative". If we were to consider incoming transactions, but they would eventually fail, we could allow transactions that we shouldn't have. On the other hand, if we consider a spending transaction and it fails, the worst case is that a subsequent transaction would have to unjustifiably wait a few more layers.
2. **Transactions are maxed out for conservative balance tracking.** When gas metering is enabled in the future, the `max_gas` will be deducted from the conservative balance. In addition, part of the principle interface is supporting a `max_balance_transfer(transaction)` method and the output of that method is also deducted. If at execution time this amount is exceeded, the transaction fails and pays the `max_gas`.

# Implementation Details
The Spacemesh protocol has a weak assumption that nodes keep transactions when they are first published. Thus, Spacemesh proposals and blocks don't contain actual transactions, but only references to them. If a node is missing a referenced transaction it can request it from its peers, but if many nodes had to sync most transactions referenced in proposals during a Hare run, it would put considerable strain on the gossip network and put the Hare protocol (which is sensitive to timing) success at risk.

For this reason, mempool transactions should be persisted on-disk (we use the name "mempool" liberally).

## Storing Transactions and Maintaining Their State

Transactions should be stored in a `transactions` SQLite table. The table should contain the following fields:
- `transaction_id`: 20-byte prefix of the Sha256 sum.
- `principle_address`: The source of funds for the transaction.
- `nonce`: 128-bit nonce value of the transaction.
- `raw`: The raw serialized transaction bytes.
- `applied`: Integer indicating the layer in which this transaction has been applied to global state (or NULL).

The following indices should be defined:
- `[transaction_id]`
- `[applied, transaction_id]` (partial index where `applied IS NULL`)
- `[principle_address, nonce]`

[...]

PR #3152, opened 2022-03-25
[Merged by Bors] - conservative cache and mempool iterator

### LINKED ISSUE #3034: Mempool Ordering
# Goal
When fetching the next best transaction from the mempool, we want to fetch the transaction with the highest gas price out of the set of transactions that can be applied next.

# Solution Overview
Pending transactions are stored in the Conservative State and there's no need to maintain a separate index for the mempool. Whenever the node needs to construct a proposal, it will generate an index on-the-fly to sort transactions in the desired order. Creating the index on-the-fly allows us to reuse the validation logic from the conservative state to determine the up-to-date applicability of each pending transaction, while not needing to constantly maintain this additional structure every time the context changes.

Since (for now) every account only has at most a single transaction that can be applied next, we create a priority queue (implemented using a heap) of applicable transactions, prioritized by their gas price and then timestamp (highest gas price first, ties broken by earliest timestamp).

We either pop a transaction from the queue, if the account has no further applicable transactions, or we replace it with the next applicable transaction from the same account and then re-establish the heap order (requires only one heap ordering, instead of the two required to pop the root and then insert a new item).

# Implementation Details
### `MempoolItem`
A struct with the following fields:
- ~`TransactionID`~ In order to access the transaction in the conservative state, we actually need `Principle` and `Nonce`, we can then get the transaction ID from the conservative state, by accessing `conservative_cache[principle]['mempool'][nonce][0]` (or a method providing similar functionality).
- `GasPrice`
- `Timestamp` - when the transaction was first processed by the node (we should add this to the Conservative State)
- `PosteriorBalance` - the conservative account balance after this transaction is applied.

### `priorityQueue`
A list of `MempoolItem`s. Should implement the [Heap](https://pkg.go.dev/container/heap) interface. There's [an example](https://pkg.go.dev/container/heap#example-package-PriorityQueue) of how to implement a priority queue using this package, which is very close to what we need.

### `MempoolIterator`
A public wrapper for `priorityQueue` which abstracts away the heap implementation details. For example, it calls `heap.Push(mi.pq, item)` when its own `mi.Push(item)` is called.

### `NewMempoolIterator()`
This method should do the following:
- Initialize the heap.
- Iterate over accounts in conservative state that have transactions in the `mempool` layer.
- For each account, determine the next applicable transaction and the posterior conservative state. This requires querying the global state. Convert each transaction to a `MempoolItem` and add it to the PQ.

### `MempoolIterator.Pop()`
1. Peak at the top item in the PQ and obtain the conservative state transaction - this is the return value.
2. Does the transaction's principle account have another applicable transaction after this?
  - Yes: Fetch it from conservative state, update the `MempoolItem` and call `heap.Fix()`
  - No: Pop the item.
  - **Important note:** there's no need to query the account state at this point, so there's no disk access. The conservative state can be determined using the nonce from the transaction + the `PosteriorBalance` from the `MempoolItem`.
3. Return the transaction. 

## Concurrency
The process of iterating over the mempool assumes that the mempool transactions in the conservative state don't change mid-process, so protections must be implemented. Possible protections range from locking the entire conservative balance while constructing a proposal to locking it only during the initialization of the PQ, and creating a snapshot of each principle's `mempool` layer and using the copy to determine the following transactions in each account.

countvonzero:
## Motivation

Closes #3034
Partial implementation of #3094

## Changes

- mempool iterator spec'ed in #3034 
- add conservative cache spec'ed in #3094
a notable difference btwn the spec and the implementation is the lack of the layer index in the cache.
my first draft of implementation organized the data structure as spec'ed
```
principle
 ↳ layer
   ↳ nonce
     ↳ (transaction_id, c_amount, c_fee)
```
however, due to the fact that layer is an artificial property of the account data (used when packing a new proposal) and that  ultimately transactions need to be applied in nonce order, the extra layer index becomes redundant and cumbersome. this is because even if we gather all the transactions marked as the layer to apply, one still needs to check if the transactions are applied in nonce order.

moreover, the layer information is used onl when collecting mempool transactions into a proposal. i.e. a tx with an empty layer ID is eligible for the next proposal. it does not really matter which layer it is in, as long it is after the last applied layer. 

the implementation now uses layer as the property of each nonce. 
```
principle
   ↳ nonce (with layer as property)
     ↳ (transaction_id, c_amount, c_fee)
```
that way, we simply apply transactions in nonce order, and error out otherwise. 
when packing the next proposal, we find the first nonce that doesn't have layer attached, and pack the proposal from that nonce and up from the cache.

this is the first PR of the series. the next PR will do the integration with mesh and the database.

## Test Plan

unit tests

## TODO

## DevOps Notes


txs/cache.go
```diff
@@ -0,0 +1,671 @@
...
+	}
+}
+
+func (n *nanoTX) maxSpending() uint64 {
+	// TODO: create SVM methods to calculate these two fields
+	return n.fee + n.amount
+}
+
+func (n *nanoTX) better(other *nanoTX) bool {
+	return other == nil || n.fee > other.fee || n.fee == other.fee && n.received.Before(other.received)
```
dshulyak: why other can be nil? i think that a validation problem 

why do we track time? nodes in the cluster will receive transactions in different order, so time is not deterministic property. if we can't guarantee determinism then it doesn't matter which tx is better if the fee is the same. if we want determinism for some reason we should compare by id

txs/cache.go
```diff
@@ -0,0 +1,671 @@
...
+	best        *nanoTX
+	postBalance uint64
+	all         map[types.TransactionID]*nanoTX
+}
+
+func (s *sameNonceTXs) layer() types.LayerID {
+	return s.best.layer
+}
+
+func (s *sameNonceTXs) txs() []*nanoTX {
```
dshulyak: i would rework it without allocating array. such hidden allocations can easily become an issue

txs/cache.go
```diff
@@ -0,0 +1,671 @@
...
+func (n *nanoTX) updateLayerMaybe(lid types.LayerID) {
+	if n.layer == (types.LayerID{}) || lid.Before(n.layer) {
+		n.layer = lid
+	}
+}
+
+type sameNonceTXs struct {
+	best        *nanoTX
+	postBalance uint64
+	all         map[types.TransactionID]*nanoTX
```
dshulyak: just to clarify, the reason why we want to keep `all` and not just `best` is that the next `best` may change after tx with a previous nonce got applied? or for other reasons?

i think we should have some aggressive limit for such transactions (e.g. 2-5) since we loop over all of them after applying a layer. it looks like a viable ddos, since there is no price for including multiple transactions with the same nonce

txs/cache.go
```diff
@@ -0,0 +1,671 @@
...
+		logger.With().Error("failed to set txs discarded for applied layer", lid, log.Err(err))
+		return err
+	}
+
+	return ac.resetAfterApply(logger, tp, newNonce, newBalance, lid)
+}
+
+func (ac *accountCache) resetAfterApply(logger log.Log, tp txProvider, nextNonce, newBalance uint64, applied types.LayerID) error {
+	offset := nextNonce - ac.startNonce
+	ac.txsByNonce = ac.txsByNonce[offset:]
```
dshulyak: this is a memory leak. together with append on L123 it will cause oom

txs/cache.go
```diff
@@ -0,0 +1,671 @@
...
+	maxTXsPerAcct = 100
+)
+
+var (
+	errBadNonce            = errors.New("bad nonce")
+	errNonceTooBig         = errors.New("nonce too big")
+	errInsufficientBalance = errors.New("insufficient balance")
+	errNotInCache          = errors.New("not in cache")
+	errNotInAcctCache      = errors.New("not in account cache")
+	errNonceAppliedOOO     = errors.New("nonce applied out of order")
```
dshulyak: typo in the name?

txs/cache.go
```diff
@@ -0,0 +1,671 @@
...
+	lid types.LayerID,
+	bid types.BlockID,
+	appliedByNonce map[uint64]*nanoTX,
+) error {
+	nextNonce := ac.startNonce
+	var toApply, toDiscard []types.TransactionID
+	for len(appliedByNonce) > 0 {
+		offset := nextNonce - ac.startNonce
+		if _, ok := appliedByNonce[nextNonce]; !ok {
+			allNonce := make([]uint64, 0, len(appliedByNonce))
```
dshulyak: allocation should be moved to nonceMarshaller, so at least it will be lazily allocated only when needed

txs/store.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+func (s *store) add(tx *types.Transaction, received time.Time) error {
+	return nil
+}
+
+func (s *store) has(tid types.TransactionID) bool {
+	has, err := transactions.Has(s.db, tid)
+	return err == nil && has
```
dshulyak: i don't think that we should ignore db errors. on any unexpected db error app should not continue normal flow

txs/cache.go
```diff
@@ -0,0 +1,671 @@
...
+	if err := ac.discardOldNonce(logger, tp, nonce2TXs); err != nil {
+		return err
+	}
+
+	if !ac.hasGapInDB && len(nonce2TXs) > 0 {
+		for nonce, txs := range nonce2TXs {
+			for _, ntx := range txs {
+				logger.With().Error("unprocessed tx in batch",
+					ntx.tid,
+					log.Named("intended", ac.addr),
```
dshulyak: please use Stringer, i will remove Named eventually

txs/cache.go
countvonzero: the reason why i keep `all` is to be able to revert correctly. 

say an account has 3 competing txs (tx0, tx2, tx3) in the same nonce bucket. the same account also have 2 txs (tx1, tx4) that are rejected due to insufficient balance earlier. all 5 txs are saved in db, but only tx0, tx2, tx3 are in `all`.

P: pending
D: discarded
A: applied

before this nonce is applied in layer 9, their db state are
|  tid  |  state  |  block    |   layer
| ----- | -------- | ---------- | --------
| tx0  |      P    |  blockA |      9
| tx1  |      D    |  {}         |      {}
| tx2  |      P    |  {}         |      9
| tx3  |      P    |  blockB |      9
| tx4  |      D    |  blockC|      {}

when layer 9 / blockA is applied, their db state become
|  tid  |  state  |  block    |   layer
| ----- | -------- | ---------- | --------
| tx0  |      A    |  blockA |      9
| tx1  |      D    |  {}         |      {}
| tx2  |      D    |  {}         |      9
| tx3  |      D    |  {}         |      9
| tx4  |      D    |  {}         |      {}

note that tx2 and tx3 have layer retained, indicating that they are discarded due to layer 9 being applied.

when we want to revert back to layer 8, we then reset all txs with layer == 9.
|  tid  |  state  |  block    |   layer
| ----- | -------- | ---------- | --------
| tx0  |      P    |   {}        |      {}
| tx1  |      D    |  {}         |      {}
| tx2  |      P    |  {}         |      {}
| tx3  |      P    |  {}         |      {}
| tx4  |      D    |  {}         |      {}

note that tx1 and tx4 are still discarded. 

so the purpose of having `all` is that at the time of applying a layer, we know which txs to set discarded with layer set properly.

alternatively, all these can be removed if i add nonce to the transactions table. i have been fighting the extra field, but it looks like the code would be safer/simpler if nonce field exists. that way the db operations can base on addr+nonce. tho with this approach, i won't be able to distinguish tx1/tx4 (discarded due to insufficient balance) from tx2/tx3 (discarded due to competing tx was applied) without adding more states. 
they will all be reset to PENDING after the revert, which i think is probably more correct.

i think i'll add a nonce field in the transactions table.

txs/cache.go
countvonzero: done

txs/store.go
countvonzero: fixed

txs/cache.go
countvonzero: done. tho i wish we have `any` available here

txs/cache.go
countvonzero: was going for short-hand of OutOfOrder but apparently too unreadable. changed

txs/cache.go
countvonzero: as discussed, when a slice is re-sliced it points to the new array index and the slice's capacity is reduced. combined with the append() that potentially will reallocate an array, this approach is not a memory leak per se, but will definitely hold onto memory longer than desired.

i've made sure in code that this array is kept under `maxTXsPerAcct` currently set to 100.

btw, we discussed also to use container.List with the benefit of push front. however, i cannot directly index into the list and for that reason i stayed with the array approach.

txs/cache.go
countvonzero: removed `all` and added `nonce` in the database table (in a followup PR). the addition of nonce allowed me to keep those competing TXs in db (and not in memory) and only operate on them when the cache balance changes for an account 
- after a layer is applied
- after a better TX replaces an old TX in a nonce group

txs/cache.go
countvonzero: removed all per discussion so no longer a issue

txs/cache.go
countvonzero: > why other can be nil? i think that a validation problem

pure convenience. changed the code to check for nil before calling this.

> why do we track time? 

per discussion on slack, we allow the selection to be non-deterministic here, but will guarantee determinism at the time of generating a block.

countvonzero:
@dshulyak please take another look at the code. i an adding test to cache.go. thanks

common/types/transaction.go
```diff
@@ -213,3 +215,49 @@ type Reward struct {
 	SmesherID           NodeID
 	Coinbase            Address
 }
+
+// NanoTX represents minimal info about a transaction for the conservative cache/mempool.
+type NanoTX struct {
```
dshulyak: why it cannot be left unexported and remain in `txs` module?

common/types/transaction.go
```diff
@@ -213,3 +215,49 @@ type Reward struct {
...
+}
+
+// MaxSpending returns the maximal amount a transaction can spend.
+func (n *NanoTX) MaxSpending() uint64 {
+	// TODO: create SVM methods to calculate these two fields
+	return n.Fee + n.Amount
+}
+
+// Better returns true if this transaction takes priority than `other`.
+func (n *NanoTX) Better(other *NanoTX) bool {
```
dshulyak: i would leave this function in txs objects, it is also doesn't need to be a method

```
func pickBest(a, b *tx) *tx
```

common/types/transaction.go
```diff
@@ -213,3 +215,49 @@ type Reward struct {
...
+	return n.Fee > other.Fee || n.Fee == other.Fee && n.Received.Before(other.Received)
+}
+
+// UpdateLayerMaybe updates the layer of a transaction if it's lower than the current value.
+func (n *NanoTX) UpdateLayerMaybe(lid LayerID, bid BlockID) {
```
dshulyak: doesn't make sense as a generic method, can we leave it in txs module?

dshulyak (approved):
i would keep nanoTx unexported if possible, methods and data structure doesn't look solid

dshulyak:
thought about it, i would rename NanoTx to InternalTxHeader and refactor those methods as functions and leave them unexported iin `txs` module

common/types/transaction.go
countvonzero: there was a circular dependency in the mocks package during testing. if i keep them in the txs module, i'd have to create more sub module inside txs module. i'll do as you suggested

countvonzero:
> 

i moved NanoTX to a package inside txs. it still public but i think it's obvious not to be used by packages outside txs.
i think the code looks more readable this way.

countvonzero:
bors merge

PR #3161, opened 2022-03-30
[Merged by Bors] - remove ntp from node

dshulyak:
this pr removes code that was comparing local time with several public ntp pools. it could cause issues if in some locations some of those pools were not available, like in China. instead, we continue to use a protocol that compares the local clock with the peers clock.


countvonzero (approved):

countvonzero:
plz add description for the PR

dshulyak:
bors merge

PR #3155, opened 2022-03-28
[Merged by Bors] - systest: update README and Makefile

countvonzero:
## Motivation

add more instructions for running tests locally

## Changes

- fix Makefile to remove node selector
- update README

## Test Plan

manual testing

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #3162, opened 2022-03-31
[Merged by Bors] - fix broken makefile

countvonzero:
## Motivation

broken by https://github.com/spacemeshos/go-spacemesh/pull/3155

## Test Plan

manual testing

## DevOps Notes

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -7,7 +7,7 @@ clusters ?= 1
 size ?= 10
 level ?= debug
 bootstrap ?= 5m
-node_selector ?=
+node_selector ?= cloud.google.com/gke-nodepool=
 
 .PHONY: docker
 docker:
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -7,7 +7,7 @@ clusters ?= 1
 size ?= 10
 level ?= debug
 bootstrap ?= 5m
-node_selector ?= cloud.google.com/gke-nodepool=
+node_selector ?=
 
 .PHONY: docker
 docker:
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,8 +83,6 @@ COPY . .
 
 # And compile the project
 RUN make build
-RUN make hare
-RUN make p2p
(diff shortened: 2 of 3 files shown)


systest/Makefile
```diff
@@ -7,7 +7,7 @@ clusters ?= 1
 size ?= 10
 level ?= debug
 bootstrap ?= 5m
-node_selector ?=
+node_selector ?= cloud.google.com/gke-nodepool=
```
dshulyak: but it wont work on minikube. how does it fail?
i think there should be a way to leave node_selector empty

countvonzero: `invalid value "" for flag -node-selector:  must be formatted as key=value`

after i added this line, it started the job

countvonzero: it works on minikube after this edit on my computer.

dshulyak: > invalid value "" for flag -node-selector: must be formatted as key=value

we should fix this flag https://github.com/spacemeshos/go-spacemesh/blob/develop/systest/testcontext/string_to_string.go#L11-L23 so that it skips empty string properly

> after i added this line, it started the job

as i mentioned it will not work with minikube, since there are no nodes with such selector

dshulyak: > it works on minikube after this edit on my computer.

that's weird, it shouldn't. something else is off then

countvonzero: > that's weird, it shouldn't. something else is off then

hmmm... i mean it started the pod pod/systest-068ea83e, but doesn't seem to start any nodes...

dshulyak: > hmmm... i mean it started the pod pod/systest-068ea83e, but doesn't seem to start any nodes...

yes that makes sense. scheduler can't find any nodes with specified node_selector

countvonzero: updated PR. plz take another look. thanks

dshulyak (approved):

countvonzero:
bors merge

PR #3163, opened 2022-04-07
[Merged by Bors] - hare: delay gossip msg till after the node catches up

countvonzero:
## Motivation

when starting a new node, the log is filled with hare warning message about validation error.
the node is not synced and not participating in the hare protocol. these spam logs just impede debugging.

## Changes

only start listening to hare message when gossip sync

## Test Plan

unit tests, systests

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -160,6 +160,7 @@ type Service interface {
 // HareService is basic definition of hare algorithm service, providing consensus results for a layer.
 type HareService interface {
 	Service
+	GetHareMsgHandler() pubsub.GossipHandler
 }
 
 // TickProvider is an interface to a glopbal system clock that releases ticks on each layer.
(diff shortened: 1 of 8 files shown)


countvonzero:
bors try

countvonzero:
bors try

cmd/node/node.go
```diff
@@ -675,6 +676,10 @@ func (app *App) initServices(ctx context.Context,
 	app.host.Register(activation.AtxProtocol, pubsub.ChainGossipHandler(syncHandler, atxDB.HandleGossipAtx))
 	app.host.Register(txs.IncomingTxProtocol, pubsub.ChainGossipHandler(syncHandler, txHandler.HandleGossipTransaction))
 	app.host.Register(activation.PoetProofProtocol, poetListener.HandlePoetProofMessage)
+	hareGossipHandler := rabbit.GetHareMsgHandler()
+	if hareGossipHandler != nil {
```
dshulyak: this can be nil only because of SuperHare? i would rather exit with error or panic if hare is not registered

turbohare/hare.go
```diff
@@ -113,3 +114,8 @@ func (h *SuperHare) Start(ctx context.Context) error {
 func (h *SuperHare) Close() {
 	close(h.closeChannel)
 }
+
+// GetHareMsgHandler returns the gossip handler for hare protocol message.
+func (h *SuperHare) GetHareMsgHandler() pubsub.GossipHandler {
+	return nil
```
dshulyak: i think it is better to return noop handler instead of nil

dshulyak (approved):

cmd/node/node.go
countvonzero: yes. done

turbohare/hare.go
countvonzero: done

countvonzero:
bors merge

Issue #3139, opened 2022-02-21
hare, proposals: don't vote for proposals with missing transactions


dshulyak:
if a proposal with the transaction that is not available to any node gets into hare output it will halt the network. since we don't plan to prune transactions the simplest solution is to change this https://github.com/spacemeshos/go-spacemesh/blob/develop/layerfetcher/layers.go#L674-L688 method so that it will fail even if single transaction is not found.

kanghua309:
if only need return “err” instead of “continue” in method GetTxs in file layers？

PR #3166, opened 2022-04-11
dont accept proposals/blocks with missing transactions

kanghua309:
## Motivation

fix issue - hare, proposals: don't vote for proposals with missing transactions #3139

## Changes

if a proposal with the transaction that is not available to any node gets, it should be treated as a error and return

### DIFF
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -664,7 +664,7 @@ func (l *Logic) GetTxs(ctx context.Context, IDs []types.TransactionID) error {
 		res := <-resC
 		if res.Err != nil {
 			l.log.WithContext(ctx).With().Error("cannot find tx", log.String("hash", hash.String()), log.Err(res.Err))
-			continue
+			return res.Err
 		}
 		if !res.IsLocal {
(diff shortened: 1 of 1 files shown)


layerfetcher/layers.go
```diff
@@ -664,7 +664,7 @@ func (l *Logic) GetTxs(ctx context.Context, IDs []types.TransactionID) error {
 		res := <-resC
 		if res.Err != nil {
 			l.log.WithContext(ctx).With().Error("cannot find tx", log.String("hash", hash.String()), log.Err(res.Err))
-			continue
+			return res.Err
```
countvonzero, line 667: yes. this would be the obvious fix. but internally we haven't decided what the expected behavior should be after this method returns error, so there are a lot more work (including test changes and other component changes) when this line is changed.

dshulyak: as i remember it was only me who was against this, in order to enable pruning. but i think that it is already confirmed that there will be no pruning for transactions in genesis

dshulyak:
superseded by another change

PR #3168, opened 2022-04-12
Bugfix issue3049-prune proposals from database

kanghua309:
## Motivation

fix issue 3049 - prune proposals from database

## Changes

add remove proposals after create unify block

### DIFF
--- a/hare/hare.go
+++ b/hare/hare.go
@@ -241,6 +241,9 @@ func (h *Hare) collectOutput(ctx context.Context, output TerminationOutput) erro
 			return fmt.Errorf("hare save block: %w", err)
 		} else {
 			hareOutput = block.ID()
+			if err = h.pdb.DelProposals(pids); err != nil {
+				h.WithContext(ctx).With().Warning("failed to remove proposals from database", log.Err(err))
+			}
 		}
 	}
 	if err := h.mesh.ProcessLayerPerHareOutput(ctx, layerID, hareOutput); err != nil {
--- a/hare/hare_test.go
+++ b/hare/hare_test.go
@@ -155,6 +155,11 @@ func TestHare_collectOutputAndGetResult(t *testing.T) {
 			assert.ElementsMatch(t, proposalIDs, pids)
 			return proposals, nil
 		}).Times(1)
+	h.mockProposalDB.EXPECT().DelProposals(gomock.Any()).DoAndReturn(
+		func(pids []types.ProposalID) error {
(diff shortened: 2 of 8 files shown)


proposals/proposaldb.go
```diff
@@ -57,6 +57,15 @@ func (db *DB) AddProposal(ctx context.Context, p *types.Proposal) error {
 	return nil
 }
 
+// DelProposal dels a proposal from the database
+func (db *DB) DelProposal(ctx context.Context, id types.ProposalID) error {
```
countvonzero: can you just name it `DeleteProposal`? thanks

sql/proposals/proposals.go
```diff
@@ -159,6 +159,20 @@ func Add(db sql.Executor, proposal *types.Proposal) error {
 	return nil
 }
 
+// Del removes a proposal for a given ID.
+func Del(db sql.Executor, id types.ProposalID) error {
```
countvonzero, line 163: :s/Del/Delete

hare/hare.go
```diff
@@ -241,6 +241,9 @@ func (h *Hare) collectOutput(ctx context.Context, output TerminationOutput) erro
 			return fmt.Errorf("hare save block: %w", err)
 		} else {
 			hareOutput = block.ID()
+			if err = h.pdb.DelProposals(pids); err != nil {
```
countvonzero, line 244: we should not delete it right away for two reasons
- some peers may be trying to fetch proposals from the node
- debugging

perhaps for now we can keep a grace period say one epoch?

hare/hare_test.go
```diff
@@ -478,6 +504,7 @@ func TestHare_onTick_NoGoodBallots(t *testing.T) {
 
 	lyrID := types.GetEffectiveGenesis().Add(1)
 	beacon := types.RandomBeacon()
+	time.Sleep(1 * time.Millisecond)
```
countvonzero, line 507: what's the need to sleep here?

proposals/proposaldb.go
```diff
@@ -57,6 +57,28 @@ func (db *DB) AddProposal(ctx context.Context, p *types.Proposal) error {
 	return nil
 }
 
+// DelProposal dels a proposal from the database
+func (db *DB) DelProposal(id types.ProposalID) error {
```
countvonzero, line 61: no need to have two API here. just do this inside `DeleteProposals`

Issue #3164, opened 2022-04-08
sync: data sync keeps going while validation stuck


countvonzero:
## Description

at the beginning of every epoch, there is a period of time when beacon is unavailable (until enough ballots are seen by the node and beacon value determined)
during this time, syncer will see that beacon is not available and will not request tortoise to verify a layer.

but the problem is: syncer uses two goroutines
goroutine 0: sync data from processedLayer (say 100) -> currentLayer-1 (say 10_000)
goroutine 1: request tortoise to verify layer

say at the beginning of epoch 3 (layer 120) the above beacon check stops goroutine1 but not goroutine0. so syncer will end up syncing data all the way from 120 -> 10_000 (while verified layer get stuck at 119) for goroutine0 to exit, and allow the next syncer run to proceed.

this is wildly inefficient.

## Steps to reproduce
1. starts a new node and one can see log message from syncer `skip validating layer: beacon not available`

## Actual Behavior

see above

## Expected Behavior

verifying a layer should be able to progress after the beacon value is available for that epoch.

kanghua309:
"wildly inefficient"  refer to sync duplicated layer content ？

********** below logic ？****** 
for layerID := s.mesh.ProcessedLayer().Add(1); layerID.Before(s.ticker.GetCurrentLayer()); layerID = layerID.Add(1) { 	
     if layerID.After(lastSynced) { 		
          lastSynced = layerID 		
          if err := s.syncLayer(ctx, layerID); err != nil { 	
		return false 		} 	
    } 	
    vQueue <- layerID 	logger.With().Debug("finished data sync", layerID) 
}

countvonzero:
> "wildly inefficient" refer to sync duplicated layer content?

it means we need to wait for all layer data to be synced (120 -> 10_000) before we have a chance to retry verifying a layer that gets stuck earlier (120) due to lack of beacon, since beacon should be available after a few more layers are synced for each epoch.

PR #3169, opened 2022-04-13
Bugfix issue3164

kanghua309:
## Motivation

try to fix “sync: data sync keeps going while validation stuck #3164”

## Changes

The main idea of modification is: when beacon delay occurs, to rerun validate process from the layer of delayed beacon! key point is ：rerun check routine is triggered in the synchronization layer loop 

## Test Plan

add unit test TestSynchronize_BeaconDelayRerun for the this issue scenario

### DIFF
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -664,7 +664,7 @@ func (l *Logic) GetTxs(ctx context.Context, IDs []types.TransactionID) error {
 		res := <-resC
 		if res.Err != nil {
 			l.log.WithContext(ctx).With().Error("cannot find tx", log.String("hash", hash.String()), log.Err(res.Err))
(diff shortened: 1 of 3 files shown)


countvonzero:
instead of adding another channel, i think a simpler approach would be to get rid of vQueue in each run and have a (maybe dedicated) goroutine continuously try to validate layers from `s.mesh.ProcessedLayer()` to the last synced layer. this will also address the TODO in https://github.com/spacemeshos/go-spacemesh/blob/bf4f2b0b13eec5adf0f08e53ee9f5fd7e35bb4bb/syncer/syncer.go#L563

countvonzero:
https://github.com/spacemeshos/go-spacemesh/issues/3164 is closed

Issue #2952, opened 2021-11-17
beacon sync: should determine beacon when all blocks from epoch are seen  


countvonzero:
## Description

as a followup to issue https://github.com/spacemeshos/go-spacemesh/issues/2938
(https://github.com/spacemeshos/go-spacemesh/pull/2951)

in the joiner's case, we use a param `tortoise-beacon-sync-num-blocks` to control the number of blocks we've collect beacon values from to determine the beacon value for the previous epoch. on the odd chance that there aren't enough blocks in the epoch, we need to determine the beacon value when all layers in the epoch is data-synced.

## Actual Behavior

when there aren't enough blocks in the epoch, beacon value will not be determined

## Expected Behavior

when the node has seen all blocks from an epoch, it should determine the beacon value despite not reaching `tortoise-beacon-sync-num-blocks`

PR #3170, opened 2022-04-14
fix issue "beacon sync: should determine beacon when all blocks from …

kanghua309:
…epoch are seen #2952"

## Motivation

fix issue “beacon sync: should determine beacon when all blocks from epoch are seen #2952”

## Changes

 when the node has seen all blocks from an epoch(i.e all layers in the epoch is data-synced),it should determine the beacon value despite not reaching beacon-sync-num-blocks
noted. modify findMostWeightedBeaconForEpoch to add ablove logic

## Test Plan

add unit test TestBeacon_findMostWeightedBeaconForEpoch_NotEnoughBlocks_ButSync

### DIFF
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/spacemeshos/go-spacemesh/sql/layers"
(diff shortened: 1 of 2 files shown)


kanghua309:
fix some bug

PR #3171, opened 2022-04-14
[Merged by Bors] - tx selection: conservative state tracking

### LINKED ISSUE #3094: Conservative State Tracking

countvonzero:
## Motivation

Closes #3094 

## Changes

- conservative state saves all tx in db
  - new table that captures relationship btwn proposals and txs
  - new table that captures relationship btwn blocks and txs
- update conservative state to use conservative cache

## Test Plan

unit test

## DevOps Notes


api/grpcserver/mesh_service.go
```diff
@@ -323,8 +323,8 @@ func (s MeshService) readLayer(ctx context.Context, layerID types.LayerID, layer
 		}
 
 		var pbTxs []*pb.Transaction
```
evt:
```suggestion
		pbTxs := make([]*pb.Transaction, 0, len(mtxs))
```

countvonzero: done. thanks

countvonzero:
bors try

dshulyak:
@countvonzero i am confused why we are using heap. in previous pr i thought that we will maintain sorted data structure (e.g. it won't be ephemeral), but in this pr heap is still completely ephemeral. after it is created, we insert one transaction from every account, pop sorted transactions and garbage collect the heap. so basically heap here is just a way to sort transactions. it will be definitely more efficient to use a straighforward approach with a transactions slice.

UPD: missed some additional code in the pop, i guess it makes sense now

also in some version of the previous pr many transactions for the same nonce were kept in memory, i guess i missed that they were removed. can you clarify why?

countvonzero:
> @countvonzero i am confused why we are using heap. in previous pr i thought that we will maintain sorted data structure (e.g. it won't be ephemeral), but in this pr heap is still completely ephemeral. after it is created, we insert one transaction from every account, pop sorted transactions and garbage collect the heap. so basically heap here is just a way to sort transactions. it will be definitely more efficient to use a straighforward approach with a transactions slice.
> 
> UPD: missed some additional code in the pop, i guess it makes sense now

yeah. thanks for the update. it's hard to guarantee total order with tx slice, as within each account we also need to maintain nonce ordering.

note on GC. because the heap and the cache use the same set of NanoTX objects, the GC on txs will happen when txs are applied.
 
> also in some version of the previous pr many transactions for the same nonce were kept in memory, i guess i missed that they were removed. can you clarify why?

this is about correctness. the conservative cache only counts outgoing spending and not incoming funds. therefore, whenever a layer is applied, it's possible that a previously TX rejected due to insufficient balance is viable again. keeping all same-nonce TXs in memory become problematic. 

for example, layer 99 starts, account A has next nonce 3 and balance 10, and 3 txs at nonce 3 in the mempool.
tx0: fee: 2, amount 5
tx1: fee: 3, amount 8
tx2: fee: 1, amount 2

in previous implementation i keep all same-nonce txs in cache, with a best set. in this case, tx1 will be rejected (and not kept in cache) while tx0 and tx1 are, tx0 being the best.

when layer 99 is applied, account A gets incoming fund of 3. so after layer 99 is applied, account A's new balance is 13. now tx1 is viable but it's not in cache. to correct this, the code needs to query DB again for any previously rejected TXs for account A at nonce 3 after layer 99 is applied. there is also potential DOS that you pointed out.

since i need to query DB again whenever a layer is applied (and potentially changes an account's conservative balance), i changed it to only keeping the best TX in the cache for any given nonce. that way, we are sure not to miss any previously rejected TXs that are made possible with the new balance.

dshulyak (approved):
we need to adjust https://github.com/spacemeshos/go-spacemesh/blob/develop/systest/tests/transactions_test.go#L21 if current version of the mempool supports multiple concurrent transactions. previously more than 1 always failed

countvonzero:
bors try

countvonzero:
> we need to adjust https://github.com/spacemeshos/go-spacemesh/blob/develop/systest/tests/transactions_test.go#L21 if current version of the mempool supports multiple concurrent transactions. previously more than 1 always failed

changed to 3 and fixed bugs revealed by that.

countvonzero:
bors merge

PR #3178, opened 2022-04-18
support dns addresses

### LINKED ISSUE #3156: p2p: support dns entrypoints
Libp2p accepts DNS entrypoint for the node. Example:

```
/dns4/bootstrap-0.mainnet.filops.net/tcp/1347/p2p/12D3KooWCVe8MmsEMes2FzgTpt9fXtmCY7wrq91GRiaC8PHSCCBj
```

In the existing deployment scheme we create several nodes and then connect every other node to them. It makes deployment more complex, and more importantly, relies on IP addresses being the same all the time.

Instead, in the new deployment scheme, we will generate entrypoints in advance and share them for every node in the cluster.

abergasov:
## Motivation
Now for bootnodes can be used links with dns addresses 
Closes #3156

## TODO

## DevOps Notes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/libp2p/go-tcp-transport v0.2.8
 	github.com/mitchellh/mapstructure v1.4.1
 	github.com/multiformats/go-multiaddr v0.4.0
+	github.com/multiformats/go-multiaddr-dns v0.3.1
 	github.com/nullstyle/go-xdr v0.0.0-20180726165426-f4c839f75077
 	github.com/prometheus/client_golang v1.11.0
 	github.com/prometheus/common v0.30.0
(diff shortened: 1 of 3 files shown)


dshulyak:
i thought that it should be possible to pass dns entries directly to libp2p [host](https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/peerexchange/crawler.go#L90). and it will resolve them internally when connection is initialized. in general we have to deal with two shortcomings:
- host should be able to pickup a change in dns to ip address mapping
- if dns name is not resolvable when host is started we shouldn't crush the node, it would be better to resolve it in the background when actual connection is made

both of them will be solved if host actually capable of resolving dns entries when connection is made

abergasov:
> i thought that it should be possible to pass dns entries directly to libp2p [host](https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/peerexchange/crawler.go#L90). and it will resolve them internally when connection is initialized. in general we have to deal with two shortcomings:
> 
> * host should be able to pickup a change in dns to ip address mapping
> * if dns name is not resolvable when host is started we shouldn't crush the node, it would be better to resolve it in the background when actual connection is made
> 
> both of them will be solved if host actually capable of resolving dns entries when connection is made

`err = r.host.Connect(ctx, *ainfo)` will work with dns string, so it do not need changes. 
But app now if fail with start, if in bootNodes not an ip address. So we need or skip it, if string has prefix `dns` or resolve dns as i wrote in pr

dshulyak:
it means that we need to make changes in the spacemesh app, so that it will be ready to accept dns

abergasov:
> it means that we need to make changes in the spacemesh app, so that it will be ready to accept dns

now it ready accept dns)

dshulyak:
as explained above this is not a solution

PR #3183, opened 2022-04-21
[Merged by Bors] - limit parallel systest jobs to 1

dshulyak:


countvonzero:
@dshulyak do you think two instances of CI systest interfere with each other?

dshulyak:
@countvonzero yes, they are deployed on the same cluster and are using the same resources pool (cpu and memory)

countvonzero (approved):

dshulyak:
bors merge

PR #3187, opened 2022-04-27
[Merged by Bors] - systest: allow to reuse cluster for multiple independent tests

### LINKED ISSUE #3186: systest: make cluster reusable
part of https://github.com/spacemeshos/go-spacemesh/issues/3172

the end goal is to allow isolated tests to reuse already deployed cluster.

## make cluster reusable

cluster setup must be able to recover cluster configuration from the deployed cluster

also tests should not use hardcoded layer. all expectations must be set relatively to the start of the test (e.g. +X layers to the current layer, and not genesis.).

### endpoints

recover from the deployed pod

### configuration

recover from spacemesh API

### private keys with funds

persist them with configmap, and recover from configmap

### nonce

recover nonce from API

dshulyak:
closes https://github.com/spacemeshos/go-spacemesh/issues/3186

this change allows to reuse same cluster for multiple independent test runs. account keys with funds and configuration are persisted using config map. all other information is recovered from pods info.

not all tests can benefit from cluster re-usability. if test requires specific cluster configuration or creates irrecoverable issues such it won't be possible to reuse cluster after such test. other tests have to be implemented using relative layers, e.g. in terms of difference with a current layer.

 for regular use cases it basically removes bootstrap time, so all additional tests won't have to wait that time and can start immediately:

`make run test_name=TestShortTimeskew size=20 bootstrap=1m keep=true`

It will print namespace address and after that you can re-use it as:

`make run test_name=TestShortTimeskew size=20 bootstrap=0m keep=true namespace=test-yyyi`


dshulyak:
bors try

systest/cluster/cluster.go
```diff
@@ -43,6 +49,19 @@ func WithKeys(n int) Opt {
 	}
 }
 
+// Reuse will try to recover cluster from the given namespace, if not found
+// it will create a new one.
```
countvonzero, line 54: what does reuse entail? i see you saved accounts in the configmap. how about mesh state?

dshulyak: reuse means that it will recover (rediscover) existing cluster, for example that was deployed as a part of the previous test run

systest/cluster/cluster.go
```diff
@@ -43,6 +50,19 @@ func WithKeys(n int) Opt {
...
+func Reuse(cctx *testcontext.Context, opts ...Opt) (*Cluster, error) {
+	cl := New(cctx, opts...)
+	if err := cl.reuse(cctx); err != nil {
+		if errors.Is(err, errNotInitialized) {
+			return Default(cctx, opts...)
+		}
+		return nil, err
```
evt, line 61: are we supposed to use `fmt.Errorf` instead of returning plain errors to the caller?

dshulyak: there is no rule to wrap them blindly, i am in general wrapping errors on the module boundary. for example cluster.reuse -> accounts.Recover -> wrapped error

systest/cluster/cluster.go
```diff
@@ -86,10 +107,71 @@ type Cluster struct {
...
+	flags, err := recoverFlags(ctx)
+	if err != nil {
+		return err
+	}
+	c.smesherFlags[genesisTimeFlag] = flags[genesisTimeFlag]
+	c.smesherFlags[accountsFlag] = flags[accountsFlag]
+	if !reflect.DeepEqual(c.smesherFlags, flags) {
+		return fmt.Errorf("configuration doesn't match %+v != %+v", c.smesherFlags, flags)
+	}
+	c.persistedFlags = true
```
evt, line 139: does it mean we can't persist again after recover? 🤔

dshulyak: why would it be needed?

evt: can't think of any real use case :) ok cheers

evt (approved):

systest/tests/smeshing_test.go
```diff
@@ -31,7 +31,13 @@ func TestSmeshing(t *testing.T) {
 func testSmeshing(t *testing.T, tctx *testcontext.Context, cl *cluster.Cluster) {
 	const limit = 15
 
-	createdch := make(chan *spacemeshv1.Proposal, cl.Total()*limit)
+	first := currentLayer(tctx, t, cl.Client(0))
+	// TODO fetch epoch size from API
+	first = nextFirstLayer(first, 4)
```
evt, line 36: magic number?

dshulyak: it is a number o layers in the epoch, need to add an api to fetch it from the node

dshulyak:
bors try

systest/cluster/cluster.go
```diff
@@ -248,3 +378,29 @@ func defaultTargetOutbound(size int) int {
...
+	_, err := ctx.Client.CoreV1().ConfigMaps(ctx.Namespace).Apply(ctx, cfgmap, metav1.ApplyOptions{FieldManager: "test"})
+	if err != nil {
+		return fmt.Errorf("failed to persist accounts %+v %w", data, err)
+	}
+	return nil
+}
+
+func recoverFlags(ctx *testcontext.Context) (map[string]DeploymentFlag, error) {
+	flags := map[string]DeploymentFlag{}
+	cfgmap, err := ctx.Client.CoreV1().ConfigMaps(ctx.Namespace).Get(ctx, "flags", metav1.GetOptions{})
```
countvonzero, line 398: curious: metav1.GetOtions{} doesn't to match metav1.ApplyOptions{}. is "test" the default value?

systest/tests/smeshing_test.go
```diff
@@ -15,7 +15,7 @@ import (
 
 func TestSmeshing(t *testing.T) {
 	tctx := testcontext.New(t, testcontext.Labels("sanity"))
-	cl, err := cluster.Default(tctx, cluster.WithKeys(10))
+	cl, err := cluster.Reuse(tctx, cluster.WithKeys(10))
```
countvonzero, line 18: how does this affect test running locally?

systest/cluster/cluster.go
```diff
@@ -86,10 +107,71 @@ type Cluster struct {
 	poets     []string
 }
 
+func (c *Cluster) persist(ctx *testcontext.Context) error {
+	if err := c.accounts.Persist(ctx); err != nil {
+		return err
+	}
+	return c.persistFlags(ctx)
```
countvonzero, line 114: how about the config file? say if i were testing a change to config/config.go or config/preset?

dshulyak: configuration is provided using flags. what about config file?

systest/tests/smeshing_test.go
dshulyak: it shouldnt, unless you want to reuse cluster

systest/cluster/cluster.go
dshulyak: FieldManager needs to be set when object is created, it is not relevant for GetOptions

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #3143, opened 2022-02-25
[Merged by Bors] - POET: do not re-validate a existing proof

countvonzero:
## Motivation

skip expensive validation for existing proof

## Changes

- add a sql existence lookup by the proof reference
- ignore the p2p msg if the proof already exists 
- skip validating PoET proof with empty membership

## Test Plan

unit test

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #3191, opened 2022-04-29
PoET: skip validating PoET proof with empty membership

### LINKED ISSUE #3153: PoET: skip validating PoET proof if it already exists
## Description

we validate every syntactically correct PoET proof regardless whether it's already present locally.

from Noam:
```
Validating a PoET proof is definitely more expensive than calculating a Sha256 sum of the proof.
It involves calculating many smaller Sha256 sums (I think the lower bound for the number of
hashed bytes is proof_size*log(number_of_poet_iterations)), as well hashes for the leaves. 
So the number of hashed bytes for validation is a multiple of the proof, likely 10x or more.
```

## Actual Behavior

validate regardless

## Expected Behavior

only validate when needed

## Additional Resources 

i made an attempt with https://github.com/spacemeshos/go-spacemesh/pull/3143 and turned out that apptest failed because rejecting duplicate msg is considered as a p2p publish failure.

evt:
## Motivation
Closes #3153 

## Changes
Skip validating PoET proof with empty membership

## DevOps Notes

### DIFF
--- a/activation/poetdb.go
+++ b/activation/poetdb.go
@@ -64,6 +64,10 @@ func (db *PoetDb) Validate(proof types.PoetProof, poetID []byte, roundID string,
 		return types.ProcessingError(fmt.Sprintf("invalid poet id %x", poetID))
 	}
 	root, err := calcRoot(proof.Members)
+	// we shouldn't care about poet proof with empty membership as it's not relevant.
+	if len(proof.Members) == 0 {
+		return nil
+	}
 	if err != nil {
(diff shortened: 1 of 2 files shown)


countvonzero:
this still doesn't quite close https://github.com/spacemeshos/go-spacemesh/issues/3153
it only skips poet proof with empty membership, but we are still validating proof that already exist in db if we get the same poetproof gossip multiple times. the two PRs need to be combined.

evt:
Merged into https://github.com/spacemeshos/go-spacemesh/pull/3143, thank you @countvonzero

PR #3192, opened 2022-05-02
[Merged by Bors] - fix flaky TransactionService test

countvonzero:
## Motivation

fixed flaky TransactionService test caused by https://github.com/spacemeshos/go-spacemesh/pull/3190

## Changes

this test becomes flaky after the said PR. the root cause is that once syncer starts syncing 
(pointless since it's a single node test), it will take the node out of sync and stop listening to gossip.

since we are testing TransactionService, there is not point for the syncer to start. set the interval long
enough so it doesn't.

## Test Plan

unit tests

## DevOps Notes


dshulyak (approved):

countvonzero:
bors merge

Issue #1798, opened 2020-02-11
Fees are distributed to miners even if the transaction can't be applied


noamnelke:
When calculating miners' rewards, fees from all transactions in contextually valid blocks are summed up and distributed to miners. Some transactions may not end up being executed (due to wrong nonce, insufficient balance or even non-existing account), but the fee in them still gets distributed.

This can be trivially abused to inflate the money supply.

avive:
We have lots of work on this area and the currently implemented solution is obviously lacking in many ways and you didn't even mention smart contract transactions. Even with the restricted set of smart contract transactions - we only know the fee range in advance, but we can't compute how much the tx fee will cost be in practice until we execute them on the current global state.

In my mind this is one of the core cryptocurrency design issues that we must address in the coming weeks that is not resolved yet. 

Specifically - I can't see how tx fees can be distributed separately form the state transition function itself. In my view, we can only distribute tx fees to smesher for a layer after the state transition function ran on that layer and determined which coin and smart contract transactions are valid, and how much gas was actually spent in each smart contract transaction. So, the tx fee reward is just computed in the wrong time and place in the code. It should be applied to a layer after the STF was actually executed and the STF should sum the tx fees for that layer based on each tx execution.
 
@lrettig @barakshani @tal-m @iddo333

lrettig:
> we only know the fee range in advance, but we can't compute how much the tx fee will cost be in practice until we execute them on the current global state

This is troubling. Why not?

> I can't see how tx fees can be distributed separately form the state transition function itself

Agree here - this seems dangerous to me for a bunch of reasons. Maybe the solution is to create some sort of "holding account" that collects all of the fees in a given layer, as the transactions in that layer are applied to the global state. Then, "at the end" of that layer, the contents of that "holding account" are distributed to the miners, in some sort of second, independent state transition.

avive:
> This is troubling. Why not?

The halting problemo :-)

lrettig:
> The halting problem

We can't know for sure, but we _can_ "estimate." You can't know for sure in Ethereum for the same reason, but wallets like Metamask still calculate an estimated gas cost (see [estimategas](https://web3js.readthedocs.io/en/v1.2.6/web3-eth.html#estimategas)).

Also, this is much less of an issue if the language is decidable like [clarity](https://docs.blockstack.org/core/smart/overview.html). Even if the entire language isn't, individual hardcoded smart contracts could be.

avive:
[...]

PR #3194, opened 2022-05-04
[Merged by Bors] - vm: flat accounts storage

### LINKED ISSUE #3173: state trie implementation plan for genesis
## use cases

- access to the account storage, including historical access

it has to be very efficient as it will be used extensively in the hot path of the consensus, and also can be used from API layer (example get erc20 balance).

historical access means that some clients will want to have access not only to the most recent state, but also to some state in the past.

- state root computation

computed once per layer when new block is applied. initial implementation doesn't have to be super efficient, as there will be a window in seconds before this root is needed for consensus.

## account storage

account state will be stored natively in the suitable key/value storage. initial implementation will use sqlite, per my estimation it will be performant and since all state will be a part of the same db it simplifies consistency.

primary index for the account is an address and a layer when state was updated, sorted by the layer in descending order. layer is encoded in big endian. with this key we will provide efficient native access to the state: to find a state that was effective at layer X we will query for a first record with layer equal or below layer X. this is optimal as it will require a single query in any key/value storage (b+tree (sqlite), lsm (leveldb), or lmdb variant).

what is stored?
- nonce
- balance
- state
- state hash
- code (do we need this field for genesis?)
- code hash (computed from code)
- hash (computed from nonce/balance/storage hash/code hash)

the hash from account data will be used in the state trie.

in ethereum state is stored not as a simple binary blob. instead every account state is a separate patricia trie, where leafs are variables (see part 6.5 [eth data structures](https://arxiv.org/pdf/2108.05513.pdf)). it makes reads/writes/hashing on updates to the account state much more efficient. we need to make a decision if we need it before genesis:

- this scheme will affect state root
- what will be the size of the precompiles? 10kb, 100kb, 10mb? at what point it will be infeasible to rehash the state at every update?

## immutable data structure

we will use sparse merkle tree (see [diem components 2020](https://diem-developers-components.netlify.app/papers/the-diem-blockchain/2020-05-26.pdf) 4.3) for details. 

### state tree

tree state will be persisted on disk. inner nodes will store records with the following information:
- autoincremented integer (primary index)
- hash (used to compute updates)
- insert_layer (index used for revert)
- obsolete_layer (index used for pruning)
- left/right pointers (type of the child and integer that points)

left/right pointers should have enough information in order to load the node from disk. leafs are stored in the separate table (as discussed above).

pointer to the root should be stored in the table with all layers, in order to provide mapping from layer to the root.

## revert

revert N last applied layers, and applies new afterwards. technically on revert we will remove records from account state and inner nodes from state tree that are inserted after or in the reverted layer.

one way to implement it efficiently is to have an index for every row with a layer when this record was inserted. it will increase used space, but it can be evaluated after initial implementation.

## pruning

pruning removes historical data that is not needed for a particular client.

in practice it means removing records that were updated at a later layers. for example if current is 100 and we keep only state 20 last layers. we need to prune every record with layer that became obsolete before or in layer 80.

one way to implement it efficiently is to have an index for every record with a layer when such record became obsolete. again, it will increased used space, but this is simple and efficient implementation that can be evaluated and improved later.

## links

- [erigon state. youtube video](https://www.youtube.com/watch?v=aAZoiJIQiTE)
- [diem components 2020](https://diem-developers-components.netlify.app/papers/the-diem-blockchain/2020-05-26.pdf)
- [eth data structures](https://arxiv.org/pdf/2108.05513.pdf)

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3173

- removes geth trie and most of the svm code
- renames svm to vm
- added accounts table, that stores every account modification
- revert is supported by removing modifications that were made after specific layer
- interface changed so that disk errors are not ignored
- pruning of the historical account data will be enabled in the separate pr


dshulyak:
bors try

sql/layers/layers.go
```diff
@@ -95,6 +95,55 @@ func UnsetAppliedFrom(db sql.Executor, lid types.LayerID) error {
 	return nil
 }
 
+// UpdateStateHash for the layer.
+func UpdateStateHash(db sql.Executor, lid types.LayerID, hash types.Hash32) error {
+	if _, err := db.Exec(`insert into layers (id, state_hash) values (?1, ?2)
```
countvonzero, line 100: this is slightly confusing. this implies that we may update state hash before we set the applied_block?
i.e. if we insert a new row here, this is a programming error.
what do you think just do `update layers set state_hash where id = lid`?

sql/layers/layers.go
```diff
@@ -95,6 +95,55 @@ func UnsetAppliedFrom(db sql.Executor, lid types.LayerID) error {
...
+			stmt.BindBytes(2, hash[:])
+		}, nil); err != nil {
+		return fmt.Errorf("set applied %s: %w", lid, err)
+	}
+	return nil
+}
+
+// GetLatestStateHash loads latest state hash.
+func GetLatestStateHash(db sql.Executor) (rst types.Hash32, err error) {
+	if rows, err := db.Exec("select state_hash from layers where state_hash is not null;",
```
countvonzero, line 113: why isn't this querying the max layer id where state_hash is not null?

vm/changes.go
```diff
@@ -0,0 +1,133 @@
...
+		layer:   layer,
+	}
+}
+
+type changes struct {
+	log log.Log
+	db  *sql.Database
+
+	layer   types.LayerID
+	order   list.List
```
countvonzero, line 31: what is this trying to capture? please add some comment

countvonzero (approved):

sql/layers/layers.go
dshulyak: i guess it would make sense to update block and hash at the same time, but lets switch to update for now

sql/layers/layers.go
dshulyak: layers are stored sorted by descending order, so we just go down and pick first where state_hash is not empty

sql/layers/layers.go
dshulyak: actually let me keep it like this in this change. i will shortly start working on https://github.com/spacemeshos/go-spacemesh/issues/1798 , and i will try to update state hash and block for the layer at the same time

dshulyak:
bors merge

dshulyak:
bors try

sql/layers/layers.go
countvonzero: on second thoughts, i might be safer to explicitly choose the max in case of a schema change in the future?

dshulyak: i think it is correct to rely on the order, moreover i dont know if adding a condition will result in the full scan or just pick the top layer. if it relies on the scan it will be a problem very fast. if someone accidentally changes schema there will be a test failure

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3198, opened 2022-05-08
[Merged by Bors] - signing: use ECVRF-EDWARDS25519-SHA512-ELL2 for vrf

### LINKED ISSUE #2269: NodeID should contain a single ed25519 key
## Description
Right now, NodeID contains two keys, an edwards public key used for signing and a BLS-derived key used for VRF. We want to collapse these down to a single ed25519 key.

In future we may want to allow the node operator to rotate the signing key, but for now we're only allowing a single key that cannot be changed.

Replace BLS with a library that implements the scheme described [here](https://community.spacemesh.io/t/research-agenda-2020-08-10/104/5). There's more background info in [this thread](https://community.spacemesh.io/t/need-for-two-keys-in-nodeid/139).

## Affected code
- Initialization code in cmd/node.go
- NodeID struct and all code that uses it
- SmesherID (derived from NodeID) and all code that uses it

This issue appears in commit hash: 35432834b1bcc8aa4110e033c9d4e2e9701bc7d7

## Related issues

Replaces #1391 
Obviates #936

dshulyak:
closes https://github.com/spacemeshos/go-spacemesh/issues/2269

- use ecvrf implementation from https://github.com/oasisprotocol/curve25519-voi . it changes vrf size to 80 bytes
- change types.NodeID to 32 bytes array


dshulyak:
bors try

beacon/beacon.go
```diff
@@ -898,7 +898,7 @@ func atxThresholdFraction(kappa uint64, q *big.Rat, epochWeight uint64) *big.Flo
 
 // TODO(nkryuchkov): Consider having a generic function for probabilities.
 func atxThreshold(kappa uint64, q *big.Rat, epochWeight uint64) *big.Int {
-	const signatureLength = 64 * 8
+	const signatureLength = 80 * 8
```
countvonzero: use a named constants for 80

countvonzero (approved):

dshulyak:
bors merge

noamnelke:

Yawning:
FYI: https://github.com/oasisprotocol/curve25519-voi/commit/2879729f5ee0d5166b6b288b9e3025aa775e7911

Sorry for the inconvenience, the newer draft(s) that post-date when ECVRF support was added, changed the semantics for pi generation.  The  new `_v10` methods can be used to maintain exact backward compatible behavior.

PR #3201, opened 2022-05-09
[Merged by Bors] - p2p periodically check peers from addressBook for dead|alive 

### LINKED ISSUE #3132: p2p: stale records in the address book
There are several problems with address book that we are using:

1. Peers that are long gone are not removed from the address book.
2. If local address changes (for example due to nat update) we do not notify our peers in the network about it.

Both issues can be handled by periodically exchanging discovery data with N peers from the address book. If peers doesn't respond for T attempts it must be removed from the local address book. Peer is expected to update local addresses based on the received data, but this needs to be tested.

Additionally, address book code was copied from somewhere, and then it got borked. Many original features are not used and the code itself doesn't look good. Would be nice to rewrite it properly before mainnet.

abergasov:
## Motivation
Closes #3132

## Changes
added method which in ticker take random X addresses and sen request for it

## Test Plan
unit tests

## DevOps Notes


p2p/nodechecker_test.go
```diff
@@ -0,0 +1,212 @@
...
+			expected: []int{oldAddr},
+			block:    []int{newAddr},
+		},
+		{
+			title:    "change local address of node and all are acceptable",
+			expected: []int{newAddr, oldAddr},
+			block:    []int{},
+		},
+	}
+	for _, testCase := range table {
```
dshulyak: testCase needs to be moved out of loop if t.Parallel is used, otherwise it will run the same test case but print it with a different name.

```
=== RUN   TestHost_LocalAddressChange
=== RUN   TestHost_LocalAddressChange/change_local_address_of_node_and_all_not_acceptable
=== PAUSE TestHost_LocalAddressChange/change_local_address_of_node_and_all_not_acceptable
=== RUN   TestHost_LocalAddressChange/change_local_address_of_node_and_old_not_acceptable
=== PAUSE TestHost_LocalAddressChange/change_local_address_of_node_and_old_not_acceptable
=== RUN   TestHost_LocalAddressChange/change_local_address_of_node_and_new_not_acceptable
=== PAUSE TestHost_LocalAddressChange/change_local_address_of_node_and_new_not_acceptable
=== RUN   TestHost_LocalAddressChange/change_local_address_of_node_and_all_are_acceptable
=== PAUSE TestHost_LocalAddressChange/change_local_address_of_node_and_all_are_acceptable
=== CONT  TestHost_LocalAddressChange/change_local_address_of_node_and_all_not_acceptable
ACTUAL TEST CASE: change local address of node and all are acceptable
=== CONT  TestHost_LocalAddressChange/change_local_address_of_node_and_new_not_acceptable
ACTUAL TEST CASE: change local address of node and all are acceptable
=== CONT  TestHost_LocalAddressChange/change_local_address_of_node_and_old_not_acceptable
ACTUAL TEST CASE: change local address of node and all are acceptable
=== CONT  TestHost_LocalAddressChange/change_local_address_of_node_and_all_are_acceptable
ACTUAL TEST CASE: change local address of node and all are acceptable
```
also please run a test with -race

p2p/peerexchange/bookchecker.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+}
+
+func (d *Discovery) CheckPeers() {
+	peers := d.getRandomPeers(peersNumber)
+	qCtx, cancel := context.WithTimeout(context.Background(), checkTimeout)
+	defer cancel()
+	if _, err := d.crawl.query(qCtx, peers); err != nil {
+		d.logger.Error("failed to query bootstrap node: %s", err)
```
dshulyak: error message?

p2p/peerexchange/bookchecker.go
```diff
@@ -0,0 +1,89 @@
...
+			d.logger.Debug("checking book")
+			d.CheckPeers()
+			d.logger.Debug("checking book done")
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (d *Discovery) CheckPeers() {
```
dshulyak: new context should wrap context that is used by errogroup

p2p/peerexchange/bookchecker.go
```diff
@@ -0,0 +1,89 @@
...
+// CheckBook periodically checks the book for dead|alive peers.
+func (d *Discovery) CheckBook(ctx context.Context) {
+	ticker := time.NewTicker(checkInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			d.logger.Debug("checking book")
+			d.CheckPeers()
```
dshulyak: ctx to CheckPeers

p2p/peerexchange/bookchecker.go
```diff
@@ -0,0 +1,89 @@
...
+	"math/rand"
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+const (
+	checkInterval = 3 * time.Minute  // Interval to check for dead|alive peers in the book.
+	checkTimeout  = 30 * time.Second // Timeout to check for dead|alive peers in the book.
+	peersNumber   = 10               // Number of peers to check for dead|alive peers in the book.
```
dshulyak: can be it be a part of config?

dshulyak (changes requested):
1. we should protect bootnodes from being removed from address book, even if they are temporarily unavailable
2. if connection with a peer exists peer should not be queried
3. if peer was queried not long time (maybe 30m-60m?) ago we shouldn't query it again
4. each query will create a connection, i think checker should disconnect a peer if connection was specifically created for a short check

p2p/nodechecker_test.go
abergasov: race generate my helper struct with blackList addresses, when access without mutex

p2p/nodechecker_test.go
```diff
@@ -0,0 +1,241 @@
...
+	require.NoError(t, err)
+	return &hostWrapper{
+		host:      host,
+		transport: addrFactory,
+	}
+}
+
+func discoveryBootstrap(disc *peerexchange.Discovery) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
dshulyak: why it needs defer cancel() if function exits only if Bootstrap is canceled?

p2p/peerexchange/bookchecker.go
```diff
@@ -0,0 +1,133 @@
...
+		return
+	}
+	qCtx, cancel := context.WithTimeout(ctx, d.cfg.CheckTimeout)
+	defer cancel()
+	if _, err := d.crawl.query(qCtx, peers); err != nil {
+		d.logger.Error("failed to check nodes: %s", err)
+		return
+	}
+	// check peers are updated in host peerBook.
+	for peerID, addresses := range d.peersToMap(peers) {
```
dshulyak, line 47: i dont understand why addresses needs to be an array. we get N random peers on L35. is it possible for that result to contain duplicates?

p2p/peerexchange/bookchecker.go
```diff
@@ -0,0 +1,133 @@
...
+			continue
+		}
+		// in case if there are more than one address for peerID is available - for us this is not important, just take first one.
+		newAddr, err := parseAddrInfo(peerStoreAddresses[0].String() + "/p2p/" + peerID.Pretty())
+		if err != nil {
+			d.logger.Error("failed to parse address: %s", err)
+			continue
+		}
+		// update node address in book.
+		d.book.AddAddress(newAddr, addresses[1])
```
dshulyak: why is it addresses[1]? should it be addresses[0]? but maybe we can get rid from duplicates in our peer book instead

p2p/peerexchange/bookchecker.go
abergasov: can be situation, when some nodeA update local address. After updating - node will be trigger notify event and peerBook of nodeB add new address and will contain multiply addresses for nodeA. that's why peerBook contains multiply addresses per node

while our addreesBook store address in `map[peerID]*addrInfo`. in method `GetRandomPeers` L105 i check, does peerBook have extra address for nodeA. if so - add additional address for check.

than check all addresses of node. if one of them is not available - remove. if only one - update to alive address

p2p/peerexchange/bookchecker.go
abergasov: yes, forgot rollback after tests. `addresses[0]` should be here

in case of multiply addresses valid after check - just take first. in our book will be only one address

p2p/nodechecker_test.go
abergasov: removed this

dshulyak (approved):

.github/workflows/ci.yml
```diff
@@ -65,7 +65,6 @@ jobs:
           make
           make tidy
           make test-fmt
-          make lint
```
countvonzero, line 68: we are removing lint check? why?

abergasov: We use Golint in this command. This tool is deprecated and frozen. Also this tool have not annotations
We also using golangci-lint instead as separate step `make golangci-lint-github-action`

It just cleanup)

abergasov:
bors try

abergasov:
bots try

abergasov:
bors try

abergasov:
bors merge

PR #3204, opened 2022-05-09
testonly: delete unused code

countvonzero:
## Motivation

TEST

## TODO

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -41,7 +41,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/layerfetcher"
 	"github.com/spacemeshos/go-spacemesh/layerpatrol"
 	"github.com/spacemeshos/go-spacemesh/log"
-	"github.com/spacemeshos/go-spacemesh/log/errcode"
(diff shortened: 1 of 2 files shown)


countvonzero:
bors try

cmd/node/node.go
```diff
@@ -296,34 +295,6 @@ func (app *App) introduction() {
 	log.Info("Welcome to Spacemesh. Spacemesh full node is starting...")
 }
 
-type clockErrorDetails struct {
```
dshulyak, line 299: hm we actually should use this error for peersync failure

dshulyak: i think Nikita used it only for ntp which was a mistake

dshulyak:
does it break something? i would rather open an issue so that this error is used by peersync verifier. smapp expects this error to show meaningful error to the user

countvonzero:
> does it break something? i would rather open an issue so that this error is used by peersync verifier. smapp expects this error to show meaningful error to the user

no. i was using this PR to test CI. feel free to open an issue

PR #3205, opened 2022-05-10
[Merged by Bors] - systest: cleanup namespaces created for the test after test exited

dshulyak:
this change makes systest cleanup more robust. even if test will crash with a bug (or some less likely OOM) resources will be cleaned up.

- namespaces for the test will be created with a testid label
- after test completes all namespaces that were created for that testid will be cleaned up. unless they have `keep=true`, in such case it is up to dev to clean  them up


countvonzero (approved):

dshulyak:
bors merge

PR #3213, opened 2022-05-12
[Merged by Bors] - layerfetcher: fail tx hashes fetch when at least 1 fails

### LINKED ISSUE #3139: hare, proposals: don't vote for proposals with missing transactions

countvonzero:
## Motivation

Closes #3139

## Changes

- fail tx hashes fetch when any of them fails
- change atx/tx to share the same code with ballot/proposal/ballots in layerfetcher and add tests

## DevOps Notes

### DIFF
--- a/layerfetcher/layers.go
+++ b/layerfetcher/layers.go
@@ -424,6 +424,8 @@ func notifyLayerDataResult(ctx context.Context, layerID types.LayerID, layerDB l
 	}
 
 	if result.Err == nil {
+		// FIXME should not blindly save hare output from peer, potentially overwriting local hare's
+		// consensus result. see https://github.com/spacemeshos/go-spacemesh/issues/3200
 		if err := layerDB.SaveHareConsensusOutput(ctx, layerID, lyrResult.hareOutput); err != nil {
 			logger.With().Error("failed to save hare output from peers", log.Err(err))
 			result.Err = err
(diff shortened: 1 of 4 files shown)


countvonzero:
bors try

dshulyak (approved):

layerfetcher/layers.go
```diff
@@ -552,27 +542,14 @@ func (l *Logic) FetchAtx(ctx context.Context, id types.ATXID) error {
...
-		}
-		if !res.IsLocal {
-			err := l.getAtxResults(ctx, res.Hash, res.Data)
-			if err != nil {
-				return err
-			}
-		}
+	l.log.WithContext(ctx).With().Debug("requesting atxs from peer", log.Int("num_atxs", len(ids)))
+	hashes := types.ATXIDsToHashes(ids)
+	if errs := l.getHashes(ctx, hashes, fetch.ATXDB, l.atxHandler.HandleAtxData); len(errs) > 0 {
```
evt, line 551: is it a common practice to return just the first error?

evt (approved):

layerfetcher/layers.go
countvonzero: ppl tend to invent their own solutions with returning multiple errors. some wrap multiple errors with errors.Wrap but i don't really like that, as for me that implies their are in the same chain of event and is confusing. some created their own customized error to include a list of errors. in this case, i just don't care enough to do that.

my rationale here is that when i'm debugging, i'll look at the log anyway and each of the missing hash has its corresponding log. the point here is to fail the upstream Block/Proposal handler. so i took the simplest approach to just return the first error.

countvonzero:
bors merge

PR #3218, opened 2022-05-20
2703 hash not found

### LINKED ISSUE #2703: devnet: log "hash not found in response from peer" & "remote peer requested non existing hash"
## Description

these two log messages are paired up by the requesting and responding peer of a fetch request.

the majority of these logs can be explained by the disconnect between the low level fetcher and the higher level syncer and gossipHander.

- gossip
when we received a gossip msg, we process the data but throw away the sender. so when we need to fetch anything the data references, we ask fetcher to pick a random peer to fetch that hash. this can result in the above log message, and fetcher succeed in retries. but whoever gossiped that msg to us, should already have all data downloaded. one example https://kibana-tests.spacemesh.io/goto/c77705fd890b66531e6ede7b2b9e4e2d fetching the ref block took 12 retries, despite knowing where we get the block from

fetch-retries

- syncer
when  syncer asks for blockIDs/atxIDs, we know exactly which peer that send us those IDs, but then when we ask fetcher to fetch data, syncer doesn't pass on the peer to fetcher. fetcher just picks a random peer to fetch the data. this also may result in the log msg above.
(i have an issue for this https://github.com/spacemeshos/go-spacemesh/issues/2472 but haven't got time to work on it)

## Steps To Reproduce
1. make dockertest-late-nodes namespace=$YOUR_NAMESPACE dump=True
hash not found in response from peer

## Expected Behavior

fetcher should accept a peer as the preferred destination peer. the caller of fetcher should provide the src peer they get the hash from. the consideration for this is potentially overloading a certain peer.

evt:
## Motivation
Closes #2703 

## DevOps Notes


fetch/fetch.go
```diff
@@ -467,30 +468,47 @@ func (f *Fetch) receiveResponse(data []byte) {
...
+	for i := range b {
+		b[i] = letters[rand.Intn(len(letters))]
+	}
+	return string(b)
+}
+
 // sendBatch dispatches batched request messages.
-func (f *Fetch) sendBatch(requests []requestMessage) {
+func (f *Fetch) sendBatch(p p2p.Peer, requests []requestMessage) {
+	id := randSeq(10)
```
countvonzero, line 507: what is this used for? how is this different from batch.ID?
if you want it for cross-requestst tracing, use a context with the same request ID

fetch/fetch.go
```diff
@@ -521,37 +543,78 @@ func (f *Fetch) sendBatch(requests []requestMessage) {
...
+
+		// if peer is not provided - send to random one
+		if p2p.IsAnyPeer(peer) {
+			peer = GetRandomPeer(f.net.GetPeers())
+			peerUpdated = true
+		}
+
+		// if we've tried peer already - try to change (just once to avoid infinite loop if only 1 peer available)
+		if _, exists := seenPeers[peer]; exists {
+			oldPear := peer
```
countvonzero, line 561: oldPeer

fetch/fetch.go
```diff
@@ -521,37 +543,78 @@ func (f *Fetch) sendBatch(requests []requestMessage) {
...
 			return
 		}
 
-		// get random peer
-		p := GetRandomPeer(f.net.GetPeers())
-		f.log.With().Debug("sending request batch to peer",
+		peerUpdated := false
+
+		// if peer is not provided - send to random one
+		if p2p.IsAnyPeer(peer) {
```
countvonzero, line 554: this should be outside the loop. it can only be empty when it enter this method.

fetch/fetch.go
```diff
@@ -521,37 +543,78 @@ func (f *Fetch) sendBatch(requests []requestMessage) {
...
+			peer = GetRandomPeer(f.net.GetPeers())
+			peerUpdated = true
+		}
+
+		// if we've tried peer already - try to change (just once to avoid infinite loop if only 1 peer available)
```
countvonzero, line 559: don't think this logic is necessary.

say we have retries == 100. but we only have 9 peers. this logic will effectively reduce the retries to 9.
a peer may not have a hash at t0, but it may have the hash at t1.

layerfetcher/layers.go
```diff
@@ -370,7 +370,7 @@ func (l *Logic) receiveLayerContent(ctx context.Context, layerID types.LayerID,
 	logger.Debug("received layer content from peer")
 	peerRes := extractPeerResult(logger, layerID, data, peerErr)
 	if peerRes.err == nil {
-		if err := l.fetchLayerData(ctx, logger, layerID, peerRes.data); err != nil {
+		if err := l.fetchLayerData(ctx, logger, peer, layerID, peerRes.data); err != nil {
```
countvonzero, line 373: i think i commented about this before in slack.
say we have 3 peers ABC. the peers return layer data in the following time sequence CAB. and they gave us exactly the same layer data. 
before this PR, all hashes will be distributed evenly between A, B and C.
after this PR, block/ballot hashes will all go to peer C.

i don't think this is a good behavioral change. 
we can wait until all peers reported back and evenly distribute the work amongst them in fetchLayerData

countvonzero (changes requested):
this is the intuitive approach for the issue indeed. but as commented below we should try harder to dish out the work evenly. note that your PR only affect syncing block/ballot. synced txs and gossiped block/ballot/txs are not handled.

i wonder if we should solve this problem from a different angle. if we create a fixed-size FIFO cache of hash->set of peer, then we can solve this more generally (without adding peer to all the APIs)

for example, in the syncing case.
when receiving layer data from peers A, B, C. we came compile a mapping that 
ballot_X_hash: A, B
ballot_Y_hash: A, B, C
ballot_Z_hash: C
so when it comes time to fetch ballot XYZ, we look up from the map and choose a random one in its peer list.

if a hash doesn't exist in the mapping, we just choose a random peer.

also, block can be fetched as a result of data integrity check when handling a ballot. say ballot_X contains votes for block_G. in this case when we ask to fetch block_G_hash, we could optionally provide a referral hash ballot_X_hash and from the mapping we know we can ask peer A or B for block_G_hash because if they don't have it, then they are giving us incomplete data and therefore malicious.

fetch/fetch.go
evt: this is just for debugging purposes to see log messages for the same `sendBatch` call. I have now a case where peer is attached to request (and thus should have data we need), but somehow hash is still not found, so debugging to find out why.

evt:
Closed in favour of https://github.com/spacemeshos/go-spacemesh/pull/3244

Issue #3036, opened 2021-12-28
Block Construction (Unification of Proposals)


noamnelke:
This issue covers how to construct a block after the Hare converged on a set of proposals. This is done as part of the Hare certification phase and the Hare certifies the resulting `block_id`.

1. Calculate the **block seed**:
	1. Sort proposal IDs lexicographically.
	2. Initialize a Sha256 hash and write the sorted proposal IDs into it.
	3. Calculate the Sha256 sum and use it as the block seed.
2. **Combine** transactions from entire set of proposals into a single set.
3. **Fetch** available transactions from mempool and missing transactions from peers.
4. If more than one transaction competes for the same principal and nonce:
	1. Pick the one that pays the highest gas (`max_gas * gas_price`).
	2. If multiple transactions pay the same highest gas - use a tie breaker: Calculate the Sha256 sum of the concatenation of the block seed and the transaction ID and pick the one that's lowest, lexicographically. Hashing the transaction ID with the block seed makes the selection unpredictable, preventing users from being able to replace transactions without bumping the fee.
	3. **Remove all remaining transactions** from the set.
5. Perform **initial shuffling** of transactions:
	1. Sort transaction IDs lexicographically.
	2. Initialize a Mersenne Twister with the block seed.
	3. Use the MT as a source of randomness for a Fisher-Yates shuffle of the previously-sorted transaction IDs.

    Something similar, but for blocks, is implemented today [here](https://github.com/spacemeshos/go-spacemesh/blob/9972b2c9c7ee79a2ef913eeb0d8ead05ee74769c/mesh/mesh.go#L603-L627) and can be used as a reference.

6. Perform **intra-principal sorting by nonce**. We want transactions associated with a principal account which have a lower nonce value to appear before transactions associated with the same principal account that have a higher nonce value. We want to do this without affecting the order of other transactions.
E.g. if `a3` is a transaction with principal `a` and nonce `3`, then the shuffled list of transactions:
`[ a3 c1 a0 b0 a2 c0 b1 ]`
will become, after intra-principal sorting:
`[ a0 c0 a2 b0 a3 c1 b1 ]` 
Note that the principal in each slot in the list never changes. One way to implement this (but feel free to come up with a more efficient method) is:
    ```python
    txs_by_principal = defaultdict(list)
    for tx in shuffled_list:
        txs_by_principal[tx.principal].append(tx)
    for txs in txs_by_principal.values():
	    # sort in decsending order, because pop() pops from the end.
[...]

PR #3219, opened 2022-05-21
[Merged by Bors] - tx selection: distribute rewards proportional to proposal/ballot weight

### LINKED ISSUE #1798: Fees are distributed to miners even if the transaction can't be applied

### LINKED ISSUE #3057: mesh: when state is reverted applied transactions and rewards must be reverted as well
Transactions store a reference to a block when they were applied, if the state is reverted some transactions may not be applied at all, or applied in the block that is different from the original.

Rewards are computed based on applied block, if the block is reverted rewards need to be adjusted,

countvonzero:
## Motivation

part of #3036 
Closes #3057
Closes #1798

## Changes

- move ballot weight calculation to proposals pkg from tortoise pkg 
- change Block.Rewards to a list of tuple [coinbase, proposal_weight]
- move rewards calculation to vm (after transactions are applied)
- revert rewards when state is reverted
- remove API that query rewards by smesher ID

## Test Plan

unit tests, systest

## DevOps Notes


countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/6534281078?check_suite_focus=true)

21T02:42:19.144Z`	INFO	using	{"namespace": "test-txsp"}
=== CONT  TestShortTimeskew
    assertion_compare.go:342: 
        	Error Trace:	timeskew_test.go:69
        	Error:      	"11" is not less than or equal to "9"
        	Test:       	TestShortTimeskew
        	Messages:   	[]
    logger.go:130: 2022-05-21T02:52:40.503Z	INFO	namespace was deleted	{"namespace": "test-ensf"}
--- FAIL: TestShortTimeskew (1[71](https://github.com/spacemeshos/go-spacemesh/runs/6534058268?check_suite_focus=true#step:11:72)0.84s)

blocks/generator.go
```diff
@@ -131,13 +136,44 @@ func (g *Generator) calculateSmesherRewards(logger log.Log, layerID types.LayerI
...
+	sort.Slice(coinbases, func(i, j int) bool {
+		return bytes.Compare(coinbases[i].Bytes(), coinbases[j].Bytes()) < 0
+	})
+	rewards := make([]types.AnyReward, 0, len(weights))
+	for _, coinbase := range coinbases {
+		weight, ok := weights[coinbase]
+		if !ok {
+			g.logger.With().Fatal("coinbase missing", coinbase)
+		}
+		wb, err := weight.GobEncode()
```
dshulyak: this is definitely wrong. it is weird that we use []byte to store weight, but why it gets encoded with gob

dshulyak (changes requested):
we can't use GobEncode to encode rewards, moreover it is weird that we are using []byte to store it in the database. we have to use that fixed float library if we want to use block weight for this purposes

dshulyak:
or we can store two uint64 (rat.Num and rat.Denom), but it is also weird

mesh/mesh.go
```diff
@@ -545,6 +546,11 @@ func (msh *Mesh) revertState(logger log.Log, revertTo types.LayerID) error {
 		return fmt.Errorf("revert state to layer %v: %w", revertTo, err)
 	}
 	logger.With().Info("successfully reverted state", log.Stringer("state_root", root))
+
+	if err = rewards.RevertToLayer(msh.db, revertTo); err != nil {
```
dshulyak: this revert needs to be atomic with accounts revert. also rewards should be saved in the vm as well, so that accounts changes are atomic with rewards

vm/vm.go
```diff
@@ -60,23 +80,28 @@ func (vm *VM) SetupGenesis(genesis *config.GenesisConfig) error {
 // ApplyLayer applies the given rewards to some miners as well as a vector of
 // transactions for the given layer. to miners vector for layer. It returns an
 // error on failure, as well as a vector of failed transactions.
-func (vm *VM) ApplyLayer(lid types.LayerID, txs []*types.Transaction, rewards []types.AnyReward) ([]*types.Transaction, error) {
+func (vm *VM) ApplyLayer(lid types.LayerID, txs []*types.Transaction, rewards []types.AnyReward) ([]*types.Transaction, []*types.Reward, error) {
```
dshulyak: i think types.Reward should be removed from Apply signature

sql/migrations/0001_initial.sql
```diff
@@ -41,14 +41,14 @@ CREATE TABLE mesh_status
 
 CREATE TABLE rewards
 (
-    smesher      CHAR(64),
     coinbase     CHAR(20),
     layer        INT NOT NULL,
     total_reward UNSIGNED LONG INT,
-    layer_reward UNSIGNED LONG INT,
-    PRIMARY KEY (smesher, layer)
+    lyr_reward   UNSIGNED LONG INT,
```
dshulyak: why it got changed?

sql/rewards/rewards.go
```diff
@@ -7,79 +7,45 @@ import (
...
-	stmt.ColumnBytes(0, reward.SmesherID[:])
-	stmt.ColumnBytes(1, reward.Coinbase[:])
-	return reward, nil
-}
-
-// FilterByCoinbase filters rewards from all layers by coinbase address.
-func FilterByCoinbase(db sql.Executor, address types.Address) (rst []types.Reward, err error) {
-	if _, err := db.Exec("select smesher, coinbase, layer, total_reward, layer_reward from rewards where coinbase = ?1;",
+// RevertToLayer reverts the rewards to the specified layer.
+func RevertToLayer(db sql.Executor, revertTo types.LayerID) error {
```
dshulyak: can we write more compact apis -> Revert(db, to Layer)? why it needs to have same stuff repeated multiple times

sql/rewards/rewards.go
```diff
@@ -7,79 +7,45 @@ import (
...
 	}
-	return rst, err
+	return nil
 }
 
-// FilterBySmesher filters rewards from all layers by smesher address.
-func FilterBySmesher(db sql.Executor, address []byte) (rst []types.Reward, err error) {
-	if _, err := db.Exec("select smesher, coinbase, layer, total_reward, layer_reward from rewards where smesher = ?1",
+// GetRewards get rewards from all layers for the coinbase address.
+func GetRewards(db sql.Executor, coinbase types.Address) (rst []*types.Reward, err error) {
```
dshulyak: rewards is in module name

mesh/mesh.go
countvonzero: 2 questions:
- why should we save rewards in vm? when the layers are reverted, they account balance were just reverted. why should we treat rewards differently inside vm?
- the mesh data and vm are are updated separately in the codebase right now. we have not discussed any effort to keep their operation atomic. i don't see how we can do this easily either.

vm/vm.go
countvonzero: i don't understand this comment. do you mean that we only keep rewards data inside vm?

countvonzero:
> we can't use GobEncode to encode rewards, moreover it is weird that we are using []byte to store it in the database. we have to use that fixed float library if we want to use block weight for this purposes

you are right. at first i tried the fixed float library ("github.com/spacemeshos/fixed"), it doesn't not get encoded correctly. then i tried rat.Num/Denom but it seems go-lang specific. so i thought using []byte. but then i neglected the fact that GobEncode is also golang specific....sigh..

i am not sure about options here other than representing weight with fixed precision float in 
- text
- a tuple of uint64 (a, b) where it gets translated to a.b
- add serialization support in "github.com/spacemeshos/fixed" lib

ideas? @dshulyak @noamnelke

mesh/mesh.go
dshulyak: > why should we save rewards in vm? when the layers are reverted, they account balance were just reverted. why should we treat rewards differently inside vm?

so lets say it is not atomic, node failed to revert rewards. now api will provide rewards that are inconsistent with the actual balance (or vice versa if they are updated in a different order)

> the mesh data and vm are are updated separately in the codebase right now. we have not discussed any effort to keep their operation atomic. i don't see how we can do this easily either.

so first of all current codebase has consistency bugs, just for now we ignore them
also what is a mesh data?
- rewards
- applied layer
- transaction status (failure/success and for genesis we will add fake-receipts)
- anything else?

there are 2 options:
- change this things atomically in a single tx, this is viable since all that data is managed by vm. this approach is the simplest code wise and fail safe
- manually restore consistency of the data. so for example we need to answer how we handle consistency of the balances<>transactions and balances<>rewards, and transactions<>rewards in the event of the crash. in practice manual solution almost certainly will have tricky bugs. maybe always do vm revert if applied layer wasn't updated, but this is not a fail-safe approach

dshulyak:
> a tuple of uint64 (a, b) where it gets translated to a.b

why is it golang specific? it just represents a float without a loss of precision. i think any language should have some wrapper for this type (example in rust https://docs.rs/fraction/latest/fraction/#simple-use).

i looked up that github.com/spacemeshos/fixed stores value in two uint64, so encoding for both choices should be very similar. maybe it can be packed into one uint64 without a loss @noamnelke  

i would say that both approaches are ok, maybe "fixed" is cleaner for consistency reasons

common/util/weight.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (w Weight) String() string {
+	if w.IsNil() {
+		return "0"
+	}
+	if w.Rat.IsInt() {
+		return w.Rat.Num().String()
+	}
+	return w.Rat.FloatString(2)
```
dshulyak, line 96: btw fixed has super ugly String output. it is literally unreadable and useless for debugging, to the point where i would keep using Rat in tortoise even if it leads to some code duplication

mesh/mesh.go
countvonzero: > > why should we save rewards in vm? when the layers are reverted, they account balance were just reverted. why should we treat rewards differently inside vm?
> 
> so lets say it is not atomic, node failed to revert rewards. now api will provide rewards that are inconsistent with the actual balance (or vice versa if they are updated in a different order)
> 
> > the mesh data and vm are are updated separately in the codebase right now. we have not discussed any effort to keep their operation atomic. i don't see how we can do this easily either.
> 
> so first of all current codebase has consistency bugs, just for now we ignore them also what is a mesh data?
> 
> * rewards
> * applied layer
> * transaction status (failure/success and for genesis we will add fake-receipts)
> * anything else?
> 
> there are 2 options:
> 
> * change this things atomically in a single tx, this is viable since all that data is managed by vm. this approach is the simplest code wise and fail safe
> * manually restore consistency of the data. so for example we need to answer how we handle consistency of the balances<>transactions and balances<>rewards, and transactions<>rewards in the event of the crash. in practice manual solution almost certainly will have tricky bugs. maybe always do vm revert if applied layer wasn't updated, but this is not a fail-safe approach

the 1st is the obvious choice here. but my reservation is that accounts data are in sqlite only for genesis. if we make rewards and txs applied atomically, does that mean when we have a VM and account state trie, we will need to figure this out again?

if memory serves, i think geth simply apply rewards as internal transfers and record the block coinbase in the block header. so maybe rewards and accounts tables can be removed together once state is properly implemented.

countvonzero:
> > a tuple of uint64 (a, b) where it gets translated to a.b
> 
> why is it golang specific? it just represents a float without a loss of precision. i think any language should have some wrapper for this type (example in rust https://docs.rs/fraction/latest/fraction/#simple-use).
> 
> i looked up that github.com/spacemeshos/fixed stores value in two uint64, so encoding for both choices should be very similar. maybe it can be packed into one uint64 without a loss @noamnelke
> 
> i would say that both approaches are ok, maybe "fixed" is cleaner for consistency reasons

i am leaning towards using 2 uint64 for now with big.Rat.Num/Denom. we only use fixed in bincdf calculation and nowhere else. for consistency, i'm inclined to use the Weight from tortoise.

mesh/mesh.go
dshulyak: > the 1st is the obvious choice here. but my reservation is that accounts data are in sqlite only for genesis. if we make rewards and txs applied atomically, does that mean when we have a VM and account state trie, we will need to figure this out again?

there is an option to keep state in the node and provide it to the svm over some interface. it can be cgo api, or share a C reference to sqlite object 

> if. memory serves, i think geth simply apply rewards as internal transfers and record the block coinbase in the block header. so maybe rewards and accounts tables can be removed together once state is properly implemented.

state for accounts can remain in the flat storage (sql table). there is nothing improper about it. commitment (e.g. inner nodes of the trie) can be stored separately. basically we will follow erigon design, i described it some time ago. 

specifically accounts table doesnt have to be removed. it can be extended with accounts_inner_nodes table that will store inner nodes of the commitment tree. 

if you want rewards can be removed even now, but we will also have to drop (rework?) API for them

countvonzero: for now i changed to atomic update with accounts data.

vm/vm.go
countvonzero: done.

sql/migrations/0001_initial.sql
countvonzero: accident.

countvonzero:
@dshulyak made two major changes based on your review
- move all rewards db update inside vm and atomic with accounts data
- use two uint64 for weight in block rewards

countvonzero:
bors try

sql/rewards/rewards.go
```diff
@@ -7,79 +7,45 @@ import (
...
-func Add(db sql.Executor, lid types.LayerID, reward *types.AnyReward) error {
-	if _, err := db.Exec(`insert into rewards 
-			(smesher, coinbase, layer, total_reward, layer_reward) 
-			values (?1, ?2, ?3, ?4, ?5) 
-		on conflict(smesher,layer) 
-			do update set 
-				total_reward=add_uint64(total_reward, ?4),
-		 		layer_reward=add_uint64(layer_reward, ?5);`,
+func Add(db sql.Executor, reward *types.Reward) error {
+	if _, err := db.Exec(`insert into rewards (coinbase, layer, total_reward, layer_reward) values (?1, ?2, ?3, ?4);`,
```
dshulyak: it was written with do update so that there is no need for creating a map with coinbases

sql/rewards/rewards.go
```diff
@@ -7,79 +7,45 @@ import (
...
 	}
-	return rst, err
+	return nil
 }
 
-// FilterBySmesher filters rewards from all layers by smesher address.
-func FilterBySmesher(db sql.Executor, address []byte) (rst []types.Reward, err error) {
-	if _, err := db.Exec("select smesher, coinbase, layer, total_reward, layer_reward from rewards where smesher = ?1",
+// Get rewards from all layers for the coinbase address.
+func Get(db sql.Executor, coinbase types.Address) (rst []*types.Reward, err error) {
```
dshulyak: List ?

dshulyak (approved):

sql/rewards/rewards.go
countvonzero: what do you mean by `a map with coinbase`? i don't understand this comment.
in the old code, `do update` loses the info at which layer the reward was given

countvonzero: oh i was wrong. it has the on conflict check. sorry.

countvonzero: changed it back to the old code. also updated test

sql/rewards/rewards.go
countvonzero: done

sql/rewards/rewards.go
dshulyak: yes, so the idea was that it is possible to write many rewards for the same layer. so if there is a list with rewards there is no need to create a map with a unique coinbases

countvonzero:
bors merge

countvonzero:
bors merge

PR #3223, opened 2022-05-27
[Merged by Bors] - sql: fix bug where update layer fails after SetNextLayer

countvonzero:
## Motivation

updating the layer/block fails when the transaction's layer/block was reset previously when applying a layer.

## Changes

when a transaction TX is included in a proposal P11 at layer N but not included in the layer's block B. after B is applied, TX's block/layer will be reset to the next lowest layer of the proposal that it was included.

in the case where tx is not included in any more proposals, the block/layer is set to NULL after `SetNextLayer` (part of applying block B).
when the next layer proposal P33 includes TX, `UpdateIfBetter` is called on TX.

before PR
- update fails because sql query didn't consider NULL case
after PR
- update succeeds

## Test Plan

unit test that fails before this PR and passes after

## DevOps Notes

### DIFF
--- a/sql/transactions/transactions.go
+++ b/sql/transactions/transactions.go
@@ -96,7 +96,9 @@ func HasBlockTX(db sql.Executor, bid types.BlockID, tid types.TransactionID) (bo
 func UpdateIfBetter(db sql.Executor, tid types.TransactionID, lid types.LayerID, bid types.BlockID) (int, error) {
 	rows, err := db.Exec(`
 		update transactions set layer = ?5, block = ?6
-		where id = ?1 and applied = ?2 and (layer = ?3 or layer > ?5 or (layer = ?5 and block = ?4)) returning id`,
+		where id = ?1 and applied = ?2 and
+		(layer is null or layer = ?3 or layer > ?5 or (layer = ?5 and (block = ?4 or block is null)))
+		returning id`,
 		func(stmt *sql.Statement) {
 			stmt.BindBytes(1, tid.Bytes())
 			stmt.BindInt64(2, statePending)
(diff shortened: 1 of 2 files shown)


sql/transactions/transactions.go
```diff
@@ -96,7 +96,9 @@ func HasBlockTX(db sql.Executor, bid types.BlockID, tid types.TransactionID) (bo
 func UpdateIfBetter(db sql.Executor, tid types.TransactionID, lid types.LayerID, bid types.BlockID) (int, error) {
 	rows, err := db.Exec(`
 		update transactions set layer = ?5, block = ?6
-		where id = ?1 and applied = ?2 and (layer = ?3 or layer > ?5 or (layer = ?5 and block = ?4)) returning id`,
+		where id = ?1 and applied = ?2 and
+		(layer is null or layer = ?3 or layer > ?5 or (layer = ?5 and (block = ?4 or block is null)))
```
evt, line 100: having so many or-s in request should probably make all indexes skipped to use. Or we don't much care about request performance in this case?

countvonzero: hmmmm.... i care about performance. what do you mean by `make all indexes skipped to use`?

`id` is the primary key, and should be the only index used in this query.

evt: ah, okay, sorry :) what I mean is indexes never used for `or` but as long as `id` is enough to find a record - should be ok, cheers :)

evt (approved):

sql/transactions/transactions.go
countvonzero: thanks for the review. yeah. the primary key already narrow it down to 0 or 1 record. we can't do better that. more index will not help with the value comparisons on fields. this query essentially does 0 or 1 row update.

please do keep reminding in the future about index. it's always an important question to ask when writing sql queries.

countvonzero:
bors merge

PR #3224, opened 2022-05-29
[Merged by Bors] - tx selection: proposal/block construction

### LINKED ISSUE #3035: Proposal Construction

### LINKED ISSUE #3036: Block Construction (Unification of Proposals)

countvonzero:
## Motivation

Closes #3035
Closes #3036

## Changes

see #3035 and #3036 for implementation details.

proposal and block constructions uses the same code (cache.go) to construct the list of transactions. 

key points:
- txs balance check and nonce ordering (conservative cache)
  - proposals construction uses a conservative cache backed by database data for getting more eligible txs
  - block construction uses a conservative cache without a database, since the set of txs are fixed by the proposals  
- txs selection
  - proposal construction uses mempool_iterator to sort the best N txs, constraints by gas limit, for a proposal, and then sort them in nonce order within each account.
  - block construction has to include all txs from proposals, it put these txs in a stable order while keeping nonce order within each account, and pick them in sequence up to gas limit.
- selection stability
  - proposal construction uses local randomness and is not stable
  - block construction uses a Mersenne Twister (seeded by the hare consensus output) as source of randomness and is stable.
- optimistic filtering for block
  - ON: the conservative cache will use the real state for balance/nonce check and drop those that fail the check
  - OFF: the conservative cache will use a fake state that allow all txs to pass the check, hence keeping all txs

## Test Plan

unit tests, systests

## DevOps Notes

### DIFF
--- a/common/types/proposal.go
+++ b/common/types/proposal.go
@@ -45,6 +45,11 @@ type InnerProposal struct {
 	Ballot
 	// smesher's content proposal for a layer
 	TxIDs []TransactionID
+	// aggregated hash up to the layer before this proposal.
+	MeshHash Hash32
+	// state root up to the layer before this proposal.
+	// note: this is needed in addition to mesh hash to detect bug in SVM
+	StateHash Hash32
 }
 
 // Initialize calculates and sets the Proposal's cached proposalID.
@@ -100,6 +105,8 @@ func (p *Proposal) MarshalLogObject(encoder log.ObjectEncoder) error {
 		}
 		return nil
 	}))
+	encoder.AddString("mesh_hash", p.MeshHash.String())
+	encoder.AddString("state_root", p.StateHash.String())
 	p.Ballot.MarshalLogObject(encoder)
 	return nil
 }
--- a/common/types/proposal.go
+++ b/common/types/proposal.go
@@ -47,9 +47,10 @@ type InnerProposal struct {
 	TxIDs []TransactionID
 	// aggregated hash up to the layer before this proposal.
 	MeshHash Hash32
+	// TODO add this when a state commitment mechanism is implemented.
 	// state root up to the layer before this proposal.
 	// note: this is needed in addition to mesh hash to detect bug in SVM
-	StateHash Hash32
+	// StateHash Hash32
 }
 
 // Initialize calculates and sets the Proposal's cached proposalID.
@@ -106,7 +107,6 @@ func (p *Proposal) MarshalLogObject(encoder log.ObjectEncoder) error {
 		return nil
 	}))
 	encoder.AddString("mesh_hash", p.MeshHash.String())
-	encoder.AddString("state_root", p.StateHash.String())
 	p.Ballot.MarshalLogObject(encoder)
 	return nil
 }
--- a/api/grpcserver/grpcserver_test.go
+++ b/api/grpcserver/grpcserver_test.go
@@ -60,7 +60,7 @@ const (
 	genTimeUnix      = 1000000
 	layerDurationSec = 10
 	layerAvgSize     = 10
-	txsPerBlock      = 99
+	txsPerProposal   = 99
 	layersPerEpoch   = uint32(5)
 	networkID        = 120
 	atxPerLayer      = 2
@@ -1081,7 +1081,7 @@ func TestSmesherService(t *testing.T) {
 
 func TestMeshService(t *testing.T) {
 	logtest.SetupGlobal(t)
-	grpcService := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerBlock)
+	grpcService := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerProposal)
 	shutDown := launchServer(t, grpcService)
 	defer shutDown()
 
@@ -1135,7 +1135,7 @@ func TestMeshService(t *testing.T) {
 			logtest.SetupGlobal(t)
 			response, err := c.MaxTransactionsPerSecond(context.Background(), &pb.MaxTransactionsPerSecondRequest{})
 			require.NoError(t, err)
-			require.Equal(t, uint64(layerAvgSize*txsPerBlock/layerDurationSec), response.MaxTxsPerSecond.Value)
+			require.Equal(t, uint64(layerAvgSize*txsPerProposal/layerDurationSec), response.MaxTxsPerSecond.Value)
 		}},
 		{"AccountMeshDataQuery", func(t *testing.T) {
 			logtest.SetupGlobal(t)
@@ -2112,7 +2112,7 @@ func checkLayer(t *testing.T, l *pb.Layer) {
 
 func TestAccountMeshDataStream_comprehensive(t *testing.T) {
 	logtest.SetupGlobal(t)
-	grpcService := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerBlock)
+	grpcService := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerProposal)
 	shutDown := launchServer(t, grpcService)
 	defer shutDown()
 
@@ -2446,7 +2446,7 @@ func TestLayerStream_comprehensive(t *testing.T) {
 	}
 	logtest.SetupGlobal(t)
 
-	grpcService := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerBlock)
+	grpcService := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerProposal)
 	shutDown := launchServer(t, grpcService)
 	defer shutDown()
 
@@ -2680,7 +2680,7 @@ func TestMultiService(t *testing.T) {
 	logtest.SetupGlobal(t)
 	cfg.GrpcServerPort = 9192
 	svc1 := NewNodeService(&networkMock, meshAPI, &genTime, &SyncerMock{}, &ActivationAPIMock{})
-	svc2 := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerBlock)
+	svc2 := NewMeshService(meshAPI, conStateAPI, &genTime, layersPerEpoch, networkID, layerDurationSec, layerAvgSize, txsPerProposal)
 	shutDown := launchServer(t, svc1, svc2)
 	defer shutDown()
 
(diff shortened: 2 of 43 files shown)


countvonzero:
bors try

common/types/proposal.go
```diff
@@ -45,6 +45,11 @@ type InnerProposal struct {
 	Ballot
 	// smesher's content proposal for a layer
 	TxIDs []TransactionID
+	// aggregated hash up to the layer before this proposal.
+	MeshHash Hash32
+	// state root up to the layer before this proposal.
+	// note: this is needed in addition to mesh hash to detect bug in SVM
+	StateHash Hash32
```
dshulyak: i thought that we agreed to drop this field for genesis. we will use it internally in the observability tooling to detect such bugs

dshulyak (approved):
so this change guarantees that transactions will be sorted in the correct order? what if node received a block with txs out of order - is such block discarded, or saved and potentially can be applied?

countvonzero:
> so this change guarantees that transactions will be sorted in the correct order? what if node received a block with txs out of order - is such block discarded, or saved and potentially can be applied?

this PR guarantees honest nodes will create a block with txs sorted in stable order. there is no guarantee in general what blocks we will receive, malicious or not. and if tortoise verify such a block (w txs out of order) the state will still apply txs until it cannot, and declare those not-applied txs failed. the logic here didnt change.

common/types/proposal.go
countvonzero: my recollection is that we use some kind of hash (the account diff) for this purpose for now. do you see a big problem here? i am ok w dropping it but thought once we have a real account state, we dont have to change node code. @dshulyak

dshulyak:
why do we save such block then? if we discard it, because it is syntactically invalid - all honest nodes will automatically vote against it. so if there is an overwhelming honest majority - tortoise will eventually reach consensus against such block.

common/types/proposal.go
dshulyak: i personally don't see any issues if we will keep it. besides that after commitment trie will be introduced this hash will be computed differently and it may cause some confusion if  commitment trie will be built starting from older layers.

countvonzero: it may force us to do a hard fork when account state is ready, but i dont see how that (account state) isnt a hard fork anyway. but i can leave it out out Proposal for now and keep the warning log.

dshulyak: commitments and account state is not the same thing

countvonzero:
> why do we save such block then? if we discard it, because it is syntactically invalid - all honest nodes will automatically vote against it. so if there is an overwhelming honest majority - tortoise will eventually reach consensus against such block.

true. i will file an issue and do that in a separate PR.

common/types/proposal.go
countvonzero: i had assume that this is the root of state trie. what is commitment then?

dshulyak: i was referring to the comment about account state. account state is a leaf in the state trie, but it doesn't have to be coupled with the state trie. so e.g. when account state is read or updated there is no reason to even think about state trie.

state trie is used as a state commitment mechanism, and can be implemented relatively independently.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/6712146015?check_suite_focus=true)

filed https://github.com/spacemeshos/go-spacemesh/issues/3233

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/6719694870?check_suite_focus=true)
```
=== CONT  TestShortTimeskew
    assertion_compare.go:342: 
        	Error Trace:	timeskew_test.go:[69](https://github.com/spacemeshos/go-spacemesh/runs/6719396707?check_suite_focus=true#step:11:70)
        	Error:      	"11" is not less than or equal to "9"
        	Test:       	TestShortTimeskew
        	Messages:   	[]
    logger.go:130: 2022-06-03T03:15:13.193Z	INFO	namespace was deleted	{"namespace": "test-qnaq"}
--- FAIL: TestShortTimeskew (1810.38s)
```

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/6720196640?check_suite_focus=true)

```
=== CONT  TestShortTimeskew
    assertion_compare.go:342: 
        	Error Trace:	timeskew_test.go:69
        	Error:      	"11" is not less than or equal to "9"
        	Test:       	TestShortTimeskew
        	Messages:   	[]
    logger.go:130: 2022-06-03T04:21:08.368Z	INFO	namespace was deleted	{"namespace": "test-gpmb"}
```

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/6727846010?check_suite_focus=true)

```
--- FAIL: TestHost_LocalAddressChange (0.00s)
    --- FAIL: TestHost_LocalAddressChange/change_local_address_of_node_and_new_not_acceptable (5.36s)
        nodechecker_test.go:152: 
            	Error Trace:	nodechecker_test.go:152
            	Error:      	Not equal: 
            	            	expected: 1
            	            	actual  : 0
            	Test:       	TestHost_LocalAddressChange/change_local_address_of_node_and_new_not_acceptable
            	Messages:   	nodeB should have 1 addresses
FAIL
FAIL	github.com/spacemeshos/go-spacemesh/p2p	10.[77](https://github.com/spacemeshos/go-spacemesh/runs/6727716867?check_suite_focus=true#step:5:78)8s
```
filed https://github.com/spacemeshos/go-spacemesh/issues/3236

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/6728957010?check_suite_focus=true)

succeeded on retry
https://github.com/spacemeshos/go-spacemesh/runs/6729288062?check_suite_focus=true

countvonzero:
bors merge

PR #3225, opened 2022-05-30
[Merged by Bors] - p2p address book refactor to separate package

### LINKED ISSUE #3132: p2p: stale records in the address book

abergasov:
## Motivation
Closes #3132 

## Changes
move addressbook to separate package with some refactoring

## Test Plan
unit tests

## DevOps Notes


log/zap.go
```diff
@@ -36,6 +36,11 @@ func (l Log) Error(format string, args ...interface{}) {
 	l.logger.Sugar().Errorf(format, args...)
 }
 
+// LogError prints message with fields.
+func (l Log) LogError(msg string, err error, fields ...LoggableField) {
```
dshulyak: please just wrap the error as log.Err field and remove this method

log/zap.go
```diff
@@ -278,6 +283,11 @@ func (fl FieldLogger) Error(msg string, fields ...LoggableField) {
 	fl.l.Error(msg, unpack(append(fields, String("name", fl.name)))...)
 }
 
+// LogError prints message with fields.
+func (fl FieldLogger) LogError(msg string, err error, fields ...LoggableField) {
+	fl.l.Error(msg, unpack(append(fields, Err(err), String("name", fl.name)))...)
```
dshulyak: same

p2p/addressbook/addrbook.go
```diff
@@ -283,46 +228,113 @@ func (a *addrBook) pickTried(bucket int) *knownAddress {
...
+	}
+	if len(nonEmptyBuckets) == 0 {
+		return nil // no addresses to extract.
+	}
+	large := 1 << 30
+	factor := 1.0
+	for {
+		// pick a random bucket.
+		bucket := nonEmptyBuckets[a.rand.Intn(len(nonEmptyBuckets))]
+		if len(buckets[bucket]) == 0 {
```
dshulyak: how it can be zero if it was filtered out above?

p2p/addressbook/addrbook.go
```diff
@@ -283,46 +228,113 @@ func (a *addrBook) pickTried(bucket int) *knownAddress {
...
+
+		// Pick a random entry in it.
+		var ka *knownAddress
+		nth := a.rand.Intn(len(buckets[bucket]))
+		for _, value := range buckets[bucket] {
+			if nth == 0 {
+				ka = value
+			}
+			nth--
+		}
```
dshulyak, line 267: why ka is not buckets[bucket][nth] ? maybe i just dont understand this loop

p2p/addressbook/addrbook.go
```diff
@@ -283,46 +228,113 @@ func (a *addrBook) pickTried(bucket int) *knownAddress {
...
+		if ka == nil {
```
dshulyak: why it can be nil?

p2p/addressbook/addrbook.go
```diff
@@ -283,46 +228,113 @@ func (a *addrBook) pickTried(bucket int) *knownAddress {
...
+	if numAddresses > a.cfg.GetAddrMax {
+		numAddresses = a.cfg.GetAddrMax
 	} else if numAddresses == 0 {
 		numAddresses = len(allAddr)
 	}
 
-	// Fisher-Yates shuffle the array. We only need to do the first
-	// `numAddresses' since we are throwing the rest.
+	result := make([]*AddrInfo, 0, numAddresses)
+	a.mu.RLock()
```
dshulyak, line 290: it is preferable to always defer unlock for safety if logic allows

p2p/addressbook/addrbook.go
```diff
@@ -283,46 +228,113 @@ func (a *addrBook) pickTried(bucket int) *knownAddress {
...
+
+	if len(a.lastAnchorPeers) == 0 {
+		return addresses
+	}
+
+	anchorPeers := make([]*knownAddress, len(a.lastAnchorPeers), len(a.lastAnchorPeers))
+	copy(anchorPeers, a.lastAnchorPeers)
+	for i := 0; i < a.cfg.AnchorPeersCount; i++ {
+		j := rand.Intn(len(anchorPeers)-i) + i // pick a number between current index and the end.
+		anchorPeers[i], anchorPeers[j] = anchorPeers[j], anchorPeers[i]
```
dshulyak, line 322: are you shuffling? there is a function rand.Shuffle

p2p/addressbook/addrbook.go
```diff
@@ -283,46 +228,113 @@ func (a *addrBook) pickTried(bucket int) *knownAddress {
...
+func (a *AddrBook) BootstrapAddressCache() []*AddrInfo {
+	addresses := a.AddressCache()
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+
+	if len(a.lastAnchorPeers) == 0 {
+		return addresses
+	}
+
+	anchorPeers := make([]*knownAddress, len(a.lastAnchorPeers), len(a.lastAnchorPeers))
```
dshulyak: why do you need to specify both length and capacity if they are equal?

p2p/addressbook/addrbook.go
```diff
@@ -283,46 +228,113 @@ func (a *addrBook) pickTried(bucket int) *knownAddress {
...
+	copy(anchorPeers, a.lastAnchorPeers)
+	for i := 0; i < a.cfg.AnchorPeersCount; i++ {
+		j := rand.Intn(len(anchorPeers)-i) + i // pick a number between current index and the end.
```
dshulyak: why a.rand is not used here?

dshulyak (changes requested):

p2p/addressbook/addrbook.go
abergasov: forget to drop)

p2p/addressbook/addrbook.go
abergasov: replaced to shuffle

p2p/addressbook/addrbook.go
abergasov: structure is `map[peer.ID]*knownAddress`, so we should iterate in loop, cause rand return only int

p2p/addressbook/addrbook.go
abergasov: forget to drop 2)

dshulyak (approved):

abergasov:
bors try

abergasov:
bors try

abergasov:
bors try

abergasov:
bors try

abergasov:
bors merge

Issue #3220, opened 2022-05-22
vm: account abstraction and new tx format


dshulyak:
At genesis we will launch vm with 3 precompiled templates:

- simple (single-sig) wallet
- multisig wallet
- vesting contract

Precompiles will be implemented as a simple golang modules, while trying to preserve boundaries discussed during design.

## Templates

Templates are identified by an address, and will have their own address space.

For genesis we will define 3 constant addresses:
- 0x01 - singlesig wallet
- 0x02 - multisig wallet
- 0x03 - vesting

TODO maybe start fron 0x00, any use case for empty address?

For spawned accounts we will store one of this addresses in the account table.

#### Load

Loads stored state or initializes state using arguments from a self spawn transaction.

#### Parse (ParsePayload)

decodes generic arguments from the concrete payload

- method arguments
- min/max layer
- nonce counter/bitfield
- max gas 
- gas price
- max spend ?

some of the fields might not be specified by the transaction, by design it depends on the template. for example max gas is known in advance, since we are using fixed gas estimator. nonce for spawn transaction is 0.

result of the Parse returned to the mempool.

#### MaxSpend

inputs: selector, method arguments, state

TODO merge it with Parse for now

#### Verify

transaction validity

input: raw transaction and state

TODO this is really weird that we need to read transaction from the end when we parsed everything before the signature.

TODO can we unmarshal signature into concrete type?

#### Exec

Apply changes to the account state, by executing one the methods. 

## Accounts

every account with funds but without attached template is a stub account and needs to be spawned before it can be used.

In the database stub account will have null template field.

### On gossip

Before accepting transaction into the local database it needs to pass validation.

Next sequence of template APIs is executed:
1. Load (use conservative state)
Probably will be implemented by conservative passing AccountState from conservative cache. Or maybe method will accept callback that will load state from conservative state.
2. Parse
Optionally after Parse tx can be validated by the conservative state, but with current API it is simpler after Verify.
3. Verify
After verify tx is saved to the pool, and conservative state may do other things.

If verify fails - transaction is discarded, and peer is blocked (tm).
Otherwise mempool may do anything with transaction.

### On apply

1. Load (non-conservative state)
2. Parse
Most of the transaction will be stored as a blob for sync efficiency, so we will Parse it again. 
3. No Verify
No need to execute signature verification, but need to consume gas.

Verify has access only to immutable state so it is guaranteed to pass.
4. Exec

[...]

PR #3229, opened 2022-06-01
[Merged by Bors] - vm: implementation that supports account abstraction model (includes signle key wallet)

dshulyak:
related https://github.com/spacemeshos/go-spacemesh/issues/3220

- state

account extend with Template and State fields.
stub accounts will have both Template and State as nil. When account is spawned with template - Template will store the address of the template, and State will store mutable and immutable state encoded using scale.

- address

when account is spawned address is computed as `sha256(template_address, spawn_args)`. this must be taken into an account when configuring the coinbase for a node.

- validation and mempool

tx id is computed as a `sha256(raw_tx)`, note that raw_tx is a full body of the transaction including signature.
tx header can't be decoded without executing template code (specifically parse_payload and max_spend methods). 

in the upcoming pr - mempool will call into vm to initialize a validation flow (which includes parsing of the tx header). validation api is a 2 step process:
- parse. returns transaction header that can be validated against conservative state
- verify. executes templates verify method. it is assumed that this method costs much more than parse, so it makes sense to call it only if conservative validation succeeded.

- applying transaction

vm.Apply API method accepts layer id, transactions and rewards. Every transaction loads state, executes parse payload, initializes state and executed. Template/Handler API is specified in `genvm/core/types.go`.

Transaction can be skipped for 3 reasons:
- transaction is malformed (expectation is that such transactions should not be added to the block at all)
- account is not spawned (same as a nonce mismatch)
- nonce doesn't match the expected nonce of the account
- can't cover max gas with the account balance

If transaction is skipped it is returned back to the caller (presumably conservative state and the caller is free to do anything about it). Otherwise transaction is consumed and will pay fee to smeshers (in the followup pr for every consumed transaction we will store execution result). Logic for the fee distribution is not changed.


countvonzero:
some questions:

> - address
when account is spawned address is computed as sha256(template_address, spawn_args). this must be taken into an account when configuring the coinbase for a node.

what do you mean by `this must be taken into an account when configuring the coinbase for a node`
do you expect the coinbase addresses to be stubs?

> verify. executes templates verify method. it is assumed that this method costs much more than parse, so it makes sense to call it only if conservative validation succeeded.

so in the future PR, the happy flow is expected to be
1. conservative state (CS) passes tx to vm for parse()
2. CS checks balance/nonce+bitfield/[min|max]Layer
3. CS passes tx to vm for verify()
4. CS adds tx to mempool

do i understand this correctly?

> Transaction can be skipped for 3 reasons:
> - transaction is malformed (expectation is that such transactions should not be added to the block at all)
> - account is not spawned (same as a nonce mismatch)
> - nonce doesn't match the expected nonce of the account
> - can't cover max gas with the account balance

how about the min/max layer check?

dshulyak:
> what do you mean by this must be taken into an account when configuring the coinbase for a node
do you expect the coinbase addresses to be stubs?

yes, if they want to spend funds they need to be able to self-spawn themself at some point

> so in the future PR, the happy flow is expected to be

in general yes. but i after looking into code i want to keep verify change as a future improvement. i don't want to change conservative cache flow too much 

> how about the min/max layer check?

i think that the block should be invalid if there is such transaction. why would honest smeshers include it? blocks with malformed transactions should probably should be invalid as well

.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
 
 env:
-  go-version: '1.17.6'
+  go-version: '1.18.2'
```
countvonzero, line 4: what features in 18 are you going for? i don't see any in the PR

common/types/address.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/spacemeshos/go-scale"
```
countvonzero, line 8: why not use the go impl of  parity scale codec https://github.com/itering/scale.go?
what are the differences?

common/types/account.go
```diff
@@ -1,10 +1,39 @@
...
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+//go:generate scalegen
+
 // Account represents account state at a certain layer.
 type Account struct {
 	Layer       LayerID
 	Address     Address
 	Initialized bool
```
countvonzero, line 13: what does `Initialized` mean in the account abstraction model?

common/types/bytes.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import "github.com/spacemeshos/go-scale"
+
+// Bytes64 is 64 byte array.
+type Bytes64 [64]byte
+
+// EncodeScale implements scale codec interface.
+func (b *Bytes64) EncodeScale(encoder *scale.Encoder) (int, error) {
+	return scale.EncodeByteArray(encoder, b[:])
```
countvonzero, line 10: the scale lib doesn't support generic byte[] with scalegen?

common/types/hashes.go
```diff
@@ -112,6 +112,11 @@ func CalcHash12(data []byte) (h Hash12) {
 	return
 }
 
+// DecodeScale implements scale codec interface.
+func (h *Hash20) DecodeScale(d *scale.Decoder) (int, error) {
```
countvonzero: maybe a symmetric EncodeScale() for Hash20?

genvm/core/hash.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// Hash bytes into sha256 hash.
+func Hash(bufs ...[]byte) Hash32 {
+	hasher := hash.New()
+	for _, buf := range bufs {
+		hasher.Write(buf)
+	}
+	var rst Hash32
+	hasher.Sum(rst[:])
```
countvonzero: i don't understand this line. why isn't this `rst := hasher.Sum(nil)` or `hasher.Sum(rst[:0])`?

genvm/templates/wallet/handler.go
```diff
@@ -0,0 +1,100 @@
...
+	// TemplateAddress is an address of the Wallet template.
+	TemplateAddress core.Address
+)
+
+type handler struct{}
+
+// Parse header and arguments.
+func (*handler) Parse(ctx *core.Context, method uint8, decoder *scale.Decoder) (header core.Header, args scale.Encodable, err error) {
+	switch method {
+	case 0:
```
countvonzero: how about name constants for methods?

genvm/core/context.go
```diff
@@ -0,0 +1,111 @@
...
+	Template Template
+
+	Account Account
+
+	Header Header
+	Args   scale.Encodable
+
+	// consumed and transferred is for MaxGas/MaxSpend validation
+	consumed    uint64
+	transferred uint64
```
countvonzero, line 25: i recall in the word doc gas price and transfer amount are uint32. what changed?

genvm/core/context.go
```diff
@@ -0,0 +1,111 @@
...
+	c.Account.Balance -= amount
+	account.Balance += amount
+	return nil
+}
+
+// Consume gas from the account after validation passes.
+func (c *Context) Consume(gas uint64) error {
+	amount := gas * c.Header.GasPrice
+	if amount > c.Account.Balance {
+		return ErrNoBalance
```
countvonzero, line 83: is there a need to differentiate out of gas from insufficient balance?

genvm/core/staged_cache.go
```diff
@@ -0,0 +1,62 @@
...
+func (ss *StagedCache) Update(account Account) error {
+	sacc, exist := ss.cache[Address(account.Address)]
+	if !exist || !sacc.Changed {
+		ss.touched = append(ss.touched, Address(account.Address))
+	}
+	ss.cache[Address(account.Address)] = stagedAccount{
+		Account: account,
+		Changed: true,
+	}
+	return nil
```
countvonzero, line 46: doesn't ever return error

genvm/vm.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+	}
+
+	hasher := hash.New()
+	encoder := scale.NewEncoder(hasher)
+	ss.IterateChanged(func(account *core.Account) bool {
+		account.Layer = lid
+		v.logger.With().Debug("update account state", log.Inline(account))
+		err = accounts.Update(tx, account)
+		account.EncodeScale(encoder)
```
countvonzero: if there is a sql error, should it still be included in the diff hash?

genvm/vm.go
```diff
@@ -0,0 +1,307 @@
...
+	} else {
+		template = account.Template
+	}
+	if template == nil {
+		return nil, nil, nil, fmt.Errorf("%w: %s", core.ErrNotSpawned, principal.String())
+	}
+
+	handler := reg.Get(*template)
+	if handler == nil {
+		return nil, nil, nil, fmt.Errorf("%w: unknown template %s", core.ErrMalformed, *account.Template)
```
countvonzero: should be just `*template` in the returned error

genvm/templates/wallet/handler.go
```diff
@@ -0,0 +1,100 @@
...
+
+const (
+	// TotalGasSpawn is consumed from principal in case of successful spawn.
+	TotalGasSpawn = 200
+	// TotalGasSpend is consumed from principal in case of successful spend.
+	TotalGasSpend = 100
+)
+
+func init() {
+	TemplateAddress[len(TemplateAddress)-1] = 1
```
countvonzero, line 27: please add comment on why the last bit is 1. i forgot already.

genvm/templates/wallet/wallet.go
```diff
@@ -0,0 +1,47 @@
+package wallet
+
+import (
+	"fmt"
+
+	"github.com/oasisprotocol/curve25519-voi/primitives/ed25519"
```
countvonzero, line 6: this is a little confusing. what's the reason not using `github.com/spacemeshos/ed25519`?

genvm/core/context.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	c.consumed += amount
+	if c.consumed > c.Header.MaxGas {
+		return ErrMaxGas
```
countvonzero: i thought we should charge gas when this happens? i.e. the account should have zero balance after this? 
and c.consumed should be set to c.Header.MaxGas

genvm/vm.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+		fee, err := ctx.Apply(ss)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %s", core.ErrInternal, err.Error())
+		}
+		fees += fee
+	}
+
+	// TODO(dshulyak) why do we fail if there are no rewards? can we just burn them?
+	if len(rewards) > 0 {
```
countvonzero, line 158: why is this block valid if there are no rewards?

genvm/vm.go
```diff
@@ -0,0 +1,307 @@
...
+
+	var principal core.Address
+	if _, err := principal.DecodeScale(decoder); err != nil {
+		return nil, nil, nil, fmt.Errorf("%w failed to decode principal: %s", core.ErrMalformed, err)
+	}
+	method, _, err := scale.DecodeCompact8(decoder)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("%w: failed to decode method selector %s", core.ErrMalformed, err.Error())
+	}
+	account, err := loader.Get(principal)
```
countvonzero, line 255: what would happen to a stub account here?

countvonzero (approved):

common/types/account.go
dshulyak: Initialized is used with old vm. i will remove this field in next pr

common/types/address.go
dshulyak: but thats not from parity. the implementation of the codec in that repo doesn't make much sense. the simplest decoding of boolean https://github.com/itering/scale.go/blob/ba81649be3be3d65b5ffd78ff32e625edcf6223b/types/base.go#L95-L98 is completely misimplemented

common/types/bytes.go
dshulyak: in this case i need this field to implement scale interface. if bytes field is a part of struct scalegen will derive codec for it correctly https://github.com/spacemeshos/go-scale/blob/master/examples/bytes.go , but `type Bytes []byte` is not supported yet

but beside that signature should not be encoded as a []byte, because it will be length prefixed

common/types/hashes.go
dshulyak: this is probably some leftover, i didn't intend to add codec fields for hash20

genvm/core/context.go
dshulyak: that doc was written with an assumption that we are using codec that doesn't compress integers. since uint64 won't use additional space we can use it (when encoded)

.github/workflows/ci.yml
dshulyak: both generics and fuzzing.

go-scale uses generics and fuzzing, and there also a couple of fuzz functions in this repo

genvm/core/context.go
dshulyak: i think out of gas is either about balance or max gas. there is a different error for second case

genvm/core/hash.go
dshulyak: thanks, thats a bug. i used this function both for signing and verifying

genvm/templates/wallet/wallet.go
dshulyak: oasis library is much more efficient, we are also using it for ecvrf. and beside that i in general have more confidence in it, since it is used in existing blockchain projects (including tendermint)

genvm/vm.go
dshulyak: it actually won't matter since we will terminate execution if there is an error, but i can skip

genvm/vm.go
dshulyak: nothing special. it loads state that it has (it might have balance for example, or all fields can be zero)

genvm/vm.go
dshulyak: why is it invalid? can't there be a consensus on a block without rewards? do we verify it when accepting a block?

genvm/templates/wallet/handler.go
dshulyak: byte

genvm/core/staged_cache.go
dshulyak: yeah, i initially assumed that it may write to database if there will be problems with memory

common/types/hashes.go
dshulyak: removed it for now, maybe will add later if it will be used for smth

genvm/core/context.go
dshulyak: so are you saying that we should consume full account balance or only c.Header.MaxGas amount? i think latter makes sense, it will be probably more relevant for full svm, but i will make a change

genvm/templates/wallet/wallet.go
dshulyak: spacemeshos will remain to be used for recovery, unless someone extends oasis

dshulyak:
bors merge

genvm/vm.go
countvonzero: we don't verify it now, but think we should. 

i don't know if rewards was part of the consensus. but as current code stands, producing a block without rewards is not following the protocol.

genvm/templates/wallet/wallet.go
countvonzero: what do you mean by "used for recovery"? 
my motivation is that if oasis is more efficient, can the core go-spacemesh code use it?

common/types/address.go
countvonzero: interesting. substrate listed it as their go impl of scale https://docs.substrate.io/v3/advanced/scale-codec/
too bad.

genvm/templates/wallet/wallet.go
dshulyak: public key recovery. public key recovery is impossible with canonical ed25519 signature

genvm/vm.go
dshulyak: i opened this issue so that we wont forget https://github.com/spacemeshos/go-spacemesh/issues/3251

but i really doubt that it makes sense to invalidate such block. for example what is the difference between block with empty rewards and the block with rewards given to random non-existing addresses? but lets continue in the issue

PR #3231, opened 2022-06-02
[Merged by Bors] - p2p write address book file atomically

### LINKED ISSUE #2886: p2p/peerexchange: write addrbook file atomically
peerexchange periodically writes addrbook to the file. The current approach is naive and in case of the update may lead to a corrupted state of the file. If update fails for any reason (power failure, the software is stopped) new file may be overwritten only partially, in such case previous version will be discarded but the new one is not yet written and persisted on disk. Also CRC check is needed to ensure that what we read wasn't corrupted due to the disk failure.

We do not terminate software yet in case of such failure, but it will force the node to reinitialize addrbook from scratch using boot nodes.

abergasov:
## Motivation
Closes #3231

## Changes
write file atomically, add crc32 checksum to validate content

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/p2p/addressbook/store.go
+++ b/p2p/addressbook/store.go
@@ -3,17 +3,17 @@ package addressbook
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"hash/crc32"
+	"io/ioutil"
 	"os"
 	"time"
 
-	"github.com/pkg/errors"
-
 	"github.com/libp2p/go-libp2p-core/peer"
-
 	atomicfile "github.com/natefinch/atomic"
+	"github.com/pkg/errors"
 
 	"github.com/spacemeshos/go-spacemesh/log"
 )
@@ -64,17 +64,6 @@ func (a *AddrBook) persistPeers() {
 		}
 	}
 
-	//w, err := os.Create(path)
-	//if err != nil {
-	//	a.logger.Error("Error creating file: %v", err)
-	//	return
-	//}
-	//enc := json.NewEncoder(w)
-	//defer w.Close()
-	//if err = enc.Encode(&sam); err != nil {
-	//	a.logger.Error("Failed to encode file %s: %v", a.path, err)
-	//	return
-	//}
 	data, err := json.Marshal(sam)
 	if err != nil {
 		a.logger.Error("Failed to encode data %s", err)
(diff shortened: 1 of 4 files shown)


p2p/addressbook/store.go
```diff
@@ -58,19 +64,53 @@ func (a *AddrBook) persistPeers(path string) {
...
-	defer w.Close()
-	if err := enc.Encode(&sam); err != nil {
-		a.logger.Error("Failed to encode file %s: %v", path, err)
+	if err = a.atomicallySaveToFile(data); err != nil {
+		a.logger.Error("Failed to encode file %s: %v", a.path, err)
 		return
 	}
 }
 
+func (a *AddrBook) atomicallySaveToFile(data []byte) error {
```
dshulyak, line 77: why backup is needed? documentation for WriteFile says https://github.com/natefinch/atomic/blob/master/atomic.go#L12-L17

> // WriteFile atomically writes the contents of r to the specified filepath.  If
// an error occurs, the target file is guaranteed to be either fully written, or
// not written at all.  WriteFile overwrites any file that exists at the
// location (but only if the write fully succeeds, otherwise the existing file
// is unmodified).
func WriteFile(filename string, r io.Reader) (err error) {

regarding crc check it is used to detect unexpected changes on disk (either disk malfunction, or someone messed with data accidentally). what makes sense to me:

on write:
- compute crc from json data
- write crc || json data

on read:
- decode crc from first 4(or 8) bytes
- compute another crc from the rest of the content
- compare both with each other
- if they are not equal - file can be discarded, as it got corrupted

abergasov: after write `crc || json data` to file, we can straightaway load data and check crc32
if mismatch - then rollback to backup

dshulyak: what does it solve?

abergasov: in case file was corrupted after save - we can restore backup, instead reset book

dshulyak: corruption may happen not only during write, but at any point in time while file is stored on the disk. verifying that file content is consistent right after writing it doesn't do much at all

abergasov: should we give possibility to node rollback to previous version in case last file is corrupted?

for example: 

1. peers.json is broken
2. try peers.json.backup. if ok - use this
3. peers.json.backup is broken - reset book

dshulyak: in my opinion backups don't add any additional safety in this case, as they are stored in the same place. if main file got corrupted, what are the chances that backup wasn't corrupted? we can store 100 backups for example and hope that atleast one of them will be in consistent state, but it don't improve fault tolerance that much 

in my understanding adding crc is useful not because we can eliminate such failure completely, but to speed up recovery if it occurs. without crc  node will have to connect with peers from potentially invalid book, but all records might be invalid in some subtle ways. node will have to spend significant amount of time removing them from the book.

p2p/addressbook/store.go
```diff
@@ -1,13 +1,19 @@
...
 	"encoding/json"
 	"fmt"
+	"hash/crc32"
+	"io/ioutil"
 	"os"
 	"time"
 
 	"github.com/libp2p/go-libp2p-core/peer"
+	atomicfile "github.com/natefinch/atomic"
+	"github.com/pkg/errors"
```
dshulyak, line 16: why to use this module?

abergasov: for wrap errors and keep erorr trace
like 
```
errors.Wrap(err, "Something")
```

dshulyak: does it work with errors.Is/errors.As ?

abergasov: yes, fully
they use native go `errors` for this methods

p2p/addressbook/store.go
```diff
@@ -169,19 +197,18 @@ func (a *AddrBook) decodeFrom(path string) error {
...
 		return
 	}
 
 	for {
 		select {
 		case <-ticker.C:
-			a.persistPeers(path)
-			a.logger.Debug("saved peers to file %v", path)
+			a.persistPeers()
+			a.logger.Debug("saved peers to file %v", a.path)
```
countvonzero: i think you need to at least return the error from persistPeers to know if it's successful before logging this

p2p/addressbook/store.go
```diff
@@ -104,8 +118,22 @@ func (a *AddrBook) decodeFrom(path string) error {
 	}
 	defer r.Close()
 
+	data, err := ioutil.ReadAll(r)
+	if err != nil {
+		return errors.Wrap(err, "error reading file")
+	}
+
+	if len(data) < 4 {
```
countvonzero, line 126: what is 4? plz give it a meaning

abergasov: before put data to file, we calc crc32. In golang lib it return uin32, this is 4 bytes.
So in file goes something like `123{some_json}`

So here we checking, is length enough? If so - split to `123` and `{some_json}`, verify that json str is not corrupted. 
123 as uint32 takes 4 bytes

to avoid panic acces out of range - check lenght

countvonzero (approved):

abergasov:
bors try

dshulyak (approved):

abergasov:
bors merge

PR #3237, opened 2022-06-03
[Merged by Bors] - flacky test fix 2

### LINKED ISSUE #3236: flaky test: nodechecker_test.go
## Description

https://github.com/spacemeshos/go-spacemesh/runs/6719396707?check_suite_focus=true

```
--- FAIL: TestHost_LocalAddressChange (0.00s)
    --- FAIL: TestHost_LocalAddressChange/change_local_address_of_node_and_new_not_acceptable (5.36s)
        nodechecker_test.go:152: 
            	Error Trace:	nodechecker_test.go:152
            	Error:      	Not equal: 
            	            	expected: 1
            	            	actual  : 0
            	Test:       	TestHost_LocalAddressChange/change_local_address_of_node_and_new_not_acceptable
            	Messages:   	nodeB should have 1 addresses
FAIL
FAIL	github.com/spacemeshos/go-spacemesh/p2p	10.[77](https://github.com/spacemeshos/go-spacemesh/runs/6727716867?check_suite_focus=true#step:5:78)8s
```

abergasov:
## Motivation
Closes #3236

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/p2p/nodechecker_test.go
+++ b/p2p/nodechecker_test.go
@@ -109,7 +109,7 @@ func TestHost_LocalAddressChange(t *testing.T) {
 
 			require.Eventually(t, func() bool {
 				return len(nodeB.host.Peerstore().Addrs(nodeA.host.ID())) == 1
-			}, 2*time.Second, 100*time.Millisecond, "nodeB should have 1 address of nodeA")
+			}, 4*time.Second, 100*time.Millisecond, "nodeB should have 1 address of nodeA")
 
 			// assing new address to nodeA, expect nodeB to be notified.
 			require.NoError(t, nodeA.host.Network().Listen(addrANew), "nodeA should be able to listen on new address")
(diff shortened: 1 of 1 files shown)


abergasov:
bors try

p2p/nodechecker_test.go
```diff
@@ -109,7 +109,7 @@ func TestHost_LocalAddressChange(t *testing.T) {
 
 			require.Eventually(t, func() bool {
 				return len(nodeB.host.Peerstore().Addrs(nodeA.host.ID())) == 1
-			}, 2*time.Second, 100*time.Millisecond, "nodeB should have 1 address of nodeA")
+			}, 4*time.Second, 100*time.Millisecond, "nodeB should have 1 address of nodeA")
```
dshulyak, line 112: doesn't it seem strange that 2s is not enough for address to update?

dshulyak (approved):

p2p/nodechecker_test.go
abergasov: usually enough 100ms. But when i run `go test --count 100` one time get fail here. Probably, as we run multiply times, it generate high load at machine, and not enough time for exchange messages between nodes.

as we run it in github ci, and do not know in which condition instance - better set more time, in most cases it will finish early than in 4 sec. But for cases when performance is low - it may help.

abergasov:
bors try

p2p/nodechecker_test.go
countvonzero: > Probably, as we run multiply times, it generate high load at machine, and not enough time for exchange messages between nodes.

isn't this unittest? what do you mean between nodes? 
without looking at the code, is there potentially lock contention?

abergasov: yes, this is unit test. Usually was used `mocknet` tool for test behavior in unit tests. But there too much mocks and you can't inject some modules.

So it actually this test create 2 nodes on same host, but on different ports. And after some actions with nodeA check that nodeB catch changes

countvonzero (approved):

abergasov:
bors merge

PR #3243, opened 2022-06-06
[Merged by Bors] - p2p  periodically check for dead|alive peers also update address

### LINKED ISSUE #2891: p2p, peerexchange: after initial bootstrap periodically connect with peers
We maintain configured number of outbound connections. Once this number goes below required we will trigger peerexcahgen bootstrap process, and this process will run until we have required number of connections again.

The limitation is that after we finished with this we won't learn about new nodes and won't propagate our address to other nodes.

The solution should start a sub-process in peerexchange.Discovery that will periodically (every 10m for example) will connect with 1-3 random peers from address book. This will update local address book and notify other peers about our node.

abergasov:
## Motivation
Closes #2891

## Changes
added test for check that periodically check for dead|alive peers also update address…

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/p2p/peerexchange/bookchecker_test.go
+++ b/p2p/peerexchange/bookchecker_test.go
@@ -2,6 +2,7 @@ package peerexchange
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


p2p/peerexchange/bookchecker_test.go
```diff
@@ -154,6 +155,93 @@ func TestDiscovery_GetRandomPeers(t *testing.T) {
...
+		bAddresses := nodeB.book.GetAddresses()
+		if len(aAddresses) != 1 || len(bAddresses) != 1 {
+			return false
+		}
+		return aAddresses[0].ID.String() == nodeB.host.ID().String() && bAddresses[0].ID.String() == nodeA.host.ID().String()
+	}, 4*time.Second, 100*time.Millisecond, "nodeA should have only addresss of nodeB and nodeB should have only addresss of nodeA")
+
+	// disconnect nodeA and node. sleep 3 sec need for addresses become outdated and ready for check
+	require.NoError(t, nodeB.host.Network().ClosePeer(nodeA.host.ID()))
+	time.Sleep(3 * time.Second)
```
countvonzero: rm

p2p/peerexchange/bookchecker_test.go
```diff
@@ -154,6 +155,93 @@ func TestDiscovery_GetRandomPeers(t *testing.T) {
...
+	wrapDiscovery(instances)
+
+	// check that after bootstrap, nodeB has only addresss of nodeA and nodeA has only addresss of nodeB
+	require.Eventually(t, func() bool {
+		aAddresses := nodeA.book.GetAddresses()
+		bAddresses := nodeB.book.GetAddresses()
+		if len(aAddresses) != 1 || len(bAddresses) != 1 {
+			return false
+		}
+		return aAddresses[0].ID.String() == nodeB.host.ID().String() && bAddresses[0].ID.String() == nodeA.host.ID().String()
```
countvonzero: it seems this should be checked outside of require.Eventually(). once we are sure they each have size-1 address book, we should immediately exit this require.Eventually()

p2p/peerexchange/bookchecker_test.go
```diff
@@ -154,6 +155,93 @@ func TestDiscovery_GetRandomPeers(t *testing.T) {
...
+		if len(addresses) != 3 {
+			return false
+		}
+		for _, address := range addresses {
+			valid := address.ID.String() == hostsAdditional[0].ID().String() ||
+				address.ID.String() == hostsAdditional[1].ID().String() ||
+				address.ID.String() == hosts[1].ID().String()
+			if !valid {
+				return false
+			}
```
countvonzero: same here. once we are sure about the size. we should exit require.Eventually() and check the expected values outside.

p2p/peerexchange/bookchecker_test.go
```diff
@@ -154,6 +155,93 @@ func TestDiscovery_GetRandomPeers(t *testing.T) {
...
+		addresses := nodeB.book.AddressCache()
+		if len(addresses) != 3 {
+			return false
+		}
+		for _, address := range addresses {
+			valid := address.ID.String() == hostsAdditional[0].ID().String() ||
+				address.ID.String() == hostsAdditional[1].ID().String() ||
+				address.ID.String() == hosts[1].ID().String()
+			if !valid {
+				return false
```
countvonzero: ditto

p2p/peerexchange/bookchecker_test.go
abergasov: this is need cause for get addresses, we watch at `address.LastUsage: time.Time` 
I cant modify address in addressBook, cause it in private struct. Only write special method, for tests, but it is not very good. Better just wait. This sleep not mean that some request will be executed, so it is not a crime here)

countvonzero: i don't usually speak in absolute but this is a special case. putting sleep in tests is always bad. this is an indication of the code not being testable, which is a real problem. 

please change the code such that you can get a signal that address book is changed (using chan?). it's fine to add harness like that for testing.
see https://github.com/spacemeshos/go-spacemesh/issues/2067

abergasov: yes, if find way to avoid sleep)

dshulyak (approved):

countvonzero (approved):
thank you

abergasov:
bors merge

PR #3244, opened 2022-06-07
[Merged by Bors] - 2703 hash not found

### LINKED ISSUE #2703: devnet: log "hash not found in response from peer" & "remote peer requested non existing hash"

evt:
## Motivation
Closes #2703 

## Changes
Introduced hash-to-peers cache in a fetcher to know which peer to poll for specific caches to avoid `hash not found` warning as much as possible. The issue still remains for ATXDB and PoetDB but still it has improved a lot.

Cache hit rate is as follows:
```
fbe52.layerFetcher .fetch - "hitrate":"100.00"
fbe52.layerFetcher .fetch - "hitrate":"52.94"
fbe52.layerFetcher .fetch - "hitrate":"45.00"
fbe52.layerFetcher .fetch - "hitrate":"42.86"
fbe52.layerFetcher .fetch - "hitrate":"58.62"
fbe52.layerFetcher .fetch - "hitrate":"67.57"
fbe52.layerFetcher .fetch - "hitrate":"73.91"
fbe52.layerFetcher .fetch - "hitrate":"77.78"
fbe52.layerFetcher .fetch - "hitrate":"80.95"
fbe52.layerFetcher .fetch - "hitrate":"83.10"
fbe52.layerFetcher .fetch - "hitrate":"85.00"
fbe52.layerFetcher .fetch - "hitrate":"86.36"
fbe52.layerFetcher .fetch - "hitrate":"79.38"
fbe52.layerFetcher .fetch - "hitrate":"77.78"
fbe52.layerFetcher .fetch - "hitrate":"79.44"
fbe52.layerFetcher .fetch - "hitrate":"79.63"
fbe52.layerFetcher .fetch - "hitrate":"81.20"
fbe52.layerFetcher .fetch - "hitrate":"82.26"
fbe52.layerFetcher .fetch - "hitrate":"84.17"
fbe52.layerFetcher .fetch - "hitrate":"83.57"
fbe52.layerFetcher .fetch - "hitrate":"83.69"
fbe52.layerFetcher .fetch - "hitrate":"83.22"
fbe52.layerFetcher .fetch - "hitrate":"83.33"
fbe52.layerFetcher .fetch - "hitrate":"83.45"
fbe52.layerFetcher .fetch - "hitrate":"83.56"
fbe52.layerFetcher .fetch - "hitrate":"83.67"
fbe52.layerFetcher .fetch - "hitrate":"83.78"
fbe52.layerFetcher .fetch - "hitrate":"83.89"
fbe52.layerFetcher .fetch - "hitrate":"84.00"
fbe52.layerFetcher .fetch - "hitrate":"84.11"
```

## DevOps Notes


fetch/metrics/hash_to_peers_cache.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var (
+	totalHits   = metrics.NewCounter("total_hits", subsystem, "Total cache hits", nil)
+	totalMisses = metrics.NewCounter("total_misses", subsystem, "Total cache misses", nil)
+)
+
+// HashToPeersCacheCollector implement metrics.Reporter
+// that keeps track of the number of messages sent and received per protocol.
+type HashToPeersCacheCollector struct{}
```
dshulyak: whats the point of this object? why not to use totalHits and totalMisses directly

fetch/metrics/hash_to_peers_cache.go
```diff
@@ -0,0 +1,34 @@
...
+
+// NewHashToPeersCacheCollector creates a new HashToPeersCacheCollector.
+func NewHashToPeersCacheCollector() *HashToPeersCacheCollector {
```
dshulyak: even if you need an object for some reason, what the point of empty initializer? why would someone even use it instead of zero value?

dshulyak:
> you actually don't need to do this cleaning. if you evict hash at the top level, you don't need to care about individual hash. it'll be bound by the number of peers we can connect to at p2p config.

not sure if this is still relevant, but we specify only target outbound peers, which is way different than the max peers. max peers can be over 100.

dshulyak:
thats unfortunate that original idea with passing peer as a value didn't workout, i think it would be much safer and nicer then shared state approach. 

can you please describe how eviction is handled, i wasn't able to understand from code

evt:
> thats unfortunate that original idea with passing peer as a value didn't workout, i think it would be much safer and nicer then shared state approach.

as discussed earlier with @countvonzero shared state is needed to load balance between peers so that all requests don't go to the first one having all the hashes. Or you think it can be done in some other way?

> 
> can you please describe how eviction is handled, i wasn't able to understand from code

eviction is not there yet. Do you any ideas for a better approach than looping over hashes and evicting those older certain timestamp?

evt:
> > you actually don't need to do this cleaning. if you evict hash at the top level, you don't need to care about individual hash. it'll be bound by the number of peers we can connect to at p2p config.
> 
> not sure if this is still relevant, but we specify only target outbound peers, which is way different than the max peers. max peers can be over 100.

hmm, did you know about this @countvonzero ?

countvonzero:
> thats unfortunate that original idea with passing peer as a value didn't workout, i think it would be much safer and nicer then shared state approach.

i have not looked at the code. what is your concern about the cache approach, other than the common problem with cache memory leak? i think mem leak we can address by fixed-size cache. @dshulyak 

let me try to explain the history and rationale. @evt first started with passing the peer. i made this suggestion of using a cache approach and he took my suggestion.

the low-level fetch library was written with 2 principals (as far as i can tell). it tries to
- batch requests instead of greedily send to peer
- evenly distribute the requests amongst peers

passing peer from gossip handler / sync would work, but sub-optimally because
- the hash->peer relationship a one-time knowledge and expires as soon as the request finishes
- with sync, it will concentrate all requests to the same peer. (this is fixable but the code will get even uglier)

the cache approach is more general and is shared across requests. the relationship between data objects is also loosely based on time. for example, ballots in the same layer tend to reference the same set of data.
 
> can you please describe how eviction is handled, i wasn't able to understand from code

i haven't look at the code. but i think the eviction part of this PR is WIP still.

countvonzero:
> > > you actually don't need to do this cleaning. if you evict hash at the top level, you don't need to care about individual hash. it'll be bound by the number of peers we can connect to at p2p config.
> > 
> > 
> > not sure if this is still relevant, but we specify only target outbound peers, which is way different than the max peers. max peers can be over 100.
> 
> hmm, did you know about this @countvonzero ?

as long as we evict hashes correctly in a FIFO approach, is 10 that different from 100?

evt:
> > > > you actually don't need to do this cleaning. if you evict hash at the top level, you don't need to care about individual hash. it'll be bound by the number of peers we can connect to at p2p config.
> > > 
> > > 
> > > not sure if this is still relevant, but we specify only target outbound peers, which is way different than the max peers. max peers can be over 100.
> > 
> > 
> > hmm, did you know about this @countvonzero ?
> 
> as long as we evict hashes correctly in a FIFO approach, is 10 that different from 100?

I wonder how much over 100 it can be... 🤔

dshulyak:
regarding peers: "true" max peers are set using HighPeers option and it is set to 100 by default

in my opinion it is better to avoid any shared state if there is an option. it adds more problems (such as mem management, eviction and synchronization). and it is easier to provide bug-free implementation without using shared state.

> batch requests instead of greedily send to peer

it is definitely possible to aggregate requests to the same peer, so batching should be supported. 
but i also would argue that batching doesn't do much because all requests are random reads and we don't create any expensive connections.

> evenly distribute the requests amongst peers

as you noted it is fixable by providing a set of peers that own the data, which doesn't seem very ugly to me.

beside that i probably don't understand why this state needs to be shared between requests. if current approach works well and there is an easy solution for eviction - no reason to switch to a different approach obviously.

p2p/peer.go
```diff
@@ -0,0 +1,14 @@
+package p2p
+
+import "github.com/libp2p/go-libp2p-core/peer"
+
+// Peer is an alias to libp2p's peer.ID.
+type Peer = peer.ID
+
+// AnyPeer is used when peer doesn't matter.
+const AnyPeer Peer = ""
```
countvonzero: how about renaming this to NoPeer.
the idea of any peer is specific to the fetch package

layerfetcher/layers.go
```diff
@@ -337,11 +337,16 @@ func (l *Logic) fetchLayerData(ctx context.Context, logger log.Log, layerID type
 	}
 	l.mutex.Unlock()
 
+	l.fetcher.TrackBallotsPeer(ctx, peer, ballotsToFetch)
+
 	logger.With().Debug("fetching new ballots", log.Int("to_fetch", len(ballotsToFetch)))
-	if err := l.GetBallots(ctx, ballotsToFetch); err != nil {
+	if err := l.GetBallots(ctx, ballotsToFetch, peer); err != nil {
```
countvonzero: i don't understand this, and the need to add peer to the APIs.
you already register hashes with the peer in the cache. why do you need to pass the peer again here?

layerfetcher/layers.go
```diff
@@ -659,3 +670,18 @@ func (l *Logic) GetPoetProof(ctx context.Context, id types.Hash32) error {
 	}
 	return nil
 }
+
+// TrackBlocksPeer is a wrapper around fetcher's TrackBlocksPeer.
+func (l *Logic) TrackBlocksPeer(ctx context.Context, peer p2p.Peer, ids []types.BlockID) {
```
countvonzero: let's not do one-line wrapper. just inline the methods...

fetch/fetch.go
```diff
@@ -189,21 +196,42 @@ type Fetch struct {
...
 	doneChan             chan struct{}
 	dbLock               sync.RWMutex
+	// FIFO cache of hash->[peer set] to keep track of peers to pull blocks/ballots/txs from
+	hashToPeers      map[types.Hash32]*hashPeers
+	hashToPeersM     sync.RWMutex
+	hashToPeersStats CacheStats
+}
+
+// CacheStats stores FIFO cache statistics.
+type CacheStats struct {
```
countvonzero: please move all the cache related stuff to a new file. readers of fetch.go doesn't really need to know how cache works. 

i think for memory management we should do a size-based. can you try to adapt to using *lru.Cache (see AtxCache) and use a size that's adequate based on your systests? it's not FIFO but close enough.

layerfetcher/layers.go
```diff
@@ -337,11 +337,16 @@ func (l *Logic) fetchLayerData(ctx context.Context, logger log.Log, layerID type
 	}
 	l.mutex.Unlock()
 
+	l.fetcher.TrackBallotsPeer(ctx, peer, ballotsToFetch)
```
countvonzero: do this in `receiveLayerContent` after `extract extractPeerResult()`
there you should just gather a big list of hashes (ballots/blocks) and register that peer with those hashes.
the API in fetch for registering hashes does not need to know whether what data type this hash is.

fetch/fetch.go
```diff
@@ -479,26 +507,115 @@ func (f *Fetch) requestHashBatchFromPeers() {
...
+
+	for _, req := range requests {
+		hashPeersMap, exists := f.hashToPeers[req.Hash]
+		var hashPeers []p2p.Peer
+		if exists {
+			f.HashToPeersCacheHit()
+			hashPeers = peerMapToList(hashPeersMap.peers)
+		}
+
+		f.HashToPeersCacheMiss()
```
countvonzero: else

fetch/fetch.go
```diff
@@ -479,26 +507,115 @@ func (f *Fetch) requestHashBatchFromPeers() {
...
+
+		var p p2p.Peer
+		if len(hashPeers) > 0 {
+			p = hashPeers[rng.Intn(len(hashPeers))]
+		} else {
+			p = GetRandomPeer(f.net.GetPeers())
+		}
```
countvonzero, line 498: you should be able to combine the two if/else blocks above

fetch/fetch.go
```diff
@@ -479,26 +507,115 @@ func (f *Fetch) requestHashBatchFromPeers() {
...
-func (f *Fetch) sendBatch(requests []requestMessage) {
+func (f *Fetch) organizeRequests(requests []requestMessage) map[p2p.Peer][][]requestMessage {
+	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
+	peer2requests := make(map[p2p.Peer][]requestMessage)
+
+	// max retries to find another peer with fewer requests attached
+	const maxRetries = 3
+
+	for _, req := range requests {
+		hashPeersMap, exists := f.hashToPeers[req.Hash]
```
countvonzero: don't you need to take the read lock?

fetch/fetch.go
```diff
@@ -479,26 +507,115 @@ func (f *Fetch) requestHashBatchFromPeers() {
...
 }
 
-// sendBatch dispatches batched request messages.
-func (f *Fetch) sendBatch(requests []requestMessage) {
+func (f *Fetch) organizeRequests(requests []requestMessage) map[p2p.Peer][][]requestMessage {
+	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
+	peer2requests := make(map[p2p.Peer][]requestMessage)
+
+	// max retries to find another peer with fewer requests attached
+	const maxRetries = 3
```
countvonzero: unused?

fetch/fetch.go
```diff
@@ -650,3 +756,90 @@ func (f *Fetch) GetHash(hash types.Hash32, h Hint, validateHash bool) chan HashD
 
 	return resChan
 }
+
+func (f *Fetch) mapPeerToHash(hash types.Hash32, peer p2p.Peer) {
```
countvonzero: these functions are repetitive. please simply organize a big list of hashes on the caller side.
fetch just needs to have one interface like `RegisterPeerHashes(p2p.Peer, []types.Hash32)`

fetch/fetch.go
```diff
@@ -479,26 +507,115 @@ func (f *Fetch) requestHashBatchFromPeers() {
...
+			}
+		}
+
+		if len(requests) == 0 {
+			break
 		}
-		f.sendBatch(requestList[i:j])
+
+		retries++
+		if retries > f.cfg.MaxRetriesForBatch {
```
countvonzero: why should we need to add another level of retries?
per peer and per request seems adequate already. whenever a request fails, it'll be back to the active queue and go through this again anyway. it's set to 100 times now, but i am hoping with you change we can reduce it to something like 5.

countvonzero (changes requested):
i have not done a super-detailed review. just some big ticket items that need addressing for simplifying code.

countvonzero:
> in my opinion it is better to avoid any shared state if there is an option. it adds more problems (such as mem management, eviction and synchronization). and it is easier to provide bug-free implementation without using shared state.

i recommended using lru by hashicorp. i thought that was simple enough.


> beside that i probably don't understand why this state needs to be shared between requests. 

for me, this will come in handy handling gossiped proposals. a node will receive multiple gossip proposals concurrently, and they tend to fetch the same set of blocks/ballots/txs. so when we register those hashes referenced in the proposals/ballots after we decode the proposals, we can distribute those requests across multiple peers. this PR only focus on the sync flow (i think).

fetch/metrics/hash_to_peers_cache.go
evt: fixed

fetch/metrics/hash_to_peers_cache.go
evt: there was a point but not after simplifying it :) removed

p2p/peer.go
evt: sure :) done.

layerfetcher/layers.go
evt: `GetBallots` calls `HandleBallotData` after response received where ballot blocks are fetched and they need peer aswell - https://github.com/spacemeshos/go-spacemesh/blob/2703-hash-not-found-new/proposals/handler.go#L399

layerfetcher/layers.go
evt: to clarify: do you mean embedding `fetcher` into `Logic` like this?
```
type Logic struct {
	fetch.Fetcher
...
```
?

fetch/fetch.go
evt: > please move all the cache related stuff to a new file. readers of fetch.go doesn't really need to know how cache works.

this will be `CacheStats` moved only as `HashToPeersCacheHit` / `HashToPeersCacheMiss` / `getHashToPeersCacheStats` are `Fetch` methods and probably no point to move object methods partly to a new file, no?

fetch/fetch.go
evt: fixed, cheers

fetch/fetch.go
evt: good point, fixed, cheers

fetch/fetch.go
evt: fixed, cheers

fetch/fetch.go
evt: fixed, thanks

layerfetcher/layers.go
countvonzero: first of all, you don't need to pass peer to `GetBallots()` because you can get peers that has this hash from the cache in `getHahes()` and call `receiver(ctx, res.Data)` with any of them.

but you can simplify this even more by adding an interface `AddPeersFromHash()` to `system.Fetcher`. so in `HandleBallotData()` before calling `processBallot`, you can do the following
```
bidHash := b.ID().AsHash32()
h.fetcher.AddPeersFromHash(bidHash, collectHashes(b))
```
in fetcher.go for `AddPeersFromHash` you just get the peers registered for `bidHash` and add them to each of the hashes in the ballot.
the idea is, whoever peers give you this ballot ID, they should have all the hashes contained in the ballot.

`collectHashes()` is just gathering all the hashes in the ballot. this way you don't have to pass peer downstream to from `processBallot` because you have pre-registered all the hashes in the ballot.
```
func collectHashes(b *types.Ballot) []types.Hash32 {
	hashes := types.BlockIDsToHashes(ballotBlockView(b))
	if b.RefBallot != types.EmptyBallotID {
		hashes = append(hashes, b.RefBallot.AsHash32())
	}
	hashes = append(hashes, b.Votes.Base.AsHash32())
	return hashes
}
```

fetch/fetch.go
countvonzero: no. i mean move the datastructure and CacheStats. take a look of `AtxCache` impl and add `CacheStats` to that. this cache will have API like below
```
TrackHashes([]types.Hash32, p2p.Peer)
Get(types.Hash32)[]p2p.Peer
```

layerfetcher/layers.go
countvonzero: that's not what i meant.

you are already doing what i meant by always doing `l.fetcher.TrackXXXPeer()` in the code. these 3 methods are not used.

fetch/fetch.go
countvonzero: or just `Add` and `Get`. i tend to be too verbose

layerfetcher/layers.go
evt: thanks, done

layerfetcher/layers.go
evt: If I don't have it then the following compile error is produced:
```
cmd/node/node.go:583:27: cannot use layerFetch (variable of type *layerfetcher.Logic) as type system.Fetcher in assignment:
        *layerfetcher.Logic does not implement system.Fetcher (missing AddPeersFromHash method)
```
here:
```
	layerFetch := layerfetcher.NewLogic(ctx, app.Config.FETCH, poetDb, atxDB, msh, app.host, dataHanders, dbStores, app.addLogger(LayerFetcher, lg))
	fetcherWrapped.Fetcher = layerFetch
```

fetch/fetch.go
evt: done, cheers

layerfetcher/layers.go
evt: done, thanks

fetch/fetch.go
evt: done, thanks

fetch/fetch.go
evt: hmm, at that moment I was sure it was needed :) removed.

fetch/fetch.go
```diff
@@ -130,6 +132,7 @@ type responseBatch struct {
 type Config struct {
 	BatchTimeout         int // in milliseconds
 	MaxRetriesForPeer    int
+	MaxRetriesForBatch   int
```
countvonzero: rm

fetch/fetch.go
```diff
@@ -189,10 +193,12 @@ type Fetch struct {
 	onlyOnce             sync.Once
 	doneChan             chan struct{}
 	dbLock               sync.RWMutex
+	hashToPeers          HashPeersCache
+	hashToPeersM         sync.RWMutex
```
countvonzero: you don't need this mutex. lru cache is thread-safe

fetch/cache.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Get returns hash peers, it also returns a boolean to indicate whether the item
+// was found in cache.
+func (hpc HashPeersCache) Get(hash types.Hash32) (HashPeers, bool) {
+	item, found := hpc.Cache.Get(hash)
+	if !found {
+		return nil, false
+	}
+	return item.(HashPeers), true
```
countvonzero, line 76: we should record the hit/miss counts here.

fetch/cache.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// cacheStats stores hash-to-peers cache hits & misses.
+type cacheStats struct {
+	// Hits is a number of successfully found hashes
+	Hits int64 `json:"hits"`
+	// Misses is a number of not found hashes
+	Misses int64 `json:"misses"`
```
countvonzero: can you consider exporting the hit rate instead of hit/misses?
the absolute number in metrics doesn't really help. to me the more important thing is hit*100/total to show how effective this cache is.

fetch/fetch.go
```diff
@@ -511,47 +578,36 @@ func (f *Fetch) sendBatch(requests []requestMessage) {
...
-				log.String("peer", p.String()),
-				log.Int("retries", retries))
-		} else {
+		err = f.net.Request(context.TODO(), p, bytes, f.receiveResponse, errorFunc)
+		if err == nil {
 			break
 		}
+
+		retries++
+		if retries > f.cfg.MaxRetriesForPeer {
```
countvonzero, line 572: in the old code, it's called MaxRetriesForPeer, but it was really trying different peer for each retry.

i wanted to drop this param too, but i guess we can keep it in case of transient network error for the same peer.

fetch/fetch.go
```diff
@@ -650,3 +706,45 @@ func (f *Fetch) GetHash(hash types.Hash32, h Hint, validateHash bool) chan HashD
 
 	return resChan
 }
+
+func (f *Fetch) mapPeerToHash(hash types.Hash32, peer p2p.Peer) {
+	f.hashToPeersM.Lock()
+	defer f.hashToPeersM.Unlock()
+
+	peers, exists := f.hashToPeers.Get(hash)
```
countvonzero: we should hide this logic in cache. i.e. hashToPeers.Add() should check its internal cache first before adding.
and you should change the hashToPeers.Add() to take a set of peers. otherwise we will get the same hash from cache `len(peers)` times.

layerfetcher/layers.go
```diff
@@ -305,8 +305,30 @@ func (l *Logic) PollLayerContent(ctx context.Context, layerID types.LayerID) cha
...
+	}
+
+	l.fetcher.RegisterPeerHashes(peer, layerHashes)
+
+	return
+}
+
 // fetchLayerData fetches the all content referenced in layerData.
-func (l *Logic) fetchLayerData(ctx context.Context, logger log.Log, layerID types.LayerID, blocks *layerData) error {
+func (l *Logic) fetchLayerData(ctx context.Context, logger log.Log, peer p2p.Peer, layerID types.LayerID, blocks *layerData) error {
```
countvonzero: rm peer arg. unused

layerfetcher/layers.go
countvonzero: sigh.... a dependency mess.

countvonzero (commented):
can you share in the systest what the hit rate is?

evt:
> can you share in the systest what the hit rate is?

will do once find out why error is back in systests after the latest changes
Screenshot 2022-06-15 at 00 40 00

fetch/fetch.go
evt: removed

fetch/fetch.go
evt: logic hidden, cheers

layerfetcher/layers.go
evt: done

fetch/cache.go
evt: done

fetch/fetch.go
evt: done

fetch/cache.go
evt: done

fetch/cache.go
```diff
@@ -0,0 +1,86 @@
...
+	peers[peer] = struct{}{}
+	hpc.Cache.Add(hash, peers)
+}
+
+// Get returns hash peers, it also returns a boolean to indicate whether the item
+// was found in cache.
+func (hpc *HashPeersCache) Get(hash types.Hash32) (HashPeers, bool) {
+	item, found := hpc.Cache.Get(hash)
+	if !found {
+		hpc.miss()
```
countvonzero, line 63: this is trickier. since Add() now call Get, we cannot blindly inc stats here. 
i think you may need a `get()` to actually get the item from cache. `Add()` and `Get()` both call `get()`, but stats are only inc'ed in `Get()`

fetch/cache_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/fetch"
+	"github.com/spacemeshos/go-spacemesh/p2p"
+)
+
+func TestCache(t *testing.T) {
+	t.Run("1Hash3Peers", func(t *testing.T) {
```
countvonzero, line 16: can you also add a test case with multiple hashes?

fetch/fetch.go
```diff
@@ -189,10 +191,11 @@ type Fetch struct {
 	onlyOnce             sync.Once
 	doneChan             chan struct{}
 	dbLock               sync.RWMutex
+	hashToPeers          *HashPeersCache
 }
 
 // NewFetch creates a new Fetch struct.
-func NewFetch(ctx context.Context, cfg Config, h *p2p.Host, dbs map[Hint]database.Getter, logger log.Log) *Fetch {
+func NewFetch(_ context.Context, cfg Config, h *p2p.Host, dbs map[Hint]database.Getter, logger log.Log) *Fetch {
```
countvonzero: maybe just remove the context all together?

layerfetcher/layers.go
```diff
@@ -305,6 +305,28 @@ func (l *Logic) PollLayerContent(ctx context.Context, layerID types.LayerID) cha
 	return resChannel
 }
 
+// registerLayerHashes registers provided hashes with provided peer.
+func (l *Logic) registerLayerHashes(peer p2p.Peer, blocks *layerData) {
+	if blocks == nil {
```
countvonzero: rename blocks to something else. it's from very old code.

layerfetcher/layers.go
```diff
@@ -305,6 +305,28 @@ func (l *Logic) PollLayerContent(ctx context.Context, layerID types.LayerID) cha
...
+		return
+	}
+	var layerHashes []types.Hash32
+	for _, ballotID := range blocks.Ballots {
+		layerHashes = append(layerHashes, ballotID.AsHash32())
+	}
+	for _, blkID := range blocks.Blocks {
+		layerHashes = append(layerHashes, blkID.AsHash32())
+	}
```
countvonzero, line 312: i'd also add HareOutput just in case it's not included in layerData.Blocks.

countvonzero (approved):

countvonzero (commented):
please update the PR description and describe the changes

layerfetcher/layers.go
evt: done

layerfetcher/layers.go
evt: done

fetch/fetch.go
evt: thanks, done

fetch/cache_test.go
evt: done

fetch/cache.go
evt: done, cheers

countvonzero:
> please update the PR description and describe the changes

@evt in the description, please share the cache hitrate during system test. thanks

evt:
bors try

evt:
> > please update the PR description and describe the changes
> 
> @evt in the description, please share the cache hitrate during system test. thanks

```
fbe52.layerFetcher .fetch - "hitrate":"100.00"
fbe52.layerFetcher .fetch - "hitrate":"52.94"
fbe52.layerFetcher .fetch - "hitrate":"45.00"
fbe52.layerFetcher .fetch - "hitrate":"42.86"
fbe52.layerFetcher .fetch - "hitrate":"58.62"
fbe52.layerFetcher .fetch - "hitrate":"67.57"
fbe52.layerFetcher .fetch - "hitrate":"73.91"
fbe52.layerFetcher .fetch - "hitrate":"77.78"
fbe52.layerFetcher .fetch - "hitrate":"80.95"
fbe52.layerFetcher .fetch - "hitrate":"83.10"
fbe52.layerFetcher .fetch - "hitrate":"85.00"
fbe52.layerFetcher .fetch - "hitrate":"86.36"
fbe52.layerFetcher .fetch - "hitrate":"79.38"
fbe52.layerFetcher .fetch - "hitrate":"77.78"
fbe52.layerFetcher .fetch - "hitrate":"79.44"
fbe52.layerFetcher .fetch - "hitrate":"79.63"
fbe52.layerFetcher .fetch - "hitrate":"81.20"
fbe52.layerFetcher .fetch - "hitrate":"82.26"
fbe52.layerFetcher .fetch - "hitrate":"84.17"
fbe52.layerFetcher .fetch - "hitrate":"83.57"
fbe52.layerFetcher .fetch - "hitrate":"83.69"
fbe52.layerFetcher .fetch - "hitrate":"83.22"
fbe52.layerFetcher .fetch - "hitrate":"83.33"
fbe52.layerFetcher .fetch - "hitrate":"83.45"
fbe52.layerFetcher .fetch - "hitrate":"83.56"
fbe52.layerFetcher .fetch - "hitrate":"83.67"
fbe52.layerFetcher .fetch - "hitrate":"83.78"
fbe52.layerFetcher .fetch - "hitrate":"83.89"
fbe52.layerFetcher .fetch - "hitrate":"84.00"
fbe52.layerFetcher .fetch - "hitrate":"84.11"
```

countvonzero:
@evt i meant to update the PR's Changes section to describe what you did in this PR and document the hit rate in systest. that way future readers can just read the PR description w/o following the whole conversation to the end

evt:
> @evt i meant to update the PR's Changes section to describe what you did in this PR and document the hit rate in systest. that way future readers can just read the PR description w/o following the whole conversation to the end

thanks, done

evt:
bors merge

evt:
bors merge

dshulyak:
i don't get it why hitrate is not 100. can someone clarify?

evt:
> i don't get it why hitrate is not 100. can someone clarify?

I'd say this is because we don't register peer hashes everywhere we're supposed to yet. Is it correct @countvonzero ?

countvonzero:
> > i don't get it why hitrate is not 100. can someone clarify?
> 
> I'd say this is because we don't register peer hashes everywhere we're supposed to yet. Is it correct @countvonzero ?

@dshulyak 
this PR does not completely handle all data types. so far it tackles only ballots/blocks. for ATXs and TXs, those hashes will be misses. i am expecting a follow-up PR for those. @evt you can verify this easily by not querying cache for hints with atxdb and txdb.

after that, if the hitrate is still not close to 100%, then i'll try adjusting the cache size.

evt:
> > > i don't get it why hitrate is not 100. can someone clarify?
> > 
> > 
> > I'd say this is because we don't register peer hashes everywhere we're supposed to yet. Is it correct @countvonzero ?
> 
> @dshulyak this PR does not completely handle all data types. so far it tackles only ballots/blocks. for ATXs and TXs, those hashes will be misses. i am expecting a follow-up PR for those. @evt you can verify this easily by not querying cache for hints with atxdb and txdb.
> 
> after that, if the hitrate is still not close to 100%, then i'll try adjusting the cache size.

tested with cache disabled for ATXs, TXs, poets and proposals (left blocks and ballots only) - all good - hit rate is 100% then.

evt:
bors merge

Issue #3247, opened 2022-06-10
p2p: reduce network protocols names to a smaller size


dshulyak:
we have overly verbose network protocols, each of them will add at least 10bytes of unncessary traffic per message.

there is no need to make it verbose or even human readable, every protocol can be described with one or two letters and a version.

PR #3248, opened 2022-06-10
[Merged by Bors] - change protocol versions

dshulyak:
i think we should change relevant protocol versions when related types are changed.

for genesis we will probably reset them to 1, and maintain backward compatibility. but for now it is safer to change versions so that not updated clients won't cause issues for new network.

related https://github.com/spacemeshos/go-spacemesh/issues/3247

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -30,7 +30,7 @@ var (
 )
 
 // AtxProtocol is the protocol id for broadcasting atxs over gossip.
-const AtxProtocol = "AtxGossip"
+const AtxProtocol = "AtxGossip/1"
 
 const defaultPoetRetryInterval = 5 * time.Second
 
--- a/activation/poetlistener.go
+++ b/activation/poetlistener.go
@@ -12,7 +12,7 @@ import (
 )
 
 // PoetProofProtocol is the name of the PoetProof gossip protocol.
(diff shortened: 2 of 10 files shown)


evt (approved):

evt:
@dshulyak just curious, isn't it better to use semantic versioning (e.g. `x.y.z`)?

dshulyak:
> @dshulyak just curious, isn't it better to use semantic versioning (e.g. x.y.z)?

i think it is better only if changes will be handled differently. are we going handle changes in `y` differently from `x` ?

evt:
> > @dshulyak just curious, isn't it better to use semantic versioning (e.g. x.y.z)?
> 
> i think it is better only if changes will be handled differently. are we going handle changes in `y` differently from `x` ?

I had a feeling we should :) you should know better of course. Or you mean semantic versioning is used for some global version instead?

dshulyak:
> I had a feeling we should :) you should know better of course. Or you mean semantic versioning is used for some global version instead?

we should use x.y.z versioning somewhere, probably for go-spacemesh itself. but i can't imagine now how we may handle `y` or `z` version difference in the protocols

dshulyak:
bors merge

PR #3249, opened 2022-06-10
[Merged by Bors] - SQL collect metrics for sqlite db

### LINKED ISSUE #3197: sql, observability: implement prometheus collector for sqlite
see https://www.sqlite.org/dbstat.html

need to collect total size for each table and index, as well as total db size. table and index should be prometheus label

## implementation 

1. make sure that sqlite is compiled with https://www.sqlite.org/compile.html#enable_dbstat_vtab
2. select name, sum(pgsize) from dbstat;
3. name is a label
4. run collector every 5 minutes

abergasov:
## Motivation

Closes #3197

## Changes
added prometheus collector for table|index sizes

## DevOps Notes

### DIFF
--- a/sql/metrics/prometheus.go
+++ b/sql/metrics/prometheus.go
@@ -67,6 +67,7 @@ func NewDBMetricsCollector(ctx context.Context, db *sql.Database, logger log.Log
 	return collector
 }
 
+// Close closes DBMetricsCollector.
 func (d *DBMetricsCollector) Close() {
 	d.cancel()
 	if err := d.eg.Wait(); err != nil {
--- a/sql/metrics/prometheus.go
+++ b/sql/metrics/prometheus.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/metrics"
@@ -23,6 +24,8 @@ type DBMetricsCollector struct {
 	checkInterval time.Duration
 	db            *sql.Database
 	tablesList    map[string]struct{}
+	eg            errgroup.Group
+	cancel        context.CancelFunc
 
 	tableSize *prometheus.GaugeVec
 	indexSize *prometheus.GaugeVec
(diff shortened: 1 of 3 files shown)


dshulyak:
does it work with the sqlite that we are using?

dshulyak:
just to clarify, we need a way to build with stats enabled. otherwise, what the point of doing this change?

abergasov:
we can enable|disable collect statistic before build

```
// enable collect statistic
export CGO_CFLAGS := "-DSQLITE_ENABLE_DBSTAT_VTAB=1"
go build 

// disable collect statistic
export CGO_CFLAGS := "-DSQLITE_ENABLE_DBSTAT_VTAB=0"
go build 
```

dshulyak (approved):

sql/metrics/prometheus.go
```diff
@@ -0,0 +1,130 @@
...
+		return
+	}
+
+	collector.tablesList, err = collector.getListOfTables()
+	if err != nil {
+		collector.logger.With().Error("error get list of tables", log.Err(err))
+		return
+	}
+	collector.logger.With().Info("start collect stat")
+	go collector.CollectMetrics(ctx)
```
countvonzero: please try to use errgroup to manage go routines

sql/metrics/prometheus.go
```diff
@@ -0,0 +1,134 @@
...
+		return
+	}
+
+	collector.tablesList, err = collector.getListOfTables()
+	if err != nil {
+		collector.logger.With().Error("error get list of tables", log.Err(err))
+		return
+	}
+	collector.logger.With().Info("start collect stat")
+	collector.eg.Go(func() error {
```
dshulyak, line 63: i think what @countvonzero meant is that it might be useful to wait for collector to finish. otherwise there are no reasons to use errgroup in here

dshulyak (approved):

dshulyak (approved):

abergasov:
bors try

sql/metrics/prometheus.go
countvonzero: yes. we want to actively manage go routines to have a clean shutdown.

countvonzero (approved):

abergasov:
bors merge

PR #3250, opened 2022-06-10
txs: integrate account abstraction vm

### LINKED ISSUE #3254: txs: bypass conservative cache during sync
https://github.com/spacemeshos/go-spacemesh/discussions/3252

changes in https://github.com/spacemeshos/go-spacemesh/discussions/3252#discussioncomment-2933674 will be sufficient to use new types from https://github.com/spacemeshos/go-spacemesh/pull/3250

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3254

- integration with mempool

for gossiped transaction nothings changes. transaction is added to mempool only after it completely passes validation (parsing/verify methods). there is a room for an optimization - verify can be called after mempool will do its own validation, but i will leave it for another change

transactions referenced in the block are saved without validation. when the layer is applied - those transactions will be verified, and if any of them fails whole block will be terminated. invalid block needs to be pruned from the database
the rest of the changes are related to using new types and tx flow.

transactions referenced by the proposal are validated immediately. if nodes vm state lags behind it may not be able to parse and validate such transaction. in such case hare will not construct unified block and will fallback to sync. so technicall transactions referenced in proposal are handled the same way as gossiped.

- streaming tx headers

headers will be streamed after parsing them and only for the principal. previously they were streamed also for recipient.
related changes in api repo https://github.com/spacemeshos/api/pull/164

- system tests 

if expected nonce is zero account creates self-spawn transaction.


dshulyak:
bors try

genvm/sdk/options.go
```diff
@@ -0,0 +1,42 @@
+package sdk
```
countvonzero, line 1: i find the package name confusing. sdk to me is the set of libraries/tools to program against a specific platform for developers to install. here this is more like a testlib kind of package.

genvm/sdk/options.go
```diff
@@ -0,0 +1,42 @@
...
+// Defaults returns default Options.
+func Defaults() *Options {
+	return &Options{GasPrice: 1}
+}
+
+// Options to modify common transaction fields.
+type Options struct {
+	Principal *core.Address
+	GasPrice  uint64
+	Nonce     *core.Nonce
```
countvonzero: why should Nonce be passed as Options? there are no default value for a principal's nonce.
why not make Nonce an arg to Spend()?

genvm/sdk/wallet/tx.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	payload := wallet.SpawnPayload{}
+	payload.GasPrice = options.GasPrice
+	copy(payload.Arguments.PublicKey[:], pk[32:])
+	var principal core.Address
+	if options.Principal != nil {
+		principal = *options.Principal
```
countvonzero: in this case, should the library check whether `core.ComputePrincipal(wallet.TemplateAddress, &payload.Arguments)` matches option.Principal?

genvm/sdk/wallet/tx.go
```diff
@@ -0,0 +1,93 @@
...
+		_, err := field.EncodeScale(encoder)
+		if err != nil {
+			panic(err)
+		}
+	}
+	return buf.Bytes()
+}
+
+// SelfSpawn create spawn transaction.
+func SelfSpawn(pk signing.PrivateKey, opts ...sdk.Opt) []byte {
```
countvonzero, line 42: for transactions, should we just require the client to just use oasisprotocol's ed25519?

genvm/sdk/wallet/tx.go
```diff
@@ -0,0 +1,93 @@
...
+		opt(options)
+	}
+	if options.Nonce == nil {
+		panic("nonce is required for spend")
+	}
+
+	var principal types.Address
+	if options.Principal == nil {
+		spawnargs := wallet.SpawnArguments{}
+		copy(spawnargs.PublicKey[:], pk[32:])
```
countvonzero: this is confusing. reader needs to understand how keys are presented internally in signing.PrivateKey. can you  do pk.Public() or something self-explainatory?

genvm/sdk/wallet/tx.go
```diff
@@ -0,0 +1,93 @@
...
+	options := sdk.Defaults()
+	for _, opt := range opts {
+		opt(options)
+	}
+	if options.Nonce == nil {
+		panic("nonce is required for spend")
+	}
+
+	var principal types.Address
+	if options.Principal == nil {
```
countvonzero: why not always require to pass in Principal in the Spend case? in multi-sig scenario, is it safe to derive if it's not passed in as an option?

genvm/templates/wallet/handler.go
```diff
@@ -17,13 +17,7 @@ const (
 	TotalGasSpend = 100
 )
 
-const (
-	methodSpawn = 0
-	methodSpend = 1
-)
-
```
countvonzero, line 24: undo the edit

genvm/templates/wallet/handler.go
```diff
@@ -17,13 +17,7 @@ const (
...
 func init() {
-	// single key wallet reserves template at the address ending with 1.
```
countvonzero, line 26: undo the edit

genvm/vm.go
```diff
@@ -51,28 +53,83 @@ func New(db *sql.Database, opts ...Opt) *VM {
...
+	return layers.GetStateHash(v.db, lid)
+}
+
+// GetLayerApplied returns layer of the applied transaction.
+func (v *VM) GetLayerApplied(tid types.TransactionID) (types.LayerID, error) {
+	return transactions.GetAppliedLayer(v.db, tid)
+}
+
+// GetStateRoot gets the current state root hash.
+func (v *VM) GetStateRoot() (types.Hash32, error) {
```
countvonzero, line 80: i always have double with this API. why is it useful to return a state root without specifying the layer?
why don't we return both hash and layer?

systest/tests/transactions_test.go
```diff
@@ -94,7 +102,7 @@ func testTransactions(t *testing.T, tctx *testcontext.Context, cl *cluster.Clust
 		}
 	}
 
-	diff := batch * amount * int(sendFor) * cl.Accounts()
+	diff := batch * amount * int(sendFor-1) * cl.Accounts()
```
countvonzero, line 105: why the -1?

txs/conservative_state.go
```diff
@@ -146,67 +155,98 @@ func (cs *ConservativeState) RevertState(revertTo types.LayerID) (types.Hash32,
...
 	}
 
-	txs, err := cs.getTXsToApply(toApply)
+	txs, raw, err := cs.getTXsToApply(toApply)
 	if err != nil {
 		return nil, err
 	}
 
-	failedTxs, err := cs.vmState.ApplyLayer(toApply.LayerIndex, txs, toApply.Rewards)
+	skipped, err := cs.vmState.Apply(toApply.LayerIndex, raw, toApply.Rewards)
```
countvonzero, line 170: couldn't you pass []Transaction to vmState.Apply()? since a Transaction is implicitly a RawTx, is it possible for vmState.Apply() to take []*Transaction or []*RawTx. that way you can keep the return types getTXsToApply unchanged.

txs/conservative_state.go
```diff
@@ -146,67 +155,98 @@ func (cs *ConservativeState) RevertState(revertTo types.LayerID) (types.Hash32,
...
 	}
-	return failedTxs, nil
+	return skipped, nil
 }
 
-func (cs *ConservativeState) getTXsToApply(toApply *types.Block) ([]*types.Transaction, error) {
+func (cs *ConservativeState) getTXsToApply(toApply *types.Block) ([]*types.Transaction, []types.RawTx, error) {
+	// TODO this allocates an additional map for all tx ids in the block
+	// it would make much more sense to load transactions in the loop below
+	// and terminate with error if transaction is missing
```
countvonzero: i don't understand this comment. what map are you referring to?

txs/conservative_state_test.go
```diff
@@ -693,16 +699,95 @@ func TestApplyLayer_VMError(t *testing.T) {
 	}
 }
 
+func TestConsistentHandling(t *testing.T) {
+	// there are two different workflows for transactions
+	// 1. receive gossiped transaction and verify it immediately
+	// 2. receive synced transaction and delay verification
```
countvonzero, line 705: i feel we should differentiate synced transaction while gossiping and synced transaction while syncing.
the former case we should still be able to go through the gossip codepath and verify immediately.

countvonzero (approved):

txs/conservative_state_test.go
dshulyak: it is hard for me to tell if thats going to work
the problem is that it is possible to safely verify transaction from layer N, only when layer N-1 was applied

countvonzero: i think this will happen naturally when conservative cache is only working after the node is synced.

> the problem is that it is possible to safely verify transaction from layer N, only when layer N-1 was applied

is that true for mempool transactions too? i'd assume that verifying transactions are independent of layer. 
whatever transactions that failed to verify will stay in db (not conservative cache), and when a layer is applied, the conservative cache will query db again for all feasible transactions again.

countvonzero: > i think this will happen naturally when conservative cache is only working after the node is synced.

and i meant to work on that next.

dshulyak: > is that true for mempool transactions too? i'd assume that verifying transactions are independent of layer.

verification will always fail if account wasn't spawned

> whatever transactions that failed to verify will stay in db (not conservative cache),

so are you going to save all transactions, even those that have failed during gossip? i doubt that it will work

> and i meant to work on that next.

i am pretty sure that you will have to always save transactions that are referenced somewhere without verifying them if layers are applied in the parallel

genvm/sdk/options.go
dshulyak: sdk for working with spacemesh transactions. it can be used in smrepl as well, there is nothing testing specific in this module

genvm/sdk/wallet/tx.go
dshulyak: i don't see why, it can be signed using any ed25519 library. in this case i think it is actually signed by using oasis lib

genvm/sdk/wallet/tx.go
dshulyak: multisig will have its own module. the whole point of this module is to avoid passing million of parameters as it results in unintelligible code

genvm/vm.go
dshulyak: just copy pasted it from the old module. is it even used anywhere?

systest/tests/transactions_test.go
dshulyak: in the first layer accounts are spawned

txs/conservative_state.go
dshulyak: vm doesn't need anything beyond raw tx, by providing txs with headers someone can decide to reuse headers in the vm

txs/conservative_state.go
dshulyak: i will remove. i don't recall, but there were unnecessary creation of the maps and arrays somewhere

genvm/sdk/wallet/tx.go
dshulyak: since i called it SelfSpawn i can remove `principal` option completely

genvm/vm.go
countvonzero: it's only used in the api GlobalStateService.GlobalStateHash, but then it queries LatestLayerInState(). i think it just make more sense to return Layer in this case. in any case, not related to your PR. i'll change it later.

txs/conservative_state.go
countvonzero: i was thinking doing
`skipped, err := cs.vmState.Apply(toApply.LayerIndex, txs, toApply.Rewards)`
(assuming txs is []Transaction)
and keep vmState.Apply() to take []RawTx. 

does that work with golang 1.8, since a Transaction is a RawTx also?

dshulyak: it definitely won't work with slice. not sure if it will work with a single object

dshulyak:
bors merge

lrettig:
@dshulyak were TODO issues opened for these? Could you please link them here for completeness? Thanks!

> there is a room for an optimization - verify can be called after mempool will do its own validation, but i will leave it for another change (TODO issue for that)

> invalid block needs to be pruned from the database (TODO open an issue for that).

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
@countvonzero please take a look at changes in fetcher and txs/handler . now transactions from proposals are going through the same code path in the mempool as gossiped transactions.

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
i think it doesn't work because this pr number was merged by bors already

PR #3258, opened 2022-06-14
SQLite Concurrent transactions

### LINKED ISSUE #3195: sql: consider to enable begin concurrent transactions
[BEGIN CONCURRENT](https://www.sqlite.org/cgi/src/doc/begin-concurrent/doc/begin_concurrent.md) enables optimistic concurrency, so that there can be more than 1 writer at any time. COMMIT are still serialized, and on conflict transaction has to be retried.

As of now this enhancement is not merged into trunk and maintained separately in its own [branch](https://www.sqlite.org/src/timeline?r=begin-concurrent). If we want to use we will have to compile with sqlite from that branch. Originally, when sqlite was introduced, i wanted to enable this extension, but didn't put effort. Not that this branch is used in https://bedrockdb.com/ and generally considered production ready.

Enabling this extension will allow us to write db changes spread across several modules (mesh, txs, vm) in a single transaction, without worrying that it will lock database for too long. In such case transaction can be passed over the module boundaries.

abergasov:
## Motivation
Closes #3195

## Changes
create fork of https://github.com/crawshaw/sqlite with `C` files with `BEGIN TRANSACTION` implementation and set as dependency for project - https://github.com/spacemeshos/sqlite

## Test Plan
unit tests

## DevOps Notes


dshulyak:
i think we also need a method that makes it easy to use and hides retries. i was thinking something like that:

```
type WithRetryBusy func(n int) func(opts *TxOptions) {
   return ...
}

func (db *Database) WithTxConcurrent(func(db.Executor) error, ...TxOpt) error

func (db *Database) WithTx(func(db.Executor) error, ...TxOpt) error  
```

sql/database.go
```diff
@@ -142,6 +142,16 @@ func (db *Database) Tx(ctx context.Context) (*Tx, error) {
 	return tx, tx.begin()
 }
 
+// TxConcurrent starts a concurrent transaction. Locking of db is deferred until commit will be called.
+func (db *Database) TxConcurrent(ctx context.Context) (*Tx, error) {
```
dshulyak: please add a link to the sqlite documentation about `begin concurrent`

sql/database.go
```diff
@@ -204,46 +188,22 @@ func exec(conn *sqlite.Conn, query string, encoder Encoder, decoder Decoder) (in
...
-	err       error
-}
-
-func (tx *Tx) begin() error {
-	stmt := tx.conn.Prep("BEGIN;")
-	_, err := stmt.Step()
-	if err != nil {
-		return fmt.Errorf("begin: %w", err)
+// GetSQLiteError returns sqlite error for the given error.
+func (db *Database) GetSQLiteError(err error) *sqlite.Error {
```
dshulyak, line 189: is it not possible to use errors.As? why?

sql/transactions.go
```diff
@@ -0,0 +1,177 @@
...
+
+// WithRetryBusy number of retry db transactions.
+func WithRetryBusy(n int) TxRetryOpt {
+	return func(t *txRetryer) {
+		t.retryNum = n
+	}
+}
+
+// WithRetryTimeout timeout for retrying db transactions.
+func WithRetryTimeout(milliseconds int) TxRetryOpt {
```
dshulyak: i think it is always better to use time.Duration for anything related to duration in golang

sql/database.go
abergasov: sqllite not use go's original error. They have their own struct, which just implement `error` interface. When they call `Error() string` - it include also issue with sql, which can be different any time.

so need deep unwrap to get this object and get code.

abergasov:
```
testErr := errors.Wrap(errors.Wrap(sqlite.Error{
  Code:  1,
  Query: "123",
  Msg:   "234",
}, "wrapped sqlite error"), "wrapped error")
resErr := sqlite.Error{Code: 2, Query: "1234", Msg: "234"}
r := errors.As(testErr, &resErr) // true
```
this code will return true even if sqlCode is different

dshulyak:
lets put it on hold. i tend to think that it is premature to merge it without deciding where and why concurrent txs will be used

dshulyak:
lets reconsider in the future

PR #3260, opened 2022-06-15
[Merged by Bors] - systest: merge systest module into root module

dshulyak:
enables reusing code from go-spacemesh in systest without additional complexity dependencies  https://github.com/spacemeshos/go-spacemesh/pull/3250. 

the size of the go-spacemesh binary remains the same.


evt (approved):

countvonzero (approved):
as i read the code change, there is no need to update systest's README. correct?

dshulyak:
> as i read the code change, there is no need to update systest's README. correct?

right, it surprisingly worked without changing makefile commands

dshulyak:
bors merge

Issue #2085, opened 2020-07-28
Not all status updates are being consumed by API streams


lrettig:
They're probably being created faster than the service can consume and stream them. We may need to increase the buffer size to accommodate this:

```
> grpcurl -plaintext -d '' localhost:9092 spacemesh.v1.NodeService/StatusStream
> go-spacemesh --grpc node
...
2020-07-28T11:51:05.529-0400    INFO    00000.defaultLogger     GRPC NodeService.StatusStream
2020-07-28T11:51:11.033-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:21.028-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:31.028-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:41.030-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:51.029-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.429-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.438-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:53.438-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.438-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:53.438-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
2020-07-28T11:51:53.438-0400    INFO    00000.defaultLogger     reported status update
2020-07-28T11:51:53.438-0400    INFO    00000.defaultLogger     not reporting status update as no one is listening
[...]

PR #3263, opened 2022-06-17
[Merged by Bors] - enable transactions results

### LINKED ISSUE #3267: vm: transaction results (while we don't have receipts)
After transaction was applied to the vm and consumed (not skipped!) it should have a transaction result that includes minimal information on the status of the transaction.

Below is the API object. It should be possible to query historically applied transactions and subscribe to live transactions that are applied to the state.

```
message TransactionResult {
    enum Status {
        SUCCESS = 0;
        FAILURE = 1;
    }

    Transaction tx = 1;
    Status status = 2;
    string message = 3;
    uint64 gas = 4;
    uint64 fee = 5;
    bytes block = 6;
    uint32 layer = 7;
    repeated bytes addresses = 8;
}
```

dshulyak:
closes https://github.com/spacemeshos/go-spacemesh/issues/3267

- transactions result

created for transactions that were applied to the state, either successfully or simply consumed. stores information about consumed gas, fee (multiple by gas price in the header), tx status with error string, updated addresses. maybe will be added for invalid transactions in the future https://github.com/spacemeshos/go-spacemesh/issues/3272

when saved into the database indexed by updated addresses, so that both principal and other modified addresses can fine this transaction efficiently.

all changes in the txs module are because vm returns this type on apply.

- transactions results stream

new endpoint for transactions service. can request a stream of historical and live transactions results. filterable by: tx id, address, start/end layer. in order to watch new data - watch field must be set to true. single stream will provide results without duplicates.

also adjusted systest for transactions to use new streaming endpoint.

api change: https://github.com/spacemeshos/api/pull/165

- events: buffered subscription type

allows to subscribe for events of a particular type with a buffer. on buffer overflow subscription is closed and subscriber must terminate with error. this was existing behavior for api streams, i just refactored it a bit.

- database: immediate tx

sqlite backend terminates transaction with sqlite_busy without waiting if transaction is initialized as reader and then upgrades to writes later. i added a helper to initialize tx with `begin immediate` that will initialize transaction in a writer mode immediately.


api/config/genesis.go
```diff
@@ -11,6 +11,18 @@ type GenesisConfig struct {
 	Accounts map[string]uint64 `mapstructure:"accounts"`
 }
 
+// ToAccounts creates list of types.Account instance from config.
+func (g *GenesisConfig) ToAccounts() []types.Account {
+	var rst []types.Account
```
evt:
```suggestion
	rst := make([]types.Account, 0, len(g.Accounts))
```
or it's not that big? :)

dshulyak: it is probably not that big. but in general i don't think that it is important, i tend to ignore preallocations when the code has no chance to have an impact on performance (e.g tests or node initialization)

dshulyak:
bors try

dshulyak:
bors try

api/grpcserver/transaction_service.go
```diff
@@ -3,23 +3,29 @@ package grpcserver
...
 	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/sql"
+	"github.com/spacemeshos/go-spacemesh/sql/transactions"
 	"github.com/spacemeshos/go-spacemesh/txs"
 )
 
 // TransactionService exposes transaction data, and a submit tx endpoint.
 type TransactionService struct {
+	db        *sql.Database
```
countvonzero, line 28: yay.

sql/transactions/transactions.go
```diff
@@ -438,3 +430,42 @@ func queryPending(db sql.Executor, query string, encoder func(*sql.Statement), e
...
+		if _, err := db.Exec(`insert into transactions_results_addresses 
+		(address, tid) values (?1, ?2);`,
+			func(stmt *sql.Statement) {
+				stmt.BindBytes(1, rst.Addresses[i][:])
+				stmt.BindBytes(2, id[:])
+			}, nil); err != nil {
+			return fmt.Errorf("add address %s to %s: %w",
+				rst.Addresses[i].String(), id[:], err)
+		}
+	}
```
countvonzero, line 469: i had assume our paradigm is to make the caller responsible for whether db_API_1 and db_API_2 should be executed in the same db transaction. in this paradigm, each API is atomic. i.e. we will have the execution against `transactions` and `transactions_result_addresses` in separate db APIs. otherwise it's prone to misuse.

i'll prefer breaking this into two APIs.

api/grpcserver/transaction_service.go
```diff
@@ -272,3 +280,135 @@ func (s TransactionService) TransactionsStateStream(in *pb.TransactionsStateStre
...
+		}
+		ierr = stream.Send(castResult(rst))
+		return ierr == nil
+	})
+	if err == nil {
+		err = ierr
+	}
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			return nil
```
countvonzero, line 337: should check if sub == nil before returning?

api/grpcserver/transaction_service.go
```diff
@@ -272,3 +280,135 @@ func (s TransactionService) TransactionsStateStream(in *pb.TransactionsStateStre
...
+		lid := types.NewLayerID(in.Start)
+		filter.Start = &lid
+	}
+	if in.End > 0 {
+		if in.Watch {
+			return status.Error(codes.InvalidArgument, "watch stream should have an empty End argument")
+		}
+		lid := types.NewLayerID(in.End)
+		filter.End = &lid
+	}
```
countvonzero, line 311: what happens if none of the criteria is set?

events/subscription.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// BufferedSubscription is meant to be used by API subscribers.
+//
+// Majority of the events are published on the critical consensus path, and they
+// can't block consensus progress if consumer is slow.
+// To account for slow consumer each subscription maintains an internal buffer
+// with specific events, if this buffer overflows consumer should be dropped
+// and stream restarted.
+type BufferedSubscription[T any] struct {
```
countvonzero, line 62: yay

sql/database.go
```diff
@@ -136,7 +136,19 @@ func (db *Database) Tx(ctx context.Context) (*Tx, error) {
 		return nil, ErrNoConnection
 	}
 	tx := &Tx{db: db, conn: conn}
-	return tx, tx.begin()
+	return tx, tx.deferred()
+}
+
+// TxImmediate creates immediate transaction.
+//
+// IMMEDIATE cause the database connection to start a new write immediately, without waiting // for a write statement. The BEGIN IMMEDIATE might fail with SQLITE_BUSY if another write //// transaction is already active on another database connection.
```
countvonzero: fix this comment.

do you mean BEGIN might fail with SQLITE_BUSY?

txs/conservative_state.go
```diff
@@ -178,35 +181,23 @@ func (cs *ConservativeState) ApplyLayer(toApply *types.Block) ([]types.Transacti
...
 		return nil, errs[0]
 	}
 	return skipped, nil
 }
 
-func (cs *ConservativeState) getTXsToApply(toApply *types.Block) ([]*types.Transaction, []types.RawTx, error) {
+func (cs *ConservativeState) getTXsToApply(toApply *types.Block) ([]types.RawTx, error) {
+	// TODO this allocates an additional map for all tx ids in the block
+	// it would make much more sense to load transactions in the loop below
+	// and terminate with error if transaction is missing
```
countvonzero: i think this is from a previously removed comment.

countvonzero (approved):

sql/transactions/transactions.go
dshulyak: i didn't had such assumptions. it is just the first case when i needed two distinct actions to be a part of a single logical action

this indeed makes it prone to misuse. but still adding two different functions is not a good option

i actually thought about changing signature to accept *sql.Tx as a first argument

api/grpcserver/transaction_service.go
dshulyak: why? subscription will close in defer if it was initialized
i wanted to avoid printing error in the grpc logs if client went away before consuming whole result, need to doublecheck that it works the way i expect though.

api/grpcserver/transaction_service.go
dshulyak: it constructs a query without `where` statement

sql/database.go
dshulyak: noticed this as well

yes, so if there is another write in progress database will wait sufficiently long before terminating tx with SQLITE_BUSY. but when it is initialized tx with read statement, and then upgrades to write - it doesn't wait and terminates immediately

api/grpcserver/transaction_service.go
countvonzero: i thought if there was no historical data but there is a subscription for future data the subscription will be closed prematurely.

sql/transactions/transactions.go
countvonzero: i think taking *sqlTx is a good signal. maybe just for function that needs to make multiple stmts.

events/subscription.go
lrettig: @dshulyak it looks like this addresses #2085, could you confirm?

dshulyak: we were already using buffer for internal events. i just refactored it

grpc stream has an internal buffer - 64kb. if consumer is alive it is really hard to overflow it with small events. 
i don't know what is the problem in that issue. if you think that it was due to the lack of buffer - then yes we have additional buffering now

lrettig: is it still the case that events are created faster than a downstream client can sync them via the stream? (esp. during initial sync) - that's what that issue is about.

dshulyak: interesting, i didn't pay attention to the amount of events during sync. i would expect that current buffering should be enough if consumer is not slacking. but it needs to be verified, since i am not really sure at what rate we are emitting events during sync

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

events/subscription.go
lrettig: okay, let's follow up in #2085 (and keep that issue open for now)

PR #3295, opened 2022-06-24
[Merged by Bors] - txs: integrate account abstraction vm

### LINKED ISSUE #3254: txs: bypass conservative cache during sync

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3254

- integration with mempool

for gossiped transaction nothings changes. transaction is added to mempool only after it completely passes validation (parsing/verify methods). there is a room for an optimization - verify can be called after mempool will do its own validation, but i will leave it for another change

transactions referenced in the block are saved without validation. when the layer is applied - those transactions will be verified, and if any of them fails whole block will be terminated. invalid block needs to be pruned from the database
the rest of the changes are related to using new types and tx flow.

transactions referenced by the proposal are validated immediately. if nodes vm state lags behind it may not be able to parse and validate such transaction. in such case hare will not construct unified block and will fallback to sync. so technicall transactions referenced in proposal are handled the same way as gossiped.

- streaming tx headers

headers will be streamed after parsing them and only for the principal. previously they were streamed also for recipient.
related changes in api repo https://github.com/spacemeshos/api/pull/164

- system tests

if expected nonce is zero account creates self-spawn transaction.


dshulyak:
moved from https://github.com/spacemeshos/go-spacemesh/pull/3250

abergasov (approved):

dshulyak:
bors merge

evt (approved):

PR #3296, opened 2022-06-24
3222 estimate user reward for layer

### LINKED ISSUE #3222: Tweak SmesherService API
## Rationale
We need to implement some APIs to make it possible to show all interesting information to our smeshers.
Check out the issue: https://github.com/spacemeshos/smapp/issues/912

## Required API endpoints

**SmesherService.EstimatedRewards**
- Exists in API: https://github.com/spacemeshos/api/blob/master/proto/spacemesh/v1/smesher.proto#L39
- But not implemented in go-spacemesh: https://github.com/spacemeshos/go-spacemesh/blob/ffc0c927d0d3e96dce55597cba2da8a54bcfc4e6/api/grpcserver/smesher_service.go#L154
- In addition to it, probably we'd like to have time estimation (layer number), and also query estimation of rewards for the ongoing epoch. Or probably this endpoint should accept a layer/epoch number as an argument and return the estimation. We need to discuss it.

**SmesherService.SmeshingStatus**
I think we also need a stream or only stream that sends the latest state to new subscribers before sending the next updates.

- It does not exist. Most likely we need to have a discussion about it.

- This is some kind of extended version of `IsSmeshing`, it tells API consumers what really happens under the hood:
   - Not smeshing
   - Creating PoS data
   - Waiting for PoET
   - GeneratingPoST
   - Eligible for rewards
   - Publishing ballot
   - doing something else
   - Not eligible: error

- We should pay attention, that some of these statuses make sense for the upcoming epoch.
   E.G. we're waiting for PoET for the next layer, but we're already eligible in this epoch and even publishing ballots from time to time.
   So probably it should return statuses for current and next epochs separately.

---

Please, share your ideas and thoughts, maybe you know how to solve the problem better and simpler.
cc: @noam, @moshababo, @countvonzero, @dshulyak, @oriya, @avive

abergasov:
## Motivation

Closes #3222

## Changes
implement endpoint, which try estimate reward for layer X.

for past layers - simply look at db.
for current or future layers - calc it from weight and activations

pr for api modify https://github.com/spacemeshos/api/pull/167

## TODO

## DevOps Notes


sql/rewards/rewards.go
```diff
@@ -54,3 +54,22 @@ func List(db sql.Executor, coinbase types.Address) (rst []*types.Reward, err err
 		})
 	return
 }
+
+// RewardPerLayer returns the reward per layer from `rewards` table.
+func RewardPerLayer(db sql.Executor, layerID uint32) (rwd *types.Reward, err error) {
```
dshulyak: please don't use uint32 for layer id in the code, the type should be consistent with the rest of the codebase

dshulyak: but in general i don't get the idea. there are on average 50 rewards per layer for different coinbases. all of them may have different  relative weight. which one do you want to get?

activation/activation.go
```diff
@@ -433,6 +442,31 @@ func (b *Builder) SetMinGas(value uint64) {
...
+		return 0, errors.Wrap(err, "failed to get epoch weight")
+	}
+	if len(activeSet) == 0 {
+		return 0, errors.New("empty active set for epoch")
+	}
+	expNumSlots, err := proposals.GetNumEligibleSlots(b.pendingATX.GetWeight(), totalWeight, uint32(b.layerAvgSize), b.layersPerEpoch)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to estimate number of eligible slots")
+	}
+	return layerReward / uint64(expNumSlots), nil
```
dshulyak, line 467: this doesn't make sense

layer reward is divided between builders of the block, based on their relative weight

abergasov: `layerReward` is just basic reward, which can be get for layer. Now it hardcoded value, but later it will be replaces with inflation formula

so we need manually divided to all block builders

dshulyak: exp num slots is not a number of block builders (proposers) in a single layer. expected num slots is the number of proposals a particular smesher is eligible for in a whole epoch

sql/rewards/rewards.go
abergasov: > there are on average 50 rewards per layer for different coinbases. all of them may have different weight behind them. which one do you want to get?

but node will get reward only for block, which will be constructed after `Hair`. It can be several blocks in hair output, but is should be extraordinary situations, and our estimation will not work in that cases

dshulyak: its hare, not hair

it always outputs a single block, but i don't understand why is it even important

i don't know how to explain it better, but this function will return a first reward in the requested layer.. it may not even be related to your smesher id, and will be based on the different relative weight. thats why it doesn't make any sense to me

 moreover i don't understand where idea to get an estimated reward for a previous layer came from. api can provide already applied rewards for applied layers. i think this need to be removed from the change completely

dshulyak: based on your messages i thought that you are going to provide an estimate for this layer based on the previous layers results, but this is not what is going on in this change

abergasov: yep, mistake in hare)

> it always outputs a single block, but i don't understand why is it even important

i thought, that node receive reward only after Hare create block. so in that case not important, how many block can be before

> api can provide already applied rewards for applied layers

yep, we have `AccountDataQuery` endpoint, but you should specify coinbaseAddress for load reward per layer, which can be problem if coinbaseAddress was changed

> provide an estimate for this layer based on the previous layers results

only if it make prediction more accurate. Idea was just show approximately, how much reward user will get in current|next layer. If for some reasons need be calculated reward for previous layers - that's why `layerID` was added in request

activation/activation.go
abergasov: got it, because `totalWeight` now is calculation of weight for all epoch

if `totalWeight` somehow way will be result of weight of layer, it should be set number of block builders to `expNumSlots`, right?

PR #3299, opened 2022-06-26
[Merged by Bors] - Replace log.Inline with entity IDs to reduce log size

evt:


countvonzero:
i'd prefer try downgrading some info/warning logs first, as missing info logs impacts debugging ability in devnet.

do you have data regarding the distribution of info logs? i am looking for something like "X% come from module A and Y% from module B".

evt:
@countvonzero here is INFO log line count distribution by module (600+ Mb):
```
proposalListener: 660587 (45.90%)
SVM: 454086 (31.55%)
mesh: 64318 (4.47%)
trtl: 58913 (4.09%)
sync: 48928 (3.40%)
atxDb: 35184 (2.44%)
....
```
As you can see, `proposalListener` and `SVM` make most of it. And here is a distribution of message count for those 2 modules:
- proposalListener:
```
[proposalListener] new ballot: 220149 (33.33%)
[proposalListener] ballot eligibility verified: 220148 (33.33%)
[proposalListener] processing ballot: 220020 (33.31%)
[proposalListener] processing proposal: 135 (0.02%)
...
```

- SVM:
```
[SVM] update account: 220096 (48.47%)
[SVM] update rewards: 220090 (48.47%)
[SVM] apply layer to vm: 6947 (1.53%)
...
```

please let me know if anything of above can be downgraded.

countvonzero:
> here is INFO log line count distribution by module (600+ Mb):

is line count proportional to data size? i know some line log heavier than others. like when we receive a proposal/ballot, we call log.inline() on the whole object. that log line should have more data than other line.

> please let me know if anything of above can be downgraded.

i cannot tell you which module can be downgraded. the point is that a sweeping measure at log level/module granularity does not work here. i'll recommend looking at the top offender modules/log lines here and see if it make sense to downgrade them.

countvonzero:
some ideas:
- the log.inline() on atx/ballot/proposal/tx can probably be downgraded. in these logs, logging a simple ID is probably sufficient.
- the logged IDs (atx/proposal/ballot/block/tx) can be a short string.
- in each gossip workflow, we'll have multiple layers of logging saying the same thing. take atx for example, we have
  - `got new atx 5b2b9df516`
  - `validating nipost`
  - `processing atx`
  - `atx is valid`
  - `finished storing atx in epoch`
  - `stored and propagated new syntactically valid atx`

  this is overly verbose for one single object. similar for other objects (propsal/ballot/block/tx)

- most of the SVM log can be downgraded (per account/reward update)
- hare has tons of WARN/INFO log even before the node is synced (hare shouldn't be participating before the node is synced)

these are the low hanging fruits i can think of. cc @dshulyak

evt:
@countvonzero good point re: data size, cheers. Top modules by log size (in bytes) looks as follows:
```
proposalListener: 375586070 (62.63%)
SVM: 143314358 (23.90%)
atxDb: 19471943 (3.25%)
trtl: 18539898 (3.09%)
mesh: 15764322 (2.63%)
sync: 13190092 (2.20%)
...
```
and then top messages (by size in bytes) for top 2 modules (`proposalListener` and `SVM`):
```
[proposalListener] new ballot: 227067124 (60.46%)
[proposalListener] ballot eligibility verified: 80785116 (21.51%)
[proposalListener] processing ballot: 67544712 (17.98%)
[proposalListener] new proposal: 154288 (0.04%)
...
[SVM] update rewards: 72265308 (50.42%)
[SVM] update account: 67506049 (47.10%)
[SVM] rewards info for layer: 1819234 (1.27%)
...
```

pigmej:
@countvonzero 

> hare has tons of WARN/INFO log even before the node is synced (hare shouldn't be participating before the node is synced)

Did the node try to execute Hare or only the logging part somehow got executed?

countvonzero:
> @countvonzero
> 
> > hare has tons of WARN/INFO log even before the node is synced (hare shouldn't be participating before the node is synced)
> 
> Did the node try to execute Hare or only the logging part somehow got executed?

the consensus process doesn't not start until the node is synced. so no, hare is not executed. 
but it appears that hare is listening to hare message all the time, producing a lot of spam because it cannot verify the hare messages in the network.

pigmej:
Thx.

countvonzero:
> the consensus process doesn't not start until the node is synced. so no, hare is not executed. but it appears that hare is listening to hare message all the time, producing a lot of spam because it cannot verify the hare messages in the network.

maybe not all the time, but definitely minutes before the node is synced. 

we may want to start exploring the sampling feature of zapcore for some of the modules so that only every Mth log is preserved.

dshulyak:
i thought about it changing it to warn only for smapp users

> svm logs

they were downgraded in the new vm version

blocks/handler.go
```diff
@@ -71,7 +71,8 @@ func (h *Handler) HandleBlockData(ctx context.Context, data []byte) error {
 		logger.Info("known block")
 		return nil
 	}
-	logger.With().Info("new block", log.Inline(&b))
+	logger.With().Info("new block ID", log.String("block_id", b.ID().AsHash32().ShortString()))
```
countvonzero: please keep the log as "new block/ballot/proposal/". the goal is logging an mesh event, not describing what we are logging.

countvonzero: no need to do log.String on the IDs. just 

`logger.With().Info("new block", b.ID())`. all types implement `Field()` for logging

countvonzero (commented):
- please update the PR description
- based on systest, can you describe how much log size is decreased by this PR?

dshulyak:
i want to point out that we can't just drop log.Inline. this information is absolutely required to understand what is going on with the protocol. one option is to drop it for info level, but keep it for debug. 

we can enable debug level for each module separately, but in order to do that we need to make debug actually readable in that module, right now debug level mostly just spits overwhelming amount of info.

the bottomline is that it won't be possible to merge this change if it removes log.Inline without adjusting logging more systematically

countvonzero:
yep. sorry. it was my idea to drop the inline all together. sigh... i simply dont feel like i have the capacity to worry about this now....
@dshulyak

dshulyak:
@evt lets restore inline for ballot and merge this change

config/logging.go
```diff
@@ -6,6 +6,7 @@ import "go.uber.org/zap/zapcore"
 type LogEncoder = string
 
 const (
+	// Info level causes too much of logs produced (400+ Mb for ~2 hours).
```
dshulyak: is it helpful to keep this log?

sql/layers/layers_test.go
```diff
@@ -147,7 +147,7 @@ func TestSetHashes(t *testing.T) {
 			got, err := GetHash(db, types.NewLayerID(lid))
 			require.NoError(t, err)
 			require.Equal(t, types.EmptyLayerHash, got)
-			got, err = GetAggregatedHash(db, types.NewLayerID(lid))
+			got, _ = GetAggregatedHash(db, types.NewLayerID(lid))
 			require.NoError(t, nil)
```
dshulyak: nice test

dshulyak (approved):

sql/layers/layers_test.go
evt: :)) fixed, cheers

config/logging.go
evt: nope, removed, cheers

evt:
bors merge

PR #3301, opened 2022-06-27
[Merged by Bors] - crash: fatal error: concurrent map iteration and map write

### LINKED ISSUE #3300: crash: fatal error: concurrent map iteration and map write
## Description

https://github.com/spacemeshos/go-spacemesh/runs/7075903106?check_suite_focus=true
```
2022-06-27 09:13:17 | /go/src/github.com/spacemeshos/go-spacemesh/fetch/fetch.go:243 +0x3aa
-- | --
  |   | 2022-06-27 09:13:17 | created by github.com/spacemeshos/go-spacemesh/fetch.(*Fetch).handleNewRequest
  |   | 2022-06-27 09:13:17 | /usr/local/go/src/runtime/asm_amd64.s:1571 +0x1 fp=0xc0010adfe8 sp=0xc0010adfe0 pc=0x474f41
  |   | 2022-06-27 09:13:17 | runtime.goexit()
  |   | 2022-06-27 09:13:17 | /go/src/github.com/spacemeshos/go-spacemesh/fetch/fetch.go:243 +0x26 fp=0xc0010adfe0 sp=0xc0010adfc8 pc=0xd9de46
  |   | 2022-06-27 09:13:17 | github.com/spacemeshos/go-spacemesh/fetch.(*Fetch).handleNewRequest.func1()
  |   | 2022-06-27 09:13:17 | /go/src/github.com/spacemeshos/go-spacemesh/fetch/fetch.go:430 +0x516 fp=0xc0010adfc8 sp=0xc0010adda0 pc=0xda0896
  |   | 2022-06-27 09:13:17 | github.com/spacemeshos/go-spacemesh/fetch.(*Fetch).requestHashBatchFromPeers(0xc0003ff800)
  |   | 2022-06-27 09:13:17 | /go/src/github.com/spacemeshos/go-spacemesh/fetch/fetch.go:441 +0x7f fp=0xc0010adda0 sp=0xc0010adcd0 pc=0xda095f
  |   | 2022-06-27 09:13:17 | github.com/spacemeshos/go-spacemesh/fetch.(*Fetch).send(0xc0003ff890?, {0xc0010ec000, 0x16, 0x20})
  |   | 2022-06-27 09:13:17 | /go/src/github.com/spacemeshos/go-spacemesh/fetch/fetch.go:459 +0x812 fp=0xc0010adcd0 sp=0xc0010ad8e0 pc=0xda12b2
  |   | 2022-06-27 09:13:17 | github.com/spacemeshos/go-spacemesh/fetch.(*Fetch).organizeRequests(0xc0003ff800, {0xc0010ec000, 0x16, 0x7fd5a6d0c090?})
  |   | 2022-06-27 09:13:17 | /go/src/github.com/spacemeshos/go-spacemesh/fetch/cache.go:28
  |   | 2022-06-27 09:13:17 | github.com/spacemeshos/go-spacemesh/fetch.HashPeers.ToList(...)
  |   | 2022-06-27 09:13:17 | /usr/local/go/src/runtime/map.go:871 +0x4eb fp=0xc0010ad8e0 sp=0xc0010ad870 pc=0x41992b
  |   | 2022-06-27 09:13:17 | runtime.mapiternext(0xc0010adb10?)
  |   | 2022-06-27 09:13:17 | /usr/local/go/src/runtime/panic.go:992 +0x71 fp=0xc0010ad870 sp=0xc0010ad840 pc=0x441f71
  |   | 2022-06-27 09:13:17 | runtime.throw({0x127f3f9?, 0x0?})
  |   | 2022-06-27 09:13:17 | goroutine 8517 [running]:
  |   | 2022-06-27 09:13:17 |  
  |   | 2022-06-27 09:13:17 | fatal error: concurrent map iteration and map write
```

evt:
Closes #3300

## DevOps Notes


fetch/cache.go
```diff
@@ -42,17 +51,15 @@ func NewHashPeersCache(size int) *HashPeersCache {
 
 // Add adds peer to a hash.
 func (hpc *HashPeersCache) Add(hash types.Hash32, peer p2p.Peer) {
-	hpc.mu.Lock()
-	defer hpc.mu.Unlock()
-
 	peers, exists := hpc.get(hash)
+	hpc.mu.Lock()
```
countvonzero: this is a race condition. and the code is getting complicated for simply getting a list of peers. you now have 3 levels of get (GetList, Get, get)

how about just change Get() to return a list instead of a map? both callers of Get() can work with list.

fetch/cache.go
```diff
@@ -15,20 +15,29 @@ import (
...
 
 // HashPeers holds registered peers for a hash.
 type HashPeers map[p2p.Peer]struct{}
 
-// ToList converts hash peers map to a list.
-func (hp HashPeers) ToList() []p2p.Peer {
-	result := make([]p2p.Peer, 0, len(hp))
-	for k := range hp {
+// GetList returns hash peers as a list.
+func (hpc *HashPeersCache) GetList(hash types.Hash32) ([]p2p.Peer, bool) {
```
dshulyak: is it possible to implement required api without copying? it is really bad pattern to copy in get, especially to copy a collection. better to use it only as a stop-gap, and definitely doesn't look good in newly written code

for example iterating over map is randomized in golang, i did some experiments and unfortunately distribution is not uniform (1st element in the map will be picked more frequently than other)

maybe it is worth doing smth like this:

```
GetRandom(seed int) {
    peers := ...
    n := seed % len(peers)
    for addr := range peers {
         i++
         if i == n {return addr}
    }
}
```

dshulyak: another option is to store collection in the ordered map

fetch/fetch.go
```diff
@@ -452,11 +452,10 @@ func (f *Fetch) organizeRequests(requests []requestMessage) map[p2p.Peer][][]req
 	peer2requests := make(map[p2p.Peer][]requestMessage)
 
 	for _, req := range requests {
-		hashPeersMap, exists := f.hashToPeers.Get(req.Hash)
+		hashPeers, exists := f.hashToPeers.GetList(req.Hash)
```
countvonzero: btw, AddPeersFromHash() suffer the same problem, concurrent map iteration and map write.

fetch/cache.go
evt: I think I’m missing something 🤔  `GetRandom` would need to convert peer map to a list too and then find a random peer there, right? What is a profit then? also I don’t quite understand how ordered map would help to find a random element?

evt: As discussed on a call, ordered map is needed if we want `GetRandom` to return the same peer for the same feed which we probably don't need for now.

fetch/cache.go
evt: replaced GetList + GetRandom + fixed race condition

fetch/cache.go
evt: added GetRandom, thank you.

fetch/fetch.go
evt: made mutex public + now using it in `AddPeersFromHash`

fetch/cache.go
```diff
@@ -67,13 +68,21 @@ func (hpc *HashPeersCache) Get(hash types.Hash32) (HashPeers, bool) {
 	return hashPeers, true
 }
 
-// get is the same as Get but doesn't affect cache stats.
-func (hpc *HashPeersCache) get(hash types.Hash32) (HashPeers, bool) {
-	item, found := hpc.Cache.Get(hash)
-	if !found {
-		return nil, false
+// GetRandom returns random peer for a given hash.
+func (hpc *HashPeersCache) GetRandom(hash types.Hash32, rng *rand.Rand) (p2p.Peer, bool) {
```
countvonzero, line 78: is there a lock missing here?

fetch/fetch.go
```diff
@@ -645,6 +640,9 @@ func (f *Fetch) RegisterPeerHashes(peer p2p.Peer, hashes []types.Hash32) {
 
 // AddPeersFromHash adds peers from one hash to others.
 func (f *Fetch) AddPeersFromHash(fromHash types.Hash32, toHashes []types.Hash32) {
+	f.hashToPeers.Mu.Lock()
+	defer f.hashToPeers.Mu.Unlock()
```
countvonzero: if you manage to contain the locking inside cache.go, then there is no need to export the lock. just like what you suggested in DM, add AddPeersFromHash() to cache too and make this the wrapper.

countvonzero:
@evt can you write a test that failed before and now passed with `-count 100` and `-race`?

fetch/cache.go
evt: yup, added, cheers

fetch/fetch.go
evt: done, cheers

fetch/cache.go
```diff
@@ -70,6 +70,9 @@ func (hpc *HashPeersCache) Get(hash types.Hash32) (HashPeers, bool) {
 
 // GetRandom returns random peer for a given hash.
 func (hpc *HashPeersCache) GetRandom(hash types.Hash32, rng *rand.Rand) (p2p.Peer, bool) {
+	hpc.mu.Lock()
+	defer hpc.mu.Unlock()
+
 	hashPeersMap, exists := hpc.Get(hash)
```
countvonzero: it's bad design to have a public API Get that's not thread-safe, but others are (GetRandom/Add).
i think you should rename `Get` to `getWithStats` instead.

evt: now we have get, getWithStats and Get as well as Add and add (the latter is needed in AddPeersFromHash)

countvonzero: hmmmmm... no. my comment above is to `rename`... 

we should only have
GetRandom/RegisterPeerHashes/AddPeersFromHash as the public API of cache.go. they all take a lock at the beginning.
then there is get/getWithStats/add that do not take locks.

countvonzero: or consider adding a boolean to `get()` to indicate whether to update stats or not.

evt: Removed cache.Get, added getCachedEntry helper in tests as you suggested :) cheers. Plus added tests for GetRandom/RegisterPeerHashes/AddPeersFromHash.

countvonzero (approved):

evt:
bors merge

PR #3307, opened 2022-06-29
[Merged by Bors] - janitor: push all network requests to fetcher

countvonzero:
## Motivation

refactor the fetch package to achieve separation of concerns:
- layer.go handle the per epoch/layer data
- handler.go handle peer requests
- fetch.go does all the actually peer requests

## Changes

- create a handler.go to decouple the circular dependency of Logic and Fetcher and allows mocking out network/peers
- move the actual peer requests to Fetcher. Logic will handle the sync logic by epoch/layer

## Test Plan

unit test, systest

## DevOps Notes

### DIFF
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -414,20 +414,25 @@ func (h *Handler) handleAtxData(ctx context.Context, data []byte) error {
 // FetchAtxReferences fetches positioning and prev atxs from peers if they are not found in db.
 func (h *Handler) FetchAtxReferences(ctx context.Context, atx *types.ActivationTx) error {
 	logger := h.log.WithContext(ctx)
+	var atxIDs []types.ATXID
 	if atx.PositioningATX != *types.EmptyATXID && atx.PositioningATX != h.goldenATXID {
 		logger.With().Debug("going to fetch pos atx", atx.PositioningATX, atx.ID())
-		if err := h.fetcher.FetchAtx(ctx, atx.PositioningATX); err != nil {
-			return fmt.Errorf("fetch positioning ATX: %w", err)
-		}
+		atxIDs = append(atxIDs, atx.PositioningATX)
 	}
 
 	if atx.PrevATXID != *types.EmptyATXID {
 		logger.With().Debug("going to fetch prev atx", atx.PrevATXID, atx.ID())
-		if err := h.fetcher.FetchAtx(ctx, atx.PrevATXID); err != nil {
-			return fmt.Errorf("fetch previous ATX ID: %w", err)
+		if len(atxIDs) < 1 || atx.PrevATXID != atxIDs[0] {
+			atxIDs = append(atxIDs, atx.PrevATXID)
 		}
 	}
-	logger.With().Debug("done fetching references for atx", atx.ID())
+	if len(atxIDs) == 0 {
+		return nil
+	}
 
+	if err := h.fetcher.GetAtxs(ctx, atxIDs); err != nil {
+		return fmt.Errorf("fetch referenced atxs: %w", err)
+	}
+	logger.With().Debug("done fetching references for atx", atx.ID())
 	return nil
 }
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -17,19 +17,27 @@ import (
 	"github.com/spacemeshos/go-spacemesh/p2p/server"
 )
 
-var emptyHash = types.Hash32{}
-
 const (
-	fetchProtocol = "/sync/2"
-	batchMaxSize  = 20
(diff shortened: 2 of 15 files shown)


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7123410838?check_suite_focus=true)

unit test timed out

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7124404654?check_suite_focus=true)

https://github.com/spacemeshos/go-spacemesh/issues/3300

activation/handler.go
```diff
@@ -414,20 +414,25 @@ func (h *Handler) handleAtxData(ctx context.Context, data []byte) error {
...
+	var atxIDs []types.ATXID
 	if atx.PositioningATX != *types.EmptyATXID && atx.PositioningATX != h.goldenATXID {
 		logger.With().Debug("going to fetch pos atx", atx.PositioningATX, atx.ID())
-		if err := h.fetcher.FetchAtx(ctx, atx.PositioningATX); err != nil {
-			return fmt.Errorf("fetch positioning ATX: %w", err)
-		}
+		atxIDs = append(atxIDs, atx.PositioningATX)
 	}
 
 	if atx.PrevATXID != *types.EmptyATXID {
```
evt, line 423: I wonder why we still don't have `atx.PrevATXID.IsEmpty()` instead of this 🤔

activation/handler.go
```diff
@@ -414,20 +414,25 @@ func (h *Handler) handleAtxData(ctx context.Context, data []byte) error {
...
 		logger.With().Debug("going to fetch prev atx", atx.PrevATXID, atx.ID())
-		if err := h.fetcher.FetchAtx(ctx, atx.PrevATXID); err != nil {
-			return fmt.Errorf("fetch previous ATX ID: %w", err)
+		if len(atxIDs) < 1 || atx.PrevATXID != atxIDs[0] {
```
evt, line 425: Does `< 1` mean `== 0` here? :) asking because 4 lines later you check it as `if len(atxIDs) == 0 {`

fetch/fetch.go
```diff
@@ -629,6 +578,46 @@ func (f *Fetch) GetHash(hash types.Hash32, h datastore.Hint, validateHash bool)
...
+// GetLayerData get layer data from peers.
+func (f *Fetch) GetLayerData(ctx context.Context, lid types.LayerID, okCB func([]byte, p2p.Peer, int), errCB func(error, p2p.Peer, int)) error {
+	remotePeers := f.host.GetPeers()
+	numPeers := len(remotePeers)
+	if numPeers == 0 {
+		return errNoPeers
+	}
+
+	for _, p := range remotePeers {
+		peer := p
```
evt, line 585: why do we need this copy?

activation/handler.go
countvonzero: there is/was a struct somewhere that uses *ATXID ptr to indicate it being optional field. don't know the exact story about that. beside that, almost all types have EmptyXXX to compare with, because the EmptyXXX means something is used for special occasion.

activation/handler.go
countvonzero: yes. it has to be.

fetch/fetch.go
countvonzero: this is the classic error of passing an iteration variable to function executed in a separate goroutine.
it's the same thing that we have `tc := tc` in the test setup.

this remind me that i haven't added tests for this. now added. PTAL @evt

countvonzero: btw, you can try this out. if you don't copy, the okCB created at iteration 0 can end up using the peer value at iteration 3.

evt: I've missed goroutine which is probably run in `f.lyrSrv.Request` :)

evt (approved):

countvonzero:
bors merge

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7140021559?check_suite_focus=true)

https://github.com/spacemeshos/go-spacemesh/runs/7140018303?check_suite_focus=true
all tests passed but ci-stage1 still exits with 1 😒

countvonzero:
bors merge

PR #3313, opened 2022-07-01
[Merged by Bors] - deflake cmd/node tests

dshulyak:
- replace time sleeps with require.Eventually
- wait for event subscription to initialize, otherwise there is always a chance to miss an event and get stuck
- removed debug logs from test and added back -v so that we can see which test hangs


evt (approved):

Makefile
```diff
@@ -121,7 +121,7 @@ endif
 test: UNIT_TESTS = $(shell go list ./...  | grep -v systest)
 
 test: get-libs
-	$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) go test -timeout 0 -p 1 $(UNIT_TESTS)
+	$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) go test -timeout 0 -v -p 1 $(UNIT_TESTS)
```
countvonzero, line 124: yay.... was not happy when it was removed a few months back....

countvonzero (approved):

dshulyak:
bors merge

PR #3314, opened 2022-07-01
3281 crud epoch status draft

### LINKED ISSUE #3281: SmesherService.SmeshingStatus API Endpoint Proposal
### Discussed in https://github.com/spacemeshos/go-spacemesh/discussions/3266

<div type='discussions-op-text'>

create new endpoint `SmesherService.SmeshingStatus`, which will serve info about epochs

by default, endpoint should serve return a list of `SmeshingStatusResponse`, which contain info about `current` epoch, `next` and `next after next`. Other epoch can be requested from query.
```
type SmeshingStatusResponse struct {
  status: 'not_eligible',
  epoch: uint64;
  estimatedTime: number;
  reason: string;
  processedBytes: string // like "Processed 176 GB of 256 GB"
  processedPercentage: string // like "86% completed"
}
```

<sup>Originally posted by **brusherru** June 19, 2022</sup>
## Rationale
We want to make clear to Smeshers what is going on at every moment in the time:
- does his Node smeshes properly
- does he eligible for rewards in the current/next epoch
- if not — then why

Let's refresh our memory on how it works, assign some statuses to each "step" and clarify which additional data we may require.
1. Not smeshing yet
    Status: Not smeshing
1. Smesher set up smeshing: define PoS directory, data units amount, coinbase. The node starts creating PoS data.
    Status: Generating PoS data
2. When it is done, the Node waits for PoET
    Status: Waiting for PoET challenge
    Extra: estimated time / layer
3. Node generates proof-of-space
    Status: Generating Proof-of-Space
    Extra: time estimation
4. Node publishes ATX and starts waiting for the next epoch
    Status: Waiting for the next epoch
    Extra: time estimation
5. If there were no errors and timing issues — Smesher is eligible for rewards in the started epoch
    Status: Smeshing on epoch 23 (current)
6. If Smesher is eligible for rewards, then he will start receiving them within this epoch
    Status: Smeshing on epoch 23 (current)
7. Meanwhile, Node starts waiting for PoET challenge for the next epoch (goto 3)

Here is a good illustration by @noamnelke:
smeshing scheme
We already had some related discussion, so you can check it out: https://github.com/spacemeshos/smapp/issues/721

## Problem
Some of the steps, mentioned above, take a long time.
Some of them may finish with failure, which means that time from the 2nd step to the 6th (getting rewards) may take much more time. So we need to notify Smesher about the current status.

We already have some API endpoints, but they have a narrow scope and do not cover all the steps from the list above:
- `SmesherService.IsSmeshing -> bool` — true if PoS creation complete and "ready to send ATXs". But actually, if the node does not sync, it is still true. So from my point of view, it seems almost useless (the same as `PostSetupStatus === complete`). Fix me if I miss something.
- `SmesherService.PostSetupStatus` — enum (not_started | in_progress | complete | error). Related only to the PoS creating process.

So we need some API endpoints that will provide other statuses. Or, I think it will be much better, one API endpoint that will answer all these questions.

## Proposal
Let's make a brand-new `SmesherService.SmeshingStatus` endpoint, that will return any of the statuses mentioned above with required extra data.
Such a solution will encapsulate some logic on the Node side, like requesting/listening for ATXs, rewards, PoET, etc.
Actually, I've already created one issue for it some time ago: https://github.com/spacemeshos/go-spacemesh/issues/3222

So I think it might be a nullary endpoint, which returns something like:
```
SmeshingStatusResponse =
  | { status: 'not_started' }
  | {
       status: 'generating_pos_data' | 'generating_pos_proof' | ...
       estimatedTime: number;
    }
  | ...
  | {
       status: 'waiting_for_poet',
       epoch: number;
       estimatedTime/Layer: number;
    }
  | {
       status: 'smeshing', 
       epoch: number;
    }
  | {
       status: 'not_eligible',
       epoch: uint64;
       reason: string;
    };
```

## Multiple statuses at once
Actually, Smesher can have more than one status at a time, regarding different epochs.

For example:
- Epoch 23: smeshing
- Epoch 24: waiting for PoET

Or even:
- Epoch 23: smeshing finished
- Epoch 24: Not eligible, missed timeframe
- Epoch 25: Waiting for PoET

So probably this endpoint should return a list of `SmeshingStatusResponse`.
But in this case, we have an open question, which epochs should be on the list. Putting all previous epochs is a bad idea.

Can we implement only streaming API endpoint, but with one requirement: can it send "current status" to the new client on connection?

## What's next
Please, share your thoughts about the problem and my proposal.
If you found that such implementation will be complicated / not optimal / etc — tell us about it, please.
If I miss something — feel free to fix me. If you have other ideas — share them.

---
P.S.
We also had a discussion about API for retrieving ATXs. But from my point of view, it is not required for Smapp/Smrepl, in case all this "smeshing status logic" will be encapsulated on the Node side, as proposed above.
However, most likely we will require such an API for Explorer. And I think we should not make Smapp/Smrepl more complicated, and just put a link to explorer where it is needed (aka on Smesher screen).

So I propose to leave API for retrieving ATXs is out of scope, and open another discussion/issue in the context of Explorer.

---
cc: @oriya @lrettig @noamnelke @dshulyak @countvonzero</div>

abergasov:
## Motivation
Closes #3281

## Changes
related api changes pr https://github.com/spacemeshos/api/pull/170
added endpoint serving status current epoch.
```
type SmeshingStatusResponse struct {
  status: enum,
  epoch: uint64;
  estimatedTime: number; // optional. for statuses 'generating_pos_data' | 'generating_pos_proof' | 'waiting_for_poet'
  reason: string; // optional. possible error of file generating
}
```
to existing endpoints was added fields for show progress of post generating
```
message PostSetupStatus {
    ...
    string processed_hrp = 5; // The processed size, like `176 GB`
    string processing_total_size = 6; // Total size, that should be processed, like `256 GB`
    uint32 processed_percentage = 7; // The processed percentage, like `50`
}
```

endpoint select info and show result. 

#### Showing data about smeshing progress
as we have indexes in table `atxs` for `(epoch, smesher)` - in endpoint request need `smesherID` which will be used for load data

## TODO

## DevOps Notes


dshulyak:
post is not built in every epoch. it is built once at the start, and on demand if smesher wants to resize it

atxs are built for each epoch. for all past epochs smeshing status can be based on the atx that is stored in the database (or not stored). for current epoch it can be based on the state of the atx builder. saving additional data will introduce inconsistencies between the actual state of the atx builder and what is stored in the database.

i would suggest the following implementation plan:
- dont add new tables
- for current epoch get the smeshing status based on the state of the atx builder
- for past epochs drop smeshing status. for all UI purposes existing activations can be used
- consider providing stream api for smeshing statuses

Issue #3199, opened 2022-05-08
hare: save certificate on completion


dshulyak:
- [x] 1. fix https://github.com/spacemeshos/go-spacemesh/issues/2942
completed in #3449 
- [x] 2. persist certificate
completed in #3449 

below moved to https://github.com/spacemeshos/go-spacemesh/issues/3588 for separate tracking
- [ ] 3. evict it after hdist layers
https://github.com/spacemeshos/go-spacemesh/issues/3199#issuecomment-1221321627

it will be used by peers that want to download recent hare output

jonZlotnik:
completed in https://github.com/spacemeshos/go-spacemesh/pull/3449

jonZlotnik:
#3449

jonZlotnik:
completed in #3449

countvonzero:
@jonZlotnik 
i dont understand why you keep closing this issue when its not closed. the PR explicitly said it is partly done. the eviction is not done yet, and its pending verifying/changing the tortoise code.

jonZlotnik:
got too excited when #3449  merged 🤦
I thought this issue just failed to close automatically.

countvonzero:
moved the 3rd item to https://github.com/spacemeshos/go-spacemesh/issues/3588
and closing this one.

PR #3332, opened 2022-07-05
3283 certified block upon hare termination

### LINKED ISSUE #3283: hare: produce a Certified Block upon hare termination
AS A smeshing node who missed a hare consensus process and/or is syncing within hdist, I WANT a canonical certificate-like message that I can request from my peers on SO THAT I don't need to wait for Tortoise in order to vote and produce ballots on said layer.

### Implementation proposal overview
A new rolacle verifiable role in the hare protocol is defined as follows:
Upon receiving f+1 notify messages and generating the layer's block, if chosen by the rolacle, this party should sign the block and gossip both their eligibility proof for this role and the block signature. (block not included in message)

All other parties (active or not) should always be listening for the messages specified above. If one is received, it should be validated, verified, and stored for hdist layers. If ever requested for a Certified Block by a peer, they should provide the full Block along with as many of those signature messages as they've received for that layer up to f+1.

As a node requesting a Certified Block, I may need to contact multiple of my peers, and/or continue listening to gossip in order to accumulate enough block signatures to convince myself that it was truly the output of hare. This is especially the case in the couple layers leading up to the current layer.

### Potential for Failure
It is possible that the producing of a Certified Block for a given layer may fail (i.e. there is no guarantee that your peers will have collected f+1 signatures for you or that you will see them in gossip). In which case a smeshing node, who hadn't witnessed the hare protocol, may need to wait for Tortoise to catch up (worst case is waiting until the layer in question exits hdist).

jonZlotnik:
## Motivation

Closes #3283 

## TODO

## DevOps Notes


hare/interfaces.go
```diff
@@ -15,7 +15,9 @@ type layerPatrol interface {
 
 // Rolacle is the roles oracle provider.
 type Rolacle interface {
-	Validate(context.Context, types.LayerID, uint32, int, types.NodeID, []byte, uint16) (bool, error)
+	Validate(ctx context.Context, layerID types.LayerID,
```
countvonzero: why make it different from the rest of the interface?

hare/certifiedblock.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return service, fmt.Errorf("not yet implemented")
+}
+
+func (s *BlockCertifyingService) Start(ctx context.Context) error {
+
+	return fmt.Errorf("not yet implemented")
+}
+
+func (s *BlockCertifyingService) GossipHandler() pubsub.GossipHandler {
```
countvonzero: how is this going to be invoked? are you planning on creating a new func for each gossip msg? if so, why?

countvonzero (commented):
thanks for sharing. tho there isn't enough for me to understand what you are trying to do. 

mainly i'd like to understand
- what is the criteria for a node to be eligible to certify/sign the BlockID?
- do we expect all nodes to receive gossip of terminationSignature and therefore can generate the certificate on its own if its online?
- is BlockCertifyingService going to be a long-running goroutine or instantiated per layer? what is its relationship with the main hare process?
- which entities are subscribing to wait for the BlockCertificate?
- if you plan to also share the terminationSignature with the peers when they ask, how?
- how is the certificate persisted?

hare/interfaces.go
jonZlotnik: I will updated the whole interface to have named parameters.

hare/certifiedblock.go
jonZlotnik: No, it's just a "getter" for the func.
It will be called at service registration time in the node.go file

jonZlotnik:
> thanks for sharing. tho there isn't enough for me to understand what you are trying to do.
Will change implementation to use the database as "blackboard" to comply with patterns seen in the rest of the codebase.
> 
> mainly i'd like to understand
> 
> * what is the criteria for a node to be eligible to certify/sign the BlockID?
To be eligible, you need to be chosen by the oracle. That's it. A vrf to prove role just like every other committee in hare.
> * do we expect all nodes to receive gossip of terminationSignature and therefore can generate the certificate on its own if its online?
Yes. Any online node can "generate" the certificate. It's just a collection of signatures.
A full certificate should only be sent over wire if directly requested from a peer (fetcher).

> * is BlockCertifyingService going to be a long-running goroutine or instantiated per layer? what is its relationship with the main hare process?
From the perspective of the exported API, it should look like it does nothing but return the gossip handler. It's role will be clearer once I push the implementation. It's meant to be injected into short-lived/ephemeral objects of the surrounding code. 
BlockCertifyingService is responsible for orchestrating collection and caching of individual signatures, and persisting certificates upon collecting enough validated signatures.

> * which entities are subscribing to wait for the BlockCertificate?
> * if you plan to also share the terminationSignature with the peers when they ask, how?
> * how is the certificate persisted?

jonZlotnik:
> thanks for sharing. tho there isn't enough for me to understand what you are trying to do.
Will change implementation to use the database as "blackboard" to comply with patterns seen in the rest of the codebase.
> 
> mainly i'd like to understand
> 
> * what is the criteria for a node to be eligible to certify/sign the BlockID?
To be eligible, you need to be chosen by the oracle. That's it. A vrf to prove role just like every other committee in hare.
> * do we expect all nodes to receive gossip of terminationSignature and therefore can generate the certificate on its own if its online?

Yes. Any online node can "generate" the certificate. It's just a collection of signatures.
A full certificate should only be sent over wire if directly requested from a peer (fetcher).

> * is BlockCertifyingService going to be a long-running goroutine or instantiated per layer? what is its relationship with the main hare process?

From the perspective of the exported API, it should look like it does nothing but return the gossip handler. It handles the time-based logic, and vrf-signature calls to the roracle. It's role will be clearer once I push the implementation. It's meant to be injected into short-lived/ephemeral objects of the surrounding code. 

> * which entities are subscribing to wait for the BlockCertificate?

So I originally thought the syncer would want to do that. But as discussed last week, the syncer and fetcher use more of a blackboard pattern with the database. So the channel getter will probably not be exported in the end and I'll only use it for deciding when to commit to database.

> * if you plan to also share the terminationSignature with the peers when they ask, how?

Single terminationSignatures should only be shared through gossip.
Ideally, upon request, a group of signatures (even if not enough to certify a block) should be able to be fetched from a peer.
That will probably be a "next step".

> * how is the certificate persisted?

Once enough signatures are collected, they're persisted in the database as discussed.

hare/certifiedblock.go
countvonzero: yeah... but it is not consistent with other handlers. other handlers simply pass the func ptr directly. what is the advantage of you doing differently?

hare/interfaces.go
countvonzero: i actually hold the opposite view. the named parameters doesn't really help me understand what the code is doing. i find it easier to just read the code/implementation. these interfaces are defined mostly for tests.

blockcerts/certificatestoring.go
```diff
@@ -0,0 +1,26 @@
...
+func certificateStoringLoop(ctx context.Context,
+    completedCerts <-chan types.BlockCertificate, db sql.Executor, logger log.Logger) {
+
+    logger = logger.WithContext(ctx)
+    for {
+        select {
+        case <-ctx.Done():
+            return
+        case cert := <-completedCerts:
+            err := certifiedblocks.Add(db, &cert)
```
dshulyak, line 20: please do it in the handler (or in the cache) no need to spin up another goroutine

blockcerts/blocksigning.go
```diff
@@ -0,0 +1,83 @@
...
+                break
+            }
+            err = gossipPublisher.Publish(ctx, BlockSigTopic, msgBytes)
+            if err != nil {
+                logger.Error("BlockCertifyService: failed to publish"+
+                    "block signature: %v", err)
+                break
+            }
+
+            signatureCache.CacheBlockSignature(ctx, blockSigMsg)
```
dshulyak: handler will be called on publish

blockcerts/blocksigning.go
```diff
@@ -0,0 +1,83 @@
...
+
+            msgBytes, err := codec.Encode(blockSigMsg)
+            if err != nil {
+                logger.Error("BlockCertifyService: failed to encode"+
+                    "block signature: %v", err)
+                break
+            }
+            err = gossipPublisher.Publish(ctx, BlockSigTopic, msgBytes)
+            if err != nil {
+                logger.Error("BlockCertifyService: failed to publish"+
```
dshulyak, line 75: please use logger with fields like the rest of the codebase

hare/hare.go
```diff
@@ -1,24 +1,24 @@
...
-	"github.com/spacemeshos/go-spacemesh/hare/config"
-	"github.com/spacemeshos/go-spacemesh/log"
-	"github.com/spacemeshos/go-spacemesh/p2p"
-	"github.com/spacemeshos/go-spacemesh/p2p/pubsub"
-	"github.com/spacemeshos/go-spacemesh/sql"
-	"github.com/spacemeshos/go-spacemesh/sql/ballots"
-	"github.com/spacemeshos/go-spacemesh/sql/layers"
-	"github.com/spacemeshos/go-spacemesh/sql/proposals"
-	"github.com/spacemeshos/go-spacemesh/system"
+    "context"
```
dshulyak, line 4: something got messed up in this file

blockcerts/signaturetracker.go
```diff
@@ -0,0 +1,145 @@
...
+
+type trackableBlockSignature struct {
+    BlockID types.BlockID
+    certtypes.BlockSignature
+}
+type blockSignatureTracker struct {
+    layerID        types.LayerID
+    signatures     map[types.NodeID]trackableBlockSignature
+    thresholdCount map[types.BlockID]int
+    sync.Mutex
```
dshulyak: please rename to `mu sync.Mutex`, this can't be public

blockcerts/signaturetracker.go
```diff
@@ -0,0 +1,145 @@
...
+    completedCerts chan<- certtypes.BlockCertificate
+    ifNotAlready   sync.Once // completedCerts channel only gets one cert
```
dshulyak: there is no need for that, everything is already wrapped in mutex

blockcerts/signaturetracker.go
```diff
@@ -0,0 +1,145 @@
...
+
+    return &newTracker
+}
+
+// addSig doesn't do any validation or verification and is a blocking operation.
+// It should be run in its own goroutine.
+func (t *blockSignatureTracker) addSig(sig certtypes.BlockSignatureMsg) {
+    t.logger.Debug("addSig entry")
+    t.Lock()
+    t.logger.Debug("addSig: acquired Write Lock on self (sigTracker)")
```
dshulyak: i would avoid such low level debug logs

blockcerts/signaturetracker.go
```diff
@@ -0,0 +1,145 @@
...
+    logger         log.Logger
+}
+
+func newBlockSignatureTracker( // did I forget to pass in the threshold?
+    layerID types.LayerID, completedCerts chan<- certtypes.BlockCertificate, logger log.Logger) *blockSignatureTracker {
+    newTracker := blockSignatureTracker{
+        layerID:        layerID,
+        signatures:     map[types.NodeID]trackableBlockSignature{},
+        thresholdCount: map[types.BlockID]int{},
+        Mutex:          sync.Mutex{},
```
dshulyak: zero value is ready to be used

blockcerts/signaturetracker.go
```diff
@@ -0,0 +1,145 @@
...
+// It should be run in its own goroutine.
+func (t *blockSignatureTracker) addSig(sig certtypes.BlockSignatureMsg) {
+    t.logger.Debug("addSig entry")
+    t.Lock()
+    t.logger.Debug("addSig: acquired Write Lock on self (sigTracker)")
+    defer t.Unlock()
+    // check if a BlockID reached threshold # of signatures
+    var majorityBlockID types.BlockID
+    var thresholdReached = false
+    for blockID, threshold := range t.thresholdCount {
```
dshulyak: why to iterate over all blocks when adding a single sig?

blockcerts/types/types.go
```diff
@@ -0,0 +1,24 @@
...
+type BlockCertificateConfig struct {
+    CommitteeSize int
+}
+type BlockCertificate struct {
+    BlockID               types.BlockID
+    LayerID               types.LayerID
+    TerminationSignatures []BlockSignature
+}
+type BlockSignature struct {
+    SignerNodeID         types.NodeID
```
dshulyak, line 11: this should be recoverable from signature

blockcerts/certifiedblock.go
```diff
@@ -0,0 +1,117 @@
...
+    "github.com/spacemeshos/go-spacemesh/p2p/pubsub"
+    "github.com/spacemeshos/go-spacemesh/signing"
+    "github.com/spacemeshos/go-spacemesh/sql"
+    "math"
+    "sync"
+)
+
+const (
+    blockCertifierRole uint32 = math.MaxUint32 - 2        // for Rolacle
+    BlockSigTopic             = "trackableBlockSignature" // for gossip
```
dshulyak: this would need to be reduced, see other names

blockcerts/certifiedblock.go
```diff
@@ -0,0 +1,117 @@
...
+    }
+    service.completedCertsCh = make(chan certtypes.BlockCertificate, 100)
+    // TODO: empirically decide on adequate channel size
+
+    service.logger = logger
+    return service, nil
+}
+
+func (s *BlockCertifyingService) Start(ctx context.Context) error {
+    go blockSigningLoop(ctx, s.hareTerminationsCh,
```
dshulyak: we are using errgroup to control service goroutines lifetimes

blockcerts/certifiedblock.go
```diff
@@ -0,0 +1,117 @@
...
+        s.blockSigner, s.config.CommitteeSize, s.rolacle,
+        s.gossipPublisher, &s.signatureCache, s.logger,
+    )
+    go certificateStoringLoop(ctx, s.completedCertsCh, s.db, s.logger)
```
dshulyak: this is unnecessary

blockcerts/certifiedblock.go
```diff
@@ -0,0 +1,117 @@
...
+            log.Error("hare termination gossip: %w", err)
+            return pubsub.ValidationReject
+        }
+        if !isValid {
+            logger.With().Debug("certified block: gossip hander: received "+
+                "invalid block signature", msg.BlockID, msg.LayerID, msg.SignerNodeID)
+            return pubsub.ValidationReject
+        }
+
+        s.SignatureCacher().CacheBlockSignature(ctx, msg)
```
dshulyak: i suggest to save certificate after calling this function, e.g

```
cert := Cache(...)
if cert != nil {
   db.Add(...)
}
```

or any other synchronous interface, if you don't like my suggestion

sql/migrations/0001_initial.sql
```diff
@@ -159,4 +159,11 @@ CREATE TABLE accounts
     PRIMARY KEY (address, layer_updated DESC)
 );
 
-CREATE INDEX accounts_by_layer_udated ON accounts (layer_updated);
\ No newline at end of file
+CREATE INDEX accounts_by_layer_udated ON accounts (layer_updated);
+
+CREATE TABLE certified_blocks
```
dshulyak, line 165: why not to store it in the blocks table? i think there is already a field that tracks if block is a hare output or not

blockcerts/certifiedblock.go
```diff
@@ -0,0 +1,117 @@
...
+    go certificateStoringLoop(ctx, s.completedCertsCh, s.db, s.logger)
+    return nil
+}
+
+func (s *BlockCertifyingService) SignatureCacher() SigCacher {
+    return &s.signatureCache // exported functions are thread safe
+}
+
+// GossipHandler returns a function that handles incoming BlockSignatureMsg gossip.
+func (s *BlockCertifyingService) GossipHandler() pubsub.GossipHandler {
```
dshulyak, line 91: i don't get why to create a clojure from this method

`GossipHandler(ctx context.Context, peerID p2p.Peer, bytes []byte)`

blockcerts/blocksigning.go
```diff
@@ -0,0 +1,83 @@
...
+                logger.Error("could not retrieve eligibility proof "+
+                    "from oracle: %v", err)
+                break
+            }
+            committeeSeatCount, err := rolacle.CalcEligibility(ctx,
+                hareTermination.LayerID, blockCertifierRole, committeeSize,
+                hareTermination.TerminatingNodeID, proof)
+
+            blockIDBytes := hareTermination.BlockID.Bytes()
+            blockIDSignature := blockSigner.Sign(blockIDBytes)
```
dshulyak, line 57: why only block id is signed? i would expect that whole message needs to be signed

dshulyak (changes requested):

sql/migrations/0001_initial.sql
dshulyak: hare output is stored in layers for some reason. but anyway, the certificate can be stored in blocks table. just add another field

common/types/hashes.go
```diff
@@ -16,18 +16,18 @@ import (
 const (
 	// Hash32Length is 32, the expected length of the hash.
 	Hash32Length = 32
-	hash20Length = 20
-	hash12Length = 12
+	Hash20Length = 20
+	Hash12Length = 12
```
dshulyak, line 20: unrelated change

p2p/pubsub/pubsub.go
```diff
@@ -27,6 +27,7 @@ type Config struct {
 }
 
 // New creates PubSub instance.
+// TODO: unexport or refactor away (only used in tests and host.go)
```
dshulyak, line 30: New is used in the cmd/node. please remove this comment

dshulyak: please remove this comment, how you can unexport it and still use in host.go?

blockcerts/certifiedblock.go
jonZlotnik: So that if we want to change where the Gossip Handler is defined in the future (or we want to switch it out depending on some state of the service or add extra "call only once logic") we don't need to break the exported api of the BlockCertifyingService. Plenty of good reasons, I think?
It decouples the handler implementation from the BlockCertifyingService exported API.

sql/migrations/0001_initial.sql
countvonzero: agreed.

hare output is in table `layers` as there is strict 1-1 relationship layer->hare output.
anyway it's orthogonal to this. the certificate should be stored in table `blocks`

blockcerts/types/types.go
```diff
@@ -0,0 +1,21 @@
...
+
+type BlockCertificate struct {
+    BlockID               types.BlockID
+    LayerID               types.LayerID
+    TerminationSignatures []BlockSignature
+}
+type BlockSignature struct {
+    SignerNodeID         types.NodeID
+    SignerRoleProof      []byte
+    SignerCommitteeSeats uint16
```
countvonzero, line 13: - shorten the name to RolProof and CommitteeSeats
- add comments on what each field represents

blockcerts/types/types.go
```diff
@@ -0,0 +1,21 @@
...
+
+    BlockIDSignature []byte
```
countvonzero, line 15: add comment on what you are signing here.

blockcerts/types/types.go
```diff
@@ -0,0 +1,21 @@
+package types
+
+import "github.com/spacemeshos/go-spacemesh/common/types"
+
+type BlockCertificate struct {
+    BlockID               types.BlockID
+    LayerID               types.LayerID
+    TerminationSignatures []BlockSignature
```
countvonzero, line 8: shorten to `Signatures`

blockcerts/types/types.go
```diff
@@ -0,0 +1,21 @@
...
+    SignerRoleProof      []byte
+    SignerCommitteeSeats uint16
+
+    BlockIDSignature []byte
+}
+type BlockSignatureMsg struct {
+    LayerID types.LayerID
+    BlockID types.BlockID
+    BlockSignature
+}
```
countvonzero, line 21: why is there LayerID and BlockID in both here and BlockSignature?

hare/eligibility/oracle.go
```diff
@@ -561,8 +561,9 @@ func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID) (map[ty
 	return activeMap, nil
 }
 
-// IsIdentityActiveOnConsensusView returns true if the provided identity is active on the consensus view derived
-// from the specified layer, false otherwise.
+// IsIdentityActiveOnConsensusView returns true if the provided identity is
+// active on the consensus view derived from the specified layer, false
+// otherwise. Active means ready to request eligibility.
```
countvonzero, line 566: Active means it has submitted an ATX in the previous epoch and is eligible to participate in the protocol this epoch.
any node can request eligibility....

hare/haretypes.go
```diff
@@ -22,8 +22,14 @@ const (
 	pre      messageType = 10
 )
 
+// declare epi-round constants
 const preRound uint32 = math.MaxUint32
 
+// terminationRound is not actually a round. A UInt32, however, is required by
+// the Rolacle for the eligibility.buildKey function. Thus, we define a constant
+// that represents the termination certification process for every layer.
+const terminationRound uint32 = math.MaxUint32 - 1
```
countvonzero, line 31: where is this used? 

and please don't do what the previous author did again. please don't define it as a hare round.

hare/haretypes.go
```diff
@@ -22,8 +22,14 @@ const (
 	pre      messageType = 10
 )
 
+// declare epi-round constants
```
countvonzero, line 25: what does `epi-round` mean?

sql/certifiedblocks/certifiedblocks.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/spacemeshos/go-spacemesh/codec"
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+func Add(db sql.Executor, blockCert *certtypes.BlockCertificate) error {
+    certBytes, err := codec.Encode(blockCert)
+    if err != nil {
+        return fmt.Errorf("encoding cert for block %s: %w",
+            blockCert.BlockID.String(), err)
```
countvonzero, line 15: %s will call String() for you

sql/certifiedblocks/certifiedblocks.go
```diff
@@ -0,0 +1,61 @@
...
+    }
+    _, err = db.Exec(
+        `insert into certified_blocks (block_id, layer_id, cert)
```
countvonzero, line 18: no need to associate layer_id to block cert. it should be 1-1 btwn blockID->cert

sql/migrations/0001_initial.sql
```diff
@@ -159,4 +159,11 @@ CREATE TABLE accounts
...
+(
+    block_id CHAR(20) PRIMARY KEY,
+    layer_id CHAR(32),
```
countvonzero, line 168: layer_id is saved as INT everywhere else.

blockcerts/blocksigning.go
jonZlotnik: I was under the impression that the public key was the same for the VRF signer and the regular ed signer. 
https://github.com/spacemeshos/go-spacemesh/blob/eedff1e659de089ab4497c3614291b213f256479/signing/signer.go#L137

If this is the case, then it seems redundant to sign the whole message since every other field goes into the VRF verifier anyways.

Come to think of it, maybe we want early warning of tampering in transit so that we don't need to run the oracle validation if the msg signature is invalid: we can reject sooner. 
I guess it's also more idiomatic to sign a message than only the BlockID.

I'm convinced.

blockcerts/certifiedblock.go
countvonzero: please don't code for the future. i call this premature abstraction.
rather, code consistently with the rest of the codebase.

if you are convinced that we need it, change all the gossip handlers in a separate PR.

blockcerts/certifiedblock.go
```diff
@@ -0,0 +1,118 @@
...
+            logger.Debug("certified block: gossip handler: failed to "+
+                "decode message from peer: %s", peerID)
+            return pubsub.ValidationReject
+        }
+        isValid, err := s.rolacle.Validate(ctx,
+            msg.LayerID, blockCertifierRole, s.config.CommitteeSize,
+            msg.SignerNodeID, msg.SignerRoleProof, msg.SignerCommitteeSeats)
+        if err != nil {
+            log.Error("hare termination gossip: %w", err)
+            return pubsub.ValidationReject
```
countvonzero: this is internal error, we should just ignore instead.

blockcerts/certifiedblock.go
```diff
@@ -0,0 +1,118 @@
...
+        }
+        if !isValid {
+            logger.With().Debug("certified block: gossip hander: received "+
+                "invalid block signature", msg.BlockID, msg.LayerID, msg.SignerNodeID)
+            return pubsub.ValidationReject
```
countvonzero: i am not sure we want to reject here. reject will lower the peer score. but it's possible that the peer just have a different active set than ours, which is not uncommon/malicious.

blockcerts/signaturetracker.go
```diff
@@ -0,0 +1,139 @@
...
+    certtypes.BlockSignature
+}
+type blockSignatureTracker struct {
+    layerID        types.LayerID
+    signatures     map[types.NodeID]trackableBlockSignature
+    signatureCount map[types.BlockID]int
+    sync.Mutex
+    completedCerts     chan<- certtypes.BlockCertificate
+    signaturesRequired int
+    ifNotAlready       sync.Once // completedCerts channel only gets one cert
```
countvonzero: unused?

blockcerts/signaturetracker.go
```diff
@@ -0,0 +1,139 @@
...
+    CacheBlockSignature(ctx context.Context,
+        blockSignature certtypes.BlockSignatureMsg)
+}
+
+type sigCache struct {
+    logger             log.Logger
+    blockSigsByLayer   sync.Map // entries are written once but read many times
+    cacheBoundary      types.LayerID
+    cacheBoundaryMutex sync.RWMutex
+    signaturesRequired int
```
countvonzero: i don't understand the design of sigCache and blockSignatureTracker here.

you basically need a object, say certificateTracker that holds a `sigs map[blockID][]signatures`, and when the gossip handler call CacheBlockSignature:
```
sigs[bid] = append(sigs[bid], sig)
if len(sigs) >= signaturesRequired {
  t.completedCerts <- generateCert(sigs[bid])
}
delete(sigs, bid)
```
what are `blockSigsByLayer` and `cacheBoundary` for?

countvonzero (changes requested):
a rough review. didn't dive into the impl details too much. general suggestion

- move the cert impl to blocks package, as it is certificate for a block.
- there is no need to have a separate file for just 1 function. i.e you can merge all code logic into one file that contains BlockCertifyingService.
- shorten the naming to cert. no need to prepend block everywhere. there is only one cert in the system
- the history of this PR is already quite long. i'd prefer to remove all the unrelated changes (style/comment/rename) just so the discussion can be focused on the logic.

blockcerts/blocksigning.go
countvonzero: i don't get your reasoning on rejecting sooner. 

the point of signing is either you can trust the whole msg is from the signer or not. for example, node X can use node Y's signature on blockID and provides bad proof or wrong layer id. how are you going to figure that out if you only sign blockID?

jonZlotnik: Because node X has a different public key from node Y. 
You need to check that the regular ED signature matches the VRF signature public key whether you sign the whole message with EdSigner or just the BlockID.

The "rejecting or ignoring sooner" is just saving us the time spent for VRF proof validation in the case that some node signed the message with their public key but used some other node's VRF proof.

countvonzero: hmmm... using your code now, i can get away with this without causing an error.
```
blockSigMsg.LayerID = bogusLayerID
blockSigMsg.SignerNodeID = Y.NodeID
blockSigMsg.SignerRoleProof = Y.proof
blockSigMsg.BlockID = hareTermination.BlockID
blockSigMsg.BlockIDSignature = X.blockIDSignature
```
in the gossip handler, you need to 
- extract the public key from the signed msg after you decode it
- call roracle.Validate() with the extracted public key, not msg.SignerNodeID
that way the above scenario will be rejected. 

> The "rejecting or ignoring sooner" is just saving us the time spent for VRF proof validation in the case that some node signed the message with their public key but used some other node's VRF proof.

no. when you extract the public key from the signed message, you have no idea yet whether the public key is valid or not until you do the VRF validate. there is no saving to be had.

jonZlotnik: Yeah, this code doesn't check that yet. I was trying to find the second public key that didn't exist to make sure I wasn't missing something.

Oh, yeah I guess if we extract the public key from the signature, then we end up with nothing really useful because obviously the public key will match the signature. The only "early warning" we get is if we fail to extract a public key.

jonZlotnik: Okay, so then I finally maintain that a signature of only the BlockID is required.

countvonzero: say you are only signing BlockID, consider this:

BadBob receives a msg generated by from GoodAlice, changed `blockSigMsg.proof` to garbage, and relayed this gossip to InnocentJoe. when InnocentJoe sees that proof doesn't validate with GoodAlice's public key extracted from the blockID signature, it it marks GoodAlice malicious.

blockcerts/types/types.go
jonZlotnik: I don't understand the question. It's not in both of them.
We need the LayerID and BlockID the message structure to provide completeness such that we can validate the message upon receiving it without making assumptions about what layer/block it's referring to.
We don't need it in the BlockSignature struct because the BlockSignature struct will always be stored in a context where the BlockID that was signed is obvious. (e.g. in a BlockSignatureMsg or in a signature tracker which is counting the number of signatures for a specific block/layer id)

common/types/hashes.go
jonZlotnik: Oh, yeah I thought i needed the Hash20Length const somewhere but ended up not needing it.
I don't mind reverting, but they do seem like things that should be exported.

hare/haretypes.go
jonZlotnik: It isn't. This is left over from my trying to reuse some of the hare logic. Will be deleted.

jonZlotnik: I was worried about defining the "round parameter" or "role index" for the oracle far from the other declarations. 
Didn't want accidental overlap between hare roles and the block certifier/signer role

blockcerts/types/types.go
countvonzero: i mistook BlockSignature as BlockCertificate. sorry about the confusion.

but i don't get in general what does LayerID has to do with the certificate? the certificate is about a block. why do we care about the layer ID?

countvonzero: ok. it is needed in the Validate() call to verify proof. 

but i actually don't think we should blindly trust the layer ID specified in the msg. we have BlockID->LayerID stored in db and should look it up from there to avoid malicious message.

countvonzero:
implemented. see #3199

PR #3336, opened 2022-07-06
[Merged by Bors] - 3315 Genesis address format - address size and type change

### LINKED ISSUE #3315: Genesis address format
See https://github.com/spacemeshos/pm/issues/140

- [x] choose and import a bech32 implementation (see [this list](https://github.com/spacemeshos/pm/issues/106#issuecomment-1171613630)) (chosen github.com/cosmos/btcutil)
- [x] use this to write a simple address parser (string -> `Address`) with error-handling code for wrong format, too long, bad HRP, etc. (this should probably be published as a separate library so it can be used elsewhere)
- [x] change `Address` type to `[24]byte` (i.e., change `AddressLength`)
- [x] update API code to receive address as string and parse it
  - [x] go-spacemesh (https://github.com/spacemeshos/go-spacemesh/pull/3336)
  - [x] API (https://github.com/spacemeshos/api/pull/173)
- [x] update addresses in genesis account config and tests
- [x] register HRPs (mainnet, testnet) in [SLIP-0713](https://github.com/satoshilabs/slips/blob/master/slip-0173.md)
  - [x] wait pr to be merged https://github.com/satoshilabs/slips/pull/1386

abergasov:
## Motivation

Closes #3315

## Changes
* address moved to separate pkg. probably in future will be as independent library
* address still array of bytes ([24]byte), but implement `bech32` algorithm. Additional first byte reserved for contain HumanReadablePart of address (now contain info about networkID - test|main net), so result is [25]byte
* all api endpoints will be changed to use string instead []byte (related pr https://github.com/spacemeshos/api/pull/173)

## Test Plan
unit tests

## TODO

## DevOps Notes


dshulyak:
could you please use aliases for such refactoring?

```go
package types

type Address = address.Address
```

they were created specifically to make such changes reviewable

dshulyak (changes requested):
hrp part wont be stored on the mesh, even if it is 1 byte. it has other implications beside space

moreover:
- there will be no network ids. they will be replaced by hashes instead. so the idea to compress hrp wont work, but it is not even important
- there is no reason to hardcode every single known testnet name in advance. there are solutions that avoid that

common/types/address/address_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func generatePublicKey(t *testing.T) []byte {
+	buff := signing.NewEdSigner().ToBuffer()
+	acc1Signer, err := signing.NewEdSignerFromBuffer(buff)
+	require.NoError(t, err)
+	return acc1Signer.PublicKey().Bytes()
+}
+
+//func intToAddress(t *testing.T, num uint32) *address.Address {
```
dshulyak: rm

api/grpcserver/mesh_service.go
```diff
@@ -178,7 +179,11 @@ func (s MeshService) AccountMeshDataQuery(ctx context.Context, in *pb.AccountMes
 	filterActivations := in.Filter.AccountMeshDataFlags&uint32(pb.AccountMeshDataFlag_ACCOUNT_MESH_DATA_FLAG_ACTIVATIONS) != 0
 
 	// Gather transaction data
-	addr := types.BytesToAddress(in.Filter.AccountId.Address)
+	addr, err := address.BytesToAddress(in.Filter.AccountId.Address)
+	if err != nil {
+		log.Error(fmt.Sprintf("failed to parse Filter.AccountId address: %v", err))
```
dshulyak: please stick to logs with fields

also:
- this is a warning, not the error
- it might be already logged by grpc middleware. please remove such logs if this is the case

common/types/address/address.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+// ErrWrongAddressLength is returned when the length of the address is not correct.
+var ErrWrongAddressLength = errors.New("wrong address length")
+
+// Address represents the address of a spacemesh account with AddressLength length.
+type Address [AddressLength]byte // contains slice 8 bytes unsigned integers
+
+// NewAddress returns a new Address from a byte slice.
+func NewAddress(src string) (Address, error) {
```
dshulyak: FromString

common/types/address/config.go
```diff
@@ -0,0 +1,26 @@
+package address
+
+// Config is the configuration of the address package.
+type Config struct {
+	NetworkName string `mapstructure:"network-name"`
```
dshulyak: i think only hrp should be required to configure. why do we need NetworkName?

cmd/node/node.go
```diff
@@ -597,8 +598,12 @@ func (app *App) initServices(ctx context.Context,
 
 	nipostBuilder := activation.NewNIPostBuilder(nodeID[:], postSetupMgr, poetClient, poetDb, sqlDB, app.addLogger(NipostBuilderLogger, lg))
 
-	coinbaseAddr := types.HexToAddress(app.Config.SMESHING.CoinbaseAccount)
+	var coinbaseAddr address.Address
 	if app.Config.SMESHING.Start {
+		coinbaseAddr, err = address.NewAddress(app.Config.SMESHING.CoinbaseAccount)
+		if err != nil {
+			app.log.Panic("failed to parse CoinbaseAccount address `%s`: %v", app.Config.SMESHING.CoinbaseAccount, err)
```
dshulyak, line 608: clearly can't panic in API request, return error to notify node administrator that node is miconfigured

api/config/genesis.go
```diff
@@ -15,25 +19,39 @@ type GenesisConfig struct {
 func (g *GenesisConfig) ToAccounts() []types.Account {
 	var rst []types.Account
 	for addr, balance := range g.Accounts {
+		var (
+			err         error
+			genesisAddr address.Address
+		)
+		// this condition need in case address in config with specific hrp prefix, like smt1abc...
+		// in code we have constants which contain only bytes, without hrp prefix.
+		if strings.HasPrefix(addr, "0x") {
```
dshulyak: please remove this. either:

- use default hrp for this addresses
- check where this constants are used and remove them

api/grpcserver/globalstate_service.go
```diff
@@ -203,7 +212,11 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 	if in.Filter.AccountDataFlags == uint32(pb.AccountDataFlag_ACCOUNT_DATA_FLAG_UNSPECIFIED) {
 		return status.Errorf(codes.InvalidArgument, "`Filter.AccountDataFlags` must set at least one bitfield")
 	}
-	addr := types.BytesToAddress(in.Filter.AccountId.Address)
+	addr, err := address.BytesToAddress(in.Filter.AccountId.Address)
```
dshulyak: it should be string with hrp

api/grpcserver/globalstate_service.go
```diff
@@ -203,7 +212,11 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
...
+	if err != nil {
+		log.Error(fmt.Sprintf("failed to parse in.Filter.AccountId.Address address: %v", err))
```
dshulyak: remove or downgrade and use fields

api/grpcserver/mesh_service.go
```diff
@@ -437,7 +442,11 @@ func (s MeshService) AccountMeshDataStream(in *pb.AccountMeshDataStreamRequest,
 	if in.Filter.AccountMeshDataFlags == uint32(pb.AccountMeshDataFlag_ACCOUNT_MESH_DATA_FLAG_UNSPECIFIED) {
 		return status.Errorf(codes.InvalidArgument, "`Filter.AccountMeshDataFlags` must set at least one bitfield")
 	}
-	addr := types.BytesToAddress(in.Filter.AccountId.Address)
+	addr, err := address.BytesToAddress(in.Filter.AccountId.Address)
+	if err != nil {
+		log.Error(fmt.Sprintf("failed to parse in.Filter.AccountId.Address address: %v", err))
```
dshulyak: same

common/types/address/address.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Bytes gets the string representation of the underlying address.
+func (a Address) Bytes() []byte { return a[:] }
+
+// Big converts an address to a big integer.
+func (a Address) Big() *big.Int { return new(big.Int).SetBytes(a[:]) }
+
+// Hex returns an EIP55-compliant hex string representation of the address.
+func (a Address) Hex() string {
```
dshulyak: should be removed, this is replaced with bech32 string

common/types/address/address.go
```diff
@@ -0,0 +1,156 @@
...
+
+// Field returns a log field. Implements the LoggableField interface.
+func (a Address) Field() log.Field {
+	return log.String("address", a.String())
+}
+
+// Short returns the first 7 characters of the address hex representation (incl. "0x"), for logging purposes.
+func (a Address) Short() string {
+	hx := a.String()
+	return hx[:util.Min(7, len(hx))]
```
dshulyak: this needs to be fixed, since String will include hrp

common/types/address/address.go
```diff
@@ -0,0 +1,156 @@
...
+
+// DecodeScale implements scale codec interface.
+func (a *Address) DecodeScale(d *scale.Decoder) (int, error) {
+	return scale.DecodeByteArray(d, a[:])
+}
+
+// GenerateAddress generates an address from a public key.
+func GenerateAddress(publicKey []byte) Address {
+	var addr Address
+	addr.setBytes(publicKey)
```
dshulyak: please double check with @lrettig if we need to leave 4 top bytes empty, he mentioned it somewhere in the github issues

dshulyak (changes requested):
why cosmos/bech32 and not one from btcd repo? are they different?

common/types/address/address.go
lrettig: NewAddressFromString?

common/types/address/address.go
lrettig: correct, thanks @dshulyak

common/types/address/address.go
dshulyak: address is a part of the module, so what looks better when you use it?

address.NewAddressFromString or address.FromString

ps: i think the 1st one is incredibly ugly)

api/grpcserver/globalstate_service.go
abergasov: it will be changed, when PR  https://github.com/spacemeshos/api/pull/173 will be merged
i can add cast to string, which will be excess after api method change signature, or just wait until pr merge, that i'll replace to `FromString`

cmd/node/node.go
abergasov: this is not an api request, it node starting. On next lines, panic can happen if address check failing
```
if coinbaseAddr.Big().Uint64() == 0 {
	app.log.Panic("invalid coinbase account")
}
```

common/types/address/address.go
abergasov: i removed this method, as it used for 1 plase in logs. Just change to `String`

common/types/address/config.go
abergasov: we do not need it. Probably we will later stream it via api for better ui usability?

cmd/node/node.go
dshulyak: sorry :) for some reason i thought that i saw same line in two places

abergasov:
> why cosmos/bech32 and not one from btcd repo? are they different?

source code almost the same. small difference between const|wrappers, but in generally code the same. it can easy replaced by fixing `go.mod`

i chose cosmos cause it's approach - to get necessary functions and types, for crypto projects. in case btcd repo - we need to vendor other blockchein implementation, that's look little bit weird for me.

api/config/genesis.go
abergasov: in `address.BytesToAddress` will be used default hrp from `address.Config`

this consts are in `api/config/genesis.go`
they are used in `DefaultGenesisConfig` function 
```
type GenesisConfig struct {
	Accounts map[string]uint64 `mapstructure:"accounts"`
}

// Account1Address is the address from Account1Private.
const Account1Address = "0x0002abedccd3ffcbf46..."

// Account2Address is the address from Account2Address.
const Account2Address = "0xfeaf3d38b6ef43093..."

func DefaultGenesisConfig() *GenesisConfig {
	// NOTE(dshulyak) keys in default config are used in some tests
	g := GenesisConfig{}

	// we default to 10^5 SMH per account which is 10^17 smidge
	// each genesis account starts off with 10^17 smidge
	g.Accounts = map[string]uint64{
		Account1Address: 100000000000000000,
		Account2Address: 100000000000000000,
	}
	return &g
}
```

while able situation, when user can override with bech32 address like `sm1abc...`. 
So we should add limitation for config genesis address (like not use bech32 in address, but put `0x0123` string for genesis addr or other encoding style). Or just use such `if else`

dshulyak: sorry i don't follow. please replace 0x addresses with addresses prefixed using hrp and remove this if condition. spacecraft and system tests have to be adjusted accordingly.

common/types/address/address.go
abergasov: done, added bytes reservation + tests

common/types/address/config.go
dshulyak: i don't know why we would need to stream it. for now it just looks like totally unnecessary config option that will have be changed for every network. lets remove it

countvonzero (commented):
i am generally surprised by the number of files changed for this PR.

what is the advantage of having a separate pkg for address?
in another words, what is the disadvantage of maintaining the flat structure inside common/types?

abergasov:
> i am generally surprised by the number of files changed for this PR.
> 
> what is the advantage of having a separate pkg for address? in another words, what is the disadvantage of maintaining the flat structure inside common/types?

alias `type Address = address.Address` in common/type reduce total edited files =)

one of idea was to move address implementation to separate library, which will be included as dependency.
in this case  i moved to nested pkg. when we will publish as single library - just change imports

lrettig:
> one of idea was to move address implementation to separate library, which will be included as dependency

This is not a terrible idea - do we think other projects will have this dependency? (And would they want to use just an address module without relying on all of go-spacemesh?) E.g., sm-repl, dash/explorer, etc.

countvonzero:
> > one of idea was to move address implementation to separate library, which will be included as dependency
> 
> This is not a terrible idea - do we think other projects will have this dependency? (And would they want to use just an address module without relying on all of go-spacemesh?) E.g., sm-repl, dash/explorer, etc.

but why? what is the problem with having the impl inside common/types/address.go? the functionality you are providing in the new pkg look highly similar to the old ones. @abergasov

common/types/address/address_test.go
```diff
@@ -0,0 +1,133 @@
...
+		name string
+		src  string
+		err  error
+	}{
+		{
+			name: "correct",
+			src:  "smt1fejq2x3d79ukpkw06t7h6lndjuwzxdnj59npghsywyusj",
+		},
+		{
+			name: "wrong length",
```
lrettig: please test too long as well as too short

also, bad bech32 formats (missing hrp, invalid characters, etc.)

common/types/address/config.go
```diff
@@ -0,0 +1,23 @@
...
+
+// DefaultAddressConfig returns the default configuration of the address package.
+func DefaultAddressConfig() *Config {
+	return conf
+}
+
+// DefaultTestAddressConfig returns the default test configuration of the address package.
+func DefaultTestAddressConfig() *Config {
+	conf = &Config{
+		NetworkHRP: "smt",
```
lrettig: "smt" is too similar to "sm", this is too error-prone. per https://github.com/spacemeshos/pm/issues/106#issuecomment-1170509194, let's use "stest" for the testnet.

abergasov:
> This is not a terrible idea - do we think other projects will have this dependency? (And would they want to use just an address module without relying on all of go-spacemesh?) E.g., sm-repl, dash/explorer, etc.

there is no strong arguments that we should move it to separate pkg, only this 2:
1. our possibly moving `address` in separate lib in future 
1. current situation in `common/types`, where we have 25 different files aka 25 different types (there are total 33 files with tests) and some of this files contains lot of code - this look like place where we can add more pkg separation and make it more strict

if you think that it better move back to `common/types` - ok, i'll do. arguments for separate pkg not very strong

lrettig (changes requested):

countvonzero:
> > This is not a terrible idea - do we think other projects will have this dependency? (And would they want to use just an address module without relying on all of go-spacemesh?) E.g., sm-repl, dash/explorer, etc.
> 
> there is no strong arguments that we should move it to separate pkg, only this 2:
> 
> 1. our possibly moving `address` in separate lib in future

i think this is a future thing and should be separated from this discussion.

> 2. current situation in `common/types`, where we have 25 different files aka 25 different types (there are total 33 files with tests) and some of this files contains lot of code - this look like place where we can add more pkg separation and make it more strict

right. but changing only 1 makes it even more confusing. i'd support if you make a sweeping change later to change all of them. but the code inside each file (tho large) are mostly very simple code... i personally don't think it's worth the effort now.
 
> if you think that it better move back to `common/types` - ok, i'll do. arguments for separate pkg not very strong

that will be my preference. and worry about separate library when the use case arise.

common/types/address/address_test.go
abergasov: done
but actually it will just bech32 lib tests) all this cases will generate err on decode and it immediately returns

dshulyak:
bors try

common/types/address.go
```diff
@@ -1,84 +1,120 @@
...
+func DefaultTestAddressConfig() *Config {
+	conf = &Config{
+		NetworkHRP: "stest",
+	}
+	return conf
+}
+
 // Address represents the address of a spacemesh account with AddressLength length.
-type Address [AddressLength]byte
+type Address [AddressLength]byte // contains slice 8 bytes unsigned integers
```
dshulyak: > contains slice 8 bytes unsigned integers

what does it mean? where are the integers?

common/types/address.go
```diff
@@ -1,84 +1,120 @@
...
 
-// BytesToAddress returns Address with value b.
-// If b is larger than len(h), b will be cropped from the left.
-func BytesToAddress(b []byte) Address {
-	var a Address
-	a.SetBytes(b)
-	return a
+// FromString returns a new Address from a given string like `sm1abc...`.
+func FromString(src string) (Address, error) {
```
dshulyak: it needs to be more specific now, since the module was removed. maybe StringToAddress if there is already BytesToAddress

common/types/address.go
```diff
@@ -109,6 +136,17 @@ func (a *Address) DecodeScale(d *scale.Decoder) (int, error) {
...
 func GenerateAddress(publicKey []byte) Address {
 	var addr Address
-	addr.SetBytes(publicKey)
+	if len(publicKey) > len(addr) {
+		publicKey = publicKey[len(publicKey)-AddressLength:]
+	}
+	copy(addr[AddressLength-len(publicKey):], publicKey)
+	for i := 0; i < AddressReservedSpace; i++ {
+		addr[i] = 0
+	}
```
dshulyak, line 142: copy(addr[4:], publicKey[len(publicKey)-24:])

common/types/address.go
```diff
@@ -1,84 +1,120 @@
...
+
+	// AddressLength+1 cause ConvertBits append empty byte to the end of the slice.
+	if len(dataConverted) != AddressLength+1 {
+		return addr, fmt.Errorf("expected %d bytes, got %d: %w", AddressLength, len(data), ErrWrongAddressLength)
+	}
+
+	if conf.NetworkHRP != hrp {
+		return addr, fmt.Errorf("wrong network id: expected `%s`, got `%s`: %w", conf.NetworkHRP, hrp, ErrUnsupportedNetwork)
+	}
```
dshulyak, line 77: it needs a check that first 4 bytes are empty

config/presets/testnet.go
```diff
@@ -34,11 +36,11 @@ func testnet() config.Config {
...
-			"0x8868600d2835019249a0f756f754d5a5ab17ebf0": 100000000000000000,
-			"0x9f120e40e330affe6aa27d5bda49ac0960df7d6f": 100000000000000000,
-			"0xafed9a1c17ca7eaa7a6795dbc7bee1b1d992c7ba": 100000000000000000,
-			"0xb20c3a31f973231e01bf2d6b5c00a22cc13b5c63": 100000000000000000,
-			"0xff083c9a22e05d3459b03f3dbed61c7ad6e0a209": 100000000000000000,
+			"stest1rxklau52r6ygyxn2638e2vqu0qv98whcmftjvds0r3764": 100000000000000000,
+			"stest1jt38hng8nmgywqc8vg83gffgfsajkxn9wjwr39qka8kuj": 100000000000000000,
+			"stest15e933zkyfvsgkxj0qxpx9fequcuhkzf2n2lzg2spvwvlf": 100000000000000000,
+			"stest13zwpcc8z0z7xcnu3jm9fxkudvzk70e6d5zvnerqjhu96v": 100000000000000000,
+			"stest1ty8kqh4w58s9qyytp7lc6ddvc84wvsy4rcvjtas3w5rkm": 100000000000000000,
```
dshulyak: this are the same addresses just re-encoded or from different private keys?

common/types/address.go
abergasov: by default slice is uint8, value can be stored from 0-256 (base 2^8=256)
for bech32 need convert to 5-bit integers (2^5=32). it still uint8, but value up to 32
```
bech32.ConvertBits(addr[:], 8, 5, true) -> before encode to `sm1abc`
bech32.ConvertBits(addr[:], 5, 8, true) -> after encode from `sm1abc`
```

after this conversion slice length make bigger. so we keep as usually, uint8

common/types/address.go
abergasov: `BytesToAddress` will be dropped, when we merge api changes
I'll rename to `StringToAddress`

common/types/address.go
dshulyak: please remove this comment. it doesn't store uint64 (8 bytes unsigned inteegers) in any shape. what you provided above is a specific encoding of the bech32 format

config/presets/testnet.go
abergasov: yes, i make ` util.FromHex(str)` and create address from it
as we keep algorithm how convert bytes to address (get last 20 bytes from byte slice)  and we fill first 4 bytes with zero - we can just regenerate this addresses

abergasov:
bors try

dshulyak:
what about system tests?

abergasov:
> what about system tests?

it fails in same place `cluster.Reuse` with ctx timeout. Probably node fails to start in case invalid genessis|coinbase accounts

abergasov:
bors try

abergasov:
bors try

abergasov:
bors try

dshulyak (approved):

abergasov:
bors merge

abergasov:
bors merge

api/grpcserver/http_server.go
```diff
@@ -6,7 +6,7 @@ import (
 	"net/http"
 	"sync"
 
-	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	runtime2 "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
```
lrettig: Why rename this?

common/types/address.go
```diff
@@ -1,84 +1,121 @@
...
+	AddressReservedSpace = 4
 )
 
+var (
+	// ErrWrongAddressLength is returned when the length of the address is not correct.
+	ErrWrongAddressLength = errors.New("wrong address length")
+	// ErrUnsupportedNetwork is returned when a network is not supported.
+	ErrUnsupportedNetwork = errors.New("unsupported network")
+	// ErrDecodeBech32 is returned when an error occurs during decoding bech32.
+	ErrDecodeBech32 = errors.New("error decode to bech32")
```
lrettig:
```suggestion
	ErrDecodeBech32 = errors.New("error decoding bech32")
```

common/types/address.go
```diff
@@ -1,84 +1,121 @@
...
+		return addr, fmt.Errorf("%s: %w", ErrDecodeBech32, err)
+	}
 
-// BigToAddress returns Address with byte values of b.
-// If b is larger than len(h), b will be cropped from the left.
-func BigToAddress(b *big.Int) Address { return BytesToAddress(b.Bytes()) }
+	// for encoding bech32 uses slice of 5-bit unsigned integers. convert it back it 8-bit uints.
+	dataConverted, err := bech32.ConvertBits(data, 5, 8, true)
+	if err != nil {
+		return addr, fmt.Errorf("error convert bits to 8 bits: %w", err)
```
lrettig:
```suggestion
		return addr, fmt.Errorf("error converting bech32 bits: %w", err)
```

common/types/address.go
```diff
@@ -1,84 +1,121 @@
...
 func (a Address) String() string {
-	return a.Hex()
+	dataConverted, err := bech32.ConvertBits(a[:], 8, 5, true)
+	if err != nil {
+		log.Panic("error convert bits to 8 bits: ", err.Error())
+	}
+
+	result, err := bech32.Encode(conf.NetworkHRP, dataConverted)
+	if err != nil {
+		log.Panic("error encode to bech32: ", err.Error())
```
lrettig:
```suggestion
		log.Panic("error encoding to bech32: ", err.Error())
```

common/types/address.go
```diff
@@ -1,84 +1,121 @@
...
-	return "0x" + string(result)
+	return true
 }
 
 // String implements fmt.Stringer.
 func (a Address) String() string {
-	return a.Hex()
+	dataConverted, err := bech32.ConvertBits(a[:], 8, 5, true)
+	if err != nil {
+		log.Panic("error convert bits to 8 bits: ", err.Error())
```
lrettig:
```suggestion
		log.Panic("error converting bech32 bits: ", err.Error())
```

common/types/address_test.go
```diff
@@ -0,0 +1,149 @@
...
+		name string
+		src  string
+		err  error
+	}{
+		{
+			name: "correct",
+			src:  "stest1qqqqqqy0dd83jemjmfj3ghjndxm0ndh0z2rymaqyp0gu3",
+		},
+		{
+			name: "corect address, but no reserved space",
```
lrettig:
```suggestion
			name: "correct address, but no reserved space",
```

common/types/address_test.go
```diff
@@ -0,0 +1,149 @@
...
+			src:  "stest1fejq2x3d79ukpkw06t7h6lndjuwzxdnj59npghsg43mh4",
+			err:  types.ErrMissingReservedSpace,
+		},
+		{
+			name: "incorrect cahrset", // chars iob is not supported by bech32.
```
lrettig:
```suggestion
			name: "incorrect charset", // chars `iob` not supported by bech32.
```

common/types/address_test.go
```diff
@@ -0,0 +1,149 @@
...
+			src:  "stest1fejq2x3d79ukpkw06t7h6lniobuwzxdnj59nphsywyusj",
+			err:  types.ErrDecodeBech32,
+		},
+		{
+			name: "missing hrp",
+			src:  "1fejq2x3d79ukpkw06t7h6lniobuwzxdnj59nphsywyusj",
+			err:  types.ErrDecodeBech32,
+		},
+		{
+			name: "to big length",
```
lrettig:
```suggestion
			name: "too long",
```

common/types/address_test.go
```diff
@@ -0,0 +1,149 @@
...
+			src:  "stest1qw508d6e1qejxtdg4y5r3zarvax8wucu",
+			err:  types.ErrWrongAddressLength,
+		},
+		{
+			name: "to small length",
```
lrettig:
```suggestion
			name: "too short",
```

common/types/address_test.go
```diff
@@ -0,0 +1,149 @@
...
+			src:  "stest1qw504y5r3zarva2v6vda",
+			err:  types.ErrWrongAddressLength,
+		},
+		{
+			name: "wrong network",
```
lrettig:
```suggestion
			name: "wrong network (hrp)",
```

lrettig (changes requested):

abergasov:
bors try

lrettig (approved):

dshulyak:
i thought that it was merged. is there any reason why it is still open?

abergasov:
bors try

abergasov:
bors merge

Issue #3333, opened 2022-07-05
database: atomic apply and revert


countvonzero:
## Description

when a layer is applied or reverted, the goroutine spans across three components:

mesh -> conservative state -> vm

two issues:
- inconsistent data: each component updates database independently. should any component error out, the other components don't undo its update.
- database deadlock: each of them can potentially start a database transaction while calling the other that also starts its own database transaction.

i made a proposal to pass the `*sql.Tx` object in apply/revert goroutines but concerns were raised about making such move without more data like code instrumentation with metrics to track db saturation (e.g how many time we spend in waiting for a write or for a db handle) and some metrics to track request rates. 

fixing this issue https://github.com/spacemeshos/go-spacemesh/pull/3258 can also add to the confidence, but there's more work to be done there too.

@dshulyak @pigmej

dshulyak:
I still believe that it will be a huge mess if opened transactions will be passed across modules.

What i would do instead is to check why we need to update state in mesh and conservative cache after revert and try to eliminate those updates.

dshulyak:
After latest refactoring only results are not saved/reverted atomically with the rest of the state. I am not sure if thats possible (definitely not easily) due to optimistic filtering

Issue #3188, opened 2022-04-27
TestShortSkew is flaky


dshulyak:
https://github.com/spacemeshos/go-spacemesh/runs/6194047609?check_suite_focus=true

```
=== CONT  TestShortTimeskew
    assertion_compare.go:342: 
        	Error Trace:	timeskew_test.go:63
        	Error:      	"13" is not less than or equal to "11"
        	Test:       	TestShortTimeskew
        	Messages:   	[]
```

PR #3338, opened 2022-07-06
[Merged by Bors] - improve mesh and transaction data consistency

### LINKED ISSUE #3311: mesh: inconsistent data in db
reported by @brusherru 

node is stuck with the following log.
```
{"L":"INFO","T":"2022-06-30T17:50:00.413+0300","N":"eb074.sync         ","M":"processing synced layers","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"sync","in_state":"10745","last_synced":"10839","name":"sync"}
{"L":"INFO","T":"2022-06-30T17:50:00.413+0300","N":"eb074.mesh         ","M":"processing layer","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"mesh","layer_id":10746}
{"L":"INFO","T":"2022-06-30T17:50:00.413+0300","N":"eb074.trtl         ","M":"on layer terminated","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"trtl","layer_id":10746,"name":"trtl"}
{"L":"INFO","T":"2022-06-30T17:50:00.415+0300","N":"eb074.mesh         ","M":"tortoise results","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"mesh","layer_id":10746,"verified":10783,"name":"mesh"}
{"L":"INFO","T":"2022-06-30T17:50:00.415+0300","N":"eb074.mesh         ","M":"trying to set processed layer to an older layer","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"mesh","processed":10784,"layer_id":10746,"name":"mesh"}
{"L":"INFO","T":"2022-06-30T17:50:00.415+0300","N":"eb074.mesh         ","M":"pushing layers to state","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"mesh","from_layer":"10746","to_layer":"10784"}
{"L":"INFO","T":"2022-06-30T17:50:00.415+0300","N":"eb074.conState     ","M":"applying layer to conservative state","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"conState","layer_id":10746,"block_id":"091c0b752c"}
{"L":"ERROR","T":"2022-06-30T17:50:00.416+0300","N":"eb074.conState     ","M":"layer not applied in order","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"conState","expected":"10745","to_apply":"10746","name":"conState"}
{"L":"ERROR","T":"2022-06-30T17:50:00.416+0300","N":"eb074.mesh         ","M":"failed to apply transactions","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"mesh","layer_id":10746,"num_failed_txs":0,"errmsg":"layers not applied in order","name":"mesh"}
{"L":"ERROR","T":"2022-06-30T17:50:00.417+0300","N":"eb074.mesh         ","M":"failed to push layers to state","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"mesh","layer_id":10746,"errmsg":"failed to update state 10746: apply layer: layers not applied in order","name":"mesh"}
{"L":"WARN","T":"2022-06-30T17:50:00.417+0300","N":"eb074.sync         ","M":"mesh failed to process layer from sync","node_id":"eb074e3c8588fd378111ef54332be8a3d0fa76c073aab0eff5f426f9e73d2804","module":"sync","errmsg":"failed to update state 10746: apply layer: layers not applied in order","name":"sync"}
```

there are two components at play here:
- syncer: thinks the latest layer applied in state is `10745` (from table `mesh_status`). so it trigger `mesh.ProcessLayer(10746)`
- conservative cache: thinks the latest applied in state is `10744` (from table `layers`), so it rejects to apply layer for `10746`.

the code in mesh.go was not written in such a way that guarantees database integrity. specifically here
https://github.com/spacemeshos/go-spacemesh/blob/ff1a96c97412c0fdcf2bf1954dca6346019ae0d6/mesh/mesh.go#L629
```
if err := layers.SetApplied(msh.cdb, layerID, applied); err != nil {
		return fmt.Errorf("set applied for %v/%v: %w", layerID, applied, err)
}

if err := msh.setLatestLayerInState(layerID); err != nil {
		return err
}
```
it first updates the `layers` table, then updates the `mesh_status` table without using a db transaction to guarantee atomicity.

the situation is similar with the revert flow.
https://github.com/spacemeshos/go-spacemesh/blob/ff1a96c97412c0fdcf2bf1954dca6346019ae0d6/mesh/mesh.go#L320

my guess is that the code is interrupted when a revert took place. as `mesh_status` table had the higher layer as applied than the `layers` table. unfortunately, my theory cannot be confirmed as the earlier log isn't available.

in any case, mesh.go needs to be reworked such that db updates are atomic.

countvonzero:
## Motivation

Closes #3311

## Changes

- remove mesh_status table
- get last layer from max(ballots.layer)
- get applied layer from max(layers.id) where applied_block not null
- get processed layer from max(layers.id) where processed is true
- save hash/aggregated hash in the same sql stmt
- in mesh.go, update applied layer with hashes atomically
- change conservative cache to use database directly (instead of an interface)
- in cache.go, update transactions within the same account atomically

database consistency is still not guaranteed. see https://github.com/spacemeshos/go-spacemesh/issues/3333

## Test Plan

unittest, systest

## DevOps Notes


sql/ballots/ballots.go
```diff
@@ -173,3 +173,17 @@ func GetRefBallot(db sql.Executor, epochID types.EpochID, pubkey []byte) (ballot
 	}
 	return ballotID, nil
 }
+
+// LatestLayer gets the highest layer with ballots.
+func LatestLayer(db sql.Executor) (types.LayerID, error) {
+	var lid types.LayerID
+	if _, err := db.Exec("select max(layer) from ballots;",
```
evt, line 180: do we need a semicolon here?

sql/migrations/0001_initial.sql
```diff
@@ -28,17 +28,13 @@ CREATE TABLE layers
...
     weak_coin       SMALL INT,
     hare_output     VARCHAR,
+    processed       SMALL INT,
     applied_block   VARCHAR,
     state_hash      CHAR(32),
     hash            CHAR(32),
     aggregated_hash CHAR(32)
 ) WITHOUT ROWID;
-
-CREATE TABLE mesh_status
```
evt, line 37: are we supposed to have `drop table` migration if we remove a table?

sql/ballots/ballots.go
countvonzero: it will work without. but keeping consistent with the rest of the code

sql/migrations/0001_initial.sql
countvonzero: we only have the initial.sql file. we will only need to add "drop table" stmt when we start to have diff versions.

countvonzero:
bors try

evt (approved):

txs/txs_builder.go
```diff
@@ -168,7 +169,7 @@ func orderTXs(logger log.Log, pmd *proposalMetadata, realState stateFunc, blockS
 	// we do not want to access database here, as all transactions are already in memory.
 	dbLessCache := &cache{
 		logger:    logger,
-		tp:        &nopTP{},
+		db:        sql.InMemory(),
```
dshulyak: i didn't notice this before but this seems somewhat strange, do you mean that db is not used in the dbLessCache?

countvonzero: correct. the DB-less cache is used for constructing a block. we fill the cache with all the proposal txs, reuse the logic there to organize by accounts, and spit them out flat for a block.

in the regular cache for mempool, db's main purpose is to get more/better txs from db when a earlier nonce is applied (and may relax the balance constraint). for block construction, we already have all the txs we need. @dshulyak

countvonzero: to reduce confusion, i've removed db from cache all together, and pass in db as needed.

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7255978229?check_suite_focus=true)

https://github.com/spacemeshos/go-spacemesh/issues/3343

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7256875968?check_suite_focus=true)

https://github.com/spacemeshos/go-spacemesh/issues/3188 is back

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #3349, opened 2022-07-12
[Merged by Bors] - txs: change key to hash32

dshulyak:
small refactoring

- replace key from hex string to 32 bytes, this is a potential source of bugs as 5 bytes clearly doesn't provide same collision resistance as 32 bytes hash. and super inefficient, for every hash there is a 64 bytes allocation plus the cost of hex encoding 
- don't pass function as a variable, it makes sense only for closures


txs/conservative_state.go
```diff
@@ -103,7 +103,7 @@ func (cs *ConservativeState) SelectBlockTXs(lid types.LayerID, proposals []*type
 		myHash = types.EmptyLayerHash
 	}
 
-	md, err := checkStateConsensus(cs.logger, cs.cfg, lid, proposals, myHash, cs.GetMeshTransaction)
+	md, err := checkStateConsensus(cs.logger, cs.cfg, lid, proposals, myHash, cs)
```
abergasov, line 107: if he passing full cs object - that we can simplify func signature `checkStateConsensus(lid, proposals, myHash, cs)`
or just make `func (cs *ConservativeState) checkStateConsensus` as it use only in this case

countvonzero (approved):

txs/conservative_state.go
countvonzero: > don't pass function as a variable, it makes sense only for closures

i don't understand why it does not make sense.

the original idea was that static functions were more testable. passing a function ptr is also easier for testing.
one does not have to create an object just for its method.

in any case, i didn't end up testing these private static functions... so.... either way.

dshulyak: maybe i am wrong. but if function is passed as a variable i can't easily find all implementations that satisfy this func signatures without looking up all code paths. there is a support for that if interfaces or regular objects are used.

for closures or in functional approach, where functions are chained or combined in some non-trivial way it makes more sense, but it is still not friendly in golang

dshulyak:
bors merge

dshulyak:
bors merge

PR #3351, opened 2022-07-12
txs: error on transactions that failed conservative cache validation

dshulyak:
- transaction that doesn't pass conservative cache validation won't be stored by the node
- proposal that includes tx with invalid nonce won't be stored by the node


countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

txs/conservative_state.go
```diff
@@ -133,13 +133,15 @@ func (cs *ConservativeState) Validation(raw types.RawTx) system.ValidationReques
 // AddToCache adds the provided transaction to the conservative cache.
 func (cs *ConservativeState) AddToCache(tx *types.Transaction) error {
 	received := time.Now()
+	if err := cs.cache.Add(cs.db, tx, received, nil); err != nil {
```
dshulyak, line 136: i think changed order makes tests flaky

countvonzero: i patched your change in my PR and ran bors try. so far systests passed 2 out of 2 times that systests get to run. the other bors try attempts failed at other flaky unittests or dockerpush failed or systests infra failed...

countvonzero: if it passes again, i am going to merge it and this PR will not be necessary anymore. @dshulyak 
will update if that happens or the same failure occurs.

dshulyak: unfortunately regardless if sanity tests are passing transactions handling is still broken. some nodes are simply getting stuck. i don't know what is the root cause though

countvonzero: noted

Issue #2507, opened 2021-07-05
mesh hash resolution with peers


countvonzero:
# design discussion
https://github.com/spacemeshos/SMIPS/issues/51 
https://github.com/spacemeshos/SMIPS/issues/60

# implementation plan
## fetch
add the following to fetch interface
```
type MeshHashRequest struct {
    From, To     types.LayerID
    Delta, Steps uint32
}

type MeshHashes struct {
    Layers []types.LayerID
    Hashes []types.Hash32
}

type EpochData struct {
    Beacon types.Beacon
    Weight uint64
    AtxIDs []types.ATXID
}

type layerFetcher interface {
    Peers() []p2p.Peer
    PeerEpochInfo(context.Context, p2p.Peer, types.EpochID) (*fetch.EpochData, error)
    PeerMeshHashes(context.Context, p2p.Peer, *fetch.MeshHashRequest) (*fetch.MeshHashes, error)
}
```

## syncer
syncer periodically check peers’ agreement on the mesh hash (aka aggregated hash) along with its layer opinions.

given a list of opinions for layer `N` from peers, syncer first finds those that differ from the node’s mesh hash at layer `N-1`. for those peers. pseudo code below:

```
getMeshAgreement(lid types.LayerID, opinions []*fetch.LayerOpinion) (*fetch.LayerOpinion, error) {
    maxIgnoredWeight uint64
    candidate LayerOpinion
    for opn in opinions {
        if opn.meshHash == nodeMeshHash {
            // update last agreement with peer
            forkFinder.UpdateAgreement(opn.Peer(), lid.Sub(1))
        else {
            epochData = fetch.PeerEpochInfo()
            missing = set_diff(peerAtxIDs, nodeAtxIDs)  // in peerAtxIDs but not in nodeAtxIDs
            err = fetchATX(missing)
            if err {
                // if the node cannot download the ATXs claimed by this peer, 
                // it does not trust this peer's mesh
                continue
            }
            ignoredWeight = total_weight(missing)
            if ignoredWeight > maxIgnoredWeight {
                maxIgnoredWeight = ignoredWeight
                candidate = opn
            }
        }
    }
    
    threshold := (nodeEpochWeight + maxIgnoredWeight).Fraction(s.conf.TortoiseGlobalThreshold)
    if nodeEpochWeight > threshold {
        // the node should still be confident of its mesh
        return nil, nil
    }

    // find the divergent layer and adopt the peer's mesh from there
[...]

PR #3352, opened 2022-07-13
[Merged by Bors] - p2p gossipsub: turn on peer scoring

### LINKED ISSUE #3247: p2p: reduce network protocols names to a smaller size

countvonzero:
## Motivation

part of #2507
Closes #3247

## Changes

- turn on gossipsub peer scoring. params are copied from FIL (https://github.com/filecoin-project/lotus)
- shorten protocol names and moved them to p2p/pubsub pkg
- gossip validation return `ValidationReject` on malformed msgs for atx/poetproof/block proposals 

## Test Plan

unit test, systests

### DIFF
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -23,8 +23,8 @@ import (
 const (
 	// RoundsPerIteration is the number of rounds per iteration in the hare protocol.
 	RoundsPerIteration = 4
-	// ProtoName is the protocol indicator for hare gossip messages.
-	ProtoName = "HARE_PROTOCOL/1"
+	// HareProtocol is the protocol indicator for hare gossip messages.
+
 )
 
 type role byte
@@ -496,7 +496,7 @@ func (proc *consensusProcess) sendMessage(ctx context.Context, msg *Msg) bool {
 	)
 	logger := proc.WithContext(ctx)
 
-	if err := proc.publisher.Publish(ctx, ProtoName, msg.Bytes()); err != nil {
+	if err := proc.publisher.Publish(ctx, pubsub.HareProtocol, msg.Bytes()); err != nil {
 		logger.With().Error("could not broadcast round message", log.Err(err))
 		return false
 	}
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -23,8 +23,6 @@ import (
 const (
 	// RoundsPerIteration is the number of rounds per iteration in the hare protocol.
 	RoundsPerIteration = 4
-	// HareProtocol is the protocol indicator for hare gossip messages.
-
 )
 
 type role byte
--- a/p2p/peerexchange/discovery.go
+++ b/p2p/peerexchange/discovery.go
@@ -49,7 +49,7 @@ type Discovery struct {
 }
 
 // New creates a Discovery instance.
-func New(logger log.Log, h host.Host, config Config) (*Discovery, bool, error) {
+func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	d := &Discovery{
 		cfg:    config,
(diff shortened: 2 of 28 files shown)


p2p/peerexchange/discovery.go
```diff
@@ -49,7 +49,7 @@ type Discovery struct {
 }
 
 // New creates a Discovery instance.
-func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
+func New(logger log.Log, h host.Host, config Config) (*Discovery, bool, error) {
```
dshulyak: this should not be a part of New signature

p2p/pubsub/pubsub.go
```diff
@@ -14,6 +15,61 @@ import (
...
+	pubsub.GossipSubD = 8
+	pubsub.GossipSubDscore = 6
+	pubsub.GossipSubDout = 3
+	pubsub.GossipSubDlo = 6
+	pubsub.GossipSubDhi = 12
+	pubsub.GossipSubDlazy = 12
+	pubsub.GossipSubDirectConnectInitialDelay = 30 * time.Second
+	pubsub.GossipSubIWantFollowupTime = 5 * time.Second
+	pubsub.GossipSubHistoryLength = 10
+	pubsub.GossipSubGossipFactor = 0.1
```
dshulyak, line 31: is anything above changed from the default value? if so what was the reason?

p2p/pubsub/pubsub_test.go
```diff
@@ -23,9 +24,10 @@ func TestGossip(t *testing.T) {
 	count := n * n
 	received := make(chan []byte, count)
 
+	fmt.Printf("num in mesh %d\n", len(mesh.Hosts()))
```
dshulyak: rm

hare/algorithm.go
```diff
@@ -23,8 +23,8 @@ import (
 const (
 	// RoundsPerIteration is the number of rounds per iteration in the hare protocol.
 	RoundsPerIteration = 4
-	// ProtoName is the protocol indicator for hare gossip messages.
-	ProtoName = "HARE_PROTOCOL/1"
+	// HareProtocol is the protocol indicator for hare gossip messages.
```
dshulyak: rm ?

p2p/pubsub/pubsub.go
countvonzero: all these are different from the default. i copied this from filecoin. my rationale is that since i also copied their params for scoring, might as well use their default values.

https://github.com/filecoin-project/lotus/blob/d9311b200ca87d50243cd8deae525f62e78f26b4/node/modules/lp2p/pubsub.go#L28

p2p/peerexchange/discovery.go
countvonzero: rm'ed

p2p/pubsub/pubsub_test.go
countvonzero: done

p2p/upgrade.go
```diff
@@ -63,6 +64,23 @@ type Host struct {
 	bootstrap *bootstrap.Bootstrap
 }
 
+func isBootnode(h host.Host, bootnodes []string) (bool, error) {
+	best, err := peerexchange.BestHostAddress(h)
```
dshulyak: there is an h.ID, best is needed only for discovery

p2p/upgrade.go
```diff
@@ -75,35 +93,36 @@ func Upgrade(h host.Host, opts ...Opt) (*Host, error) {
 		opt(fh)
 	}
 	cfg := fh.cfg
-	var err error
-	fh.PubSub, err = pubsub.New(fh.ctx, fh.logger, h, pubsub.Config{
+	bootnode, err := isBootnode(h, cfg.Bootnodes)
```
dshulyak, line 92: this probably should be a part of the user provided config. technically bootnodes list should not include its own node, e.g. it requires preparing full libp2p entry before starting a bootnode. i don't do that in system tests, and *i think* we don't do that for public testnets as well

dshulyak (approved):
see comment about IsBootnode. i think current approach won't have an effect in practice

p2p/upgrade.go
countvonzero: > it requires preparing full libp2p entry before starting a bootnode

curious. what does "preparing full libp2p entry" mean? how does that compare to a regular node?

 > this probably should be a part of the user provided config
 
 i am trying to limit the scope just to gossipsub in this PR. but maybe it's worth creating an issue for the more general approach if bootnodes require a different setup? (i would do it myself if i know the difference :P)

p2p/upgrade.go
countvonzero: done

p2p/upgrade.go
dshulyak: the deployment follows this process for bootnode:
1. launch a node
2. node will generate p2p private key
3. request p2p public key from api
4. create libp2p entry from public key + ip/port
5. put that entry into config for other nodes to use (they may be bootnodes or regular nodes)

so as you see bootnode will not have its own entry in the config

countvonzero: thanks for the explanation.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

Issue #3289, opened 2022-06-22
txs: prune blocks and txs that have malformed transactions


dshulyak:
Transactions that are referenced in the block are saved to the database without validating them, when it is time to apply a layer transaction will be validated, and if it is not valid - block is not syntactically valid and should be pruned from the database.

PR #3370, opened 2022-07-19
[Merged by Bors] - decoupling vm from consensus

### LINKED ISSUE #3272: vm: transactions that were included into the block but can't be executed
Current vm behavior is to skip transaction in two cases:
- out of order nonce
- not enough funds to cover for max gas

There is an argument that we shouldn't skip them, but consume since they are taking space in the mesh. Pending on research

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3272
supersedes: https://github.com/spacemeshos/go-spacemesh/issues/3289

- malformed and unverifiable transactions are ineffective and will be returned back to conservative cache
- transactions that weren't executed due to block gas limit are also returned as ineffective to conservative cache
- if transaction was verified when accepted by the node it won't be verified again by the vm when executed. conservative cache should save headers for verified transactions
- conservative cache doesn't parse/verify transactions before sending them to vm. it will use vm output to restore its own consistency after applying layer


dshulyak:
@countvonzero please take a look at this change

genvm/vm.go
```diff
@@ -246,52 +254,92 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
...
+			raw:     txs[i].RawTx(),
+			decoder: decoder,
+			limit:   limit,
+		}
+
+		header, err := req.Parse()
 		if err != nil {
-			v.logger.With().Warning("skipping transaction. failed to parse", log.Err(err))
-			skipped = append(skipped, tx.ID)
+			v.logger.With().Warning("innefective transaction. failed to parse",
```
countvonzero: typo: ineffective. here and for the rest of this file

genvm/vm.go
```diff
@@ -246,52 +254,92 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
...
+				log.Object("header", header),
+				log.Object("account", &ctx.Account),
+			)
+			ineffective = append(ineffective, types.Transaction{RawTx: tx.RawTx()})
+			invalidTxCount.Inc()
+			continue
+		}
+		if limit < ctx.Header.MaxGas {
+			v.logger.With().Warning("innefective transaction. out of block gas",
+				log.Uint64("block gas limit", v.cfg.GasLimit),
```
countvonzero, line 304: i think logging the actual limit is more meaningful. we can easily see the configured gas limit.

genvm/vm.go
```diff
@@ -246,52 +254,92 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
...
+			ineffective = append(ineffective, types.Transaction{RawTx: tx.RawTx()})
 			invalidTxCount.Inc()
 			continue
 		}
+
+		t2 := time.Now()
+		v.logger.With().Debug("applying transaction",
+			log.Object("header", header),
+			log.Object("account", &ctx.Account),
+		)
```
countvonzero: move t2 and this logging after the nonce check to be consistent.

mesh/mesh.go
```diff
@@ -375,7 +375,7 @@ func (msh *Mesh) ProcessLayer(ctx context.Context, layerID types.LayerID) error
 
 	// pass the layer to tortoise for processing
 	newVerified := msh.trtl.HandleIncomingLayer(ctx, layerID)
-	logger.With().Info("tortoise results", log.FieldNamed("verified", newVerified))
```
countvonzero, line 378: can we keep this log as Info? or what is the other log stmt i can check for tortoise progress?

genvm/vm.go
```diff
@@ -327,31 +372,37 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.RawTx)
 			return nil, nil, 0, fmt.Errorf("%w: %s", core.ErrInternal, err.Error())
```
countvonzero, line 373: it doesn't look like `ctx.Apply` can ever return an error. `StagedCache.Update` is just changing in-memory data.

genvm/vm.go
```diff
@@ -327,31 +372,37 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.RawTx)
...
 // After Parse is executed - conservative cache may do validation and skip Verify
 // if transaction can't be executed.
 type Request struct {
-	vm *VM
+	vm    *VM
+	cache *core.StagedCache
 
 	raw     types.RawTx
 	ctx     *core.Context
+	args    scale.Encodable
```
countvonzero: add comment about ctx/args are only set after Parse.

genvm/vm.go
```diff
@@ -327,31 +372,37 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.RawTx)
...
 	decoder *scale.Decoder
+	limit   uint64
 }
 
 // Parse header from the raw transaction.
 func (r *Request) Parse() (*core.Header, error) {
 	start := time.Now()
-	header, ctx, _, err := parse(r.vm.logger, r.vm.registry, core.NewStagedCache(r.vm.db), r.raw.ID, r.decoder)
+	header, ctx, args, err := parse(r.vm.logger, r.vm.registry, r.cache, r.raw.ID, r.decoder)
```
countvonzero: it doesn't look like `r.raw.ID` is used in `parse()`

txs/conservative_state.go
```diff
@@ -154,78 +154,33 @@ func (cs *ConservativeState) RevertState(revertTo types.LayerID) (types.Hash32,
...
-		return nil, errs[0]
+	logger.With().Debug("applying layer to cache",
+		log.Int("num_txs_skipped", len(ineffective)),
+		log.Int("num_txs_applied", len(results)),
+	)
+	if _, errs := cs.cache.ApplyLayer(ctx, cs.db, block.LayerIndex, block.ID(),
+		results, ineffective); len(errs) > 0 {
+		return errs[0]
 	}
-	cacheApplyDuration.Observe(float64(time.Since(t0)))
```
countvonzero, line 185: please keep

genvm/vm.go
```diff
@@ -246,52 +254,92 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
...
+				log.Object("header", header),
+				log.Object("account", &ctx.Account),
+			)
+			ineffective = append(ineffective, types.Transaction{RawTx: tx.RawTx()})
+			invalidTxCount.Inc()
+			continue
+		}
+
+		// NOTE this part is executed only for transactions that weren't verified
+		// when accepted into mempool
```
countvonzero: this comment is not true. mempool never accepts tx that's not parsed/verified. do you mean when `saved into database`?

txs/conservative_state.go
```diff
@@ -291,6 +246,27 @@ func (cs *ConservativeState) GetMeshTransactions(ids []types.TransactionID) ([]*
...
+		}
+		if mtx.State == types.APPLIED {
+			continue
+		}
+		if mtx.TxHeader == nil {
+			rawTxCount.WithLabelValues(rawFromDB).Inc()
+			txs = append(txs, mtx.RawTx)
+		} else {
+			txs = append(txs, types.VerifiedTx(mtx.RawTx))
+		}
```
countvonzero, line 264: i don't get the need for this if/else block, and for `VerifiedTx`/`ExecutableTx`

it seems you want to save a Verify() call in VM if a tx is already verified by mempool. 
why not just check for missing Header in VM and call Verify() if it's nil?

genvm/vm.go
```diff
@@ -246,52 +254,92 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
...
+		rd.Reset(tx.RawTx().Raw)
+		req := &Request{
+			vm:      v,
+			cache:   ss,
+			raw:     txs[i].RawTx(),
+			decoder: decoder,
+			limit:   limit,
+		}
+
+		header, err := req.Parse()
```
countvonzero, line 280: > if transaction was verified when accepted by the node it won't be verified again by the vm when executed. conservative cache should save headers for verified transactions

why are we parsing the tx again here? purely for getting req.ctx and req.args?
are we confident that the two parses yield the same header?

genvm/vm.go
```diff
@@ -246,52 +254,92 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
...
+				log.Object("account", &ctx.Account),
+			)
+			ineffective = append(ineffective, types.Transaction{RawTx: tx.RawTx()})
+			invalidTxCount.Inc()
+			continue
+		}
+
+		// NOTE this part is executed only for transactions that weren't verified
+		// when accepted into mempool
+		if !tx.Verified() && !req.Verify() {
```
countvonzero, line 316: i think it's more straightforward to simply check `txs[i].TxHeader == nil`, instead of defining `ExecutableTx` interface and two impls of them.

mesh/mesh.go
dshulyak: there are logs in the tortoise

genvm/vm.go
dshulyak: it can potentially flush data on disk

genvm/vm.go
dshulyak: ok i will rephrase it, because i am calling txs module mempool

genvm/vm.go
dshulyak: parsing because arguments can't be parsed in another way. yes they are the same

genvm/vm.go
dshulyak: i don't want vm module to have access to already parsed tx header, someone may want to reuse it and it will cause confusion. this interface on the other hand is optional, in a sense that you may use it to improve perf

txs/conservative_state.go
dshulyak: sorry bad merge

txs/conservative_state.go
dshulyak: ok, i removed that

countvonzero (approved):

genvm/vm.go
countvonzero: hmmmm.... the current impl always return nil.

dshulyak: yes, i know. i meant that it can be changed without changing caller code

dshulyak:
bors try

dshulyak:
bors merge

PR #3377, opened 2022-07-24
[Merged by Bors] - Migrate to scale for blocks

### LINKED ISSUE #3271: codec: migrate to scale
Use https://github.com/spacemeshos/go-scale in the codec module.

before working on this issue implement https://github.com/spacemeshos/go-scale/issues/2

Using scale for majority of the objects should be trivial:
- add go generate with scalegen
- for every collection define sensible limit using tag name defined above
- add a fuzz tests and a golden tests (you should be able to find utilities for both in the lib above)

If some objects uses collections with pointers in them - they should be replaced with collections without pointers.

There is also some mess with types from poet/post, but i don't recall details.

### Modules to enable scale codec for:
- [x] transactions - https://github.com/spacemeshos/go-spacemesh/pull/3380
- [x] blocks - https://github.com/spacemeshos/go-spacemesh/pull/3377
- [x] ballots - https://github.com/spacemeshos/go-spacemesh/pull/3383
- [x] proposals - https://github.com/spacemeshos/go-spacemesh/pull/3383
- [x] fetch - https://github.com/spacemeshos/go-spacemesh/pull/3466
- [x] SQL proposal TX IDs - https://github.com/spacemeshos/go-spacemesh/pull/3448
- [x] poet proofs - https://github.com/spacemeshos/go-spacemesh/pull/3395
- [x] sync - https://github.com/spacemeshos/go-spacemesh/pull/3395
- [x] activation types - https://github.com/spacemeshos/go-spacemesh/pull/3413
- [x] hare - https://github.com/spacemeshos/go-spacemesh/pull/3478 
- [x] p2p server - https://github.com/spacemeshos/go-spacemesh/pull/3476
- [x] disable XDR codec + final fixes: https://github.com/spacemeshos/go-spacemesh/pull/3496, https://github.com/spacemeshos/go-spacemesh/pull/3482

evt:
## Motivation

Partly Closes #3271 (for blocks)

## Changes
Applied scale codec to blocks

## Test Plan


dshulyak (approved):
i am not sure why do we need hand written tests. there is fuzz test that generates random objects, encode/decodes them. it should provide exhaustive coverage

evt:
> i am not sure why do we need hand written tests. there is fuzz test that generates random objects, encode/decodes them. it should provide exhaustive coverage

does it run fuzzy tests in CI/CD? they never end when I run them locally.

dshulyak:
> does it run fuzzy tests in CI/CD? they never end when I run them locally.

it doesn't run fuzz by default. we can add another job that runs selected number of fuzz tests. i think it runs only one test instance if fuzz is executed with go test without -fuzz flag.

as for go test with -fuzz flag it needs a timeout, otherwise it will run forever

evt:
bors try

evt:
bors merge

PR #3189, opened 2022-04-27
[Merged by Bors] - layerfetcher: syntactically invalid blocks are expected from malicious peers

evt:
## Motivation
Closes #2697 

## Changes
Skip returning error for invalid blocks

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #3382, opened 2022-07-26
[Merged by Bors] - sync: ignore malicious ballots during sync

### LINKED ISSUE #2697: syncer fails sync too aggressively on invalid blocks
## Description

when debugging late-nodes system test failure on develop branch, one of the issues that came up was that syncer fails sync as soon as block fails syntactic validity check (caused by nodes disagreeing on different tortoise beacon value).

this behavior is too aggressive, as syntactically invalid blocks are expected from malicious peers.

## Expected Behavior

when syncer 
- gets database.ErrNotFound from peer, it should retry to get blocks from other peers until it exhausted all peers
- receives an syntactically invalid block, it should skip it instead of failing sync

countvonzero:
## Motivation

Closes #2697 

## Changes

the block part is done in https://github.com/spacemeshos/go-spacemesh/pull/3189
same should be done for ballots.

### DIFF
--- a/fetch/layers.go
+++ b/fetch/layers.go
@@ -218,13 +218,27 @@ func (l *Logic) fetchLayerData(ctx context.Context, logger log.Log, layerID type
 
 	logger.With().Debug("fetching new ballots", log.Int("to_fetch", len(ballotsToFetch)))
 	if err := l.GetBallots(ctx, ballotsToFetch); err != nil {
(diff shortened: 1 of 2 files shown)


countvonzero:
@evt please review. thank you.

evt (approved):

countvonzero:
bors merge

PR #3390, opened 2022-07-29
[Merged by Bors] - systest: multiple poet servers

### LINKED ISSUE #3387: systest: allow multiple poet servers
## Description

longevity tests only has 1 poet server. with 50 nodes, TestScheduleBasic will inevitably fail within 5-6 hrs of running. 
this makes reproducing bugs time-consuming.

having multiple poet servers should allow longevity test to run longer

countvonzero:
## Motivation

Closes #3387

## Changes

- allow multiple poet servers
- allow multiple tries on submitting txs

## Test Plan

manual

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -8,6 +8,8 @@ test_pod_name ?= systest-$(version_info)-$(shell date +'%s')
 keep ?= false
 clusters ?= 1
 size ?= 10
+psize ?= 1
+retries ?= 0
 level ?= debug
 bootstrap ?= 5m
 storage ?= standard=1Gi
@@ -31,7 +33,7 @@ run:
 	--restart=Never \
 	--labels="testid=$(test_id)" \
     --image-pull-policy=IfNotPresent -- \
-    tests -test.v -test.timeout=0 -test.run=$(test_name) -namespace=$(namespace) -clusters=$(clusters) -storage=$(storage) -size=$(size) -image=$(smesher_image) -level=$(level) -node-selector=$(node_selector) -bootstrap=$(bootstrap) -keep=$(keep) -testid=$(test_id) -labels=$(label)
+    tests -test.v -test.timeout=0 -test.run=$(test_name) -namespace=$(namespace) -clusters=$(clusters) -storage=$(storage) -size=$(size) -poet-size=$(psize) -retries=$(retries) -image=$(smesher_image) -level=$(level) -node-selector=$(node_selector) -bootstrap=$(bootstrap) -keep=$(keep) -testid=$(test_id) -labels=$(label)
 	-@kubectl wait --timeout=10s --for=condition=ready pod/$(test_pod_name)
 	@kubectl logs $(test_pod_name) -f
 
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -8,8 +8,7 @@ test_pod_name ?= systest-$(version_info)-$(shell date +'%s')
 keep ?= false
 clusters ?= 1
 size ?= 10
-psize ?= 1
-retries ?= 0
+poet_size ?= 1
 level ?= debug
 bootstrap ?= 5m
 storage ?= standard=1Gi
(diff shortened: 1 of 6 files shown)


systest/tests/common.go
```diff
@@ -292,15 +292,20 @@ func (c *txClient) nonce(ctx context.Context) (uint64, error) {
 	return getNonce(ctx, c.node, c.account.Address)
 }
 
-func (c *txClient) submit(ctx context.Context, tx []byte) (*txRequest, error) {
-	txid, err := submitTransacition(ctx, tx, c.node)
-	if err != nil {
-		return nil, fmt.Errorf("submit to node %s: %w", c.node.Name, err)
+func (c *txClient) submit(ctx context.Context, tx []byte, tries int) (*txRequest, error) {
```
dshulyak: tries should not be a required parameter to submit function. it either should be an option or removed from submit completely, and set when txClient is initialized

systest/Makefile
```diff
@@ -8,6 +8,8 @@ test_pod_name ?= systest-$(version_info)-$(shell date +'%s')
 keep ?= false
 clusters ?= 1
 size ?= 10
+psize ?= 1
+retries ?= 0
```
dshulyak: why it make sense to set retries in env variables? currently env variables control only environment and resources, they weren't meant to parametrize test logic

would it work to provide some reasonable number of default retries and remove this variable completely?

systest/cluster/nodes.go
```diff
@@ -185,6 +184,8 @@ func deployNodes(ctx *testcontext.Context, name string, from, to int, flags []De
 		i := i
 		eg.Go(func() error {
 			setname := fmt.Sprintf("%s-%d", name, i)
+			idx := i % ctx.PoetSize
+			flags = append(flags, PoetEndpoint(poetEndpoint(idx)))
```
dshulyak: this will extend flags on every iteration, and potentially will even modify original flags array

systest/Makefile
```diff
@@ -8,6 +8,8 @@ test_pod_name ?= systest-$(version_info)-$(shell date +'%s')
 keep ?= false
 clusters ?= 1
 size ?= 10
+psize ?= 1
```
dshulyak: would be better if this variable was consistent with flag

systest/testcontext/context.go
```diff
@@ -239,6 +243,8 @@ func New(t *testing.T, opts ...Opt) *Context {
 		TestID:            *testid,
 		Keep:              *keep,
 		ClusterSize:       *clusterSize,
+		PoetSize:          *poetSize,
```
dshulyak, line 258: needs to be saved in labels like keep and cluster size

systest/cluster/nodes.go
countvonzero: copied it out. thanks for catching this

systest/testcontext/context.go
countvonzero: done

systest/tests/common.go
countvonzero: used a default value for number of attempts. 

i feel eventually may need to wrap retries around the gRPC and rest/k8s calls. it's frustrating to fail the test immediately when those fail.

systest/Makefile
countvonzero: hehe... i really have no idea what the intended design for env variables was in the systest. 
now i know. done.

systest/cluster/nodes.go
dshulyak: this is not safe. if original flags slice has a capacity it will be extended with new flag value, and the next one will overwrite it

```
a := make([]string, 1, 2)
a[0] = "1"
b := a
b = append(b, "2")
c := a
c = append(c, "3")
fmt.Println(b, c)
// [1, 3], [1,3]
```

countvonzero: sorry. bad with slices. fixed.

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors try

countvonzero:
bors merge

dshulyak:
does it break something of there is an environment issue?

countvonzero:
> does it break something of there is an environment issue?

there's def something wrong with my PR. 
i've used this PR to run longevity tests many times (prior to the last few edits) :( 
will debug.

systest/cluster/nodes.go
```diff
@@ -183,8 +182,12 @@ func deployNodes(ctx *testcontext.Context, name string, from, to int, flags []De
 	)
 	for i := from; i < to; i++ {
 		i := i
+		finalFlags := make([]DeploymentFlag, len(flags)+1)
+		copy(finalFlags, flags)
 		eg.Go(func() error {
 			setname := fmt.Sprintf("%s-%d", name, i)
+			idx := i % ctx.PoetSize
+			finalFlags[len(finalFlags)-1] = PoetEndpoint(poetEndpoint(idx))
 			if err := deployNode(ctx, setname, labels, flags); err != nil {
```
dshulyak: this should use finalFlags :)

countvonzero: yes. i just realized that too. soooooooo dumb.....thanks for helping

countvonzero:
bors merge

PR #3399, opened 2022-08-02
[Merged by Bors] - atxs: changes related to poet ticks

dshulyak:
the weight of the atx computed from tick count multiplied by the number of units.
tick height of the atx computed from base tick height plus tick count.
base tick height is a tick height of the positioning atx.
tick count is a number of leaves in the poet proof divided by tick size.
tick size is a protocol parameter, meant to kept the size of the weight in uint64 bounds.

positioning atx is an atx with the highest tick height from the last epoch.


common/types/activation.go
```diff
@@ -135,10 +142,40 @@ func (atxh *ActivationTxHeader) SetID(id *ATXID) {
 	atxh.id = id
 }
 
-// GetWeight returns the ATX's weight = numUnits * ticks.
+// GetWeight of the atx.
+// Will panic if read before Verify is called.
```
countvonzero: this comment, and below. the code below doesn't panic. do you mean to add the panic stmt?

what do you think about adding an Initialize() to be consistent with ballot/proposal, set ID and verify there? 
it may mean you have to query the position atx early and shift stuff around, but i think it'll make the code easier to follow.

activation/handler.go
```diff
@@ -405,14 +385,16 @@ func (h *Handler) handleAtxData(ctx context.Context, data []byte) error {
 	if err != nil {
 		return fmt.Errorf("received syntactically invalid atx %v: %v", atx.ShortString(), err)
 	}
-
+	if !atx.Verified() {
```
countvonzero: this check here is out of place for me. if it's not verified, `SyntacticallyValidateAtx` should return an error, right?

activation/handler.go
```diff
@@ -288,9 +293,12 @@ func (h *Handler) StoreAtx(ctx context.Context, ech types.EpochID, atx *types.Ac
 	h.Lock()
 	defer h.Unlock()
 
-	// todo: maybe cleanup handler if failed by using defer (#1921)
-	if err := h.storeATXInDB(atx); err != nil {
-		return fmt.Errorf("store ATX in db: %w", err)
+	if err := atxs.Add(h.cdb, atx, time.Now()); err != nil {
+		if errors.Is(err, sql.ErrObjectExists) {
+			// exists - how should we handle this?
```
countvonzero, line 298: i think we should ignore this.

when the node is syncing from scratch, we are fetching ATXID one by one in that epoch. depending on the order of hashes, we can be fetching an ATX in one goroutine while processing the same ATX in the other.

in the gossip case, we should have all the prev/pos ATXs locally. in only scenario i can see is network rejoin where the above sync-from-scratch scenario could happen.

common/types/activation.go
```diff
@@ -135,10 +142,40 @@ func (atxh *ActivationTxHeader) SetID(id *ATXID) {
...
 func (atxh *ActivationTxHeader) GetWeight() uint64 {
-	// TODO: Limit the number of bits this can occupy
-	return uint64(atxh.NumUnits) * (atxh.EndTick - atxh.StartTick)
+	return uint64(atxh.NumUnits) * (atxh.tickCount)
+}
+
+// BaseTickHeight is a tick height of the positional atx.
+// Will panic if read before Verify is called.
+func (atxh *ActivationTxHeader) BaseTickHeight() uint64 {
```
countvonzero, line 151: i don't understand why this BaseTickHeight() and TickCount() is needed. once we verify an ATX's tick height, why can't we just record the sum in the db instead of recording them separately?

countvonzero (commented):
two things:
- the verify flow is not intuitive. i'd prefer to move things around and make sure once the ID is set, the tick height is also recorded.
- is it possible to combine the API for base tick height and tick count into just tick height?

dshulyak:
> the verify flow is not intuitive. i'd prefer to move things around and make sure once the ID is set, the tick height is also recorded.

id needs to be computed earlier, in order to check data in the database. ticks can be recorded only after validation
i will check if i can shift that around in a way that allows to bypass expensive validation, but likely it won't be possible

> is it possible to combine the API for base tick height and tick count into just tick height?

base tick height is used for proposals
tick count is used for weight computation

activation/handler.go
dshulyak: it is probably not my comment

common/types/activation.go
dshulyak: removed the comment.

there are two things about Initialize

1. Verify needs to be called long after id is computed. this is too expensive to validate atx every time 
2. this "caching" in initialize is a terrible pattern and i don't think that it makes sense to waste time spreading it around codebase

i wanted to introduce `type VerifiedAtx` that has this fields set and ready to be used after validating atx or after loading atx from database. this way you can rely on compiler to give you warnings that you are using wrong type. however it requires some refactoring and i would rather avoid doing it in this change

activation/handler.go
dshulyak: removed

common/types/activation.go
dshulyak: see reply below. it is not possible to merge them, we are using both of them for different use cases

activation/handler.go
dshulyak: yes i factored away the original method and copied this part here. 

do you want me to remove this comment?

dshulyak:
opened an issue https://github.com/spacemeshos/go-spacemesh/issues/3434

countvonzero (approved):

dshulyak:
bors try

dshulyak:
bors merge

Issue #3212, opened 2022-05-11
TestSmeshing/Proposals flake


dshulyak:
https://github.com/spacemeshos/go-spacemesh/pull/3203#issuecomment-1122654636

moshababo:
@dshulyak still relevant?

countvonzero:
this is happening because the network produce N proposals, but the hare output <N proposals. causing the discrepency.

https://github.com/spacemeshos/go-spacemesh/runs/8076053876?check_suite_focus=true 

```
 status:Included]" should have 31 item(s), but has 30
        	Test:       	TestSmeshing/Proposals
        	Messages:   	client=0 layer=15
```
logs

```
{"L":"INFO","T":"2022-08-29T13:35:34.010-0400","N":"fe963.hare         ","M":"starting hare","node_id":"fe963b48cadef98289da6430fac134345e86073735aa7c9bed50c1efd5d4b07f","module":"hare","protocol":"hr1_tickloop","sessionId":"1ebf06e9-6c53-41b4-afb5-d66c163feb5b","layer_id":15,"num_proposals":30,"name":"hare"}
{"L":"INFO","T":"2022-08-29T13:35:42.097-0400","N":"fe963.hare         ","M":"hare terminated with success","node_id":"fe963b48cadef98289da6430fac134345e86073735aa7c9bed50c1efd5d4b07f","module":"hare","protocol":"hr1_outputloop","sessionId":"5140ad2d-d157-4883-952f-b999931ae1ca","layer_id":15,"num_proposals":30,"name":"hare"}
```
but it did receive 31 proposals

![Screenshot from 2022-08-29 13-03-46](https://user-images.githubusercontent.com/30611210/187268237-ad497141-4311-4444-a2ec-6e1db0892198.png)

PR #3403, opened 2022-08-03
[Merged by Bors] - allow nonce gap in mempool and exeuction

### LINKED ISSUE #3389: conservative cache: apply layer should not return cache-only error
## Description

when conservative cache apply an layer, the main purpose is to
- update db w.r.t transaction state
- reset the cache based on vm after the layer is applied

any cache-only error should be just warning and should not interrupt the above two tasks.

### LINKED ISSUE #3401: vm: allow nonce gap for execution
## Description

see https://github.com/spacemeshos/pm/issues/150

genvm/vm.go: change inequality (validity predicate) to compare >= NextNonce()

### LINKED ISSUE #3402: conservative cache: allow nonce gap in mempool
## Description

tx model (`txs/cache.go`, `txs/conservative_state.go` refactoring)
- logic in `applyLayer` cannot loop over incremental nonces, this needs to be refactored; it cannot calculate `numApplied` with simple subtraction. change inequality (validity predicate) to check that tx nonce is `>= NextNonce()`
- `accountCache.moreInDB`: update logic to allow for nonce gaps

countvonzero:
## Motivation

Closes #3401 
Closes #3402
Closes #3389 

## Changes

- in vm relaxes the nonce comparison from `tx.Nonce == acct.NextNonce()` to `tx.Nonce >= acct.NextNonce()`
- in conservative cache
  - use container.List to keep mempool transactions
  - allow nonce gap

## Test Plan

unit test, systest

### DIFF
--- a/genvm/vm.go
+++ b/genvm/vm.go
@@ -323,8 +323,6 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.Transa
 			continue
 		}
 
-		// TODO to be changed after nonces are defined
-		// https://github.com/spacemeshos/go-spacemesh/issues/3273
 		if ctx.Account.NextNonce() > ctx.Header.Nonce.Counter {
 			v.logger.With().Warning("ineffective transaction. failed nonce check",
 				log.Object("header", header),
--- a/genvm/vm.go
+++ b/genvm/vm.go
@@ -325,7 +325,7 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.Transa
 
 		// TODO to be changed after nonces are defined
 		// https://github.com/spacemeshos/go-spacemesh/issues/3273
-		if ctx.Account.NextNonce() != ctx.Header.Nonce.Counter {
+		if ctx.Account.NextNonce() > ctx.Header.Nonce.Counter {
 			v.logger.With().Warning("ineffective transaction. failed nonce check",
 				log.Object("header", header),
 				log.Object("account", &ctx.Account),
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,7 +206,7 @@ jobs:
 
     - name: Run tests
       timeout-minutes: 60
-      run: cd systest && make run test_name=. size=50 bootstrap=10m level=info clusters=2 node_selector=cloud.google.com/gke-nodepool=gha label=sanity
+      run: cd systest && make run test_name=. size=50 bootstrap=10m level=info clusters=2 node_selector=cloud.google.com/gke-nodepool=gha label=sanity storage=premium-rwo=10Gi
 
     - name: Systemtest result
       run: |
--- a/genvm/vm_test.go
+++ b/genvm/vm_test.go
@@ -624,23 +624,23 @@ func TestWorkflow(t *testing.T) {
 						spendWallet{0, 11, 100}.withNonce(core.Nonce{Counter: 2}),
 						spendWallet{0, 10, 100}.withNonce(core.Nonce{Counter: 1}),
 					},
-					ineffective: []int{1},
+					ineffective: []int{2},
 					headers: map[int]struct{}{
-						1: {},
+						2: {},
 					},
 					expected: map[int]change{
 						0: spawned{
 							template: wallet.TemplateAddress,
 							change:   spent{amount: 100 + defaultGasPrice*(wallet.TotalGasSpawn+wallet.TotalGasSpend)},
 						},
-						10: earned{amount: 100},
-						11: same{},
+						10: same{},
+						11: earned{amount: 100},
 					},
 				},
 				{
 					txs: []testTx{
-						spendWallet{0, 10, 100}.withNonce(core.Nonce{Counter: 2}),
-						spendWallet{0, 12, 100}.withNonce(core.Nonce{Counter: 3}),
+						spendWallet{0, 10, 100}.withNonce(core.Nonce{Counter: 3}),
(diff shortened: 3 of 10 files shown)


genvm/vm.go
```diff
@@ -325,7 +325,7 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.Transa
 
 		// TODO to be changed after nonces are defined
```
dshulyak: remove todo?

dshulyak (approved):

dshulyak:
i realized that it introduces very unpleasant side effect. if multiple transactions are sent at the same time and received out of order - some of them may get skipped even if proposal is not full, simply because of the order they are received.

is it possible to pack transactions in order if block is not full, but allow out of order execution in vm?

countvonzero:
@dshulyak 
> i realized that it introduces very unpleasant side effect. if multiple transactions are sent at the same time and received out of order - some of them may get skipped even if proposal is not full, simply because of the order they are received.

if such transactions arrive only after a node's proposal is created, yes, they will be skipped for this node's proposal. but they are likely to be included in other nodes' proposals.
 
> is it possible to pack transactions in order if block is not full, but allow out of order execution in vm?

i am assuming you mean `proposal`. proposals are indeed packed in order still, just with gaps. 

for example, for account A with next nonce being 1.
node1 received (A, 3), (A, 6), (A, 7) before packing proposal P1, and (A, 4) afterwards, P1 will have [(A, 3), (A, 6), (A, 7)] 
node2 received (A, 3), (A, 6), (A, 4) before packing proposal P2, and (A, 7) afterwards, P2 will have [(A, 3), (A, 4), (A, 6)] 

if hare output includes both P1 and P2, the block will include [(A, 3), (A, 4), (A, 6), (A, 7)]

but then of course there is a chance that (A, 4) was delayed and neither P1 or P2 include it, then the final block will have [(A, 3), (A, 6), (A, 7)]. (A, 4) will become invalid once this block is executed.

is the latter case what you are concerned about?

dshulyak:
> is the latter case what you are concerned about?

yes
i thought maybe it is possible to keep packing in order, but allow out of order execution. so that with honest nodes - proposals with out of order txs will come up only if they are full

countvonzero:
> > is the latter case what you are concerned about?
> 
> yes i thought maybe it is possible to keep packing in order, but allow out of order execution. so that with honest nodes - proposals with out of order txs will come up only if they are full

concluding this after the discussion during dev sync, i think we should reconsider after we collect some empirical data. there are other scenarios where an earlier nonce will be render invalid when a  layer is executed. for example, 
- when hare output empty for layer N, since there are no overlap btwn txs in proposal N and N+1, almost all txs in layer N will be rendered invalid after layer N+1 executed. 
- on a much smaller scale, when some lower nonces txs are only packed in one proposal (barely make the cut broadcasting to one miner), and that proposal didn't make it to hare consensus output.

btw, i don't understand why being full or not make a difference...

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

dshulyak:
> btw, i don't understand why being full or not make a difference...

i don't know as well. but it might be a sign that there is a mistake in the implementation

@lrettig @noamnelke can you clarify when it makes a difference?

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7677902358?check_suite_focus=true)

https://github.com/spacemeshos/go-spacemesh/issues/3426

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7679046982?check_suite_focus=true)

https://github.com/spacemeshos/go-spacemesh/issues/3212

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7685394195?check_suite_focus=true)

this is again #3212
also verified with logs that 36 proposals were created. it's most likely being processed late.

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7685915744?check_suite_focus=true)

@dshulyak 
it seems this PR failed more frequently with https://github.com/spacemeshos/go-spacemesh/issues/3212
i up'ed the batch size to 10 in TestSmeshing/Transactions.
the network did produce 37 proposals, but when the test polls, it only got 36.
do you have any suggestion to fix this flakiness? thanks

dshulyak:
@countvonzero can you try with ssd disk type? https://github.com/spacemeshos/go-spacemesh/blob/develop/.github/workflows/ci.yml#L209

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7696158757?check_suite_focus=true)

canceled this manually to rm a dbg stmt from PR

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #3405, opened 2022-08-03
[Merged by Bors] - proposals: add tick height from proposals to blocks

dshulyak:
- add highest known base tick height from activation that are referenced in the proposals to the block. this will be used later in the tortoise to prevent adversarial votes using ballot with lower tick heights
- when choosing valid block from multiple valid blocks use the lowest tick height. the idea is that the lower the tick height - the more time honest voters had to vote


common/types/block.go
```diff
@@ -156,9 +158,14 @@ func ToBlockIDs(blocks []*Block) []BlockID {
 	return ids
 }
 
-// SortBlocks sort blocks by their IDs.
+// SortBlocks sort blocks tick height, if height is equal by lexicographic order.
 func SortBlocks(blks []*Block) []*Block {
-	sort.Slice(blks, func(i, j int) bool { return blks[i].ID().Compare(blks[j].ID()) })
+	sort.Slice(blks, func(i, j int) bool {
+		if blks[i].TickHeight < blks[j].TickHeight {
```
countvonzero: i think it's worth adding a test here.

countvonzero (approved):

common/types/block.go
dshulyak: i would rather make it private and move to mesh. it doesn't need to be exported and in common package if it used once

countvonzero: yes. make sense

dshulyak:
bors merge

Issue #3134, opened 2022-02-20
logging: support log collection from external smeshers


dshulyak:
We are missing important information that are required to troubleshoot concurrency-related issues. For instance in devnet220 in layer 8345 there are 2 proposals that are created with old base ballots and without votes for recent layers. It indicates that the hare didn't complete on those smeshers, while the rest of the cluster completed hare successfully.

There are two options:
1. Integrate https://github.com/elastic/beats/tree/main/libbeat into spacemesh binary
2. Integrate https://github.com/elastic/beats/tree/main/filebeat as a daemon, and manage that daemon from smapp

I prefer to go with 1st option

dshulyak:
With loki there is simpler version of the [client](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/client/client.go) . it needs to be integrated as a WriteSyncer to zap.Core . The client itself batches log entries in the background and instrumented with relevant metrics, so that we will know if logger will use a lot of traffic.

PR #3428, opened 2022-08-05
[Merged by Bors] - logging: reduce log verbosity for hare and others

countvonzero:
## Motivation

part of #3134

## Changes

using systests TestAddNodes as measurement of the log file for a node.

- before 4.3M
- with hare reduced 1.6M (1st commit) (-62.79%)
- with the rest of the change 1.2M (-9.3%)

### DIFF
--- a/common/types/ballot.go
+++ b/common/types/ballot.go
@@ -227,16 +227,10 @@ func (b *Ballot) MarshalLogObject(encoder log.ObjectEncoder) error {
 	encoder.AddUint32("epoch_id", uint32(b.LayerIndex.GetEpoch()))
 	encoder.AddString("smesher", b.SmesherID().String())
 	encoder.AddString("base_ballot", b.Votes.Base.String())
-	encoder.AddInt("supports", len(b.Votes.Support))
-	encoder.AddInt("againsts", len(b.Votes.Against))
-	encoder.AddInt("abstains", len(b.Votes.Abstain))
+	encoder.AddInt("support", len(b.Votes.Support))
+	encoder.AddInt("against", len(b.Votes.Against))
+	encoder.AddInt("abstain", len(b.Votes.Abstain))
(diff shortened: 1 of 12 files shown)


common/types/ballot.go
```diff
@@ -227,16 +227,10 @@ func (b *Ballot) MarshalLogObject(encoder log.ObjectEncoder) error {
...
 	encoder.AddString("smesher", b.SmesherID().String())
 	encoder.AddString("base_ballot", b.Votes.Base.String())
-	encoder.AddInt("supports", len(b.Votes.Support))
-	encoder.AddInt("againsts", len(b.Votes.Against))
-	encoder.AddInt("abstains", len(b.Votes.Abstain))
+	encoder.AddInt("support", len(b.Votes.Support))
+	encoder.AddInt("against", len(b.Votes.Against))
+	encoder.AddInt("abstain", len(b.Votes.Abstain))
 	encoder.AddString("atx_id", b.AtxID.String())
-	encoder.AddArray("eligibilities", log.ArrayMarshalerFunc(func(encoder log.ArrayEncoder) error {
```
dshulyak, line 234: why? it should be once per received ballot. if there are more inlines they should be reduced to id

dshulyak (approved):

common/types/ballot.go
countvonzero: i can revert this, but it contribute so little, if at all, to debugging. 
```
type VotingEligibilityProof struct {
	// the counter value used to generate this eligibility proof. if the value of J is 3, this is the smesher's
	// eligibility proof of the 3rd ballot/proposal in the epoch.
	J uint32
	// the VRF signature of some epoch specific data and J. one can derive a Ballot's layerID from this signature.
	Sig []byte
}
```

i am not sure what you mean by `reduced to id` tho.

countvonzero:
bors merge

PR #3429, opened 2022-08-05
[Merged by Bors] - vm: multisig template

### LINKED ISSUE #3268: vm: multisig wallet
See [singlesig wallet](https://github.com/spacemeshos/go-spacemesh/blob/develop/genvm/templates/wallet/wallet.go) for reference.

Multisig wallet should be spawned with 2 values:
- array with public keys
- number of required signatures

Spend arguments are the same as for singlesig.

Multisig signature should be composed from required number of signatures where every signature part has:
- field that references public key used to spawn the wallet
- ed25519.Sign result

One option is to use:

```go
type Signature struct {
      // 2 is the number of required signatures
     Parts [2]Part
}

type Part struct {
      // Reference to the public key in the spawned array
      Ref uint8
      Sig [64]byte
}
``` 

Verify method should decode signatures, find a reference public key and do the verification using oasis BatchVerifier.

Note that same template should be generic enough to be usable with several configurations (e.g 3/5/7/9 keys and different number of required keys). Each configuration will be registered at its own address.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3268
depends on: https://github.com/spacemeshos/go-scale/pull/16

vm:

changed method signature for Verify from 
`Verify(ctx Context, raw []byte) bool`
 to
`Verify(ctx Context, raw []byte, dec *scale.Decoder) bool`
it simplifies decoding and enables variable sized fields in the signature.

multisig:

there are 3 multisig (K/N) variants 1/N, 2/N, 3/N
N is a number of keys that own multisig, N is limited by 10 and provided when account is spawned.
K is a number of signatures required to pass threshold, template can't have less keys than K.

multisig relies on canonical ed25519 signatures. K owners needs to sign a hash of the payload.
see type definitions in https://github.com/spacemeshos/go-spacemesh/blob/20f5a38aab38daff830b08f611ed5aef2e1bd53d/genvm/templates/multisig/types.go
public keys for spawn are encoded as a variable sized collection of 32byte arrays.
signature is encoded as a constant sized collection with tuples [(ref uint8, sig [64]byte)]. collection has to be sorted in the ascending order.


genvm/sdk/multisig/tx.go
```diff
@@ -0,0 +1,120 @@
...
+
+	payload := multisig.SpawnPayload{}
+	payload.GasPrice = options.GasPrice
+	payload.Arguments.PublicKeys = make([]core.PublicKey, len(pubs))
+	for i := range pubs {
+		copy(payload.Arguments.PublicKeys[i][:], pubs[i])
+	}
+	principal := core.ComputePrincipal(template, &payload.Arguments)
+
+	tx := encode(&zero, &principal, &zero, &template, &payload)
```
countvonzero: the named variable here `zero` really doesn't make the encoding sequence more readable.
maybe define one transaction type (0) and two method types (0 and 1).

genvm/templates/multisig/handler.go
```diff
@@ -0,0 +1,152 @@
...
+
+const (
+	methodSpawn = 0
+	methodSpend = 1
+)
+
+func init() {
+	TemplateAddress1[len(TemplateAddress1)-1] = 2
+	TemplateAddress2[len(TemplateAddress2)-1] = 3
+	TemplateAddress3[len(TemplateAddress3)-1] = 4
```
countvonzero, line 41: these template addresses should really be defined in the same place as with `TemplateAddress`.
seem error prone to scatter them in different packages, and duplicate addresses are not checked.

perhaps worth adding a sanity check test too.

genvm/templates/multisig/handler.go
```diff
@@ -0,0 +1,152 @@
...
+		header.MaxGas = h.totalGasSpend
+	default:
+		return header, args, fmt.Errorf("%w: unknown method %d", core.ErrMalformed, method)
+	}
+	return header, args, nil
+}
+
+// Init wallet.
+func (h *handler) Init(method uint8, args any, state []byte) (core.Template, error) {
+	if method == 0 {
```
countvonzero: :s/0/methodSpawn

genvm/templates/multisig/handler.go
```diff
@@ -0,0 +1,152 @@
...
+		return &MultiSig{
+			PublicKeys: args.(*SpawnArguments).PublicKeys,
+			k:          h.k,
+		}, nil
+	}
+	decoder := scale.NewDecoder(bytes.NewReader(state))
+	ms := MultiSig{k: h.k}
+	if _, err := ms.DecodeScale(decoder); err != nil {
```
countvonzero, line 118: i cannot find usage of `handler.Init()` in tests, so i don't know what would be in the passed-in state.
what if `state`, after decoded, overwrote the k value in ms?

genvm/vm_test.go
```diff
@@ -41,16 +45,78 @@ func newTester(tb testing.TB) *tester {
...
+
+func (a *multisigAccount) getAddress() core.Address {
+	return a.address
+}
+
+func (a *multisigAccount) spend(to core.Address, amount uint64, nonce core.Nonce) []byte {
+	var agg *sdkmultisig.Aggregator
+	for i := 0; i < a.k; i++ {
+		part := sdkmultisig.Spend(uint8(i), a.pks[i], a.address, to, amount, nonce)
+		if agg == nil {
```
countvonzero: code here is a bit confusing....
how about do `agg := sdkmultisig.Spend(uint8(0), a.pks[0], a.address, to, amount, nonce)`
and iterate from 1 below.

genvm/templates/multisig/types.go
```diff
@@ -0,0 +1,39 @@
...
+type Signatures []Part
+
+// Part contains a reference to public key and signature from private key counterpart.
+type Part struct {
+	Ref uint8
+	Sig core.Signature
+}
+
+// SpendArguments ...
+type SpendArguments struct {
```
countvonzero: SpendArguments and SpendPayload are the same as single sig.
do you think there are downsides to use the same structs?

genvm/templates/multisig/multisig.go
```diff
@@ -0,0 +1,59 @@
...
+		return false
+	}
+	hash := core.Hash(raw[:len(raw)-n])
+	batch := ed25519.NewBatchVerifierWithCapacity(int(ms.k))
+	last := uint8(0)
+	for i, part := range sig {
+		if part.Ref >= uint8(len(ms.PublicKeys)) {
+			return false
+		}
+		if i != 0 && part.Ref <= last {
```
countvonzero, line 46: do you mean the signatures have to be in increasing order of Ref? why? is it a restriction on ed25519?

genvm/templates/multisig/multisig.go
```diff
@@ -0,0 +1,59 @@
...
+		return args.(*SpendArguments).Amount, nil
+	default:
+		return 0, fmt.Errorf("%w: unknown method %d", core.ErrMalformed, method)
+	}
+}
+
+// Verify that transaction is signed has k valid signatures.
+func (ms *MultiSig) Verify(ctx *core.Context, raw []byte, dec *scale.Decoder) bool {
+	sig := make(Signatures, ms.k)
+	n, err := scale.DecodeStructArray(dec, sig)
```
countvonzero, line 35: i don't fully understand this.
is it true that at this point of the code, `dec` contains the remaining unparsed data?

countvonzero (commented):
main question i have is about the design. why not implement single sig acct as 1-1 and we will just have one type of wallet?

countvonzero (approved):

genvm/templates/multisig/handler.go
dshulyak: k is a private field

genvm/templates/multisig/multisig.go
dshulyak: yes, by the time we need to decode signatures every other field was already decoded

genvm/templates/multisig/multisig.go
dshulyak: this is the most efficient way to guarantee that they are unique. not related to ed25519

genvm/templates/multisig/types.go
dshulyak: right, will reuse

genvm/vm_test.go
dshulyak: sure will change

genvm/templates/multisig/handler.go
dshulyak: i test for uniqueness in registry. i don't see any reason to set them in a single place, as long as uniqueness is enforced

dshulyak:
> main question i have is about the design. why not implement single sig acct as 1-1 and we will just have one type of wallet?

i don't know yet if we will drop singlesig. will do what research will decide
singlesig transactions are slightly shorter, due to the absence of var sized fields and removed `ref` field from signature

dshulyak:
bors merge

dshulyak:
bors merge

PR #3430, opened 2022-08-09
Replace []TransactionID with *TransactionIDs for proposals to enable scale codec

evt:


dshulyak:
but we don't need to replace it everywhere. scale should generate EncodeStructSlice/DecodeStructSlice for []TransactionID. this custom type should be used only on demand (e.g in database)

evt:
> but we don't need to replace it everywhere. scale should generate EncodeStructSlice/DecodeStructSlice for []TransactionID. this custom type should be used only on demand (e.g in database)

done - https://github.com/spacemeshos/go-spacemesh/pull/3432. Closing this one.

PR #3432, opened 2022-08-09
Replace []TransactionID with *TransactionIDs for SQL proposals to enable scale codec

evt:


sql/proposals/proposals.go
```diff
@@ -4,11 +4,31 @@ import (
...
 
+	"github.com/spacemeshos/go-scale"
+
 	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/sql"
 )
 
+// TransactionIDs is a list of transaction IDs.
+type TransactionIDs []types.TransactionID
```
countvonzero: shouldn't these belong in common/types?

common/types/testutil.go
```diff
@@ -148,9 +148,11 @@ func GenLayerProposal(layerID LayerID, txs []TransactionID) *Proposal {
 					},
 				},
 			},
-			TxIDs: txs,
 		},
 	}
+	if len(txs) > 0 {
+		p.InnerProposal.TxIDs = txs
```
countvonzero, line 154: if you need to change it here, why not in miner/proposal_builder.go too?

evt: I had a hare test failing due to `TxIDs: ([]types.TransactionID) {}` coming from this func and decoded by scale codec into `TxIDs: ([]types.TransactionID) <nil>` as there is no way for now to distinguish nil and empty slices in the codec. @dshulyak do we need the same change in other places (where codec is probably not applied as long as tests pass)?

sql/proposals/proposals.go
evt: good question. This type is specifically for sql/proposals with the only place where we encode/decode `[]types.TransactionID` and need this special type with proper `EncodeScale/DecodeScale` on it. So not sure. @dshulyak what do you think?

common/types/testutil.go
countvonzero: in some of the systests, we don't have any transactions in proposals. (TestAddNodes for example). you can verify whether that works, as proposals will be gossiped with no transactions inside.

sql/proposals/proposals.go
countvonzero: one would assume that there are similar issues with ballots's votes where a list of BlockIDs are encoded, sometimes empty. can we have consistent solutions with that too?

evt: @countvonzero I can't find where we encode a list of BlockIDs as a slice. We have it as a list of structs but then it's not an issue and doesn't need this workaround.

common/types/testutil.go
evt: do we add/get to/from SQL database there with sql/proposals add/get? If not then it should not be a problem. I'll run systests anyway.

sql/proposals/proposals.go
dshulyak: i don't understand why is it in sql module either. i also not aware where we need to encode BlockIDs separately

lets use scale.Encode/DecodeStructSlice to avoid such discussions

sql/proposals/proposals.go
```diff
@@ -129,9 +149,17 @@ func GetBlob(db sql.Executor, id []byte) (proposal []byte, err error) {
 	return proposal, err
 }
 
+func toTransactionIDs(txIDs []types.TransactionID) TransactionIDs {
+	if txIDs == nil {
+		return nil
+	}
+	return TransactionIDs(txIDs)
+}
```
dshulyak: whats wrong with TransactionIDs(nil) ?

common/types/testutil.go
dshulyak: > do we add/get to/from SQL database there with sql/proposals add/get?

i don't understand this question

sql/proposals/proposals.go
dshulyak: i would introduce generic codec.EncodeSlice[T scale.Encodeble](values []T) / codec.DecodeSlice[T scale.Decodable](buf []byte) ([]T, error) that will hide buffer and codec initialization in the codec module

countvonzero: > i also not aware where we need to encode BlockIDs separately

@evt @dshulyak i am referring to here 
https://github.com/spacemeshos/go-spacemesh/blob/3a4cb1bc931eb6ca34bebb62f31beaa014e006a5/common/types/ballot_scale.go#L114

votes.support and votes.Against are both []types.BlockID, which is similar to []Transaction.ID. you used `scale.EncodeStructSlice` there. why not here also?

common/types/testutil.go
countvonzero: @evt in systest like TestAddNodes, we simply don't send transactions to the network, so mempool is empty => proposals don't contain any transactions

sql/proposals/proposals.go
evt: @countvonzero this is generated by `scalegen` tool and perfectly works with a slice being a field in a struct. But not yet for a standalone slice.

countvonzero: @evt i don't understand. those []TransactionID is not standalone. they also live in the struct InnerProposal.

countvonzero: i guess you are saying you need them to be standalone for encoding in this sql function. ok. i got it.

evt: yeah, this is the only place where we run `codec.Encode` for a slice

evt: replaced  `TransactionIDs` with `EncodeSlice/DecodeSlice` as suggested by @dshulyak

sql/proposals/proposals.go
evt: removed it as per https://github.com/spacemeshos/go-spacemesh/pull/3432#discussion_r944148370

evt:
This one was automatically closed as base branch merged. Opened new PR here - https://github.com/spacemeshos/go-spacemesh/pull/3448

PR #3439, opened 2022-08-12
[Merged by Bors] - sync: advance the network when it stalls

### LINKED ISSUE #3371: sync: make sure data is synced even when hare output empty layer
## Description

when a node went offline for a brief period of time (and didn't receive ballot gossip), and hare terminated with empty layer, ballots will not be downloaded from peers.

in this case, syncer should consider empty hare output and still try to sync data in this case.

### LINKED ISSUE #3381: node: continue to advance despite poet server failures
### background
while doing longevity testing, the following scenario happened and the network simply halted.
the one PoET server became unstable in epoch 22 and requests to build NIPost all failed. 
this means miners cannot build ATXs to target epoch 23,

when epoch 24 starts, several components starts to fail due to cascade of multiple events
- miners cannot participate in beacon protocol (no ATX in epoch 23)
- miners cannot build proposals as it cannot calculate its own eligibility (no ATX in epoch 23)
- nodes fell out of sync when the latest known layer is too far behind current layer
- hare cannot run consensus protocol (nodes go out of sync, no ATX in epoch 23)
- tortoise cannot verify layers as no proposal/ballots are produced

basically the network halted.

and even if PoET servers recover, and ATXs building resumes, the code as is, cannot recover because syncer 
doesn't detect layers with absolutely zero content (no ballot and blocks).

this means that
- we need to have HighAvailability for PoET servers
- syncer needs to be able to make progress when hare doesn't (currently syncing data from peers is gated at processed layer) and when the network stopped producing ballots/blocks for many many layers
- tortoise needs to be able to produce votes when ballots are unavailable for many many layers

- [x] https://github.com/spacemeshos/go-spacemesh/issues/3440 syncer: move the network forward despite layers of empty content
- [ ] https://github.com/spacemeshos/go-spacemesh/issues/3388 tortoise: continue to verify layers despite long period of emptiness
- [ ] https://github.com/spacemeshos/go-spacemesh/issues/3441 activation: resumes building atxs after poet servers recover

### LINKED ISSUE #3440: syncer: move the network forward despite layers of empty content
## Description

part of https://github.com/spacemeshos/go-spacemesh/issues/3381

syncer needs to become a true backup for hare. this means syncer needs to move the network forward in the following scenario:
- when the node is synced but misses out on gossiped proposals due to unstable network connection
- when hare fails
- when the network stopped producing ballots/blocks for many many layers (for example poet server failures)

currently the following syncer behavior prevents the network to advance when there is no data in the network.
- syncing data from peers is gated at processed layer. and when hare fails/didn't run, processed layer cannot advance
- syncer does not sync every layer from peer. so missed gossiped messages are forever missed.
- when the network is stalled, syncer prematurely gets stuck for lack of beacon.

countvonzero:
## Motivation

Closes #3440

## Changes

- as a true backup mechanism, syncer should sync every layer from peers even when there is hare output
- remove fetching restrictions on layers higher than processed layer so that the network can make progress when there are no proposals/ballots/blocks in the network
- move the beacon check from syncer to mesh so that processed layer can advance
- removed some unused code

this is tested by 
- bringing up a cluster in GCP
- delete poet pod
- observe the network still advance with processed layer and apply empty layer when there are no data in the network. only tortoise verified layer and vm applied layer are stuck.

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -457,12 +457,6 @@ func (app *App) initServices(ctx context.Context,
 		}),
 		txs.WithLogger(app.addLogger(ConStateLogger, lg)))
 
-	var verifier blockValidityVerifier
-	msh, err := mesh.NewMesh(cdb, &verifier, app.conState, app.addLogger(MeshLogger, lg))
-	if err != nil {
-		return fmt.Errorf("failed to create mesh: %w", err)
-	}
-
 	genesisAccts := app.Config.Genesis.ToAccounts()
 	if len(genesisAccts) > 0 {
 		exists, err := state.AccountExists(genesisAccts[0].Address)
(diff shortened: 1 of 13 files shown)


countvonzero:
bors try

countvonzero:
@evt can you help review this PR? thanks

evt (approved):

countvonzero:
bors merge

Issue #3441, opened 2022-08-13
activation: resumes building atxs after poet servers recover 


countvonzero:
## Description

part of https://github.com/spacemeshos/go-spacemesh/issues/3381

the network needs to resume building ATXs after network recover from poet server failures.

countvonzero:
after further testing, when the network was able to move forward by this PR https://github.com/spacemeshos/go-spacemesh/pull/3439, it turned out that the node cannot resume building ATX.

### problem with positioning ATX
say we failed to build ATXs from epoch 5 -7, and resume building again, we will fetch the positioning ATX from epoch 4, which set the challenge's PubLayerId to somewhere in epoch 5. but epoch 5 had already expired so atx builder gets into an infinite loop. what should we do in this case?

slack convo. https://spacemesh.slack.com/archives/C89QJJY3G/p1660361378432829
still under discussion. will update GH when a solution emerges.

dshulyak:
i believe it works as expected now

PR #3442, opened 2022-08-13
[Merged by Bors] - longevity test: add step to delete/redeploy poet servers

countvonzero:
## Motivation

part of  #3441

## Changes

- systests: discover poet servers
- systests: add step to delete and redeploy poet servers
- activation: allow actual update of poet proof service and add logging

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -153,21 +153,22 @@ func NewBuilder(conf Config, nodeID types.NodeID, signer signer, cdb *datastore.
 	syncer syncer, log log.Log, opts ...BuilderOption,
 ) *Builder {
 	b := &Builder{
-		parentCtx:         context.Background(),
-		signer:            signer,
-		nodeID:            nodeID,
-		coinbaseAccount:   conf.CoinbaseAccount,
-		goldenATXID:       conf.GoldenATXID,
-		layersPerEpoch:    conf.LayersPerEpoch,
-		cdb:               cdb,
-		atxHandler:        hdlr,
-		publisher:         publisher,
-		nipostBuilder:     nipostBuilder,
-		postSetupProvider: postSetupProvider,
-		layerClock:        layerClock,
-		syncer:            syncer,
-		log:               log,
-		poetRetryInterval: defaultPoetRetryInterval,
+		parentCtx:             context.Background(),
+		signer:                signer,
+		nodeID:                nodeID,
+		coinbaseAccount:       conf.CoinbaseAccount,
+		goldenATXID:           conf.GoldenATXID,
+		layersPerEpoch:        conf.LayersPerEpoch,
+		cdb:                   cdb,
+		atxHandler:            hdlr,
+		publisher:             publisher,
+		nipostBuilder:         nipostBuilder,
+		postSetupProvider:     postSetupProvider,
+		layerClock:            layerClock,
+		syncer:                syncer,
+		log:                   log,
+		poetRetryInterval:     defaultPoetRetryInterval,
+		poetClientInitializer: defaultPoetClientFunc,
 	}
 	for _, opt := range opts {
 		opt(b)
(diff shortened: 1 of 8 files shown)


dshulyak:
there will be a significant difference between current poet config and poet that runs for the whole epoch.
currently poet runs several hash iterations (1 << n) and starts a next round. it will run multiple rounds in the same epoch.
so there is a significant chance that node can register in one of them and succeed.

poet that runs for the whole epoch starts round only once and executes it for the whole epoch. if it fails and gets restarted later - it will finish round, but with less ticks. in order for it to work it needs to have a persistent state. in current version state is deleted with the poet.

so given the above, deleting a poet completely will guarantee that smeshers won't generate activations for one epoch. does it change anything in your tests?

countvonzero:
> so given the above, deleting a poet completely will guarantee that smeshers won't generate activations for one epoch. does it change anything in your tests?

yes. that's exactly what i want to achieve. i'd like to create such scenario (no ATXs for >1 epoch) and make sure the network moves forward despite it. 

we should also test the case where poet's state persist and finish with less tick. but not the focus of this PR

activation/activation.go
```diff
@@ -394,14 +396,16 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) error {
...
 	client := b.poetClientInitializer(target)
 	// TODO(dshulyak) not enough information to verify that PoetServiceID matches with an expected one.
 	// Maybe it should be provided during update.
-	_, err := client.PoetServiceID(ctx)
+	sid, err := client.PoetServiceID(ctx)
 	if err != nil {
 		return fmt.Errorf("%w: %v", ErrPoetServiceUnstable, err)
 	}
 	b.pendingPoetClient.Store(unsafe.Pointer(&client))
+	b.log.With().Debug("preparing to update poet service", log.Binary("poet_id", sid))
```
dshulyak: i think it will encode using base64

systest/cluster/cluster.go
```diff
@@ -263,6 +273,17 @@ func (c *Cluster) AddSmeshers(cctx *testcontext.Context, n int) error {
 	return nil
 }
 
+// DeletePoet delete all poet servers.
```
dshulyak: DeletePoet deletes first poet?

systest/cluster/cluster.go
```diff
@@ -263,6 +273,17 @@ func (c *Cluster) AddSmeshers(cctx *testcontext.Context, n int) error {
...
+func (c *Cluster) DeletePoet(cctx *testcontext.Context) error {
+	for _, pod := range c.poets {
+		if err := deletePoet(cctx, pod.Name); err != nil {
+			return err
+		}
+		c.poets = c.poets[1:]
```
dshulyak, line 322: it will be using same slice, so if there are appends somewhere this will be a memory leak

systest/cluster/discover.go
```diff
@@ -7,14 +7,53 @@ import (
...
 
 	"golang.org/x/sync/errgroup"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/api/core/v1"
+	apimetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/spacemeshos/go-spacemesh/systest/testcontext"
 )
 
+func discoverPoets(ctx *testcontext.Context) ([]*v1.Pod, error) {
```
dshulyak: i think it is possible to implement it without so much duplication

dshulyak (approved):

activation/activation.go
countvonzero: changed to hex

systest/cluster/cluster.go
countvonzero: the code actually delete ALL poets, as i intended to. updated the name

systest/cluster/cluster.go
countvonzero: k. set it to nil at the beginning

systest/cluster/discover.go
countvonzero: used *NodeClient for poet too.

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

Issue #3297, opened 2022-06-25
block construction: execute txs in situ in the optimistic case


countvonzero:
## Description

see https://docs.google.com/document/d/1QSBsBdDvB3zc0J8jJgN5Y4iv9T5HDcD0QONsQbig5Po/edit#heading=h.1rzy5im1uaop

If optimistic filtering does not occur then all transactions from proposals are treated equally, since there’s no guarantee that everyone has access to the same state, even to perform a signature check.

If optimistic filtering does take place, the block is executed in situ while being constructed and any transaction that isn’t valid or where the account can’t cover the cost of gas is dropped from the block.

from Noam:
```
It's the only way to know for sure that a transaction should be included. Conservative state could 
cause over filtering and more importantly, it's not guaranteed to be identical for all nodes.

This concept basically mimics what miners do on Ethereum and other chains. If a candidate 
transaction doesn't have enough balance or the correct nonce at the slot in the block that it could 
occupy (point in time in history) then it's not included.
```

PR #3443, opened 2022-08-14
[Merged by Bors] - move block generation out of hare

countvonzero:
## Motivation

part of #3297

in order to generate block in situ in the optimistic filtering case, we need to change the block execution 
thread model to generate block layer by layer. this PR prepares for that change.

## Changes

- change block generator to run a background goroutine that listen to hare output and generate blocks
- move all block generation logic outside of hare.


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/7830828537?check_suite_focus=true)

this failed because the last commit to fix to #3446 was prematurely applied in this PR.

countvonzero:
bors try

blocks/generator.go
```diff
@@ -41,6 +55,13 @@ func defaultConfig() Config {
 // GeneratorOpt for configuring Generator.
 type GeneratorOpt func(h *Generator)
 
+// WithContext modifies parent context that is used for all operations in Sync.
+func WithContext(ctx context.Context) GeneratorOpt {
```
jonZlotnik, line 59: I don't really understand how this will be used. And should we really modify a context?
I feel like this should only ever be called once. But I do like the something.WithContext semantics.
Would a separate options struct be better than an abstract GeneratorOpt function?
I don't like the idea of Generator being mutable.
All these WithSomething functions should go on a config that can then be passed into the Generator via constructor. so: `options.WithContext(...)` and then `NewGenerator(options, ...)`

jonZlotnik (approved):

blocks/generator.go
countvonzero: you can see the usage in cmd/node/node.go

this is called golang functional options pattern. it's not about the Generator being mutable or not. 
it's about how to configure Generator in a readable but non-repetitive way. 

config is an option, but you would end up with very long struct init code. 
with this pattern, the struct just need to init the most basic default stuff, so tests do not need to pass in any options.
only when you actually need to configure the binary do you need to call the WithXXX().

this pattern is generally considered superior as i know it. and our current codebase uses it extensively.

countvonzero:
bors merge

PR #3447, opened 2022-08-16
[Merged by Bors] - tortoise: blocks with height from the future

### LINKED ISSUE #143: Implement: PoET ticks
Requires: spacemeshos/pm#124

- [x] ensure that merkle tree library supports unbalanced trees https://github.com/spacemeshos/merkle-tree/pull/16
- [x] make poet aware of epoch duration https://github.com/spacemeshos/poet/pull/112
- [x] integrate new poet config into systests https://github.com/spacemeshos/go-spacemesh/pull/3396
- [x] add poet ticks to activation and include them into weight computation https://github.com/spacemeshos/go-spacemesh/pull/3399
- [x] add poet tick to block from proposals https://github.com/spacemeshos/go-spacemesh/pull/3405
- [x] ballots should not be able to vote for future blocks https://github.com/spacemeshos/go-spacemesh/pull/3447

dshulyak:
the core of this change is that blocks with a higher height than a ballot are considered to be from future, and votes on such blocks are ignored. it changes termination condition from - all blocks must cross global threshold, to - either blocks at a specific height crossed positive threshold, and all blocks below cross negative threshold, or empty layer cross threshold.

in order to be compatible with verifying tortoise we will vote against blocks that are above block that crossed positive threshold.

verifying tortoise needs to count votes from ballots that can vote on the same blocks as this particular node. to implement it verifying tortoise has a concept of the reference height. reference height is a median of all activation tick heights that target a particular epoch. reference height of the layer is computed as a maximum height of the block below reference height of the layer. ballots that are below layer reference height won't be counted in the verifying tortoise. the assumption is that it should not have impact on verifying tortoise as long as there are no poets that are significantly faster than the rest of the network (by whole epoch).

closes: https://github.com/spacemeshos/pm/issues/143


countvonzero:
have not read the code yet. just want to get a high level understanding first.

> it changes termination condition from - all blocks must cross global threshold, to - either blocks at a specific height crossed positive threshold, and all blocks below cross negative threshold, or empty layer cross threshold.

let me use an example to understand. say a layer has 4 blocks [A, B, C, D] with height [1000, 10, 8, 0] (last one is the empty layer)
do you mean if one of the following termination condition happens, this layer is finalized.
- B crossed the + threshold and CD crossed the negative threshold
- D crossed the positive threshold

is this correct?

> reference height is a median of all activation tick heights that target a particular epoch. 

i had understood that reference height is different for smeshers and non-smeshers. a smesher uses its own height as the reference height, whereas a non-smesher uses the median height.

> reference height of the layer is computed as a maximum height of the block below reference height of the layer. ballots that are below layer reference height won't be counted in the verifying tortoise. 

this is because verifying tortoise only counts votes that agree with local opinions. and ballots that are below the layer reference height cannot vote for the same set of blocks as the node itself. correct?

is this also the algorithm for good ballots?

dshulyak:
> let me use an example to understand

yes, example is accurate. a bit confusing because it is sorted in reverse)

> i had understood that reference height is different for smeshers and non-smeshers. a smesher uses its own height as the reference height, whereas a non-smesher uses the median height.

in general right, in this change median is used for everyone. since this is not a security thing, it feels like a mess if results won't be reproducible because of some difference between nodes that can't be fixed by downloading more data.

> this is because verifying tortoise only counts votes that agree with local opinions. and ballots that are below the layer reference height cannot vote for the same set of blocks as the node itself. correct?

yes

> is this also the algorithm for good ballots?

yes, this is included into algorithm for good ballots. this is also one of the thing that i want to try to simplify, as i doesn't look strictly necessary

tortoise/full.go
```diff
@@ -28,14 +28,13 @@ type full struct {
 	abstain map[types.BallotID]map[types.LayerID]struct{}
 	base    map[types.BallotID]types.BallotID
 
+	empty map[types.LayerID]util.Weight
```
countvonzero, line 32: comment on what data this holds. is this those layers where hare output empty set?

tortoise/full.go
```diff
@@ -75,28 +70,50 @@ func (f *full) getVote(logger log.Log, ballot types.BallotID, blocklid types.Lay
...
-				current := f.weights[block]
-				switch vote {
+			if _, exist := f.empty[lid]; !exist {
+				f.empty[lid] = util.WeightFromUint64(0)
+			}
+			blocks := f.blocks[lid]
+			empty := true
+			for i := range blocks {
+				block := &blocks[i]
+				if block.height > ballot.height {
```
countvonzero, line 105: does it constitute malicious behavior if a ballot doesn't vote against this block?
if so, is this something we should check in the ballot validation flow?

tortoise/full.go
```diff
@@ -75,28 +70,50 @@ func (f *full) getVote(logger log.Log, ballot types.BallotID, blocklid types.Lay
...
+					empty = false
+					block.weight.Add(ballot.weight)
 				case against:
-					current = current.Sub(ballotWeight)
+					block.weight.Sub(ballot.weight)
 				}
-				f.weights[block] = current
+			}
+			if empty && !f.abstained(ballot.id, lid) {
+				f.empty[lid].Add(ballot.weight)
```
countvonzero, line 117: this is how i read this code:
if no block in this layer was voted by any ballot, and we are no longer waiting for hare output, the node will consider this layer empty. is this correct?

tortoise/verifying.go
```diff
@@ -74,6 +75,26 @@ type verifying struct {
...
+	if !exist {
+		current = v.layerReferenceHeight[block.LayerIndex.Sub(1)]
+	}
+	if block.TickHeight > current {
+		current = block.TickHeight
+	}
+	v.layerReferenceHeight[block.LayerIndex] = current
+}
+
+func (v *verifying) getReferenceHeight(lid types.LayerID) uint64 {
```
countvonzero: getLayerReferenceHeight?

tortoise/verifying.go
```diff
@@ -149,39 +170,43 @@ func (v *verifying) verify(logger log.Log, lid types.LayerID) bool {
...
-			)
-			return false
-		}
-		v.validity[bid] = vote
+	if verifyLayer(
+		logger,
+		v.blocks[lid],
+		v.validity,
+		func(block blockInfo) sign {
+			decision, _ := getLocalVote(v.commonState, v.Config, lid, block.id)
```
countvonzero: can we check the height before this call?

tortoise/verifying.go
```diff
@@ -149,39 +170,43 @@ func (v *verifying) verify(logger log.Log, lid types.LayerID) bool {
...
+			if block.height > v.getReferenceHeight(lid) {
+				return abstain
```
countvonzero: i recall we talked about this in a tortoise code walk with Tal.
the constant `abstain` here is confusing. abstain is relevant when voting. here we are just adding up margins. i think this constant should be renamed to `neutral` instead.

tortoise/verifying.go
```diff
@@ -149,39 +170,43 @@ func (v *verifying) verify(logger log.Log, lid types.LayerID) bool {
...
+func (v *verifying) countGoodBallots(logger log.Log, lid types.LayerID, ballots []tortoiseBallot) (util.Weight, int) {
 	sum := util.WeightFromUint64(0)
 	n := 0
 	for _, ballot := range ballots {
 		if ballot.weight.IsNil() {
+			logger.With().Debug("ballot weight is nil", ballot.id)
+			continue
+		}
+		// get height of the max votable block
+		if refheight := v.getReferenceHeight(lid.Sub(1)); refheight > ballot.height {
```
countvonzero: i am still confused about this. in your reply to my earlier questions

> in general right, in this change median is used for everyone. since this is not a security thing, it feels like a mess if results won't be reproducible because of some difference between nodes that can't be fixed by downloading more data.

but the case i am thinking is:
minerH: miner height 100
refH: epoch medium height 110
lyrRefH: layer ref height 109
blockH: block A height 108
ballotH: ballot B height 120

in this case, can we consider ballot B a good ballot? theoretically miner cannot vote for a block A, but ballot B can. but it seems this implementation allows miner to vote for block A anyway?

tortoise/verifying.go
```diff
@@ -74,6 +75,26 @@ type verifying struct {
 	abstainedWeight map[types.LayerID]util.Weight
 	// total weight of good ballots from verified + 1 up to last processed
 	totalGoodWeight util.Weight
+	// layerReferenceHeight is a highest block height below reference height
+	layerReferenceHeight map[types.LayerID]uint64
+}
+
+func (v *verifying) onBlock(block *types.Block) {
+	if block.TickHeight > v.referenceHeight[block.LayerIndex.GetEpoch()] {
```
countvonzero, line 83: this caused the node not to vote for a block higher than the medium epoch height, right?

tortoise/verifying.go
countvonzero: the code as is still votes abstain for zdist layers even when hare outputs empty layer immediately. is that deliberate or should that be a TODO?

tortoise/util.go
```diff
@@ -135,3 +137,61 @@ func maxLayer(i, j types.LayerID) types.LayerID {
...
+		if decision == abstain {
+			// all blocks with the same height should be finalized
+			if positive && block.height > supported.height {
+				decision = against
+			} else {
+				return false
+			}
+		} else if decision == support {
+			positive = true
+			supported = block
```
countvonzero, line 175: this will overwrite the previous supported in the case where we support multiple blocks in the same layer.
are you saying that if we support multiple blocks in the same layer, they must have the same height?

tortoise/full.go
dshulyak: they are not malicious. neither support or against affect the protocol

tortoise/full.go
dshulyak: yes

tortoise/util.go
dshulyak: the point here is to wait until all blocks with the same height are verified, because the same weight should vote for them

tortoise/verifying.go
dshulyak: i think right, but i am confused by wording

verifying tortoise won't count votes on a block from future, but tortoise still votes on every block

tortoise/verifying.go
dshulyak: right we can

dshulyak: are you referring to getLocalVote internals? no, it doesn't vote abstain if hare output for the layer is finalized

otherwise if block is from future tortoise will not count votes from it, hence abstain (neutral) opinion. it won't change because of zdist

tortoise/verifying.go
dshulyak: sure will add an alias

tortoise/verifying.go
dshulyak: > but it seems this implementation allows miner to vote for block A anyway?

miner height is not relevant by itself. the only relevant height is a reference height. it could be equal to miner height. but i am using median height as a reference height everywhere, so that tortoise will work consistently across the network. 

this is not about allowing miner to vote, this is about what votes will be counted

tortoise/full.go
dshulyak: this is the weight of ballots that vote for the empty layer

tortoise/verifying.go
dshulyak: in the example above you are not providing a ballot from that miner. if it had a ballot with height=100 - it won't be counted.

tortoise/full.go
dshulyak: > we are no longer waiting for hare output,

the voter is no longer waiting for hare output, not the node that counts votes. 

> the node will consider this layer empty. 

also it is not about node, it is about deciding if ballot votes for empty layer

tortoise/full.go
countvonzero: you mean because full tortoise is not counting the votes anyway?

tortoise/verifying.go
countvonzero: > this is not about allowing miner to vote, this is about what votes will be counted

right. as you can see i am consistently confusing voting vs counting votes... 🙄

countvonzero (approved):
thanks for all the answers.

tortoise/full.go
dshulyak: yes

dshulyak:
bors merge

dshulyak:
bors merge

PR #3449, opened 2022-08-16
[Merged by Bors] - block: generate certificate for hare output

### LINKED ISSUE #3199: hare: save certificate on completion

### LINKED ISSUE #3200: sync: hare output should be saved only if peer can provide a certificate for it
after: https://github.com/spacemeshos/go-spacemesh/issues/2942, https://github.com/spacemeshos/go-spacemesh/issues/3199

1. download certificate for the layer within hdist
2. validate it using eligibility oracle from hare module
3. save hare output and certificate only if validation passes

this is important for safety, otherwise it will be trivial to force a new node to vote in the specific way.

countvonzero:
## Motivation

Closes #3200 
Part of #3199 

## Changes

- certificate is saved in table `layers` instead of `blocks` because it's expected for hare to output empty block id for a layer, and that certifiers can certify an empty block id for a layer.
- certifier waits till it has f+1 signatures and produce a certificate and save to db.
- syncer fetches the certificate to use as hare output


countvonzero:
@dshulyak can you review this PR please? thank you

countvonzero:
bors try

sql/migrations/0001_initial.sql
```diff
@@ -28,6 +28,7 @@ CREATE TABLE layers
     id              INT PRIMARY KEY DESC,
     weak_coin       SMALL INT,
     hare_output     VARCHAR,
+    cert            BLOB,
```
jonZlotnik, line 31: Why did you decide to add it to the layers table instead of the block table?

common/types/block.go
```diff
@@ -166,3 +166,35 @@ type BlockContextualValidity struct {
 	ID       BlockID
 	Validity bool
 }
+
+// Certificate represents a certified block.
+type Certificate struct {
+	BlockID    BlockID
+	Signatures []CertifyMessage
```
jonZlotnik, line 173: By storing the entire message in the certificate, we'll be storing the LayerID and BlockID for every signature of every certificate. It would be best if we have one struct for the gossiped message and another for a signature stored in the certificate.

It's not really possible to do the above if we sign the entire message unless we store CertifyContent with nil values and then have some recontstuction upon certificate verification. But, if we sign only the BlockID, then the signature can be decoupled from the message. We don't need to sign the entire message because the VRF proof (which uses the same key pair as the ed signer) allows us to validate and verify the rest of the values in the message.

Extra bytes stored and then sent over the wire in fetch by keeping LayerID and BlockID in Certificate for every signature:

(hdist) * (f)(len(LayerID)+len(BlockID)) =>  200 * 400 (32 + 20) = 4.16 MB of redundant data

sql/layers/layers.go
```diff
@@ -64,6 +65,47 @@ func SetHareOutput(db sql.Executor, lid types.LayerID, output types.BlockID) err
...
 
+// SetHareOutputWithCert sets the hare output for the layer with a block certificate.
+func SetHareOutputWithCert(db sql.Executor, lid types.LayerID, cert *types.Certificate) error {
+	output := cert.BlockID
+	data, err := codec.Encode(cert)
+	if err != nil {
+		return fmt.Errorf("encode cert %w", err)
+	}
+	if _, err := db.Exec(`insert into layers (id, hare_output, cert) values (?1, ?2, ?3) 
+					on conflict(id) do update set hare_output=?2, cert=?3;`,
```
jonZlotnik, line 76: Is the goal here to disallow updating a BlockID unless it's accompanied by a certificate?

two concerns:
1. If we end up with >1 certificates for different block ids in the same layer, the block and certificate that were there previously get overwritten.
2. Why does having a certificate change things?

The HareOutput setters seem inconsistent. Is it reasonable to all of a sudden make the BlockID mutable if it's accompanied by a certificate?

Doesn't other code assume the BlockID to be immutable?

hare/haretypes.go
```diff
@@ -22,14 +22,12 @@ const (
...
 const (
-	statusRound uint32 = iota
-	proposalRound
-	commitRound
-	notifyRound
+	preRound      = eligibility.HarePreRound
+	statusRound   = eligibility.HareStatusRound
+	proposalRound = eligibility.HareProposalRound
+	commitRound   = eligibility.HareCommitRound
+	notifyRound   = eligibility.HareNotifyRound
```
jonZlotnik, line 30: very good and quick refactor

fetch/layers.go
```diff
@@ -308,10 +307,20 @@ func notifyLayerDataResult(db *sql.Database, layerID types.LayerID, msh meshProv
...
+		best             *types.Certificate
+		maxProcessed     = types.NewLayerID(0)
 	)
 	for _, res := range lyrResult.responses {
 		if res.err == nil && res.data != nil {
 			success = true
+			if res.data.HareOutput != nil {
+				if best == nil ||
+					res.data.ProcessedLayer.After(maxProcessed) ||
+					res.data.ProcessedLayer == maxProcessed && best.BlockID == types.EmptyBlockID && res.data.HareOutput.BlockID != types.EmptyBlockID {
```
jonZlotnik, line 338: Can you pls add a comment that explains what would lead up to this case evaluating to true.

fetch/layers.go
```diff
@@ -331,11 +340,11 @@ func notifyLayerDataResult(db *sql.Database, layerID types.LayerID, msh meshProv
 	}
 
 	if result.Err == nil {
-		// FIXME should not blindly save hare output from peer, potentially overwriting local hare's
-		// consensus result. see https://github.com/spacemeshos/go-spacemesh/issues/3200
-		if err := layers.SetHareOutput(db, layerID, lyrResult.hareOutput); err != nil {
-			logger.With().Error("failed to save hare output from peers", log.Err(err))
-			result.Err = err
+		if best != nil {
+			if err := layers.SetHareOutputWithCert(db, layerID, best); err != nil {
```
jonZlotnik, line 363: This is also called in certifier.go
In the last layers of sync, we might fetch a certificate that certifies a block different from what pops out of the certifier. Since `sync` is not in sync with hare, if we collect signatures from gossip that certify a block different than one we've fetched, there's no way in this implementation to know that we've been "convinced" of 2 blocks.

This comes back to the question "why do we support multiple blocks per layer in some places but not others." We  really do need a complete answer to this. And we need to decide, moving forward, what the correct response to a situation is where the possibility exists of more than one BlockId. I don't care how likely or unlikely it is, we need to know what to do if it happens or formally prove it won't happen.

fetch/handler.go
```diff
@@ -90,7 +90,9 @@ func (h *handler) handleLayerDataReq(ctx context.Context, req []byte) ([]byte, e
 		h.logger.WithContext(ctx).With().Warning("failed to get layer blocks", lyrID, log.Err(err))
 		return nil, errInternal
 	}
-	if ld.HareOutput, err = layers.GetHareOutput(h.db, lyrID); err != nil {
+
+	ld.HareOutput, err = layers.GetCert(h.db, lyrID)
```
jonZlotnik, line 92: update handleLayerDataReq comment to better explain what it's returning.
Maybe reference the layerData datastructure instead of listing all the contents.

fetch/fetch_test.go
```diff
@@ -274,7 +275,7 @@ func TestFetch_GetLayerData(t *testing.T) {
 				f.mLyrS.EXPECT().Request(gomock.Any(), p, gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(
 					func(_ context.Context, _ p2p.Peer, _ []byte, okFunc func([]byte), errFunc func(error)) error {
 						if tc.errs[idx] == nil {
-							go okFunc(generateLayerContent(false))
+							go okFunc(generateLayerContent(false, processed))
```
jonZlotnik: How did this compile before?
How did this pass CI before?

blocks/certifier.go
```diff
@@ -0,0 +1,439 @@
...
+		return errInternal
+	}
+	if block.LayerIndex != lid {
+		logger.With().Warning("unexpected layer id for block",
+			log.Stringer("expected", block.LayerIndex),
+			log.Stringer("received", lid))
+		return errMalformedData
+	}
+	return nil
+}
```
jonZlotnik: Why do we fail if the block isn't in the database?
When we sync, we need to accept blocks that have been certified by an eligible committee whether or not we've seen the BlockID before.

I don't see how we could know the LayerIndex of a block before receiving a certificate for the layer.

Does this work?

blocks/certifier.go
```diff
@@ -0,0 +1,439 @@
...
+		} else if created {
+			c.certMsgs[lid].done = true
+		}
+	}
+
+	return nil
+}
+
+func (c *Certifier) tryGenCertLocked(logger log.Log, lid types.LayerID, bid types.BlockID, sigs []types.CertifyMessage) (bool, error) {
+	if len(sigs) < c.cfg.CertifyThreshold+1 {
```
jonZlotnik: Why doesn't the CertifyThreshold already include the +1? Just seems too easy to forget the +1.

blocks/certifier.go
```diff
@@ -0,0 +1,439 @@
...
+// RegisterDeadline register the deadline to stop waiting for enough signatures to generate
+// the certificate.
+func (c *Certifier) RegisterDeadline(lid types.LayerID, bid types.BlockID, now time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if _, ok := c.certMsgs[lid]; !ok {
+		c.certMsgs[lid] = &certInfo{
+			bid:        bid,
+			deadline:   now.Add(time.Duration(c.cfg.SignatureWaitDuration) * time.Second),
+			signatures: make([]types.CertifyMessage, 0, c.cfg.CommitteeSize),
```
jonZlotnik, line 182: Why is the underlying array of size CommitteeSize and not CertifyThreshold+1?

blocks/certifier.go
```diff
@@ -0,0 +1,439 @@
...
+		}
+	}
+}
+
+// CertifyMaybe signs the hare output, along with its role proof as a certifier, and gossip the CertifyMessage
+// if the node is eligible to be a certifier.
+func (c *Certifier) CertifyMaybe(ctx context.Context, logger log.Log, lid types.LayerID, bid types.BlockID) error {
```
jonZlotnik: Rename to `CertifyIfEligible` please

`CertifyMaybe` is very ambiguous

blocks/certifier.go
```diff
@@ -0,0 +1,439 @@
...
+	} else if state == expired {
+		return errMsgTooOld
+	} else if state == certified {
+		// should still gossip this msg to peers even when this node has created a certificate
+		return nil
+	}
+
+	logger.WithFields(bid)
+	if bid != types.EmptyBlockID {
+		if err := c.verifyLayerID(logger, lid, bid); err != nil {
```
jonZlotnik: A call to database for every message in a certificate?
Seems a little excessive. That's 400 successive database calls when validating a certificate. 

Especially since the certificate needs only be internally consistent to be trusted. We can take the certificate as truth. No need to check against our local understanding of the layer.

jonZlotnik (changes requested):

blocks/certifier.go
```diff
@@ -0,0 +1,439 @@
...
+	if state, err := c.getCertState(logger, lid, bid); err != nil {
+		return err
+	} else if state == expired {
+		return errMsgTooOld
+	} else if state == certified {
+		// should still gossip this msg to peers even when this node has created a certificate
+		return nil
+	}
+
+	logger.WithFields(bid)
```
dshulyak: logger = logger.WithFields(bid)
it looks like it always creates a new instance

common/types/block.go
```diff
@@ -166,3 +166,35 @@ type BlockContextualValidity struct {
 	ID       BlockID
 	Validity bool
 }
+
```
dshulyak, line 169: please call go generate on this file. you will need scalegen binary

blocks/certifier.go
countvonzero: this is the capacity. not the size. the size is 0 at this point.

blocks/certifier.go
countvonzero: it's not about checking local understanding of layer. the point was to verify this block is for real, that we have this block.

the concept of this network is that we only accept and gossip about msg we validate.
and a msg is only valid if everything referenced in the msg were validated.

in any case, i removed this check as i later changed the code to check it earlier at getCertState. there it was just an in-memory check.

blocks/certifier.go
countvonzero: see my comment above about how we consider a msg valid.

blocks/certifier.go
countvonzero: i'm trying to be consistent with the config. the threshold here is taken from hare config F. and all checks in hare, is adding 1 to it.

common/types/block.go
countvonzero: for db: we do want to save as little as possible. i meant to address that in a following PR where it does prune the certificate. 
for over-the-wire, we do want the message to be complete for validation purpose.

in any case, i've prune the certificate data before it's saved to db and restore it upon retrieval.

countvonzero: and i deliberately avoid to have separate messages for gossip and for db. that's proved to be a little messy in the past.

fetch/fetch_test.go
countvonzero: i only changed the func signature in this PR.

fetch/layers.go
countvonzero: sync and hare are coordinated. syncer will not sync a layer from peers if hare is working on it. and will only sync if 10+ layers have passed and hare hasn't produced an output.

in my implementation, i decided to only accept the layer/blockID that block generator told it to. see `certifier.RegisterDeadline()`. the node is only responsible for the hare committee it is in. the gossip handler will only process msg that matches layer id and block id registered.

sql/layers/layers.go
countvonzero: no. in a steady state. we will always first call SetHareOutput (by mesh.go after a block is generated)
this is important so tortoise can make progress right away without waiting for the certificate.

later, when a block is certified, then it call SetHareOutputWithCert to store the cert. it can change the local opinion at this point. for example, if the local hare fails, it will call SetHareOutput() with types.EmptyBlockID. but later when we sync Certificate from a peer, we then can overwrite this to their hare output.

i don't want to put application logic here. a certificate can only be saved if it gathered f+1 signatures. any certificate that passed that threshold can be saved here.

sql/migrations/0001_initial.sql
countvonzero: i explained in the PR description. we need to certify a hare output with empty block ID too.

this is certificate on hare output for a given layer. so it is more natural to keep it in layers table.

countvonzero:
bors try

blocks/certifier.go
jonZlotnik: The size of the slice is 0, but you're allocating a CommitteeSize array in memory to hold a slice that'll only reach CommitteeSize/2 + 1

blocks/certifier.go
jonZlotnik: I understand, but it should then be `MaxAdversaries`+1 or something.
CertifyThreshold just sounds like it should contain the actual threshold required to certify a block and not be off-by-one.

blocks/certifier.go
countvonzero: as you wish. done

blocks/certifier.go
countvonzero: initialized the number to f+1 instead.

fetch/layers.go
jonZlotnik: Must have missed that 10 layer delta behavior. But @noamnelke maintains that we still need to support tracking of more than one block per layer.

sql/layers/layers.go
jonZlotnik: Okay, so we need to support multiple blocks per layer. I think we need to move the certificates to the blocks table.
@noamnelke says it's rare, but we need to support multiple blocks per layer. I don't think we can just overwrite previous opinions.

countvonzero: even if we want to accept multiple blocks per layer, can you explain how we can handle when hare output EmptyBlockID if we use blocks table?

sql/migrations/0001_initial.sql
jonZlotnik: I would say it belongs next to the block and not the layer. The other thing is that we need to support multiple blocks, so we can't have just one certificate per layer, but the 1:1 block:cert relationship still holds.

sql/layers/layers.go
jonZlotnik: I think blockIDs should remain immutable.

countvonzero: i don't think you are understanding my question. if we put the certificate in table "blocks", where will i save certificate for an empty block id from layer 100, 111, 123? they are layers that hare output empty block

cmd/node/node.go
```diff
@@ -535,9 +537,30 @@ func (app *App) initServices(ctx context.Context,
...
+		hOracle = eligibility.New(beaconProtocol, cdb, signing.VRFVerify, vrfSigner, app.Config.LayersPerEpoch, app.Config.HareEligibility, app.addLogger(HareOracleLogger, lg))
+		// TODO: genesisMinerWeight is set to app.Config.SpaceToCommit, because PoET ticks are currently hardcoded to 1
+	}
+	app.certifier = blocks.NewCertifier(sqlDB, hOracle, nodeID, sgn, app.host, clock,
+		blocks.WithCertContext(ctx),
+		blocks.WithCertConfig(blocks.CertConfig{
+			CommitteeSize:         app.Config.HARE.N,
+			CertifyThreshold:      app.Config.HARE.F + 1,
+			SignatureWaitDuration: app.Config.HARE.WakeupDelta,
+			NumLayersToKeep:       app.Config.Tortoise.Zdist,
```
jonZlotnik, line 556: Missed this before. Why Zdist instead of Hdist?

blocks/certifier.go
```diff
@@ -0,0 +1,413 @@
...
+		} else if created {
+			c.certMsgs[lid].done = true
+		}
+	}
+
+	return nil
+}
+
+func (c *Certifier) tryGenCertLocked(logger log.Log, lid types.LayerID, bid types.BlockID, sigs []types.CertifyMessage) (bool, error) {
+	if len(sigs) < c.cfg.CertifyThreshold {
```
jonZlotnik: We need to be counting eligibilities, and not signatures.
I may have made the same mistake in my PR.

blocks/certifier.go
```diff
@@ -0,0 +1,413 @@
...
+
+func (c *Certifier) getCertState(logger log.Log, lid types.LayerID, bid types.BlockID) (certState, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.getCertStateLocked(logger, lid, bid)
+}
+
+func (c *Certifier) getCertStateLocked(logger log.Log, lid types.LayerID, bid types.BlockID) (certState, error) {
+	if ci, ok := c.certMsgs[lid]; !ok {
+		logger.Error("layer not registered for cert")
```
dshulyak: this is expected if very old message is received for any reason, at most should be a warning
also  why internal?

blocks/certifier.go
```diff
@@ -0,0 +1,413 @@
...
+		return unknown, errInternal
+	} else if ci.bid != bid {
+		logger.Error("block not registered for cert")
```
dshulyak: some part of the committee can be malicious and will sign different block
we should collect certificates for all blocks, until one of the them has enough space units

blocks/certifier.go
```diff
@@ -0,0 +1,413 @@
...
+	c.certMsgs[lid].signatures = append(c.certMsgs[lid].signatures, msg)
+
+	if created, err := c.tryGenCertLocked(logger, lid, bid, c.certMsgs[lid].signatures); err != nil {
+		return err
+	} else if created {
+		c.certMsgs[lid].done = true
+	}
+	return nil
+}
+
```
dshulyak, line 362: i don't understand the purpose of this method
certificate should be generatable when handler received sufficient number of messages to cross threshold

blocks/certifier.go
dshulyak: :+1:
i think eligibility count represents weight units. and cert needs to be atleast 401 weight units

sql/layers/layers.go
```diff
@@ -64,6 +65,64 @@ func SetHareOutput(db sql.Executor, lid types.LayerID, output types.BlockID) err
...
 }
 
+func fill(lid types.LayerID, bid types.BlockID, cert *types.Certificate) {
+	for _, msg := range cert.Signatures {
+		msg.LayerID = lid
+		msg.BlockID = bid
+	}
+}
+
+func prune(cert *types.Certificate) {
```
dshulyak: not sure what was the idea but it doesn't prune stuff, it just zeroes it out

sql/migrations/0001_initial.sql
dshulyak: > i explained in the PR description. we need to certify a hare output with empty block ID too.

can you please get agreement from Tal/Noam about this. it sounds like a mistake to me

in general i would also prefer hare_output to be stored in the blocks and avoid using zeroed block id the way it is used now, but i am also fine leaving it for future

p2p/pubsub/pubsub.go
```diff
@@ -62,6 +62,9 @@ const (
 	// HareProtocol is the protocol id for hare messages.
 	HareProtocol = "/sm/hr/1"
 
+	// BlockCertify is the protocol id for block certification.
+	BlockCertify = "/sm/bc/1"
```
dshulyak: i should have paid more attention to the original change :)

we really should remove all useless stuff from this names, they will add bloat on the wire. for example `/sm/` and third slash is not needed. so as long as they are unique we are good

blocks/certifier.go
```diff
@@ -0,0 +1,413 @@
...
+func (c *Certifier) tryGenCert() error {
```
dshulyak: i don't understand the purpose of this method
certificate should be generatable when handler received sufficient number of messages to cross threshold

sql/layers/layers.go
dshulyak: > In the situation you describe, there would be either one or 0 hare outputs --- once the rejoin occurs, the previously certified hare output might end up decertified, since the threshold has risen

you can have multiple certificates when the part of the same committee signs more than one block on purpose. in such case we should vote for empty layer. in order for this to happen you need an isolated partition where only malicious nodes can communicate with both sides, and maybe even more requirements

if committee changes based on the atxs state - previous certificate will be invalid, but in such case you will still have 1 or 0.

sql/migrations/0001_initial.sql
countvonzero: will do followup PR once research conclude

sql/layers/layers.go
countvonzero: i'll do a follow up PR once research conclude on this.

sql/layers/layers.go
countvonzero: yikes. changed to use new struct so the field is nil slice

p2p/pubsub/pubsub.go
countvonzero: shortened

fetch/layers.go
countvonzero: done by popular demand.

common/types/block.go
countvonzero: done

cmd/node/node.go
countvonzero: it can be arbitrary numbers really. just the amount of old data to keep.
tortoise waits for zdist layer to give up waiting for hare output so just to be consistent.

hdist is the number of layers tortoise will use the hare output. it's not relevant here as syncer will be taking over to sync cert from peers so that tortoise can still use the synced cert within hdist.

blocks/certifier.go
countvonzero: fixed.

blocks/certifier.go
countvonzero: glad you caught me. it was meant to prune. but then i over did it. now it's just pruning old data.

blocks/certifier.go
countvonzero: fixed

blocks/certifier.go
countvonzero: changed code to allow for multiple blocks.

blocks/certifier.go
countvonzero: done.

sql/layers/layers.go
countvonzero: hmmm... this doesn't matter. the BlockID is an array and so is still zeroed out values. will revert this.

blocks/certifier.go
```diff
@@ -0,0 +1,404 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.getCertStateLocked(logger, lid, bid)
+}
+
+func (c *Certifier) getCertStateLocked(logger log.Log, lid types.LayerID, bid types.BlockID) (certState, error) {
+	if _, ok := c.certMsgs[lid]; !ok {
+		logger.Warning("layer not registered for cert")
+		return unknown, errUnexpectedMsg
+	} else if _, ok = c.certMsgs[lid][bid]; !ok {
```
dshulyak: so you won't be collecting certificates for blocks that this node didn't generate? i misunderstood it during initial review

so if for any reason (missing state/bugs) this node wasn't able to form full certificate - it will download it only from other peers

in this case it probably makes no sense to store a map with blocks

blocks/certifier.go
```diff
@@ -0,0 +1,404 @@
...
+	errInvalidCertMsg = errors.New("invalid cert msg")
+	errUnexpectedMsg  = errors.New("unexpected lid/bid")
+	errInternal       = errors.New("internal")
+)
+
+// CertConfig is the config for Certifier.
+type CertConfig struct {
+	CommitteeSize         int
+	CertifyThreshold      int
+	SignatureWaitDuration int
```
dshulyak: better to use time.Duration and avoid multiplying it by time.Second later

blocks/certifier.go
```diff
@@ -0,0 +1,404 @@
...
+func (c *Certifier) RegisterDeadline(lid types.LayerID, bid types.BlockID, now time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if _, ok := c.certMsgs[lid]; !ok {
+		c.certMsgs[lid] = make(map[types.BlockID]*certInfo)
+	}
+
+	c.certMsgs[lid][bid] = &certInfo{
+		deadline:   now.Add(time.Duration(c.cfg.SignatureWaitDuration) * time.Second),
+		signatures: make([]types.CertifyMessage, 0, c.cfg.CertifyThreshold),
```
dshulyak: i believe this makes no difference now

dshulyak (approved):

blocks/certifier.go
countvonzero: that's right. i had assume that you want me to write code such that the malicious nodes who are able to connect to all network splits can also use. i will revert back to single block assumption.

blocks/certifier.go
countvonzero: not sure what you mean. do you mean it makes no difference what the capacity is?

dshulyak: i meant that we don't expect number of signatures to be equal to this threshold. we expect eligibility count to be equal to threshold

countvonzero: right. but i still need to initialize it to some capacity, and given that we don't use fraction for eligibility count, this is a safe number to use. or the committee size.

sql/migrations/0001_initial.sql
dshulyak: for posterity - my concern had no ground

sql/layers/layers.go
countvonzero: concluded as in https://github.com/spacemeshos/go-spacemesh/pull/3449#discussion_r949231928

countvonzero:
bors try

blocks/certifier.go
jonZlotnik: Yeah, the CertifyThreshold is the maximum it can be. (i.e. everybody's signer's weight is 1)

fetch/layers.go
```diff
@@ -278,6 +278,19 @@ func (l *Logic) receiveLayerContent(ctx context.Context, layerID types.LayerID,
 		}
 	}
 
+	// process the certificate after blocks are fetched
+	if peerRes.data != nil && peerRes.data.HareOutput != nil {
+		cert := peerRes.data.HareOutput
+		if cert.BlockID != types.EmptyBlockID && !contains(peerRes.data.Blocks, cert.BlockID) {
+			logger.With().Warning("certificate block id not in peer's layer content")
+			peerRes.err = errInvalidCertificate
+		}
```
jonZlotnik, line 287: We can't error with errInvalidCertificate in this case. If we receive a valid certificate for a BlockID that we can't find anywhere, then we need to try to find it from a different peer, or try again on next sync.

jonZlotnik: Annoying, but can we have a new error for this case something like `errMissingBlock`

blocks/certifier.go
```diff
@@ -0,0 +1,394 @@
...
+
+	db         *sql.Database
+	oracle     hare.Rolacle
+	nodeID     types.NodeID
+	signer     *signing.EdSigner
+	publisher  pubsub.Publisher
+	layerClock layerClock
+
+	mu       sync.Mutex
+	certMsgs map[types.LayerID]*certInfo
```
jonZlotnik: rename to `certs`

blocks/certifier.go
```diff
@@ -0,0 +1,394 @@
...
+	bid := msg.BlockID
+	if state, err := c.getCertStateLocked(logger, lid, bid); err != nil {
+		return err
+	} else if state != pending { // already certified
+		return nil
+	}
+
+	if _, ok := c.certMsgs[lid]; !ok {
+		logger.Fatal("missing layer in cache")
+	}
```
jonZlotnik: replace lone if statement with `else if state == unknown`

`unknown` is returned from getCertStateLocked, but never checked by a calling method.

blocks/certifier.go
```diff
@@ -0,0 +1,394 @@
...
+	lid := msg.LayerID
+	bid := msg.BlockID
+	if state, err := c.getCertState(logger, lid, bid); err != nil {
+		return err
+	} else if state == expired {
+		return errMsgTooOld
+	} else if state == certified {
+		// should still gossip this msg to peers even when this node has created a certificate
+		return nil
+	}
```
jonZlotnik, line 320: `else if state == unknown` behavior missing

blocks/certifier.go
jonZlotnik: actually, this case is caught by the `err != nill`

blocks/certifier.go
jonZlotnik: same here, `err != nil` catches the unknown case.

blocks/certifier.go
jonZlotnik: So we can just get rid of this last if statement.

jonZlotnik (approved):
My last comments don't affect behavior.

fetch/layers.go
countvonzero: it's not about missing block, it's about the block in certificate isn't referenced in the peers data.
renamed to errCertifiedBlockNotReferenced

blocks/certifier.go
countvonzero: but they are not certs. i used the name `cert` for the actual certificate. i don't want my readers to confuse the two (CertifyMessage and Certificate)

countvonzero: renamed to certifyMsgs to be clear

blocks/certifier.go
countvonzero: this is there in case earlier code screwed up and didn't check for existence. the reason being i used `c.certMsgs[lid]` directly in the code below. without this check, and if previous block were removed, i'll get a nil ptr. with this check, at this i get a fatal log. either way. it dies. but fatal log is more graceful.

blocks/certifier.go
jonZlotnik: fair point, that works

blocks/certifier.go
jonZlotnik: Maybe a comment is required to explain the choice?

countvonzero: why do you think this matters so much? 

for me it really doesn't matter. whether it's CertifyThreshold or CommitteeSize, they are O(1) difference. we could also just do make([]types.CertifyMessage, 0) and call it a day...

sure. next time i have something meaningful to update, i'll add a comment.

jonZlotnik: I agree it doesn't matter so much. But I'm thinking about when we have contributors from the wild they'll want to know why. If it was enterprise software, I would have left it at CommitteeSize when I read it the first time.

The less magic and mysteries the better. It's hard for a dev new to the project to know when something matters and when something doesn't matter.

countvonzero: right. so if i change the code to `make([]types.CertifyMessage, 0)`, there will be no confusion because the array will just grow organically as we receive message. no comments needed.

jonZlotnik: I guess so. seems like a reasonable enough sacrifice in the name of understandability.

countvonzero: changed back to CommitteeSize because i love to come full circle.

jonZlotnik: omg stop whyyyy, I'm changing my review to needs work

jonZlotnik (changes requested):
I don't like circles

blocks/certifier.go
countvonzero: exactly why do you think this need work? this is the maximum capacity that can be possible. as such, no comments is required. adding unnecessary comments will only serve to confuse.

you yourself said
> If it was enterprise software, I would have left it at CommitteeSize when I read it the first time.

please unblock. this is not cool. @jonZlotnik

jonZlotnik: I would have left it at CommitteeSize because it doesn't matter in enterprise software. You have your colleagues to explain things to you so I would have just ignored that it was too big an array for what it was. `CommitteeSize` makes it more confusing than `CertifyThreshold`

The maximum possible capacity that it would grow to is `CertifyThreshold` because if we assume every committee member has one eligibility, your implementation collects signatures until CertifyThreshold number of signatures is reached, and then builds/saves the certificate. It drops any further signatures (as it should):
https://github.com/spacemeshos/go-spacemesh/blob/a5269257c54ee59fc082f37b9b2ee2f6431d1a33/blocks/certifier.go#L340-L351

If any committee member has any more than 1 eligibility, the number of signatures required to reach CertifyThreshold decreases.

jonZlotnik: I'm okay with either:
 `make([]types.CertifyMessage, 0, c.cfg.CertifyThreshold)`
or 
`make([]types.CertifyMessage, 0)`

jonZlotnik (approved):
> "do what's right and leave the consequences to code"
\- some sticker, 2022

blocks/certifier.go
countvonzero: my rationale is, committee size is the max size it can grow for any implementation. the current impl, the max is the threshold yes. but say we decided to wait for ALL msgs before we certify? and because it matters so little (800 vs 401), i decided to go for the future-safe one.

jonZlotnik: okidoki

countvonzero:
bors try

countvonzero:
bors merge

sql/layers/layers.go
dshulyak: @countvonzero that comment has nothing to do with the comment above

countvonzero: i stand corrected. it should be this one that concludes this. https://github.com/spacemeshos/go-spacemesh/pull/3449#discussion_r949139516

do you agree? @dshulyak

dshulyak: my point was that we should vote for an empty layer if committee signed more than one block intentionally, since it clearly malicious behavior. i believe that Tal mentioned it some time in the past, but i am not tracking that

countvonzero: ok. i focused on your quotes from Tal and missed the malicious behavior.

filed this issue https://github.com/spacemeshos/go-spacemesh/issues/3467 to follow up.
@dshulyak @pigmej

PR #3450, opened 2022-08-16
Remove NodeID from ATX and fix compile issues

### LINKED ISSUE #3126: Broken Test_handleTransaction_BadSignature
https://github.com/spacemeshos/go-spacemesh/runs/5227241447?check_suite_focus=true

```
 --- FAIL: Test_handleTransaction_BadSignature (0.00s)
    logger.go:130: 2022-02-17T04:56:36.697Z	INFO	got new tx%!(EXTRA types.TransactionID=0x3a8460be064d8f6b3fb63bb311681c7e71a2d0f1fc1fdd430f926485e78849bf, log.Field={nonce 18 3  <nil>}, log.Field={amount 18 10  <nil>}, log.Field={fee 18 1  <nil>}, log.Field={gas 18 100  <nil>}, log.Field={recipient 15 0 0x7e41a01100000000000000000000000000000000 <nil>}, log.Field={origin 15 0 0x472333d20a381422928A78B628CAF3eB390e6EBb <nil>})
    handler.go:67: Unexpected call to *mocks.MockconservativeState.AddressExists([[71 35 51 210 10 56 20 34 146 138 120 182 40 202 243 235 57 14 110 187]]) at /home/runner/work/go-spacemesh/go-spacemesh/txs/handler.go:67 because: there are no expected calls of the method "AddressExists" for that receiver
```

Tests expectations are invalid. RandomBytes can produce valid signature, which obviously won't belong to the signer that is generated for the test, but it will allow to recover some address.

### LINKED ISSUE #3216: atx: remove node id field from atx

_Originally posted by @noamnelke in https://github.com/spacemeshos/go-spacemesh/issues/3198#issuecomment-1128749597_

fasmat:
## Motivation
Closes #3216

## Changes
- Removed NodeId from ATX
- Derive NodeId from signature where needed
- Fix compile problems and tests

## Test Plan
- Unit tests via CI

## TODO

## DevOps Notes


fasmat:
Superseded by https://github.com/spacemeshos/go-spacemesh/pull/3454

PR #3454, opened 2022-08-18
Remove NodeID field from ATX

### LINKED ISSUE #3216: atx: remove node id field from atx

fasmat:
## Motivation
Closes #3216

## Changes
- Removed NodeId from ATX
- Derive NodeId from signature where needed
- Fix compile problems and tests

## Test Plan
- Unit tests via CI

## TODO

## DevOps Notes

### DIFF
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -149,13 +149,6 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error
 //   - The ATX view of the previous epoch contains ActiveSetSize activations.
 func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.ActivationTx) error {
 	events.ReportNewActivation(atx)
-	pub, err := ExtractPublicKey(atx)
-	if err != nil {
-		return fmt.Errorf("cannot validate atx sig atx id %v err %v", atx.ShortString(), err)
-	}
-	if !bytes.Equal(atx.NodeID().ToBytes(), pub.Bytes()) {
-		return fmt.Errorf("node ids don't match")
-	}
 
 	if atx.PositioningATX == *types.EmptyATXID {
 		return fmt.Errorf("empty positioning atx")
@@ -212,7 +205,7 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
 		// as expected from the initial Post.
 		initialPostMetadata := *atx.NIPost.PostMetadata
 		initialPostMetadata.Challenge = shared.ZeroChallenge
-		if err := h.nipostValidator.ValidatePost(pub.Bytes(), atx.InitialPost, &initialPostMetadata, uint(atx.NumUnits)); err != nil {
+		if err := h.nipostValidator.ValidatePost(atx.NodeID().ToBytes(), atx.InitialPost, &initialPostMetadata, uint(atx.NumUnits)); err != nil {
 			return fmt.Errorf("invalid initial Post: %v", err)
 		}
 	}
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -120,12 +120,12 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error
 	h.log.WithContext(ctx).With().Info("processing atx",
 		atx.ID(),
 		epoch,
-		log.FieldNamed("atx_node_id", atx.NodeID),
+		log.FieldNamed("atx_node_id", atx.NodeID()),
 		atx.PubLayerID)
 	if err := h.ContextuallyValidateAtx(&atx.ActivationTxHeader); err != nil {
 		h.log.WithContext(ctx).With().Warning("atx failed contextual validation",
 			atx.ID(),
-			log.FieldNamed("atx_node_id", atx.NodeID),
+			log.FieldNamed("atx_node_id", atx.NodeID()),
 			log.Err(err))
 	} else {
 		h.log.WithContext(ctx).With().Info("atx is valid", atx.ID())
@@ -153,7 +153,7 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
 	if err != nil {
 		return fmt.Errorf("cannot validate atx sig atx id %v err %v", atx.ShortString(), err)
 	}
-	if !bytes.Equal(atx.NodeID[:], pub.Bytes()) {
+	if !bytes.Equal(atx.NodeID().ToBytes(), pub.Bytes()) {
 		return fmt.Errorf("node ids don't match")
 	}
 
@@ -167,9 +167,9 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
 			return fmt.Errorf("validation failed: prevATX not found: %v", err)
 		}
 
-		if prevATX.NodeID != atx.NodeID {
+		if prevATX.NodeID() != atx.NodeID() {
 			return fmt.Errorf("previous atx belongs to different miner. atx.ID: %v, atx.NodeID: %v, prevAtx.NodeID: %v",
-				atx.ShortString(), atx.NodeID, prevATX.NodeID)
+				atx.ShortString(), atx.NodeID(), prevATX.NodeID())
 		}
 
 		prevEp := prevATX.PubLayerID.GetEpoch()
@@ -245,7 +245,7 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
 
 	h.log.WithContext(ctx).With().Info("validating nipost", log.String("expected_challenge_hash", expectedChallengeHash.String()), atx.ID())
 
-	pubKey := signing.NewPublicKey(atx.NodeID[:])
+	pubKey := signing.NewPublicKey(atx.NodeID().ToBytes())
 	leaves, err := h.nipostValidator.Validate(*pubKey, atx.NIPost, *expectedChallengeHash, uint(atx.NumUnits))
 	if err != nil {
 		return fmt.Errorf("invalid nipost: %v", err)
@@ -258,10 +258,10 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
 // If a previous ATX is not referenced, it validates that indeed there's no previous known ATX for that miner ID.
 func (h *Handler) ContextuallyValidateAtx(atx *types.ActivationTxHeader) error {
 	if atx.PrevATXID != *types.EmptyATXID {
-		lastAtx, err := atxs.GetLastIDByNodeID(h.cdb, atx.NodeID)
+		lastAtx, err := atxs.GetLastIDByNodeID(h.cdb, atx.NodeID())
 		if err != nil {
 			h.log.With().Error("could not fetch node last atx", atx.ID(),
-				log.FieldNamed("atx_node_id", atx.NodeID),
+				log.FieldNamed("atx_node_id", atx.NodeID()),
 				log.Err(err))
 			return fmt.Errorf("could not fetch node last atx: %w", err)
 		}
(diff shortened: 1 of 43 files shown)


dshulyak:
there is a lot of unrelated changes in this pr

fasmat:
> there is a lot of unrelated changes in this pr

There are a few changes I had to do because of updated dependencies. Do you want me to create a new PR with only those changes and rebase this PR after it is merged?

activation/handler.go
```diff
@@ -153,7 +153,7 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
 	if err != nil {
 		return fmt.Errorf("cannot validate atx sig atx id %v err %v", atx.ShortString(), err)
 	}
-	if bytes.Compare(atx.NodeID[:], pub.Bytes()) != 0 {
+	if !bytes.Equal(atx.NodeID().ToBytes(), pub.Bytes()) {
```
dshulyak: i think we can drop this validation

fasmat:
I removed all unrelated changes and will open a new PR fixing all complaints by `go-staticcheck`

EDIT: new PR for staticcheck fixes can be found here: https://github.com/spacemeshos/go-spacemesh/pull/3471

dshulyak (approved):

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

sql/atxs/atxs.go
```diff
@@ -24,11 +24,15 @@ func Get(db sql.Executor, id types.ATXID) (atx *types.ActivationTx, err error) {
 		tickCount := uint64(stmt.ColumnInt64(2))
 		v.SetID(&id)
 		v.Verify(baseTickHeight, tickCount)
+		buf := make([]byte, len(types.NodeID{}))
```
countvonzero: i think we should be consistent with other places by using `stmt.ColumnLen(3)` here

dshulyak: i haven't noticed but you don't need that at all. 

```go
id := NodeID{}
stmt.ColumnBytes(3, id[:])
```

dshulyak:
the reason why ci fails is because of

```
{"L":"WARN","T":"2022-08-22T09:30:09.367-0400","N":"5b27e.layerFetcher ","M":"failed fetching new ballots","node_id":"5b27e2d0f57c6a81ac16f1d205853e7f742851cab7884bedf81134f7893a0f03","module":"layerFetcher","sessionId":"618772c9-f5db-4404-b9c4-c90e943e8e6f","layer_id":15,"peer":"12D3KooWHgMeRUW2z2Q4x8re92yjef18D539f4JUVsF7AokSvdvG","num_ballots":1,"num_blocks":0,"ballot_ids":["dca568d894"],"errmsg":"syntactic-check ballot: fetch ballots: syntactic-check ballot: fetch referenced ATXs: proposal get ATXs: received syntactically invalid atx 2bd79145d4: previous atx belongs to different miner. atx.ID: 2bd79145d4, atx.NodeID: 214b9a69525c21730a6ac6d9eb1e2a8fee54eac2c6f14f6aacf1c428f042fba0, prevAtx.NodeID: 6a70574f406fb558c013f9b08a92c54751dafc46cc3d852ce1ee15c45192a3b2","name":"layerFetcher"}
```

previous atx should be from the same node. i wasn't able to understand why it started to fail with this change

fasmat:
@countvonzero I cannot reproduce the failing test `TestShortTimeskew` locally. I'll look into what @dshulyak found.

countvonzero:
> the reason why ci fails is because of
> 
> ```
> {"L":"WARN","T":"2022-08-22T09:30:09.367-0400","N":"5b27e.layerFetcher ","M":"failed fetching new ballots","node_id":"5b27e2d0f57c6a81ac16f1d205853e7f742851cab7884bedf81134f7893a0f03","module":"layerFetcher","sessionId":"618772c9-f5db-4404-b9c4-c90e943e8e6f","layer_id":15,"peer":"12D3KooWHgMeRUW2z2Q4x8re92yjef18D539f4JUVsF7AokSvdvG","num_ballots":1,"num_blocks":0,"ballot_ids":["dca568d894"],"errmsg":"syntactic-check ballot: fetch ballots: syntactic-check ballot: fetch referenced ATXs: proposal get ATXs: received syntactically invalid atx 2bd79145d4: previous atx belongs to different miner. atx.ID: 2bd79145d4, atx.NodeID: 214b9a69525c21730a6ac6d9eb1e2a8fee54eac2c6f14f6aacf1c428f042fba0, prevAtx.NodeID: 6a70574f406fb558c013f9b08a92c54751dafc46cc3d852ce1ee15c45192a3b2","name":"layerFetcher"}
> ```
> 
> previous atx should be from the same node. i wasn't able to understand why it started to fail with this change

so i tracked down this particular atx id `2bd79145d4`

```
{"L":"INFO","T":"2022-08-22T09:27:55.157-0400","N":"214b9.atxBuilder   ","M":"atx published","node_id":"214b9a69525c21730a6ac6d9eb1e2a8fee54eac2c6f14f6aacf1c428f042fba0","module":"atxBuilder","event":true,"challenge":"0xf1a2832ae4552766d983c742609263933026264572537de08ac5bf2984904133","id":"2bd79145d4","sender_id":"214b9a69525c21730a6ac6d9eb1e2a8fee54eac2c6f14f6aacf1c428f042fba0","prev_atx_id":"6967a7ce80","pos_atx_id":"ba002e0322","coinbase":"stest1qqqqqqp3qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqql50dsa","pub_layer_id":8,"epoch":2,"num_units":2,"sequence_number":1,"size":239,"name":"atxBuilder"}
```
the previous atx id is `6967a7ce80` and found taht two separate node id published the same atx 

```
{"L":"INFO","T":"2022-08-22T09:26:55.324-0400","N":"214b9.atxBuilder   ","M":"atx published","node_id":"214b9a69525c21730a6ac6d9eb1e2a8fee54eac2c6f14f6aacf1c428f042fba0","module":"atxBuilder","event":true,"nipost":"nonce: 6, indices: \"6008…\"","challenge":"0xc518895430d1ae9dc9aedf4b308ed55b94ab1b948d112a440db84a8992060270","id":"6967a7ce80","sender_id":"214b9a69525c21730a6ac6d9eb1e2a8fee54eac2c6f14f6aacf1c428f042fba0","prev_atx_id":"0000000000","pos_atx_id":"0000000000","coinbase":"stest1qqqqqqp3qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqql50dsa","pub_layer_id":4,"epoch":1,"num_units":2,"sequence_number":0,"size":249,"name":"atxBuilder"}
{"L":"INFO","T":"2022-08-22T09:26:55.075-0400","N":"6a705.atxBuilder   ","M":"atx published","node_id":"6a70574f406fb558c013f9b08a92c54751dafc46cc3d852ce1ee15c45192a3b2","module":"atxBuilder","event":true,"nipost":"nonce: 4, indices: \"6008…\"","challenge":"0xc518895430d1ae9dc9aedf4b308ed55b94ab1b948d112a440db84a8992060270","id":"6967a7ce80","sender_id":"6a70574f406fb558c013f9b08a92c54751dafc46cc3d852ce1ee15c45192a3b2","prev_atx_id":"0000000000","pos_atx_id":"0000000000","coinbase":"stest1qqqqqqp3qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqql50dsa","pub_layer_id":4,"epoch":1,"num_units":2,"sequence_number":0,"size":249,"name":"atxBuilder"}
```

@dshulyak 
@fasmat

btw, the namespace on CI for these logs is `test-zwar`
the queries i used are

{namespace="test-zwar"} | json | module="atxBuilder" | M="atx published" | id="2bd79145d4"
{namespace="test-zwar"} | json | module="atxBuilder" | M="atx published" | id="6967a7ce80"

fasmat:
> the previous atx id is `6967a7ce80` and found taht two separate node id published the same atx

this gives me an idea! Maybe the reason is a similar one as for the other failing test we fixed already? ATX ID calculation used to include the Node ID, but now doesn't any more:

If both ATXs are identical (= reference the same previous ATX and have the same metadata) the ATX IDs are identical even if produced by two different nodes and signed by two different keys.

fasmat:
Specifically I mean here: https://github.com/spacemeshos/go-spacemesh/blob/fix-%233216/common/types/activation.go#L288

and here https://github.com/spacemeshos/go-spacemesh/blob/fix-%233216/common/types/hashes.go#L177

Essentially only these 3 fields are used to derive the ATX ID:

https://github.com/spacemeshos/go-spacemesh/blob/fix-%233216/common/types/activation.go#L104-L106

I'm just not sure if it's just the test that causes these to be the same or if this is actually an oversight that needs fixing.

countvonzero:
> this gives me an idea! Maybe the reason is a similar one as for the other failing test we fixed already? ATX ID calculation used to include the Node ID, but now doesn't any more:

what's wrong to me is that the two nodes should have different challenge. but in the ATX they each build, they had the same challenge `0xc518895430d1ae9dc9aedf4b308ed55b94ab1b948d112a440db84a8992060270`

countvonzero:
> > this gives me an idea! Maybe the reason is a similar one as for the other failing test we fixed already? ATX ID calculation used to include the Node ID, but now doesn't any more:
> 
> what's wrong to me is that the two nodes should have different challenge. but in the ATX they each build, they had the same challenge `0xc518895430d1ae9dc9aedf4b308ed55b94ab1b948d112a440db84a8992060270`

but i am not completely sure. @dshulyak do you know?

countvonzero:
> > > this gives me an idea! Maybe the reason is a similar one as for the other failing test we fixed already? ATX ID calculation used to include the Node ID, but now doesn't any more:
> > 
> > 
> > what's wrong to me is that the two nodes should have different challenge. but in the ATX they each build, they had the same challenge `0xc518895430d1ae9dc9aedf4b308ed55b94ab1b948d112a440db84a8992060270`
> 
> but i am not completely sure. @dshulyak do you know?

regardless, the two axt have different nonce. why do they still hash to the same thing?

fasmat:
> regardless, the two axt have different nonce. why do they still hash to the same thing?

I don't think the nonce is used for ID calculation. Which field encodes the nonce in an ATX?

If it's `ActivationTx.InnerActivationTx.InitialPost.Nonce` or `ActivationTx.InnerActivationTx.NIPost.Post.Nonce` then it is not part of `CalcAndSetID` (only `ActivationTx.InnerActivationTx.ActivationTxHeader` is).

dshulyak:
I found it very surprising that we computed atx id from atx header and not the whole structure (including signature). I guess it worked before (since node id was in the header), but this is really weird optimization. So i think we should change it to compute id using all data.

Regarding post (not post data) - i think it is not bound to the specific id, so thats why challenge might be the same.

fasmat:
bors try

fasmat:
bors try

dshulyak (changes requested):
waiting for @noamnelke to clarify if we should rely on InitialPostIndices to avoid collision, or should we hash signature into atx id

noamnelke:
@fasmat can you please help me validate my suspicion about the cause of the collisions?

I suspect that we have an issue with the difficulty calculation, causing a bias towards small index values. Can you please add a log for the actual indexes and try to find a collision? If you send me the actual colliding indexes I can test my hypothesis.

I just looked at the default parameters and I was a little surprised that `k1` is set to 2000... This means that with a default space unit of 2048, a single space unit smesher will basically only determine which 48 of their 2048 possible indexes **not** to include. The number of possible combinations is still huge, so we still need a bug or other issue to explain this.

I think that we can safely set this value to something MUCH smaller. Even if we set `k1` to 10, with a 256GiB space unit, the space of possibilities is immense (and the size of the proof would be tiny). Anyway, this is a separate discussion, not for right now.

PR #3464, opened 2022-08-19
[Merged by Bors] - vm: txs that can't cover max_gas are failed by consuming the remaining balance

### LINKED ISSUE #3391: vm: transactions that can't cover max_gas

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3391

this change separates fixed gas from max gas.
fixed gas is known in advance for every template/method combination, it is returned by parse_payload call.
max gas additionally includes storage cast, that is computed as a tx length in bytes  multiplied by storage cost factor (vm parameter, defaults to 2 as a placeholder)

if account can't cover fixed gas transaction will be ineffective, since vm can't run verify for free. if account can cover fixed gas, but can't cover max gas tx will be consumed and failed with out of balance. in this case all remaining balance is consumed and paid as a fee to smeshers. in other cases transactions are executed. if transaction is consumed (regardless of the result) block gas limit will be reduced by max gas.


dshulyak:
bors try

common/types/account.go
```diff
@@ -11,26 +11,17 @@ type Account struct {
 	Layer       LayerID
 	Address     Address
 	Initialized bool
-	Nonce       uint64
+	NextNonce   uint64
```
countvonzero, line 14: i prefer this term as well.

genvm/core/context.go
```diff
@@ -121,13 +126,13 @@ func (c *Context) Consumed() uint64 {
...
+	return c.fee
 }
 
 // Updated list of addresses.
 func (c *Context) Updated() []types.Address {
-	rst := make([]types.Address, len(c.touched)+1)
+	rst := make([]types.Address, 0, len(c.touched)+1)
 	rst = append(rst, c.Account.Address)
-	copy(rst[1:], c.touched)
+	rst = append(rst, c.touched...)
```
countvonzero, line 136: curious about why you made a impl change here? is it performance or memory related?

genvm/core/gas.go
```diff
@@ -0,0 +1,6 @@
+package core
+
+// ComputeGasCost computes gas cost from fixed gas plus tx size multiplied by storage cost.
+func ComputeGasCost(fixedGas uint64, tx []byte, storageFactor uint64) uint64 {
+	return fixedGas + uint64(len(tx))*storageFactor
```
countvonzero, line 5: this means each byte will cost `storageFactor` gas uinit?

genvm/core/types.go
```diff
@@ -58,3 +58,11 @@ type AccountLoader interface {
 type AccountUpdater interface {
 	Update(Account) error
 }
+
+// ParseOutput is a tuple containing all fields that are returned
+// by ParsePayload call.
```
countvonzero: :s/ParsePayload/Parse/

genvm/rewards.go
```diff
@@ -11,15 +11,17 @@ import (
 
 // Config defines the configuration options for Spacemesh rewards.
 type Config struct {
-	GasLimit   uint64
-	BaseReward uint64 `mapstructure:"base-reward"`
+	GasLimit          uint64
```
countvonzero, line 14: didn't notice this before, should GasLimit also have a json tag?

genvm/vm.go
```diff
@@ -266,6 +266,7 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.Transa
 		limit       = v.cfg.GasLimit
 	)
 	for i := range txs {
+		logger := v.logger.WithFields(log.Int("i", i))
```
countvonzero: name it "ith" or "i-th" for the sake of easier search?

genvm/vm.go
```diff
@@ -291,18 +292,18 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.Transa
 		}
 		ctx := req.ctx
 		args := req.args
-
-		if ctx.Account.Balance < ctx.Header.MaxGas*ctx.Header.GasPrice {
-			v.logger.With().Warning("ineffective transaction. can't cover gas and max spend",
+		if ctx.Account.Balance < ctx.ParseOutput.FixedGas*ctx.ParseOutput.GasPrice {
```
countvonzero, line 295: why isn't this compared to ctx.Header.MaxGas * ctx.ParseOutput.GasPrice?
i.e. we should also consider if the principal can afford the storage cost too

countvonzero (approved):

genvm/vm.go
dshulyak: > if account can cover fixed gas, but can't cover max gas tx will be consumed and failed with out of balance.

this is the other part of this pr. the gas is consumed in ctx.Consume before ctx.Exec

genvm/rewards.go
dshulyak: i don't know. the reason why i didn't add a tag for GasLimit is because mempool and vm should use same parameter

genvm/core/gas.go
dshulyak: yes

genvm/core/context.go
dshulyak: neither) i just noticed that i made a mistake
this way it is harder to mess up what gets copied into the source array

genvm/vm.go
countvonzero: right. so checking it here confuses me. 

theoretically `ctx.Account.Balance < ctx.ParseOutput.FixedGas*ctx.ParseOutput.GasPrice` cannot be true because the code already checks if the balance can cover max_gas*gas_price

dshulyak: it can be true. fixed gas is lower than max gas. this check is *before* doing Consume

if account can't cover fixed gas - tx ineffective
if account can cover [fixed gas, max gas) - tx will be failed and remaining balance is consumed
if account - [max gas, ...) - tx is executed

countvonzero: yikes. i confused execute() with handler.Exec(). sorry. make sense. thanks

dshulyak:
bors merge

PR #3465, opened 2022-08-19
[Merged by Bors] - Added wait to systests cleaning process

pigmej:
## Motivation

It was reported today that sometimes we have flaky systests because of resource intensive items that are not yet cleaned.

That should address that issue.

## Changes

Added kubectl wait after ns and pod deletions.

## TODO

## DevOps Notes

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -41,10 +41,13 @@ run:
 clean:
 	@echo "deleting test pods with testid=$(test_id)"
 	@kubectl delete pod --selector=testid=$(test_id)
+	@kubectl wait --for delete pod --selector=testid=$(test_id)
 	@echo "deleting ephemeral namespaces with testid=$(test_id)"
 	@kubectl delete ns --selector=testid=$(test_id),keep=false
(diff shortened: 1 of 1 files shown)


dshulyak (approved):
not completely sure that it will cover the case if k8s api is temporarily unavailable

pigmej:
> not completely sure that it will cover the case if k8s api is temporarily unavailable

It will *not* cover the case directly. My hope is that these 3s wait will allow us to for sure cleanup the NS and then API will not crash because of resources.

pigmej:
@dshulyak Actually, let me raise the timeout to 60s. :)

dshulyak:
@pigmej do you want to merge this change?

pigmej:
That's a good question. I think it could help fox some race conditions, it will not help with app unavailability. For that i can make separate script "wait for API" or STH and plug before these. 

But I think wait for delete timeout should prevent api crashes too, as it should not run out of memory.

pigmej:
bors try

dshulyak:
bors try

pigmej:
bors try

pigmej:
bors merge

pigmej:
bors merge

PR #3469, opened 2022-08-21
[Merged by Bors] - vm: add nonce to self-spawn

### LINKED ISSUE #3392: vm: include nonce in self-spawn txs
nonce was implicitly decoded by parse_payload of the template as 0. it leads to hairy edge cases if self-spawn fails but got consumed.

also enables https://github.com/spacemeshos/go-spacemesh/issues/3339

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3392

self-spawn transaction can fail if it was included into the block and account can't cover
gas required for it to be executed (in general it is not possible if all assumptions are satisfied)

however if it does happen there should be a way to retry self-spawning account, so it will have a nonce
as any other transaction. 

in system tests we don't expect self-spawn to fail, so in that case self-spawn submitted always with zero nonce.
if it fails it is definitely a bug and needs to be fixed.


genvm/templates/multisig/handler.go
```diff
@@ -86,23 +86,24 @@ func (h *handler) Parse(ctx *core.Context, method uint8, decoder *scale.Decoder)
 			return
 		}
 		args = &p.Arguments
-		header.GasPrice = p.GasPrice
-		header.MaxGas = h.totalGasSpawn
+		out.GasPrice = p.GasPrice
+		out.Nonce = p.Nonce
```
countvonzero, line 90: this reminds me, where does the code check whether this account is already spawned?

countvonzero (commented):
i need to review this after the base PR is merged.

genvm/templates/multisig/handler.go
dshulyak: https://github.com/spacemeshos/go-spacemesh/blob/c7b97f0a2f586f303d4a27211841279da083c988/genvm/core/context.go#L47-L50

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #3470, opened 2022-08-21
[Merged by Bors] - vm: support regular spawn

### LINKED ISSUE #3339: vm: regular spawn is not supported
we don't support regular spawn in the current implementation.

as i understand there is big difference in how payload is encoded for spawn and self spawn:
- for self spawn: nonce is not included into payload. zeroes are returned by parse_payload
- for regular spawn: nonce needs to be included into payload, as it is a regular transaction from another principal

this difference is not reflected in the existing spec, so it might be that i don't understand something and instead skipped regular spawn altogether.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3339

this change enables spawned account to pay gas for spawning other stub accounts.

tx is ineffective if it tries to spawn target account while its own account is still not spawned.
if target account was already spawned transaction will be consumed, paying max gas and using a nonce.

the format of the spawn matches self-spawn format:
both for spawn and self-spawn the format is `scale(tx_version, principal, method_spawn, template, envelope, arguments, signature)`
arguments should be decodable into concrete type by a template specified by transaction. signature should be verifiable by a principal template.

for example if single sig spawns multi sig, arguments should include a collection with at least K ed25519 public keys, but signature is a single ed25519. and vice-versa if multi sig spawns single sig.


dshulyak:
bors try

genvm/vm.go
```diff
@@ -455,32 +455,70 @@ func parse(logger log.Log, reg *registry.Registry, loader core.AccountLoader, st
...
 	logger.With().Debug("loaded account state", log.Inline(&account))
 
-	var template *core.Address
+	ctx := &core.Context{
+		Registry: reg,
+		Loader:   loader,
+		Account:  account,
+	}
+
+	if account.Template != nil {
```
countvonzero: i am very confused by the use of "Template". sometimes it's an address, sometimes it's a Template.
can you perhaps rename to TemplateAddr or something for those that are addresses?

genvm/vm.go
```diff
@@ -455,32 +455,70 @@ func parse(logger log.Log, reg *registry.Registry, loader core.AccountLoader, st
...
+		if err != nil {
+			return nil, nil, nil, err
+		}
+	}
+
+	var (
+		template *core.Address
+		handler  core.Handler
+	)
 	if method == 0 {
```
countvonzero, line 479: isn't there a named constant MethodSpawn you can use? if not, maybe just adding comment that this is a spawn

genvm/vm.go
countvonzero: it's hard to understand when i need to think what does `account.Template != nil` or `ctx.Handler==nil` mean in different context (spawn/self-spawn/spend).

what do you think about breaking this block in the following structure?
maybe there will be some repeated code, but those can be factored out.
```
if method == spawn {
    if toSpawn == principal {  // self-spwan
    } else {  // regular spawn
    }
} else { // spend
}
```

genvm/core/context.go
```diff
@@ -38,16 +39,40 @@ type Context struct {
...
 
 // Spawn account.
-func (c *Context) Spawn(template Address, args scale.Encodable) error {
-	principal := ComputePrincipal(template, args)
-	// TODO(dshulyak) only self-spawn is supported
-	if principal != c.Header.Principal {
-		return ErrSpawn
+func (c *Context) Spawn(args scale.Encodable) error {
+	template := c.Header.Template
+	address := ComputePrincipal(template, args)
```
countvonzero: can you rename address/account to spawned/spawnedAcct for easier differentiation with c.Account.Address

genvm/templates/wallet/handler.go
```diff
@@ -40,39 +40,39 @@ var (
...
 			err = fmt.Errorf("%w: %s", core.ErrMalformed, err.Error())
 			return
 		}
-		args = &p.Arguments
-		out.GasPrice = p.GasPrice
-		out.Nonce = p.Nonce
-		out.FixedGas = TotalGasSpend
+		output.GasPrice = p.GasPrice
+		output.Nonce = p.Nonce
+		output.FixedGas = TotalGasSpend
```
countvonzero, line 49: these two cases are exactly the same now.

countvonzero (approved):
i mostly found that code changes for single/multi sigs are in parallel.

dshulyak:
bors merge

dshulyak:
bors merge

PR #3473, opened 2022-08-22
3348 genesis config

### LINKED ISSUE #3348: Split config into genesis config and mutable config
## Description
Right now, we have a single config file (`config.toml` by default) containing a number of sections. Some of these sections contain data that the node operator can update at-will, e.g., NTP server list, API config, and `logging` section. Some sections should not be updated unilaterally by a node operator but could conceivably change in a coordinated fashion post-genesis, e.g., HARE config. Other sections contain things that can never change post-genesis, e.g., `main.genesis-time` and `main.golden-atx`, and the entire `genesis` section (containing genesis accounts and balances). Some of the `beacon` and `tortoise` sections of the config should probably not be able to change after genesis, we need to double check this. (For now, we'll leave alone the second class of config items, those that should not be updated unilaterally; we may want to revisit this in future.)

The task: split out immutable, genesis-related config items from the main `BaseConfig` object and move them into the `Genesis` config object. Split this from `config.toml` and move it into a new standalone `genesis.toml` file.

Every time the node starts, check that the genesis config has not changed (and panic/refuse to start if it has): this may involve saving a hash of the config to the database, something along these lines.

## Affected code
- api/config/genesis.go
- cmd/node/node.go
- config/config.go
- config.toml

See also:
- https://github.com/spacemeshos/go-spacemesh/pull/2581
- https://github.com/spacemeshos/go-spacemesh/issues/2004
- https://github.com/spacemeshos/pm/issues/135

This issue appears in commit hash: 5eb9998e5ea11667be53d8a2bd2e8d634c23c463

WilfredTA:
## Motivation
Closes #3348, #3503, #3504, #3505, #3506, #3507 

## Changes
- `[genesis]` section in config.toml that is enforced to not change.
- Hash of genesis section in config.toml is stored in a `genesis.conf` file
- GenesisID is calculated from `genesis-time` and `genesis-extradata` values taken from `[genesis]` section of config.toml
- P2P.NetworkId is set to the Uint32(genesisId)
- Upon startup of node, genesis configuration is loaded from the config file. The node then checks if a `genesis.conf` file exists. If it does, the node compares `genesis.conf` contents to the genesis information in the config file. If they are unequal, the node returns an error. If the genesis.conf file does not exist, the node creates the file in readonly mode
- The genvm SDK provides an option to be initialized with genesis id: `WithGenesisId`
- The VM can be initialized with Genesis ID as well
- A lightweight wrapper around EdSigner and EDVerifier has been added, called `GenesisBoundEdSigner` and `GenesisBoundEdVerifier` which *prepends* the genesis Id to the signed message or checks that the first 20 bytes of the message is the correct genesis id, respectively.
- GoldenATX is set to Genesis ID and then zero-padded to be type-compatible with the ATX type (GenesisID is 20 bytes, but ATX ID is defined as 32-byte slice)

## Test Plan
- [x] Test Genesis config changes
- [ ]  Test Signer changes
- [x]  Test VM changes
- [ ]  Update e2e tests to reflect changes introduced by this PR

## TODO
- [x] Complete issue #3505 
- [x] Add modified ED signer & verifier
- [x] Remove P2P Network Id & set to genesis id
- [x] Add genesis ID to VM and SDK
- [x] Modify GenesisBoundEd[Signer|Verifier] to treat genesis ID as prepended (rather than appended) to msg per #3503
- [x] Modify Verify method in templates per issue #3504 
- [ ] Address bullet points 2 and 3 in #3506 
- [x] Complete #3507 

## DevOps Notes


dshulyak:
this change needs to be splitted before review

cmd/node/node.go
```diff
@@ -457,7 +458,7 @@ func (app *App) initServices(ctx context.Context,
 		}),
 		txs.WithLogger(app.addLogger(ConStateLogger, lg)))
 
-	genesisAccts := app.Config.Genesis.ToAccounts()
+	genesisAccts := app.Config.Genesis.Accounts.ToAccounts()
```
evt, line 461: Thinking if `app.Config.Genesis.Accounts.ToList()` makes more sense here as you convert accounts map to a list here 🤔

config/genesis.go
```diff
@@ -0,0 +1,214 @@
...
+		digest := hasher.Sum([]byte{})
+		return types.BytesToHash(digest).ToHash20(), nil
+	} else {
+		hasher := hash.New()
+		hasher.Write(genesis_time_b)
+		hasher.Write(genesis_extra_data)
+		// full_b := make([]byte, len(genesis_time_b)+len(genesis_extra_data))
+		// copy(full_b[:], genesis_time_b)
+		// copy(full_b[len(genesis_time_b):], genesis_extra_data)
+		// hasher.Write(full_b)
```
evt, line 185: do we still need this code?

config/genesis.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (cfg *GenesisConfig) CalcGenesisID() {
+	gen_id, _ := CalcGenesisId([]byte(cfg.ExtraData), cfg.GenesisTime)
+
+	cfg.GenesisID = gen_id.Hex()
+}
+
+func CalcGoldenATX(genesis_data []byte, genesis_time string) (types.Hash32, error) {
+	if id, err := CalcGenesisId([]byte(genesis_data), genesis_time); err == nil {
```
evt, line 209: `[]byte(genesis_data)` => `genesis_data`

config/genesis.go
```diff
@@ -0,0 +1,214 @@
...
+		return id.ToHash32(), nil
```
evt, line 210: why doesn't `CalcGenesisId` return `Hash32`? 🤔

cmd/node/node.go
```diff
@@ -1088,6 +1091,40 @@ func (app *App) Start() error {
...
+			// Write to readonly file
+			if writeErr := ioutil.WriteFile(config.DefaultGenesisDataPath, json_cfg, 0444); writeErr != nil {
+				return fmt.Errorf("failed to save genesis config with data: %s to %s.\nGot error: %s",
+					string(json_cfg), config.DefaultGenesisDataPath, writeErr)
+			}
+		}
+		// Persist it
+	}
+	// Note to self: Configuration should be finished after this
+	// Panic should be before this
```
evt, line 1127: Looks like a good candidate for a separate function :)

config/genesis.go
evt: Found why in https://github.com/spacemeshos/SMIPS/issues/75, please skip it.

dshulyak:
implemented elsewhere

PR #3477, opened 2022-08-23
[Merged by Bors] - beacon: fix test that expects corrupted signature and some times fails

dshulyak:
the message is not necessarily corrupted without first byte


countvonzero (approved):

dshulyak:
bors merge

PR #3478, opened 2022-08-23
[Merged by Bors] - Enable scale codec for hare builder

evt:
## Motivation
Part of #3271

## Changes
enable scale codec for hare builder

## DevOps Notes


evt:
bors try

evt:
bors try

hare/builder.go
```diff
@@ -10,17 +10,19 @@ import (
 	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
+//go:generate scalegen
+
 // Message is the tuple of a message and its corresponding signature.
 type Message struct {
 	Sig      []byte
-	InnerMsg *innerMessage
+	InnerMsg *InnerMessage
```
dshulyak, line 18: i think this should not be an option

hare/builder.go
```diff
@@ -41,120 +43,132 @@ func (m *Message) Field() log.Field {
...
 	msg   *Msg
-	inner *innerMessage
+	inner *InnerMessage
 }
 
 // newMessageBuilder returns a new, empty message builder.
 // One should not assume any values are pre-set.
-func newMessageBuilder() *messageBuilder {
-	m := &messageBuilder{&Msg{Message: &Message{}, PubKey: nil}, &innerMessage{}}
+func newMessageBuilder() *MessageBuilder {
```
dshulyak: definitely shouldn't generate scale codec for MessageBuilder

dshulyak (changes requested):

dshulyak:
please fix two things, and i think we can merge it

hare/builder.go
evt: done, cheers

hare/builder.go
evt: if I make it a struct rather than a pointer then how would I check `InnerMessage` is not empty? I mean what field shall I look at?

hare/builder.go
```diff
@@ -10,17 +10,19 @@ import (
 	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
+//go:generate scalegen -types Message,Certificate,AggregatedMessages,InnerMessage
+
 // Message is the tuple of a message and its corresponding signature.
 type Message struct {
 	Sig      []byte
-	InnerMsg *innerMessage
+	InnerMsg *InnerMessage
```
dshulyak, line 18: i meant that it should not be a pointer, so that it won't be encoded as an option

evt, line 18: I mean:
```
	if m.Message.InnerMsg == nil {
		logger.Error("onEarlyMessage called with nil message")
		return
	}

	if m.InnerMsg == nil {
		logger.Error("onEarlyMessage called with nil inner message")
		return
	}
```
what shall I replace it with if `InnerMsg` is not a pointer anymore?

hare/builder.go
dshulyak: what do you mean? it can't be nil if it won't be a pointer

hare/builder.go
dshulyak: that code can be removed

dshulyak (approved):
nevermind, no reason to refactor previous mistakes in this pr

evt:
bors try

evt:
bors merge

PR #3494, opened 2022-08-25
[Merged by Bors] - tortoise: ensure that blocks and ballots are processed after ref height was computed

dshulyak:
reference height is computed when the first layer of the epoch is terminated
before that all ballots and blocks that are passed to OnBlock/OnBallot should be ignored, 
processing them will lead to inconsistent state. namely ballots wont have votes for all blocks in the previous layers 

if there a gap in the layer that were sent to tortoise (due to the failed hare), tortoise waits only for zdist until moving on.
once it decided to move on it will download all relevant data (blocks and ballots) after ref height was computed.

### DIFF
--- a/tortoise/full.go
+++ b/tortoise/full.go
@@ -171,7 +171,8 @@ func (f *full) verify(logger log.Log, lid types.LayerID) bool {
 		return false
 	}
 	isEmpty := empty.Cmp(f.globalThreshold) > 0
-	if len(f.blocks) == 0 {
+	blocks := f.blocks[lid]
+	if len(blocks) == 0 {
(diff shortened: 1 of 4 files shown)


dshulyak:
bors try

dshulyak:
bors try

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors merge

PR #3496, opened 2022-08-25
[Merged by Bors] - Disable XDR in codec + fixed applying scale codec in beacon and fetch

evt:
## Motivation
Part of #3271

## Changes
disable XDR in codec + fixed applying scale codec in beacon and fetch

## DevOps Notes


beacon/beacon.go
```diff
@@ -922,7 +937,15 @@ func buildSignedProposal(ctx context.Context, signer signing.Signer, epoch types
 	return signature
 }
 
-func signMessage(signer signing.Signer, message interface{}, logger log.Log) []byte {
+func signFirstVotingMessage(signer signing.Signer, message *FirstVotingMessageBody, logger log.Log) []byte {
```
countvonzero: what's the diff btwn signFirstVotingMessage and signFollowingVotingMessage?

beacon/beacon.go
```diff
@@ -795,7 +800,12 @@ func (pd *ProtocolDriver) sendFirstRoundVote(ctx context.Context, epoch types.Ep
 		types.FirstRound,
 		log.String("message", m.String()))
 
-	pd.sendToGossip(ctx, pubsub.BeaconFirstVotesProtocol, m)
+	serialized, err := codec.Encode(&m)
```
countvonzero, line 808: do you want to move 803-806 into pd.sendToGossip()? the code here and below are repetitive

beacon/beacon_test.go
```diff
@@ -730,41 +730,6 @@ func TestBeacon_buildProposal(t *testing.T) {
 	}
 }
 
-func TestBeacon_signMessage(t *testing.T) {
```
countvonzero, line 733: why rm this test?

beacon/beacon.go
evt: they accept different message types. Scale codec doesn't work with `interface{}`. Alternatively we can move `codec.Encode` to the caller side but then only `signer.Sign` is left and no point in a function :)

beacon/beacon.go
evt: nope, I don't :) scale codec doesn't work with `interface{}`

beacon/beacon_test.go
evt: gonna add separate tests for different message types. scale codec doesn't work with `interface{}`

beacon/beacon.go
countvonzero: right. in that case maybe just get rid of signXXXMessage all together?

beacon/beacon_test.go
countvonzero: i am not sure this test is meaningful anyway....

evt: true, I'll think if I can make it :)

beacon/beacon.go
evt: and done

evt:
bors try

beacon/beacon.go
dshulyak: you could replace interface{} with codec.Encodable (which should be an alias for scale.Encodable)

codec/codec.go
```diff
@@ -6,46 +6,27 @@ import (
...
 
-func init() {
-	// xdr will fail with overflow if slice size is larger than 1mb
-	// see BenchmarkInvalidLength
-	xdr.SliceLimit = 1 << 20
-}
-
 // Encodable is an interface that must be implemented by a struct to be encoded.
-type Encodable interface{}
+type Encodable interface {
```
dshulyak: maybe Encodable = scale.Encodable

dshulyak (approved):

evt:
bors try

evt:
bors merge

evt:
bors merge

PR #3497, opened 2022-08-26
[Merged by Bors] - certifier: accept early certify message

### LINKED ISSUE #3475: hare certification: sync flow is not working yet tests passed
## Description

two issues in the implementation:
- early gossip messages are not handled properly, causing the following two logs
```
{"L":"ERROR","T":"2022-08-22T10:51:10.345-0400","N":"3dd22.blockCert    ","M":"failed to process certify message gossip","node_id":"3dd225dd843c54f9e711d71db4d013f283aca653f55632a1293f393b7cd6d8dc","module":"blockCert","requestId":"d47b5112-7630-49c8-b772-776accb6d6e8","errmsg":"unexpected lid/bid","name":"blockCert"}

{"L":"WARN","T":"2022-08-22T10:51:10.274-0400","N":"ec05f.blockCert    ","M":"layer not registered for cert","node_id":"ec05f85c6f081b7dcf0842a855aa37add882bfcaf8b94ee7463d2f17c2f6f6e1","module":"blockCert","requestId":"2ee0f936-bd18-4751-923f-1f148b52c2fc","layer_id":11,"block_id":"65a664148b"}
```

- sync'ed certificate is not actually verified, due to the layer/block check
```
{"L":"WARN","T":"2022-08-22T09:30:01.659-0400","N":"3d88b.layerFetcher ","M":"failed to handle certificate","node_id":"3d88bc875aabe5c1b59803b5d814d22886177bf5434693f99e670da07ca053e6","module":"layerFetcher","sessionId":"76ec9ad8-472e-4a7c-baa4-2c3ea7585b2a","layer_id":15,"peer":"12D3KooWSaj7fTVv33Be6BbXUdQqxxbuBXy9qA8VToQfq4EmE1qz","errmsg":"cert msg too old","name":"layerFetcher"}
```
another important question: why do all system tests passed?

countvonzero:
## Motivation

Closes #3475

## Changes

- block certifier accept early CertifyMessage
- syncer: 
  - separate fetching layer data and layer opinions (only certificate for now)
  - only fetch certificates from peers within hdist layers of processed layer. processed layer is tortoise's last layer
- fix sql bug where certificate is retrieved correctly and not overwritten once set
- systest
  - fixed systests TestAddNode so that the test fails with the certificate sync failures
  - make systests TestFailedNodes to reassess layer hashes from nodes. because blocks are applied optimistically, the applied block can be changed after tortoise verify a layer

### DIFF
--- a/blocks/certifier.go
+++ b/blocks/certifier.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"sync"
-	"time"
 
 	"github.com/spacemeshos/ed25519"
 	"golang.org/x/sync/errgroup"
@@ -23,6 +22,8 @@ import (
 	"github.com/spacemeshos/go-spacemesh/system"
 )
 
+const numEarlyLayers = uint32(1)
+
 var (
 	errInvalidCert        = errors.New("invalid certificate")
 	errInvalidCertMsg     = errors.New("invalid cert msg")
@@ -33,18 +34,16 @@ var (
 
 // CertConfig is the config for Certifier.
 type CertConfig struct {
-	CommitteeSize         int
-	CertifyThreshold      int
-	SignatureWaitDuration time.Duration
-	NumLayersToKeep       uint32
+	CommitteeSize    int
+	CertifyThreshold int
+	NumLayersToKeep  uint32
 }
 
 func defaultCertConfig() CertConfig {
 	return CertConfig{
-		CommitteeSize:         10,
-		CertifyThreshold:      6,
-		SignatureWaitDuration: time.Second * 10,
-		NumLayersToKeep:       5,
+		CommitteeSize:    10,
+		CertifyThreshold: 6,
+		NumLayersToKeep:  5,
 	}
 }
 
@@ -73,21 +72,10 @@ func WithCertifierLogger(logger log.Log) CertifierOpt {
 }
 
 type certInfo struct {
-	bid              types.BlockID
-	deadline         time.Time
-	done             bool
+	registered, done bool
 	totalEligibility uint16
 	signatures       []types.CertifyMessage
 }
-
-type certState int
-
-const (
-	unknown certState = iota
-	early
-	pending
-	certified
-)
 
 // Certifier collects enough CertifyMessage for a given hare output and generate certificate.
 type Certifier struct {
@@ -107,8 +95,7 @@ type Certifier struct {
 	beacon     system.BeaconGetter
 
 	mu          sync.Mutex
-	certifyMsgs map[types.LayerID]*certInfo
-	buffer      map[types.LayerID][]*types.CertifyMessage
+	certifyMsgs map[types.LayerID]map[types.BlockID]*certInfo
 }
 
 // NewCertifier creates new block certifier.
@@ -127,8 +114,7 @@ func NewCertifier(
 		publisher:   p,
 		layerClock:  lc,
 		beacon:      b,
-		certifyMsgs: make(map[types.LayerID]*certInfo),
-		buffer:      make(map[types.LayerID][]*types.CertifyMessage),
+		certifyMsgs: make(map[types.LayerID]map[types.BlockID]*certInfo),
 	}
 	for _, opt := range opts {
 		opt(c)
@@ -168,33 +154,51 @@ func (c *Certifier) run() error {
 	for layer := c.layerClock.GetCurrentLayer(); ; layer = layer.Add(1) {
 		select {
 		case <-c.layerClock.AwaitLayer(layer):
-			_ = c.genCertificate()
+			c.prune()
 		case <-c.ctx.Done():
 			return fmt.Errorf("context done: %w", c.ctx.Err())
 		}
 	}
 }
 
-// RegisterDeadline register the deadline to stop waiting for enough signatures to generate
-// the certificate.
-func (c *Certifier) RegisterDeadline(ctx context.Context, lid types.LayerID, bid types.BlockID, now time.Time) error {
-	logger := c.logger.WithContext(ctx).WithFields(lid, bid)
-	logger.Info("certifier for layer registered")
+func (c *Certifier) prune() {
 	c.mu.Lock()
 	defer c.mu.Unlock()
+
+	cutoff := types.GetEffectiveGenesis()
+	current := c.layerClock.GetCurrentLayer()
+	if current.Uint32() > c.cfg.NumLayersToKeep {
+		cutoff = c.layerClock.GetCurrentLayer().Sub(c.cfg.NumLayersToKeep)
+	}
+	for lid := range c.certifyMsgs {
+		if lid.Before(cutoff) {
+			c.logger.With().Debug("removing layer from cert registration", lid)
+			delete(c.certifyMsgs, lid)
+		}
+	}
+}
+
+func (c *Certifier) createIfNeeded(lid types.LayerID, bid types.BlockID) {
 	if _, ok := c.certifyMsgs[lid]; !ok {
-		c.certifyMsgs[lid] = &certInfo{
-			bid:        bid,
-			deadline:   now.Add(c.cfg.SignatureWaitDuration),
+		c.certifyMsgs[lid] = make(map[types.BlockID]*certInfo)
+	}
+	if _, ok := c.certifyMsgs[lid][bid]; !ok {
+		c.certifyMsgs[lid][bid] = &certInfo{
 			signatures: make([]types.CertifyMessage, 0, c.cfg.CommitteeSize),
 		}
 	}
-	for _, msg := range c.buffer[lid] {
-		if err := c.saveMessageLocked(logger, *msg); err != nil {
-			return err
-		}
-	}
-	delete(c.buffer, lid)
+}
+
+// RegisterForCert register to generate a certificate for the specified layer/block.
+func (c *Certifier) RegisterForCert(ctx context.Context, lid types.LayerID, bid types.BlockID) error {
+	logger := c.logger.WithContext(ctx).WithFields(lid, bid)
+	logger.Info("certifier registered")
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.createIfNeeded(lid, bid)
+	c.certifyMsgs[lid][bid].registered = true
+	c.tryGenCert(logger, lid, bid)
 	return nil
 }
 
@@ -266,35 +270,6 @@ func (c *Certifier) NumCached() int {
 	return len(c.certifyMsgs)
 }
 
-func (c *Certifier) getCertState(logger log.Log, lid types.LayerID, bid types.BlockID) (certState, error) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	return c.getCertStateLocked(logger, lid, bid)
-}
-
-func (c *Certifier) getCertStateLocked(logger log.Log, lid types.LayerID, bid types.BlockID) (certState, error) {
-	if ci, ok := c.certifyMsgs[lid]; !ok {
-		current := c.layerClock.GetCurrentLayer()
-		// only accept early msgs within a range and with limited size to prevent DOS
-		if !lid.Before(current) && lid.Before(current.Add(c.cfg.NumLayersToKeep)) {
-			if c.buffer[lid] == nil || len(c.buffer[lid]) < c.cfg.CommitteeSize {
-				logger.Debug("received early certify message")
-				return early, nil
-			}
-		}
-		logger.Warning("layer not registered for cert")
-		return unknown, errUnexpectedMsg
-	} else if bid != ci.bid {
-		logger.With().Warning("block not registered for cert", log.Stringer("cert_block_id", ci.bid))
-		return unknown, errUnexpectedMsg
-	}
-
-	if c.certifyMsgs[lid].done {
-		return certified, nil
-	}
-	return pending, nil
-}
-
 // HandleSyncedCertificate handles Certificate from sync.
 func (c *Certifier) HandleSyncedCertificate(ctx context.Context, lid types.LayerID, cert *types.Certificate) error {
 	logger := c.logger.WithContext(ctx).WithFields(lid, cert.BlockID)
@@ -317,10 +292,22 @@ func (c *Certifier) HandleSyncedCertificate(ctx context.Context, lid types.Layer
 		return err
 	}
 
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	delete(c.buffer, lid)
 	return nil
+}
+
+func (c *Certifier) certified(lid types.LayerID, bid types.BlockID) bool {
+	if _, ok := c.certifyMsgs[lid]; ok {
+		if _, ok := c.certifyMsgs[lid][bid]; ok {
+			return c.certifyMsgs[lid][bid].done
+		}
+	}
+	return false
+}
+
+func expectedLayer(clock layerClock, lid types.LayerID) bool {
+	current := clock.GetCurrentLayer()
+	// only accept early msgs within a range and with limited size to prevent DOS
+	return !lid.Before(current) && !lid.After(current.Add(numEarlyLayers))
 }
 
 func (c *Certifier) handleRawCertifyMsg(ctx context.Context, data []byte) error {
@@ -337,9 +324,13 @@ func (c *Certifier) handleRawCertifyMsg(ctx context.Context, data []byte) error 
 	if _, err := c.beacon.GetBeacon(lid.GetEpoch()); err != nil {
 		return errBeaconNotAvailable
 	}
-	if state, err := c.getCertState(logger, lid, bid); err != nil {
-		return err
-	} else if state == certified {
+
+	if !expectedLayer(c.layerClock, lid) {
+		logger.Debug("received message for unexpected layer")
+		return errUnexpectedMsg
+	}
+
+	if c.certified(lid, bid) {
 		// should still gossip this msg to peers even when this node has created a certificate
 		return nil
 	}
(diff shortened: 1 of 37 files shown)


dshulyak:
> block certifier tries to wait for all CertifyMessage. the behavior now is a background thread generating certificate / pruning messages when every layer ticks. this can be reduced to smaller interval but i don't find it necessary. the reason to wait for all messages is that nodes in the network have different epoch weights. when node A finds a msg valid, node B may disagree. since these certificates are going to be pruned, size is not an issue. therefore we want do make best effort for as many nodes to generate a certificate, or accept one from peers, by having as many CertifyMessage in each Certificate.

but that is not relevant if someone disagrees. cert message is either valid according to your view or invalid, so once there are valid msgs with total weight > 400, whats the point to keep adding them?

i also think that it creates problems for p2p network, since anybody can broadcast invalid data and every peer will re-broadcast it. the only good option is to ignore certificates that are invalid according to your view

dshulyak:
> block certifier accept early CertifyMessage and put them to a cache until the node's block is generatede.

eligibility can be validated even if message is early

hare/hare.go
```diff
@@ -221,17 +221,16 @@ func (h *Hare) collectOutput(ctx context.Context, output TerminationOutput) erro
 		postNumProposals.Add(float64(set.len()))
 		pids = make([]types.ProposalID, 0, set.len())
 		pids = append(pids, set.elements()...)
+		h.blockGenCh <- LayerOutput{
```
dshulyak: btw i didn't notice this before but it is safer to

```
select {
  case h.blockGen <- ...:
  case ctx.Done():
}
```

otherwise it is very common to get stuck in such cases when node is being shut down

mesh/mesh.go
```diff
@@ -55,17 +55,26 @@ type Mesh struct {
 	// if contextual validity changed for blocks in that layer, we need to
 	// double-check whether we have applied the correct block for that layer.
 	minUpdatedLayer atomic.Value
+
+	// used for caching hare certificate when the node sync data from peers.
+	// a synced certificate cannot be verified before the node learns the beacon value
+	// of the epoch. any certificate passed on by the syncer will stay here until
+	// beacon is available.
+	certCache    map[types.LayerID][]*types.Certificate
```
dshulyak: i don't think that it is ok to add more mutable state randomly across the code. is there a way to refactor it in a better way?

can syncer download ballots and blocks before downloading opinions on them?

systest/cluster/cluster.go
```diff
@@ -309,10 +310,15 @@ func (c *Cluster) AddSmeshers(cctx *testcontext.Context, n int) error {
 		return err
 	}
 	bootnodes := c.clients[:c.bootnodes]
+	smeshers := c.clients[c.bootnodes:]
 	c.clients = nil
+	smeshers = append(smeshers, clients...)
+	sort.Slice(smeshers, func(i, j int) bool {
+		return decodeOrdinal(smeshers[i].Name) < decodeOrdinal(smeshers[j].Name)
+	})
```
dshulyak: i think it is sufficient to sort them in deployNodes function

blocks/certifier.go
```diff
@@ -163,7 +168,7 @@ func (c *Certifier) run() error {
 	for layer := c.layerClock.GetCurrentLayer(); ; layer = layer.Add(1) {
 		select {
 		case <-c.layerClock.AwaitLayer(layer):
-			_ = c.prune()
+			_ = c.genCertificate()
```
dshulyak, line 171: certificate should be generatable after required weight was received. i don't understand how it makes sense to you

dshulyak (commented):
- `buffer` is unnecessary. eligibility can be verified when received and saved into map[Layer][BlockID]
- i don't think that it is a good option to store all signatures. i also don't understand whats the problem with doing it in a clean way. i would rather output certificate in one of two events:
   - register is called and we already have a certificate
   - otherwise when required units were received after register
- it is pretty clear that we can't just keep them, as they are not validated when received. so anybody can spam node with fake certificates until it blows up. i think syncer should not be asking for certificates until it downloaded required ballots for beacon

countvonzero:
> but that is not relevant if someone disagrees. cert message is either valid according to your view or invalid, so once there are valid msgs with total weight > 400, whats the point to keep adding them?

so during TestAddNodes i found that a new node that joined the network always have higher total weight than the online nodes, 
because it fetched ATXs from its peers. so it often rejects the certificates it synced because it finds most those CertifyMessage invalid. this happens amongst the online nodes too.

say there are 800 msgs in the network (assuming every miner has eligibility 1)
node A collects [0, 400] to create CertificateA
node B collects [3, 403] to create CertificateB
you don't think node C (new node with 20(?) peers) has a better chance to accept node A/B's certificate if every certificate has all 800 CertifyMessage?

dshulyak:
> you don't think node C (new node with 20(?) peers) has a better chance to accept node A/B's certificate if every certificate has all 800 CertifyMessage?

i think it has no chances to accept certificate if it is using different state for eligibility distribution

the way i understand it is lets say there is some state X (which is prepared based on atxs X1, X2, X3)
if both node A and node B are using state X for eligibility - they will both compute the same eligibility distribution for certifiers. That certificate can be composed from different signatures but it is still the same certificate.

the eligibility distribution will change completely if node C is using different set of atxs (lets say from state X plus atx Y1). so all eligibilities that were computed using state X will be invalid if they are verified using state Y.

countvonzero:
> > you don't think node C (new node with 20(?) peers) has a better chance to accept node A/B's certificate if every certificate has all 800 CertifyMessage?
> 
> i think it has no chances to accept certificate if it is using different state for eligibility distribution
> 
> the way i understand it is lets say there is some state X (which is prepared based on atxs X1, X2, X3) if both node A and node B are using state X for eligibility - they will both compute the same eligibility distribution for certifiers. That certificate can be composed from different signatures but it is still the same certificate.
> 
> the eligibility distribution will change completely if node C is using different set of atxs (lets say from state X plus atx Y1). so all eligibilities that were computed using state X will be invalid if they are verified using state Y.

thanks for the explanation. i'll revert back to the old behavior. 
and will file an issue to record the weight differences btwn nodes and debug/resolve after feature complete.

countvonzero:
> * `buffer` is unnecessary. eligibility can be verified when received and saved into map[Layer][BlockID]

agreed 

> * it is pretty clear that we can't just keep them, as they are not validated when received. so anybody can spam node with fake certificates until it blows up. i think syncer should not be asking for certificates until it downloaded required ballots for beacon

note that msgs stored in buffer ARE validated before storing into buffer. but as you stated above. changing the data structure to map[Layer][BlockID] is cleaner

mesh/mesh.go
countvonzero: ok. i modified syncer to fetch layer data and layer opinions separately.
the layer opinion is part of the state sync, after beacon is available.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
@dshulyak please take another look. thank you.

added the following change
- certifier accepts early message in the same data structure and generates certificate as early as possible
- syncer fetch layer data and opinions separately
- change TestFailedNodes to get layer hashes multiple times. for the hashes of earlier layers, nodes may change its mind about which block to apply after tortoise verifies.
- change vm to return empty hash when no layers were applied. this was causing nodes to get in a crash loop during TestFailedNodes when mesh is trying to revert to last applied layer 8 (which was empty)
- change new join nodes to wait for the next epoch to publish its first ATX

activation/activation.go
```diff
@@ -301,6 +317,39 @@ func (b *Builder) generateProof() error {
...
+		log.Stringer("current_epoch", currentLayer.GetEpoch()),
+		log.Stringer("wait_till", waitTill),
+		log.Stringer("wait_till_epoch", waitTill.GetEpoch()))
+	select {
+	case <-ctx.Done():
+		return
+	case <-b.layerClock.AwaitLayer(waitTill):
+		// this estimate work if the majority of the nodes use poet servers that are configured the same way.
+		// TODO: do better when nodes use different poet services
+		timer := time.NewTimer(time.Second * 35)
```
dshulyak: comment doesn't make sense.
are you waiting for next epoch or 35 seconds after the epoch? what is special about those 35 seconds?

config/presets/testnet.go
```diff
@@ -56,6 +56,9 @@ func testnet() config.Config {
 	conf.POST.MaxNumUnits = 4
 	conf.POST.MinNumUnits = 2
 
+	conf.POET.CycleGap = 10 * time.Second
+	conf.POET.PhaseShift = 20 * time.Second
```
dshulyak: it is better to remove poet config from this pr completely, it will only cause confusion since it is not used

mesh/mesh.go
```diff
@@ -389,7 +375,8 @@ func (msh *Mesh) ProcessLayer(ctx context.Context, layerID types.LayerID) error
 	logger.Info("processing layer")
 
 	// pass the layer to tortoise for processing
-	newVerified := msh.triggerTortoise(ctx, logger, layerID)
+	newVerified := msh.trtl.HandleIncomingLayer(ctx, layerID)
+	logger.With().Debug("tortoise results", log.FieldNamed("verified", newVerified))
```
dshulyak: log.Stringer

mesh/mesh.go
```diff
@@ -410,7 +397,7 @@ func (msh *Mesh) ProcessLayer(ctx context.Context, layerID types.LayerID) error
 
 	if !to.Before(from) {
 		if err := msh.pushLayersToState(ctx, from, to, newVerified); err != nil {
-			logger.With().Error("failed to push layers to state", log.Err(err))
+			logger.With().Warning("failed to push layers to state", log.Err(err))
```
dshulyak: why warning? this is not expected event in any circumstances

activation/activation.go
```diff
@@ -301,6 +317,39 @@ func (b *Builder) generateProof() error {
 	return nil
 }
 
+func (b *Builder) waitForFirstATX(ctx context.Context) {
```
dshulyak, line 320: should it be called waitForNextEpoch ?

config/presets/fastnet.go
```diff
@@ -50,6 +50,9 @@ func fastnet() config.Config {
 	conf.POST.MaxNumUnits = 4
 	conf.POST.MinNumUnits = 2
 
+	conf.POET.CycleGap = 10 * time.Second
+	conf.POET.PhaseShift = 20 * time.Second
```
dshulyak: they should be set in systest if you want to set them

systest/cluster/cluster.go
```diff
@@ -309,10 +309,12 @@ func (c *Cluster) AddSmeshers(cctx *testcontext.Context, n int) error {
...
 	}
 	bootnodes := c.clients[:c.bootnodes]
+	smeshers := c.clients[c.bootnodes:]
 	c.clients = nil
+	smeshers = append(smeshers, clients...)
 	c.clients = append(c.clients, bootnodes...)
-	c.clients = append(c.clients, clients...)
-	c.smeshers = len(clients)
+	c.clients = append(c.clients, smeshers...)
+	c.smeshers = len(smeshers)
```
dshulyak: i think less prone to bugs would be to avoid append on smeshers slice, e.g

```
c.clients = append(..., bootnodes...)
c.clients = append(..., smeshers...)
c.clients = append(...., clients...)
c.smeshers += len(clients)
```

activation/activation.go
```diff
@@ -279,6 +293,8 @@ func (b *Builder) run(ctx context.Context) {
 		return
 	}
 
+	b.waitForFirstATX(ctx)
```
dshulyak: it shouldn't be in initializer

activation/activation.go
```diff
@@ -301,6 +317,39 @@ func (b *Builder) generateProof() error {
...
+		timer := time.NewTimer(time.Second * 35)
+		defer timer.Stop()
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
+		}
+	}
+	b.log.WithContext(ctx).With().Info("ready to build first atx",
+		log.Stringer("current_layer", currentLayer))
```
dshulyak: currentLayer reports old layer, that was computed before waiting

activation/activation.go
dshulyak: i looked up logs and see why you set 35, but it doesn't make any sense. something is wrong with time/layer computation in go-spacemesh

atx's that target next epoch are published at epoch_start + phase_shift - cycle_gap. so after first layer of the epoch started you need to wait phase_shift - cycle_gap + expected broadcast duration (for fastnet we can set it to 5s)

UPD:
clock is 1-based so if genesis is at 03:01 - first layer of the first epoch will be ticked at 03:01
so according to the clock first layer of the 4th epoch should be layer 17
but the computation in https://github.com/spacemeshos/go-spacemesh/blob/develop/common/types/activation.go#L41
is 0-based so first layer of the 4th epoch is layer 16

we will probably have to change the clock to be 0-based, otherwise there will be a mess

genvm/vm.go
```diff
@@ -88,7 +88,11 @@ func (v *VM) GetLayerApplied(tid types.TransactionID) (types.LayerID, error) {
 
 // GetStateRoot gets the current state root hash.
 func (v *VM) GetStateRoot() (types.Hash32, error) {
-	return layers.GetLatestStateHash(v.db)
+	root, err := layers.GetLatestStateHash(v.db)
+	if errors.Is(err, sql.ErrNotFound) {
+		return types.Hash32{}, nil
+	}
+	return root, err
```
dshulyak, line 98: not sure if this is correct. i would rather pass empty layer to vm and let it persist empty (or previous if we will use cumulative) hash. but no reason to do it in this pr

activation/activation.go
countvonzero: yikes. for some reason i didn't update this to use the values set in config. the 35 sec was done for systest testing :(

activation/activation.go
countvonzero: it's in `run()`. not the initializer

config/presets/testnet.go
countvonzero: i was gonna use this in the wait computation and forgot (and used 35 instead...)

mesh/mesh.go
countvonzero: it is expected when a certificate is not available. we cannot find a block to apply in that case.

activation/activation.go
countvonzero: i will file an issue for the clock for now and set the config values in systest.

systest/cluster/cluster.go
countvonzero: done

config/presets/fastnet.go
countvonzero: done

mesh/mesh.go
countvonzero: changed back to error. it is an error for online nodes.

activation/activation.go
countvonzero: https://github.com/spacemeshos/go-spacemesh/issues/3508

genvm/vm.go
countvonzero: added todo to rethink this. it currently can cause the TestFailedNodes to stall and timeout the CI test

activation/activation.go
countvonzero: fixed

countvonzero:
bors try

countvonzero:
@dshulyak all feedback addressed. TestAddNodes passed 3 times in longevity cluster test-sltm test-piwv test-jvkg

activation/activation.go
dshulyak: you need to wait longer if that clock is not fixed. if you check miner smesher-27 logs `{namespace="test-piwv", pod="smesher-27-0"}  |= "atxBuilder"` - the first atx is still messed up: 

```
{"L":"INFO","T":"2022-08-28T10:07:52.828-0400","N":"9272a.atxBuilder   ","M":"new atx challenge is ready","node_id":"9272a5bf98750e7d5009db903e1b872d80077c1b1f7db2870cc24bd2e527ceaa","module":"atxBuilder","sessionId":"b52ebb53-04c3-411b-a903-2d98b06f3f57","target_epoch":"4","current_epoch":"4","name":"atxBuilder"}
```

and then it fixes atx in the next epoch:

```
{"L":"INFO","T":"2022-08-28T10:09:16.057-0400","N":"9272a.atxBuilder   ","M":"new atx challenge is ready","node_id":"9272a5bf98750e7d5009db903e1b872d80077c1b1f7db2870cc24bd2e527ceaa","module":"atxBuilder","sessionId":"3ee492d6-a97f-4120-9c5d-88ba224181d6","target_epoch":"6","current_epoch":"5","name":"atxBuilder"}
```

to fix this issue either grace period needs to be adjusted by one more layer duration (so total will be 20s for fastnet), or better waitTill should be equal to nextEpoch.FirstLayer().Add(1)

---

on the other hand if it doesn't break the test - it can be fixed in https://github.com/spacemeshos/go-spacemesh/issues/3508

dshulyak (approved):
please see a note regarding wait time

countvonzero:
> please see a note regarding wait time

it's surprising easy to fix the clock to start from layer 0. it's done as part of this PR.
however, after the clock fix, the phase-shift-cycle-gap+grace-period with the values as is, doesn't work. (see test-drcj)
so i simply changed to wait for 1 more layer as you suggested (see test-djfr) and removed all the poet config.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/8061868706?check_suite_focus=true)

@dshulyak 
i spoke too soon. TestShortTimeskew does get flaky when i test with other cluster too. 
haven't looked into it yet. will appreciate any pointers. thanks

dshulyak:
> it's surprising easy to fix the clock to start from layer 0. it's done as part of this PR.
however, after the clock fix, the phase-shift-cycle-gap+grace-period with the values as is, doesn't work. (see test-drcj)
so i simply changed to wait for 1 more layer as you suggested (see test-djfr) and removed all the poet config.

waiting for one more layer without poet config is totally random, it may work only in systests case because phase_shift - cycle gap + broadcast is 15s and layer is 15s as well
i suggested to wait one more layer as a workaround for clock issue, not as a replacement for everything. 

what exactly doesn't work? in the shared namespace both smesher-26 and smesher-27 created correct atxs that target next epoch, so wait time was correct

```
 {"L":"INFO","T":"2022-08-28T17:46:04.002-0400","N":"962f0.atxBuilder   ","M":"new atx challenge is ready","node_id":"962f0c8e101da4273999ae798bfcc6449c99a411fb34ec35a7c791a54bcc9563","module":"atxBuilder","sessionId":"1581ecc9-d9f1-49f9-a64e-b1e67a2fa6e3","target_epoch":"5","current_epoch":"4","name":"atxBuilder"}
```

--- 

also in test-drc https://github.com/spacemeshos/go-spacemesh/issues/3508 wasn't fixed fully . check `release tick` in the logs
i think it was better not to touch clock in this pr, and do a local workaround so that it can be merged

activation/activation.go
```diff
@@ -301,6 +303,30 @@ func (b *Builder) generateProof() error {
 	return nil
 }
 
+func (b *Builder) waitForFirstATX(ctx context.Context) {
+	if b.layerClock.GetCurrentLayer().GetEpoch() == 0 { // genesis miner
+		return
+	}
+	if _, err := b.cdb.GetPrevAtx(b.nodeID); err == nil { // not a new miner
+		return
+	}
```
dshulyak: what if miner was temporarily offline?

timesync/converter.go
```diff
@@ -17,13 +17,13 @@ func (lc LayerConv) TimeToLayer(t time.Time) types.LayerID {
...
 		return types.LayerID{}
 	}
-	return types.NewLayerID(uint32(t.Sub(lc.genesis)/lc.duration) + 1)
+	return types.NewLayerID(uint32(t.Sub(lc.genesis) / lc.duration))
 }
 
 // LayerToTime returns the time of the provided layer.
 func (lc LayerConv) LayerToTime(id types.LayerID) time.Time {
-	if !id.After(types.LayerID{}) { // layer 1 is genesis, consider 0 also as genesis
+	if id == (types.LayerID{}) { // layer 1 is genesis, consider 0 also as genesis
```
dshulyak: it will work without this check as well

dshulyak (commented):
maybe t.Skip TestAddNodes and remove atx/clock changes, i think it will be better to merge them in a follow up. clock fix is not complete, atx changes also need a bit more work

TestShortTimeSkew failed because part of the cluster votes for layer 9 as empty, and the other votes to support `{namespace="test-ovbi", pod="smesher-1-0"} |= "block generated" | json | block_id="1f15d3c2b0"` . I am not completely sure why it only surfaced in this change. Maybe worth to skip and follow up on it as well

countvonzero:
>TestShortTimeSkew failed because part of the cluster votes for layer 9 as empty, and the other votes to support {namespace="test-ovbi", pod="smesher-1-0"} |= "block generated" | json | block_id="1f15d3c2b0" . I am not completely sure why it only surfaced in this change. Maybe worth to skip and follow up on it as well

i don't really understand this test. 

```
	var confirmed uint32
	watchLayers(ctx, eg, client, func(layer *spacemeshv1.LayerStreamResponse) (bool, error) {
		if layer.Layer.Number.Number == stopTest {
			return false, nil
		}
		if layer.Layer.Status == spacemeshv1.Layer_LAYER_STATUS_CONFIRMED {
			tctx.Log.Debugw("confirmed layer",
				"layer", layer.Layer.Number.Number,
				"hash", prettyHex(layer.Layer.Hash),
			)
			confirmed = layer.Layer.Number.Number
			if confirmed >= stopSkew {
				return false, nil
			}
		}
		return true, nil
	})
	require.NoError(t, eg.Wait())
	require.LessOrEqual(t, int(stopSkew), int(confirmed))
```
the code stops watchLayers() when `confirmed >= stopSkew`. but then require that `stopSkew <= confirmed` 
doesn't that essentially mean that it requires `confired == stopSkew`?

in any case, i'm skipping this test as well. but i am not sure how i can fix this

countvonzero:
bors try

dshulyak:
> the code stops watchLayers() when confirmed >= stopSkew. but then require that stopSkew <= confirmed
doesn't that essentially mean that it requires confirmed == stopSkew?

i think it was in case if layers are not confirmed one by one, so that test is fine as long as confirmed advanced atleast to stopSkew level before stopTest layer is reached.

countvonzero:
bors try

countvonzero:
> > the code stops watchLayers() when confirmed >= stopSkew. but then require that stopSkew <= confirmed
> > doesn't that essentially mean that it requires confirmed == stopSkew?
> 
> i think it was in case if layers are not confirmed one by one, so that test is fine as long as confirmed advanced atleast to stopSkew level before stopTest layer is reached.

k. i'll work on re-enabling these two tests in followup PRs

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

PR #3510, opened 2022-08-29
[Merged by Bors] - replace remaining fnv with repo's default hash

### LINKED ISSUE #1508: calcValue in hare beacon uses fnv
hare/eligibility/beacon.go

The func calcValue (of the beacon) uses fnv instead of our agreed (low collision hash).
To my understanding, we still haven't agreed on such hash.
We use Hash32 for blocks and we also have Hash12. We should decide what we want to use for the beacon.

@barakshani any input on that?

jonZlotnik:
## Motivation

Closes #1508

The last remnants of FNV were in the hare package.

## Changes

Replaced usage of FNV with truncated SHA256 (what's used everywhere else)

## Test Plan

No extra tests. Existing tests must continue to pass.

### DIFF
--- a/hare/notifytracker.go
+++ b/hare/notifytracker.go
@@ -2,7 +2,8 @@ package hare
 
 import (
 	"encoding/binary"
-	"hash/fnv"
+
+	"github.com/spacemeshos/go-spacemesh/hash"
 )
 
 // notifyTracker tracks notify messages.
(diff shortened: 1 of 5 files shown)


hare/notifytracker.go
```diff
@@ -48,21 +50,17 @@ func (nt *notifyTracker) NotificationsCount(s *Set) int {
...
-	hash := fnv.New32()
+func calcID(k uint32, set *Set) types.Hash32 {
+	h := hash.New()
 
 	// write K
 	buff := make([]byte, 4)
 	binary.LittleEndian.PutUint32(buff, k) // K>=0 because this not pre-round
-	hash.Write(buff)
-
-	// TODO: is this hash enough for this usage?
```
jonZlotnik, line 59: Switching from 32bits to 256bits (32 bytes) because the set ids aren't sent over the wire and aren't stored on disk. Easier than trying to prove the absence of a vulnerability caused by a hash collision.

If Hare setIDs become a performance or memory issue, we can change algo later on without worry as they're not sent over the wire.

jonZlotnik:
bors try

dshulyak (approved):
certificates map was changed to use Hash32 key. does it make sense to change the other map as well and remove cast to uint32?

fasmat (approved):

jonZlotnik:
> certificates map was changed to use Hash32 key. does it make sense to change the other map as well and remove cast to uint32?

Yep, makes sense to remove confusing casting that doesn't have a good reason.

jonZlotnik:
bors try

jonZlotnik:
bors merge

PR #3513, opened 2022-08-30
[Merged by Bors] - tortoise: not using base ballots with abstain votes for empty layers

### LINKED ISSUE #3446: hare: should not output empty block ID when hare fails
## Description

when hare fails locally, it should not output empty block ID. instead, it should either
- wait for the sync flow to sync the block certificate from its peers
- wait out the zdist layers. after zdist layer has passed, hare output no longer matter for that layer because tortoise will use global opinions for that layer.

this is related to https://github.com/spacemeshos/go-spacemesh/issues/3445.
in that case, hare output empty block ID. TestScheduleBasic/verify fails because it happens to use this node as the reference hash, and all other nodes have the same hash.
 
Screenshot from 2022-08-14 16-17-41

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3446

to encode votes tortoise picks base ballot and encodes voting diff (support, against for blocks, abstain for layers) explicitly. if hare didn't terminate with empty layer immediately then tortoise will vote abstain for that layer within zdist. unless it receives certificate and updates local opinion it should vote against that layer after zdist layers passed.

it is obvious that we can't encode against votes for an empty layer explicitly, because it has no blocks and there is nothing to add to the diff. therefore we should not pick base ballot that votes abstain for a layer if it is empty and already terminated by hare.

this change also updates hare to not output empty block in case of failures, as discussed in the issue above.

### DIFF
--- a/tortoise/tortoise.go
+++ b/tortoise/tortoise.go
@@ -72,10 +72,6 @@ func (t *turtle) cloneTurtleParams() *turtle {
 
 func (t *turtle) init(ctx context.Context, genesisLayer *types.Layer) {
 	// Mark the genesis layer as “good”
-	t.logger.WithContext(ctx).With().Info("initializing genesis layer for verifying tortoise",
-		genesisLayer.Index(),
-		genesisLayer.Hash().Field(),
-	)
 	genesis := genesisLayer.Index()
 	for _, blk := range genesisLayer.Blocks() {
 		t.blockLayer[blk.ID()] = genesis
@@ -239,10 +235,9 @@ func (t *turtle) EncodeVotes(ctx context.Context, conf *encodeConf) (*types.Vote
 	}
 
 	logger.With().Info("choose base ballot",
-		ballotID,
-		ballotLID,
+		log.Stringer("base_layer", ballotLID),
 		log.Stringer("voting_layer", last),
-		log.Object("votes", votes),
+		log.Inline(votes),
 	)
 
 	metrics.LayerDistanceToBaseBallot.WithLabelValues().Observe(float64(t.last.Value - ballotLID.Value))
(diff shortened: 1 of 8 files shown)


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

countvonzero:
@dshulyak instead of not choosing a base ballot that abstains, does it make sense to create vote diff that support EmptyBlockID instead?

dshulyak:
> instead of not choosing a base ballot that abstains, does it make sense to create vote diff that support EmptyBlockID instead?

@countvonzero can you elaborate? empty block id can't be used as it not sufficient to identify which layer is empty. there is an option to add one more slice, `empty` with layer ids that are voted as empty. it makes encoding shorter, but increases complexity and ambiguity

countvonzero:
> > instead of not choosing a base ballot that abstains, does it make sense to create vote diff that support EmptyBlockID instead?
> 
> @countvonzero can you elaborate? empty block id can't be used as it not sufficient to identify which layer is empty. there is an option to add one more slice, `empty` with layer ids that are voted as empty. it makes encoding shorter, but increases complexity and ambiguity

you are right. there is no layer attached to just a block id.

tortoise/tortoise.go
```diff
@@ -291,6 +285,19 @@ func (t *turtle) firstDisagreement(ctx context.Context, blid types.LayerID, ball
...
+			// it is not possible to encode against for empty layer
+			// because there are no blocks
+			// therefore we should not pick base ballot that votes abstain on
+			// layer outside zdist if layer is empty
+			if !isUndecided(t.Config, t.decided, lid, last) && t.full.abstained(ballotID, lid) {
+				t.logger.With().Debug("ballot votes abstain on a layer without blocks. can't use as a base ballot",
+					ballotID,
+					lid,
+				)
+				return types.LayerID{}, nil
```
countvonzero, line 298: returning layer id 0, instead of lid, here effectively move this ballot to the back of the base ballot sort?

countvonzero (approved):

tortoise/tortoise.go
dshulyak: yes

dshulyak:
bors merge

dshulyak:
bors merge

PR #3518, opened 2022-08-31
[Merged by Bors] - Nonce arithmetic and types in sqlite

### LINKED ISSUE #3386: Nonce arithmetic and types in sqlite
https://github.com/spacemeshos/go-spacemesh/blob/dd50f53d404055568c823748503b2d9427cbad38/sql/transactions/transactions.go#L187

On this line, and a few other places in this file, uint64 nonces are converted to int64 without checking for overflow. Can we add overflow protection here?

evt:
## Motivation
SQLite doesn't support `uint64` nonces we sort and compare by.

Closes #3386

## Changes
Store `nonce` as a byte representation of a uint64, using big endian encoding

## Test Plan
Existing tests

## DevOps Notes


dshulyak:
please add tests that would fail before and won't fail now

evt:
> please add tests that would fail before and won't fail now

done

sql/migrations/0001_initial.sql
```diff
@@ -57,7 +57,7 @@ CREATE TABLE transactions
     layer       INT,
     block       CHAR(20),
     principal   CHAR(20),
-    nonce       UNSIGNED LONG INT,
+    nonce       BLOB,
```
dshulyak, line 60: char(8) ?
i don't know if it changes anything for sqlite storage-wise, but probably won't hurt to specify exact size

dshulyak (approved):

sql/migrations/0001_initial.sql
evt:
```
TEXT. The value is a text string, stored using the database encoding (UTF-8, UTF-16BE or UTF-16LE).
BLOB. The value is a blob of data, stored exactly as it was input.
```
do we need to store bytes using database encoding? 🤔

dshulyak: interesting observation. i don't see why we would need to encode nonce to a utf8 string, so blob makes more sense

evt:
bors try

evt:
bors merge

PR #3519, opened 2022-08-31
[Merged by Bors] - Introduce VerifiedAtx as type

### LINKED ISSUE #3434: atxs: refactor atx type to avoid struct update
atx struct is updated in two cases:

- when received to compute id 
- and will be updated after it was verified, to set correct base_tick_height and tick_count https://github.com/spacemeshos/go-spacemesh/pull/3399

any mistakes can be catched only during runtime and only if we have sufficient test coverage. moreover this is prone to concurrency issues, if fields that are updated will be read concurrently in the other goroutine.

instead we should treat data as data and use different struct when we need more data. for example after atx is verified and verified data is required downstream we should use type:

```go
type VerifiedAtx struct {
     Atx
     BaseTickHeight uint64
     TickCount uint64
}
```

fasmat:
## Motivation
Fixes https://github.com/spacemeshos/go-spacemesh/issues/3434

## Changes
- `ActivationTxHeader` and `ActivationTx` are now independent types instead of being nested. A header can still be generated from an ATX by calling the `Header()` method.
- `VerifiedActivationTxHeader` is a new type that is returned by calling `Verify` on an ATX.
  - Naming could be improved - open for suggestions
  - The method now ensures that the ID and Node ID fields are set, if not it calculates those
  - Additionally it requires the caller to specify `baseTickHeight` and `tickCount`
  - This type is now used everywhere where `baseTickHeight` and `tickCount` are expected to be set, which ensures compile time checks where previously errors would only happen during runtime

## Test Plan
All existing tests still pass

## TODO

## DevOps Notes


activation/handler.go
```diff
@@ -139,7 +139,7 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error
 // SyntacticallyValidateAtx ensures the following conditions apply, otherwise it returns an error.
 //
 //   - If the sequence number is non-zero: PrevATX points to a syntactically valid ATX whose sequence number is one less
-//     than the current ATX's sequence number.
+//     than the current ATXs sequence number.
```
countvonzero, line 142: i don't think SequenceNumber used anymore. since you are here. can you update the comment so it make sense to you?

sql/atxs/atxs.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/spacemeshos/go-spacemesh/sql"
 )
 
-// Get gets an ATX by a given ATX ID.
-func Get(db sql.Executor, id types.ATXID) (atx *types.ActivationTx, err error) {
+// ByID gets an ATX by a given ATX ID.
+func ByID(db sql.Executor, id types.ATXID) (atx *types.ActivationTx, err error) {
```
countvonzero: `Get` is the consistent term in all the sql sub pkgs. i think we should either change all or leave it as Get

same for other methods. GetByXXXX

if you want change, please send a separate PR that updates all sql sub pkgs

countvonzero (commented):
this PR is bigger than it should be.

please separate the renaming to a separate PR. and when you do, please make sure its consistent across pkgs inside sql.

fasmat:
> this PR is bigger than it should be.
> 
> please separate the renaming to a separate PR. and when you do, please make sure its consistent across pkgs inside sql.

I'll undo most of the renaming and leave it for a different PR.

activation/handler.go
fasmat: The method still behaves as described in the comment and the builder produces ATXs that pass this check, besides that I didn't see sequence number being used anywhere.

I will create a separate small PR that removes the Sequence number from the NIPostChallenge struct.

fasmat: PR can be found here: https://github.com/spacemeshos/go-spacemesh/pull/3520

countvonzero: oh. wait. i thought there was a field called SequenceNumber (as the comment indicated). 
but i don't know about the field Sequence tho. it may be still useful. i don't know. sorry for the confusion.

activation/handler.go
```diff
@@ -112,7 +112,7 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error
 	h.processAtxMutex.Lock()
 	defer h.processAtxMutex.Unlock()
 
-	existingATX, _ := h.cdb.GetAtxHeader(atx.ID())
+	existingATX, _ := h.cdb.GetAtxByID(atx.ID())
```
evt: do we skip error here on purpose? 🤔

fasmat: We are only checking if the ATX with the given ID already exists in the cache / db. If it does, `existingATX` will be non-nil and `err` will be nil. If the ATX does not exist in the cache and cannot be fetched it will be `nil` and `err` will be non-nil.

The returned error will be some SQL error that would need to be parsed to be handled correctly. The only error that could give extra info besides "atx hasn't been stored yet" are DB connection issues. These arguably aren't needed to be handled here, since they will be handled a few lines further down if storing the ATX fails anyway.

common/types/activation.go
```diff
@@ -317,6 +296,119 @@ func (atx *ActivationTx) GetShortPoetProofRef() []byte {
...
+}
+
+// SetNodeID sets the Node ID in the ATX's cache.
+func (atx *ActivationTx) SetNodeID(nodeID *NodeID) {
+	atx.nodeID = nodeID
+}
+
+// Verify an ATX for a given base TickHeight and TickCount.
+func (atx *ActivationTx) Verify(baseTickHeight, tickCount uint64) *VerifiedActivationTx {
+	// TODO(mafa): should we also invoke calcAndSetID() as well as calcAndSetNodeID() here? maybe after checking that all required fields are set?
```
fasmat: Opinions? I could get rid of these two methods because in nearly all places the call chain looks like:

```golang
atx := newAtx()
atx.CalcAndSetID()
atx.CalcAndSetNodeID()
vAtx := atx.Verify(...)
```

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

activation/handler.go
```diff
@@ -139,112 +139,112 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error
...
-		return fmt.Errorf("invalid nipost: %v", err)
+		return nil, fmt.Errorf("invalid nipost: %v", err)
 	}
-	atx.Verify(baseTickHeight, leaves/h.tickSize)
-	return nil
+
+	vAtx := atx.Verify(baseTickHeight, leaves/h.tickSize)
+
+	// TODO(mafa): instead of reporting the event immediately, should we only report it after validation passed?
+	events.ReportNewActivation(vAtx)
```
countvonzero: i don't know why it was reported immediately before, but i think the right thing to do is to report it after ProcessAtx() succeed, before "got new atx" logged.

common/types/activation.go
```diff
@@ -180,18 +162,13 @@ func (atxh *ActivationTxHeader) TickHeight() uint64 {
...
-	atxh.tickCount = tickCount
-}
-
 // NIPostChallenge is the set of fields that's serialized, hashed and submitted to the PoET service to be included in the
 // PoET membership proof. It includes ATX sequence number, the previous ATX's ID (for all but the first in the sequence),
 // the intended publication layer ID, the PoET's start and end ticks, the positioning ATX's ID and for
 // the first ATX in the sequence also the commitment Merkle root.
 type NIPostChallenge struct {
+	// Sequence number counts the number of ancestors of the ATX. It sequentially increases for each ATX in the chain.
+	// It can be used in proving to other nodes that a malicious party created more ATXs than they would are allowed to.
```
countvonzero: rephrase to "Two ATXs with the same sequence number from the same miner can be used as the proof of malfeasance against that miner."

common/types/activation.go
```diff
@@ -223,9 +200,17 @@ func (challenge *NIPostChallenge) String() string {
 // structure is serialized and signed. It includes the header fields, as well as the larger fields that are only used
 // for validation: the NIPost and the initial Post.
 type InnerActivationTx struct {
```
countvonzero, line 192: what is the motivation to remove the embedded-ness of the header?

common/types/activation.go
```diff
@@ -274,11 +258,6 @@ func (atx *ActivationTx) MarshalLogObject(encoder log.ObjectEncoder) error {
 	encoder.AddUint32("epoch", uint32(atx.PubLayerID.GetEpoch()))
 	encoder.AddUint64("num_units", uint64(atx.NumUnits))
 	encoder.AddUint64("sequence_number", atx.Sequence)
-	if atx.verified {
-		encoder.AddUint64("base_tick_height", atx.baseTickHeight)
-		encoder.AddUint64("tick_count", atx.tickCount)
-		encoder.AddUint64("weight", atx.GetWeight())
```
countvonzero, line 280: we should still log these. use atx.Header()?

common/types/activation.go
countvonzero, line 192: do you have a test that guarantee that InnerActivationTx has everything public field of ActivationTxHeader?
i.e. how do we keep these two struct in sync over time?

datastore/store.go
```diff
@@ -51,13 +51,12 @@ func (db *CachedDB) GetAtxHeader(id types.ATXID) (*types.ActivationTxHeader, err
 		return nil, fmt.Errorf("get ATXs from DB: %w", err)
 	}
 
-	db.atxHdrCache.Add(id, &atx.ActivationTxHeader)
-	return &atx.ActivationTxHeader, nil
+	return atx.Header(), nil
```
countvonzero: here you are not getting it from the cache anymore. you are creating a new struct?

go.sum
```diff
@@ -818,8 +818,8 @@ golang.org/x/net v0.0.0-20210726213435-c6fcb2dbf985/go.mod h1:9nx3DQGgdP8bBQD5qx
 golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd/go.mod h1:CfG3xpIq0wQ8r1q4Su4UZFWDARRcnwPjda9FqA0JpMk=
 golang.org/x/net v0.0.0-20220225172249-27dd8689420f/go.mod h1:CfG3xpIq0wQ8r1q4Su4UZFWDARRcnwPjda9FqA0JpMk=
 golang.org/x/net v0.0.0-20220624214902-1bab6f366d9e/go.mod h1:XRhObCWvk6IyKnWLug+ECip1KBveYUHfp+8e9klMJ9c=
-golang.org/x/net v0.0.0-20220822230855-b0a4917ee28c h1:JVAXQ10yGGVbSyoer5VILysz6YKjdNT2bsvlayjqhes=
-golang.org/x/net v0.0.0-20220822230855-b0a4917ee28c/go.mod h1:YDH+HFinaLZZlnHAfSS6ZXJJ9M9t4Dl22yv3iI2vPwk=
+golang.org/x/net v0.0.0-20220826154423-83b083e8dc8b h1:ZmngSVLe/wycRns9MKikG9OWIEjGcGAkacif7oYQaUY=
+golang.org/x/net v0.0.0-20220826154423-83b083e8dc8b/go.mod h1:YDH+HFinaLZZlnHAfSS6ZXJJ9M9t4Dl22yv3iI2vPwk=
```
countvonzero: why is this needed? shouldn't dependabot do this?

sql/atxs/atxs.go
```diff
@@ -20,14 +20,14 @@ func Get(db sql.Executor, id types.ATXID) (atx *types.ActivationTx, err error) {
...
 		v.SetID(&id)
-		v.Verify(baseTickHeight, tickCount)
 		nodeID := types.NodeID{}
 		stmt.ColumnBytes(3, nodeID[:])
 		v.SetNodeID(&nodeID)
-		atx, err = &v, nil
+
+		baseTickHeight := uint64(stmt.ColumnInt64(1))
+		tickCount := uint64(stmt.ColumnInt64(2))
+		atx, err = v.Verify(baseTickHeight, tickCount), nil
```
countvonzero: why ignore the error here?

common/types/activation_scale.go
```diff
@@ -227,6 +256,29 @@ func (t *ActivationTx) DecodeScale(dec *scale.Decoder) (total int, err error) {
 	return total, nil
 }
 
+func (t *VerifiedActivationTx) EncodeScale(enc *scale.Encoder) (total int, err error) {
```
dshulyak: this needs to be removed

common/types/activation.go
fasmat: In an `ActivationTx` these values are always `0` because they have not been set. Only after calling `Verify` these values are set in the `VerifiedActivationTx`

Basically `atx.verified` is now guaranteed to be `false`.

Also `MarshalLogObject()` on a `VerifiedActivationTx` will log those values.

datastore/store.go
fasmat: `GetAtxHeader` calles `GetFullAtx` if the header wasn't found in the cache. `GetFullAtx` already adds the header to the cache, no need to do it again here.

common/types/activation.go
fasmat: The header always has `baseTickHeight` and `tickCount` set, while an unverified ActivationTx does not. Making these two true separate objects removes the need for run time checks that these values are set.

You cannot get an `ActivationTxHeader` from an `ActivationTx` only from a `VerifiedActivationTx`.

go.sum
fasmat: `golang.org/x/net` and `google.golang.org/genproto` are libraries where google doesn't create new releases for - they don't tag them with a version number. These libraries can only be referenced by their commit hash. Every new commit is a bugfix or added feature though.

Dependabot doesn't update to unreleased versions, so as long as no version is tagged for these repos, dependabot won't update them automatically.

sql/atxs/atxs.go
fasmat: Which error? It's a double assignment:

```golang
atx = v.Verify(baseTickHeight, tickCount)
err = nil
```

before it was

```golang
v := functionCall()
atx, err = &v, nil
```

I'll reformat the code to make it clearer

common/types/activation_scale.go
fasmat: This is generated by `go generate`. The only way to remove `EncodeScale` and `DecodeScale` for `VerifiedActivationTx` is to move them to a separate file, which I now did.

common/types/activation.go
```diff
@@ -317,6 +296,54 @@ func (atx *ActivationTx) GetShortPoetProofRef() []byte {
...
+	atx.id = id
+}
+
+// SetNodeID sets the Node ID in the ATX's cache.
+func (atx *ActivationTx) SetNodeID(nodeID *NodeID) {
+	atx.nodeID = nodeID
+}
+
+// Verify an ATX for a given base TickHeight and TickCount.
+func (atx *ActivationTx) Verify(baseTickHeight, tickCount uint64) *VerifiedActivationTx {
```
poszu: This method's comment and name imply that this function performs **verification**. But it only constructs a "verified" type assuming the caller has already verified it somehow. It's misleading. Perhaps rename the method to something like `ToVerified`.

common/types/verified_activation.go
```diff
@@ -0,0 +1,69 @@
...
+	return vatx.tickCount
+}
+
+// TickHeight returns a sum of base tick height and tick count.
+func (vatx *VerifiedActivationTx) TickHeight() uint64 {
+	return vatx.baseTickHeight + vatx.tickCount
+}
+
+// Header returns the header of the atx.
+func (vatx *VerifiedActivationTx) Header() *ActivationTxHeader {
```
poszu: What's the motivation to return a pointer here instead of just a value? The users of this API might think that this method returns a pointer to the insides of `vatx`, but it returns a separate object (a copy basically). And they must consider what happens if `nil` is returned (I know it cannot return a `nil` but to know that you need to read the implementation).

I'd suggest returning by value. What do you think?

common/types/verified_activation.go
```diff
@@ -0,0 +1,69 @@
...
+type VerifiedActivationTx struct {
+	*ActivationTx
+
+	baseTickHeight uint64
+	tickCount      uint64
+}
+
+// GetWeight of the atx.
+func (vatx *VerifiedActivationTx) GetWeight() uint64 {
+	return uint64(vatx.NumUnits) * (vatx.tickCount)
```
poszu: This might be a stupid question... I don't know the practical bounds of these variables yet :)

Can it, in practice, overflow here? It's a uint32 * uint64. Perhaps consider annotating in a comment why it is safe to multiply.

Similarly when you add two `uint64` in `TickHeight()`.

common/types/activation.go
fasmat: While true, I avoided doing too many changes in one PR; the method already existed before and did the same thing (except for setting fields in `ActivationTx` instead of returning a new `VerifiedActivationTx`). I can rename it to `ToVerifiedActivationTx(baseTickHeight, tickCount uint64)`.

@dshulyak I also added a TODO in this method. I think ID and Node ID should be calculated here since I would expect a `VerifiedActivationTx` to have these two fields set. At the moment this isn't guaranteed and adding those would to some degree justify calling this method `Verify`.

common/types/activation.go
fasmat: I don't have a test, I'm also not sure how such a test would look like.

The only way to create an `ActivationTxHeader` (besides calling `new(ActivationTxHeader)`) is to call the method `VerifiedActivationTx.Header()`. So any field that a Header has needs to be set by this method from a field of a `VerifiedActivationTx` which includes `InnerActivationTx`.

common/types/verified_activation.go
fasmat: You are correct. `Header()` can never return `nil` not even conceptually, since every ATX has a Header. The main reason is again avoiding too many changes at once.

`Header()` returns a pointer to the object because everywhere we receive an `ActivationTxHeader` we expect a pointer instead of a value, i.e. https://github.com/spacemeshos/go-spacemesh/blob/store-and-chacke-full-atxs/activation/handler.go#L125 If header returns a value instead of a reference I would need to either change `ContextuallyValidateAtx` or change this code to something like:

```golang
header := atx.Header()
h.ContextuallyValidateAtx(&header)
```

fasmat: Also returning it as a value wouldn't be enough. `NIPostChallenge` contains a byte slice `InitialPostIndices` that would require a copy as well to make it a truly separate object.

I'm not sure about the size of this array, but if I remember correctly a Header is expected to be in the range of ~ 100 byte in size for the main net.

common/types/verified_activation.go
fasmat: It's a valid point. I'm not sure about the expected size of these variables as well. The biggest I have seen where < 1000, which after multiplication would still easily fit into a `uint64`, but maybe @dshulyak can give some insights here? :-)

sql/atxs/atxs.go
```diff
@@ -15,7 +15,7 @@ func Get(db sql.Executor, id types.ATXID) (atx *types.VerifiedActivationTx, err
 		stmt.BindBytes(1, id.Bytes())
 	}
 	dec := func(stmt *sql.Statement) bool {
-		var v types.VerifiedActivationTx
```
countvonzero: there should be a test that failed before your fix but passed with this.
i.e. 
- add a atx to db
- get the blob out
- parse it correctly

countvonzero (commented):
please also address questions in the earlier review. thanks

datastore/store.go
countvonzero: it adds to the cache in `GetFullAtx`, but you are not returning the cached value. you are creating a new struct. what am i missing?

countvonzero: also, what is the value of the header cache if one can arbitrarily call atx.Header() now?
should Header() be a private method in the cache instead?

go.sum
countvonzero: but this change is unrelated right?
i've had the unpleasant experience of binary search on a failed deps update along with feature change...

common/types/verified_activation.go
countvonzero: my understanding was that headers are always cached, hence the ptr

datastore/store.go
fasmat: I didn't change any behaviour. The code behaves the same as before:

- If you call either `GetFullAtx(id)` or `GetAtxHeader(id)` you will receive the Atx / header for the given ID
- If `GetAtxHeader(id)` is called and the value is cached, the cached value is returned
- If `GetFullAtx(id)` or `GetAtxHeader(id)` are called when the header for the given ID is not cached it will be cached

The only thing that changed is: if a header isn't cached for a given ID it is now only added to the cache once instead of twice.

common/types/activation.go
countvonzero: i am in favor if you can make it work.

datastore/store.go
fasmat: > also, what is the value of the header cache if one can arbitrarily call atx.Header() now? should Header() be a private method in the cache instead?

It's the same as before: one could have arbitrarily called `atx.ActivationTxHeader` instead of fetching the header from the cache and `Header()` can't be a private method because it is required to be called in packages other than `common/types`

countvonzero: not what my reading.
old code:
```
func (db *CachedDB) GetAtxHeader(id types.ATXID) (*types.ActivationTxHeader, error) {
	if id == *types.EmptyATXID {
		return nil, errors.New("trying to fetch empty atx id")
	}

	if atxHeader, gotIt := db.atxHdrCache.Get(id); gotIt {
		return atxHeader, nil
	}
	atx, err := db.GetFullAtx(id)
	if err != nil {
		return nil, fmt.Errorf("get ATXs from DB: %w", err)
	}

	db.atxHdrCache.Add(id, &atx.ActivationTxHeader)
	return &atx.ActivationTxHeader, nil
}
```
it cached the same ptr twice (once in `GetFullAtx()`, and once in `GetAtxHeader()`)
but it is returning the cached ptr. you are now returning a new struct.

go.sum
fasmat: If this PR gets merged first: https://github.com/spacemeshos/go-spacemesh/pull/3530 a rebase will get rid of this change, but for some reason bors seems to not work for the last few hours...

datastore/store.go
countvonzero: > It's the same as before: one could have arbitrarily called atx.ActivationTxHeader instead of fetching the header from the cache and Header() can't be a private method because it is required to be called in packages other than common/types

no. before all the header ptrs are cached after retrieving the full atx from db. here you are creating a new struct every time Header() is called.

common/types/verified_activation.go
fasmat: The change I made is instead of calling `header := &atx.ActivationTxHeader` one now calls `header := atx.Header()` unless one explicitly fetches the ATX or its header from the store it will NOT be cached and wasn't cached before.

datastore/store.go
fasmat: I get your point, but what is the behaviour we are expecting?

```golang
id := 1234
header1 := db.GetAtxHeader(id)

// make some change
header1.PrevATXID = 5678

header2 := db.GetAtxHeader(id)
same := header1 == header2
equal := reflect.DeepEqual(header1, header2)
```

Is `same` and `equal` expected to be `true` or `false`? Because before it could be either - `true` as long as the value was cached (last 600 entries fetched from store) and `false` if the cache became invalid between the two calls to `db.GetAtxHeader`.

countvonzero: i don't understand your example. i don't think that's how the codebase use the atx header cache....

the expected behavior is caching the header, not creating more memory than we need to.
now if you want to argue the need of the cache, that's a different matter.

fasmat: I removed the `Header()` method from `VerifiedActivationTx`. Now a header can only be fetched from the store/cache and is cached correctly after fetching it (or the full ATX with the same ID).

sql/atxs/atxs.go
fasmat: There is a test in `store_test.go` that asserts this: `TestBlobStore_GetATXBlob`

countvonzero: but why didn't it fail when systest failed when the wrong type was parsed?

fasmat:
bors try

activation/handler.go
```diff
@@ -139,117 +139,116 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.ActivationTx) error
...
-			return fmt.Errorf("validation failed: prevATX not found: %v", err)
+			return nil, fmt.Errorf("validation failed: prevATX not found: %v", err)
 		}
 
-		if prevATX.NodeID() != atx.NodeID() {
-			return fmt.Errorf("previous atx belongs to different miner. atx.ID: %v, atx.NodeID: %v, prevAtx.NodeID: %v",
-				atx.ShortString(), atx.NodeID(), prevATX.NodeID())
+		if prevATX.NodeID != atx.NodeID() {
+			return nil, fmt.Errorf(
+				"previous atx belongs to different miner. atx.ID: %v, atx.NodeID: %v, prevAtx.NodeID: %v",
```
countvonzero: can you also print out prevAtx's ID short string? thanks

api/grpcserver/grpcserver_test.go
```diff
@@ -114,10 +116,33 @@ var (
...
 
+	atx := types.NewActivationTx(challenge, addr1, nipost, numUnits, nil)
+	if err := activation.SignAtx(signer, atx); err != nil {
+		log.Println("failed to sign atx:", err)
+		os.Exit(1)
+	}
+	var err error
+	globalAtx, err = atx.Verify(0, 1)
+	if err != nil {
+		log.Println("failed to verify atx", err)
```
countvonzero: does the err needs formatting?

common/types/verified_activation.go
```diff
@@ -0,0 +1,54 @@
...
+
+type VerifiedActivationTx struct {
+	*ActivationTx
+
+	baseTickHeight uint64
+	tickCount      uint64
+}
+
+// GetWeight of the atx.
+func (vatx *VerifiedActivationTx) GetWeight() uint64 {
```
countvonzero, line 18: the result should always be the same as ActivationTxHeader. maybe worth factoring this out to a func that takes unit and tickcount. similarly for TickHeight.

countvonzero (dismissed):
nicely done. thanks

api/grpcserver/grpcserver_test.go
fasmat: TIL: `fmt.Println` (and `log` is just an extension of that) will not only check if an object implements the `fmt.Stringer` interface, but also `error` and automatically call `err.Error()` to get the string for err. So if just the message needs to be printed, no extra formatting is needed.

proof: https://go.dev/play/p/Jd5JrIsVVQc

common/types/activation.go
```diff
@@ -119,9 +126,24 @@ func (atxh *ActivationTxHeader) TargetEpoch() EpochID {
 	return atxh.PubLayerID.GetEpoch() + 1
 }
 
-// GetWeight of the atx.
+// GetWeight of the ATX. The total weight of the epoch is expected to fit in a uint64 and is
+// sum(atx.NumUnits * atx.TickCount for each ATX in a given epoch).
+// Space Units sizes are chosen such that NumUnits for all ATXs in an epoch is expected to be < 10^6.
```
noamnelke: `10^6` for all ATXs is wrong. Using Chia as a reference, their historical maximum netspace was >36 EiB which is 151,000,000 space units of 256 GiB. So we should expect / be prepared for `10^9`.

common/types/activation.go
```diff
@@ -119,9 +126,24 @@ func (atxh *ActivationTxHeader) TargetEpoch() EpochID {
...
+	return safeMul(uint64(atxh.NumUnits), (atxh.TickCount))
+}
+
+func safeMul(a, b uint64) uint64 {
+	if a > 0 && b > math.MaxUint64/a {
+		panic("uint64 overflow")
+	}
+	if b > 0 && a > math.MaxUint64/b {
+		panic("uint64 overflow")
+	}
```
noamnelke: Why is the second check needed? I assume it's because of rounding down that happens in the division. You could do one check if you did (on line 140) `b >= math.MaxUint64/a + 1`, but all of this is actually really confusing and error prone...

I find [Rob Pike's version](https://groups.google.com/g/golang-nuts/c/h5oSN5t3Au4/m/KaNQREhZh0QJ) with a single division more intuitively obvious - I need to think less to see that it's safe.

Adapted here it would be:
```go
func safeMul(a, b uint64) uint64 {
	c := a * b
	if a > 1 && b > 1 && c/b != a {
		panic("uint64 overflow")
	}
	return c
}
```

I'd be curious to see a benchmark comparison of the 3 options (and also to make sure that my version passes tests - maybe I'm missing something).

noamnelke (changes requested):
(second comment is optional - ignore if you disagree)

common/types/activation.go
fasmat: A very quick benchmark shows that this is not only faster than the double check version, but also hardly slower than multiplying without overflow check:

```
goos: linux
goarch: arm64
pkg: github.com/spacemeshos/go-spacemesh/common/types
BenchmarkSafeMul-6    	940181595	         1.256 ns/op	       0 B/op	       0 allocs/op
BenchmarkSafeMul2-6   	1000000000	         0.6292 ns/op	       0 B/op	       0 allocs/op
BenchmarkMul-6   	1000000000	         0.4933 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/spacemeshos/go-spacemesh/common/types	2.459s
```

I'll update the implementation to this version.

fasmat:
bors try

poszu (dismissed):
My comments have been addressed. LGTM 👍

noamnelke (dismissed):

fasmat:
bors merge

fasmat:
bors try

fasmat:
bors merge

countvonzero:
@fasmat i think i prefer to up the overall run tests time to 70min, and leave the systest part as 50min.

common/types/activation.go
```diff
@@ -133,7 +133,11 @@ func (atxh *ActivationTxHeader) TargetEpoch() EpochID {
 // to produce more over time. A uint64 should be large enough to hold the total weight of an epoch,
 // for at least the first few years.
 func (atxh *ActivationTxHeader) GetWeight() uint64 {
-	return safeMul(uint64(atxh.NumUnits), (atxh.TickCount))
+	return getWeight(uint64(atxh.NumUnits), (atxh.TickCount))
```
countvonzero: nit: (atxh.TickCount) the parenthesis not needed.

common/types/verified_activation.go
```diff
@@ -9,9 +9,14 @@ type VerifiedActivationTx struct {
...
-// GetWeight of the atx.
+// GetWeight of the ATX. The total weight of the epoch is expected to fit in a uint64 and is
+// sum(atx.NumUnits * atx.TickCount for each ATX in a given epoch).
+// Space Units sizes are chosen such that NumUnits for all ATXs in an epoch is expected to be < 10^6.
+// PoETs should produce ~10k ticks at genesis, but are expected due to technological advances
+// to produce more over time. A uint64 should be large enough to hold the total weight of an epoch,
+// for at least the first few years.
 func (vatx *VerifiedActivationTx) GetWeight() uint64 {
-	return uint64(vatx.NumUnits) * (vatx.tickCount)
+	return getWeight(uint64(vatx.NumUnits), (vatx.tickCount))
```
countvonzero: ditto

fasmat:
bors try

fasmat:
> @fasmat i think i prefer to up the overall run tests time to 70min, and leave the systest part as 50min.

It works but is extremely close - the systests finished in 49 minutes; I would feel more comfortable if we would increase the timeout to 60 min.

EDIT: I decreased the timeout of the run tests step to 50 minutes (from 60) and increased the timeout for the systest job from 50 to 60 minutes

countvonzero (approved):

fasmat:
bors merge

PR #3520, opened 2022-08-31
Remove sequence number from NIPostChallenge

fasmat:
## Motivation
Sequence number isn't used any more.

## Changes
Remove field from NIPostChallenge and remove validation for it

## Test Plan
Existing Tests

## TODO

## DevOps Notes


dshulyak (changes requested):
why are you removing it? i don't get. you removed a place where it was incremented, and saying that it is not used

PR #3523, opened 2022-09-01
[Merged by Bors] - vm: vesting with vault

### LINKED ISSUE #3269: vm: vesting wallet
Vesting wallet will have funds locked in it, which will be unlocked over the time (based on the layer).

This is pending on additional input from research

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3269

this change implements:
- vesting template

vesting is built on top of regular multisig template and reuses multisig code. spawn and spend transactions are encoded the same way as for multisig template. the difference is that vesting template supports drain vault method call. drain vault is a restricted case of the remote execution. arguments of the drain vault method call: vault address, recipient, amount.  if we would support generic remote execution at genesis this template type won't be needed.

when drain vault transaction is received it goes through regular verification flow, but during execution we load the state of the account specified at vault address. verify that account is spawned with a vault type. and execute Spend method on the remote (vault) account, while paying gas from principal (vesting) account.  

- vault template

vault stores locked funds that can be withdrawn (drained) after certain period. vault is initialized with a following variables - owner, vesting start, vesting end, total amount and initial unlock amount.

before vesting starts nothing can be withdrawn from a vault. at and after vesting end everything can be withdrawn.
during the vesting period the amount is computed as follows:
available = initial + (total - initial) * (block_layer - vesting_start) / (vesting_end - vesting_start)

vault account can't pay for gas, hence Verify always returns false. it cannot be self-spawned and it can't spend its own funds. in other words it can be executed only by another account remotely. it can be spawned by any account, but only owner can spend funds locked in the vault.


genvm/templates/vault/handler.go
```diff
@@ -0,0 +1,79 @@
...
+func (h *handler) Load(state []byte) (core.Template, error) {
+	dec := scale.NewDecoder(bytes.NewBuffer(state))
+	vault := &Vault{}
+	if _, err := vault.DecodeScale(dec); err != nil {
+		return nil, fmt.Errorf("%w: %s", core.ErrInternal, err)
+	}
+	return vault, nil
+}
+
+// Exec spawn or spend based on the method selector.
```
countvonzero: comment different from code.

genvm/templates/vault/handler.go
```diff
@@ -0,0 +1,79 @@
...
+func (h *handler) Exec(host core.Host, method uint8, args scale.Encodable) error {
+	if method != core.MethodSpend {
+		return fmt.Errorf("%w: unknown method %d", core.ErrMalformed, method)
+	}
+	spend := args.(*SpendArguments)
```
countvonzero, line 66: Exec() doesn't support MethodSpawn because accounts of this template is always spawned by other accounts?

genvm/templates/vault/vault.go
```diff
@@ -0,0 +1,72 @@
...
+	VestingEnd          core.LayerID
+
+	DrainedSoFar uint64
+}
+
+func (v *Vault) isOwner(address core.Address) bool {
+	return v.Owner == address
+}
+
+func (v *Vault) Available(lid core.LayerID) uint64 {
```
countvonzero, line 35: i know that the research was quite clear about not using layer id to determine the vesting schedule.
to me it's strange to merge this version of vault implementation and change it again later.
but it's fruitless to argue about it now as there is no spec. i guess we'll just merge this and change it later.

countvonzero (approved):

genvm/templates/vault/vault.go
dshulyak: but are you sure about that? both in Lane's economics implementation and Wojtek's documents layer id is used to determine vesting schedule

it is even discussed in the document that i shared in slack. i know that it is not finalized, but i didn't see even a note that layer id can't be used for vesting anywhere

genvm/templates/vault/handler.go
dshulyak: yes

genvm/templates/vault/vault.go
noamnelke: I've never heard any objections for using `LayerID` for the vesting schedule.

The issuance schedule is layer-based. I can't imagine how (or why) to design an issuance schedule that is based on anything else and it doesn't make sense not to tie vesting to the issuance.

@countvonzero do you have any reference to the objection? If it's not written anywhere - who objected? @tal-m?

dshulyak:
bors merge

genvm/templates/vault/vault.go
countvonzero: i thought the research want to use TTL to determine whether the vesting account can even start vesting (1 yr),
and not layer ID. or maybe i am confused with not having layer ID available in Verify()? @noamnelke

countvonzero: >@countvonzero do you have any reference to the objection? If it's not written anywhere - who objected? @tal-m?

i think it was in one of the video calls where Tal want to use LayerID to determine when the vesting wait time (1 year) has ended and the rest of the research members scoffed at Tal for that, as you seem to just agreed to use TLL and not LayerID for that purpose.

PR #3524, opened 2022-09-02
[Merged by Bors] - syncer do not set hare output for empty layer

### LINKED ISSUE #3522: layer fetcher erroneously sets hare_output to zero value
this is what is causing TestShortTimeskew to fail after i fixed issue with abstain votes in https://github.com/spacemeshos/go-spacemesh/pull/3513

at layer 11 `{namespace="test-koki"} |= "choose base ballot" | json | voting_layer="11"` , there will be a split in votes. half of the nodes will be voting for layer 9 as empty, and other will vote to support block. this split will remain for hdist since we are voting according to hare output within it.

both smesher-21 and smesher-1 correctly generated a block. however smesher-1 loaded hare_output into tortoise `{namespace="test-koki", pod="smesher-1-0"} |= "on hare output" | json | layer_id=9` and smesher-21 loaded empty layer `{namespace="test-koki", pod="smesher-21-0"} |= "on hare output" | json | layer_id=9`.

looking further smesher-21 actually tried to save hare output right before calling tortoise `{namespace="test-koki", pod="smesher-21-0"} |= "saving hare output for layer" | json | layer_id=9`. see a timing, it is right before a call to the tortoise. but it failed to update hare_output since syncer already updated layer as a zero `{namespace="test-koki", pod="smesher-21-0"} |= "tagging zero block layer" | json | layer_id=9`.

@countvonzero i don't recall what is the purpose of this code https://github.com/spacemeshos/go-spacemesh/blob/develop/fetch/layers.go#L434-L438  . it looks like that it needs to be removed

countvonzero:
## Motivation

Closes #3522 

## Changes

- when syncing, remove call to SetHareOutput on empty block ID when layer is empty. it should wait for a certificate instead.
- undo retry stream in TestFailedNodes.
- emit event Layer_LAYER_STATUS_APPLIED when a block is applied and use that for systest

### DIFF
--- a/api/grpcserver/mesh_service.go
+++ b/api/grpcserver/mesh_service.go
@@ -573,6 +573,8 @@ func convertLayerStatus(in int) pb.Layer_LayerStatus {
 		return pb.Layer_LAYER_STATUS_APPROVED
 	case events.LayerStatusTypeConfirmed:
 		return pb.Layer_LAYER_STATUS_CONFIRMED
+	case events.LayerStatusTypeApplied:
+		return pb.Layer_LAYER_STATUS_APPLIED
 	default:
 		return pb.Layer_LAYER_STATUS_UNSPECIFIED
 	}
--- a/events/reporter.go
+++ b/events/reporter.go
@@ -331,6 +331,7 @@ const (
 	LayerStatusTypeUnknown   = iota
 	LayerStatusTypeApproved  // approved by Hare
 	LayerStatusTypeConfirmed // confirmed by Tortoise
+	LayerStatusTypeApplied   // applied to state
 )
 
 // LayerUpdate packages up a layer with its status (which a layer does not ordinarily contain).
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/prometheus/common v0.37.0
 	github.com/pyroscope-io/pyroscope v0.27.0
 	github.com/seehuhn/mt19937 v1.0.0
-	github.com/spacemeshos/api/release/go v1.4.1-0.20220808081923-57864fb8ea2e
+	github.com/spacemeshos/api/release/go v1.4.1-0.20220905145225-2baea06a0206
 	github.com/spacemeshos/ed25519 v0.0.0-20220819111725-6f9b8412778f
 	github.com/spacemeshos/fixed v0.0.0-20210523192743-8d17e03c169a
 	github.com/spacemeshos/go-scale v0.0.0-20220825075539-b6b3deb9834c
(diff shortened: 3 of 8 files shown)


countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/runs/8148423413?check_suite_focus=true)

@dshulyak can you help me confirming this error?
i checked the nodes, they recovered from the missing hare certificate as expected (layer 15). so i think the failure is caused by grpc stream timeout, as indicated by log below  {namespace="test-ejzd", pod="boot-0-0"} 

do you agree?
```
{"L":"ERROR","T":"2022-09-01T22:10:28.711-0400","N":"1cc8a.app          .grpc         ","M":"finished streaming call with code Unknown","node_id":"1cc8ab7e9b0b3a78e1c62543df148a056e8b76ab1b2c51334ed6a3198c69c448","module":"app","module":"grpc","grpc.start_time":"2022-09-01T21:51:44-04:00","grpc.request.deadline":"2022-09-01T22:49:32-04:00","system":"grpc","span.kind":"server","grpc.service":"spacemesh.v1.MeshService","grpc.method":"LayerStream","peer.address":"10.8.3.28:44042","error":"send to stream: rpc error: code = Canceled desc = context canceled","grpc.code":"Unknown","grpc.time_ms":1124113.8}
```

i think the retry logic in TestFailedNode i added is faulty
https://github.com/spacemeshos/go-spacemesh/blob/75ea165d6fb450160a36679861ebb5d00d1a0b9f/systest/tests/nodes_test.go#L179

the stream will subscribe to the latest layer event and would not get the hashes from the older (disagreed) layers again.
what do you think about doing LayerQuery instead?

dshulyak:
> what do you think about doing LayerQuery instead?

sorry missed that. it is indeed doesn't make sense to retry a stream in such case
it won't be possible to use LayerQuery in watchStreams function, but generally can be used for this test

i also think it makes sense to emit event that layer is applied, so that event will be delivered even in case of revert. and use stream without retries

dshulyak:
Actually sorry i don't understand the workaround and what current failure means.

The only place where layer is confirmed is at https://github.com/spacemeshos/go-spacemesh/blob/develop/tortoise/algorithm.go#L221-L224 . The fact that hare output was applied immediately shouldn't make any difference. Also the rerun is not triggered in this test, so we pretty much can't receive more than one event for the same layer.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

PR #3525, opened 2022-09-04
[Merged by Bors] - Fix data races in flows_test.go

piersy:
## Motivation
Data races in test code can cause tests to be flaky.

## Changes
Added synchronisation to test code to ensure race free access to test results. Details of the data races can be found in the commit messages.

## Test Plan
I used the following command to run the offending tests multiple times and witnessed no data races.
```
ulimit -n 9999; CGO_LDFLAGS="-L/Users/piersy/projects/go-spacemesh/build/ -Wl,-rpath,/Users/piersy/projects/go-spacemesh/build/" TEST_LOG_LEVEL= go test -timeout 0 -v -count 10 -p 1 -race ./hare/... -run Test_multiple
```

## DevOps Notes


hare/flows_test.go
```diff
@@ -404,6 +404,7 @@ func Test_multipleCPs(t *testing.T) {
 	var pubsubs []*pubsub.PubSub
 	scMap := NewSharedClock(totalNodes, totalCp, time.Duration(50*int(totalCp)*totalNodes)*time.Millisecond)
 	outputs := make([]map[types.LayerID]LayerOutput, totalNodes)
```
countvonzero, line 406: thanks for submitting a PR @piersy to fix this. highly appreciated.
tho i don't think this fixes the data race. the race happens because the goroutines are modifying `output` at the same time. 
making sure the goroutines are done before we check the result is good, but not enough to fix the race.

piersy: Good point, I wonder why this doesn't seem to be triggering any data races?

I'll update to sync over the map.

hare/flows_test.go
```diff
@@ -448,6 +454,7 @@ func Test_multipleCPs(t *testing.T) {
 	for _, h := range test.hare {
 		close(h.blockGenCh)
 	}
+	outputsWaitGroup.Wait()
 	for _, out := range outputs {
```
countvonzero, line 455: you need to also lock here.

piersy: Hi @countvonzero I'm not seeing why that should be required.

My understanding is that after `outputsWaitGroup.Wait()` returns there is only one running go-routine (the main one) that has access to the outputs map, and consequently there is no need for locking synchronisation at this point.

countvonzero: it's more about consistency than correctness at the moment.
the lock is about data, not about how the program is written (which can change in the future).

my preference is create a separate struct to hold the output and use thread-safe getter/setter to write/read.

piersy: Hi @countvonzero i realised I was mistakenly thinking `outputs` is a map, but its not, its a slice of maps.

So actually there is no need to lock around the writes to `outputs` because each write is updating a different element. That's why the writes to outputs were not triggering any data races. I believe the original version of this PR was correct with respect to the use of `outputs`. So given that no locking is required here I'm wondering if it still makes sense to create a thread-safe struct?

dshulyak: ~it would be more convenient to review if trace with a race was added to the change.~ Found the race in the commit message. it will be sufficient to wait until goroutines have exited

piersy: @dshulyak & @countvonzero if you are agreed on simply having the wait group, would you prefer that I revert the commits that added a mutex or force push the branch back to the initial PR state?

countvonzero: i was wrong. sorry for the trouble. 
yes please revert the commits back to only the waitgroup. @piersy

countvonzero: > i realised I was mistakenly thinking outputs is a map, but its not, its a slice of maps.

you and me both

piersy: Hi @dshulyak & @countvonzero  I've reverted the addition of the mutexes.

countvonzero (approved):

dshulyak (approved):

dshulyak:
bors merge

PR #3527, opened 2022-09-05
[Merged by Bors] - vm: relay to allow execution on a remote state

dshulyak:
part of: https://github.com/spacemeshos/go-spacemesh/issues/3269

previously all method calls were executed on the principal account (same account that pays for gas).
relay allows to execute method on any remote account while paying gas from principal.

for genesis it will be used only for vesting/vault pair that will be introduced in a followup.


genvm/core/context.go
```diff
@@ -15,10 +15,12 @@ import (
 type Context struct {
 	Registry HandlerRegistry
 	Loader   AccountLoader
-	Handler  Handler
-	Template Template
 
-	Account Account
+	LayerID LayerID
```
countvonzero, line 20: what is this layer id associate with?
i cannot see in this PR how this can be used.

and if i remember correctly, the research team doesn't think the vesting account should use layer id for verify.

countvonzero:
> relay allows to execute method on any remote account while paying gas from principal.

can you give me a concrete example of such case in the vesting/vault scenario that require remote execution?

genvm/core/types.go
```diff
@@ -77,6 +89,19 @@ type HandlerRegistry interface {
 	Get(Address) Handler
 }
 
+// Host ...
```
countvonzero: please document for real what this interface is for. is this a regular account that can also relay execution to a remote account? or any account can be Host?

dshulyak:
> can you give me a concrete example of such case in the vesting/vault scenario that require remote execution?

vesting account will be an owner of the vault and will pay for drain vault transaction

countvonzero:
> > can you give me a concrete example of such case in the vesting/vault scenario that require remote execution?
> 
> vesting account will be an owner of the vault and will pay for drain vault transaction

thanks. (and this is why i need a spec...). so is the following mapping correct>
addrVaultOwner -> principal
addrVault -> remote account
addrRecipient -> recipient of the drained amount

if it's correct, how does one know that addrVaultOwner can execute for addrVault?
(maybe pointing me to the code would be better)

dshulyak:
seems correct 

https://github.com/spacemeshos/go-spacemesh/blob/90a7fba4c0494136a76c899e0ba16efaa31e3bc5/genvm/templates/vault/vault.go#L49

genvm/core/context.go
```diff
@@ -69,40 +81,86 @@ func (c *Context) Spawn(args scale.Encodable) error {
...
+
+	c.transferred += amount
+	from.Balance -= amount
+	account.Balance += amount
+	c.change(account)
+	return nil
+}
+
+// Relay call to the remote account.
+func (c *Context) Relay(remoteTemplate, address Address, call func(Host) error) error {
```
countvonzero, line 118: i don't understand this API. why care about remoteTemplate when you can get the template from loading address?

genvm/core/context.go
```diff
@@ -164,9 +225,38 @@ func (c *Context) load(address types.Address) (*Account, error) {
...
+}
+
+// Handler ...
+func (r *RemoteContext) Handler() Handler {
+	return r.handler
+}
+
+// Transfer ...
+func (r *RemoteContext) Transfer(to Address, amount uint64) error {
+	if err := r.transfer(r.remote, to, amount, r.remote.Balance); err != nil {
```
countvonzero: should maxSpend here be `amount` instead?

genvm/core/context.go
dshulyak: this is used to prevent executing remote calls on any other template type

genvm/core/context.go
dshulyak: doesn't matter really, can be amount

genvm/core/types.go
dshulyak: this is not about accounts. this is a set of methods that vm makes available to template

genvm/core/context.go
dshulyak: layer id is used by vault to compute amount that can be drained. not related to verify

genvm/core/context.go
countvonzero: so i guess i had to see the usage (other than in the test). can you point me to the usage in later PR?

if the Principal account can act on `address`, why do we care about which remote template it is?
or is there only one template can be remotely executed?

countvonzero (approved):

genvm/core/context.go
dshulyak: other template types don't have any auth in place. in full svm auth must be much more nuanced compared with what is needed for genesis

https://github.com/spacemeshos/go-spacemesh/blob/90a7fba4c0494136a76c899e0ba16efaa31e3bc5/genvm/templates/vesting/handler.go#L106

genvm/core/context.go
countvonzero: so how do you determine that the vault is properly vested (1yr) before it can be drained?

dshulyak: lets discuss it in vesting pr

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3534, opened 2022-09-07
[Merged by Bors] - consistent active set for hare and tortoise in the network

countvonzero:
## Motivation
in the hare active set calculation. we calculate a safe layer range for each target layer.
in the system test, there are 4 layers per epoch. here is the mapping btwn target layers and their safe layer range (test only ran to layer 41)
```
target layer -> safe layer range
8-32        ->      [7,7]
33-36       ->      [8,8]
37-40       ->      [12,12]
41          ->      [16,16]
```

during systest, each miner has different total weight when calculating the active set for a layer.
hare oracle only starts using hare active set (activeset from ref ballots in safe layer range) from layer 33.
i.e. for the first 9 epochs, it's using tortoise active set (i.e. all ATX in database)

this PR include changes 3 changes
- syncer fetch ATXs from peers at the beginning of every epoch so that nodes have consistent tortoise active set
- remove caching tortoise active set so that it does not pollute the hare active set in future epoch.
- adjusted hare eligibility config for fastnet/systests

### DIFF
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -549,7 +549,6 @@ func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID) (map[ty
 	}
 
 	// if we failed to get a Hare active set, we fall back on reading the Tortoise active set targeting this epoch
-	// TODO: do we want to cache tortoise active set too?
 	if safeLayerStart.GetEpoch().IsGenesis() {
 		logger.With().Debug("no hare active set for genesis layer range, reading tortoise set for epoch instead",
 			targetLayer.GetEpoch())
@@ -575,9 +574,6 @@ func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID) (map[ty
 		activeMap[atxHeader.NodeID()] = atxHeader.GetWeight()
 	}
 	logger.With().Debug("got tortoise active set", log.Int("count", len(activeMap)))
-
-	// update cache and return
-	o.activesCache.Add(targetLayer.GetEpoch(), activeMap)
 	return activeMap, nil
 }
 
--- a/config/presets/fastnet.go
+++ b/config/presets/fastnet.go
@@ -43,6 +43,9 @@ func fastnet() config.Config {
 	conf.LayerDurationSec = 15
 	conf.LayersPerEpoch = 4
 
+	conf.HareEligibility.ConfidenceParam = 2 // half epoch
+	conf.HareEligibility.EpochOffset = 0
+
 	conf.POST.BitsPerLabel = 8
 	conf.POST.K1 = 2000
 	conf.POST.K2 = 4
--- a/hare/eligibility/config/config.go
+++ b/hare/eligibility/config/config.go
@@ -8,5 +8,5 @@ type Config struct {
 
 // DefaultConfig returns the default configuration for the oracle package.
 func DefaultConfig() Config {
-	return Config{25, 0}
+	return Config{ConfidenceParam: 25, EpochOffset: 0}
 }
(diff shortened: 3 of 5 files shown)


countvonzero:
@dshulyak 
the tortoise active set weight discrepancy was causing the new nodes not accepting others' certificate.
my plan is to add ignore list for ATX + trigger tortoise rerun after this PR.

countvonzero:
bors try

dshulyak:
how often active set will be computed without a cache? it is going be computed for every hare message? if so it clearly can't be removed


> during systest, each miner has different total weight when calculating the active set for a layer.

i don't understand how this is possible if syncer downloads all atxs before everything else. i think there was a bug where some nodes were outputting late atxs (one that you fixed). in this case yes it is possible that the weight will be different. this is why we need to use safe layer range for hare as even single atx can make a difference (in tortoise single atx won't make any difference)

also safe layer range doesn't work as intended because systest is using wrong [config](https://github.com/spacemeshos/go-spacemesh/blob/develop/hare/eligibility/config/config.go#L5-L6)

> add ignore list for ATX

what is the idea of ignore list?

hare/eligibility/oracle.go
```diff
@@ -575,9 +574,6 @@ func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID) (map[ty
 		activeMap[atxHeader.NodeID()] = atxHeader.GetWeight()
 	}
 	logger.With().Debug("got tortoise active set", log.Int("count", len(activeMap)))
-
-	// update cache and return
-	o.activesCache.Add(targetLayer.GetEpoch(), activeMap)
```
countvonzero, line 580: @dshulyak 
this key `targetLayer.GetEpoch()` is different from the key used for hare active set `safeLayerStart.GetEpoch()`. this means the tortoise active set can pollute the data for hare active set, since `targetLayer` is always higher than `safeLayerStart`. that's why this cached value was never used in L445 because the epoch value is different.
i.e. the tortoise active set is repeatedly calculated (3 or 4 times per oracle Validate() call) until epoch 9 before the config fix in this PR.

i've considered creating a separate cache for tortoise active set. but now that the config is fixed, only genesis epoch will fall back on tortoise active set. and it seems pointless to create a cache for that.

countvonzero:
> how often active set will be computed without a cache? it is going be computed for every hare message? if so it clearly can't be removed

explained in the other comment where the cache was removed, the epoch used to cache for hare active set and tortoise active set is different. therefore the latter can pollute the former.
 
> > during systest, each miner has different total weight when calculating the active set for a layer.
> 
> i don't understand how this is possible if syncer downloads all atxs before everything else. i think there was a bug where some nodes were outputting late atxs (one that you fixed). in this case yes it is possible that the weight will be different. this is why we need to use safe layer range for hare as even single atx can make a difference (in tortoise single atx won't make any difference)
> 
> also safe layer range doesn't work as intended because systest is using wrong [config](https://github.com/spacemeshos/go-spacemesh/blob/develop/hare/eligibility/config/config.go#L5-L6)

it seems the config fix works. the safe layer range now is consistent across the network.

UPD: i mean the config allows the nodes to use hare active set, which yields consistent weight across the network.
 
> > add ignore list for ATX
> 
> what is the idea of ignore list?

for the network to heal from a split, and to avoid frequent tortoise rerun, Tal proposed to create an ignore list for ATXs.
when an ATX is received late, we put it in a ignore list (maybe just a database boolean).
then tortoise will periodically check whether the ignored ATXs accumulated enough weight to matter. if so, do a rerun considering those ignored ATX weight.

countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

evt (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #3544, opened 2022-09-12
[Merged by Bors] - tortoise: rerun verifying tortoise if hare output changed within hdist

### LINKED ISSUE #3308: implementation/protocol doesn't handle well short network splits
- consider node that goes offline for less then one layer, but for enough time to miss most of the hare
- eventually it will fail and terminate with empty layer. local opinion within hdist = empty layer
- tortoise can't verify such layer using verifying mode, since everyone else votes differently from the local opinion
- at this point node is effectively in a fork with the rest of the network
- it will stay in a fork for hdist layers. or until sync fixes local opinion (right now it doesn't because it considers this layer as terminated) and triggers rerun
- hdist is planned to be 200 layers with 5 min each. node is effectively useless for almost a whole day, so it rather weird to just wait
- rerun is not really cheap even with verifying tortoise, and this way sync will have to trigger it every time when it sees different layer hash

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3308

if hare output changed within hdist (for example from empty to a certified block) node should rerun
verifying tortoise. this will allow it to make progress and cast votes that are aligned with the rest of the network.

code changes:
- refactored hareOutput to a single map that is keyed by layerId instead of blockId, since we always support only one hare output per layer
- added OnHareOutput, that is called every time cert is saved to database


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

tortoise/sim/generator.go
```diff
@@ -247,3 +247,8 @@ func (g *Generator) generateAtxs() {
 		}
 	}
 }
+
+// Layer returns generated layer.
+func (g *Generator) Layer(i int) *types.Layer {
+	return g.layers[i+1]
```
evt, line 253: is it possible we try to access non-existing layer and then it panics?

tortoise/tortoise.go
```diff
@@ -764,9 +762,29 @@ func (t *turtle) onHareOutput(lid types.LayerID, bid types.BlockID) {
 		return
 	}
 	t.logger.With().Debug("on hare output", lid, bid, log.Bool("empty", bid == types.EmptyBlockID))
-	t.decided[lid] = struct{}{}
-	if bid != types.EmptyBlockID {
-		t.hareOutput[bid] = support
+	previous, exists := t.hareOutput[lid]
+	if !exists || (exists && previous != bid) {
```
evt: is it better to have early return like
```
if exists && previous == bid {
  return
}
```
?

blocks/certifier.go
```diff
@@ -426,5 +428,6 @@ func (c *Certifier) save(logger log.Log, lid types.LayerID, cert *types.Certific
 		logger.Error("failed to save block cert", log.Err(err))
 		return err
 	}
+	c.tortoise.OnHareOutput(lid, cert.BlockID)
```
countvonzero, line 431: note that in the current flow, this may get called N times where the node asks N peers for opinions in one shot.
i have a pending change to change that but not yet ready.

tortoise/algorithm.go
```diff
@@ -202,6 +202,8 @@ func (t *Tortoise) EncodeVotes(ctx context.Context, opts ...EncodeVotesOpts) (*t
 
 // HandleIncomingLayer processes all layer block votes
 // returns the old verified layer and new verified layer after taking into account the blocks votes.
+//
+// TODO(dshulyak) open an issue to split this method into.
```
countvonzero: into?

tortoise/tortoise.go
```diff
@@ -764,9 +762,29 @@ func (t *turtle) onHareOutput(lid types.LayerID, bid types.BlockID) {
...
+		t.hareOutput[lid] = bid
+		if lid.Before(t.minprocessed) && t.mode.isVerifying() && withinHdist(t.Config, lid, t.last) {
+			t.logger.With().Info("local opinion changed within hdist",
+				lid,
+				log.Stringer("verified", t.verified),
+				log.Stringer("previous", previous),
+				log.Stringer("new", bid),
+			)
+			t.verifying.resetWeights()
```
countvonzero: why is this needed? all the ballots should remain the same, right?

tortoise/tortoise.go
```diff
@@ -764,9 +762,29 @@ func (t *turtle) onHareOutput(lid types.LayerID, bid types.BlockID) {
...
+
+			// if local opinion within hdist was changed about the layer
+			// that was already verified we need to revert that
+			t.verified = minLayer(t.verified, lid)
+			for target := t.verified; !target.After(t.processed); target = target.Add(1) {
```
countvonzero: does this have the same effect as doing rerun from t.verified layer?
conceptually it's easier to understand that if local opinion changes about hare output for layer N, then we do a rerun from layer N.

blocks/certifier.go
dshulyak: thats fine. a lot of data though

tortoise/tortoise.go
dshulyak: all ballots are re-counted based on the "goodness" computed with new opinion, this method resets previously counted weight

tortoise/tortoise.go
dshulyak: > does this have the same effect as doing rerun from t.verified layer?

yes, this is basically a "cheap" rerun. it counts votes based on new local opinion that changed because of the hare output update

tortoise/sim/generator.go
dshulyak: this is used exclusively for testing
it surely will panic if non-existing layers is requested

tortoise/algorithm.go
dshulyak: that was a leftover. i will describe it in the issue instead

evt (approved):

countvonzero (approved):

dshulyak:
bors merge

PR #3549, opened 2022-09-12
[Merged by Bors] - Avoid scale code generation for ActivationTxHeader

poszu:
## Motivation
I noticed that a file was changed after running `make generate`. It seems that the code was not regenerated after changes introduced in #3519. It actually turns out that scale (de)serialization code should not be generated for it (https://github.com/spacemeshos/go-spacemesh/pull/3549#issuecomment-1243936342).

## Changes
Moved `ActivationTxHeader` to a separate file.

## Test Plan
unittests, systests

## DevOps Notes


dshulyak (dismissed):
@fasmat it looks like a bug, those fields shouldn't be encoded

fasmat:
> @fasmat it looks like a bug, those fields should be encoded

I think ActivationTxHeader shouldn't be encoded at all, since it doesn't go over the wire or does it?

fasmat:
I checked, `ActivationTxHeader` is now only written to and read from DB or generated from an existing `ActivationTx` but not encoded with scale any more.

@poszu To fix this I suggest moving the `ActivationTxHeader` type and its methods to a separate source file so no `Encode`/`Decode` methods are generated for it. Another `make generate` should remove the existing generated methods.

Also `activation_test.go` only contains tests relevant for `VerifiedActivationTx` and `ActivationTxHeader` so renaming that file would make easier to associate the tests with their implementation

fasmat (changes requested):
See my previous comment

poszu:
> I checked, `ActivationTxHeader` is now only written to and read from DB or generated from an existing `ActivationTx` but not encoded with scale any more.
> 
> @poszu To fix this I suggest moving the `ActivationTxHeader` type and its methods to a separate source file so no `Encode`/`Decode` methods are generated for it. Another `make generate` should remove the existing generated methods.
> 
> Also `activation_test.go` only contains tests relevant for `VerifiedActivationTx` and `ActivationTxHeader` so renaming that file would make easier to associate the tests with their implementation

@fasmat, thanks for pointing this out. I moved the struct into a separate file to avoid code generation for it.

poszu:
bors try

fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3550, opened 2022-09-13
[Merged by Bors] - ATX: add API to differentiate publish and target epoch

countvonzero:
## Motivation

for code clarity, explicitly use PublishEpoch() vs TargetEpoch()

## Changes

- move TargetEpoch to challenge struct and add PublishEpoch API
- simplify StoreAtx API

### DIFF
--- a/activation/activation_test.go
+++ b/activation/activation_test.go
@@ -90,7 +90,7 @@ func (n *NetMock) hookToAtxPool(transmission []byte) {
 				if err != nil {
 					panic(err)
 				}
-				if err := atxDb.StoreAtx(context.TODO(), atx.PubLayerID.GetEpoch(), vAtx); err != nil {
+				if err := atxDb.StoreAtx(context.TODO(), vAtx); err != nil {
 					panic(err)
 				}
 			}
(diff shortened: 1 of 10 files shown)


activation/handler.go
```diff
@@ -299,7 +297,7 @@ func (h *Handler) StoreAtx(ctx context.Context, ech types.EpochID, atx *types.Ve
 		delete(h.atxChannels, atx.ID())
 	}
 
-	h.log.WithContext(ctx).With().Info("finished storing atx in epoch", atx.ID(), ech)
+	h.log.WithContext(ctx).With().Info("finished storing atx in epoch", atx.ID(), atx.PublishEpoch())
```
fasmat, line 300: Should we use a formatted string here: "... atx %v in epoch %d" ?

activation/activation_test.go
```diff
@@ -282,9 +282,7 @@ func assertLastAtx(r *require.Assertions, posAtx, prevAtx *types.VerifiedActivat
 }
 
 func storeAtx(r *require.Assertions, atxHdlr *Handler, atx *types.VerifiedActivationTx, lg log.Log) {
```
fasmat: should we inline this function? `storeAtx(r, atxHdlr, atx, nil)` is now just 

`r.NoError(atxHdlr.StoreAtx(context.TODO(), atx))`

fasmat (approved):

activation/handler.go
countvonzero: that's inconsistent with the current practice.
these labeled field is searchable in loki. whereas formatted string can only be regex'ed when searching through log.
Dmitry and i were moving away from the formatted string in logging.

activation/activation_test.go
countvonzero: done.

countvonzero:
bors merge

PR #3554, opened 2022-09-14
[Merged by Bors] - tortoise: revisit thresholds and margin computation 

### LINKED ISSUE #3119: tortoise: compute correct global and local threshold
In the codebase, we default to 6/10 and 2/10 for global and local thresholds respectively. We discussed the rationale behind these values on 1st of February (details are recorded in the call). The issue is to revisit these values based on that information.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3119

see https://community.spacemesh.io/t/requirements-for-local-and-global-thresholds/235 and https://community.spacemesh.io/t/uncounted-weight-and-the-global-threshold/228 

local threshold is selected in a way so that two honest nodes can't be on a different side of the local threshold if they receive inconsistent votes from adversary (e.g. one receives support for the block, and another received votes that are against the block). 1/3 is a safe upper bound for this parameter. 

global threshold is selected so that if honest node crossed global threshold and adversary cancels weight (by assumption 1/3 of total weight) the honest node will still cross the local threshold.

full tortoise verifies layer when votes margin crosses global threshold

verifying tortoise uses conservative margin, we assume that all weight that wasn't counted by a verifying tortoise votes against local opinion. verifying tortoise counts only weight from ballots that vote consistently with local opinion.

other changes:
- some tests were refactored as layers are verified faster

### DIFF
--- a/tortoise/threshold.go
+++ b/tortoise/threshold.go
@@ -10,6 +10,16 @@ import (
 	"github.com/spacemeshos/go-spacemesh/datastore"
 )
 
+var (
+	// by assumption adversarial weight can't be larger than 1/3.
+	adversarialWeightFraction = big.NewRat(1, 3)
+	// nodes should not be on different sides of the local threshold if they receive different adversarial votes.
+	localThresholdFraction = big.NewRat(1, 3)
+	// for comleteness:
+	// global threshold is set in a such way so that if adversary
+	// cancels their weight (adversarialWeightFraction) - honest nodes should still cross local threshold.
+)
+
 func getBallotHeight(cdb *datastore.CachedDB, ballot *types.Ballot) (uint64, error) {
 	atx, err := cdb.GetAtxHeader(ballot.AtxID)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

countvonzero (approved):

dshulyak:
bors merge

tortoise/threshold.go
```diff
@@ -10,6 +10,16 @@ import (
 	"github.com/spacemeshos/go-spacemesh/datastore"
 )
 
+var (
+	// by assumption adversarial weight can't be larger than 1/3.
+	adversarialWeightFraction = big.NewRat(1, 3)
+	// nodes should not be on different sides of the local threshold if they receive different adversarial votes.
+	localThresholdFraction = big.NewRat(1, 3)
+	// for comleteness:
```
poszu, line 18:
```suggestion
	// for completeness:
```

poszu (approved):

PR #3559, opened 2022-09-16
[Merged by Bors] - vm, blocks: integrate rewards from updated economics

### LINKED ISSUE #151: Implement genesis rewards
See #5

# Spec

Rewards are split into two portions, tx-fee-rewards and block-subsidy-rewards. Both portions depend upon knowing the per-coinbase layer weight.

## Calculating weight

Total layer weight $\gamma_\ell$ for layer $\ell$ is defined as the sum of the weights of all valid proposals (i.e., those passed by hare) $P_\ell$ in the layer:

$$ \gamma_\ell \equiv \sum_{p \in P_\ell} w(p)$$

where $w(p)$ is the weight (in accumulated PoET ticks, recorded in the referenced ATX) of valid proposal $p$.

The weight associated with a particular coinbase $c$ in layer $\ell$, $\gamma_{\ell c}$, is calculated as:

$$ \gamma_{\ell c} \equiv \sum_{p \in P_\ell, cb(p) = c} w(p)$$

where $cb(p)$ is the coinbase associated with the proposal (recorded in the referenced ATX).

## tx-fee-rewards

We "smooth" the tx-fee-rewards for each layer among all miners who participated in building that layer (i.e., submitted one or more proposal that was passed by hare). First, we must calculate the weight per coinbase, then the total weight for the layer, and finally, the reward breakdown per smesher.

### Block construction

We cannot store absolute values for tx-fee-rewards in a block, since tx fees are contextual and the fees collected for the txs in the block could change if the block context changes. Instead, we store the coinbase layer weights $\gamma_{\ell c}$ (as defined above).

The coinbase layer weights for the layer should be stored explicitly in the block as a list mapping coinbase addresses to rational numbers $num/denom$ where both $num$ and $denom$ are `uint64`. This is necessary since the weights can only be calculated from valid proposals, which are ephemeral.

### VM execution

To calculate tx-fee-rewards $v_B$ for a given block, the VM must execute each block tx in order, _in situ,_ and sum the total fees paid by each tx:

$$ v_B = \sum_{T \in B_T} T_g$$

where $B_T$ is the set of transactions in block $B$, and $T_g$ is the total gas paid by the transaction (zero if it's ineffective).

To calculate the coinbase layer reward for a given coinbase $v_{\ell c}$, take the total tx-fee-rewards for the block $v_B$, multiply by the coinbase layer weight $\gamma_{\ell c}$ (calculated and stored in the block as described above), and divide by the total layer weight $\gamma_\ell$. The coinbase layer reward $v_{\ell c}$ should be added to the coinbase account balance $\sigma[c]_B$ _at the end of the block_, i.e., after all other block transactions have been applied:

```math
\begin{align}
v_{\ell c} & = & v_B \times \frac{\gamma_{\ell c}}{\gamma_\ell} \\
\sigma'[c]_B & \equiv & \sigma[c] + v_{\ell c}
\end{align}
```

See the note on rounding, below.

### Key math

The Spacemesh economic model provides a function from layer number $\ell$ to the total smidge that should have been issued as block subsidy as of that layer, based on an exponential decay function with a particular half life. The function is as follows:

$$
% GitHub markdown does not support equation numbering so we need to jump through hoops to make this work; due to a bug, using \tag also does not work
\begin{align}
CumulativeSubsidy_\ell \equiv TotalSubsidy \times (1-\exp(-\lambda \ell ))&&\text{(1)}
\end{align}
$$

Note that the layer number $\ell$ is 1-indexed; see note below about how we index layers.

$TotalSubsidy$ includes only the portion of total issuance issued as a block subsidy to miners. In particular, it does not include the portion of issuance in the initial vaults (nor does it include tx-fee-rewards as outlined above). It is calculated as:

$$ TotalSubsidy \equiv TotalIssuance - TotalVaulted$$

where $TotalIssuance$ is set to the total, eventual issuance of 2.4bn smesh and $TotalVaulted$ is set to the total vault amount of 150m smesh; i.e., $TotalSubsidy$ is 2.25bn smesh.

$\lambda$ (lambda) is the _decay parameter_ corresponding to our half life, defined as:

$$
\begin{align}
\lambda \equiv \frac{\ln(2)}{HalfLife}&&\text{(2)}
\end{align}
$$

$HalfLife$ is calculated such that, exactly ten years post-genesis, $TenYearTarget$ (600m smesh, _including the vaulted amount_) has been issued and is circulating; i.e., in addition to $TotalVaulted$, we want to issue precisely $TenYearTarget - TotalVaulted = 600 \times 10^6 - 150 \times 10^6 = 450 \times 10^6$ smesh.

(The value of these constants is set in https://github.com/spacemeshos/economics/blob/main/constants/constants.go.)

$HalfLife$ is denominated in layers and defined to be:

$$ HalfLife \equiv TenYears \times \frac{-\ln(2)}{\ln(1-\frac{TenYearTarget-TotalVaulted}{TotalSubsidy})}$$

We can thus calculate $HalfLife$ and $\lambda$ ([derivation](https://www.wolframalpha.com/input?i=log%282%29%2F%5B1051201+*+%5B-log%282%29%2Flog%281-%28450e6%2F2.25e9%29%29%5D%5D)) as:

$$
\begin{eqnarray}
HalfLife & \equiv & TenYears & \times & \frac{-\ln(2)}{\ln(1-\frac{TenYearTarget-TotalVaulted}{TotalSubsidy})} & \\
HalfLife & \equiv & 1051201 & \times & \frac{-\ln(2)}{\ln(1-\frac{450 \times 10^6}{2.25 \times 10^9})} & \approx 3.265328552 \ldots \times 10^6
\end{eqnarray}
$$

This is between layer 3,265,328 and layer 3,265,329, or approximately [31.04 years post-genesis](https://www.wolframalpha.com/input?i=1051201+*+%5B-log%282%29%2Flog%281-%28450e6%2F2.25e9%29%29%5D*5+minutes).

$$ \lambda = \frac{\ln(2)}{HalfLife} = \frac{\ln(2)}{3.265328552 \ldots \times 10^6} \approx 2.122748659 \ldots \times 10^{-7}$$

### Effective genesis

Spacemesh includes both a _genesis_ layer and an _effective genesis_ layer. The genesis layer is the zeroth layer in the zeroth epoch. The first set of smeshers submits their initial ATX in epoch 1, and become eligible to smesh in epoch 2. No transactions are mined and no blocks produced until epoch 2 (chronologically, the third epoch). Therefore, no rewards are issued in epochs 0 and 1.

The effective genesis layer is the first layer in epoch 2. This is when rewards begin to be issued. **_For the purposes of all rewards-related math and economics, we treat the effective genesis layer as the genesis (zeroth) layer._**

### Layer indexing and handling time

For our purposes, we define one year to be exactly 365 24-hour days, or exactly 105120 layers of 5 minutes each. Note that equation $\text{(1)}$ above assumes that layers are indexed from 1. However, we zero-index layers: layer 0 is the effective genesis layer. We therefore add one to each layer number in equation $\text{(1)}$ so that layers are indexed from 1; otherwise, there would be zero issuance in the first layer (the effective genesis layer). In other words, the $LayerNum$ for the effective genesis layer is 1.

For any abstract time interval, e.g., $OneYear$, this has absolutely no impact: a year is always 105120 layers. However, for any fixed time interval, e.g., ten years post-genesis, we need to add one layer to account for this indexing: "ten years post genesis" is the effective genesis layer (1 layer) plus $10 \times OneYear$. For this reason we set the ten year mark, at which point we want to have achieved the $TenYearTarget$ issuance target, to ten years' worth of layers plus one:

$$ OneYear \equiv 105120$$

$$ TenYears \equiv 10 \times OneYear + 1 = 1051201$$

### Number format

The key constant, $HalfLife$, is defined precisely and may be found [in the tests](https://github.com/spacemeshos/economics/blob/045462b489e85ff948414ba1ec60a16301463071/rewards/rewards_test.go#L131). All other constants, including $\lambda$, can be derived from $HalfLife$ (see equation $\text{(2)}$).

We perform only fixed-point integer arithmetic using 128 bit unsigned integers rounded towards zero, as implemented by https://github.com/ericlagergren/decimal/releases/tag/v3.3.1. We run tests across all relevant platforms and architectures against a suite of hardcoded, expected outputs to ensure that the results of this math are deterministic.

### Subsidy calculation

The formula for the total subsidy issued as of layer $\ell$ is given in equation $\text{(1)}$ above. We can therefore calculate the _additional_ subsidy $R_\ell$ to be issued in $\ell$ in a straightforward fashion as:

$$
R_\ell \equiv
\begin{align}
\lfloor
\begin{cases}
CumulativeSubsidy_{\ell} - CumulativeSubsidy_{\ell-1} & \ell > 0 \\
CumulativeSubsidy_{\ell} & \ell = 0
\end{cases}
\rfloor&\text{(3)}
\end{align}
$$

#### Rounding

Note that, due to the nature of the math in equation $\text{(1)}$, the $CumulativeSubsidy$ values will contain a fractional component. Note the rounding (down to the nearest integer units of smidge) in equation $\text{(3)}$, the per-layer subsidy calculation. In particular, the rounding happens _after_ the layer-to-layer difference is calculated, which means that rounded-off amounts will be accounted for in later layers.

#### Tail issuance

With exponential decay, issuance approaches zero asymptotically but never actually reaches zero. For this reason, asking, "When does issuance cease?" is not a well-defined question. We can be more precise and provide three answers.

We can calculate the layer when the final _full smidge_ and _full smesh_ are issued using the derivative of equation $\text{(1)}$:

$$
\begin{align}
CumulativeSubsidy_\ell & \equiv TotalSubsidy \times (1-\exp(-\lambda \ell )) \\
f(\ell) & = TotalSubsidy \times (1-\exp(-\lambda \ell )) \\
f'(\ell) & = TotalSubsidy \times \lambda \exp(-\lambda \ell) = 1 \\
\ell & = \frac{ln(\lambda \times TotalSubsidy)}{\lambda} \\
\ell_{LastFullSmesh} \equiv \ell & = 1 \text{ Smesh} \\
\ell_{LastFullSmidge} \equiv \ell & = 1 \text{ Smidge}
\end{align}
$$

The last full smesh will be issued in layer 29,060,492, approximately [276 years](https://www.wolframalpha.com/input?i=29060492+*+5+minutes) post-genesis, and the last full smidge will be issued in layer 126,685,172, approximately [1204 years](https://www.wolframalpha.com/input?i=126685172+*+5+minutes) post-genesis. The full derivations of the above can be found here: [last smesh](https://www.wolframalpha.com/input?i=log%28%5Blog%282%29%2F%5B1051201+*+%5B-log%282%29%2Flog%281-%28450e6%2F2.25e9%29%29%5D%5D%5D*2.25e9%29%2F%5Blog%282%29%2F%5B1051201+*+%5B-log%282%29%2Flog%281-%28450e6%2F2.25e9%29%29%5D%5D%5D), [last smidge](https://www.wolframalpha.com/input?i=log%28%5Blog%282%29%2F%5B1051201+*+%5B-log%282%29%2Flog%281-%28450e6%2F2.25e9%29%29%5D%5D%5D*2.25e18%29%2F%5Blog%282%29%2F%5B1051201+*+%5B-log%282%29%2Flog%281-%28450e6%2F2.25e9%29%29%5D%5D%5D).

However, due to the way per-layer subsidy issuance is defined as the difference between the previous layer and the current layer (equation $\text{(3)}$), even after per-layer issuance drops below one smidge (the minimum unit of value), issuance will continue (with less than one smidge issued per layer on average) until the _total amount left to be issued_ falls below one smidge. After this point, no further issuance will occur since the amount will always round down to zero. We call this layer $\ell_{Final}$.

Note that this also means that the last smidge will never actually be issued, and the final, actual total issuance will in fact be one smidge less than $TotalIssuance$. $\ell_{Final}$ can be calculated by solving for $\ell$ in equation $\text{(1)}$:

$$
\begin{align}
CumulativeSubsidy_\ell & \equiv TotalSubsidy \times (1-\exp(-\lambda \ell )) \\
\ell & = -\frac{ln(1-\frac{CumulativeSubsidy_\ell}{TotalSubsidy})}{\lambda} \\
\ell_{Final} & = -\frac{ln(1-\frac{TotalSubsidy-1}{TotalSubsidy})}{\lambda}
\end{align}
$$

$\ell_{Final}$ is [199,069,549](https://www.wolframalpha.com/input?i=-log%281-%282249999999999999999%2F2250000000000000000%29%29%2F%5Blog%282%29%2F%5B1051201+*+%5B-log%282%29%2Flog%281-%28450e6%2F2.25e9%29%29%5D%5D%5D), which is approximately [1,892 years](https://www.wolframalpha.com/input?i=199069549+*+5+minutes) post-genesis.

### Per-coinbase subsidy

The total layer subsidy $R_\ell$ should be split among the layer coinbases, weighted in the same way as in tx-fee-rewards, above. The _sum of the tx-fee-rewards and block-subsidy-rewards_ should be rounded down to the nearest smidge _after_ the two are summed and _before_ being added to the coinbase balance. These rounded-down "dust" smidge units are effectively burnt. Thus the subsidy awarded to a given coinbase in a given layer $R_{\ell c}$ is calculated as:

```math
\begin{align}
R_{\ell c} & = & R_B \times \frac{\gamma_{\ell c}}{\gamma_\ell} \\
\sigma'[c]_B & \equiv & \lfloor \sigma[c] + v_{\ell c} + R_{\ell c} \rfloor
\end{align}
```

(for avoidance of doubt, if smesher A registered coinbase X, and smeshers B and C registered coinbase Y, and coinbase X is due 100.5 smidge, and coinbase Y is due 100.3 smidge from smesher B and 100.8 smidge from smesher C, then ultimately coinbase X will receive 100 smidge and coinbase Y will receive 201 smidge for the layer; the remaining 0.5 + 0.1 = 0.6 smidge is burnt. the rounding down happens _per coinbase_ not _per smesher._)

### Block construction

Unlike tx-fee-rewards, which are contextual (the outcome of a transaction depends upon the transactions that came before it, which can change), block-subsidy-rewards are not contextual and will never change for a given block or layer. For this reason, unlike tx-fee-rewards, they could be added to the block as a set of explicit transactions (one per coinbase). However, this is unnecessary since they may be trivially calculated from the set of coinbase layer weights, included in the block, and the block total subsidy, which is a pure function of the layer number.

### Empty layers

Note that the algorithm defined here (and the implementation, below) has no context about the actual conditions of the blockchain, e.g., whether or not Hare succeeded, whether any blocks were successfully produced for a layer, what the network conditions are, etc. As such, the total subsidy values for a given layer $R_\ell$ should be considered a "platonic, ideal, best case scenario", or in other words, a ceiling on the total issuance. The most obvious example where less effective issuance will occur is an empty layer: in this case, since the protocol failed to come to consensus on the contents of a layer, no subsidy issuance can occur, and total effective issuance will be less than this "platonic ideal." _No affordance is made to increase issuance in later layers to account for issuance that failed to occur on older layers_ and in such cases the issuance that failed to be effected should be considered burnt from an economic perspective.

### Code spec

The block-subsidy-rewards portion of this specification has been implemented in https://github.com/spacemeshos/economics, which may be called directly: `rewards.TotalAccumulatedSubsidyAtLayer` corresponds to equation $\text{(1)}$ and `rewards.TotalSubsidyAtLayer` corresponds to equation $\text{(3)}$. (Note that the code spec does _not_ include the tx-fee-rewards, nor the weight calculation.)

# Dev task breakdown

Weight calculation and tx-fee-rewards code is already in place. The only thing to be added is block-subsidy-rewards:

- [ ] `genvm/vm.go`
  - [ ] `Apply()`: `blockRewards` should never be empty (this would cause issuance to be wrong)
- [ ] `genvm/rewards.go`
  - [ ] remove `cfg.BaseReward`
  - [ ] import github.com/spacemeshos/economics
  - [ ] modify `calculateLayerReward` to receive layerID and pass this layer (minus effective genesis) into `rewards.TotalSubsidyAtLayer`
- [x] https://github.com/spacemeshos/go-spacemesh/issues/3408
- [ ] tests
  - [ ] issuance at genesis layer
  - [ ] issuance at effective genesis
  - [ ] random sampling of issuance layer to layer
  - [ ] test round down of "dust" in layer subsidy
  - [ ] test that rounding happens per coinbase not per smesher
  - [ ] test rounding of fees

### LINKED ISSUE #3251: blocks: invalidate block without rewards (or leave everything as it is)
Do we consider block with empty rewards as syntactically invalid? If yes - how is it different from a block with a single reward, to a completely random address?

There is no incentive for creating such block, but on the code level we need to have a defined handling for it. Specifically if block with empty rewards is invalid:
- any block with empty rewards should be discarded in sync
- vm should not apply the layer without rewards

if such block is valid:
- sync should save such block normally
- vm should apply layer without rewards

### LINKED ISSUE #3408: Replace float64 with fixed point arithmetic
We're already using fixed-point arithmetic in hare, see hare/eligibility/oracle.go.

The only place I'm aware of this being an issue today is rewards:

https://github.com/spacemeshos/go-spacemesh/blob/5234b07a2bff387ee83d3f5d7ef4ec66713333d3/genvm/rewards.go#L67-L75

But this code will be replaced anyway as part of https://github.com/spacemeshos/pm/issues/151.

There's nothing wrong with rounding down to the nearest smidge, but without fixed-point arithmetic we cannot be sure that precisely the same operation will happen regardless of platform, architecture, etc.

Double-check that we're not doing this anywhere else.

dshulyak:
closes: https://github.com/spacemeshos/pm/issues/151
closes: https://github.com/spacemeshos/go-spacemesh/issues/3408
closes: https://github.com/spacemeshos/go-spacemesh/issues/3251

this change enforces next rules for block syntactic validity:
- every block must have at least one reward (empty layers are an obvious exception)
- to make block more compact there should be at most one reward per coinbase (smeshers are not recorded in rewards) - this is already like that in the code that generates block, i am adding a validation for it
- weight is recorded as fractional int using two uint64 (num/denom), if any of those is 0 - block is syntactically invalid

total rewards is a sum of fees and subsidy (subsidy is computed using github.com/spacemeshos/economics). total rewards are splited between coinbases recorded in the block. each coinbase share is a relative to the total weight recorded in the block. share is computed using big.Rat from stdlib on integer values.

absolute value is computed by multiplying total reward by big.Rat numerator and divides by denominator using big.Int from stdlib (no floating operations).

### DIFF
--- a/blocks/handler.go
+++ b/blocks/handler.go
@@ -15,8 +15,9 @@ import (
 )
 
 var (
-	errMalformedData = errors.New("malformed data")
-	errDuplicateTX   = errors.New("duplicate TxID in proposal")
+	errMalformedData  = errors.New("malformed data")
+	errInvalidRewards = errors.New("invalid rewards")
+	errDuplicateTX    = errors.New("duplicate TxID in proposal")
 )
 
 // Handler processes Block fetched from peers during sync.
@@ -65,7 +66,7 @@ func (h *Handler) HandleSyncedBlock(ctx context.Context, data []byte) error {
 	b.Initialize()
 
 	if err := vm.ValidateRewards(b.Rewards); err != nil {
-		return err
+		return fmt.Errorf("%w: %s", errInvalidRewards, err.Error())
 	}
 
 	logger = logger.WithFields(b.ID())
--- a/blocks/handler.go
+++ b/blocks/handler.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/spacemeshos/go-spacemesh/common/types"
+	vm "github.com/spacemeshos/go-spacemesh/genvm"
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/sql"
 	"github.com/spacemeshos/go-spacemesh/sql/blocks"
@@ -60,9 +61,13 @@ func (h *Handler) HandleSyncedBlock(ctx context.Context, data []byte) error {
 		logger.With().Error("malformed block", log.Err(err))
 		return errMalformedData
 	}
-
 	// set the block ID when received
 	b.Initialize()
+
+	if err := vm.ValidateRewards(b.Rewards); err != nil {
+		return err
+	}
+
 	logger = logger.WithFields(b.ID())
 
 	if exists, err := blocks.Has(h.db, b.ID()); err != nil {
--- a/blocks/handler_test.go
+++ b/blocks/handler_test.go
@@ -41,6 +41,9 @@ func createBlockData(t *testing.T, layerID types.LayerID, txIDs []types.Transact
 		InnerBlock: types.InnerBlock{
 			LayerIndex: layerID,
 			TxIDs:      txIDs,
+			Rewards: []types.AnyReward{
+				{Weight: types.RatNum{Num: 1, Denom: 1}},
+			},
 		},
 	}
 	block.Initialize()
@@ -51,11 +54,14 @@ func createBlockData(t *testing.T, layerID types.LayerID, txIDs []types.Transact
 
 func Test_HandleBlockData_MalformedData(t *testing.T) {
 	th := createTestHandler(t)
-	layerID := types.NewLayerID(99)
-	txIDs := createTransactions(t, max(10, rand.Intn(100)))
+	assert.ErrorIs(t, th.HandleSyncedBlock(context.TODO(), []byte("malformed")), errMalformedData)
+}
 
-	_, data := createBlockData(t, layerID, txIDs)
-	assert.ErrorIs(t, th.HandleSyncedBlock(context.TODO(), data[1:]), errMalformedData)
+func Test_HandleBlockData_InvalidRewards(t *testing.T) {
+	th := createTestHandler(t)
+	buf, err := codec.Encode(&types.Block{})
+	require.NoError(t, err)
+	require.Error(t, th.HandleSyncedBlock(context.TODO(), buf))
 }
 
 func Test_HandleBlockData_AlreadyHasBlock(t *testing.T) {
--- a/blocks/handler_test.go
+++ b/blocks/handler_test.go
@@ -61,7 +61,7 @@ func Test_HandleBlockData_InvalidRewards(t *testing.T) {
 	th := createTestHandler(t)
 	buf, err := codec.Encode(&types.Block{})
 	require.NoError(t, err)
-	require.Error(t, th.HandleSyncedBlock(context.TODO(), buf))
+	require.ErrorIs(t, th.HandleSyncedBlock(context.TODO(), buf), errInvalidRewards)
 }
 
 func Test_HandleBlockData_AlreadyHasBlock(t *testing.T) {
--- a/genvm/rewards.go
+++ b/genvm/rewards.go
@@ -7,6 +7,7 @@ import (
 	erewards "github.com/spacemeshos/economics/rewards"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/common/util"
 	"github.com/spacemeshos/go-spacemesh/genvm/core"
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/sql"
@@ -19,23 +20,14 @@ func ValidateRewards(rewards []types.AnyReward) error {
 		return fmt.Errorf("empty rewards")
 	}
 	unique := map[core.Address]struct{}{}
-	base := rewards[0].Weight.Denom
-	total := uint64(0)
 	for _, reward := range rewards {
 		if reward.Weight.Num == 0 || reward.Weight.Denom == 0 {
 			return fmt.Errorf("reward with invalid (zeroed) weight (%d/%d) included into the block for %v", reward.Weight.Num, reward.Weight.Denom, reward.Coinbase)
-		}
-		if base != reward.Weight.Denom {
-			return fmt.Errorf("all rewards must have the same base %d != %d", base, reward.Weight.Denom)
 		}
 		if _, exists := unique[reward.Coinbase]; exists {
 			return fmt.Errorf("multiple rewards for the same coinbase %v", reward.Coinbase)
 		}
 		unique[reward.Coinbase] = struct{}{}
-		total += reward.Weight.Num
-	}
-	if total != base {
-		return fmt.Errorf("total %d is inconsistent with a base %d", total, base)
 	}
 	return nil
 }
@@ -46,15 +38,18 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
 		subsidy                     = erewards.TotalSubsidyAtLayer(layersAfterEffectiveGenesis)
 		total                       = subsidy + fees
 		transferred                 uint64
+		totalWeight                 = util.WeightFromUint64(0)
 	)
 	for _, blockReward := range blockRewards {
-		fraction := new(big.Int).SetUint64(blockReward.Weight.Num)
-		base := new(big.Int).SetUint64(blockReward.Weight.Denom)
+		totalWeight.Add(util.WeightFromNumDenom(blockReward.Weight.Num, blockReward.Weight.Denom))
+	}
+	for _, blockReward := range blockRewards {
+		relative := util.WeightFromNumDenom(blockReward.Weight.Num, blockReward.Weight.Denom).Div(totalWeight)
 
 		totalReward := new(big.Int).SetUint64(total)
 		totalReward.
-			Mul(totalReward, fraction).
-			Quo(totalReward, base)
+			Mul(totalReward, relative.Num()).
+			Quo(totalReward, relative.Denom())
 		if !totalReward.IsUint64() {
 			return fmt.Errorf("%w: total reward %v for %v overflows uint64",
 				core.ErrInternal, totalReward, blockReward.Coinbase)
@@ -62,8 +57,8 @@ func (v *VM) addRewards(lctx ApplyContext, ss *core.StagedCache, tx *sql.Tx, fee
 
 		subsidyReward := new(big.Int).SetUint64(subsidy)
 		subsidyReward.
-			Mul(subsidyReward, fraction).
-			Quo(subsidyReward, base)
+			Mul(subsidyReward, relative.Num()).
+			Quo(subsidyReward, relative.Denom())
 		if !subsidyReward.IsUint64() {
 			return fmt.Errorf("%w: subsidy reward %v for %v overflows uint64",
 				core.ErrInternal, subsidyReward, blockReward.Coinbase)
--- a/genvm/rewards.go
+++ b/genvm/rewards.go
@@ -2,79 +2,112 @@ package vm
 
 import (
 	"fmt"
-	"math"
+	"math/big"
+
+	erewards "github.com/spacemeshos/economics/rewards"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
-	"github.com/spacemeshos/go-spacemesh/common/util"
+	"github.com/spacemeshos/go-spacemesh/genvm/core"
 	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/sql"
+	"github.com/spacemeshos/go-spacemesh/sql/rewards"
 )
 
-// Config defines the configuration options for Spacemesh rewards.
-type Config struct {
-	GasLimit          uint64
-	StorageCostFactor uint64 `mapstructure:"vm-storage-cost-factor"`
-	BaseReward        uint64 `mapstructure:"base-reward"`
-}
-
-// DefaultConfig returns the default RewardConfig.
-func DefaultConfig() Config {
-	return Config{
-		GasLimit:          100_000_000,
-		StorageCostFactor: 2,
-		BaseReward:        50 * uint64(math.Pow10(12)),
+// ValidateRewards syntactically validates rewards.
+func ValidateRewards(rewards []types.AnyReward) error {
+	if len(rewards) == 0 {
+		return fmt.Errorf("empty rewards")
 	}
-}
-
-func calculateLayerReward(cfg Config) uint64 {
-	// todo: add inflation rules here
-	return cfg.BaseReward
-}
-
-// calculateRewards splits layer rewards and total fees fairly between coinbases recorded in rewards.
-func calculateRewards(logger log.Log, cfg Config, lid types.LayerID, totalFees uint64, rewards []types.AnyReward) ([]*types.Reward, error) {
-	logger = logger.WithFields(lid)
-	totalWeight := util.WeightFromUint64(0)
-	byCoinbase := make(map[types.Address]util.Weight)
+	unique := map[core.Address]struct{}{}
+	base := rewards[0].Weight.Denom
+	total := uint64(0)
 	for _, reward := range rewards {
-		weight := util.WeightFromNumDenom(reward.Weight.Num, reward.Weight.Denom)
-		logger.With().Debug("coinbase weight", reward.Coinbase, log.Stringer("weight", weight))
-		totalWeight.Add(weight)
-		if _, ok := byCoinbase[reward.Coinbase]; ok {
-			byCoinbase[reward.Coinbase].Add(weight)
-		} else {
-			byCoinbase[reward.Coinbase] = weight
+		if reward.Weight.Num == 0 || reward.Weight.Denom == 0 {
+			return fmt.Errorf("reward with invalid (zeroed) weight (%d/%d) included into the block for %v", reward.Weight.Num, reward.Weight.Denom, reward.Coinbase)
 		}
+		if base != reward.Weight.Denom {
+			return fmt.Errorf("all rewards must have the same base %d != %d", base, reward.Weight.Denom)
+		}
+		if _, exists := unique[reward.Coinbase]; exists {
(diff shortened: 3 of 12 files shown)


genvm/templates/vault/vault.go
```diff
@@ -41,7 +41,7 @@ func (v *Vault) Available(lid core.LayerID) uint64 {
 	}
 	incremental := new(big.Int).SetUint64(v.TotalAmount - v.InitialUnlockAmount)
 	incremental.Mul(incremental, new(big.Int).SetUint64(uint64(lid.Difference(v.VestingStart))))
-	incremental.Div(incremental, new(big.Int).SetUint64(uint64(v.VestingEnd.Difference(v.VestingStart))))
+	incremental.Quo(incremental, new(big.Int).SetUint64(uint64(v.VestingEnd.Difference(v.VestingStart))))
```
dshulyak, line 44: this is for consistency with rewards, Div and Quo works the same for uint64

dshulyak:
bors try

genvm/vm.go
```diff
@@ -37,6 +37,20 @@ func WithLogger(logger log.Log) Opt {
 	}
 }
 
+// Config defines the configuration options for Spacemesh rewards.
```
countvonzero: update comment

blocks/handler.go
```diff
@@ -60,9 +61,13 @@ func (h *Handler) HandleSyncedBlock(ctx context.Context, data []byte) error {
 		logger.With().Error("malformed block", log.Err(err))
 		return errMalformedData
 	}
-
 	// set the block ID when received
 	b.Initialize()
+
+	if err := vm.ValidateRewards(b.Rewards); err != nil {
```
countvonzero, line 68: why do it here instead of after the de-duplication check?

blocks/handler.go
```diff
@@ -60,9 +61,13 @@ func (h *Handler) HandleSyncedBlock(ctx context.Context, data []byte) error {
...
+		return err
```
countvonzero: using `fmt.Errorf` for cross-module error?

blocks/handler_test.go
```diff
@@ -51,11 +54,14 @@ func createBlockData(t *testing.T, layerID types.LayerID, txIDs []types.Transact
...
+	assert.ErrorIs(t, th.HandleSyncedBlock(context.TODO(), []byte("malformed")), errMalformedData)
+}
 
-	_, data := createBlockData(t, layerID, txIDs)
-	assert.ErrorIs(t, th.HandleSyncedBlock(context.TODO(), data[1:]), errMalformedData)
+func Test_HandleBlockData_InvalidRewards(t *testing.T) {
+	th := createTestHandler(t)
+	buf, err := codec.Encode(&types.Block{})
+	require.NoError(t, err)
+	require.Error(t, th.HandleSyncedBlock(context.TODO(), buf))
```
countvonzero: i think we should best-effort check for exact error here if it's not too much trouble.
defining an new error for invalid rewards?

genvm/rewards.go
```diff
@@ -2,79 +2,107 @@ package vm
...
+				core.ErrInternal, totalReward, blockReward.Coinbase)
+		}
+
+		subsidyReward := new(big.Int).SetUint64(subsidy)
+		subsidyReward.
+			Mul(subsidyReward, relative.Num()).
+			Quo(subsidyReward, relative.Denom())
+		if !subsidyReward.IsUint64() {
+			return fmt.Errorf("%w: subsidy reward %v for %v overflows uint64",
+				core.ErrInternal, subsidyReward, blockReward.Coinbase)
```
countvonzero, line 64: looks like you can factor out an utility method to do the division and check uint64

genvm/rewards.go
```diff
@@ -2,79 +2,107 @@ package vm
...
+		if err := ss.Update(account); err != nil {
+			return fmt.Errorf("%w: %s", core.ErrInternal, err.Error())
+		}
+		transferred += totalReward.Uint64()
 	}
-	return finalRewards, nil
+	v.logger.With().Debug("rewards for layer",
+		lctx.Layer,
+		log.Uint32("after genesis", layersAfterEffectiveGenesis),
+		log.Uint64("subsidy estimated", subsidy),
```
countvonzero, line 97: why the term "estimated" here and for total?

countvonzero (approved):

countvonzero (commented):
- can also close https://github.com/spacemeshos/go-spacemesh/issues/3251
- i think we should burn the subsidy if there are empty layers

blocks/handler.go
dshulyak: because syntactic validation is cheap. i can move it afterwards, doesn't matter much to me

genvm/rewards.go
dshulyak: it is looks cleaner without such method

genvm/rewards.go
dshulyak: actual amount computed from how much was transfered

dshulyak:
> i think we should burn the subsidy if there are empty layers

they are already burned

dshulyak:
bors merge

countvonzero:
> > i think we should burn the subsidy if there are empty layers
> 
> they are already burned

@dshulyak sorry i missed that. where does that happen?

dshulyak:
> @dshulyak sorry i missed that. where does that happen?

what do you mean? rewards that are not transferred are burned. so in case of empty layer - rewards are always burned

countvonzero:
> > @dshulyak sorry i missed that. where does that happen?
> 
> what do you mean? rewards that are not transferred are burned. so in case of empty layer - rewards are always burned

but we don't call vm.Apply() on types.EmptyBlockID

dshulyak:
> but we don't call vm.Apply() on types.EmptyBlockID

burning doesn't modify any state

countvonzero:
> > but we don't call vm.Apply() on types.EmptyBlockID
> 
> burning doesn't modify any state

oh ok. its in the math (num layers since genesis). i get it now. thanks

genvm/rewards.go
lrettig: I don't understand - "estimated" here means "before rounding"? or something else?

dshulyak: before rounding

PR #3561, opened 2022-09-16
[Merged by Bors] - Properly shot down harness for poet

### LINKED ISSUE #3562: PoET isn't properly shut down after tests
## Description
PoET instance started in tests isn't properly shut down at the end of the testrun

## Steps to reproduce
1. Execute tests in `go-spacemesh` repository
2. Observe a process running after the tests complete (`poet`) that is listening on port `18550` and 18551`

## Actual Behavior
At the end of the test run processes are running that were stated by tests.

## Expected Behavior
At the end of the test run no process that was started for a test should still be running

## Environment
 - OS: Linux arm64
 - Node Version: develop as of 2022-09-15

fasmat:
## Motivation
Fix for a flacky test and cleanup PoET tests correctly (fixes #3562)

## Changes
Shutdown poet properly at the end of the test, make a blocking dial to the server.

## Test Plan
Test still passes

## TODO

## DevOps Notes


activation/nipost_test.go
```diff
@@ -205,19 +206,21 @@ func buildNIPost(tb testing.TB, r *require.Assertions, postCfg PostConfig, nipos
...
 	var postProvider PostSetupProvider
 	postProvider, err = NewPostSetupManager(minerID, postCfg, logtest.New(tb))
 	r.NoError(err)
 	r.NotNil(postProvider)
-	defer func() {
+	tb.Cleanup(func() {
 		r.NoError(postProvider.LastError())
 		r.NoError(postProvider.StopSession(true))
-	}()
+	})
```
poszu, line 223: Just being curious - why plain `defer` was not enough?

cmd/node/node_test.go
```diff
@@ -426,49 +426,28 @@ func TestSpacemeshApp_GrpcService(t *testing.T) {
...
-
-	// We expect this one to fail
-	_, err = c.Echo(context.Background(), &pb.EchoRequest{
-		Msg: &pb.SimpleString{Value: message},
-	})
-	r.Error(err)
-	r.Contains(err.Error(), "rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp")
-	r.NoError(conn.Close())
+	conn, err := grpc.Dial(fmt.Sprintf("localhost:%d", port), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock(), grpc.WithTimeout(3*time.Second))
+	r.ErrorContains(err, "context deadline exceeded")
```
poszu: This check verifies it there hadn't been already _something_ leasing on that port. In this case 3s timeout seems excesive (as the test will wait for 3 seconds on **happy path**). It will increase run time of the test suite. Perhaps you could shrink it down to, let's say 100ms?

fasmat:
bors try

cmd/node/node_test.go
```diff
@@ -426,49 +426,28 @@ func TestSpacemeshApp_GrpcService(t *testing.T) {
...
-
-	require.Eventually(t, func() bool {
-		var err error
-		conn, err = grpc.Dial(fmt.Sprintf("localhost:%d", port), grpc.WithTransportCredentials(insecure.NewCredentials()))
-		return err == nil
-	}, 2*time.Second, 100*time.Millisecond)
-	defer conn.Close()
+	r.True(app.Config.API.StartNodeService)
 
+	conn, err = grpc.Dial(fmt.Sprintf("localhost:%d", port), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
```
poszu: What do you plan for handling timeouts (`grpc.Dial()` here will wait for connection indefinetely)? 

I suggest to use `grpc.WithTimeout(time.Second*5)` option.

poszu (commented):
Great! Thanks :)

activation/nipost_test.go
fasmat: - The time of execution and the guarantees are different: `tb.Cleanup` is executed at the end of the test, `defer` when the function returns.
- `tb.Cleanup` is also executed when a test panics, `defer` isn't (unless one `recover`s withn the `defer`). 
-  `tb.Cleanup` gives an indication about the intend of the code

Here in this specific case it probably doesn't matter if one uses `t.Cleanup` or `defer`.

cmd/node/node_test.go
fasmat: Yes, before the changes a lazy connection was opened and then node client was initialised with that. The assertion then checked if sending the first message failed because the then triggered grpc dialup failed, if something else would have listened on that port the test would fail because it would have had received a different error and not the one expected.

So the behaviour of the test is still the same: if something is listening on `port` (1242 here) the test fails. Independently of if this is a spacemesh node or something else.

Arguably this check can be removed completely because it doesn't test for correct behaviour of our code but if the environment where it is executed fits some "arbitrary" restrictions (nothing is listening on 1242).

cmd/node/node_test.go
fasmat: I'll add timeouts to the Dials with the same values as the `r.Eventually` used before.

poszu (approved):

fasmat:
bors merge

PR #3567, opened 2022-09-19
[Merged by Bors] - Add ATX with highest perceived height to PoST commitment

### LINKED ISSUE #3556: Add positioning ATX to commitment for PoST initialisation
## Description

- When starting a session in the `PostSetupManager` PoST is initialised with the VRF public key (or `NodeId`). Instead it should be initialised with the hash digest of the concatenation of the `ID` of the positioning ATX and the `NodeId`.
- The chosen positioning ATX needs to be added to the initial ATX as part of the `NIPostChallenge`.
  - For incoming initial ATXs check that they have this field set and that it references an existing valid ATX. For all other ATXs ensure that this feed is not set.
- When verifying a PoST use the hash digest of the concatenation of the `ID` of the positioning ATX and the `NodeId` as input instead of just the `NodeId`.

## Implementation hints

- PoST initialisation https://github.com/spacemeshos/go-spacemesh/blob/1c6c817f3797dedf65921cc87baf04c7c65b80d6/activation/post.go#L263 
- PoST verification: https://github.com/spacemeshos/go-spacemesh/blob/1c6c817f3797dedf65921cc87baf04c7c65b80d6/activation/handler.go#L209
- Selecting an positioning ATX: https://github.com/spacemeshos/go-spacemesh/blob/1c6c817f3797dedf65921cc87baf04c7c65b80d6/activation/activation.go#L437
- ATX validation: https://github.com/spacemeshos/go-spacemesh/blob/1c6c817f3797dedf65921cc87baf04c7c65b80d6/activation/handler.go#L214

fasmat:
## Motivation
Closes #3556 

## Changes
- a "commitment" ATX is chosen when PoST is initialised
  - this new ATX is referenced in the `NiPoSTChallenge` for the initial ATX of a node
  - incoming ATX are verified to have a commitment ATX when they are initial and verified to not have one if they are not the first ATX of a Node
- Instead of only using the Nodes public key as commitment for post, the digest of the concatenation of public key and the ID of the chosen commitment ATX are used
  - this requires after determining which ATX to use for the commitment to store it in the local DB, since it is reused for the creation of NiPoST challenges.
- I did some cleanup of the tests in the `activation` package
  - `handler_test.go`: now uses `t.Run` to isolate various test cases from each other more clearly
  -  imho there are still too many tests cases within single tests (especially for `SyntacticallyValidate`)
  - there is too much shared setup between tests in the activation package in general, which makes it hard to modify or add tests without breaking several existing tests
  - For the last two points I created the following issue to clean up the code: https://github.com/spacemeshos/go-spacemesh/issues/3590

## Test Plan
- All existing tests still pass / are updated to incorporate changes
- New tests where added for new functions / methods and to test new behaviour

## TODO

## DevOps Notes


fasmat:
bors try

activation/activation.go
```diff
@@ -106,14 +105,15 @@ type Builder struct {
 	postSetupProvider PostSetupProvider
 	challenge         *types.NIPostChallenge
 	initialPost       *types.Post
+	commitmentAtx     *types.ATXID
```
poszu: I understand that this field works as a cache and it should never be accessed outside of `func getCommitmentAtx()`, right? How about leaving a comment here to reduce the chances of somebody mistakenly using `b.commitmentAtx` in the future instead of `b.commitmentAtx()`? Like:
```suggestion
	// A cached value, meant to be accessed from `commitmentAtx()` only.
	// Use the getter method to access its value.
	commitmentAtx     *types.ATXID
```

Alternatively, and even better, the logic of getting and caching the commitment ATX could be factored out to a separate structure (in a separate pkg to prevent access to its private fields) which sole responsibility would be getting and caching a value. The suggested comment would be absolutely redundant then. Consider an implementation I quickly sketched out:
```go
package cached

type CacheInitializer[T any] func() (*T, error)

type Cached[T any] struct {
	cached      *T
	initializer CacheInitializer[T]
}

func New[T any](initializer CacheInitializer[T]) Cached[T] {
	return Cached[T]{
		cached:      nil,
		initializer: initializer,
	}
}

func (self *Cached[T]) Get() (*T, error) {
	if self.cached == nil {
		value, err := self.initializer()
		if err != nil {
			return nil, err
		}
		self.cached = value
	}

	return self.cached, nil
}
```

and stick a `commitmentAtx Cached[types.ATXID]` into `Builder` + provide it with a closure to run the DB query. What do you think?

fasmat: We already have ways of caching values in our application that are concurrency safe (e.g. cachedDB), I would prefer using / extending those over implementing another solution.

The suggested change would probably work, but complicates testing a bit. I don't think accessing `commitmentAtx` directly within the `activation` package is an issue. Currently it behave the same as other fields of the builder some of which also never change during the lifetime of it and are not protected from direct access: `goldenATXID`, `coinbaseAccount`, `nodeID`, `initialPost`, etc.

I intentionally did not include any bigger refactorings to limit the size of this PR; the only bigger re-writes I did was re-implementing tests since fixing the existing ones would have been more effort than just implementing all tests anew.

fasmat:
bors try

fasmat:
bors try

Makefile
```diff
@@ -177,12 +169,16 @@ golangci-lint-github-action:
...
+	  $(foreach pkg,$(UNIT_TESTS),\
 		go test -coverprofile=cover.out -covermode=count $(pkg);\
 		tail -n +2 cover.out >> cover-all.out;)
 	go tool cover -html=cover-all.out
 .PHONY: cover
 
+cover-profile:
+	@$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" go test -coverprofile=cover.out -timeout 0 -p 1 $(UNIT_TESTS)
+.PHONY: cover-profile
+
```
countvonzero: please remove the lint/coverage/ci.yml change from this PR.

activation/handler.go
```diff
@@ -202,12 +205,36 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
...
+			}
+			if !atx.PubLayerID.After(commitmentAtx.PubLayerID) {
+				return nil, fmt.Errorf("atx layer (%v) must be after commitment atx layer (%v)", atx.PubLayerID, commitmentAtx.PubLayerID)
+			}
+			if d := atx.PubLayerID.Difference(commitmentAtx.PubLayerID); d > h.layersPerEpoch {
+				return nil, fmt.Errorf("expected distance of one epoch (%v layers) from commitment atx but found %v", h.layersPerEpoch, d)
+			}
+		} else {
+			publicationEpoch := atx.PublishEpoch()
+			if !publicationEpoch.NeedsGoldenPositioningATX() {
```
countvonzero: why not just use IsGenesis()? to me it's the same difference.

activation/handler.go
```diff
@@ -202,12 +205,36 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
...
+
 		// Use the NIPost's Post metadata, while overriding the challenge to a zero challenge,
 		// as expected from the initial Post.
 		initialPostMetadata := *atx.NIPost.PostMetadata
 		initialPostMetadata.Challenge = shared.ZeroChallenge
-		if err := h.nipostValidator.ValidatePost(atx.NodeID().ToBytes(), atx.InitialPost, &initialPostMetadata, uint(atx.NumUnits)); err != nil {
-			return nil, fmt.Errorf("invalid initial Post: %v", err)
+
+		commitment := sha256.Sum256(append(atx.NodeID().ToBytes(), atx.CommitmentATX[:]...))
+		if err := h.nipostValidator.ValidatePost(commitment[:], atx.InitialPost, &initialPostMetadata, uint(atx.NumUnits)); err != nil {
```
countvonzero: curious. why copy here?

activation/handler.go
```diff
@@ -220,27 +247,40 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
...
-		return nil, fmt.Errorf("failed to compute NIPost's expected challenge hash: %v", err)
+		return nil, fmt.Errorf("failed to compute NIPost's expected challenge hash: %w", err)
 	}
 
 	h.log.WithContext(ctx).With().Info("validating nipost", log.String("expected_challenge_hash", expectedChallengeHash.String()), atx.ID())
 
-	pubKey := signing.NewPublicKey(atx.NodeID().ToBytes())
-	leaves, err := h.nipostValidator.Validate(*pubKey, atx.NIPost, *expectedChallengeHash, uint(atx.NumUnits))
+	commitmentATX := atx.CommitmentATX
+	if atx.CommitmentATX == *types.EmptyATXID {
```
countvonzero: the syntactic check of CommitmentATX is scattered in 2, 3 different places. is it possible to re-organize SyntacticallyValidateAtx a little such that the expectation for CommitmentATX is in its own place/method?

activation/post.go
```diff
@@ -360,14 +366,16 @@ func (mgr *PostSetupManager) StopSession(deleteFiles bool) error {
...
+func (mgr *PostSetupManager) GenerateProof(challenge []byte, commitmentAtx types.ATXID) (*types.Post, *types.PostMetadata, error) {
 	state := mgr.getState()
 
 	if state != postSetupStateComplete {
 		return nil, nil, errNotComplete
 	}
 
-	prover, err := proving.NewProver(config.Config(mgr.cfg), mgr.LastOpts().DataDir, mgr.id)
+	// TODO(mafa): id field in post package should be renamed to commitment otherwise error messages are confusing
+	commitment := sha256.Sum256(append(mgr.id, commitmentAtx.Bytes()...))
```
countvonzero: have a method for calculation commitment bytes? as is, if we were to change hashing algo, we need to change it in many places...

activation/activation.go
```diff
@@ -483,58 +520,64 @@ func (b *Builder) Coinbase() types.Address {
...
+	if b.commitmentAtx != nil {
+		return b.commitmentAtx, nil
 	}
 
-	if len(bts) > 0 {
-		var tp types.NIPostChallenge
-		if err = codec.Decode(bts, &tp); err != nil {
-			return fmt.Errorf("parse NIPost challenge: %w", err)
-		}
+	syncedCh := b.syncer.RegisterChForSynced(ctx)
```
countvonzero: why do we need the node to be synced to get the commitment ATX?

sql/migrations/0001_initial.sql
```diff
@@ -132,13 +132,12 @@ CREATE TABLE poets
 
 CREATE INDEX poets_by_service_id_by_round_id ON poets (service_id, round_id);
 
-CREATE TABLE niposts
+CREATE TABLE state
```
countvonzero: i think the name state is too vague, especially when the value is BLOB. what's wrong with `niposts`?

sql/store/niposts.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+const (
+	NIPostChallengeKey    = "NIPost"
+	NIPostBuilderStateKey = "NIPostBuilderState"
+)
+
+// AddNIPostChallenge adds the data for nipost to the key-value store.
+func AddNIPostChallenge(db sql.Executor, ch *types.NIPostChallenge) error {
```
countvonzero: can you combine niposts.go and commitment.go in on file?
each sql pkg now represents one table and has one file. what's the rationale to deviate from current pattern?

sql/store/commitment.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+const CommitmentATXKey = "CommitmentATX"
+
+// AddCommitmentATX adds the id for the commitment atx to the key-value store.
+func AddCommitmentATX(db sql.Executor, atx types.ATXID) error {
```
dshulyak: why does it need to be stored like this? as you mentioned commitment atx is referenced in the nipost challenge in the first atx.

please remove this file

dshulyak (dismissed):

activation/post.go
```diff
@@ -11,9 +12,12 @@ import (
 	"github.com/spacemeshos/post/proving"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/datastore"
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
+//go:generate mockgen -package=activation -destination=./post_mock.go -source=./post.go
```
dshulyak: mocks are generated in the mocks directory. please do it consistently with the rest

activation/post.go
```diff
@@ -1,6 +1,7 @@
 package activation
 
 import (
+	"crypto/sha256"
```
dshulyak: please use spacemeshos/go-spacemesh/hash package

activation/post.go
fasmat: That's not true, there are mocks in the `activations` directory, at least one of which I moved with this PR to the `mocks` directory.

I can move mocks for this file to the mock directory too, but then I run into cyclic dependencies that are resolved by moving all types in this file into the `common/types` package. That is imho where they belong anyway, but will cause the PR to become even bigger ;-)

sql/store/commitment.go
fasmat: The reason it is stored like this is because a) its simpler and faster to look it up than having to look up the first ATX of the node itself and get it from there and b) the commitment ATX has to be determined before the first ATX is published.

It is similar to the `NiPoSTChallenge`, that is also handled by this package. It needs to be persisted before the ATX using it is published because the node could shut down in between and then needs to restore the same value to continue work where it left off.

sql/store/niposts.go
fasmat: I agree with every table one package, but not with every package only has one file.

By splitting up the code into multiple files it is imho easier to read the code and understand what it does, since everything related to each other is in the same file and files stay relatively small.

Imho having 20 files here with 3-5 functions each is much easier to parse as a developer than one file with > 1000 lines of code.

sql/migrations/0001_initial.sql
fasmat: Yes, I couldn't come up with a better name for the table.

The table is a key-value store. Before this PR it was used to store two values: `NIPost` and `NIPostBuilderState` both of which had an object serialized using the scale codec as value. Now a 3rd value is stored in this table `CommitmentATX` which stores the ID of the selected ATX for the PoST commitment.

All of these values have in common that they are determined at some point, (temporarily) persisted in this table and published later as a part of an ATX. The reason for the temporary persistence is because if the node restarts between determining these values and publishing the ATX that contains them it must continue to use the same values instead of determining new ones.

I just generalised the table to store any key-value pair and moved / renamed the existing functions accordingly.

activation/activation.go
fasmat: Because the commitment ATX selected by a node should be the one with the highest height it has seen before starting to initialise PoST.

If we don't sync here, the selected commitment ATX will always be the golden ATX.

activation/post.go
fasmat: Yes, I will add a method.

Imho this should be part of the PoST library, just like the parameter taking the commitment should be called commitment and not `id` (leads to confusing error messages in the worst case).

activation/handler.go
fasmat: I will take a look at `SyntacticallyValidateAtx` and try to reduce its cyclomatic complexity by splitting it up in multiple methods, probably one for each field that is validated.

activation/handler.go
fasmat: I used this method because it is also used for the positioning ATX; I wasn't aware there are two methods doing essentially the same.

Since `NeedsGoldenPositioningATX` seems to be exclusively used in this method, while `IsGenesis` is used in many different places, I will get rid of it and replace it.

activation/activation.go
poszu: I'm fine with using any existing solution. I didn't know about them. 

>The suggested change would probably work, but complicates testing a bit. 

How would it complicate testing?

IMHO it would actually simplify the code (by hiding caching code into a specialized container) and make it error-resistant (by making improper usage impossible). I consider using the `commitmentAtx` field directly improper as it would bring a non-obvious dependency where the getter must be called first (to fill the variable). IMO it's better to write code that is impossible to misuse rather than hope that nobody will use it wrongly in the future. Do you see any cons?

>I intentionally did not include any bigger refactorings to limit the size of this PR

It's not a refactor, that is a totally new piece of code.

activation/activation.go
countvonzero: to clarify. 
- for genesis miners (miners that start smeshing from epoch 0), they will have have golden ATX for commitment ATX. 
- for semshers that join at epoch 1, it's possible that some of them have golden ATX for commitment ATX, and some of them have valid ATX for commitment ATX.
- for smeshers that join at/after epoch 2, they should all have valid ATX for commitment ATX

do i understand correctly?

sql/migrations/0001_initial.sql
countvonzero: yes. i still think `nipost` is a better name than `state`. as these k/v pairs are all about nipost.

sql/store/niposts.go
countvonzero: sure. but like you said, this is a table that store key value pairs. using your argument, we should have 3 files, each serves a different key for this table? IMO it's over kill. but ok.

activation/activation.go
fasmat: After the discussion with Tal today: 

- The commitment ATX should be the ATX with the maximum hight the node as seen when it starts smeshing. This is not validated / enforced by other nodes.
- A node can choose any ATX for its commitment as it wants, but choosing the ATX with the highest available hight gives the node the highest identity tick height.
- This identity tick height is used in case of an attack against the tortoise beacon where honest parties prioritize IDs with the highest tick numbers.

sql/store/niposts.go
fasmat: You're right, I should split up the NiPoST and NiPoSTBuilder.

I tried to make `store`/`get`/`clear` generic but didn't find an approach that I was satisfied with. As an example:

```golang
type key string

const (
   CommitmentATX key = key("CommitmentATX")
   NiPost key = key("NiPost")
)

func StoreValueForKey(db sql.Executor, key key, value scale.Encodable) error

// or

func StoreValueForKey[T scale.Encodable](db sql.Executor, key key, value T) error
```

Both allow for any value to be written for a key to the DB as long as it implements the `scale.Encodable` interface. We could need to check during runtime via reflection if the provided value is of the type we expect to avoid someone by mistake writing something else to the store. 

At that point I thought that this is becoming more complicated than it needs to be and errors would still only be caught during runtime and not compile time.

sql/migrations/0001_initial.sql
fasmat: For the table I see your point. But calling the package that contains the code `nipost` is a bit confusing imho.

`nipost.AddNIPostChallenge`, `nipost.StoreChallenge` or similar variations to me are less clear about what they do and/or have more repetition in their naming than `store.AddNiPostChallenge`.

The package handles the storing of data not `NIPosts` in general, which are handled by multiple packages.

activation/post.go
fasmat: I changed it by using the `hash` package.

activation/handler.go
fasmat: I'm not sure I understand? To b able to validate the post challenge I need to use the hash of the concatenation of the node ID and the commitment ATX as input.

activation/post.go
fasmat: refactored package to have mock in its own package.

fasmat:
bors try

activation/handler.go
fasmat: I broke the method up into 3 individual methods, one for validating initial ATX, one for validating non-initial ATX and one that applies validations that are relevant for both.

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

activation/handler.go
countvonzero: yes. but why make a copy at call site? isn't this an implementation details for `ValidatePost()`?
`h.nipostValidator.ValidatePost(commitment[:],...)`

activation/activation.go
countvonzero: right. with Tal's info, can you confirm the effective behavior i listed if a node were to follow the protocol?

if what i stated is right. then all genesis miners will have empty commitment ATX and will be de-prioritized in beacon protocol.

sql/migrations/0001_initial.sql
countvonzero: my thinking is that it's all storing nipost-related data. 

i am sure a new person reading the code will ask, what is in `state` table? state about what? everything?
how about calling it `kvstore`? 
and i do support having generic interface store/get/clear. it doesn't have to be the public interface. for example, you can still keep `AddNIPostChallenge()` but call a `store()` that's shared by all 3 keys.

sql/store/niposts.go
countvonzero: as the code stands now, i don't understand the logical divide of these two files.

i think for me, it is simpler if this table is declared to be a kv table, and that caller simply store []byte inside. it's up to the caller to decide how to encode/decode the data. this table simply stores random kv pair.

activation/handler.go
```diff
@@ -220,72 +174,176 @@ func (h *Handler) SyntacticallyValidateAtx(ctx context.Context, atx *types.Activ
...
+
+	if !bytes.Equal(atx.InitialPost.Indices, atx.InitialPostIndices) {
+		return fmt.Errorf("initial Post indices included in challenge does not equal to the initial Post indices included in the atx")
+	}
+
+	if atx.CommitmentATX == nil {
+		return fmt.Errorf("no prevATX declared, but commitmentATX is missing")
+	}
+
+	if *atx.CommitmentATX == h.goldenATXID && !atx.PublishEpoch().IsGenesis() {
```
countvonzero, line 223: i am missing something for the commitment ATX logic.

for genesis miners, they can only have golden atx id for commitment atx.
and given that miners don't change commitment atx for the same node id, all their subsequent atxs are going to have golden atx as commitment atx.

syncer/syncer.go
```diff
@@ -89,9 +89,13 @@ type Syncer struct {
 	lastATXsSynced    atomic.Value
 
 	mu sync.Mutex
+
 	// awaitSyncedCh is the list of subscribers' channels to notify when this node enters synced state
 	awaitSyncedCh []chan struct{}
```
countvonzero: this can be removed. only activation pkg uses this, and you can replace them with the new api.

activation/activation.go
```diff
@@ -483,58 +508,63 @@ func (b *Builder) Coinbase() types.Address {
...
 	}
-
-	return nil
 }
 
-func (b *Builder) loadChallenge() error {
-	bts, err := niposts.Get(b.cdb, getNIPostKey())
-	if err != nil {
-		return fmt.Errorf("get NIPost challenge from store: %w", err)
+func (b *Builder) getCommitmentAtx(ctx context.Context) (*types.ATXID, error) {
```
countvonzero, line 510: how do you ensure that
- all atxs of the same node id uses the same commitment atx
- if identity file change (node id change), this commitment cache is cleared

common/types/post.go
```diff
@@ -0,0 +1,31 @@
+package types
+
+// PostConfig is the configuration of the Post protocol, used for data creation, proofs generation and validation.
+type PostConfig struct {
```
countvonzero: i don't understand this move. 

common/types, as it stand today, are for data. options specific to packages reside in separate package.
this move makes the code organization more confusing. what about other options scattered in its own package?

please undo this move. and maybe use a separate PR to bring up discussion for this purpose.

go.mod
```diff
@@ -14,14 +14,14 @@ require (
...
-	github.com/libp2p/go-libp2p v0.22.0
+	github.com/libp2p/go-libp2p v0.23.0
 	github.com/libp2p/go-libp2p-pubsub v0.8.1
 	github.com/libp2p/go-libp2p-testing v0.12.0
 	github.com/minio/sha256-simd v1.0.0
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/multiformats/go-multiaddr v0.7.0
 	github.com/natefinch/atomic v1.0.1
-	github.com/oasisprotocol/curve25519-voi v0.0.0-20220708102147-0a8a51822cae
+	github.com/oasisprotocol/curve25519-voi v0.0.0-20220914075539-c3db9d0435d1
```
countvonzero: separate PR for go.mod/go.sum changes? i don't see this PR needs this update.

activation/handler.go
fasmat: That's mostly correct. Only the very first ATX published by a node has `atx.CommitmentATX` set and the commitment ATX of a node doesn't change over its lifetime.

Subsequent ATXs don't have a Commitment ATX but the NiPosts in them will use the commitment ATX set in the first ATX of the node.

activation/activation.go
fasmat: Add 1): For every ATX a node receives it fetches the very first ATX published by it and verifies that the NiPost still uses the same commitment ATX.

Add 2): Does the identity of the node change while it is running? Isn't a re-creation of PoST necessary when the identity changes? This would also trigger a new initial ATX which would then set a new commitment ATX for the node.

This method `getCommitmentAtx` only gets the commitment ATX for the node itself, not for other nodes.

common/types/post.go
fasmat: This is a necessity to do what @dshulyak asked me to do: move the mocks to a `mocks` package.

If the PostConfig stays in the `activation` package I cannot create a mock using that type since then `mocks` would depend on the `activation` package and `activation` package would depend on `mocks` causing a cyclic dependency.

go.mod
fasmat: The PR can be found here: https://github.com/spacemeshos/go-spacemesh/pull/3595

fasmat:
bors try

fasmat:
bors try

common/types/post.go
fasmat: Mocks have been moved in https://github.com/spacemeshos/go-spacemesh/pull/3598

fasmat:
bors try

fasmat:
After pulling out all only tangental related changes of this PR into their own PRs it should be easier to review now. The following changes were pulled out and have already been merged:

- https://github.com/spacemeshos/go-spacemesh/pull/3616
- https://github.com/spacemeshos/go-spacemesh/pull/3615

fasmat:
bors try

activation/activation.go
```diff
@@ -498,14 +518,46 @@ func (b *Builder) loadChallenge() error {
 	return nil
 }
 
+func (b *Builder) getCommitmentAtx(ctx context.Context) (*types.ATXID, error) {
```
countvonzero, line 510: i think this method should be thread-safe. otherwise the commitment atx could be changed

activation/activation.go
```diff
@@ -207,22 +211,30 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts atypes.PostSetupOpt
...
-	b.exited = exited
+	ctx, stop := context.WithCancel(b.parentCtx)
+	b.stop = stop
+	b.exited = make(chan struct{})
+
+	commitmentAtx, err := b.getCommitmentAtx(ctx)
+	if err != nil {
+		close(b.exited)
+		b.started.Store(false)
+		b.mu.Unlock()
```
countvonzero: this is error prone. i'd prefer to do unlock prior to call to getCommitmentAtx() and make sure getCommitmentAtx it self is thread-safe.

activation/activation.go
```diff
@@ -239,6 +251,21 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts atypes.PostSetupOpt
 	return nil
 }
 
+// findCommitmentAtx determines the best commitment ATX to use for the node.
+// It will use the ATX with the highest height seen by the node and defaults to the goldenATX,
+// when no ATXs have yet been published.
+func (b *Builder) findCommitmentAtx() (types.ATXID, error) {
+	posAtx, err := atxs.GetAtxIDWithMaxHeight(b.cdb)
+	if errors.Is(err, sql.ErrNotFound) {
+		b.log.With().Info("using golden atx as commitment atx", posAtx)
```
countvonzero: no need to log posAtx in this case

activation/activation.go
```diff
@@ -239,6 +251,21 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts atypes.PostSetupOpt
 	return nil
 }
 
+// findCommitmentAtx determines the best commitment ATX to use for the node.
+// It will use the ATX with the highest height seen by the node and defaults to the goldenATX,
+// when no ATXs have yet been published.
+func (b *Builder) findCommitmentAtx() (types.ATXID, error) {
+	posAtx, err := atxs.GetAtxIDWithMaxHeight(b.cdb)
```
countvonzero: the name posAtx is misleading. maybe commitAtx?

activation/activation.go
```diff
@@ -498,14 +518,46 @@ func (b *Builder) loadChallenge() error {
...
+	}
+
+	id, err := kvstore.GetCommitmentATX(b.cdb)
+	switch {
+	case errors.Is(err, sql.ErrNotFound):
+		id, err := b.findCommitmentAtx()
+		if err != nil {
+			return nil, fmt.Errorf("failed to determine commitment ATX: %w", err)
+		}
+		if err := kvstore.AddCommitmentATX(b.cdb, id); err != nil {
```
countvonzero: there is still no way to ensure that this commitment atx is tied to the node id.

one way is to prepend node id here (similar to `GetCommitmentBytes`) to the atx id and save it to kvstore.
and when retrieved, verify that node id remain the same.
tho this scheme essentially make swapping ID file difficult, as the operator needs to copy the commitment atx key/value too to the new smesher machine.

i still think the most elegant solution is to store the commitment atx id with the ID file.

sql/atxs/atxs.go
```diff
@@ -73,6 +72,29 @@ func GetTimestamp(db sql.Executor, id types.ATXID) (timestamp time.Time, err err
...
+	}
+	dec := func(stmt *sql.Statement) bool {
+		stmt.ColumnBytes(0, id[:])
+		return true
+	}
+
+	if rows, err := db.Exec(`
+		select id from atxs 
+		where smesher = ?1
+		order by epoch asc, timestamp asc
```
countvonzero: ordering by timestamp seems redundant. we simply don't care. for 2 reasons:
- timestamp is when the node received this ATX, this can be via gossip or via sync (much later than the ATX's actual gossip time)
- there is just 1 atx per smesher per epoch. 

so `order by epoch asc limit 1` is sufficient

sql/atxs/atxs.go
```diff
@@ -73,6 +72,29 @@ func GetTimestamp(db sql.Executor, id types.ATXID) (timestamp time.Time, err err
...
+		limit 1;`, enc, dec); err != nil {
+		return types.ATXID{}, fmt.Errorf("exec id %v: %w", id, err)
```
countvonzero: this error string does not mke sense. id is always empty in this case. i think you want `nodeID` instead. actually this is the same for other methods as well.

activation/activation.go
countvonzero: or if you do prefer the db approach, once the prepended node id doesnt match, find the right one and update kvstorewith it

activation/activation.go
fasmat: I added a mutex that should make `getCommitmentAtx` thread safe.

activation/activation.go
fasmat: made `getCommitmentAtx()` threadsafe

activation/activation.go
fasmat: I updated the logic a little bit:

1. commitment ATXs are now stored on a per-nodeId basis in the `kvstore`. This way if the node id changes  a new commitment ATX has to be determined.
2. If the node has already published an ATX before, this initial ATX is used to get the commitment ATX (instead of looking it up in the `kvstore`). So the commitment ATX in `kvstore` is now really transient and only used for the time between a node starts for the first time and publishes its first ATX.

sql/atxs/atxs.go
fasmat: Fixed, also for the other methods

fasmat:
bors try

sql/kvstore/commitment.go
```diff
@@ -0,0 +1,32 @@
...
+// GetCommitmentATXForNode returns the id for the commitment atx from the key-value store.
+func GetCommitmentATXForNode(db sql.Executor, nodeId types.NodeID) (types.ATXID, error) {
+	var res types.ATXID
+	if err := getKeyValue(db, getKeyForNode(nodeId), &res); err != nil {
+		return *types.EmptyATXID, err
+	}
+	return res, nil
+}
+
+func ClearCommitmentAtx(db sql.Executor, nodeId types.NodeID) error {
```
countvonzero: why is this necessary? isn't it true that once stored for a node ID it should never change?
in another words, it should incur an error if the code tries to overwrite a kv pair

activation/activation.go
countvonzero: perfect. thanks

sql/kvstore/commitment.go
fasmat: Yes and no, after **publishing** the initial ATX the commitment ATX never changes for a node. Before that a node is free to choose any commitment ATX it deems satisfactory (and theoretically also change its choice and recreate PoST). After the initial ATX has been published we don't need the kvstore any more to get the commitment ATX, we can just get it from the first ATX published by the node.

The kvstore is not a source of truth for the commitment ATX, it is only a transient storage for a node that is currently initializing PoST and hasn't published its initial ATX yet.

countvonzero: agreed. but current code does not try to choose a better committment ATX before it publish the initial ATX.
having this interface is confusing until code is added to choose a better one.

sql/kvstore/commitment.go
```diff
@@ -15,18 +16,42 @@ func getKeyForNode(nodeId types.NodeID) string {
...
+		stmt.ColumnBytes(0, val[:])
+		return true
+	}); err != nil {
+		return *types.EmptyATXID, fmt.Errorf("failed to get value: %w", err)
+	} else if rows == 0 {
+		return *types.EmptyATXID, fmt.Errorf("failed to get value: %w", sql.ErrNotFound)
+	}
+
 	var res types.ATXID
-	if err := getKeyValue(db, getKeyForNode(nodeId), &res); err != nil {
```
countvonzero: not sure why you don't use `getKeyValue` here.

countvonzero (approved):

sql/kvstore/commitment.go
fasmat: OK, then I'll remove the `ClearCommitmentAtx` function and update the setter and getters for commitmentATX

sql/kvstore/commitment.go
fasmat: Reverted GetCommitmentATXForNode

fasmat:
bors merge

PR #3568, opened 2022-09-19
cmd/node: configuration changes for genesis id

### LINKED ISSUE #3505: cmd/node: configuration changes for genesis id

- add a genesis section into config
- for this issue genesis section needs to include: genesis time and genesis extra data (other fields can be moved into this config later)
   - genesis time is a string with time encoded according to RFC3339
   - genesis extra data is a string with a max size of 255 chars. defaults to `mainnet`
- when node is started check if genesis section is already stored in the data directory (example: /data/genesis.conf)
   - if it is not stored, encode genesis section with json and persist it on disk
   - if stored - decode object with json into another instance of genesis config, compute diff using https://github.com/google/go-cmp/ and print fatal error to the operator with a diff and a path to persisted genesis conf
- compute genesis id as a hash from genesis time and genesis extra data
- golden atx should be set to genesis id

evt:
## Motivation
Closes #3505 

## Changes
Adds a genesis section into config.

## DevOps Notes


config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+		ExtraData:   DefaultTestGenesisExtraData(),
+		GoldenATXID: DefaultTestGoldenATXId().Hash32().Hex(),
+		GenesisID:   CalcGenesisId([]byte(DefaultTestGenesisExtraData()), DefaultTestGenesisTime()).Hex(),
+		Path:        DefaultGenesisDataDir(),
+	}
+}
+
+func (gc *GenesisConfig) Compare(payload *GenesisConfig) error {
+	if diff := cmp.Diff(payload, gc); diff != "" {
+		fmt.Printf("config files mismatch (-want +got):\n%s", diff)
```
countvonzero: rm

cmd/node/node.go
```diff
@@ -1164,3 +1170,35 @@ func decodeLoggers(cfg config.LoggerConfig) (map[string]string, error) {
...
+			return fmt.Errorf("failed to read stored genesis config file: %s", err)
+		}
+
+		var payload config.GenesisConfig
+
+		if err = json.Unmarshal(genesisData, &payload); err != nil {
+			return fmt.Errorf("failed to unmarshal stored genesis config file: %w", err)
+		}
+		if err = app.Config.Genesis.Compare(&payload); err != nil {
+			return fmt.Errorf("failed to compare stored genesis config file: %w", err)
```
countvonzero: the txt is misleading. how about "genesis config changed from previous run"?

cmd/node/node.go
```diff
@@ -1164,3 +1170,35 @@ func decodeLoggers(cfg config.LoggerConfig) (map[string]string, error) {
...
+		}
+		return nil
+	}
+	jsonCfg, err := json.Marshal(app.Config.Genesis)
```
countvonzero, line 1192: only persist when the genesis config is absent?

countvonzero (commented):
i don't know the spec well enough to approve. just some general comment

cmd/node/node.go
```diff
@@ -1164,3 +1170,35 @@ func decodeLoggers(cfg config.LoggerConfig) (map[string]string, error) {
 	}
 	return rst, nil
 }
+
+func (app *App) checkAndStoreGenesisConfig() error {
+	genesisConfigPath := filepath.Join(app.Config.Genesis.Path, config.DefaultGenesisConfigFileName)
```
dshulyak: this should use configured data dir, not some new parameter

cmd/node/node.go
```diff
@@ -1164,3 +1170,35 @@ func decodeLoggers(cfg config.LoggerConfig) (map[string]string, error) {
...
+	if exists := filesystem.PathExists(genesisConfigPath); exists {
+		genesisData, err := ioutil.ReadFile(genesisConfigPath)
+		if err != nil {
+			return fmt.Errorf("failed to read stored genesis config file: %s", err)
+		}
+
+		var payload config.GenesisConfig
```
dshulyak: payload is not a good name

cmd/node/node.go
dshulyak: yes message is bad. it also needs to specify location of persisted config file

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+	DefaultGenesisConfigFileName = "genesis.conf"
+)
+
+type GenesisConfig struct {
+	Accounts    *apiConfig.GenesisAccountConfig `mapstructure:"genesis-api"`
+	GenesisTime string                          `mapstructure:"genesis-time"`
+	ExtraData   string                          `mapstructure:"genesis-extradata"`
+	GoldenATXID string
+	GenesisID   string
+	Path        string
```
dshulyak: this config should have only 3 things: Accounts (mapping), GenesisTime, ExtraData

remove everything else

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (gc *GenesisConfig) Compare(payload *GenesisConfig) error {
+	if diff := cmp.Diff(payload, gc); diff != "" {
+		fmt.Printf("config files mismatch (-want +got):\n%s", diff)
+		return errors.New("failed to match config files")
+	}
+	return nil
+}
+
+func DefaultGenesisDataDir() string {
```
dshulyak: remove

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+	return filepath.Join(defaultDataDir, DefaultGenesisConfigFileDir)
+}
+
+func DefaultGenesisTime() string {
+	return time.Now().Format(time.RFC3339)
+}
+
+func DefaultGenesisExtraData() string {
```
dshulyak: put this line into the default config or to constant

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+	return "mainnet"
+}
+
+func DefaultGoldenATXId() types.ATXID {
```
dshulyak: remove, it doesn't make sense

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+
+func DefaultTestnetGenesisExtraData() string {
+	return "testnet"
+}
+
+func DefaultTestnetGoldenATXId() types.ATXID {
+	return types.ATXID(CalcGoldenATX([]byte(DefaultTestnetGenesisExtraData()), DefaultTestnetGenesisTime()))
+}
+
+func CalcGenesisId(genesisExtraData []byte, genesisTime string) types.Hash20 {
```
dshulyak: this function is not doing a right thing, it should be

hash.Write(gentime)
hash.Write(extradata)

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	digest := hasher.Sum([]byte{})
+	return types.BytesToHash(digest).ToHash20()
+}
+
+func (cfg *GenesisConfig) CalcGenesisID() {
+	cfg.GenesisID = CalcGenesisId([]byte(cfg.ExtraData), cfg.GenesisTime).Hex()
+}
+
+func CalcGoldenATX(genesisData []byte, genesisTime string) types.Hash32 {
```
dshulyak: remove this

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+		copy(padded[0:dataLen], genesisExtraData)
+		hasher.Write(padded)
+	} else {
+		hasher.Write(genesisExtraData)
+	}
+	digest := hasher.Sum([]byte{})
+	return types.BytesToHash(digest).ToHash20()
+}
+
+func (cfg *GenesisConfig) CalcGenesisID() {
```
dshulyak: return types.Hash20 from this method

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+
+func DefaultTestnetGenesisTime() string {
+	return DefaultTestGenesisTime()
+}
+
+func DefaultTestnetGenesisExtraData() string {
+	return "testnet"
+}
+
+func DefaultTestnetGoldenATXId() types.ATXID {
```
dshulyak: remove

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+const (
+	ExtraDataLen                 = 255
+	DefaultGenesisConfigFileDir  = "data"
+	DefaultGenesisConfigFileName = "genesis.conf"
+)
+
+type GenesisConfig struct {
+	Accounts    *apiConfig.GenesisAccountConfig `mapstructure:"genesis-api"`
```
dshulyak: store accounts directly in this config

config/genesis.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (cfg *GenesisConfig) CalcGenesisID() {
+	cfg.GenesisID = CalcGenesisId([]byte(cfg.ExtraData), cfg.GenesisTime).Hex()
+}
+
+func CalcGoldenATX(genesisData []byte, genesisTime string) types.Hash32 {
+	return CalcGenesisId([]byte(genesisData), genesisTime).ToHash32()
+}
+
+func (cfg *GenesisConfig) CalcGoldenATX() {
```
dshulyak: remove this

dshulyak (changes requested):
changes in the vm module needs to be removed
also please drop all that DefaultSomethingSomething. please keep only DefaultConfig. DefaultTestnetConfig and DefaultFastnetConfig should be written in presets files

config/genesis.go
evt: I added this on purpose to see the diff in a clean way when error occurs. If we include it into error printed then JSON log escapes it and it doesn't look good to understand. Or no point to print diff at all?

cmd/node/node.go
evt: yes, correct. If it exists - we make sure it hasn't changed and exit otherwise.

config/genesis.go
evt: this is what it does, not sure what you mean

dshulyak: remove padding

cmd/node/node.go
evt: fixed

config/genesis.go
evt: done

config/genesis.go
evt: done

cmd/node/node.go
evt: fixed

cmd/node/node.go
evt: fixed

config/genesis.go
evt: removed

config/genesis.go
evt: removed as GenesisID removed

config/genesis.go
evt: done

config/genesis.go
evt: done

config/genesis.go
evt: done

config/genesis.go
evt: removed

config/genesis.go
evt: done

config/genesis.go
evt: done

evt:
> changes in the vm module needs to be removed also please drop all that DefaultSomethingSomething. please keep only DefaultConfig. DefaultTestnetConfig and DefaultFastnetConfig should be written in presets files

vm changes removed. DefaultTestnetConfig - done. DefaultFastnetConfig - do we have it?

evt:
bors try

config/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+			Address: genesisAddr,
+			Balance: balance,
+		})
+	}
+	return rst
+}
+
+func (gc *GenesisConfig) Compare(stored *GenesisConfig, path string) error {
+	if diff := cmp.Diff(stored, gc); diff != "" {
+		fmt.Printf("Genesis config %s changed from previous run (-want +got):\n%s", path, diff)
```
dshulyak: but this can't stay like this, can you find a cleaner solution to print details?

config/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/spacemeshos/go-spacemesh/hash"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+const (
+	DefaultGenesisConfigFileDir  = "data"
+	DefaultGenesisConfigFileName = "genesis.conf"
+	defaultGenesisExtraData      = "mainnet"
+
+	DefaultTestGenesisTime      = "2022-12-25T00:00:00+00:00"
```
dshulyak: where do we use this? should be possible to cleanup

config/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/google/go-cmp/cmp"
+
+	apiConfig "github.com/spacemeshos/go-spacemesh/api/config"
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/hash"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+const (
+	DefaultGenesisConfigFileDir  = "data"
```
dshulyak: so given that spacemesh is configured with datadir - /state/ . genesis config will saved into /state/data/genesis.conf. does it make sense? why is data needed?

config/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+
+func defaultGenesisConfig() *GenesisConfig {
+	return &GenesisConfig{
+		Accounts:    apiConfig.DefaultGenesisAccounts(),
+		GenesisTime: time.Now().Format(time.RFC3339),
+		ExtraData:   defaultGenesisExtraData,
+	}
+}
+
+func GenesisDataDir(dataDirParent string) string {
```
dshulyak: can you explain why do we need this?

api/config/genesis.go
```diff
@@ -3,51 +3,23 @@ package config
 import (
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/common/util"
-	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
-// GenesisConfig defines accounts that will exist in state at genesis.
```
dshulyak, line 10: why other things weren't moved?

config/genesis.go
evt: Not needed really. Removed.

config/genesis.go
evt: Removed along with https://github.com/spacemeshos/go-spacemesh/pull/3568#discussion_r977193346

config/genesis.go
evt: replaced with log error

api/config/genesis.go
evt: moved it all to `config/genesis.go`

config/genesis.go
evt: moved it to `cmd/node/node_test.go`

config/genesis.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		rst = append(rst, types.Account{
+			Address: genesisAddr,
+			Balance: balance,
+		})
+	}
+	return rst
+}
+
+func (gc *GenesisConfig) Compare(stored *GenesisConfig, path string, lg log.Log) error {
```
dshulyak, line 73: this method signature is very bad

`Diff(other genesisConfig) string`

if result is not empty print a fatal error on the caller side

dshulyak:
implemented elsewhere

Issue #3551, opened 2022-09-13
p2p: provide api for app to retain / drop peers


countvonzero:
## Description

use cases
- sync: when a node is syncing, we want to be able to manage a balance btwn
  - peers that has higher tortoise-verified layer and higher ATX weight: this will help adopting their local opinions without unnecessary tortoise reruns 
  - peers that are less synced than the node: we want to help other node to sync data, but should keep them at some threshold (say 20% of the peers)

- malicious peer: when peer relay invalid data to us, we should be able to drop them as peers and/or blacklist the p2p identity

we should be able to achieve these functionality via the gossipsub scoring scheme by supplying a custom application scoring function.

dshulyak:
@countvonzero is it only to get more peers in case of eclipse attack? i added medium priority, please change if you think otherwise

PR #3570, opened 2022-09-20
[Merged by Bors] - sync: get block validity from peers

### LINKED ISSUE #3003: tortoise, sync: required data for tortoise when syncing a new node

countvonzero:
## Motivation

Closes #3003

## Changes

- syncer: move state sync code to a separate file
- moves opinion processing from fetcher to syncer pkg
- add the following data to opinions response
  - epoch weight
  - latest verified layer
  - aggregated hash
  - block validity 

opinions amongst peers are sorted in the following order
1. highest epoch weight
2. highest verified layer
3. existence of certificate

- only adopts block validity from peers whose aggregated layer hash is the same as node from the previous layer.
- block validity is saved directly to database for tortoise to consume.

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -547,13 +547,12 @@ func (app *App) initServices(ctx context.Context,
 	dataHanders := fetch.DataHandlers{
 		ATX:      atxHandler,
 		Block:    blockHandller,
-		Cert:     app.certifier,
 		Ballot:   proposalListener,
 		Proposal: proposalListener,
 		TX:       txHandler,
 		Poet:     poetDb,
 	}
-	layerFetch := fetch.NewLogic(app.Config.FETCH, sqlDB, msh, app.host, dataHanders, app.addLogger(LayerFetcher, lg))
+	layerFetch := fetch.NewLogic(app.Config.FETCH, cdb, msh, app.host, dataHanders, app.addLogger(LayerFetcher, lg))
 	fetcherWrapped.Fetcher = layerFetch
 
 	patrol := layerpatrol.New()
@@ -562,7 +561,7 @@ func (app *App) initServices(ctx context.Context,
 		HareDelayLayers:  app.Config.Tortoise.Zdist,
 		SyncCertDistance: app.Config.Tortoise.Hdist,
 	}
-	newSyncer := syncer.NewSyncer(ctx, syncerConf, clock, beaconProtocol, msh, layerFetch, patrol, app.addLogger(SyncLogger, lg))
+	newSyncer := syncer.NewSyncer(ctx, syncerConf, sqlDB, clock, beaconProtocol, msh, layerFetch, patrol, app.certifier, app.addLogger(SyncLogger, lg))
 	// TODO(dshulyak) this needs to be improved, but dependency graph is a bit complicated
 	beaconProtocol.SetSyncState(newSyncer)
 
--- a/fetch/handler.go
+++ b/fetch/handler.go
@@ -22,22 +22,24 @@ var errInternal = errors.New("unspecified error returned by peer")
 
 type handler struct {
 	logger log.Log
-	db     *sql.Database
+	cdb    *datastore.CachedDB
 	bs     *datastore.BlobStore
+	msh    meshProvider
 }
 
-func newHandler(db *sql.Database, bs *datastore.BlobStore, lg log.Log) *handler {
+func newHandler(cdb *datastore.CachedDB, bs *datastore.BlobStore, m meshProvider, lg log.Log) *handler {
 	return &handler{
 		logger: lg,
-		db:     db,
+		cdb:    cdb,
 		bs:     bs,
+		msh:    m,
 	}
 }
 
 // handleEpochATXIDsReq returns the ATXs published in the specified epoch.
 func (h *handler) handleEpochATXIDsReq(ctx context.Context, msg []byte) ([]byte, error) {
 	epoch := types.EpochID(util.BytesToUint32(msg))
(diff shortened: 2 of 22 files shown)


countvonzero:
bors try

dshulyak:
validity is imported from the same peer or each layer from different? it won't work in the second case

dshulyak:
are certificates downloaded for all layers? not only hdist?

countvonzero:
> validity is imported from the same peer or each layer from different? it won't work in the second case

as implemented it can be from different peer for each layer. it will be ideal if we can import it from the same peer. but
A. we don't have a way to retain peer now. https://github.com/spacemeshos/go-spacemesh/issues/3551
B. even if we can retain peer, there is still cases where the peer can drop from the network anytime.
for B, my plan is to switch to a different peer if the mesh hash is consistent. 

in general, if the node is syncing from peer X and Y, both have the same accumulated hash in the mesh. what is the issue you see that to import validity from X for layer N and from Y for layer N+1?

> are certificates downloaded for all layers? not only hdist?

yes. for all layers for now. https://github.com/spacemeshos/go-spacemesh/blob/e55b88add994aea08ecdc9271212f9e8b7fa35ba/syncer/state_syncer.go#L223
there is an attempt to do this, but mesh.ProcessLayer()/trtl's last layer usually is the last synced layer. this is the reason why we cannot prune the certificate yet in the network, as discussed in slack https://spacemesh.slack.com/archives/C89QJJY3G/p1661405377244839?thread_ts=1661281388.797769&cid=C89QJJY3G

we need to fix https://github.com/spacemeshos/go-spacemesh/issues/2921 before we can start pruning certificate and that a new node only need certificate at the last hdist layers.

dshulyak:
> we don't have a way to retain peer now. https://github.com/spacemeshos/go-spacemesh/issues/3551

i don't quite understand why there is a dependency on that. it is completely up to sync which peers to use. ofcourse they may go away unexpectedly, as you noted

> in general, if the node is syncing from peer X and Y, both have the same accumulated hash in the mesh. 

if it downloads opinion from the peer that atleast communicates same hash it should work. but otherwise, even if one opinion somewhere in the middle is inconsistent, verifying tortoise will be stuck

countvonzero:
> > we don't have a way to retain peer now. #3551
> 
> i don't quite understand why there is a dependency on that. it is completely up to sync which peers to use. 

so i don't fully understand how p2p module manage peers and probably assume incorrectly.

let's say the node finds peer X has a different hash at layer N, it wants to find at which layer it starts having differences.
i assume at this point we want to make best effort to retain peer X until the node gets all relevant data from X.

last time we discussed this in the dev sync, your suggestion was to use gossipsub scoring system to keep this peer around.
but since fetcher gets peers from p2p/bootstrap, which uses EventSpacemeshPeer to add/rm from the peer list, i am not sure gossipsub's scoring system can affect that.

> > in general, if the node is syncing from peer X and Y, both have the same accumulated hash in the mesh.
> 
> if it downloads opinion from the peer that atleast communicates same hash it should work. but otherwise, even if one opinion somewhere in the middle is inconsistent, verifying tortoise will be stuck

thanks. i'll make that change in this PR and notify you when its ready

countvonzero:
> i don't quite understand why there is a dependency on that. it is completely up to sync which peers to use.

@dshulyak 
i didn't explain this well. yes syncer can pick any peer to use. but if we don't have an active connection with the peer, it'll fail here
https://github.com/spacemeshos/go-spacemesh/blob/09225a3ada8939ccbe9c3c6cf9f18af91d9f2367/p2p/server/server.go#L141

and i don't want to just dial a peer randomly in this case. i'd rather like an API from the p2p package to do that consistently with its peer management scheme.

dshulyak:
> and i don't want to just dial a peer randomly in this case. i'd rather like an API from the p2p package to do that consistently with its peer management scheme.

rpc server won't dial peer if there is no connection. it has this https://github.com/spacemeshos/go-spacemesh/blob/09225a3ada8939ccbe9c3c6cf9f18af91d9f2367/p2p/server/server.go#L154 . yes it will fail if selected peer fails to respond. but it should be expected that peer can go away. node has no control over it

the only thing that be provided by p2p module is to prioritize connections with "better" peers. but it is only relevant if node will be over high watermark for number of peers (currently 100), because at that time it will prune total number of peers down to lower watermark (which is set to 40)

dshulyak (approved):

countvonzero:
> if it downloads opinion from the peer that atleast communicates same hash it should work. but otherwise, even if one opinion somewhere in the middle is inconsistent, verifying tortoise will be stuck

@dshulyak added a commit that only adopts block validity from peers with the same aggregated layer hash from the previous layer

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

PR #3571, opened 2022-09-20
[Merged by Bors] - vm: remove bitfield from nonce

dshulyak:
cleanup, genesis will be released with templates that don't support bitfield


common/types/transaction_header.go
```diff
@@ -48,6 +47,5 @@ type LayerLimits struct {
 // Nonce is for ordering transactions.
 // TODO(dshulyak) we are using only counter until bitfield is defined.
```
countvonzero: update comment

common/types/transaction_header.go
```diff
@@ -48,6 +47,5 @@ type LayerLimits struct {
...
 type Nonce struct {
-	Counter  uint64
-	Bitfield uint8
+	Counter uint64
```
countvonzero: if we are doing without bitfield, can we make it a simple uint64 nonce instead of struct? 
readers that don't know the history will be confused.

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3572, opened 2022-09-20
Cleanup p2p configuration structures

poszu:
## Motivation
The existing configuration structures are a bit messy. Some fields are doubled - they exist both in the "main" config struct for p2p as well as in the submodule's config structures (for example, fields for the book check logic).

## Changes
- cleaned up this duplication by using the submodule's config structures as subfields of the "main" p2p config struct.
-  removed fields that are not part of the _configuration_ (they cannot be _configured_ by user) out from `Config` structs. Now `Config` structs contain only settings configurable either by a config file (config.toml) or flags.
- removed `config.toml` fields that don't exist and moved the ones that changed their place in config structures.

## Test Plan
unittests

## DevOps Notes


p2p/pubsub/pubsub.go
```diff
@@ -77,21 +77,23 @@ const (
...
-	Flood          bool
-	IsBootnode     bool
-	MaxMessageSize int
+	Flood          bool `mapstructure:"flood"`
+	MaxMessageSize int  `mapstructure:"max-message-size"`
 }
 
 // New creates PubSub instance.
-func New(ctx context.Context, logger log.Log, h host.Host, cfg Config) (*PubSub, error) {
+func New(ctx context.Context, logger log.Log, h host.Host, cfg Config, isBootnode bool) (*PubSub, error) {
```
fasmat: for me booleans arguments to a function are usually an indication that that function should be two instead, in this case probably `New` and `NewBootNode`

p2p/pubsub/pubsub.go
```diff
@@ -214,7 +216,7 @@ func getOptions(cfg Config) []pubsub.Option {
 	}
 
 	// enable Peer eXchange on bootstrappers
-	if cfg.IsBootnode {
+	if isBootnode {
```
fasmat, line 218: same as above, imho this could be part of it's own function `getGossipOptions` or similar that is only called by `NewBootnode` but not be `New`

p2p/upgrade.go
```diff
@@ -78,7 +78,7 @@ func isBootnode(h host.Host, bootnodes []string) (bool, error) {
 }
 
 // Upgrade creates Host instance from host.Host.
-func Upgrade(h host.Host, opts ...Opt) (*Host, error) {
+func Upgrade(h host.Host, dataDir string, opts ...Opt) (*Host, error) {
```
fasmat, line 81: `dataDir` could be an `Opt` (that possibly defaults to some value if not set)

p2p/upgrade.go
```diff
@@ -89,37 +89,22 @@ func Upgrade(h host.Host, opts ...Opt) (*Host, error) {
...
-	if fh.discovery, err = peerexchange.New(fh.logger, h, peerexchange.Config{
-		DataDir:              cfg.DataDir,
-		Bootnodes:            cfg.Bootnodes,
-		CheckPeersNumber:     cfg.CheckPeersNumber,
-		CheckTimeout:         cfg.CheckTimeout,
-		CheckInterval:        cfg.CheckInterval,
-		CheckPeersUsedBefore: cfg.CheckPeersUsedBefore,
-		PeerExchange:         cfg.peerExchange,
-	}); err != nil {
+	if fh.discovery, err = peerexchange.New(fh.logger, h, cfg.Discovery, dataDir); err != nil {
```
fasmat: i think `logger` and `dataDir` could be part of `cfg.Discovery`

p2p/upgrade.go
```diff
@@ -89,37 +89,22 @@ func Upgrade(h host.Host, opts ...Opt) (*Host, error) {
...
+	bootnode, err := isBootnode(h, cfg.Discovery.Bootnodes)
 	if err != nil {
 		return nil, fmt.Errorf("check node as bootnode: %w", err)
 	}
-	if fh.PubSub, err = pubsub.New(fh.ctx, fh.logger, h, pubsub.Config{
-		Flood:          cfg.Flood,
-		IsBootnode:     bootnode,
-		MaxMessageSize: cfg.MaxMessageSize,
-	}); err != nil {
+	if fh.PubSub, err = pubsub.New(fh.ctx, fh.logger, h, cfg.Pubsub, bootnode); err != nil {
```
fasmat: `fh.logger` could be part of the config and `bootnode` could either be part of the config, or there could be a new function `pubsub.NewBootnode`

EDIT: `isBootnode` in my opinion probably shouldn't be its own function but rather be part of `pubsub.New`

p2p/upgrade.go
```diff
@@ -89,37 +89,22 @@ func Upgrade(h host.Host, opts ...Opt) (*Host, error) {
...
-		PeerExchange:         cfg.peerExchange,
-	}); err != nil {
+	if fh.discovery, err = peerexchange.New(fh.logger, h, cfg.Discovery, dataDir); err != nil {
 		return nil, fmt.Errorf("failed to initialize peerexchange discovery: %w", err)
 	}
-	if fh.bootstrap, err = bootstrap.NewBootstrap(fh.logger, bootstrap.Config{
-		TargetOutbound: cfg.TargetOutbound,
-		Timeout:        cfg.BootstrapTimeout,
-	}, fh, fh.discovery); err != nil {
+	if fh.bootstrap, err = bootstrap.NewBootstrap(fh.logger, cfg.Bootstrap, fh, fh.discovery); err != nil {
```
fasmat: I think `fh.logger` and `fh.discovery` don't need to be passed in explicitly since they are already passed in via `fh`.

p2p/peerexchange/discovery.go
```diff
@@ -50,15 +64,15 @@ type Discovery struct {
 }
 
 // New creates a Discovery instance.
-func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
+func New(logger log.Log, h host.Host, config Config, dataDir string) (*Discovery, error) {
```
fasmat: same here, to me the function signature could be:

`func New(h host.Host, config Config)` where the config contains both the `logger` and the `dataDir`, or optionally an `opt... Opts` if that's more convenient for the caller of that function.

On that note: the config that is passed in here and the config above don't necessarily need to be (or even should be) the same type. The one passed in here needs to contain ALL configurable parameters, while the one above ONLY contains the ones that can be set via `config.yml` and/or CLI

p2p/host.go
```diff
@@ -15,59 +15,44 @@ import (
...
-
-	peerExchange peerexchange.PeerExchangeConfig `mapstructure:"peer-exchange"`
+	Pubsub    pubsub.Config       `mapstructure:"pubsub"`
+	Discovery peerexchange.Config `mapstructure:"discovery"`
+	Bootstrap bootstrap.Config    `mapstructure:"bootstrap"`
 }
 
 // New initializes libp2p host configured for spacemesh.
-func New(_ context.Context, logger log.Log, cfg Config, opts ...Opt) (*Host, error) {
+func New(_ context.Context, logger log.Log, cfg Config, dataDir string, opts ...Opt) (*Host, error) {
```
fasmat, line 53: this `New` contains all 3 options on how to pass in configuration:

it has arguments for `logger` and `dataDir`, a `Config` object that contains all other configuration AND `opts ...Opt`. This function should in my opinion be either:

- `New(ctx context.Context, cfg Config)` or
- `New(ctx context.Context, opts ...Opt)`

with dataDir and logger being either an `Opt` or a field in the `Config`

config.toml
```diff
@@ -6,42 +6,28 @@
...
-log-encoder = "console"
 metrics = false
 metrics-port = 2020
 oracle_server = "http://localhost:3030"
 oracle_server_worldid = 0
 genesis-time = "2019-02-13T17:02:00+00:00"
 layer-duration-sec = "5"
-block-cache-size = "20"
-hdist = "5"
-coinbase = "0x1234"
```
fasmat, line 18: a lot of entries here were deleted, are they really not configurable any more?

fasmat (changes requested):
There is a wild mixture of different patterns on how to pass configuration / parameters to various components that in my opinion should be aligned.

I know this is mostly not caused by the changes of this PR specifically, but this PR might be a good place to improve on that.

config.toml
dshulyak: i don't know why we keep this config in repo

dshulyak (commented):
i haven't reviewed pr fully, i don't agree that it is needed or better. original intention was to hide implementation details (e.g the existence of bootstrap component) from operator

wrt parameters that can't be configured by operator, i don't see as well what the problem with them? in some modules we have parameters that are configured for another component, but they need to be in sync with that secondary module

dshulyak (changes requested):
i don't think that operator should be aware that we have p2p.bootstrap component or p2p.pubsub, it can get removed or updated and it will require changes in config structure

poszu:
@dshulyak 
> i don't think that operator should be aware that we have p2p.bootstrap component or p2p.pubsub, it can get removed or updated and it will require changes in the config structure

I absolutely understand your point and I actually agree :) I think there is a middle point where we can meet with this. My intent was to remove duplication and have a single source of truth for the settings that are configurable by an operator. In order to hide the implementation details from the config, I can embed these structures in the main p2p `Config` instead of having separate fields.

config.toml
poszu: While I am at it, I can remove it completely 🤷 .

p2p/upgrade.go
poszu: They must be passed. Both `logger` and `discovery` are private fields of `Host` and thus are not accessible from `bootstrap` pkg.

p2p/upgrade.go
poszu: Well, I don't think it could. The `type Opt func(fh *Host)` gets a pointer to the `Host` instance and can do something with it (i.e. set a field). However, `dataDir` is not, and needs not to be, a part of `Host` (there is no field `dataDir` in `Host` because there is no need for it).

p2p/upgrade.go
fasmat: This is true; I think they don't need to be private. They have the same scope as `fh.Peers` and `fh.PubSub` both of which are also set in this function and only used in this file.

To me the current function signature indicates that `bootstrap` could use a different logger or discovery than the host, which I don't think is intended?

dshulyak:
> In order to hide the implementation details from the config, I can embed these structures in the main p2p Config instead of having separate fields.

i don't know if parsing will work without subgroups (there was some special code in cmd/base.go), but in general sounds good

p2p/upgrade.go
poszu: In my opinion, `logger` should not be part of `cfg.Discovery` as it is not a _configuration_. It is an instance of a logger shared from its parent. 

Regarding `dataDir`, I just removed it from `Discovery` struct in order to clean up data structures that are filled by an operator via `viper` and `cobra`.

p2p/upgrade.go
fasmat: before the change `dataDir` was a field of the host and its purpose is to be passed along all the way down to the addressbook. It is also used in a call to `EnsureIdentity`.

I think it could be named `identitiesDir` instead and then be part of the hosts config, just as it was before just with a clearer name of its intended purpose.

p2p/upgrade.go
fasmat: > In my opinion, logger should not be part of cfg.Discovery as it is not a configuration. It is an instance of a logger shared from its parent.

That's a good point, then I think `fh.logger` could be accessed by `peerexchange.New` by making it public or adding a getter to `Host`. I don't think there is any reason that `peerexchange` would use a different logger than the one from host, but having it as its own parameter makes it appear as if that's the case.

p2p/upgrade.go
poszu: I also thought about moving the `isBootnode` function to the `pubsub` pkg. I think that the reason why it was placed here is that it uses the `bootnodes []string` field of the config, which is not available from `pubsub.New()`. However, we don't want to explicitly pass the list of boot nodes to `pubsub.New`. The whole list is not required there - it just needs information if we are a boot node or not.

Regarding creating separate `New(...)` and `NewBootnode(...)` I'm not sure if it is worth it. There would be 3 functions instead of just one (private `new(...)` accepting the boolean and public New(...) and `NewBootstrap(...)`). IMO it's an overkill for this usecase.

I can propose something different. Another way to get rid of `bools` (which I hate too 😉 ) is to use enums. Consider:
```go
package p2p

type NodeType uint8
const (
  Bootnode NodeType = iota
  RegularNode 
)

---

package pubsub
func New(..., nodeType NodeType) { ... }
```

poszu:
> > In order to hide the implementation details from the config, I can embed these structures in the main p2p Config instead of having separate fields.
> 
> i don't know if parsing will work without subgroups (there was some special code in cmd/base.go), but in general sounds good

It should work with a `squash` mapstructure tag. But I will need to rename the structs (they are named just `Config` now) to avoid name collision when embedded in a single struct.

p2p/upgrade.go
fasmat: Using multiple "constructors" like `New` and `NewBootnode` feels more idiomatic to me in go, but I would be OK with an enum for the NodeType as well.

p2p/upgrade.go
poszu: Yes, it was a field, but it doesn't need to be as it is used one time during the initialization.

fasmat: I think as a caller of the `Upgrade` function it is confusing to a) not know what `dataDir` is used for (which data is stored in that dir?) and b) having to set it explicitly when other configuration is set via `opts` and/or the `Host` object.

poszu:
@dshulyak, I embedded `pubsub`, `discovery` and `bootstrap` to hide them from an operator.

PR #3574, opened 2022-09-20
[Merged by Bors] - Use gotestsum to run unit tests

poszu:
## Motivation
`go test` doesn't print a summary of run tests at the end of a run. In particular, it doesn't print **failed** tests. This makes it difficult to quickly find the failing test. I propose to use [gotestsum](https://github.com/gotestyourself/gotestsum) wrapper instead.

## Changes
- use gotestsum to run unit tests and create a junit XML report.
- added a step to parse and present the junit test report.

An example of failed unit tests ran with gotestsum: https://github.com/spacemeshos/go-spacemesh/actions/runs/3089490869#summary-8448033184

Example Junit Test Report: https://github.com/spacemeshos/go-spacemesh/runs/8448163116

 

## DevOps Notes


Makefile
```diff
@@ -123,7 +124,7 @@ endif
 test: UNIT_TESTS = $(shell go list ./...  | grep -v systest)
 
 test: get-libs
-	$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) go test -timeout 0 -v -p 1 $(UNIT_TESTS)
+	$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) gotestsum -- -timeout 0 -p 1 $(UNIT_TESTS)
```
fasmat, line 127: would `TEST_DIRECTORY` instead of `UNIT_TEST` work here as well? It's part of `gotestsum`'s documentation.

Also: why are there two `test` targets defined in this `Makefile`?

fasmat (approved):
Just a minor thing, besides that lgtm

Makefile
poszu: My guess is that both `test` target definitions are executed. 

Using `TEST_DIRECTORY` could work, I think. I didn't see any benefit so I limited the scope of changes to a minimum.

dshulyak (approved):

evt (approved):

Makefile
fasmat: I would still get rid of having two `test` targets and instead set `UNIT_TESTS` or `TEST_DIRECTORY` in the same block as calling `gotestsum`.

countvonzero (approved):

poszu:
bors merge

PR #3577, opened 2022-09-21
[Merged by Bors] - Run staticcheck in CI

### LINKED ISSUE #3576: We should start running `staticcheck` in CI
## Description
From https://staticcheck.io:
>Staticcheck is a state of the art linter for the [Go programming language](https://go.dev/). Using static analysis, it finds bugs and performance issues, offers simplifications, and enforces style rules.

There already is a target to run staticcheck in the makefile. However, CI doesn't run it. I propose to include running staticcheck in the CI workflow.

poszu:
## Motivation
We could run [staticcheck](https://staticcheck.io) linter in CI to catch common mistakes, bugs, unused code etc.
Closes #3576 

## Changes
- run statickcheck in _quicktests_  job in CI
- fixed all existing issues reported by statickcheck

## DevOps Notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,10 @@ jobs:
           make
           make tidy
           make test-fmt
+      - name: staticcheck
+        run: |
+          go install honnef.co/go/tools/cmd/staticcheck@latest
+          make staticcheck
 
   golangci-lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,7 @@ jobs:
           make tidy
           make test-fmt
       - name: staticcheck
-        run: |
-          go install honnef.co/go/tools/cmd/staticcheck@latest
-          make staticcheck
+        run: make staticcheck
 
   golangci-lint:
     runs-on: ubuntu-latest
--- a/activation/post.go
+++ b/activation/post.go
@@ -414,10 +414,3 @@ func (mgr *PostSetupManager) LastOpts() *PostSetupOpts {
 func (mgr *PostSetupManager) Config() PostConfig {
 	return mgr.cfg
 }
-
-func (mgr *PostSetupManager) getState() postSetupState {
-	mgr.mu.Lock()
-	defer mgr.mu.Unlock()
-
-	return mgr.state
-}
--- a/activation/post.go
+++ b/activation/post.go
@@ -225,9 +225,7 @@ func (mgr *PostSetupManager) Benchmark(p PostSetupComputeProvider) (int, error) 
 // It supports resuming a previously started session, as well as changing the Post setup options (e.g., number of units)
 // after initial setup.
 func (mgr *PostSetupManager) StartSession(opts PostSetupOpts) (chan struct{}, error) {
-	mgr.mu.Lock()
-	state := mgr.state
-	mgr.mu.Unlock()
+	state := mgr.getState()
 
 	if state == postSetupStateInProgress {
 		return nil, fmt.Errorf("post setup session in progress")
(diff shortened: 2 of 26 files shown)


.github/workflows/ci.yml
```diff
@@ -65,6 +65,10 @@ jobs:
           make
           make tidy
           make test-fmt
+      - name: staticcheck
```
fasmat, line 76: I think this should be part of `make install` since `make staticcheck` assumes that the tool is available already

activation/post.go
```diff
@@ -414,10 +414,3 @@ func (mgr *PostSetupManager) LastOpts() *PostSetupOpts {
 func (mgr *PostSetupManager) Config() PostConfig {
 	return mgr.cfg
 }
-
-func (mgr *PostSetupManager) getState() postSetupState {
```
fasmat: rather than removing this method, we could use it where `mgr.state` is read from, i.e. at the beginning of `StartSession` in this file and a few other places

cmd/integration/server.go
```diff
@@ -79,28 +77,3 @@ func (s *server) start(args []string) error {
 
 	return nil
 }
-
-// shutdown terminates the running node server process, and cleans up
-// all files/directories created by it.
-func (s *server) shutdown() error {
```
fasmat, line 85: I think this method should not be obsolete, just like `stop`. This feels to me like its not called where it is supposed to be called (probably in some `t.Cleanup()`s)

hare/haretypes.go
```diff
@@ -346,13 +346,6 @@ func (s *Set) add(id types.ProposalID) {
 	s.values[id] = struct{}{}
 }
 
-func (s *Set) remove(id types.ProposalID) {
```
fasmat, line 349: Something feels odd here.

Add & Remove (and some other methods) appear twice on `Set` (lower and upper case). I think the `add` method can be removed by inlining too? Same for `contains` and `len`, possibly also `initWithSize`

fasmat (changes requested):
So far many changes that I like to see! but I couldn't resist adding some small suggestions ;-)

.github/workflows/ci.yml
poszu: Good idea, done.

activation/post.go
poszu: 👍

cmd/integration/server.go
poszu: Good point. Actually, `stop()` does nothing extra besides calling `shutdown()`. I will remove `stop()` and add calling `shutdown()` in `main()`.

hare/haretypes.go
poszu: I totally agree it needs refactoring. As we discussed offline, I will restrain from too much refactoring of `Set` in this PR as it is more work and it deserves a separate PR. It also seems that there might be some race conditions.

hare/haretypes.go
```diff
@@ -94,8 +94,11 @@ func NewSet(data []types.ProposalID) *Set {
 	s.isIDValid = false
 
 	s.initWithSize(len(data))
+
+	// SAFETY: It's safe not to lock here as `s` was just
+	// created and nobody else has a access to it yet.
```
poszu:
```suggestion
	// created and nobody else has access to it yet.
```

cmd/integration/harness.go
```diff
@@ -86,6 +83,7 @@ func main() {
 		log.With().Error("harness: an error has occurred while generating a new harness:", log.Err(err))
 		log.Panic("error occurred while generating a new harness")
 	}
+	defer h.server.stop()
```
fasmat: this is unfortunately never executed, because `dummyChan` is never closed or written to so main blocks on line 103

I think what is missing here is a go-routine that listens for `SIGINT` and then executes `h.server.stop()` to shutdown all running go routines.

This would also remove the need for the goroutine below (just select on the main goroutine) and `dummyChan`

hare/haretypes.go
```diff
@@ -94,8 +94,11 @@ func NewSet(data []types.ProposalID) *Set {
 	s.isIDValid = false
 
 	s.initWithSize(len(data))
+
+	// SAFETY: It's safe not to lock here as `s` was just
+	// created and nobody else has access to it yet.
```
fasmat, line 99: 💡

fasmat (changes requested):
The `main` in the harness package can be improved, besides that it now looks ok.

cmd/integration/harness.go
poszu: Refactored the whole main function.

fasmat (approved):

poszu:
bors merge

poszu:
bors merge

Issue #3573, opened 2022-09-20
tortoise: update opinion on the terminated layers within sliding window when received new data 


dshulyak:
depends on https://github.com/spacemeshos/go-spacemesh/issues/3569

once layer is verified we do not revisit opinion on that layer unless rerun is triggered. it was assumed that syncer will trigger a rerun when it notices a good reason for it. opinion may change because of the following cases:
- hare output changed within hdist
- node downloaded atxs with additional ballots, previously counted ballots won't be sufficient to cross global threshold
- ballots from higher layers voted differently from the previous layers and changed threshold for already terminated layer
- counting delayed ballots changed opinion on already terminated layer

from that list above only 2nd case can be tracked successfully by syncer, in other cases it is either impossible (4th) or inefficient (1st and 3rd) to trigger full rerun.

## implementation plan
revisit opinions on terminated layer if changes are within sliding window
### epoch weight
it is already computed in current implementation, but there are some changes:
```go
// float64 is not used in code
func expectedWeight(epochs map[uint32]uint64, target, last layer) float64 {
  weight := 0
  if target.epoch() == last.epoch() {
    weight = epochs[target.epoch()]
    return weight*(last.ordinal()-target.ordinal())/target.epoch().length()
  }
  weight += epochs[target.epoch()]*(target.epoch().length() - target.epoch().ordinal()) / target.epoch().length()
  for epoch := target.epoch() + 1 < last.epoch(); epoch++ {
    weight += epochs[epoch]
  }
  weight += epochs[last.epoch()]*(last.epoch().ordinal() / last.epoch().length())
  return weight
}
```
### OnAtx handler
add OnAtx handler to the tortoise. every time when tortoise receives new atx update expected weight by adding full atx weight or a fraction, using same computation as above. OnAtx is expected to be called before OnBallot that references such atx.
### related tortoise state
```go
type state struct {
  // weight of individual atxs in the sliding window
  atxs map[id]uint64
  // referenceWeight is computed by dividing atx weight by an eligibility count when reference ballot is received
  // to get ballot weight - reference weight is multiplied by eligibilities attached to the ballot
  referenceWeight map[id]float64
  // weight of all epochs (only atxs that are were used within sliding window)
  epochs map[uint32]uint64
  // if missed weight is larger than 1/3 of epochs weight - trigger state reload (rerun from scratch)
  // this is specifically for atxs that weren't counted within sliding window
  // we rely on p2p layer and syncer to deduplicate this
[...]

PR #3581, opened 2022-09-22
[Merged by Bors] - tortoise: state refactor

### LINKED ISSUE #3569: tortoise: faster and memory-efficient data structure for storing votes
supersedes: https://github.com/spacemeshos/go-spacemesh/issues/3007, https://github.com/spacemeshos/go-spacemesh/issues/3005

### local opinion
```go
type id [20]byte

type vote uint8

// ~30bytes of metadata + N*blockInfo (N is expected to be 1)
type layerInfo struct {
         // empty weight is used to count weight if there are no blocks in the layer
  	 empty weight
         blocks []blockInfo
  
  	 // used by verifying tortoise
  	 verifying struct {
              // good weight is a weight of good ballots in this layer
              good weight
              // abstained weight is the weight of ballots that are counted by
              // verifying tortoise that abstained on this layer
              abstained weight
              referenceHeight uint64 
     }
}

// around ~50bytes
type blockInfo struct {
        id id
  	layer uint32
        height uint64
    
  	// hare may support only one block
  	// if hare is not termianted - all blocks will have neutral
  	hare vote
        // validity is either loaded from database or computed by comparing margin with threshold
  	validity vote // support, against, neutral
  	margin weight
}
```

### ballot opinions (global opinion)
```go
type layerVote struct {
    layer *layerInfo
    vote vote // valid values are abstain or against
    votes []*blockVote
}

type blockVote struct {
    block *blockInfo
    vote vote
}

type ballotInfo struct {
      id id
      height uint64
      weight weight
      goodness goodness
      // beacon is copied from reference ballot
      // to check if beacon is consistent with local beacon compare values
      beacon [20]byte 
      
      // copied from base ballot, updated with a difference
      // note that this is a naive version, and likely can't be used in practice
      votes []*layerVote
}
```

### full state
```go
type state struct {
  // localOpinion
  localOpinion []layerInfo
  
  // evicted starts with effective genesis - 1
  evicted uint32
  // votes are stored only for layers in the sliding window 
  votes 	   [][]ballotInfo
  
  // ballots reference is used to get base and reference ballot
  ballots map[id]*ballotInfo
  // blocks references are used t 
  blocks  map[id]*blockInfo
}
```

### on block
- add block to the correct layer in the local opinion  
- if there are ballots that can vote on this block, add vote on this block to every ballotInfo.votes
### on hare output
- expected to be called after block was added to the state  
- find a block in the local opinion, set hare validity to support for this block, set other blocks in same layer to against
### on ballot
- expected to be called after referenced blocks, base and ref ballots are added to the state  
- copy pointers to votes from base ballot  
  - with 2000 sliding window naive implementation copies ~ 16KB per ballot (2000 pointers),  to keep all votes in memory we will need ~1.6GB  
- iterate over support, against, abstain arrays - modify ballotInfo.votes based on those differences

### supported operations:
- iterate over all votes in any direction (either in order or in reverse, no need to support both directions)
- efficiently copy base ballot votes, also without huge memory waste
#### bit per vote
bit per vote encoding. for sliding window 2000 every ballot will store ~250bytes (2000 / 8). this is also assumes that there is atmost one block per layer. if not - encoding becomes much more complex
#### persistent list with votes
copy base ballot last vote from list where opinion was the same and update with new opinions.
example:
- A: 1 <- 2 <- 3 <- 4  
- B doesn't overwrite any votes of A 4 <- 5  
    - this is a perfect case. every new ballot will add only 2 pointers (16 bytes) to store votes. 800 bytes per 50 ballots. with 2 000 layers in the window - ~16mb. it doesn't take into account actual data (e.g ballot id, weight, beacon) but that is negligible in comparison with votes.  
    - we can assume that majority of the nodes will try to use this pattern, with the exception of bugs and attacks  
- C overwrites some previous votes: 2 <- 3' <- 4' <- 5  
    - synthetic worst case is if every ballot voting pattern votes differently than the base starting from effective genesis. every ballot adds 2*2000 pointers (32kb per ballot), 3.2GB per sliding window.         
### how to reset weight for verifying tortoise?  
iterate over all layers that needs to be reset and set verifying.weight and verifying.abstained to zero
### what happens with votes if verified layer is below layer that supposed to be evicted? 
the memory will grow above expected value. it could happen if partition existed for longer then sliding window. but there is no better option

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3569

main changes:
- removed auxiliary maps to track various data about blocks and ballots. all data is stored directly in layerInfo, ballotInfo or blockInfo objects.
- votes are represented as a reverse linked list. child ballot copies votes list from the base ballot and extends it with new votes. if it overwrites some of the opinions below base ballot - it creates a copy of the list starting at a layer where opinion was changed. in future there is a room for optimization by reusing list for ballots that vote consistently with local opinion.
- votes outside of the window are evicted by iterating reverse linked list and dropping pointers for votes that are outside sliding window. i didn't notice any overhead with window size of 10 000 
- removed 'abstained' condition that was used for verifying tortoise, as new data structure allows to inherit abstained votes without additional work

other info:
- this change doesn't count votes for blocks that are received after layer was terminated. change is already big and that bug fits in the scope of https://github.com/spacemeshos/go-spacemesh/issues/3573
- representing votes this way allows to simplify the process of encoding votes, i will address it in future prs
- in profile it is obvious that doing weight computation with big.Rat with large sliding window is inefficient. using https://github.com/spacemeshos/fixed allows to reduce BenchmarkRerun by almost two times and lowers memory requirement. most likely will switch to it in the future pr


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

common/util/weight.go
```diff
@@ -33,6 +33,12 @@ type Weight struct {
 
 // Add adds `other` to weight.
 func (w Weight) Add(other Weight) Weight {
+	if w.Rat == nil {
```
countvonzero, line 36: adding this check highlights the inconsistency of this file. for example, why not do the same check for `Div()`?

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+		layer    types.LayerID
+		height   uint64
+		hare     sign
+		validity sign
+		margin   weight
+	}
+
+	state struct {
+		// last received layer
+		// TODO should be last layer according to the clock
```
countvonzero, line 37: reference the issue https://github.com/spacemeshos/go-spacemesh/issues/2921

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+	verifyingInfo struct {
+		good, abstained weight
+		referenceHeight uint64
+	}
+
+	layerInfo struct {
+		lid            types.LayerID
+		empty          weight
+		hareTerminated bool
+		blocks         []*blockInfo
```
countvonzero, line 21: curious why not having []*ballotInfo here too? do layerInfo and ballotInfo have different life-span?

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+
+type votes struct {
+	tail *layerVote
+}
+
+func (v *votes) append(lv *layerVote) {
+	if v.tail == nil {
+		v.tail = lv
+	} else {
+		v.tail = v.tail.append(lv)
```
countvonzero, line 178: it seems error-prone when you don't check if layer ID is consecutive
for example, the caller loop inserted a continue stmt in later edits.

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+func (l *layerVote) append(lv *layerVote) *layerVote {
+	lv.prev = l
+	return lv
+}
+
+func (l *layerVote) update(from types.LayerID, diff map[types.LayerID]map[types.BlockID]sign) *layerVote {
+	if l.lid.Before(from) {
+		return l
+	}
+	l = l.copy()
```
countvonzero: i wonder if there are some hidden issue when you update the method receiver here. 
what's the intention here? why not simply update `l` in place? or just use a new var?

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+		votesBeforeBase bool
+	}
+
+	ballotInfo struct {
+		id         types.BallotID
+		layer      types.LayerID
+		base       baseInfo
+		height     uint64
+		weight     weight
+		beacon     types.Beacon
```
countvonzero, line 153: does this need to remain in memory after you determine `conditions.badBeacon`?

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+		*blockInfo
+		vote sign
+	}
+
+	baseInfo struct {
+		id    types.BallotID
+		layer types.LayerID
+	}
+
+	conditions struct {
```
countvonzero, line 137: how about a bit vector?

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+		// last evicted layer
+		evicted types.LayerID
+
+		localThreshold  weight
+		globalThreshold weight
+
+		// epochWeight average weight per layer of atx's that target keyed epoch
+		epochWeight map[types.EpochID]weight
+		// referenceHeight is a median height from all atxs that target keyed epoch
+		referenceHeight map[types.EpochID]uint64
```
countvonzero, line 56: given that the goal is trying to be memory efficient, maybe it's worth consolidating maps with the same key type.
i.e. combine epochWeight/referenceHeight to be a map[types.EpochID]epochInfo that has two uint64

same for maps with LayerID/BallotID

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
-type ballotInfo struct {
-	id     types.BallotID
-	height uint64
-	weight util.Weight
+type (
+	weight = util.Weight
+
+	verifyingInfo struct {
+		good, abstained weight
+		referenceHeight uint64
```
countvonzero, line 14: please keep the old comment for this ("highest block height below reference height")

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+func (s *state) addBlock(block *blockInfo) {
+	layer := s.layer(block.layer)
+	layer.blocks = append(layer.blocks, block)
+	s.blockRefs[block.id] = block
+	s.updateRefHeight(layer, block)
+}
+
+func (s *state) updateRefHeight(layer *layerInfo, block *blockInfo) {
+	if layer.verifying.referenceHeight == 0 && layer.lid.After(s.evicted) {
+		layer.verifying.referenceHeight = s.layer(layer.lid.Sub(1)).verifying.referenceHeight
```
countvonzero: say layer.lid.Sub(1) is an empty layer (no blocks). so addBlock() is not called for lid.Sub(1)
in that case, should the code keep going back in layers until it finds a non-zero layer reference height?

tortoise/tortoise.go
```diff
@@ -904,40 +831,112 @@ func (t *turtle) layerCutoff() types.LayerID {
...
+	from := base.layer
+	diff := map[types.LayerID]map[types.BlockID]sign{}
+	for vote, bids := range map[sign][]types.BlockID{
+		support: exceptions.Support,
+		against: exceptions.Against,
+	} {
+		for _, bid := range bids {
+			block, exist := t.blockRefs[bid]
+			if !exist {
+				ballot.conditions.votesBeforeBase = true
```
countvonzero, line 842: when the ballot voted for a block the nodes hasn't seen before, why is `votesBeforeBase` set to true? how is this related to the base ballot?

tortoise/tortoise.go
```diff
@@ -904,40 +831,112 @@ func (t *turtle) layerCutoff() types.LayerID {
...
+				continue
+			}
+			if block.layer.Before(from) {
```
countvonzero, line 845: compare with base.layer?

tortoise/tortoise.go
```diff
@@ -904,40 +831,112 @@ func (t *turtle) layerCutoff() types.LayerID {
...
+			}
+			layerdiff[block.id] = vote
+		}
 	}
-	genesis := types.GetEffectiveGenesis()
-	limit := genesis
-	if last.After(genesis.Add(config.Zdist)) {
-		limit = last.Sub(config.Zdist)
+	for _, lid := range exceptions.Abstain {
+		if lid.Before(from) {
```
countvonzero, line 858: from could be modified from line 849. why not directly compare with base.layer?

tortoise/state.go
```diff
@@ -5,78 +5,242 @@ import (
...
+	return lv
+}
+
+func (l *layerVote) update(from types.LayerID, diff map[types.LayerID]map[types.BlockID]sign) *layerVote {
+	if l.lid.Before(from) {
+		return l
+	}
+	l = l.copy()
+	if l.prev != nil {
+		l.prev = l.prev.update(from, diff)
```
countvonzero: not a fan of recursion. i know it's more elegant but wouldn't this be a source of memory overhead compared to an iterative approach.

as a reader, it's more intuitive to iterate the diff from `from` to `ballot.layer.Sub(1)`, and have `next` instead `prev`
the code as such, the main part of `layerVote.update()` is very similar to the last part of `decodeException` (// add new opinions after the base layer)

tortoise/state.go
dshulyak: i don't see hidden issues, but i can use new var

tortoise/state.go
dshulyak: i realize that there will be additional overhead, but it is not about elegance. list supports iteration only in one direction, for example:
1 <- 2 <- 3 <- 4
will allow to iterate from 4 to 1. what i need to do is to find the first layer where opinion changed and cope list from that point

for example child ballot changes opinion on layer 2, so the list should become

1 <- 2' <- 3' <- 4'

tortoise/state.go
dshulyak: i actually tried that and it was harder to grock, memory savings are not very important i think (4 booleans vs 1 byte)

tortoise/state.go
dshulyak: you are right, technically beacon matters only in ref ballot. but there are some nuances when beacon can be dropped from ref ballot. let me check that in the upcoming pr, as it will change when badBeacon is determined

tortoise/state.go
dshulyak: probably can be kept in the layerInfo, i will check

tortoise/state.go
dshulyak: `next` doesn't work for several reasons, but the most significant is:
 
given the same list 1 -> 2 -> 3 -> 4
and for example i want to append new opinion -> 5

i can't simply 1 -> 2 -> 3 -> 4 -> 5 because it modifies 4, but if i copy 4 i also need to copy 3 and so on

tortoise/state.go
dshulyak: looks like it is not robust indeed

tortoise/tortoise.go
dshulyak: in this case block layer can't be determined since block is not in the sliding window, so it is implicitly below base ballot, since base ballot is in the sliding window

tortoise/tortoise.go
dshulyak: so the thing here is that i want to find minimal layer that is below base.layer. and the current approach works because votesBeforeBase is set to true the first time `from` is updated

i can change it to two comparisons

tortoise/tortoise.go
dshulyak: same as above

tortoise/state.go
dshulyak: i will add a panic. that would be a programmer mistake if it happens

tortoise/state.go
dshulyak: moved to layerInfo

tortoise/state.go
dshulyak: tried that just now and i have to refactor even more tests, i would rather do it some time in the future. having those things as they are doesn't really affect memory much

tortoise/tortoise.go
dshulyak: changed to two comparisons

common/util/weight.go
dshulyak: added to other methods

tortoise/state.go
countvonzero: yeah... thinking more about it, `next` doesn't work indeed.

tortoise/state.go
dshulyak: added function to find weight from the last non-empty layer (last commit)

countvonzero (approved):

dshulyak:
bors merge

PR #3584, opened 2022-09-22
[Merged by Bors] - Enable caching Go mod and build files in CI

poszu:
## Motivation
Caching Go's package and build files should speed up CI

## Changes
Added steps to cache Go's module directory and build cache to ci.yml

## DevOps Notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,6 @@ jobs:
           key: cache-go-${{ env.go-version }}-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             cache-go-${{ env.go-version }}-
-            cache-
       - name: fmt, tidy, lint
         run: |
           make
@@ -96,7 +95,6 @@ jobs:
           key: cache-go-${{ env.go-version }}-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             cache-go-${{ env.go-version }}-
-            cache-
       - name: setup env
         run: make
       - name: golangci-lint
@@ -128,7 +126,6 @@ jobs:
           restore-keys: |
             cache-go-${{ env.go-version }}-unittests-
             cache-go-${{ env.go-version }}-
-            cache-
       - name: setup env
         run: make
       - name: Clear test cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,8 @@ jobs:
             cache-
       - name: setup env
         run: make
+      - name: Clear test cache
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +60,15 @@ jobs:
...
           go-version: ${{ env.go-version }}
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: cache-go-${{ env.go-version }}-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            cache-go-${{ env.go-version }}-
+            cache-
```
fasmat: This is a bit confusing to me; we want to restore a cache built for that exact go version and the exact dependencies. If we didn't find one, then we restore the next best built for the same go version. That's OK so far, but if we don't find one for that go version this action restores any cache?

I think just `cache-go-${{ env.go-version }}-` is enough as a fallback and we just rebuild our caches when the go version changes.

.github/workflows/ci.yml
```diff
@@ -101,8 +119,20 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ env.go-version }}
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: cache-go-${{ env.go-version }}-unittests-${{ hashFiles('**/go.sum') }}
```
fasmat, line 125: I think we can use the same cache name here as in the other steps, it's running on the same code as the other steps, no need to build a separate cache just for the unit-tests:

```yaml
key: cache-go-${{ env.go-version }}-${{ hashFiles('**/go.sum') }}
restore-keys: |
    cache-go-${{ env.go-version }}
```

fasmat (changes requested):
I would not restore any cache if we don't find one built for the exact go version that's used in the build, besides that looks ok to me.

.github/workflows/ci.yml
poszu: The problem is that multiple jobs using the same cache key that run **in parallel** race on storing the cache. It's OK for `quicktests` and `golangci-lint` as both just pull the dependency go modules. Whichever finishes first will store the cache. 

But `unittests` job also builds tests. As it takes the longest time to finish, the build cache would always be discarded, which is not what we want.  With a distinct key for `unittests`, its cache is always saved.

fasmat (approved):

poszu:
bors merge

Issue #3503, opened 2022-08-27
signer: use genesis id as domain separator for protocol messages


dshulyak:
part of: https://github.com/spacemeshos/pm/issues/135
depends on: https://github.com/spacemeshos/go-spacemesh/issues/3505

https://github.com/spacemeshos/go-spacemesh/blob/develop/signing/signer.go

- initialize EdSigner with GenesisID (preferably pass as an option to NewEdSigner/NewEdSignerFromBuffer/NewEDVerifier)
- prefix message that is to be signed and verified with GenesisID

PR #3587, opened 2022-09-25
signing: add signer with genesis id

evt:
## Motivation

Closes #[3503](https://github.com/spacemeshos/go-spacemesh/issues/3503)

## Changes

Add signer with genesis id, add test

## TODO

## DevOps Notes


activation/activation_test.go
```diff
@@ -98,7 +98,7 @@ func (n *NetMock) hookToAtxPool(transmission []byte) {
 }
 
 func NewMockSigner() *MockSigning {
-	return &MockSigning{signing.NewEdSigner()}
+	return &MockSigning{signing.NewEdSigner([20]byte{})}
```
dshulyak, line 101: i think this is why using config option was a better approach

signing/signer.go
```diff
@@ -154,23 +163,24 @@ func Verify(pubkey *PublicKey, message []byte, sign []byte) bool {
...
-
 // Verify that signature matches public key.
-func (EDVerifier) Verify(pub *PublicKey, msg, sig []byte) bool {
+func (ev EdVerifier) Verify(pub *PublicKey, msg, sig []byte) bool {
+	if len(msg) < 20 {
+		return false
+	}
+	if string(msg[0:20][:]) != string(ev.genesisID[:]) {
+		return false
+	}
```
dshulyak, line 178: this implementation is incorrect. message should not include genesis id. verifier should prefix message before attempting to extract public key

evt:
Closed in favour of [3597](https://github.com/spacemeshos/go-spacemesh/pull/3597)

activation/activation_test.go
evt: Made it as an option in https://github.com/spacemeshos/go-spacemesh/pull/3597

signing/signer.go
evt: done this way in https://github.com/spacemeshos/go-spacemesh/pull/3597

PR #3589, opened 2022-09-26
[Merged by Bors] - Add create cover profile to makefile

fasmat:
## Motivation
This adds a target to the `Makefile` that creates a cover profile. This file can be used to visualize which parts of the code are covered by tests and which not.

## Changes
New target in Makefile

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,17 +7,14 @@ include Makefile-gpu.Inc
 #include Makefile-svm.Inc
 
 DOCKER_HUB ?= spacemeshos
-TEST_LOG_LEVEL ?=
+UNIT_TESTS ?= $(shell go list ./...  | grep -v systest)
 
 COMMIT = $(shell git rev-parse HEAD)
 SHA = $(shell git rev-parse --short HEAD)
 BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 
-export GO111MODULE := on
 export CGO_ENABLED := 1
 export CGO_CFLAGS := "-DSQLITE_ENABLE_DBSTAT_VTAB=1"
-
-PKGS = $(shell go list ./...)
 
 # These commands cause problems on Windows
 ifeq ($(OS),Windows_NT)
@@ -64,7 +61,6 @@ DOCKER_IMAGE = $(DOCKER_IMAGE_REPO):$(BRANCH)
 ifeq ($(BRANCH),$(filter $(BRANCH),staging trying))
   DOCKER_IMAGE = $(DOCKER_IMAGE_REPO):$(SHA)
 endif
-
 
 install:
 	go run scripts/check-go-version.go --major 1 --minor 18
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,18 +7,15 @@ include Makefile-gpu.Inc
...
 
 DOCKER_HUB ?= spacemeshos
-TEST_LOG_LEVEL ?=
+UNIT_TESTS ?= $(shell go list ./...  | grep -v systest)
 
 COMMIT = $(shell git rev-parse HEAD)
 SHA = $(shell git rev-parse --short HEAD)
 BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 
-export GO111MODULE := on
```
poszu, line 16: 👍

Makefile
```diff
@@ -125,19 +121,18 @@ endif
 # Clear tests cache
 clear-test-cache:
 	go clean -testcache
-
-test: UNIT_TESTS = $(shell go list ./...  | grep -v systest)
+.PHONY: clear-test-cache
```
poszu, line 124: Thx, I forgot about it!

Makefile
```diff
@@ -125,19 +121,18 @@ endif
...
 
 test: get-libs
-	$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) gotestsum -- -timeout 0 -p 1 $(UNIT_TESTS)
+	@$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" gotestsum -- -timeout 0 -p 1 $(UNIT_TESTS)
 .PHONY: test
```
poszu, line 128: `TEST_LOG_LEVEL` is used in tests for tweaking the log level. If you remove this, there will be no other sign that one can tweak the log level apart from the source code. How about adding a short note in the README that log level in tests can be modified by setting `TEST_LOG_LEVEL`?

Makefile
```diff
@@ -177,12 +172,16 @@ golangci-lint-github-action:
...
-	  $(foreach pkg,$(PKGS),\
+	  $(foreach pkg,$(UNIT_TESTS),\
 		go test -coverprofile=cover.out -covermode=count $(pkg);\
 		tail -n +2 cover.out >> cover-all.out;)
 	go tool cover -html=cover-all.out
 .PHONY: cover
 
+cover-profile:
+	@$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" go test -coverprofile=cover.out -timeout 0 -p 1 $(UNIT_TESTS)
+.PHONY: cover-profile
```
poszu: It's not clear to me why there is a second coverage target. Could you please leave a comment in the Makefile justifying it?

Btw, it would be super nice to run coverage in CI 😉.

poszu (changes requested):

Makefile
fasmat: I'm not 100% sure what the other target is for. It uses `covermode=count` instead of the default `covermode=set` and it iterates over all packages manually, instead of just making one call.

What it is probably used for can be seen by the last line: it starts a server that visualizes the generated coverprofile in the browser.

The `cover.out` file created by the new `cover-profile` target is just a file that can be loaded in VScode to visualize which codepaths are covered (green background) and which are not (red background)

Makefile
fasmat: Updated README.md

Makefile
poszu: Perhaps let's initiate a discussion to see if anybody actually uses the `cover` target. Then:
- if nobody uses it - remove it,
- if somebody uses it - we will know why

@dshulyak @countvonzero @evt What do You think?

fasmat: Alternatively these targets can also be cleaned up nicely (assuming `covermode=set` works for the HTML tool as well as `covermode=count`.

```suggestion
cover.out:
	@$(ULIMIT) CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)" go test -coverprofile=cover.out -timeout 0 -p 1 $(UNIT_TESTS)
	
cover-profile: cover.out
.PHONY: cover-profile
	
cover-tool: cover.out
	@go test cover -html=cover.out
.PHONY: cover-tool

clear-cover-profile:
	@rm cover.out
.PHONY: clear-cover-profile
```

poszu (approved):

fasmat:
bors merge

PR #3594, opened 2022-09-27
[Merged by Bors] - Use a dedicated type for PoET proof ref

poszu:
## Motivation
This PR presents a practical use of leveraging a type system to increase code readability, the expressiveness of APIs and to prevent API misusage.

The idea is to use a dedicated "new" type instead of a generic primitive (for example use `type PoetProofRef []byte` instead of `[]byte`). Some of the benefits are:

- It is clear what a variable "is for" - this information is embedded in its type. For example, it is immediately clear what a `PoetProofRef` is for, in contrast to `[]byte`. Consider an example interface: https://github.com/spacemeshos/go-spacemesh/blob/develop/activation/interface.go#L12-L16. It's hard to tell what is expected to be passed as an argument exactly.
- It limits the usage of a given type of variable throughout the program. It is easier to understand where it comes from and what can be done with it (IDE's features like _find all references_ begin to be helpful).
- ~It prevents mistakes of passing a wrong thing to some API **at compile time**. It takes the burden of questioning "Am I passing the right thing to this function?" from the developer and moves it to the compiler.~

Some cons:
- it might require adding some "glue" casts here and there where APIs are not fully compatible (not 100% ported to use the new type), this especially manifests in tests. On the other hand, having to do an explicit cast should trigger thinking if one is doing the correct thing, to begin with.
- When used too much and without thinking, it might actually reduce the readability of the code.

**Edit:** I realized that Go actually allows for implicit conversions so the 3rd point is invalid :(

## Changes
Replaces usage of `[]byte` with a new type `type PoetProofRef []byte` where applicable.

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/activation/interface.go
+++ b/activation/interface.go
@@ -10,9 +10,9 @@ import (
 //go:generate mockgen -package=mocks -destination=./mocks/mocks.go -source=./interface.go
 
 type poetValidatorPersistor interface {
-	HasProof([]byte) bool
+	HasProof(types.PoetProofRef) bool
 	Validate(types.PoetProof, []byte, string, []byte) error
-	StoreProof([]byte, *types.PoetProofMessage) error
+	StoreProof(types.PoetProofRef, *types.PoetProofMessage) error
 }
 
 type nipostValidator interface {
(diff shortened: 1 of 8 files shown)


dshulyak (approved):

countvonzero (approved):

poszu:
bors merge

PR #3596, opened 2022-09-28
[Merged by Bors] - tortoise: tally votes for layers independently of the hare

### LINKED ISSUE #2919: tortoise should use latest ticked layer to offset hdist
## Description

currently tortoise uses the last layer hare outputs `t.Last` to calculate offset from hdist (the cutoff).
https://github.com/spacemeshos/go-spacemesh/blob/9c8b62c01095445d7e3c06faeb37b3803b92456d/tortoise/verifying_tortoise.go#L1032

this cutoff is used in two places:
- within hdist, we use hare opinion. outside hdist, we use global/blocks opinions and break ties with weak coin.
the weak coin for each layer is supposed to sway nodes in the same direction when votes don't cross the threshold. however, when nodes can potentially have different `t.Last` value, they can have a different cutoff value and hence different layer ranges to use weak coin

- the range of self healing. currently self-healing only heal layers in (verified, lastest-hdist). if hare get stuck, `t.Last-hdist` stays unchanged. self-healing will not heal past `t.Last-hdist`

we should use the last ticked layer (i.e. the current layer), as the rest of the system, to consider the range to use weak coin and do self-healing

### LINKED ISSUE #3575: tortoise: counting ballots from layer should not depend on hare termination
depends on #3569 

ballots from layer N are not counted until hare for layer N terminates (cert is downloaded) or zdist passes.  it doesn't have impact within hdist, because we vote according to hare output, but if votes will be invalid if node casts vote on block outside of hdist without counting all received ballots from higher layers.

this change proposes to remove OnLayerTerminated and use only data received using OnBallot/OnBlock/OnHareOutput handlers.

## counting weight dependencies
- base ballot and reference ballot should be added to the state  
- availability of the beacon in layers N epoch (matters both for full and verifying tortoise)  
- blocks from previous layers that are referenced by a ballot should be in the state  
	- if median tick height is not computed when block is added to the tortoise, tortoise won't be able to add a block  
- hare output from layer N-1 should be in the state for verifying tortoise. alternatively hare output can be added later, but it will trigger short rerun. 

## count weight in OnBallot
the simplest solution is to count votes when OnBallot handler is called, based on the above dependencies, there are two events where it can fail:
- beacon for the epoch is not stored in the state  
- reference height that is used for verifying tortoise wasn't yet computed, it is expected to be computed in the first layer of the epoch  
In both of those cases we need to have some data available locally (beacon and atxs).

## add TallyVotes(allowed layer)
TallyVotes updates highest layer number that should be safe to count. All ballots from (previous allowed, current allowed] are counted when this method is called. Afterwards, ballots that are received in OnBallot handler with layer <= current allowed are counted immediately. 

During sync - syncer will be calling this method explicitly when it knows that we have enough data to count votes (e.g beacon and atxs are downloaded). If node is in synced state - tortoise will consider layer to be safe to count when clock emits the layer.

### ballots that have different beacon should be delayed by N layers. if OnLayerTerminated is removed - when tortoise counts them?
TallyVotes will be used as a notification that it is safe to count delayed ballots
### if node was offline for a short period of time and missed broadcast of some ballots
when those ballots are downloaded they will be added to the tortoise and counted using OnBallot handler

dshulyak:
depends on: https://github.com/spacemeshos/go-spacemesh/issues/3569
closes: https://github.com/spacemeshos/go-spacemesh/issues/3575
closes: https://github.com/spacemeshos/go-spacemesh/issues/2919

1. onBlock and onBallot refactored in a way to avoid failures if required state wasn't loaded and to support counting ballots separately
 
onBlock will not fail if reference height wasn't computed yet. reference height is always computed together with the first layer in the epoch. this change makes it safe to call onBlock before such first layer was processed.

onBallot will not fail if beacon is not yet available. beacon needs to be known only before counting ballot (in both modes). this change (and changes to onBlock) allows to decode votes before calling a method to verify a layer,  in practice decoding votes will be the most expensive operation. with an exception of counting votes in full mode.

this refactoring also allows to seamlessly count votes from late ballot without any additional logic in place, this will be required for https://github.com/spacemeshos/go-spacemesh/issues/3573.

2. new TallyVotes method

this method is safe to call whenever node needs to know the latest local opinion, in current codebase it is called:
- to check that local opinion is consistent with previously applied blocks
- before casting votes. this ensures that the node always votes according to the latest changes in the opinion


dshulyak:
bors try

tortoise/tortoise.go
```diff
@@ -408,41 +407,46 @@ func (t *turtle) getFullVote(ctx context.Context, block *blockInfo) (sign, voteR
...
 
-// onLayerTerminated is expected to be called when hare terminated for a layer.
-// Internally tortoise will verify all layers before the last one if there are no gaps in
-// terminated layers.
-func (t *turtle) onLayerTerminated(ctx context.Context, lid types.LayerID) error {
-	t.logger.With().Debug("on layer terminated", lid)
+func (t *turtle) onLayer(ctx context.Context, lid types.LayerID) error {
+	t.logger.With().Debug("on layer", lid)
 	defer t.evict(ctx)
 	if err := t.updateLayer(t.logger, lid); err != nil {
```
countvonzero, line 413: this is where the reference height is calculated. isn't it still possible that onLayer() is called after the first block of the epoch is fetched?

mesh/mesh.go
```diff
@@ -368,7 +368,8 @@ func (msh *Mesh) ProcessLayer(ctx context.Context, layerID types.LayerID) error
 	logger.Info("processing layer")
 
 	// pass the layer to tortoise for processing
-	newVerified := msh.trtl.HandleIncomingLayer(ctx, layerID)
+	msh.trtl.TallyVotes(ctx, layerID)
```
countvonzero, line 371: based on the description of https://github.com/spacemeshos/go-spacemesh/issues/3575
i'd assume that we don't need this call in mesh.go anymore. in another words, tortoise should advance verified layer based on data from OnBallot(). mesh just needs to check if a new layer is verified and then apply them.

dshulyak: it would be too expensive to check if layer crosses threshold after every ballot is counted, there are two points when we need to know when threshold is crossed:
- when node wants to apply layers
- for casting votes

tortoise/tortoise.go
dshulyak: updateRefHeight for block in layer X is called only after layer X was processed

for example if tortoise receives:

on_block bid=AA lid=100. if processed is 99 update ref height is not triggered, because it may fail
on_layer lid=100. processed is set to 100, process also looks up what blocks are in the internal state and triggers update ref height for them
on_block bid=BB lid=100. processed is 100, so now failure is not expected and update ref height is triggered immediately

countvonzero (approved):

dshulyak:
bors merge

PR #3597, opened 2022-09-28
signing: add genesis id for signer and verifier

evt:
## Motivation

Closes #[3503](https://github.com/spacemeshos/go-spacemesh/issues/3503)

## TODO

## DevOps Notes


dshulyak:
Please check my comments on the original pr. this change has the same incorrect implementation. And even added genesis id to vrf which is also a mistake.

Also it really has to many changes for no reason, lets stick to original plan with using option for genesis id to minimize amount of changes. This whole thing should around 50 lines of code

dshulyak:
implemented elsewhere

Issue #3590, opened 2022-09-26
Improve tests and "Constructors" in activation package


fasmat:
## Description
The `activation` package is convoluted in terms of its testing. I see the following issues:

- Most code between tests is shared; which makes changing any of the shared code a difficult task (fixing one test breaks another, breaks another...)
- Tests are too large and test too many things at once. This can lead to unstable tests, tests that aren't actually testing what they claim to test and tests that seemingly pass, when they shouldn't, because they were executed on some leftover state from a previous test
- There is a mixture of handwritten and generated mocks. All mocks should be generated.
- "Constructors" like `NewBuilder` and `NewHandler` use too many parameters, some of which are always the same anyway (especially in tests). This can be improved by applying the [functional options pattern](https://uptrace.dev/blog/posts/go-functional-options-named-args.html)

## Affected code
Most of the code in the `activation` package.

PR #3598, opened 2022-09-28
[Merged by Bors] - Create a mock for PostSetupProvider

fasmat:
## Motivation
Part of https://github.com/spacemeshos/go-spacemesh/issues/3590. Generate a mock for PostSetupProvider

## Changes
This are the MINIMUM changes necessary to be able to generate a mock for the `PostSetupProvider` interface and move that generated mock into a package outside `activation`.

- Types used by that interface need to be moved to a new package `activation/types` and all uses of those types need to be updated
- everywhere I see `var _ SomeInterface = (*typeImplementingInterface)(nil)` to check that a type actually implements the interface
   - This is a very "Java" / "C++" way of thinking of interfaces and is driven by the fact the provider defines the interface
   - In Go this should be the other way round: the consumer defines the interface not the provider

## TODO

## DevOps Notes


countvonzero (approved):

fasmat:
bors merge

PR #3599, opened 2022-09-28
[Merged by Bors] - implement signer and verifier that prefixes message

### LINKED ISSUE #3503: signer: use genesis id as domain separator for protocol messages

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3503

- added option to prefix message before signing and verifying
- signing.ExtractPublicKey replaced ed25519.ExtractPublicKey, so that we can use prefixed verifier

more efficient version won't allocate new buffer when message is being prefixed, but that requires changes in spacemesh/ed25519 library

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -323,6 +323,8 @@ func (app *App) Initialize() (err error) {
 			return fmt.Errorf("genesis config was updated after initializing a node, if you know that update is required delete config at %s.\ndiff:\n%s", gpath, diff)
 		}
 	}
+	signing.DefaultVerifier = signing.NewEDVerifier(
+		signing.WithVerifierPrefix(app.Config.Genesis.GenesisID().Bytes()))
 
 	// tortoise wait zdist layers for hare to timeout for a layer. once hare timeout, tortoise will
 	// vote against all blocks in that layer. so it's important to make sure zdist takes longer than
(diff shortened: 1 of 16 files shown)


signing/signer.go
```diff
@@ -151,29 +168,63 @@ func (es *EdSigner) ToBuffer() []byte {
 
 // Verify verifies the provided message.
 func Verify(pubkey *PublicKey, message []byte, sign []byte) bool {
```
countvonzero: the only use of this func `Verify` in a test. rm?

cmd/node/node.go
```diff
@@ -941,7 +941,6 @@ func (app *App) LoadOrCreateEdSigner() (*signing.EdSigner, error) {
 		log.With().Info("created new identity", edSgn.PublicKey())
 		return edSgn, nil
 	}
-
 	edSgn, err := signing.NewEdSignerFromBuffer(data)
```
countvonzero: are you waiting for the genesis confg to merge before you can initialize the signer?

countvonzero (approved):

cmd/node/node.go
dshulyak: yes, i added a placeholder genesis id initially but then realized that it is better to rebase after config is merged

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #3600, opened 2022-09-28
[Merged by Bors] - implement genesis config

### LINKED ISSUE #3505: cmd/node: configuration changes for genesis id

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3505

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -309,7 +309,7 @@ func (app *App) Initialize() (err error) {
 	var existing config.GenesisConfig
 	if err := existing.LoadFromFile(gpath); err != nil {
 		if !errors.Is(err, os.ErrNotExist) {
-			return fmt.Errorf("failed to load genesis config at %s", gpath)
+			return fmt.Errorf("failed to load genesis config at %s: %w", gpath, err)
 		}
 		if err := app.Config.Genesis.Validate(); err != nil {
 			return err
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -60,7 +60,10 @@ import (
 	"github.com/spacemeshos/go-spacemesh/txs"
 )
 
-const edKeyFileName = "key.bin"
+const (
+	edKeyFileName   = "key.bin"
+	genesisFileName = "genesis.json"
+)
 
 // Logger names.
 const (
@@ -115,11 +118,11 @@ var Cmd = &cobra.Command{
 		)
 		starter := func() error {
 			if err := app.Initialize(); err != nil {
-				return fmt.Errorf("init node: %w", err)
+				return err
 			}
 			// This blocks until the context is finished or until an error is produced
 			if err := app.Start(); err != nil {
-				return fmt.Errorf("start node: %w", err)
+				return err
 			}
 
 			return nil
@@ -127,7 +130,7 @@ var Cmd = &cobra.Command{
 		err = starter()
 		app.Cleanup()
 		if err != nil {
-			log.With().Fatal("Failed to run the node. See logs for details.", log.Err(err))
+			log.With().Fatal(err.Error())
 		}
 	},
 }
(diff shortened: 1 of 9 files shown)


config/genesis.go
```diff
@@ -1,15 +1,71 @@
...
+	hh.Write([]byte(g.ExtraData))
+	rst := types.Hash20{}
+	hh.Sum(rst[:0])
+	return rst
+}
+
+// Validate GenesisConfig.
+func (g *GenesisConfig) Validate() error {
+	if len(g.ExtraData) > 255 {
+		return fmt.Errorf("extra-data %s is longer than 255 symbols", g.ExtraData)
```
countvonzero: nit: probably better to print the extra-data at the end of the sentence.

cmd/node/node.go
```diff
@@ -296,6 +299,31 @@ func (app *App) introduction() {
 
 // Initialize sets up an exit signal, logging and checks the clock, returns error if clock is not in sync.
 func (app *App) Initialize() (err error) {
+	// ensure all data folders exist
+	err = filesystem.ExistOrCreate(app.Config.DataDir())
```
countvonzero, line 303: for better usability maybe we should just go ahead to create the directory if it doesn't exist.

config/genesis.go
```diff
@@ -1,15 +1,71 @@
...
+	return json.Unmarshal(buf, g)
+}
+
+// WriteToFile writes config content to file.
+func (g *GenesisConfig) WriteToFile(filename string) error {
+	buf, err := json.Marshal(g)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(filename, buf, 0600)
```
countvonzero: 0600 seems strict.... why not 0644?

cmd/node/node_test.go
```diff
@@ -914,6 +914,90 @@ func TestConfig_GenesisAccounts(t *testing.T) {
...
+		app.Config = &config.Config{
+			BaseConfig: config.BaseConfig{
+				DataDirParent: t.TempDir(),
+			},
+			Genesis: &config.GenesisConfig{
+				ExtraData:   "test",
+				GenesisTime: time.Now().Format(time.RFC3339),
+			},
+		}
+		require.ErrorContains(t, app.Initialize(), "tortoise")
```
countvonzero: maybe you can start from DefaultTestConfig() and modify the genesis config and data dir so you don't hit the tortoise param error. see `TestInitialize_BadTortoiseParams`

cmd/node/node_test.go
```diff
@@ -914,6 +914,90 @@ func TestConfig_GenesisAccounts(t *testing.T) {
...
+				DataDirParent: t.TempDir(),
+			},
+			Genesis: &config.GenesisConfig{
+				ExtraData:   "test",
+				GenesisTime: time.Now().Format(time.RFC3339),
+			},
+		}
+
+		require.ErrorContains(t, app.Initialize(), "tortoise")
+		require.ErrorContains(t, app.Initialize(), "tortoise")
```
countvonzero: double-whammy :P

countvonzero (approved):

cmd/node/node_test.go
dshulyak: thanks, slipped from my mind

cmd/node/node.go
dshulyak: but this function is creating a directory exactly in such case. did i miss your point?

countvonzero: lol. my bad. 

the error msg is "ensure folders exist" which misled me and prolly the operator. but i guess it'll print filesystem error to give the operator more hint.

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #3601, opened 2022-09-29
[Merged by Bors] - block certifier: fix certifier concurrent map access

countvonzero:
## Motivation

race condition, added test that failed with -race -count 5 and pass after the fix


dshulyak (approved):

countvonzero:
bors merge

PR #3603, opened 2022-09-29
[Merged by Bors] - Support multiple poet services

### LINKED ISSUE #3460: atx: register with multiple poet servers
### Motivation
PoET servers are a potential single-point-of-failure. There are many scenarios where PoETs can be attacked and possibly fail. In addition, a single widely used PoET service provides its operators centralized control of the network. If a widely used PoET service is malicious or attacked it can have detrimental effects on the network.

### Mitigation
Allow smeshers to register for multiple PoET services and pick one proof to use in their ATX.

### Design
https://community.spacemesh.io/t/redundant-poet-registration/310/1

poszu:
Closes #3460 

## Changes
Added support for multiple PoET services. The design is described in detail in [Redundant PoET Registration](https://community.spacemesh.io/t/redundant-poet-registration/310). 

## Test Plan
unittests, systests

## TODO

- [x] Proper deadline calculation
- [x] Extend systests
- [x] Update Node's GRPC API to support configuring multiple PoET services

## DevOps Notes


activation/nipost.go
```diff
@@ -223,3 +195,133 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash3
...
+	}()
+	g.Wait()
+	close(poetRequestsChannel)
+	<-done
+	return poetRequests
+}
+
+// awaitPoetProof concurrently waits for proofs from all PoET services that a challenge was submitted to
+// until it gets proofs from all PoETs or a deadline channel is closed.
+// The best proof will be provided through the returned channel (if any). Otherwise, the channel is closed.
```
poszu:
```suggestion
// awaitPoetProof concurrently waits for proofs from all PoET services that a challenge was submitted to
// until it gets proofs from all PoETs or the `deadline` channel is closed.
// The best proof is returned or `nil` if proof was not received at all within the deadline.
```

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

activation/activation.go
```diff
@@ -368,6 +371,23 @@ func (b *Builder) waitForFirstATX(ctx context.Context) bool {
...
+	poetClients := func() []PoetProvingServiceClient {
+		for {
+			select {
+			case clients := <-b.pendingPoetClients:
+				return clients
+			default:
+				return nil
+			}
+		}
+	}()
```
poszu: It should pick up the last element available in the channel (the last update of poet clients). Currently, it picks up the first one (the oldest update).

cmd/root.go
```diff
@@ -38,8 +38,8 @@ func AddCommands(cmd *cobra.Command) {
 		config.OracleServer, "The oracle server url. (temporary) ")
 	cmd.PersistentFlags().IntVar(&config.OracleServerWorldID, "oracle_server_worldid",
 		config.OracleServerWorldID, "The worldid to use with the oracle server (temporary) ")
-	cmd.PersistentFlags().StringVar(&config.PoETServer, "poet-server",
-		config.PoETServer, "The poet server url. (temporary) ")
+	cmd.PersistentFlags().StringArrayVar(&config.PoETServers, "poet-server",
+		config.PoETServers, "The poet server url. (temporary) Can be passed multiple times")
```
fasmat, line 42:
```suggestion
	cmd.PersistentFlags().StringSliceVar(&config.PoETServers, "poet-server",
		config.PoETServers, "Poet server url(s) (temporary) Can be passed multiple times")
```

cmd/node/node.go
```diff
@@ -171,7 +171,7 @@ type TickProvider interface {
 	GetCurrentLayer() types.LayerID
 	StartNotifying()
 	GetGenesisTime() time.Time
-	LayerToTime(types.LayerID) time.Time
+	timesync.LayerConverter
```
fasmat, line 173: This seems to me like it doesn't fit here. `TickProvider` looks like a collection of all kinds of methods, not necessarily related to each other:

- `GetCurrentLayer()` returns the current layer of the network - while related to time this shouldn't be part of a "TimeService" - but this isn't part of the PR so lets refrain from touching this for now
- `timesync.LayerConverter` is a collection of arithmetic functions that do conversions, imho they shouldn't be part of `TickProvider`.

From an implementation standpoint both are actually different objects that should be treated such and not embedded and exposed in one another.

api/grpcserver/node_service.go
```diff
@@ -128,16 +128,16 @@ func (s NodeService) Shutdown(context.Context, *pb.ShutdownRequest) (*pb.Shutdow
...
 	if err == nil {
-		return &pb.UpdatePoetServerResponse{
+		return &pb.UpdatePoetServersResponse{
 			Status: &rpcstatus.Status{Code: int32(code.Code_OK)},
 		}, nil
 	}
 	switch {
 	case errors.Is(err, activation.ErrPoetServiceUnstable):
-		return nil, status.Errorf(codes.Unavailable, "can't reach server at %s. retry later", req.Url)
+		return nil, status.Errorf(codes.Unavailable, "can't reach poet service (%v). retry later", err)
```
fasmat, line 140:
```suggestion
		return nil, status.Errorf(codes.Unavailable, "can't reach poet service (retry later): %w", err)
```

`%w` wraps errors so they can be used with `errors.As` and `errors.Is`. `%v` just uses their string representation.

api/grpcserver/grpcserver_test.go
```diff
@@ -670,17 +670,17 @@ func TestNodeService(t *testing.T) {
...
 		{"UpdatePoetServerUnavailable", func(t *testing.T) {
 			logtest.SetupGlobal(t)
 			atxapi.UpdatePoETErr = activation.ErrPoetServiceUnstable
-			url := "test"
-			res, err := c.UpdatePoetServer(context.TODO(), &pb.UpdatePoetServerRequest{Url: url})
+			urls := []string{"test"}
+			res, err := c.UpdatePoetServers(context.TODO(), &pb.UpdatePoetServersRequest{Urls: urls})
 			require.Nil(t, res)
-			require.ErrorIs(t, err, status.Errorf(codes.Unavailable, "can't reach server at %s. retry later", url))
+			require.ErrorIs(t, err, status.Errorf(codes.Unavailable, "can't reach poet service (%v). retry later", atxapi.UpdatePoETErr))
```
fasmat, line 686:
```suggestion
			require.ErrorIs(t, err, status.Errorf(codes.Unavailable, "can't reach poet service (retry later): %w", atxapi.UpdatePoETErr))
```

`%w` wraps errors so they can be used with `errors.As` and `errors.Is`. `%v` just uses their string representation.

systest/cluster/nodes.go
```diff
@@ -71,7 +71,7 @@ func deployPoet(ctx *testcontext.Context, name string, gateways []string, flags
 		args = append(args, "--gateway="+gw)
 	}
 	ctx.Log.Debugw("deploying poet pod", "args", args, "image", ctx.PoetImage)
-	labels := nodeLabels("poet")
+	labels := nodeLabels(name)
```
fasmat: Why give multiple deployments of PoET different labels? Shouldn't they all be labeled `"poet"`?

systest/cluster/nodes.go
```diff
@@ -188,12 +188,15 @@ func deployNodes(ctx *testcontext.Context, name string, from, to int, flags []De
 	)
 	for i := from; i < to; i++ {
 		i := i
-		finalFlags := make([]DeploymentFlag, len(flags)+1)
+		finalFlags := make([]DeploymentFlag, len(flags))
```
fasmat: If you want to avoid re-allocating the slice multiple times do this:

```suggestion
		finalFlags := make([]DeploymentFlag, len(flags), len(flags)+ctx.PoetSize)
```

activation/nipost_test.go
```diff
@@ -106,18 +111,42 @@ func newPoetServiceMock(tb testing.TB) (*mocks.MockPoetProvingServiceClient, *go
...
+	return ok
+}
+
+func (p *poetDbMock) didUnsubscribe(poetID []byte) bool {
+	_, ok := p.unsubscribed[hash.Sum(poetID)]
+	return ok
+}
+
+// A compile time check to ensure that poetDbMock fully implements poetDbAPI.
+var _ poetDbAPI = (*poetDbMock)(nil)
```
fasmat: This is not needed, if the mock doesn't implement the interface you will get a compilation error even if this check doesn't exist.

activation/activation_test.go
```diff
@@ -116,25 +116,31 @@ func (ms *MockSigning) Sign(m []byte) []byte {
...
 	}
-	return newNIPostWithChallenge(challenge, np.poetRef), nil
+	return newNIPostWithChallenge(challenge, np.poetRef), 0, nil
 }
 
 type NIPostErrBuilderMock struct{}
 
-func (np *NIPostErrBuilderMock) updatePoETProver(PoetProvingServiceClient) {}
+// A compile time check to ensure that `NIPostErrBuilderMock` fully implements the `nipostBuilder` interface.
+var _ nipostBuilder = (*NIPostErrBuilderMock)(nil)
```
fasmat: This check is not needed, if the mock wouldn't implement the interface the compiler already tells you so; no need for an extra check here.

activation/activation.go
```diff
@@ -88,8 +87,8 @@ type Config struct {
 // it is responsible for initializing post, receiving poet proof and orchestrating nipst. after which it will
 // calculate total weight and providing relevant view as proof.
 type Builder struct {
-	pendingPoetClient atomic.UnsafePointer
-	started           atomic.Bool
+	pendingPoetClients chan []PoetProvingServiceClient
```
fasmat: The code becomes easier to understand if you use a `sync.Mutex` to protect the field of the struct instead of a channel.

```suggestion
	pendingPoetClientsMux sync.Mutex
	pendingPoetClients    []PoetProvingServiceClient
```

activation/activation.go
fasmat:
```suggestion
	b.pendingPoetClientsMux.Lock()
	poetClients := b.pendingPoetClients
	b.pendingPoetClients = nil
	b.pendingPoetClientsMux.Unlock()
```

activation/activation.go
poszu: The channel is not initialized anywhere.
TODO:
- [ ] fix
- [ ] write unit test

activation/activation.go
```diff
@@ -452,17 +467,22 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) error {
...
+	b.log.With().Debug("request to update poet services", log.Strings("endpoints", endpoints))
+
+	clients := make([]PoetProvingServiceClient, len(endpoints))
+	for _, endpoint := range endpoints {
+		client := b.poetClientInitializer(endpoint)
+		// TODO(dshulyak) not enough information to verify that PoetServiceID matches with an expected one.
+		// Maybe it should be provided during update.
+		sid, err := client.PoetServiceID(ctx)
+		if err != nil {
+			return fmt.Errorf("%w: failed to query Poet '%s' for ID (%v)", ErrPoetServiceUnstable, endpoint, err)
```
fasmat: It's weird to see two errors wrapped simultaneously: `ErrPoetServiceUnstable` and `err`.

The only way that I see around that would be to make `ErrPoetserviceUnstable` a proper type:

```golang
type ErrPoetServiceUnstable struct {
     Err error
}

func (e *ErrPoetServiceUnstable) Error() string { return fmt.Sprintf("builder: poet service is unstable: %v", e.Err) }

func (e *ErrPoetServiceUnstable) Unwrap() error { return e.Err }
```

and then use it here like this:

```suggestion
			return &ErrPotServiceUnstable{Err: fmt.Errorf("failed to query Poet '%s': %w", endpoint, err) }
```

activation/activation.go
dshulyak: what was the problem with atomic? i think it simpler that way

activation/activation.go
dshulyak: with atomic (or mutex locked variable also works) it will work exactly this way

activation/activation.go
```diff
@@ -452,17 +467,22 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) error {
...
+		}
+		b.log.With().Debug("preparing to update poet service", log.String("poet_id", util.Bytes2Hex(sid)))
 	}
-	b.pendingPoetClient.Store(unsafe.Pointer(&client))
-	b.log.With().Debug("preparing to update poet service", log.String("poet_id", util.Bytes2Hex(sid)))
+
+	b.pendingPoetClients <- clients
```
fasmat:
```suggestion
	b.pendingPoetClientsMux.Lock()
	b.pendingPoetClients = clients
	b.pendingPoetClientsMux.Unlock()
```

activation/activation.go
```diff
@@ -591,23 +611,26 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
 
 	// the following method waits for a PoET proof, which should take ~1 epoch
-	expireLayer := (pubEpoch + 2).FirstLayer()
-	atxExpired := b.layerClock.AwaitLayer(expireLayer) // this fires when the target epoch is over
+	pubEpoch := b.challenge.PublishEpoch()
+	// Add a 10% margin to the time it takes to generate a PoST proof
+	postDurationWithMargin := time.Duration(float64(b.lastPostGenDuration.Nanoseconds()) * 1.1)
+	postDurationInLayers := b.layerClock.DurationToLayers(postDurationWithMargin)
+	deadlineLayer := b.challenge.PubLayerID.Add(types.GetLayersPerEpoch()).Sub(postDurationInLayers)
+	deadline := b.layerClock.AwaitLayer(deadlineLayer)
```
fasmat: I feel this introduces a possible issue with deadlines. Instead of hard setting a certain layer for when expiration happens now a node can intentionally delay its PoST generation to set a longer deadline?

activation/activation.go
```diff
@@ -591,23 +611,26 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
 
-	b.log.With().Info("building NIPost", log.Stringer("pub_epoch", pubEpoch), log.Stringer("expire_layer", expireLayer))
+	b.log.With().Info("building NIPost", log.Stringer("pub_epoch", pubEpoch), log.Stringer("deadline_layer", deadlineLayer), log.Uint32("last PoST duration", postDurationInLayers))
 
-	nipost, err := b.nipostBuilder.BuildNIPost(ctx, hash, atxExpired)
+	nipost, postDuration, err := b.nipostBuilder.BuildNIPost(ctx, hash, deadline)
```
fasmat: Instead of returning postDuration here I'd prefer it being a field of `nipostBuilder` that is then read from to set `b.lastPostGenDuration` below or even replace it.

activation/interface.go
```diff
@@ -24,4 +24,6 @@ type layerClock interface {
 	AwaitLayer(layerID types.LayerID) chan struct{}
 	GetCurrentLayer() types.LayerID
 	LayerToTime(types.LayerID) time.Time
+	// Convert Duration to number of layers
+	DurationToLayers(time.Duration) uint32
```
fasmat: I don't think a Clock should do Layer arithmetic, see also my earlier comment

activation/nipost.go
```diff
@@ -222,5 +195,148 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash3
...
+		})
+	}
+	poetRequests := make([]PoetRequest, 0)
+	done := make(chan struct{})
+	go func() {
+		for request := range poetRequestsChannel {
+			poetRequests = append(poetRequests, request)
+		}
+		close(done)
+	}()
```
dshulyak: i think this is cleaner
```
requests := make(chan Request, N)
eg.Go // spawn N goroutines
eg.Wait() // because every goroutine in eg.Go wont block on send to requests
close(requests)
... loop over closed channel ...
```

activation/nipost.go
```diff
@@ -81,12 +86,15 @@ type NIPostBuilder struct {
...
 	postSetupProvider PostSetupProvider
-	poetProver        PoetProvingServiceClient
+	poetProvers       []PoetProvingServiceClient
 	poetDB            poetDbAPI
 	state             *BuilderState
 	log               log.Log
 }
 
+// A compile time check to ensure that `NIPostBuilder` fully implements the `nipostBuilder` interface.
+var _ nipostBuilder = (*NIPostBuilder)(nil)
```
fasmat: Not required. `NIPostBuilder` doesn't define the `nipostBuilder` interface, but its user `activation.Builder` does.

Even without this line the compiler would still complain that `NIPostBuilder` doesn't implement the `nipostBuilder` interface if it wouldn't.

activation/nipost.go
```diff
@@ -115,100 +123,65 @@ func NewNIPostBuilder(
...
+	nb.poetProvers = poetProvers
+	nb.log.With().Info("updated poet proof service clients", log.Int("count", len(nb.poetProvers)))
 }
 
-// BuildNIPost uses the given challenge to build a NIPost. "atxExpired" and "stop" are channels for early termination of
+// BuildNIPost uses the given challenge to build a NIPost. "deadline" is a channel for early termination of
 // the building process. The process can take considerable time, because it includes waiting for the poet service to
 // publish a proof - a process that takes about an epoch.
-func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash32, atxExpired chan struct{}) (*types.NIPost, error) {
+func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash32, deadline chan struct{}) (*types.NIPost, time.Duration, error) {
```
fasmat: I would instead of returning the duration to create a post here, make it a field that can be read from by the user of the `NIPostBuilder`.

fasmat (changes requested):
I think some parts can be improved (see my comments).

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+	nipost, _, err = nb.BuildNIPost(context.TODO(), &hash2, nil)
 	assert.Equal(4, postProvider.called)
 
 	assert.NotNil(nipost)
 	assert.NoError(err)
 }
 
+func createMockPoetService(t *testing.T, id []byte) PoetProvingServiceClient {
+	t.Helper()
+	poet := mocks.NewMockPoetProvingServiceClient(gomock.NewController(t))
```
fasmat, line 350: By using this helper method multiple `gomock` Controllers are instantiated within the same test.

In my opinion this helper can be inlined. It only saves 3 lines of code per call and is called 4 times, but hides test details outside of the test.

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+
+	challenge := types.BytesToHash([]byte("challenge0"))
+	g, ctx := errgroup.WithContext(context.TODO())
+	resultChan := make(chan struct {
+		*types.NIPost
+		error
+	})
+	deadlineChan := make(chan struct{}, 1)
+
+	g.Go(func() error {
```
fasmat, line 451: This error group is never awaited?

activation/nipost.go
```diff
@@ -222,5 +195,148 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash3
...
+		for request := range poetRequestsChannel {
+			poetRequests = append(poetRequests, request)
+		}
+		close(done)
+	}()
+	g.Wait()
+	close(poetRequestsChannel)
+	<-done
+	return poetRequests
+}
```
fasmat:
```suggestion
func (nb *NIPostBuilder) submitPoetChallenges(ctx context.Context, challenge *types.Hash32) []PoetRequest {
	poetRequestsChannel := make(chan PoetRequest, len(nb.poetProvers))
	g, ctx := errgroup.WithContext(ctx)
	for _, poetProver := range nb.poetProvers {
		poet := poetProver
		g.Go(func() error {
			if poetRequest, err := submitPoetChallenge(ctx, nb.log, poet, challenge); err == nil {
				poetRequestsChannel <- *poetRequest
			} else {
				nb.log.With().Warning("failed to submit challenge to PoET", log.Err(err))
			}
			return nil
		})
	}
	g.Wait()
	close(poetRequestsChannel)

	poetRequests := make([]PoetRequest, 0, len(nb.poetProvers))
	for request := range poetRequestsChannel {
		poetRequests = append(poetRequests, request)
	}
	return poetRequests
}
```

I completely forgot one can still read elements from a channel after it is closed (as long as it is a buffered channel).

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+
+	challenge := types.BytesToHash([]byte("challenge"))
+	g.Go(func() error {
+		nipost, _, err := nb.BuildNIPost(ctx, &challenge, deadlineChan)
+		resultChan <- struct {
+			*types.NIPost
+			error
+		}{nipost, err}
+		return nil
+	})
```
fasmat: This could be simplified to:

```suggestion
	g, ctx := errgroup.WithContext(context.TODO())
	resultChan := make(chan *types.NIPost, 1)
	deadlineChan := make(chan struct{})
	challenge := types.BytesToHash([]byte("challenge"))
	g.Go(func() error {
		nipost, _, err := nb.BuildNIPost(ctx, &challenge, deadlineChan)
		assert.NoError(err)
		resultChan <- nipost
		return nil
	})
```

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+		},
+	}
+	ref, err := proofMsg.Ref()
+	assert.NoError(err)
+	poetDb.StoreProof(ref, &proofMsg)
+	time.Sleep(time.Millisecond * 10)
+
+	// Signal that the time is out
+	close(deadlineChan)
+
```
fasmat, line 391: I would add a `g.Wait()` here just to make sure that the `errGroup` finished before continuing.

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+	// Verify
+	result := <-resultChan
+	assert.NoError(result.error)
+	assert.Equal(*result.NIPost.Challenge, challenge)
+	proof, err := poetDb.GetProof(result.NIPost.PostMetadata.Challenge)
```
fasmat: with the change above this simplifies to:

```suggestion
	result := <-resultChan
	assert.Equal(challenge, *result.Challenge)
	proof, err := poetDb.GetProof(result.PostMetadata.Challenge)
```

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+		PoetServiceID: []byte("poet1"),
+		PoetProof: types.PoetProof{
+			Members:   [][]byte{challenge[:]},
+			LeafCount: 1,
+		},
+	}
+	ref, err = proofMsg.Ref()
+	assert.NoError(err)
+	poetDb.StoreProof(ref, &proofMsg)
+
```
fasmat, line 444: I would again add a `g.Wait()` here to make sure the `errGroup` actually finished.

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+	deadlineChan := make(chan struct{}, 1)
+
+	g.Go(func() error {
+		nipost, _, err := nb.BuildNIPost(ctx, &challenge, deadlineChan)
+		resultChan <- struct {
+			*types.NIPost
+			error
+		}{nipost, err}
+		return nil
+	})
```
fasmat: Similar to above:

```suggestion
	challenge := types.BytesToHash([]byte("challenge0"))
	g, ctx := errgroup.WithContext(context.TODO())
	resultChan := make(chan *types.NIPost, 1)
	deadlineChan := make(chan struct{}, 1)

	g.Go(func() error {
		nipost, _, err := nb.BuildNIPost(ctx, &challenge, deadlineChan)
		assert.NoError(err)
		resultChan <- nipost
		return nil
	})
```

activation/nipost_test.go
```diff
@@ -281,73 +310,194 @@ func TestNIPostBuilder_BuildNIPost(t *testing.T) {
...
+	}
+	ref, err = proofMsg.Ref()
+	assert.NoError(err)
+	poetDb.StoreProof(ref, &proofMsg)
+
+	// Verify
+	result := <-resultChan
+	assert.NoError(result.error)
+	assert.Equal(*result.NIPost.Challenge, challenge)
+	proof, err := poetDb.GetProof(result.NIPost.PostMetadata.Challenge)
```
fasmat: This can be simplified to:

```suggestion
	result := <-resultChan
	assert.Equal(challenge, *result.Challenge)
	proof, err := poetDb.GetProof(result.PostMetadata.Challenge)
```

cmd/node/node.go
poszu: I agree that this interface is abused. However, `LayerConverter` is indirectly part of this interface already (via its `LayerToTime` method). I only substituted it with the whole interface to also include the `DurationToLayers` method.

This interface needs ground refactoring which is out of the scope of this PR.

cmd/root.go
poszu: `StringSliceVar` works differently and not how I wanted it to. I want users to pass each PoET separately (like `--poet-server=foo --poet-server=bar`).

fasmat: Both `StringSliceVar` and `StringArrayVar` will parse your example in the same way:`--poet-server=foo --poet-server=bar` will result in `{"foo", "bar"}`

The difference is in the following: `--poet-server=foo,bar`. With `StringArrayVar` this will be `{"foo,bar"}` with `StringSliceVar` this will be `{"foo", "bar"}`. See also https://github.com/spf13/cobra/issues/661#issuecomment-377684634 and https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#StringSlice

cmd/node/node.go
fasmat: Then please create a ticket for it. `TickProvider` and `LayerConverter` should be their own independent interfaces with their own implementations (which is already the case). Just the users of `TickProvider` should use `LayerConverter` directly instead of via `TickConverter`.

activation/activation.go
poszu: Yeah, initially I thought that using a channel will simplify things, but it actually causes a few problems I had not anticipated. I will revert back to the atomic pointer.

cmd/root.go
poszu: Yes, I know. I don't want to allow passing `--poet-server=foo,bar`.

api/grpcserver/node_service.go
poszu: status.Errorf does not support error-wrapping directive %w.

systest/cluster/nodes.go
poszu: 💡 These are not Deployments, but a Pod and a Service.

A node is given a list of the hostnames of Kubernetes Services that expose PoET Pods. There are N Services and N PoET pods. We want each Service to expose a different PoET Pod. A Service selects Pod it exposes by using a `label` selector. To be able to select a different PoET Pod in each Service we must use a different label for every Pod.

Previously, every Service exposed all available PoETs because they shared the label `app=poet`. Requests to a Service are load balanced across the matched Pods. It led to a situation when a Node thought it was connecting to `poet-0`, but it was actually being connected randomly to one of the available poets.

activation/activation.go
poszu: I don't understand what you mean. Could you elaborate?

The longer the PoST takes, the earlier the deadline is set in order to accommodate the time required to run PoST.

activation/nipost.go
poszu: It also serves as a note that `NIPostBuilder` is supposed to implement `nipostBuilder`. This note sits next to the definition of the struct. IMO it is useful.

activation/nipost_test.go
poszu: > By using this helper method multiple gomock Controllers are instantiated within the same test.
Yes, but it doesn't hurt.

IMO, it makes the test code more readable 🤷 .

activation/nipost_test.go
poszu: Well, yes, but at the same time, it moves the assert from "Verify" block into the goroutine mixing the code a bit. Anyway, a lack of tuples in Go 😠 makes my approach with the anonymous struct so ugly that I will refactor to your way.

activation/nipost_test.go
poszu: There's no need. It's synchronized via the `resultChan`.

activation/activation.go
dshulyak: was is described by someone this way? i don't understand why is this so complex, and why it measures some timings

if we are using a certain instance of poet we know when that poet is expected to terminate. proof should be available almost immediately when poet terminates. so the deadline when we can give up on waiting for that proof is a termination time + any reasonable timeout (5s for example, but should be lower than the cycle gap)

activation/nipost_test.go
poszu: Same, it's not required.

activation/activation.go
poszu: Refactored

activation/activation.go
poszu: Refactored

activation/nipost.go
poszu: It is simpler but it has the drawback that the consumer starts processing after all producers are finished. In this case, it doesn't really matter as it just collects the results and I don't expect too many PoETs so I will refactor.

However, I'd rather keep the current implementation of `awaitPoetProof`, which uses the same approach as here.

activation/activation.go
poszu: @dshulyak yes, it is described in https://community.spacemesh.io/t/redundant-poet-registration/310/11.

The deadline reflects the point up to which it is "safe" to wait for PoETs proofs. "Safe" means that there will still be enough time to build PoST using PoET proof as the challenge.

 We submit challenges to many PoETs. We want to wait for one of the following to happen:
- we got proofs from all PoETs within the deadline time✅ 
- we got proofs from **at least** on PoET and the deadline expired ⚠️ 
- deadline expired (no proofs) ❌ 

In the ✅ and ⚠️, we can proceed to build PoST. 
However, in ⚠️ we must not wait till the old `atxExpired` time because we will not be able to build ATX on time when at least one PoET service is unavailable.
In ❌ we must abort.

In order to guarantee that, when the deadline expires, there is still enough time to build PoST (and complete the NiPoST), we must shorten the deadline by the duration that PoST generation is expected to take.

cmd/node/node.go
poszu: I created a new tech-debt ticket to address it: https://github.com/spacemeshos/go-spacemesh/issues/3621

activation/activation.go
poszu: I didn't want to keep this information in the state of two structs (`Builder` and `NIPostBuilder`). It needs to be kept in `Builder` because it is initialized with the duration of the first PoST here: https://github.com/spacemeshos/go-spacemesh/blob/2e230ca4d4081e68d517edc538270a5051d3ffa6/activation/activation.go#L320

Therefore, I decided to just return it from `BuildNIPost()` and keep it in a single place (in the state of `Builder`).

activation/activation.go
dshulyak: i think you are overcomplicating things, as well as the description in that thread

there is a poet cycle-gap (delta) which is equal to 5m (for example), this 5m should include and it is important that it needs to be known in advance:
- proof delivery time
- post execution time
- post delivery time

since all this values have to be known in advance - they should be set to the max reasonable value. 

what you are doing by introducing more complexity is dynamically allow to adjust post execution time, in order to increase or decrease proof/post delivery time. it makes literally no difference because whole cycle-gap won't change because of that

activation/activation.go
poszu: They are not both wrapped. `ErrPoetServiceUnstable` is wrapped and `err` provides more details. I agree it became awkward. I will think about this. Thx

activation/activation.go
poszu: I'm not overcomplicating things but following the design.

Do you mean the deadline should be set to the start of the next epoch minus the cycle gap?

activation/nipost_test.go
fasmat: Still; if you don't plan `g.Wait()` you don't need the `errgroup` to begin with. You can just call the function directly:

```golang
go func() {
	nipost, _, err := nb.BuildNIPost(ctx, &challenge, deadlineChan)
	resultChan <- struct {
		*types.NIPost
		error
	}{nipost, err}
}
```

Instead you can return the `err` in the function body and await it with `g.Wait()` (and additionally `errgroup` signals via `context` to the goroutines that are still running that they are not needed any more and can exit).

activation/activation.go
fasmat: This line: https://github.com/spacemeshos/go-spacemesh/blob/2e230ca4d4081e68d517edc538270a5051d3ffa6/activation/activation.go#L316

is expected to be called once during the whole lifetime of the node, when creating / publishing its very first ATX. While the call to `BuildNIPost` happens during every epoch. So unless the node has never published an ATX before, this won't be called and `b.lastPostGenDuration` is 0 and later set to the return value of `BuildNIPost()`

activation/nipost.go
fasmat: The compiler reminds you of that 😉 if one of the interface `nipostBuilder` or the type `NIPostBuilder` changes, the compiler will tell you that even without this line of code. The need for a clarifying comment is IMO already an indication this might better be removed again.

activation/nipost_test.go
fasmat: Then the `errGroup` is not required 😉

activation/nipost_test.go
poszu: Fair enough

activation/activation.go
dshulyak: sorry you are right, lets do what spec suggests

utils/guard/guard.go
```diff
@@ -0,0 +1,22 @@
+package guard
+
+import "sync"
+
+// Guard is a thread-safe wrapper around T.
+// It guarantees a thread-safe access to the underlying value T.
```
countvonzero: can you also comment on the recommended use of this struct?
if a struct has 10 members that require thread-safety, do you recommend using this for every member?

activation/nipost.go
poszu: Removed

activation/activation.go
poszu: OK 👍 It can still be changed later if a better method is found 😉

utils/guard/guard.go
poszu: It depends on the particular situation. Generally, fields that must be kept "in sync" must be protected with a single lock to avoid an invalid intermediate state. Otherwise, each field could be put in a separate `Guard`.

activation/activation.go
poszu: Yes, it's executed once to calculate the first PoST. However, I still want to use this information to predict the next PoST duration.

utils/guard/guard.go
countvonzero: right. i am a little wary that future devs will use guard for fields that are always read/updated together, hence the request to add comments for guidance.

countvonzero: also, what is your consideration for not choosing atomic.Value here?
for me, it seems a more natural choice if the only purpose is a thread-safe update for a single variable (and not semaphoring a sequence of events).

fasmat: In case you want type-safety there is also https://pkg.go.dev/sync/atomic#Pointer that allows atomic access to a pointer of generic type.

poszu: > also, what is your consideration for not choosing atomic.Value here? for me, it seems a more natural choice if the only purpose is a thread-safe update for a single variable (and not semaphoring a sequence of events).

The reason for choosing this `Guard` over the previous atomic pointer is simplicity and safety. It doesn't require unsafe and pointer casting.

The reason for choosing `Guard` over just sticking a mutex next to the guarded variable into a struct is readability and safety (at the cost of the simplicity of use):
- ➕  readability: it's clear which variable is guarded. When a mutex is just stuck in a struct, it is not clear what it is supposed to protect
- ➕ safety: it's impossible to access the guarded variable without a lock
- ➖ lost simplicity: it might not be obvious/easy to use this `Apply` API

@fasmat I didn't know that generic pointer, thx! It removes my main concerns regarding typecasting and using unsafe. I'm OK with using it for the pending poet clients.

countvonzero: > I'm OK with using it for the pending poet clients.

please do. thanks. this solution is perfect. thank you @fasmat 

my rationale for singling this out in this PR (i have not reviewed the rest)
- the Apply() with func ptr smells over-engineered for simply providing thread-safety to a variable.
- it's all too easy to start using Guard for every variable for a new programmer, which will cause a lot of unnecessary locking ops.
- the lib also ignored the variable's use characteristic (read/write frequency)

either of the following approach is good for me:
- use a generic library Guard but with solutions from atomic pkg and a simple getter/setter
- use a sync.Mutex next to the variable and use simple getter/setter

poszu: @fasmat Rewrote to the generic pointer.

poszu:
@fasmat Is there anything left that you request to be changed?

activation/nipost.go
```diff
@@ -86,100 +88,65 @@ func NewNIPostBuilder(
...
+	nb.poetProvers = poetProvers
+	nb.log.With().Info("updated poet proof service clients", log.Int("count", len(nb.poetProvers)))
 }
 
-// BuildNIPost uses the given challenge to build a NIPost. "atxExpired" and "stop" are channels for early termination of
+// BuildNIPost uses the given challenge to build a NIPost. "deadline" is a channel for early termination of
 // the building process. The process can take considerable time, because it includes waiting for the poet service to
 // publish a proof - a process that takes about an epoch.
-func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash32, atxExpired chan struct{}) (*types.NIPost, error) {
+func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash32, deadline chan struct{}) (*types.NIPost, time.Duration, error) {
```
fasmat: The name `deadline` is confusing, since the deadline is set by `ctx`. Why not leave it as `atxExpired` it signales that the ATX that the NIPost is built for expired.

activation/activation.go
```diff
@@ -562,23 +562,26 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
 
 	// the following method waits for a PoET proof, which should take ~1 epoch
-	expireLayer := (pubEpoch + 2).FirstLayer()
-	atxExpired := b.layerClock.AwaitLayer(expireLayer) // this fires when the target epoch is over
+	pubEpoch := b.challenge.PublishEpoch()
+	// Add a 10% margin to the time it takes to generate a PoST proof
+	postDurationWithMargin := time.Duration(float64(b.lastPostGenDuration.Nanoseconds()) * 1.1)
+	postDurationInLayers := b.layerClock.DurationToLayers(postDurationWithMargin)
+	deadlineLayer := b.challenge.PubLayerID.Add(types.GetLayersPerEpoch()).Sub(postDurationInLayers)
+	deadline := b.layerClock.AwaitLayer(deadlineLayer)
```
fasmat: imo `AwaitLayer` should take a context and return a new context that is canceled when the layer is reached.

At the moment there are two signals that every `select` has to wait for: `<-ctx.Done()` and `<-deadlineChan`. Both of which serve the same purpose -> sending a signal to cancel when the result of `BuildNIPost` isn't needed any more.

I think with some minor changes to `layerClock` this could return a context that can then be used by `BuildNIPost` without the need of a second channel.

activation/nipost.go
poszu: I can change the name. However, `atxExpired` is no longer a valid name. It no longer means that `atx` expired. It now describes a point in time when we should stop waiting for the remaining PoET proofs and use what we collected up to this point. I can call it `poetProofDeadline` or similar. What do you propose?

activation/activation.go
fasmat: Discussion continued in https://github.com/spacemeshos/go-spacemesh/pull/3633/

activation/nipost.go
fasmat: The signal that is sent by closing this channel is that a certain layer has been reached if I'm not mistaken? Maybe `layerReachedChan`?

fasmat (approved):

activation/nipost.go
poszu: Yes, but that's meaningless for the `BuildNIPost()` and is an implementation detail - it can change in the future to something else. From the point of view of `BuildNIPost(),` it is a signal meaning "stop waiting for more poet proofs".

countvonzero:
@poszu please merge develop (with the latest activation change) first. more efficient to review after that. thanks

activation/activation.go
```diff
@@ -122,6 +112,7 @@ type Builder struct {
 	poetCfg               PoetConfig
 	poetRetryInterval     time.Duration
 	poetClientInitializer PoETClientInitializer
+	lastPostGenDuration   time.Duration
```
countvonzero: add comment on what this field is for

activation/activation.go
```diff
@@ -469,17 +464,22 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) error {
...
+		sid, err := client.PoetServiceID(ctx)
+		if err != nil {
+			return &PoetSvcUnstableError{source: fmt.Errorf("failed to query Poet '%s' for ID (%w)", endpoint, err)}
+		}
+		b.log.With().Debug("preparing to update poet service", log.String("poet_id", util.Bytes2Hex(sid)))
 	}
-	b.pendingPoetClient.Store(unsafe.Pointer(&client))
-	b.log.With().Debug("preparing to update poet service", log.String("poet_id", util.Bytes2Hex(sid)))
+
+	b.pendingPoetClients.Store(&clients)
```
countvonzero, line 498: i don't see code populating `clients`. maybe add a test for this?

activation/activation.go
```diff
@@ -618,27 +618,31 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
 
 	// the following method waits for a PoET proof, which should take ~1 epoch
-	expireLayer := (pubEpoch + 2).FirstLayer()
-	atxExpired := b.layerClock.AwaitLayer(expireLayer) // this fires when the target epoch is over
+	pubEpoch := b.challenge.PublishEpoch()
+	// Add a 10% margin to the time it takes to generate a PoST proof
+	postDurationWithMargin := time.Duration(float64(b.lastPostGenDuration.Nanoseconds()) * 1.1)
+	postDurationInLayers := b.layerClock.DurationToLayers(postDurationWithMargin)
+	deadlineLayer := b.challenge.PubLayerID.Add(types.GetLayersPerEpoch()).Sub(postDurationInLayers)
+	deadline := b.layerClock.AwaitLayer(deadlineLayer)
```
countvonzero: this is the meat of this PR. i think it's worth factoring out to a method and test separately.
with
- some comments in the method about the algorithm to calculate the deadline
- referencing the research forum thread

activation/activation_errors.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+)
+
+var (
+	// ErrStopRequested is returned when builder is stopped.
+	ErrStopRequested = errors.New("builder: stop requested")
+	// ErrATXChallengeExpired is returned when atx missed its publication window and needs to be regenerated.
+	ErrATXChallengeExpired = errors.New("builder: atx expired")
+	// ErrPoetServiceUnstable is returned when poet quality of service is low.
+	ErrPoetServiceUnstable = &PoetSvcUnstableError{}
```
countvonzero, line 14: in code you always use it with &PoetSvcUnstableError{custom msg}
and the only purpose was for node_service.go to be able to return codes.Unavailable vs codes.Internal.
i find it pointless.

i find it easier to simply remove this error all together and just do fmt.Errorf() each time. in node_service.go, just return codes.Internal on error.

activation/nipost.go
```diff
@@ -85,100 +87,81 @@ func NewNIPostBuilder(
...
+		nb.state.PoetRequests = poetRequests
 		nb.persist()
 	}
 
-	// Phase 1: receive proofs from PoET service
+	// Phase 1: receive proofs from PoET services
 	if nb.state.PoetProofRef == nil {
-		var poetProofRef []byte
+		awaitProofsCtx, cancel := context.WithCancel(ctx)
+		go func() {
```
countvonzero: please use errgroup to manage go routine.

activation/nipost.go
```diff
@@ -192,5 +175,135 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash3
...
+func (nb *NIPostBuilder) awaitPoetProof(ctx context.Context, challenge *types.Hash32) types.PoetProofRef {
+	incomingPoetProofRefs := make(chan types.PoetProofRef, len(nb.state.PoetRequests))
+	g, ctx := errgroup.WithContext(ctx)
+
+	// Spawn workers to concurrently wait for PoET proofs
+	for _, poetService := range nb.state.PoetRequests {
+		svc := poetService
+		g.Go(func() error {
+			proofSubscription := nb.poetDB.SubscribeToProofRef(svc.PoetServiceID, svc.PoetRound.ID)
+			defer nb.poetDB.UnsubscribeFromProofRef(svc.PoetServiceID, svc.PoetRound.ID)
```
countvonzero: the old code only call this `UnsubscribeFromProofRef()` when atx expired. not when shutdown.
do you know what the impact is to delete this when app shutdown (and restarted within the same layer)?

activation/nipost.go
```diff
@@ -192,5 +175,135 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash3
...
+			case <-ctx.Done():
+			}
+			return nil
+		})
+	}
+
+	result := make(chan types.PoetProofRef)
+
+	// Spawn consumer processing received proofs
+	go func() {
```
countvonzero: please use a separate top-level errgroup to manage goroutines. that way this goroutine can be terminated when app shutdown

common/types/nipost.go
```diff
@@ -1,19 +1,22 @@
 package types
 
-//go:generate scalegen -types NIPostBuilderState
+//go:generate scalegen -types NIPostBuilderState,PoetRequest
 
 // NIPostBuilderState is a builder state.
```
countvonzero: this comment doesn't belong here

activation/nipost.go
```diff
@@ -85,100 +87,81 @@ func NewNIPostBuilder(
...
-		}
-		if !membership[*nipost.Challenge] {
-			round := nb.state.PoetRound
-			nb.state.PoetRound = nil // no point in waiting in Phase 1 since we are already received a proof
-			return nil, fmt.Errorf("%w not a member of this round (poetId: %x, roundId: %s, challenge: %x, num of members: %d)", ErrPoetServiceUnstable,
-				nb.state.PoetServiceID, round.ID, *nipost.Challenge, len(membership)) // TODO(noamnelke): handle this case!
+		if poetProofRef == nil {
+			// Haven't received any poet proof
+			select {
+			case <-awaitProofsCtx.Done():
```
countvonzero: i find the original select block easier to understand.
maybe you want to add a check in case poetProofRef where context.Error() == nil so that you dont continue while app is shutting down. at least you see the logic in one code block.

the new code is similar in that after returning from `awaitPoetProof`, it still need to check for `awaitProofsCtx` again. but now the logic is broken in different places.

common/types/nipost.go
```diff
@@ -1,19 +1,22 @@
...
+type PoetRequest struct {
```
countvonzero, line 6: nit: why not `PoetService`? it's what you use for variable name in the code.

log/zap.go
```diff
@@ -158,6 +158,11 @@ func Array(name string, array ArrayMarshaler) Field {
 	return Field(zap.Array(name, array))
 }
 
+// Strings for logging slice of strings.
+func Strings(name string, slice []string) Field {
```
countvonzero: don't want to use ArrayMarshaler?

timesync/converter.go
```diff
@@ -24,3 +24,7 @@ func (lc LayerConv) TimeToLayer(t time.Time) types.LayerID {
 func (lc LayerConv) LayerToTime(id types.LayerID) time.Time {
 	return lc.genesis.Add(time.Duration(id.Uint32()) * lc.duration)
 }
+
+func (lc LayerConv) DurationToLayers(duration time.Duration) uint32 {
+	return uint32(duration.Nanoseconds() / lc.duration.Nanoseconds())
```
countvonzero: how is this supposed to be used?
say layer time is 10 sec. duration is 234. number of layers is 23.
how does the client use this API?

activation/activation.go
countvonzero: or add comments to the method that uses this field would be good too.

activation/nipost.go
fasmat: Then go with `poetProofDeadline`.

activation/activation_errors.go
fasmat: The reason `ErrPoetServiceUnstable` exists is to be able to call `errors.Is(err, ErrPoetServiceUnstable)` when checking for that specific error.

I don't mind having this variable, but it could be removed and `errors.As` be used instead of `errors.Is`.

activation/nipost.go
fasmat: This is actually a good example of why I'm not a fan of the `deadline` channel. If deadline was passed in as a timestamp this would just be.

```golang
awaitProofCtx, cancel := context.WithDeadline(ctx, deadline)
defer cancel() // just to be sure that the awaitProofCtx is properly cleaned up; doing it on return works

Comment (fasmat) [reply activation/nipost.go]: errGroup here is in my opinion overkill. There are no errors to be handled and there is only one go routine started, so no synchronisation with `errGroup.Wait()` is needed.

File: activation/activation_errors.go (hunk 35 above)
Comment (countvonzero) [reply activation/activation_errors.go:14]: right. but the only places that check errors.Is() against `ErrPoetServiceUnstable` is in node_service.go and in tests.
for node_service.go, the differentiation of codes.Unavailable and codes.Internal does not make sense.
for tests, i'd prefer more specific errors to be checked, not a category of `ErrPoetServiceUnstable`.
for example, errUnableToRegister, errServiceIdUnavailable...

File: activation/nipost.go (hunk 36 above)
Comment (countvonzero) [reply activation/nipost.go]: first of all, this section of code can be eliminated completely 
see my comments here. https://github.com/spacemeshos/go-spacemesh/pull/3603/files/721cf5c9816b29ac3f1c2531c66b2d3447b45745#r994039013

secondly, in each component, we are trying to use a top level errgroup to manage goroutines so that when the app shut down, all outstanding goroutines can be terminated. 

in this case, we must wait after deadline expires before this goroutine can exit. this is not good, because deadline can be multiple layers. for errgroup use pattern, see any background goroutines. for example tortoise/algorithm.go Tortoise.Stop()

File: activation/nipost.go (hunk 37 above)
Comment (poszu) [reply activation/nipost.go]: I'm not sure what you mean. The `poetDbAPI` interface, the `nb.poetDB` type, doesn't specify any assumptions or requirements that must be met when calling `Subscribe`/`Unsubscribe`. We can look at the implementation of this interface in `poetdb.go`, which is the only non-mock one at the moment. It stores subscriptions in memory and thus it doesn't survive a restart.

Personally, I find it correct to unsubscribe after we subscribed and `defer` makes sure of it.

File: activation/nipost.go (hunk 38 above)
Comment (poszu) [reply activation/nipost.go]: I don't really follow what you mean here. Just using an `errgroup` doesn't guarantee that a goroutine will finish, even if the parent context of the errgroup is cancelled.

What if there is a "temporary" task to do, meaning spawn a few workers, and wait for them to finish, like in this case? If I use a "parent" errgroup I will not be able to `.Wait()` for them to finish.

The way to not use raw `go func() {}()` here would be to use a **separate** (different than `g`) and **local** errgroup. But what is the advantage? `errgroup` approach is just the same error-prone as the current one (one might forget to check if `ctx.Done()`. I think `errgroup` here is overkill and would complicate the code (need to select on two channels), though it would be more explicit. I'm sorry if I'm missing some obvious benefit here. 

Here I want to guarantee that all goroutines are finished before I return from this function. This goroutine is guaranteed to finish on app shutdown because `incomingPoetProofRefs` channel will be closed and I wait for its result.

File: common/types/nipost.go (hunk 41 above)
Comment (poszu) [reply common/types/nipost.go:6]: Because it describes an in-flight request to a PoetService, not a service itself. It contains the round ID of the submitted challenge. It could be named `SubmittedPoetChallenge`, or something like that.

File: log/zap.go (hunk 42 above)
Comment (poszu) [reply log/zap.go]: Idk what you mean. `[]string` doesn't implement `ArrayMarshaler`.

File: timesync/converter.go (hunk 43 above)
Comment (poszu) [reply timesync/converter.go]: It transforms time duration to the number of layers it spans. It is described in the interface definition:
>// Convert time duration to number of layers it spreads

Correct, if the layer duration is 10s and the duration passed in is 234s, then it outputs 23.

File: activation/activation.go (hunk 33 above)
Comment (poszu) [reply activation/activation.go:498]: Good catch!

There is a test for it and it passes in CI... https://github.com/spacemeshos/go-spacemesh/blob/721cf5c9816b29ac3f1c2531c66b2d3447b45745/activation/activation_test.go#L1117-L1137

File: activation/activation_errors.go (hunk 35 above)
Comment (poszu) [reply activation/activation_errors.go:14]: IMO, the main reason to have different error **types** is to be able to react differently to different errors (control flow). If there is no possible error handling, then simple `fmt.Errorf()` is enough - I'm not a fan of creating error types which are then only logged. `ErrPoetServiceUnstable` is handled in one critical place and thus must be identified: https://github.com/spacemeshos/go-spacemesh/blob/721cf5c9816b29ac3f1c2531c66b2d3447b45745/activation/activation.go#L410

File: activation/nipost.go (hunk 36 above)
Comment (poszu) [reply activation/nipost.go]: > secondly, in each component, we are trying to use a top level errgroup to manage goroutines so that when the app shut down, all outstanding goroutines can be terminated.

Please correct me if I'm wrong. Using an errgroup doesn't guarantee that the goroutine will be terminated when the parent context is cancelled. It still needs to select on `ctx.Done()` manually.

Comment (poszu) [reply activation/nipost.go]: @fasmat 
>This is actually a good example of why I'm not a fan of the deadline channel. If deadline was passed in as a timestamp this would just be.

I'm fine with passing the deadline as a timestamp.

File: activation/activation.go (hunk 33 above)
Comment (countvonzero) [reply activation/activation.go:498]: why did it pass???

File: log/zap.go (hunk 42 above)
Comment (countvonzero) [reply log/zap.go]: instead of defining a function for every type of array you want to log, you can do something like this

https://github.com/spacemeshos/go-spacemesh/blob/a316d780f95241a38cc0db1cf3a0645ef0891811/miner/oracle.go#L175

the advantage is that the marshaling is lazy.

File: timesync/converter.go (hunk 43 above)
Comment (countvonzero) [reply timesync/converter.go]: my main thought this, as an API is not generic enough and easy to misuse.

say numLayers := lc.DurationToLayers(234s)
if the client want to wait for `duration` to elapse, then waiting till currentLayer+numLayers != 234s.

i think this should be just a line of code at the call site. `duration/types.LayerDuration()` and the context in the call site make it clear how it's going to be used.
we don't currently have `types.LayerDuration()` but you can follow the setup for `types.GetLayersPerEpoch()` from app.Config value.

File: activation/nipost.go (hunk 37 above)
Comment (countvonzero) [reply activation/nipost.go]: i was trying to make a point that in this function, when context.Done() unblocks, one doesn't really know what actually happened. is it app shutdown or has the deadline expired. i dislike this kind of multi-purpose context that makes application logic harder to understand. in this case, `UnsubscribeFromProofRef` happens to be in-memory only. but what if it is changed to persist across restart? then this method has to be refactored again.

i think the original select (with extra check on shutdown) is more readable.
```
select {
	case poetProofRef = <-nb.poetDB.SubscribeToProofRef(nb.state.PoetServiceID, nb.state.PoetRound.ID):
		if ctx.Error() != nil {
			// in case shutdown happens at the same time as poet proof arrival
			return nil, 0, fmt.Errorf("failed to get poet proofs: %w", ErrStopRequested)
		}
	case <-deadline:
		nb.poetDB.UnsubscribeFromProofRef(nb.state.PoetServiceID, nb.state.PoetRound.ID)
		return nil, 0, fmt.Errorf("%w: while waiting for poet proof, target epoch ended", ErrATXChallengeExpired)
	case <-ctx.Done():
		return nil, 0, fmt.Errorf("failed to get poet proofs: %w", ErrStopRequested)
	default:
}
```
this way you see all the possible outcome in one select block. 
**having said that, i am fine with however you decide to write the code.**
besides, if you manage to get rid of SubscribeToProofRef and UnsubscribeFromProofRef. the app logic can be even more simplified.

File: activation/nipost.go (hunk 36 above)
Comment (countvonzero) [reply activation/nipost.go]: regarding errgroup, the idea is to facilitate graceful shutdown as journaled in https://github.com/spacemeshos/go-spacemesh/issues/2094

the general approach in go-spacemesh now for each component is
- app init routine optionally passes the globalcontext to each component at startup
- for each component:
  - w global context: it creates a context and cancel func out of that global context, and use errgroup to start goroutines
  - w/o global context: it creates an internal stop channel, and use errgroup to start goroutines
  - any goroutine a component starts, it either listen to the child context of global context, or the stop channel for app shutdown
  - in its Stop() or Close(), it either calls the cancel func on the context, or close the stop channel. then block on errgroup.Wait() until it finishes the necessary shutdown routine
- when app shutdown (the global context got canceled, see code in cmd/node.go and cmd/base.go), it calls each component's Stop() explicitly in its shudown routine.

timesync/clock.go is an example that uses stop channel, and blocks/certifier.go is an example of the cancel func.

the ultimate goal is:
- do `eg, ctx := errgroup.WithContext(globalCtx)` in node.go
- pass `eg` and `ctx` to each component for necessary goroutine management

this is the reason i asked to use errgroup to manage goroutine in the first place.

would be happy if you are willing to take one step further in this PR to achieve the ultimate goal (passing the eg down to nipost.go). but following other component is fine too.

File: activation/nipost.go (hunk 38 above)
Comment (countvonzero) [reply activation/nipost.go]: please see https://github.com/spacemeshos/go-spacemesh/pull/3603#discussion_r995305653

Comment (countvonzero) [reply activation/nipost.go]: but i am hoping all these spawn goroutines for processing proofs will be removed with the lazy approach to retrieve proofs from DB

File: log/zap.go (hunk 42 above)
Comment (countvonzero) [reply log/zap.go]: but this doesnt really matter. feel free to leave the code as is.

File: timesync/converter.go (hunk 43 above)
Comment (poszu) [reply timesync/converter.go]: Removed

File: activation/activation.go (hunk 33 above)
Comment (poszu) [reply activation/activation.go:498]: Fixed

File: activation/activation.go (hunk 34 above)
Comment (poszu) [reply activation/activation.go]: It's been simplified a lot now.

**Comment by poszu** [comment]
bors try

File: activation/activation.go (hunk 34 above)
Comment (poszu) [reply activation/activation.go]: Added description in a comment.

**Comment by poszu** [comment]
bors try

**Comment by poszu** [comment]
bors try

File: activation/activation.go [hunk 44]
```diff
@@ -392,15 +394,13 @@ func (b *Builder) loop(ctx context.Context) {
...
-			b.nipostBuilder.updatePoETProver(*(*PoetProvingServiceClient)(client))
-			// CaS here will not lose concurrent update
-			b.pendingPoetClient.CompareAndSwap(client, nil)
+		if poetClients := b.receivePendingPoetClients(); poetClients != nil {
+			b.nipostBuilder.updatePoETProvers(*poetClients)
 		}
 
 		ctx := log.WithNewSessionID(ctx)
 		if err := b.PublishActivationTx(ctx); err != nil {
+			b.log.With().Warning("PublishActivationTx failed", log.Err(err))
```
countvonzero: this seems redundant. please combine with the log on L407

activation/activation.go
```diff
@@ -411,9 +411,11 @@ func (b *Builder) loop(ctx context.Context) {
 
 			switch {
 			case errors.Is(err, ErrATXChallengeExpired):
+				b.log.Info("Discarding challenge")
```
countvonzero: are these log stmts in the the switch block leftoever from debugging?
the log stmt by itself doesn't have enough information on whats wrong.
i think either rm or change to debug level. if you choose latter, better do WithContext().With() for more info.

activation/activation.go
```diff
@@ -470,17 +473,28 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) error {
...
-func (b *Builder) UpdatePoETServer(ctx context.Context, target string) error {
-	b.log.With().Debug("request to update poet service", log.String("target", target))
-	client := b.poetClientInitializer(target)
-	// TODO(dshulyak) not enough information to verify that PoetServiceID matches with an expected one.
-	// Maybe it should be provided during update.
-	sid, err := client.PoetServiceID(ctx)
-	if err != nil {
-		return fmt.Errorf("%w: %v", ErrPoetServiceUnstable, err)
+func (b *Builder) UpdatePoETServers(ctx context.Context, endpoints []string) error {
+	b.log.With().Debug("request to update poet services", log.Array("endpoints", log.ArrayMarshalerFunc(func(encoder log.ArrayEncoder) error {
```
countvonzero: here and other places. probably good to do WithContext() as it also logs requestID/sessionID

activation/activation.go
```diff
@@ -619,27 +633,52 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
 	if err != nil {
 		return nil, fmt.Errorf("getting challenge hash failed: %w", err)
 	}
 
-	// the following method waits for a PoET proof, which should take ~1 epoch
-	expireLayer := (pubEpoch + 2).FirstLayer()
-	atxExpired := b.layerClock.AwaitLayer(expireLayer) // this fires when the target epoch is over
+	pubEpoch := b.challenge.PublishEpoch()
+	// Calculate deadline for waiting for poet proofs.
+	// Deadline must fir between current poet round end and
```
countvonzero: :s/fir/fit/

:s/between current poet round end/between the end of current poet round/

activation/activation.go
```diff
@@ -619,27 +633,52 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
+	//                                ┌┐
+	//         ┌─────────────────────┐││  ┌─────────────────────┐
+	//         │     POET ROUND      │││  │     POET ROUND      │
+	// ┌───────┴──────────────────┬──┴┴┴▲─┴─────────────────▲┬──┴───► time
+	// │           EPOCH          │     │      EPOCH        ││
+	// └──────────────────────────┴─────┼───────────────────┼┴──────
+	//                                  │				    │
+	//                              DEADLINE FOR	   ATX PUBLICATION
+	//                            WAITING FOR POET       DEADLINE
+	//                                PROOFS
```
countvonzero: nice........ 

it seems the line from `ATX PUBLICATION DEADLINE` is broken

activation/activation.go
```diff
@@ -619,27 +633,52 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
+	postDurationWithMargin := time.Duration(float64(b.lastPostGenDuration.Nanoseconds()) * 1.1)
```
countvonzero, line 663: can you add link to the forum thread? future devs are going to wonder the reasoning behind this algorithm....

activation/activation.go
```diff
@@ -619,27 +633,52 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
+	// Aim no later than the middle of the cycle gap
+	if postDurationWithMargin < b.poetCfg.CycleGap/2 {
+		postDurationWithMargin = b.poetCfg.CycleGap / 2
+	}
+	poetProofDeadline := b.layerClock.LayerToTime(b.challenge.PubLayerID).Add(b.poetCfg.PhaseShift).Add(-postDurationWithMargin)
```
countvonzero: `b.poetCfg.PhaseShift` is relative to the start of an epoch. i don't understand why adding it to `b.layerClock.LayerToTime(b.challenge.PubLayerID)` means.
do you mean `b.layerClock.LayerToTime(b.challenge.PublishEpoch().FirstLayer())` for the start of the next epoch?

activation/activation.go
```diff
@@ -619,27 +633,52 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
+	if time.Now().After(poetProofDeadline) {
+		b.log.With().Error("Poet proof deadline is already up!")
```
countvonzero: should the code return here?

poszu:
bors try

activation/activation.go
```diff
@@ -619,27 +633,52 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
 	}
 
-	// the following method waits for a PoET proof, which should take ~1 epoch
-	expireLayer := (pubEpoch + 2).FirstLayer()
-	atxExpired := b.layerClock.AwaitLayer(expireLayer) // this fires when the target epoch is over
+	pubEpoch := b.challenge.PublishEpoch()
+	// Calculate deadline for waiting for poet proofs.
+	// Deadline must fir between current poet round end and
+	// start of the next. It must also accommodate for PoST duration.
+	//                               PoST
```
countvonzero: the block for PoST should be after the deadline? (i.e. you need the proof as input to do PoST)

activation/activation.go
```diff
@@ -619,27 +633,52 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
...
+	// │           EPOCH          │     │      EPOCH        ││
+	// └──────────────────────────┴─────┼───────────────────┼┴──────
+	//                                  │				    │
+	//                              DEADLINE FOR	   ATX PUBLICATION
+	//                            WAITING FOR POET       DEADLINE
+	//                                PROOFS
+	postDurationWithMargin := time.Duration(float64(b.lastPostGenDuration.Nanoseconds()) * 1.1)
+	// Aim no later than the middle of the cycle gap
+	if postDurationWithMargin < b.poetCfg.CycleGap/2 {
+		postDurationWithMargin = b.poetCfg.CycleGap / 2
```
countvonzero: i don't understand the comparison with the cycle gap here. 

what if `postDurationWithMargin > cycleGap`? do we then skip an epoch? 
for now i think you can gather empirical data from running systests, and adjust cycleGap param accordingly.

so the deadline can then be
deadline := startOfNextEpoch + phaseShift - postDurationWithMargin - gracePeriod

here grace period is for network propagation (submitting to PoET)

note that there is a very similar calculation here. it is used to decide whether a miner has missed the window to start working on its very first ATX.
https://github.com/spacemeshos/go-spacemesh/blob/ddf936ac19da83108768a029ca1bb8ebe726cfba/activation/activation.go#L354

activation/nipost.go
```diff
@@ -85,100 +86,71 @@ func NewNIPostBuilder(
...
 		startTime := time.Now()
 		proof, proofMetadata, err := nb.postSetupProvider.GenerateProof(nb.state.PoetProofRef, commitmentAtx)
 		if err != nil {
-			return nil, fmt.Errorf("failed to execute Post: %v", err)
+			return nil, 0, fmt.Errorf("failed to execute Post: %v", err)
 		}
 
-		nb.log.With().Info("finished post execution",
-			log.Duration("duration", time.Since(startTime)))
+		postGenDuration = time.Since(startTime)
```
countvonzero, line 152: consider adding metrics here so one can collect data

activation/nipost.go
```diff
@@ -192,5 +164,106 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.Hash3
...
+
+		if bestProof == nil || bestProof.leafCount < proof.LeafCount {
+			bestProof = &poetProof{
+				ref:       ref,
+				leafCount: proof.LeafCount,
+			}
+		}
+	}
+
+	// Send the best proof (if any) and close the channel.
```
countvonzero: update comment. no more channel

countvonzero (commented):
mostly questions about deadline calculation.

instead of using 1.1X of historical data dynamically, for genesis we can simply gather data from tests and hardcode the buffer. leave a TODO for the dynamic buffer calculation later? we will have a homogeneous fleet of PoET servers to begin with.

activation/activation.go
poszu: Sorry, a leftover from debugging.

activation/activation.go
poszu: No. `b.layerClock.LayerToTime(b.challenge.PubLayerID).Add(b.poetCfg.PhaseShift)` marks the point in time when the next poet round starts. Then I subtract the required "safety gap". I can rewrite it a bit.

Btw, in my understanding, `b.challenge.PublishEpoch().FirstLayer()` == `b.challenge.PubLayerID`.

countvonzero: > Btw, in my understanding, b.challenge.PublishEpoch().FirstLayer() == b.challenge.PubLayerID.

you are right. i didn't know about this, and don't know if this is guaranteed. 
i just thought that the expression is self-explanatory for start of the next epoch, with more math underneath i guess.

poszu:
bors try

systest/tests/poets_test.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		if layer.Layer.GetStatus() != spacemeshv1.Layer_LAYER_STATUS_APPLIED {
+			return true, nil
+		}
+		// don't kill a poet if this is not ~middle of epoch
+		if ((layer.Layer.GetNumber().GetNumber() + epochSize/2) % epochSize) != 0 {
+			return true, nil
+		}
+
+		poetToDelete := cl.Poet(0)
```
countvonzero, line 82: would it work if you kill a random poet server each time?

countvonzero (approved):
nice work!

systest/tests/poets_test.go
poszu: It should work

poszu:
bors merge

poszu:
bors merge

PR #3604, opened 2022-09-29
[Merged by Bors] - Fix flakiness in PoST setup test

poszu:
## Motivation
The `TestPostSetupManager` test from `activation` pkg frequently fails. It's becoming frustrating. It throws the following error:
```
post_test.go:56:

Not equal:
expected: &activation.PostSetupStatus{State:2, NumLabelsWritten:0x800, LastOpts:(*activation.PostSetupOpts)(0xc0027721b0), LastError:error(nil)}
actual  : &activation.PostSetupStatus{State:3, NumLabelsWritten:0x800, LastOpts:(*activation.PostSetupOpts)(0xc0027721b0), LastError:error(nil)}

Diff:
--- Expected
+++ Actual
@@ -1,3 +1,3 @@
 (*activation.PostSetupStatus)({
- State: (activation.postSetupState) 2,
+ State: (activation.postSetupState) 3,
  NumLabelsWritten: (uint64) 2048,
TestPostSetupManager
```

The check on line 56 verifies if the `status` received from the channel equals the status returned from a direct call to the `Status()` method. However, IMO this is racy as the `mgr` runs in the background and its state can change at any time. I don't really understand the motivation for this check (and if it should be removed or fixed). Perhaps @moshababo could throw some light on it.

In this PR I propose to remove it or have a discussion how to properly fix this test

## Changes
Removed the flaky check.

## DevOps Notes


dshulyak (approved):
removing this looks fine, but i really don't know what was the test about @moshababo

fasmat (approved):

moshababo:
The motivation for this assertion is to expose correctness and race conditions issues with the state field, which apparently exists. It doesn't affect non-test code due to how the component is used. Relaxing the test is fine because the fixing of the underlying problem (which is also commented on in this test file) can be postponed.

moshababo (approved):

poszu:
bors merge

poszu:
bors merge

PR #3605, opened 2022-09-30
[Merged by Bors] - vm: introduce genesis id to vm

### LINKED ISSUE #3504: vm: include genesisid into a tx hash that is signed

- initialize vm with GenesisID
- add GenesisID to Context type
- prefix message that is hashed with GenesisID https://github.com/spacemeshos/go-spacemesh/blob/develop/genvm/templates/multisig/multisig.go#L39 in Verify 
- for sdk module: add WithGenesisID option and prefix hashed message 
  https://github.com/spacemeshos/go-spacemesh/blob/develop/genvm/sdk/multisig/tx.go#L108
  https://github.com/spacemeshos/go-spacemesh/blob/develop/genvm/sdk/options.go#L10

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3504

motivation behind this change is a cross-network replay protection. it works consistently across default templates, but in general the implementation is a template specific and other templates may go a different route.

for all templates that will be released in genesis (singlesig, multisig, vesting) replay protection works as follows:
- signer prefixes encoded transaction body with genesis id, hashes the result and signs it 
```
sig = sign(sha256(genesisid || body))
payload = body || sig
```
- verifier receives payload of the transaction. decodes sig. prefixes body with genesisid, hashes it and passes to ed25519 verify function


dshulyak:
bors try

dshulyak:
bors try

genvm/core/types.go
```diff
@@ -100,6 +102,7 @@ type Host interface {
 	Handler() Handler
 	Template() Template
 	Layer() LayerID
+	GetGenesisID() Hash20
```
countvonzero, line 105: how about an interface `Hash([]byte)` instead and the template gets to decide how to hash?

dshulyak: i prefer to use to use simple hash function instead of some interface method on the host

countvonzero: my rationale is that in the future custom templates don't necessarily care about genesis id, as long as the template can parse its own tx correctly.

systest/cluster/cluster.go
```diff
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"crypto/sha256"
```
countvonzero: "github.com/minio/sha256-simd"?

countvonzero (approved):

genvm/core/types.go
dshulyak: it is up to template how to implement replay protection, they are not necessarily will use hashing for that

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

PR #3613, opened 2022-10-03
[Merged by Bors] - Improve checks in grpcserver_test for better failure messages

poszu:
## Motivation
Currently, when the tests in the `grpcserver_test.go` file fail (which happen sporadically - the tests are flaky), they emit an enigmatic log:

> Error:          inner account data item has wrong data type

That is not super helpful. A test could print what type it actually got vs the exptected one.

## Changes
Use `require.IsType(...)` to check if `foo interface{}` is of the expected type. It will print the actual type of the `foo` in case of failure.

## DevOps Notes


countvonzero (approved):

poszu:
bors merge

PR #3615, opened 2022-10-05
[Merged by Bors] - Add method to be able to await ATX sync

fasmat:
## Motivation
Pull out some of the changes from https://github.com/spacemeshos/go-spacemesh/pull/3567 for easier review.

## Changes
The `syncer` that allows components of go-spacemesh to register for callbacks when sync states are reached is extended to allow to register for the ATX synced state.

Additionally the activation package is updated to listen for the new ATX synced state instead of waiting for full sync.

## Test Plan
Existing tests pass, new tests for new functionality.

## TODO

## DevOps Notes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -88,13 +88,13 @@ type Syncer struct {
 	lastLayerSynced   atomic.Value
 	lastATXsSynced    atomic.Value
 
-	mu sync.Mutex
-
 	// awaitSyncedCh is the list of subscribers' channels to notify when this node enters synced state
 	awaitSyncedCh []chan struct{}
+	muSyncedCh    sync.Mutex // protects the slice above from concurrent access
 
 	// awaitATXSyncedCh is the list of subscribers' channels to notify when this node enters ATX synced state
 	awaitATXSyncedCh []chan struct{}
+	muATXSyncedCh    sync.Mutex // protects the slice above from concurrent access
 
 	shutdownCtx context.Context
 	cancelFunc  context.CancelFunc
@@ -161,8 +161,8 @@ func (s *Syncer) RegisterForSynced(ctx context.Context) chan struct{} {
 		close(ch)
 		return ch
 	}
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	s.muSyncedCh.Lock()
+	defer s.muSyncedCh.Unlock()
 	s.awaitSyncedCh = append(s.awaitSyncedCh, ch)
 	return ch
 }
@@ -174,8 +174,8 @@ func (s *Syncer) RegisterForATXSynced() chan struct{} {
 		close(ch)
 		return ch
 	}
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	s.muATXSyncedCh.Lock()
+	defer s.muATXSyncedCh.Unlock()
 	s.awaitATXSyncedCh = append(s.awaitATXSyncedCh, ch)
 	return ch
 }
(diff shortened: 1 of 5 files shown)


fasmat:
bors try

syncer/syncer_test.go
```diff
@@ -130,16 +130,28 @@ func TestStartAndShutdown(t *testing.T) {
...
 
 	// the node is synced when current layer is <= 1
 	ts.syncer.Start(context.TODO())
-	<-syncedCh
+
+	select {
+	case <-syncedCh:
+	case <-time.After(1 * time.Second):
+		require.Fail(t, "node should be synced")
+	}
```
poszu, line 147: A helper would be nice as this pattern is used **a lot**.
```suggestion
	requireClosed(t, syncedCh, time.Second, "node should be synced")
```

syncer/syncer_test.go
```diff
@@ -570,3 +633,45 @@ func TestSync_AlsoSyncProcessedLayer(t *testing.T) {
...
+
+	ts.syncer.setSyncState(context.TODO(), synced)
+
+	select {
+	case <-sync:
+	case <-time.After(1 * time.Second):
+		require.Fail(t, "should have reached synced state")
+	}
+
+	require.NotPanics(t, func() { ts.syncer.setSyncState(context.TODO(), synced) })
```
poszu, line 655: Out of curiosity - what's the purpose of `require.NotPanics`? The test should fail if the code panics anyway, shouldn't it? What is the point?

syncer/syncer.go
```diff
@@ -227,6 +247,13 @@ func (s *Syncer) isClosed() bool {
 
 func (s *Syncer) setATXSynced() {
 	s.atxSyncState.Store(synced)
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for _, ch := range s.awaitATXSyncedCh {
+		close(ch)
+	}
+	s.awaitATXSyncedCh = make([]chan struct{}, 0)
```
poszu: Nit,  the `mu` Mutex so far guarded `awaitSyncedCh` only, though it's not documented explicitly what it is meant to be protecting. 😢 

The current implementation might be too strict - `awaitSyncedCh` can be safely accessed while `awaitATXSyncedCh` is modified. How about using a separate Mutex + document what the mutexes guard?

poszu (approved):

syncer/syncer_test.go
fasmat: I prefer being explicit over factoring out small functions that are only used in a few places.

The `select` doesn't check if the channel is closed, only if it blocks upon a read. Checking for it to be closed requires additional steps that aren't necessary for what's being asserted here. In some places the inverse is done, checking if it blocks.

Introducing two additional functions with trivial logic to only save 3 lines of code is imho not aiding readability.

syncer/syncer_test.go
fasmat: The called method used to panic before it was fixed. I wrote this test to explicitly check if the bug was fixed.

If I remove the `require.NotPanics` I would need to add a comment of the form:

```golang
// call setSyncState to check that it doesn't panic
ts.syncer.setSyncState(context.TODO(), synced)
```

Also if it does throw, the require prints an error message and continues with the next test. If not recovered the panic will stop the execution of the remaining tests (at least for the package).

fasmat:
bors merge

fasmat:
bors merge

PR #3616, opened 2022-10-05
[Merged by Bors] - Generalize storing of key value pairs in DB

fasmat:
## Motivation
Pull out some of the changes from https://github.com/spacemeshos/go-spacemesh/pull/3567 for easier review.

## Changes
The activation package needs to store transient data in the DB that is needed to create ATXs for the time between syncing global state and publishing a new ATX, after which this data can be discarded.

At the moment this is done for `NIPostBuilderState` and `NIPostChallange`. The changes in this PR generalize this storing, retrieving and clearing of data. In https://github.com/spacemeshos/go-spacemesh/pull/3567 another datatype is added to the `kvstore` introduced in this PR.

## Test Plan
Existing tests pass, new code is sufficiently tested.

## TODO

- [X] Explain motivation or link existing issue(s)

## DevOps Notes


fasmat:
bors try

sql/kvstore/nipost_challenge.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/codec"
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+const NIPostChallengeKey = "NIPost"
+
+// AddNIPostChallenge adds the data for nipost to the key-value store.
```
poszu, line 10: The logic in `AddNiPostChallenge()` is 100% the same as the logic in `AddNIPostBuilderState()`. The only difference is the type (`NiPostChallenge` vs `NIPostBuilderState`) and the key. The same probably applies to the `GetX()` functions. How about extracting the core DB logic to two separate functions:

```go
func Set(db sql.Executor, key []byte, value codec.Encodable) error;
func Get(db sql.Executor, key []byte, value codec.Decodable) error;
```

or like:

```go
func Get[T codec.Decodable](db sql.Executor, key []byte) (T, error);
```

poszu (commented):
I love the idea of factoring out DB logic! However, I think that it could be made more generically 😉

sql/kvstore/nipost_challenge.go
fasmat: Pulled out shared logic and moved it to `kvstore.go`

sql/kvstore/kvstore.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spacemeshos/go-scale"
+	"github.com/spacemeshos/go-spacemesh/codec"
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+func addKeyValue(db sql.Executor, key string, value scale.Encodable) error {
```
poszu, line 12: I don't get why pass a `string` to later convert to `[]byte`. Why not just:
```suggestion
func addKeyValue(db sql.Executor, key []byte, value scale.Encodable) error {
```

sql/kvstore/kvstore.go
```diff
@@ -0,0 +1,62 @@
...
+		return true
+	}); err != nil {
+		return fmt.Errorf("failed to get value: %w", err)
+	} else if rows == 0 {
+		return fmt.Errorf("failed to get value: %w", sql.ErrNotFound)
+	}
+
+	if len(val) == 0 {
+		return nil
+	}
```
poszu: Is this check really required?  This case should be handled by `codec.Decode`, shouldn't it?

poszu (approved):
Nice!

sql/kvstore/kvstore.go
fasmat: The reason is that `[]byte` cannot be `const` so we either have to do the conversation between `string` and `[]byte` here in this function or the caller of it has to convert.

I prefer to do the conversation here, to only have to do it once instead of everywhere this function is called.

sql/kvstore/kvstore.go
fasmat: Not quite: If the value is empty for a given key it will return no error. This is the behaviour as it was before the refactoring.

If I remove this check it will instead return an error (a wrapped `io.EOF` because `codec.Decode` tries to read from a buffer that is empty).

I'm not sure what is the correct behaviour here. I'll remove this check and see if anything breaks

fasmat:
bors merge

sql/kvstore/kvstore.go
poszu: By using a `string`, the API has limited usability. The `string` type is a subset of `[]byte` type, isn't it? `string` can be converted to `[]byte`, but not the other way around.

Therefore, it will not be possible to use this API to store objects with raw byte keys (for example a hash).

PR #3618, opened 2022-10-06
[Merged by Bors] - cleanup: remove unused await sync

countvonzero:
## Motivation

remove unused code

### DIFF
--- a/syncer/syncer_test.go
+++ b/syncer/syncer_test.go
@@ -198,6 +198,18 @@ func TestSynchronize_AllGood(t *testing.T) {
 		ts.mLyrFetcher.EXPECT().PollLayerData(gomock.Any(), lid).Return(okCh())
 	}
 
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		atxSyncedCh := ts.syncer.RegisterForATXSynced()
+		select {
+		case <-atxSyncedCh:
+			wg.Done()
+		case <-time.After(1 * time.Second):
+			require.Fail(t, "node should be atx synced")
+		}
+	}()
+
 	require.True(t, ts.syncer.synchronize(context.TODO()))
 	require.Equal(t, current.Sub(1), ts.syncer.getLastSyncedLayer())
 	require.Equal(t, current.GetEpoch(), ts.syncer.getLastSyncedATXs())
@@ -205,6 +217,18 @@ func TestSynchronize_AllGood(t *testing.T) {
 	require.True(t, ts.syncer.ListenToATXGossip())
 	require.True(t, ts.syncer.ListenToGossip())
 	require.False(t, ts.syncer.IsSynced(context.TODO()))
+
+	wg.Add(1)
+	go func() {
+		atxSyncedCh := ts.syncer.RegisterForATXSynced()
+		select {
+		case <-atxSyncedCh:
+			wg.Done()
+		case <-time.After(1 * time.Second):
+			require.Fail(t, "node should be atx synced")
+		}
+	}()
+	wg.Wait()
 }
 
 func TestSynchronize_FetchLayerDataFailed(t *testing.T) {
@@ -399,15 +423,8 @@ func TestFromSyncedToNotSynced(t *testing.T) {
 	ts := newSyncerWithoutSyncTimer(t)
 	ts.mLyrFetcher.EXPECT().GetEpochATXs(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
 
-	atxSyncedCh := ts.syncer.RegisterForATXSynced()
 	require.True(t, ts.syncer.synchronize(context.TODO()))
 	require.True(t, ts.syncer.IsSynced(context.TODO()))
-
-	select {
-	case <-atxSyncedCh:
-	case <-time.After(1 * time.Second):
-		require.Fail(t, "node should be atx synced")
-	}
 
 	// cause the syncer to get out of synced and then wait again
 	lyr := types.GetEffectiveGenesis().Add(1)
--- a/syncer/syncer_test.go
+++ b/syncer/syncer_test.go
@@ -130,9 +130,6 @@ func TestStartAndShutdown(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.TODO())
 	ts := newTestSyncer(ctx, t, time.Millisecond*5)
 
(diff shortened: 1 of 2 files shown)


countvonzero:
@fasmat please review. thanks

syncer/syncer_test.go
```diff
@@ -130,31 +130,16 @@ func TestStartAndShutdown(t *testing.T) {
...
-
-	select {
-	case <-atxSynchedCh:
-	case <-time.After(1 * time.Second):
-		require.Fail(t, "node should be atx synced")
-	}
-
-	require.True(t, ts.syncer.IsSynced(context.TODO()))
-	require.True(t, ts.syncer.ListenToATXGossip())
-	require.True(t, ts.syncer.ListenToGossip())
```
fasmat, line 157: Why where these asserts removed? Shouldn't they still be part of the test even after the removal of `RegisterForSynced`?

syncer/syncer_test.go
```diff
@@ -239,20 +224,6 @@ func TestSynchronize_FetchLayerDataFailed(t *testing.T) {
...
-		require.Fail(t, "node should not be synced")
-	case <-time.After(1 * time.Millisecond):
-	}
-
-	atxSyncedCh := ts.syncer.RegisterForATXSynced()
-	select {
-	case <-atxSyncedCh:
-	case <-time.After(1 * time.Second):
-		require.Fail(t, "node should be atx synced")
-	}
```
fasmat, line 255: I think that this assert is also required; If `ListenToATXGossip` returns `true` `RegisterForATXSynced` should return a closed channel.

syncer/syncer_test.go
```diff
@@ -272,20 +243,6 @@ func TestSynchronize_FailedInitialATXsSync(t *testing.T) {
...
-	case <-syncedCh:
-		require.Fail(t, "node should not be synced")
-	case <-time.After(100 * time.Millisecond):
-	}
-
-	atxSyncedCh := ts.syncer.RegisterForATXSynced()
-	select {
-	case <-atxSyncedCh:
-		require.Fail(t, "node should not be atx synced")
-	case <-time.After(100 * time.Millisecond):
```
fasmat, line 287: Same here. If ATX Sync failed, `RegisterForATXSynced` should return a blocking channel.

syncer/syncer_test.go
```diff
@@ -634,27 +569,6 @@ func TestSync_AlsoSyncProcessedLayer(t *testing.T) {
...
-	ts.syncer.setSyncState(context.TODO(), synced)
-
-	select {
-	case <-sync:
-	case <-time.After(1 * time.Second):
-		require.Fail(t, "should have reached synced state")
-	}
-
-	require.NotPanics(t, func() { ts.syncer.setSyncState(context.TODO(), synced) })
-}
```
fasmat, line 656: How do we deal with regressions? I wrote this test when there was a bug in the code that caused a panic. This test 1) verifies that the bug was fixed and 2) prevents a regression.

fasmat (changes requested):

syncer/syncer_test.go
countvonzero: this doesn't quite make sense without notifying channels. it's basically setting variables. having this test would be more confusing.

syncer/syncer_test.go
countvonzero: why isn't require.False(t, ts.syncer.ListenToGossip()) sufficient?

syncer/syncer_test.go
countvonzero: RegisterForATXSynced is just a method. you already tested it in `TestSyncer_setATXSyncedTwice_NoError` that it behaves correctly. `TestFromSyncedToNotSynced` also tested that it channel get notified when the node is synced. it does not need to be checked in EVERY test.

syncer/syncer_test.go
countvonzero: because it's ensured in line 141 now.

countvonzero:
@fasmat i added another async test that make sure
- ch is notified after node is atx synced
- ch is notified right away after the node is atx synced

syncer/syncer_test.go
fasmat: Because after that change we have no test that checks that `RegisterForATXSynced` returns a blocked channel when the state is "not synced". Right now it does when `ts.syncer.ListenToGossip()` returns `false`, but implementations change over time.

The question is should we test as if `syncer` is a black box or not? The former leads inevitably to more tests and more test coverage, the later requires someone changing the implementation to know how it is used in other places so they don't break things.

syncer/syncer_test.go
fasmat: This bug occurred in a system test. Here are the exact circumstances:

1. A go routine called `RegisterForSynced` and received a channel that it was using to be notified when the synced state happened.
2. A second go routine called `setSyncState` with `synced` triggering the notification of the first go routine.
3. The second go routine later called `setSyncState` again, without someone Registering for it.
4. PANIC -> because `setSyncState` tried to close channels that were already closed

The fix was clearing the internal map holding the channels created by `RegisterForSynced` after closing them.

Steps I took to fix the bug:

1. Write a test that should pass but fails because of the bug
2. Fix the bug
3. Verify that the test is now passing and leave it to avoid a regression

syncer/syncer_test.go
countvonzero: added async tests in TestSynchronize_AllGood

syncer/syncer_test.go
countvonzero: ok. added the test back, in async fashion, with registering before and after

syncer/syncer_test.go
countvonzero: i am confused. i did NOT remove TestSyncer_setATXSyncedTwice_NoError.
i only removed TestSyncer_setSyncedTwice_NoError because there is no longer a internal map holding the channels for synced-ness, only for ATX synced-ness.

fasmat: 😓 seems like I need more coffee; you are right. I mistook both tests to be the same. So forget my complaints.

fasmat (approved):

fasmat:
bors merge

countvonzero:
bors merge

PR #3619, opened 2022-10-06
tortoise: use max base tick height to determine votable blocks

dshulyak:
extracted from https://github.com/spacemeshos/go-spacemesh/issues/3573

blocks are generated using max base tick height from the atxs that are referenced by the proposals. ballots that vote for this block are expected to have higher tick height by one epoch, if such condition doesn't hold votes from this ballots are ineffective.

using max base tick height assumes that no smesher is faster by a whole epoch. if such smesher exists verifying tortoise will stop working. but the actual difference between current and previous solution is minimal, in previous solution verifying tortoise will stop working as soon as superfast smesher participated in building a block, and in current version we assume that it always participates in building a block.

### DIFF
--- a/tortoise/verifying.go
+++ b/tortoise/verifying.go
@@ -56,7 +56,7 @@ func (v *verifying) countVotes(logger log.Log, lid types.LayerID, ballots []*bal
 	var (
 		sum       = util.WeightFromUint64(0)
 		n         int
-		refheight = v.findRefHeightBelow(lid)
+		refheight = v.maxBaseTickHeight[lid.Sub(1).GetEpoch()]
 	)
 	for _, ballot := range ballots {
 		base, exist := v.ballotRefs[ballot.base.id]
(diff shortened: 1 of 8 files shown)


countvonzero (approved):

tortoise/verifying.go
```diff
@@ -155,7 +157,7 @@ func (v *verifying) verify(logger log.Log, lid types.LayerID) bool {
 		logger,
 		layer.blocks,
 		func(block *blockInfo) sign {
-			if block.height > layer.verifying.referenceHeight {
+			if block.height > refheight {
 				return neutral
```
countvonzero, line 161: i keep forgetting why it votes neutral instead of against here. can you remind me please, or maybe add comment?

dshulyak: if ballot votes on a block with a higher height the vote is ignored. this is how loop looks in a full mode https://github.com/spacemeshos/go-spacemesh/blob/develop/tortoise/full.go#L51

in verifying mode we can't enforce that condition for every ballot/block pair. instead only ballots that are higher than the reference height are counted, but if block has a higher height than that reference height we don't know what fraction of weight would be ignored - therefore opinion is neutral.

PR #3620, opened 2022-10-06
[Merged by Bors] - tortoise: refactor layerVote to store only supported blocks

dshulyak:
extracted from https://github.com/spacemeshos/go-spacemesh/issues/3573

this refactoring makes it unnecessary to modify layer votes for each ballot after late block is received. 
for example late block can be received after recovering from partition.

### DIFF
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -124,11 +124,6 @@ func (s *state) updateRefHeight(layer *layerInfo, block *blockInfo) {
 }
 
 type (
-	blockVote struct {
-		*blockInfo
-		vote sign
-	}
-
 	baseInfo struct {
 		id    types.BallotID
 		layer types.LayerID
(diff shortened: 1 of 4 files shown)


tortoise/state.go
```diff
@@ -204,29 +199,39 @@ func (v *votes) cutBefore(lid types.LayerID) {
...
 	}
 	return abstain
 }
 
 type layerVote struct {
 	*layerInfo
-	vote   sign
-	blocks []blockVote
+	vote      sign
+	supported []*blockInfo
```
countvonzero, line 216: why not a map here? does order matter?

countvonzero (approved):

tortoise/state.go
dshulyak: maps have high overhead. i am trying to avoid using them where possible

tortoise/tortoise.go
```diff
@@ -904,9 +899,9 @@ func validateConsistency(state *state, config Config, ballot *ballotInfo) bool {
...
 		for j := range lvote.blocks {
-			local, _ := getLocalVote(state, config, lvote.blocks[j].blockInfo)
-			vote := lvote.blocks[j].vote
-			if vote != local {
+			block := lvote.blocks[j]
+			local, _ := getLocalVote(state, config, block)
+			if lvote.getVote(block.id) != local {
 				return false
 			}
 		}
```
poszu: How about using a `for _, block := range lvote.blocks {` loop instead?

dshulyak:
bors merge

Issue #3626, opened 2022-10-07
Test `TestSpacemeshApp_NodeService` from `cmd/node` pkg fails on Windows


poszu:
## Description
Test `TestSpacemeshApp_NodeService` from `cmd/node` pkg fails on Windows

## Steps to reproduce
go test -timeout 4m -p 1 -run "TestSpacemeshApp_NodeService" github.com/spacemeshos/go-spacemesh/cmd/node

## Actual Behavior
```sh
=== FAIL: cmd/node TestSpacemeshApp_NodeService (4.17s)
2022-10-07T14:33:49.901+0200    ERROR   00000.defaultLogger     couldn't load config file at location: ./config.toml switching to defaults
 error: failed to read config file open ./config.toml: The system cannot find the file specified..
    node_test.go:558:
                Error Trace:    C:\Users\barte\workspace\go-spacemesh\cmd\node\node_test.go:558
                Error:          Received unexpected error:
                                new harness: exec: "Stop-Process": executable file not found in %PATH%
                Test:           TestSpacemeshApp_NodeService
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x30 pc=0x7ff747d06165]

goroutine 32 [running]:
testing.tRunner.func1.2({0x7ff748266b20, 0x7ff747e92f40})
        C:/Users/barte/scoop/apps/go/current/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
        C:/Users/barte/scoop/apps/go/current/src/testing/testing.go:1399 +0x39f
panic({0x7ff748266b20, 0x7ff747e92f40})
        C:/Users/barte/scoop/apps/go/current/src/runtime/panic.go:884 +0x212
github.com/spacemeshos/go-spacemesh/cmd/node.TestSpacemeshApp_NodeService.func1()
        C:/Users/barte/workspace/go-spacemesh/cmd/node/node_test.go:553 +0x25
testing.(*common).Cleanup.func1()
        C:/Users/barte/scoop/apps/go/current/src/testing/testing.go:1041 +0x11f
testing.(*common).runCleanup(0xc000105520, 0x7ff747042d46?)
        C:/Users/barte/scoop/apps/go/current/src/testing/testing.go:1210 +0x95
testing.tRunner.func2()
[...]

Issue #3625, opened 2022-10-07
Test `TestNIPostBuilderWithClients` from `activation` pkg times out on Windows


poszu:
## Description
Test `TestNIPostBuilderWithClients` from `activation` pkg times out on Windows

## Steps to reproduce
go test --timeout=30s -run "^TestNIPostBuilderWithClients$" github.com/spacemeshos/go-spacemesh/activation

## Actual Behavior
The test hangs in clean up awaiting for the PoET harness to be shut down.

## Expected Behavior
Test passes

## Environment
 - OS:  Windows

fasmat:
Fixed with https://github.com/spacemeshos/go-spacemesh/pull/3679

Issue #3624, opened 2022-10-07
Few tests from `txs` pkg fails on Windows


poszu:
## Description
4 tests fail on Windows:
- TestSelectBlockTXs_NoOptimisticFiltering
- TestSelectProposalTXs_ExhaustMemPool
- Test_checkStateConsensus_ownMeshDiffer
- Test_checkStateConsensus_NoConsensus

## Steps to reproduce
Run on Windows:
` go test -timeout 30s -p 1 github.com/spacemeshos/go-spacemesh/txs`

#### TestSelectBlockTXs_NoOptimisticFiltering
```sh
=== FAIL: txs TestSelectBlockTXs_NoOptimisticFiltering (0.08s)
    conservative_state_test.go:252:
                Error Trace:    C:\Users\barte\workspace\go-spacemesh\txs\conservative_state_test.go:252
                Error:          Received unexpected error:
                                node has different mesh hash from majority
                Test:           TestSelectBlockTXs_NoOptimisticFiltering
```

#### TestSelectProposalTXs_ExhaustMemPool
```sh
=== FAIL: txs TestSelectProposalTXs_ExhaustMemPool (0.02s)
    conservative_state_test.go:365:
                Error Trace:    C:\Users\barte\workspace\go-spacemesh\txs\conservative_state_test.go:365
                Error:          Not equal:
[...]

PR #3627, opened 2022-10-07
[Merged by Bors] - Fix most and skip some unit tests for Windows

poszu:
## Motivation
Some unit tests fail in CI on Windows

## Changes
Fixed most unit tests. Tests that require more work are skipped. I created bug tickets for skipped tests:
- https://github.com/spacemeshos/go-spacemesh/issues/3624
- https://github.com/spacemeshos/go-spacemesh/issues/3625
- https://github.com/spacemeshos/go-spacemesh/issues/3626
- https://github.com/spacemeshos/go-spacemesh/issues/3628
- https://github.com/spacemeshos/go-spacemesh/issues/3629

## Test Plan
unit tests

## DevOps Notes


poszu:
bors try

activation/nipost_test.go
```diff
@@ -180,6 +180,9 @@ func TestPostSetup(t *testing.T) {
 }
 
 func TestNIPostBuilderWithClients(t *testing.T) {
+	if os.Getenv("GOOS") == "windows" {
```
countvonzero: plz create named constants for these OSs and env variables so every test in all packages can use them.

common/types/testutil.go
```diff
@@ -2,14 +2,12 @@ package types
...
 import (
 	"math/rand"
-	"time"
 
 	"github.com/spacemeshos/go-spacemesh/signing"
 )
 
 // RandomBytes generates random data in bytes for testing.
 func RandomBytes(size int) []byte {
-	rand.Seed(time.Now().UnixNano())
```
countvonzero, line 12: curious. what motivates removing randomness here?

config/config_test.go
```diff
@@ -25,7 +25,4 @@ func TestConfig_DataDir(t *testing.T) {
 	config.DataDirParent = "~" + sep + "space-a-mesh"
 	expectedDataDir := filesystem.GetUserHomeDirectory() + sep + "space-a-mesh"
 	assert.Equal(t, expectedDataDir, config.DataDir())
-
-	config.DataDirParent = "~" + sep + "space-a-mesh" + sep // trailing slash should be ignored
```
countvonzero: why not test this when OS is ubuntu?

go.mod
```diff
@@ -32,9 +32,9 @@ require (
 	github.com/spacemeshos/economics v0.0.0-20220810190316-045462b489e8
 	github.com/spacemeshos/ed25519 v0.0.0-20220825090204-1f103b5756da
 	github.com/spacemeshos/fixed v0.0.0-20210523192743-8d17e03c169a
```
countvonzero, line 34: are the changes in go.mod and go.sum related to the fixes? if not, can you break them into a separate PR?

poszu: They are - Poet was upgraded.

common/types/testutil.go
poszu: Time on Windows appears to have low resolution. Because of it, subsequent calls to `rand.Seed(...)` can be given the same seed value. As a result, it returns exactly the same "random" value in subsequent calls.

I'm not sure why it was given an explicit seed in the first place.

countvonzero: rand seems wonky in the codebase right now. some use directly from math/rand, some from go-spacemesh rand pkg.

go.mod
countvonzero: not a big deal. but how does poet upgrade affect these tests?

activation/nipost_test.go
poszu: I'm not certain it gives any benefit. IMHO, it would obfuscate the code. Named constants are useful if the constant might change in the future or if we want to make sure the same value is used everywhere. Then such a change is easier, a one-liner. However, here we know for sure these constants will never change. Also, using a different value just wouldn't work obviously (easy to notice). 

Anyway, where do you suggest putting them?

countvonzero: sure these constants will never change, but typos in the new code can happen.

for example
```
if os.Getenv("GOOS") != "wondows" {
    // skip
}
```
then one will never notice because it never run on any OS.
common/utils?

config/config_test.go
poszu: This test was already completely removed on `develop`.

go.mod
poszu: There was a bug in Poet integration harness that caused it hang during shutdown. See https://github.com/spacemeshos/poet/pull/128.

countvonzero (approved):
thanks for filing those issues on windows!

poszu:
bors merge

PR #3631, opened 2022-10-07
[Merged by Bors] - Don't build in `setup env` CI step in every job

poszu:
## Motivation
The default target installs deps and builds. The building part is not required to set up the environment in CI.

## Changes
Call `make install` instead of `make` in setup env step in CI.

## Test Plan
CI passes

## DevOps Notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
             ${{ runner.os }}-cache-go-${{ env.go-version }}-
       - name: fmt, tidy, lint
         run: |
-          make
+          make install
           make tidy
           make test-fmt
       - name: staticcheck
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -98,10 +98,40 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-cache-go-${{ env.go-version }}-
       - name: setup env
-        run: make
+        run: make install get-libs
```
fasmat: imo `get-libs` shouldn't be needed to be called called explicitly. If `golangci-lint-github-action` requires it, we could just add it there as a prerequisite (like for `make test` and other targets that need it).

fasmat (approved):

fasmat (approved):

poszu:
bors merge

PR #3632, opened 2022-10-10
[Merged by Bors] - tortoise: update opinion within sliding window

### LINKED ISSUE #2920: miner should use weak coin value from layer N-1 to produce block in layer N
## Description

when miner create a block and decide how to vote in that block, it uses weak coin to break tie.
the code today uses coin flip recorded in layer `t.Last-1`.
https://github.com/spacemeshos/go-spacemesh/blob/9c8b62c01095445d7e3c06faeb37b3803b92456d/tortoise/verifying_tortoise.go#L1100

this is not correct. a block produced in block N should use weak coin from layer N-1.

### LINKED ISSUE #3573: tortoise: update opinion on the terminated layers within sliding window when received new data

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3573
closes: https://github.com/spacemeshos/go-spacemesh/issues/2920
depends on: https://github.com/spacemeshos/go-spacemesh/pull/3596

- added OnAtx tortoise, similarly to OnBallot and OnBlock. 
this handler is required to receive late atxs and expected to be called before ballots that are using referenced atx
are sent to OnBallot
- ballots are counted for all layers within the window
- all layers within the window are verified on calling TallyVotes (there are optimization options, but they can be made later if needed)
- verifying tortoise is reworked so that total good weight doesn't have to be updated after verifying a layer
- integrated OnAtx to atx handler

removed code:
- separate windows for full and verifying tortoise are removed. as a simplification both modes will use same window
- mark good cut routine is removed. approach described #3637 will allow to skip that
- outdated tests are removed
- background rerun is removed


datastore/store.go
```diff
@@ -186,17 +186,3 @@ func (bs *BlobStore) Get(hint Hint, key []byte) ([]byte, error) {
 	}
 	return nil, fmt.Errorf("blob store not found %s", hint)
 }
-
-func getHeader(atx *types.VerifiedActivationTx) *types.ActivationTxHeader {
```
countvonzero, line 190: why remove this?

the motivation to keep private here is that header should be cached here. being public in common pkg encourage new code to freely create instances of the header.

tortoise/algorithm.go
```diff
@@ -130,35 +118,24 @@ func New(cdb *datastore.CachedDB, beacons system.BeaconGetter, updater blockVali
...
-			log.Stringer("historically_verified", t.cfg.MeshVerified),
+			log.Stringer("last layer", t.cfg.MeshProcessed),
 		)
 		t.eg.Go(func() error {
-			t.ready <- t.rerun(ctx)
+			for lid := types.GetEffectiveGenesis().Add(1); lid.After(t.cfg.MeshProcessed); lid = lid.Add(1) {
+				err := t.trtl.onLayer(ctx, lid)
+				if err != nil {
+					t.ready <- err
+					return err
```
countvonzero, line 129: this causes other errgroup managed goroutine to exit. is this the intention?

tortoise/full.go
```diff
@@ -60,14 +64,32 @@ func (f *full) countBallot(logger log.Log, ballot *ballotInfo) {
 	}
 }
 
+func (f *full) countForLateBlock(block *blockInfo) {
+	// ballots are always added to the state after blocks that are
+	// explicitly referenced in the ballot.
```
countvonzero, line 69: this many not be true in the near future with delayed fetching, right?

tortoise/tortoise.go
```diff
@@ -695,28 +626,43 @@ func (t *turtle) onHareOutput(lid types.LayerID, bid types.BlockID) {
...
+	}
+	if _, exist := epoch.atxs[atx.ID]; !exist {
+		t.logger.With().Debug("on atx",
+			log.Stringer("id", atx.ID),
+			log.Uint32("epoch", uint32(atx.TargetEpoch())),
+			log.Uint64("weight", atx.GetWeight()),
+		)
+		epoch.atxs[atx.ID] = atx.GetWeight()
+		epoch.weight += atx.GetWeight()
+		if atx.TargetEpoch() == t.last.GetEpoch() {
```
countvonzero: what happens when atx.TargetEpoch < t.last.GetEpoch()? in that case, shouldn't all ballots that references these ATXs added to uncounted good weight for their layers?

tortoise/tortoise.go
```diff
@@ -387,16 +369,16 @@ func (t *turtle) encodeVotes(
...
+	vote, reason := getLocalVote(verified, current, t.Config, block)
 	if !(vote == abstain && reason == reasonValidity) {
 		return vote, reason, nil
 	}
 	vote = sign(block.margin.Cmp(t.localThreshold))
 	if vote != abstain {
 		return vote, reasonLocalThreshold, nil
 	}
-	coin, err := layers.GetWeakCoin(t.cdb, t.last)
+	coin, err := layers.GetWeakCoin(t.cdb, current.Sub(1))
```
countvonzero, line 350: i don't understand how this closes https://github.com/spacemeshos/go-spacemesh/issues/2920

shouldn't this be `coin, err := layers.GetWeakCoin(t.cdb, block.layer.Sub(1))`?

datastore/store.go
dshulyak: i am using ActivationHeader in OnAtx. so i added Header() method, can change that if there is something better

tortoise/algorithm.go
dshulyak: there is only one goroutine in tortoise that loads state in the background
but in general it doesn't make sense if tortoise failed to load state

tortoise/full.go
dshulyak: in case of delayed fetching - block identifier will be added together with a first ballot that references it

datastore/store.go
countvonzero: you can simply call `cdb.GetAtxHeader`

tortoise/algorithm.go
countvonzero: i was thinking in the future we may want to pass eg from the node.go. but yeah, it doesn't make sense for the whole app to continue if tortoise doesn't load.

tortoise/tortoise.go
dshulyak: current is a layer when ballot is created, so it uses weak coin from that was collected right before

i am not aware how block is related to weak coin, is it? or that was a ballot/block typo?

tortoise/tortoise.go
dshulyak: on atx is always called before ballot that reference it is added to the state, after on atx is called - expected weight and local threshold will change. good uncounted weight will be updated after ballot that referenced this atx is counted.

good uncounted and uncounted would be distinct values. uncounted = expected weight - (total - good uncounted).

datastore/store.go
dshulyak: i am using it here https://github.com/spacemeshos/go-spacemesh/blob/68066ef3cc6c7676b5f9b3e5b76bcfe8818f9fe8/activation/handler.go#L449

not sure why it makes sense to call cdb if data is available right there

tortoise/tortoise.go
countvonzero: it is a typo. plz do `:s/block/ballot/` in my msg.

the confusing part for me is that in the call chain of `getFullVote` the variable is called `last`, which was updated with the current layer in the miner pkg. will it make sense to rename them all to current?

tortoise/tortoise.go
countvonzero: oh ok. so when processing ballot, if a ballot's layer is already processed, then the ballot is counted towards good uncounted weight.

tortoise/full.go
countvonzero: right. the struct that contains block id and tick count.

datastore/store.go
countvonzero: it's a bit awkward yes. 
if you can find a way not having to create Header instances in common/activation pkg, i'm fine with whatever.
i only care that future dev don't easily use a API like that and effectively remove the use of the header cache.

tortoise/tortoise.go
dshulyak: updated

datastore/store.go
dshulyak: i will use cdb.GetAtxHeader then

countvonzero (approved):
thanks

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

PR #3633, opened 2022-10-10
Refactor `AwaitLayer` to use contexts

poszu:
## Changes
Refactored `AwaitLayer` to base on Go's contexts instead of channels.

The context returned from `AwaitLayer` is cancelled when it is time for the given layer.

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -283,10 +283,11 @@ func (b *Builder) run(ctx context.Context) {
 	}
 
 	// ensure layer 1 has arrived
+	<-b.layerClock.AwaitLayer(ctx, types.NewLayerID(1)).Done()
 	select {
-	case <-b.layerClock.AwaitLayer(ctx, types.NewLayerID(1)).Done():
 	case <-ctx.Done():
 		return
+	default:
 	}
 
 	b.waitForFirstATX(ctx)
@@ -339,12 +340,13 @@ func (b *Builder) waitForFirstATX(ctx context.Context) bool {
 			log.Stringer("current_epoch", currEpoch),
 			log.Stringer("wait_till", waitTill),
 			log.Stringer("wait_till_epoch", waitTill.GetEpoch()))
+		<-b.layerClock.AwaitLayer(ctx, waitTill).Done()
 		select {
 		case <-ctx.Done():
 			return false
-		case <-b.layerClock.AwaitLayer(ctx, waitTill).Done():
-			wait = window
+		default:
 		}
+		wait = window
 	}
 	timer := time.NewTimer(wait)
 	b.log.WithContext(ctx).With().Info("waiting for the poet window expires",
@@ -407,6 +409,11 @@ func (b *Builder) loop(ctx context.Context) {
 				// other failures are related to in-process software. we may as well panic here
 				currentLayer := b.layerClock.GetCurrentLayer()
 				<-b.layerClock.AwaitLayer(ctx, currentLayer.Add(1)).Done()
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
 			}
 		}
 	}
@@ -574,10 +581,12 @@ func (b *Builder) createAtx(ctx context.Context) (*types.ActivationTx, error) {
 		log.FieldNamed("pub_epoch_first_layer", pubEpoch.FirstLayer()),
 		log.FieldNamed("current_layer", b.layerClock.GetCurrentLayer()),
 	)
+
+	<-b.layerClock.AwaitLayer(ctx, pubEpoch.FirstLayer()).Done()
 	select {
-	case <-b.layerClock.AwaitLayer(ctx, pubEpoch.FirstLayer()).Done():
 	case <-ctx.Done():
 		return nil, fmt.Errorf("failed to wait for publication epoch: %w", ErrStopRequested)
+	default:
 	}
 
 	b.log.Info("publication epoch has arrived!")
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -48,7 +48,7 @@ const defaultPoetRetryInterval = 5 * time.Second
 
 type nipostBuilder interface {
 	updatePoETProver(PoetProvingServiceClient)
-	BuildNIPost(ctx context.Context, challenge *types.Hash32, timeout chan struct{}) (*types.NIPost, error)
+	BuildNIPost(ctx context.Context, challenge *types.Hash32) (*types.NIPost, error)
 }
 
 type atxHandler interface {
@@ -292,7 +292,9 @@ func (b *Builder) run(ctx context.Context) {
 	}
 
 	// ensure layer 1 has arrived
-	if err := b.waitOrStop(ctx, b.layerClock.AwaitLayer(types.NewLayerID(1))); err != nil {
+	select {
+	case <-b.layerClock.AwaitLayer(ctx, types.NewLayerID(1)).Done():
+	case <-ctx.Done():
 		return
 	}
 
@@ -349,7 +351,7 @@ func (b *Builder) waitForFirstATX(ctx context.Context) bool {
 		select {
 		case <-ctx.Done():
 			return false
-		case <-b.layerClock.AwaitLayer(waitTill):
+		case <-b.layerClock.AwaitLayer(ctx, waitTill).Done():
 			wait = window
 		}
 	}
@@ -413,11 +415,7 @@ func (b *Builder) loop(ctx context.Context) {
 			default:
 				// other failures are related to in-process software. we may as well panic here
 				currentLayer := b.layerClock.GetCurrentLayer()
-				select {
-				case <-ctx.Done():
-					return
-				case <-b.layerClock.AwaitLayer(currentLayer.Add(1)):
-				}
+				<-b.layerClock.AwaitLayer(ctx, currentLayer.Add(1)).Done()
 			}
 		}
 	}
@@ -542,7 +540,7 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 	select {
 	case <-atxReceived:
 		logger.With().Info(fmt.Sprintf("received atx in db %v", atx.ID().ShortString()), atx.ID())
-	case <-b.layerClock.AwaitLayer((atx.TargetEpoch() + 1).FirstLayer()):
+	case <-b.layerClock.AwaitLayer(ctx, (atx.TargetEpoch() + 1).FirstLayer()).Done():
 		select {
 		case <-atxReceived:
 			logger.With().Info(fmt.Sprintf("received atx in db %v (in the last moment)", atx.ID().ShortString()), atx.ID())
@@ -571,11 +569,11 @@ func (b *Builder) createAtx(ctx context.Context) (*types.ActivationTx, error) {
 
 	// the following method waits for a PoET proof, which should take ~1 epoch
 	expireLayer := (pubEpoch + 2).FirstLayer()
-	atxExpired := b.layerClock.AwaitLayer(expireLayer) // this fires when the target epoch is over
+	atxExpiredContext := b.layerClock.AwaitLayer(ctx, expireLayer) // this fires when the target epoch is over
 
 	b.log.With().Info("building NIPost", log.Stringer("pub_epoch", pubEpoch), log.Stringer("expire_layer", expireLayer))
 
-	nipost, err := b.nipostBuilder.BuildNIPost(ctx, hash, atxExpired)
+	nipost, err := b.nipostBuilder.BuildNIPost(atxExpiredContext, hash)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build NIPost: %w", err)
 	}
@@ -585,9 +583,12 @@ func (b *Builder) createAtx(ctx context.Context) (*types.ActivationTx, error) {
 		log.FieldNamed("pub_epoch_first_layer", pubEpoch.FirstLayer()),
 		log.FieldNamed("current_layer", b.layerClock.GetCurrentLayer()),
 	)
-	if err := b.waitOrStop(ctx, b.layerClock.AwaitLayer(pubEpoch.FirstLayer())); err != nil {
-		return nil, fmt.Errorf("failed to wait for publication epoch: %w", err)
+	select {
+	case <-b.layerClock.AwaitLayer(ctx, pubEpoch.FirstLayer()).Done():
+	case <-ctx.Done():
+		return nil, fmt.Errorf("failed to wait for publication epoch: %w", ErrStopRequested)
 	}
+
 	b.log.Info("publication epoch has arrived!")
 	if discarded := b.discardChallengeIfStale(); discarded {
 		return nil, fmt.Errorf("%w: atx target epoch has passed during nipost construction", ErrATXChallengeExpired)
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -276,15 +276,6 @@ func (b *Builder) SignAtx(atx *types.ActivationTx) error {
 	return nil
 }
 
-func (b *Builder) waitOrStop(ctx context.Context, ch chan struct{}) error {
-	select {
-	case <-ch:
-		return nil
-	case <-ctx.Done():
-		return ErrStopRequested
-	}
-}
-
 func (b *Builder) run(ctx context.Context) {
 	if err := b.generateProof(); err != nil {
 		b.log.Error("Failed to generate proof: %v", err)
--- a/blocks/certifier.go
+++ b/blocks/certifier.go
@@ -156,7 +156,7 @@ func (c *Certifier) isShuttingDown() bool {
 func (c *Certifier) run() error {
 	for layer := c.layerClock.GetCurrentLayer(); ; layer = layer.Add(1) {
 		select {
-		case <-c.layerClock.AwaitLayer(layer):
+		case <-c.layerClock.AwaitLayer(c.ctx, layer).Done():
 			c.prune()
 		case <-c.ctx.Done():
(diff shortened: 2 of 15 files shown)


hare/flows_test.go
```diff
@@ -217,21 +217,20 @@ func (m *mockClock) LayerToTime(layer types.LayerID) time.Time {
...
-		return ch
+		cancel()
+		return ctx
 	}
-	ch := make(chan struct{})
-	m.channels[layer] = ch
-	return ch
+	cancels := m.layersSubscriptions[layer]
+	cancels = append(cancels, cancel)
+	m.layersSubscriptions[layer] = cancels
```
fasmat: I think this can probably be shortened to:

```suggestion
	m.layersSubscriptions[layer] = append(m.layersSubscriptions[layer], cancel)
```

timesync/ticker.go
```diff
@@ -179,22 +182,28 @@ func init() {
 	close(closedChan)
 }
 
-// AwaitLayer returns a channel that will be signaled when layer id layerID was ticked by the clock, or if this layer has passed
-// while sleeping. it does so by closing the returned channel.
-func (t *Ticker) AwaitLayer(layerID types.LayerID) chan struct{} {
+// AwaitLayer returns a context that will be canceled when layer with given id is ticked by the clock,
+// or if this layer has passed while sleeping.
+func (t *Ticker) AwaitLayer(ctx context.Context, layerID types.LayerID) context.Context {
```
fasmat, line 187: I think this method could just be

```golang
func (t *Ticker) AwaitLayer(ctx context.Context, layerID types.LayerID) context.Context {
	t.mu.Lock()
	defer t.mu.Unlock()

	layerCtx, cancel := context.WithDeadline(t.LayerToTime(layerID))
        t.layersSubscriptions[layerID] = append(t.layersSubscriptions[layerID], cancel)
	return layerCtx
}

**Review by fasmat — APPROVED** [review]
Looks good!

Just added two minor suggestions that might further simplify the code

File: timesync/ticker.go (hunk 2 above)
Comment (fasmat) [reply timesync/ticker.go:187]: On second thought, this could possibly break things.

At the moment `Notify` needs to be called explicitly for the listeners to be made aware that a layer has been reached. `Notify` can be called late so listeners can be notified late, this might be required.

`Notify` errors if it is called too late, but still signals to all listeners that their layer has been reached. This also means that the existing and the changed code in this PR behave possibly in an unintuitive way:

- Assume `LayerToTime` for layer `L` returns time `T`
- All go routines calling `AwaitLayer` before T receive a `context` that is not cancelled (or a channel that is not closed)
- All go routines calling `AwaitLayer` after T receive a `context` that is cancelled (or a channel that is closed)
- The early callers receive their signal only after `Notify` is called, which is expected to be up to `sendTickThreshold` after `T`; it can be even later then that, and then only `Notify` errors

**Comment by poszu** [comment]
After giving it some thought I begin to have doubts if it is a good idea to signal an event by cancelling a context. It is impossible to distinguish which of the two actually happened: the layer event or parent context cancellation.

File: activation/activation.go [hunk 3]
```diff
@@ -276,23 +276,16 @@ func (b *Builder) SignAtx(atx *types.ActivationTx) error {
...
 		b.log.Error("Failed to generate proof: %v", err)
 		return
 	}
 
 	// ensure layer 1 has arrived
-	if err := b.waitOrStop(ctx, b.layerClock.AwaitLayer(types.NewLayerID(1))); err != nil {
+	select {
+	case <-b.layerClock.AwaitLayer(ctx, types.NewLayerID(1)).Done():
+	case <-ctx.Done():
 		return
```
fasmat:
```golang
<-b.layerClock.AwaitLayer(ctx, types.NewLayerID(1)).Done()
select {
	case <-ctx.Done():
		return
	default:
}
```

If the parent context was cancelled it returns, otherwise it waits until the layer is reached; the context from `AwaitLayer` returns on `Done()` when the parent context was cancelled before the layer was reached. So it will return before waiting for the layer if signalled to do so.

activation/activation.go
```diff
@@ -349,7 +342,7 @@ func (b *Builder) waitForFirstATX(ctx context.Context) bool {
 		select {
 		case <-ctx.Done():
 			return false
-		case <-b.layerClock.AwaitLayer(waitTill):
+		case <-b.layerClock.AwaitLayer(ctx, waitTill).Done():
 			wait = window
 		}
```
fasmat:
```golang
<-b.layerClock.AwaitLayer(ctx, waitTill).Done()
select {
	case <-ctx.Done():
		return
	default:
}
wait = window
```

activation/activation.go
```diff
@@ -413,11 +406,7 @@ func (b *Builder) loop(ctx context.Context) {
 			default:
 				// other failures are related to in-process software. we may as well panic here
 				currentLayer := b.layerClock.GetCurrentLayer()
-				select {
-				case <-ctx.Done():
-					return
-				case <-b.layerClock.AwaitLayer(currentLayer.Add(1)):
-				}
+				<-b.layerClock.AwaitLayer(ctx, currentLayer.Add(1)).Done()
```
fasmat:
```golang
<-b.layerClock.AwaitLayer(ctx, currentLayer.Add(1)).Done()
select {
	case <-ctx.Done():
		return
	default:
}
```

activation/activation.go
```diff
@@ -585,9 +574,12 @@ func (b *Builder) createAtx(ctx context.Context) (*types.ActivationTx, error) {
...
 		log.FieldNamed("current_layer", b.layerClock.GetCurrentLayer()),
 	)
-	if err := b.waitOrStop(ctx, b.layerClock.AwaitLayer(pubEpoch.FirstLayer())); err != nil {
-		return nil, fmt.Errorf("failed to wait for publication epoch: %w", err)
+	select {
+	case <-b.layerClock.AwaitLayer(ctx, pubEpoch.FirstLayer()).Done():
+	case <-ctx.Done():
+		return nil, fmt.Errorf("failed to wait for publication epoch: %w", ErrStopRequested)
 	}
+
```
fasmat:
```golang
<-b.layerClock.AwaitLayer(ctx, pubEpoch.FirstLayer()).Done()
select {
	case <-ctx.Done():
		return nil, fmt.Errorf("failed to wait for publication epoch: %w", ErrStopRequested)
	default:
}
```

blocks/certifier.go
```diff
@@ -156,7 +156,7 @@ func (c *Certifier) isShuttingDown() bool {
 func (c *Certifier) run() error {
 	for layer := c.layerClock.GetCurrentLayer(); ; layer = layer.Add(1) {
 		select {
-		case <-c.layerClock.AwaitLayer(layer):
+		case <-c.layerClock.AwaitLayer(c.ctx, layer).Done():
 			c.prune()
 		case <-c.ctx.Done():
 			return fmt.Errorf("context done: %w", c.ctx.Err())
```
fasmat:
```golang
<-c.layerClock.AwaitLayer(c.ctx, layer).Done()
select {
	case <-ctx.Done():
		return fmt.Errorf("context done: %w", c.ctx.Err())
	default:
}
c.prune()
```

fasmat (commented):
I think the change to only pass one context to `BuildNIPost` is a valuable change. Sometimes in our code it doesn't even matter if the timeout happened or the parent context was cancelled.

It's only a few places where it makes a difference whether the parent or the child context was cancelled and it can be handled without adding complexity. I added suggested changes to those places.

The reason why I prefer one context over context + channel: the caller of `BuildNIPost` has full control over how to handle the situation without the need for `BuildNIPost` to return typed errors to indicate to the caller how it should handle the situation.

timesync/ticker.go
fasmat: In my opinion there are 2 possible ways to fix this issue:

- **small change**: `AwaitLayer` doesn't check `t.LayerToTime` to determine if the returned `context` should already be cancelled or not, but rather `t.lastTickedLayer`.
- **possibly bigger change**: Remove the canceling of contexts in `Notify` and use a much simpler implementation of `AwaitLayer`. `Notify` then only notifies the subscribers who called `Subscribe`, while listeners of `AwaitLayer` "handle themselves" with this implementation:

```golang
func (t *Ticker) AwaitLayer(ctx context.Context, layerID types.LayerID) (context.Context, context.CancelFunc) {
	return context.WithDeadline(ctx, t.LayerToTime(layerID))
}
```

The relevant code part in `Notify` that imo can be removed with my suggested change is the following:

```golang
	for l := t.lastTickedLayer; !l.After(layer); l = l.Add(1) {
		if cancels, found := t.layersSubscriptions[l]; found {
			for _, cancel := range cancels {
				cancel()
			}
			delete(t.layersSubscriptions, l)
		}
	}
```

poszu: @countvonzero, @dshulyak What do you think?

countvonzero: @poszu @fasmat 
can either of you elaborate in the PR description what was wrong with the original channel approach? thanks

fasmat: The original approach has the disadvantage of a function to need to listen for 2 signals to check if it should stop or not: the context and the channel that was returned by `AwaitLayer`.

Since `AwaitLayer` sets a deadline for waiting it is effectively the same as the following line of code:

```golang
ctx, _ := context.WithDeadline(ctx, t.LayerTime(layerID))
```

This achieves the same effect (sending a signal at a known time in the future). As the current implementations of `AwaitLayer`, `Notify` and `startClock()` combined, at least in terms of signalling when a layer was reached.

countvonzero: still don't understand. 

`AwaitLayer` is used in many places. the pattern in the code code either waits for a layer to arrive (in which case the channel unblocks) or waits for the app to shutdown (`context.Done()` unblocks). what is the advantage combining the two?

the disadvantage for me is that i wouldn't know whether it's the layer time arrives or the app shutdown. the former i need to do something, the latter i need to stop.

activation/activation.go
```diff
@@ -627,15 +630,15 @@ func (b *Builder) createAtx(ctx context.Context) (*types.ActivationTx, error) {
...
-	atxExpired := b.layerClock.AwaitLayer(expireLayer) // this fires when the target epoch is over
+	atxExpiredContext := b.layerClock.AwaitLayer(ctx, expireLayer) // this fires when the target epoch is over
 	commitmentAtx, err := b.getCommitmentAtx(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("getting commitment atx failed: %w", err)
 	}
 
 	b.log.With().Info("building NIPost", log.Stringer("pub_epoch", pubEpoch), log.Stringer("expire_layer", expireLayer))
 
-	nipost, err := b.nipostBuilder.BuildNIPost(ctx, hash, *commitmentAtx, atxExpired)
```
countvonzero, line 638: this ctx is the app's top level context. it is used throughout the app goroutines to listen for app shutdown.
should not combine this with whether an ATX should expire.

countvonzero (changes requested):

activation/activation.go
countvonzero: in this specific case, the code return immediately in both cases. but this is not a general case. what is so wrong with waiting for two signals (layer arrival / shutdown) separately?

fasmat: >this ctx is the app's top level context. it is used throughout the app goroutines to listen for app shutdown.

`AwaitLayer` creates a new context with an additional constraint to indicate when the layer has been reached. This new context can now be used in `BuildNIPost` to signal when it is supposed to stop creating NIPosts. The original `ctx` without this constraint is still available and can continued to be used.

Contexts in an application naturally form a tree:

image

Cancelling a parent context cancels all child contexts but not vice versa! The root context is untouched and is continued to be used by other routines that only listen for app shutdown.

fasmat: > in this specific case, the code return immediately in both cases. but this is not a general case. what is so wrong with waiting for two signals (layer arrival / shutdown) separately?

Contexts are all about specific cases:

```golang
func LongRunningFunc(rootCtx context.Context) {
	child1Ctx, cancelFunc1 := context.WithCancel(rootCtx)
        go someFunctionThatNeedsToBeCancelled(child1Ctx)

	// some logic that calls cancelFunc1

        child2Ctx, cancelFunc2 := context.WithTimeout(rootCtx, 1 * time.Second)
	defer cancelFunc2()
	someFunctionThatMustReturnQuickly(child2Ctx)
}
```

This way if `rootCtx` gets canceled `someFunctionThatMustReturnQuickly` and `someFunctionThatNeedsToBeCancelled` receive this signal too without the need to pass in `rootCtx`

EDIT: here `AwaitLayer` basically serves as `context.WithDeadline` and imo could even be replaced completely by `context.WithDeadline(ctx, b.layerClock.LayerToTime(expireLayer)`.

EDIT2: If the function should behave differently based on if the rootContext was canceled or the timeout was reached then yes, the deadline needs to be passed in explicitly (either via a channel that closes, a second context, or just the time of the deadline).

countvonzero: thanks for the illustration :P

i agree with you that context is all about specific case. but i also think it makes the code more complex than it needs to be at the moment.

my rationale:
- major: AwaitLayer is used in many places. in some the new code now insist to finish waiting for layers before the app can exit. this is the main reason i block this PR. for mainnet, the layer duration is set to be 5 min. we don't want the app shutdown to hang for that long.

- moderate: i think this should be dealt with in a case-by-case manner in each component. the original interface uses channel, which is very simple. and it's up to the component to determine its action. 
take this for example https://github.com/spacemeshos/go-spacemesh/blob/develop/activation/activation.go#L648-L652
```
select {
	case <-ctx.Done():
		return nil, fmt.Errorf("failed to wait for publication epoch: %w", err)
	case <-b.layerClock.AwaitLayer(pubEpoch.FirstLayer()):
}
```
can be rewritten to 
```
select {
	case <-ctx.Done():
		return nil, fmt.Errorf("failed to wait for publication epoch: %w", err)
	case <-b.layerClock.AwaitLayer(pubEpoch.FirstLayer()):
	    if ctx.Error() != nil {
	        return nil, fmt.Errorf("context error: %w", ctx.Error)
	    }
}
```
to differentiate the two. 

in the case where you handle them almost the same
```
	select {
	case poetProofRef = <-nb.poetDB.SubscribeToProofRef(nb.state.PoetServiceID, nb.state.PoetRound.ID):
	case <-atxExpired:
		nb.poetDB.UnsubscribeFromProofRef(nb.state.PoetServiceID, nb.state.PoetRound.ID)
		return nil, fmt.Errorf("%w: while waiting for poet proof, target epoch ended", ErrATXChallengeExpired)
	case <-ctx.Done():
		return nil, ErrStopRequested
	}
```
i simply cannot see why this code is offensive. if all 3 channels unblocks at the same time, the first case may add an extra check for ctx.Error(). but even if it doesn't, it will stop at some later check.

- minor: for a reader trying to understand the code, tracing exactly where context is updated is tedious. i always feel using a separate context for p2p request timeout is a cleaner separation for app logic error handling.

timesync/ticker.go
fasmat: > what is the advantage combining the two?

So that `AwaitLayer` stops waiting for a new layer when the app shuts down. There is a problem with this code:

```golang
select {
case <-AwaitLayer:
	// continue
case <-ctx.Done:
	// stop working and return
}
```

If both events happen at roughly the same time `select` will choose a random case. So if the app shuts down at around the time a new layer happens this select has a 50% chance of not being signalled that it is supposed to stop working and continues instead. With the change this now looks like this:

```golang
<-AwaitLayer(ctx)
select {
case <-ctx.Done():
	// stop working and return
default:
}
```

This way if both events happen the routine is guaranteed to shut down. This also doesn't make the application wait longer since the context returned by `AwaitLayer` is a child of the parent context and will also be `Done` when `ctx` is done.

countvonzero: > So that AwaitLayer stops waiting for a new layer when the app shuts down.

where does it happen? can you point me to the code where AwaitLayer is unblocked if the app shuts down?

activation/activation.go
fasmat: > AwaitLayer is used in many places. in some the new code now insist to finish waiting for layers before the app can exit.

This is not the case. `AwaitLayer` returns when the parent context is canceled. The context signals to `AwaitLayer` that it should stop waiting. These two are functionally identical:

`AwaitLayer` without context:
```golang
select {
	case <-ctx.Done():
		return nil, fmt.Errorf("failed to wait for publication epoch: %w", err)
	case <-b.layerClock.AwaitLayer(pubEpoch.FirstLayer()):
	    if ctx.Error() != nil {
	        return nil, fmt.Errorf("context error: %w", ctx.Error)
	    }
}
```

`AwaitLayer` with context:
```golang
<-b.layerClock.AwaitLayer(ctx, pubEpoch.FirstLayer())
select {
	case <-ctx.Done():
		return nil, fmt.Errorf("failed to wait for publication epoch: %w", err)
	default:
}
```

countvonzero: > This is not the case. AwaitLayer returns when the parent context is canceled.

i stand corrected. you are right.
but i still think this code is not better than 

```
select {
	case <-ctx.Done():
		return nil, fmt.Errorf("failed to wait for publication epoch: %w", err)
	case <-b.layerClock.AwaitLayer(pubEpoch.FirstLayer()):
	    if ctx.Error() != nil {
	        return nil, fmt.Errorf("context error: %w", ctx.Error)
	    }
}
```
and in this case, you can keep `AwaitLayer()` simple with channel, and not using context that dev can potentially misuse.

timesync/ticker.go
poszu: It's not unblocked when the app shuts down. Still, both events can happen at the roughly same time. Then `select` will choose randomly.

PR #3639, opened 2022-10-11
[Merged by Bors] - Rename make target golangci-lint to just lint

fasmat:
## Motivation
Before `golangcli-lint` was used `make lint` called a different tool for linting. `make golangci-lint` was added - I assume - temporarily to use both tools. Since we retired the previous tool we should also go back to `make lint`

## Changes
Only renaming `make` targets.

## TODO

## DevOps Notes


poszu (approved):

fasmat:
bors merge

PR #3647, opened 2022-10-12
[Merged by Bors] - tortoise: compute opinion hash

### LINKED ISSUE #3635: compute ballot opinion hash from received votes
opinion hash will be used to verify that signed opinion hash matches the hash computed from received votes.  for more information see https://community.spacemesh.io/t/mutable-ballots/303

## implementation

when votes are decoded compute an opinion hash from ballot votes, following next algorithm (using sha256 as a hash):
- layer hash is one of:  
  - concatenation of supported blocks (sorted by tick height asc and lex order asc)  
  - empty buffer for empty layer  
  - buffer with length of 1 byte if ballot votes abstain on the layer  
 
- opinion hash (or aggregated layer hash) defined recursively:  
  - for layer immediately after genesis an opinion hash is a layer hash  
  - otherwise an opinion hash is a hash from a concatenation of previous opinion hash and current layer hash

### affected code

- add additional types.Hash32 to layerVote  
https://github.com/spacemeshos/go-spacemesh/blob/ff1989387ad2e1f7f49906fe2159d5342addaa06/tortoise/state.go#L213-L219  
- overwriting previously computed hashes  
https://github.com/spacemeshos/go-spacemesh/blob/ff1989387ad2e1f7f49906fe2159d5342addaa06/tortoise/state.go#L244-L269  
- computing hashes on top of base ballot hash. cleaner approach would merge this and the previous operation
https://github.com/spacemeshos/go-spacemesh/blob/ff1989387ad2e1f7f49906fe2159d5342addaa06/tortoise/tortoise.go#L872-L889

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3635


tortoise/state_test.go
```diff
@@ -90,3 +91,100 @@ func TestVotesUpdate(t *testing.T) {
 		}
 	})
 }
+
+func TestComputeOpinion(t *testing.T) {
+	t.Run("single supported sorted", func(t *testing.T) {
+		v := votes{}
+		blocks := []*blockInfo{
```
poszu, line 98: These tests could run in parallel. How about adding `t.Parallel()` in the test and in each `t.Run(...)`?

poszu (approved):

hash/hash.go
```diff
@@ -9,6 +11,9 @@ const (
 	Size = sha256.Size
 )
 
+// Hash is an alias to stdlib hash.Hash interface.
+type Hash = hash.Hash
```
countvonzero: is this used?

tortoise/state_test.go
dshulyak: makes sense, i think pretty much all tests in tortoise module can be run in parallel, i will switch all of them in a separate pr

hash/hash.go
dshulyak: thanks, ended up not using it

dshulyak:
bors merge

PR #3649, opened 2022-10-13
[Merged by Bors] - Fix tests sharing state

fasmat:
## Motivation
Fix some tests to not share state: specifically tests related to PoST used to share the same directory for their session, leaking data from one test to the next.

## Changes
The change here makes sure that all tests use their own directory and properly cleanup afterwards. `t.TempDir()` creates a unique test directory upon each call that is automatically deleted when a test ends - successful or not.

## Test Plan
Updated tests still pass.

## TODO

## DevOps Notes

### DIFF
--- a/filesystem/dirs_test.go
+++ b/filesystem/dirs_test.go
@@ -14,21 +14,19 @@ import (
 var RootFolder = "/"
 
 func TestPathExists(t *testing.T) {
-	tempFile := filepath.Join(os.TempDir(), "testdir"+uuid.New().String()+"_"+t.Name())
+	tempFile := filepath.Join(t.TempDir(), uuid.New().String()+"_"+t.Name())
 	_, err := os.Create(tempFile)
 	require.NoError(t, err, "couldn't create temp path")
 	assert.True(t, PathExists(tempFile), "expecting existence of path")
-	os.RemoveAll(tempFile)
 }
 
 func TestGetFullDirectoryPath(t *testing.T) {
-	tempFile := filepath.Join(os.TempDir(), "testdir"+uuid.New().String()+"_"+t.Name())
+	tempFile := filepath.Join(t.TempDir(), "testdir"+uuid.New().String()+"_"+t.Name())
 	err := os.MkdirAll(tempFile, os.ModeDir)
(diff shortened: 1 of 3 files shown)


countvonzero (approved):

fasmat:
bors merge

filesystem/dirs_test.go
```diff
@@ -14,21 +14,19 @@ import (
...
+	tempFile := filepath.Join(t.TempDir(), uuid.New().String()+"_"+t.Name())
 	_, err := os.Create(tempFile)
 	require.NoError(t, err, "couldn't create temp path")
 	assert.True(t, PathExists(tempFile), "expecting existence of path")
-	os.RemoveAll(tempFile)
 }
 
 func TestGetFullDirectoryPath(t *testing.T) {
-	tempFile := filepath.Join(os.TempDir(), "testdir"+uuid.New().String()+"_"+t.Name())
+	tempFile := filepath.Join(t.TempDir(), "testdir"+uuid.New().String()+"_"+t.Name())
```
poszu, line 24: I don't get why this has to be that complicated. It could be just:
```suggestion
	tempFile := filepath.Join(t.TempDir(), "test_file")
```

Btw, looking at the test it's not actually treated as a path to a **file**, but a directory. So:
```suggestion
	directory := filepath.Join(t.TempDir(), "test_directory")
```

filesystem/dirs_test.go
```diff
@@ -14,21 +14,19 @@ import (
...
 	err := os.MkdirAll(tempFile, os.ModeDir)
```
poszu, line 25: Considering that the tested function _creates_ the directory if it doesn't exist:
>// GetFullDirectoryPath gets the OS specific full path for a named directory.
>// The directory is created if it doesn't exist.

It would be better to skip creating it in the test and verify if the function created it as it promises.

Btw, this function's name doesn't match its responsibilities.

poszu (changes requested):

fasmat:
@poszu: I "fixed" the `filesystem` package here: https://github.com/spacemeshos/go-spacemesh/pull/3652

PR #3651, opened 2022-10-13
[Merged by Bors] - Cache deps in systest Dockerfile

poszu:
## Motivation
It should speed up docker image rebuilds when dependencies are the same as before.

## DevOps Notes


fasmat:
That seems a bit redundant to me, since the two lines below the lines you added already mount the build cache of the host overwriting the files created by `go mod download`.

fasmat (changes requested):
See comment

poszu:
> That seems a bit redundant to me, since the two lines below the lines you added already mount the build cache of the host overwriting the files created by `go mod download`.

They don't (`build` directory is .dockerignore-d) and they shouldn't. Docker can efficiently cache layers and knows when it can reuse a layer (when go.mod/sum has not changed).

fasmat:
> They don't (build directory is .dockerignore-d) and they shouldn't. Docker can efficiently cache layers and knows when it can reuse a layer (when go.mod/sum has not changed).

The following two lines already cache the go build: even if the `build` directory is ignored.

```Dockerfile
RUN --mount=type=cache,target=/go/pkg/mod go mod download
RUN --mount=type=cache,target=/root/.cache/go-build go test -v -c -o /build/tests.test ./systest/tests/
```

I am not opposed to your change, but then you should also update the two lines to be just:

```Dockerfile
RUN go mod download
RUN go test -v -c -o /build/tests.test ./systest/tests/
```

poszu:
@fasmat Fixed.

fasmat (approved):

poszu:
bors merge

poszu:
bors merge

PR #3652, opened 2022-10-13
[Merged by Bors] - Remove filesystem package

fasmat:
## Motivation
We don't need this package, all the functionality it provides is given by the standard library of go.

## Changes
- `ExistOrCreate`: no need for this function, `os.MkdirAll` does the same thing - creates a directory or no-op if the dir already exists
- `GetUserHomeDirectory`: is the same as `os.UserHomeDir`, the later works on all platforms supported by go, not only just macOS, windows and linux
- `GetCanonicalPath`: is the same as `filepath.Clean` with the additon of replacing `~` by the users home directory. This should arguably NOT be done by an application. `~` is only an alias of the users home directory in some unix shells. On windows most shells now replace `~` with `C:\Users\myUser` - on the old command line `%userprofile%` needs to be used instead. See also https://github.com/golang/go/issues/4140#issuecomment-66070561

## TODO

## DevOps Notes


poszu (approved):

fasmat:
bors merge

PR #3653, opened 2022-10-13
[Merged by Bors] - Improve caching in Dockerfile

poszu:
## Motivation
`make dockerbuild-go` takes an insanely long time every time a file changes. It downloads libs from Internet and rebuilds everything **from scratch** every time.

This change cut rebuild after a single-line-of-code-change from 5m to 30s on my machine.

## Changes
- cache library (`make get-libs`) in  a separate docker image layer to avoid redownloading
- mount a build cache to avoid rebuilding from scratch every time
- build docker image with build kit to be able to use `--mount`

## DevOps Notes


poszu:
bors try

Makefile
```diff
@@ -113,7 +113,7 @@ endif
 # available only for linux host because CGO usage
 ifeq ($(HOST_OS),linux)
 docker-local-build: go-spacemesh hare p2p harness
-	cd build; docker build -f ../Dockerfile.prebuiltBinary -t $(DOCKER_IMAGE) .
+	cd build; DOCKER_BUILDKIT=1 docker build -f ../Dockerfile.prebuiltBinary -t $(DOCKER_IMAGE) .
```
fasmat, line 116: I think this is not required, `DOCKER_BUILDKIT=1` should be the default on any docker installation after Oct 2020.

pigmej: I'm not so sure. I'm kinda semi-sure that it's still behind a flag. Its not anymore experimental flag though.

At least that's what docs say :)

poszu: It complained on my machine without it, so I guess it's not a default.

fasmat: I was going by this blog post: https://pythonspeed.com/articles/docker-buildkit/#:~:text=Enabling%20BuildKit%20in,Docker%20Engine.

Also on my machine with and without the Environment variable I notice little to no difference 🤷

poszu: This blog post says it's ON by default for Mac and Windows and OFF for Linux. 

Anyway, IMHO setting it doesn't hurt at all.

fasmat (approved):

Makefile
pigmej: It's actually even worse. It's set only when using docker desktop app :)

And you can set docker in milion other ways that using docker desktop.

I think setting it is a must have (if it works) :).

poszu:
bors merge

PR #3658, opened 2022-10-14
[Merged by Bors] - tortoise, ballot: sign and validate opinion hash

### LINKED ISSUE #3636: sign opinion hash and validate that opinion hash matches received votes
depends on #3635 

in order to enable ballot mutability smesher will sign opinion hash instead of the votes. see https://community.spacemesh.io/t/mutable-ballots/303 for more information.

## implementation

- add new type Opinion. Opinion type should be encoded when Ballot is gossiped. The reason to encode full Opinion and not just Votes part for gossip is so that receiver can verify signature before doing everything else. Otherwise signature verification has to be moved to the last step, after tortoise decodes votes and computes opinion hash.

```go
type Opinion struct {
  Hash types.Hash32
  Votes Votes
}
```

- tortoise.EncodeVotes should return Opinion instead of Votes.
- field `Votes` should not be signed
add new type signedBallot that will be a complete copy of type Ballot, but without added Votes field
- change tortoise.OnBallot to return error and OnBallot should be called before storing ballot
verify that computed hash matches opinion.Hash, if it doesn't raise ErrValidation with a message that hash mismatches

### affected code
- encoding votes
https://github.com/spacemeshos/go-spacemesh/blob/ff1989387ad2e1f7f49906fe2159d5342addaa06/tortoise/tortoise.go#L323-L379
- ballot type
https://github.com/spacemeshos/go-spacemesh/blob/ff1989387ad2e1f7f49906fe2159d5342addaa06/common/types/ballot.go#L62
- signing in the proposal builder
https://github.com/spacemeshos/go-spacemesh/blob/develop/miner/proposal_builder.go#L244
- OnBallot call
https://github.com/spacemeshos/go-spacemesh/blob/develop/mesh/mesh.go#L646

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3636

- add opinion hash to ballot type, it replaces votes in the signed payload. it allows to re-encode votes after initial container was received to reduce the size of that container, or to prune ballots that deemed adversarial. votes container is still broadcasted and synced, but not included into the signature and id
- bulk of this change is using splitting tortoise.OnBallot into tortoise.DecodeBallot and tortoise.StoreBallot . decoding votes is expensive, so DecodeBallot caches decoded votes and stores them when StoreBallot is called. StoreBallot is conditional on other validation performed in proposals handler.
- proposals/util.go was moved to proposals/util/util.go to disentangle circular dependency between tortoise and proposals module. proposals/util_test.go wasn't moved as it uses same utils for testing. I will create a separate test suite in a follow up


common/types/ballot.go
```diff
@@ -59,7 +61,10 @@ type InnerBallot struct {
 	// the proof of the smesher's eligibility to vote and propose block content in this epoch.
 	// Eligibilities must be produced in the ascending order.
 	EligibilityProofs []VotingEligibilityProof
-	Votes             Votes
+	// OpinionHash is a aggregated opinion on the latest votable layer.
```
countvonzero: what does `the latest votable layer` mean?
is there a case where the OpinionHash layer isn't equal to `ballot.layer - 1`?

common/types/testutil.go
```diff
@@ -101,14 +101,14 @@ func RandomTransactionID() TransactionID {
 func RandomBallot() *Ballot {
 	return &Ballot{
 		InnerBallot: InnerBallot{
-			AtxID: RandomATXID(),
-			Votes: Votes{
-				Base:    RandomBallotID(),
-				Support: []BlockID{RandomBlockID(), RandomBlockID()},
-			},
+			AtxID:      RandomATXID(),
```
countvonzero, line 98: add hash?

common/types/ballot_test.go
```diff
@@ -39,36 +39,20 @@ func TestBallot_Initialize(t *testing.T) {
 func TestBallot_Initialize_BadSignature(t *testing.T) {
 	b := Ballot{
 		InnerBallot: InnerBallot{
-			AtxID: RandomATXID(),
-			Votes: Votes{
-				Base:    RandomBallotID(),
-				Support: []BlockID{RandomBlockID(), RandomBlockID()},
-			},
+			AtxID:      RandomATXID(),
```
countvonzero, line 65: add hash?

common/types/ballot.go
```diff
@@ -38,10 +38,12 @@ func (id *BallotID) DecodeScale(d *scale.Decoder) (int, error) {
 
 // Ballot contains the smesher's signed vote on the mesh history.
 type Ballot struct {
-	// the actual votes on the mesh history
+	// InnerBallot is a signed part of the ballot.
 	InnerBallot
 	// smesher's signature on InnerBallot
 	Signature []byte
+	// Votes field is not signed.
+	Votes Votes
```
countvonzero, line 47: my understanding is that we need to clear Votes before calculating ID, as Votes is not signed and is subject to change later.

common/types/ballot.go
```diff
@@ -204,11 +221,21 @@ func (b *Ballot) Bytes() []byte {
 	return data
 }
 
+// SetID from stored data.
+func (b *Ballot) SetID(id BallotID) {
+	b.ballotID = id
```
countvonzero, line 236: `SetID` and `SetSmesherID` allow violation against the immutability after `Initialize()` rule. i thought `DBBallot` was to make sure of that. why remove it?

tortoise/algorithm.go
```diff
@@ -211,6 +211,43 @@ func (t *Tortoise) OnBallot(ballot *types.Ballot) {
...
+
+// CancelWeight zeroes out ballot weight.
+func (d *DecodedBallot) CancelWeight() {
+	d.info.weight = weight{}
+}
+
+// Store adds decoded ballot to the tortoise in-memory state.
+func (d *DecodedBallot) Store() error {
+	d.t.mu.Lock()
+	defer d.t.mu.Unlock()
```
countvonzero: this shared lock seems unsafe. mesh.go have access to a `DecodedBallot` and `Tortoise` too. bad code that can cause Tortoise hanging.

mesh/mesh.go
```diff
@@ -641,45 +641,56 @@ func (msh *Mesh) AddTXsFromProposal(ctx context.Context, layerID types.LayerID,
...
-	if mal {
-		b.SetMalicious()
+	if malicious {
+		ballot.SetMalicious()
 	}
-
-	// it is important to run add ballot and set identity to malicious atomically
-	return msh.cdb.WithTx(context.Background(), func(tx *sql.Tx) error {
-		if err := ballots.Add(tx, b); err != nil && !errors.Is(err, sql.ErrObjectExists) {
+	// balots.Add and ballots.Count should be atomic
```
countvonzero: :s/balots/ballots/

mesh/mesh.go
```diff
@@ -641,45 +641,56 @@ func (msh *Mesh) AddTXsFromProposal(ctx context.Context, layerID types.LayerID,
...
+	if err != nil {
+		return fmt.Errorf("decode ballot %s: %w", ballot.ID(), err)
 	}
-	msh.trtl.OnBallot(ballot)
-	return nil
-}
-
-func (msh *Mesh) addBallot(b *types.Ballot) error {
-	mal, err := identities.IsMalicious(msh.cdb, b.SmesherID().Bytes())
+	if decoded == nil {
```
countvonzero: what are the conditions that cause decoded to return nil here? please also add comment in the interface in system pkg

system/tortoise.go
```diff
@@ -8,9 +8,16 @@ import (
...
+type DecodedBallot interface {
+	Opinion() types.Hash32
+	CancelWeight()
+	Store() error
+}
+
 // Tortoise is an interface provided by tortoise implementation.
 type Tortoise interface {
-	OnBallot(*types.Ballot)
+	DecodeBallot(*types.Ballot) (DecodedBallot, error)
```
countvonzero: this interface (based on the usage in mesh.go) is not common in that DecodedBallot can be nil when error is nil. 
can you add comment to help caller when new code is written?

tortoise/tortoise.go
```diff
@@ -653,7 +649,7 @@ func (t *turtle) onBallot(ballot *types.Ballot) error {
 		t.logger.With().Warning("base ballot not in state",
 			log.Stringer("base", ballot.Votes.Base),
 		)
-		return nil
+		return nil, nil
```
countvonzero, line 702: regarding all the above `return nil, nil` lines.
so if a ballot's base ballot is outside the window, the node will not store it in db (in mesh.go)? 
shouldn't we try harder to look up db instead?

tortoise/tortoise.go
```diff
@@ -680,22 +676,21 @@ func (t *turtle) onBallot(ballot *types.Ballot) error {
...
-			t.logger.With().Warning("invalid ballot used as a reference",
-				log.Stringer("ref", ballot.RefBallot),
-			)
-			return nil
+			return nil, fmt.Errorf("invalid ballot use as a reference %s", ballot.RefBallot)
 		}
 		refinfo = ref.reference
 	}
+
 	if !ballot.IsMalicious() {
```
countvonzero, line 737: how can this be true now that malicious is set after this method returns.

mesh/mesh.go
```diff
@@ -641,45 +641,56 @@ func (msh *Mesh) AddTXsFromProposal(ctx context.Context, layerID types.LayerID,
 
 // AddBallot to the mesh.
 func (msh *Mesh) AddBallot(ballot *types.Ballot) error {
-	if err := msh.addBallot(ballot); err != nil {
-		return err
+	decoded, err := msh.trtl.DecodeBallot(ballot)
```
countvonzero: i really don't think the code should only save the ballot to DB if this call return successfully. it is conditional on tortoise's internal state and will make ballot validation flow inconsistent.

why can't a ballot be saved when it's syntactically correct?

common/types/ballot.go
dshulyak: InnerBallot is signed, the same as before

tortoise/tortoise.go
dshulyak: it doesn't make a difference to find a base ballot without decoded votes, this is why there is sliding window

mesh/mesh.go
dshulyak: opinion hash validation is a part of the validation, received ballots with votes can be broadcasted only if computed opinion from Votes sidecar is consistent with signed opinion

> it is conditional on tortoise's internal state and will make ballot validation flow inconsistent.

btw there should be no inconsistency, decoding votes depends fully on the information provided by the ballot, and previously referenced ballots

tortoise/tortoise.go
dshulyak: this is true when previously stored ballot is loaded into tortoise from db. i decided to leave at as it is, and use DecodeBallot only when DecodeBallot method is called

common/types/ballot.go
dshulyak: DBBallot wasn't used for long time, thats why i removed it in this pr
SetID was consistent with changes to atxs

common/types/ballot.go
dshulyak: ballot should vote on all previous layers, latest layer on which ballot can vote is layer - 1

common/types/testutil.go
dshulyak: what for? not all fields are added here anyway
random hash won't make this struct valid, so not sure what the point

mesh/mesh.go
dshulyak: i also noticed that some validation in the handler is incorrect (all of validation in checkVotesConsistency is incorrect, since it doesn't decode votes, so that can be bypassed easily). i think only tortoise can do such validation properly

dshulyak:
bors try

common/types/ballot.go
countvonzero: yes. but i am talking about ID calculation.

https://github.com/spacemeshos/go-spacemesh/blob/ce3b65c90e73d4d69be78e0eec2b573427cf0068/common/types/ballot.go#L187

the code hashes the whole ballot, including `ballot.Votes`. we should hash data with `ballot.Votes` cleared so that ID doesn't change even tho the votes change.

mesh/mesh.go
countvonzero: > opinion hash validation is a part of the validation.
> all of validation in checkVotesConsistency is incorrect

in that case, why not move the decoding to the gossip handler and update the validation logic there? 
since this is how you determine whether a ballot is syntactically valid or not, it make more sense to do all the necessary decoding there and, only when it's valid, call `mesh.AddBallot()`

tortoise/tortoise.go
countvonzero: suppose we change sliding window via node's config. does that affect a ballot's syntactic validness?

mesh/mesh.go
dshulyak: sorry i dont know how to do that, decoding is quite nuanced and coupled with tortoise implementation
moving it elsewhere from the tortoise would require keeping same stuff in memory as i keep in the tortoise

tortoise/tortoise.go
dshulyak: protocol should define max distance for base ballot, sliding window should be alteast as large as that distance. i guess that should be a part of the config validation

common/types/ballot.go
dshulyak: thank i forgot that change was made. it should encode only inner and signature for that purpose

mesh/mesh.go
dshulyak: it looks like checkVotesConsistency can't be done before signed opinion is compared with computed, will have to figure out something

countvonzero: > sorry i dont know how to do that, decoding is quite nuanced and coupled with tortoise implementation
moving it elsewhere from the tortoise would require keeping same stuff in memory as i keep in the tortoise

i mean to expose system.Tortoise interface to the ballot gossip handler. not to move decode ballot outside of tortoise.

tortoise/tortoise.go
countvonzero: in that case, should the window-size be part of genesis config?

dshulyak: max base distance should be in the genesis config or just a constant protocol parameter, and sliding window should be atleast as large as max base distance

countvonzero (approved):

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

PR #3659, opened 2022-10-14
[Merged by Bors] - Fix race condition in Builder::StartSmeshing

### LINKED ISSUE #3066: `TestBuilder_RestartSmeshing` fails when `-race` flag is enabled
## Description

```
2022-01-19T20:40:16.1885250Z ?   	github.com/spacemeshos/go-spacemesh	[no test files]
2022-01-19T20:41:04.3316844Z --- FAIL: TestBuilder_RestartSmeshing (0.06s)
2022-01-19T20:41:04.3322460Z     activation_test.go:330: 
2022-01-19T20:41:04.3323133Z         	Error Trace:	activation_test.go:330
2022-01-19T20:41:04.3323679Z         	Error:      	Should be false
2022-01-19T20:41:04.3324239Z         	Test:       	TestBuilder_RestartSmeshing
2022-01-19T20:41:04.3324619Z FAIL
2022-01-19T20:41:04.3325256Z FAIL	github.com/spacemeshos/go-spacemesh/activation	37.968s
```

https://github.com/spacemeshos/go-spacemesh/runs/4873660080?check_suite_focus=true
https://github.com/spacemeshos/go-spacemesh/pull/2930

poszu:
## Motivation
There is a race condition in `StartSmeshing()` happening when it is called repeatedly with `StopSmeshing()` (not even concurrently). The race is on the atomic boolean `started`. It is set to false in the spawned goroutine (which is fine) and also in `StopSmeshing()` (which is not fine because it no longer serves as a signalling mechanism that Builder stopped smeshing).

Consider the following:

1. `StartSmeshing()`
   1. `started` is set to true
   2. goroutine is spawned
2. `StopSmeshing()`
   1. `started` is set to false. At this point, one can already restart smeshing
   2. b.stop() - cancel context, the goroutine starts to shutdown
3. `StartSmeshing()`
   1. `started` is set to true
   2. goroutine finishes. It sets `started` back to false. Builder state is broken now. It is smeshing, but `Smeshing()` returns false.

Closes #3066 

## Changes
Removed setting `started = false` in `StopSmeshing()`. Added more tests and fixed existing ones

## Test Plan
Unit tests. Tests now pass with `-race` flag.

## DevOps Notes

### DIFF
--- a/activation/activation_test.go
+++ b/activation/activation_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/spacemeshos/go-spacemesh/activation/mocks"
 	atypes "github.com/spacemeshos/go-spacemesh/activation/types"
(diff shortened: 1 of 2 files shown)


activation/activation_test.go
```diff
@@ -468,16 +468,29 @@ func TestBuilder_RestartSmeshing(t *testing.T) {
...
-		time.Sleep(400 * time.Microsecond)
-		require.Truef(t, builder.Smeshing(), "failed on execution %d", i)
-		require.NoError(t, builder.StopSmeshing(true))
-		time.Sleep(400 * time.Microsecond)
-		require.Falsef(t, builder.Smeshing(), "failed on execution %d", i)
-	}
+	t.Run("Single threaded", func(t *testing.T) {
+		for i := 0; i < 100; i++ {
+			require.NoError(t, builder.StartSmeshing(types.Address{}, atypes.PostSetupOpts{}))
+			require.Never(t, builder.Finished, 400*time.Microsecond, 50*time.Microsecond, "failed on execution %d", i)
```
fasmat: It looks a bit convoluted, but this way you don't need both `builder.Smeshing` and `builder.Finished`:

```suggestion
			require.Never(t, func() bool { return !builder.Smeshing() }, 400*time.Microsecond, 50*time.Microsecond, "failed on execution %d", i)
```

activation/activation_test.go
```diff
@@ -468,16 +468,29 @@ func TestBuilder_RestartSmeshing(t *testing.T) {
...
+			require.Truef(t, builder.Smeshing(), "failed on execution %d", i)
+			require.NoError(t, builder.StopSmeshing(true))
+			require.Eventually(t, builder.Finished, 10*time.Millisecond, time.Millisecond, "failed on execution %d", i)
```
fasmat: Same here:

```suggestion
			require.Eventually(t, func() bool { return !builder.Smeshing() }, 10*time.Millisecond, time.Millisecond, "failed on execution %d", i)
```

activation/activation_test.go
```diff
@@ -468,16 +468,29 @@ func TestBuilder_RestartSmeshing(t *testing.T) {
...
-		require.NoError(t, builder.StartSmeshing(types.Address{}, atypes.PostSetupOpts{}))
-		// NOTE(dshulyak) this is a poor way to test that smeshing started and didn't exit immediately,
-		// but proper test requires adding quite a lot of additional mocking and general refactoring.
-		time.Sleep(400 * time.Microsecond)
-		require.Truef(t, builder.Smeshing(), "failed on execution %d", i)
-		require.NoError(t, builder.StopSmeshing(true))
-		time.Sleep(400 * time.Microsecond)
-		require.Falsef(t, builder.Smeshing(), "failed on execution %d", i)
-	}
+	t.Run("Single threaded", func(t *testing.T) {
```
fasmat, line 475: I think each `t.Run` should have its own instance of `builder` otherwise subsequent tests re-use the already built state by the previous tests.

Arguably `cdb` and `atxHandler` shouldn't be reused between subtests as well

fasmat (commented):
Good find and fix! I just added some suggestions to the tests.

activation/activation_test.go
poszu: I added `Finished()` to make tests more readable :D

activation/activation_test.go
```diff
@@ -468,16 +468,29 @@ func TestBuilder_RestartSmeshing(t *testing.T) {
...
+			require.Eventually(t, builder.Finished, 10*time.Millisecond, time.Millisecond, "failed on execution %d", i)
+		}
+	})
+
+	t.Run("Multi threaded", func(t *testing.T) {
+		// Meant to be run with -race to detect races.
+		// It cannot check `builder.Smeshing()` as Start/Stop is happening from many goroutines simultaneously.
+		// Both Start and Stop can fail as it is not known if builder is smeshing or not.
+		for worker := 0; worker < 10; worker += 1 {
+			go func() {
```
fasmat: Use this in a `errGroup` or with a `WaitGroup` otherwise the test finishes before all workers are finished

fasmat (changes requested):
The last comment is critical to be fixed 😢

fasmat (approved):

poszu:
bors merge

PR #3664, opened 2022-10-17
[Merged by Bors] - conservative cache: reset accounts that only have infeasible txs in db

### LINKED ISSUE #3663: conservative cache: insufficient balance is not re-evaluated after layers are applied
## Description

@brusherru / @dshulyak reported that tx that was first not accepted by mempool due to insufficient funds are not reconsidered after layers are applied (and received rewards / incoming funds), and somehow prevent submitting other transactions even if nonce was incremented.

countvonzero:
## Motivation

Closes #3663

## Changes
- added two tests that failed with the old code
- after a layer is applied, also reset accounts that ONLY have infeasible txs in db, as they may also receive new funds/rewards in the most recent applied layer


txs/cache.go
```diff
@@ -727,13 +732,16 @@ func (c *cache) ApplyLayer(
 		}
 		acctResetDuration.Observe(float64(time.Since(t1)))
 	}
-	for principal := range skippedByPrincipal {
-		if _, ok := byPrincipal[principal]; ok {
+
+	for principal, cash := range c.pending {
```
dshulyak: what is cash? 

i looked up in the code, should it be called account? i can't make any connection to cash

dshulyak (approved):
is there any eviction policy for infeasible transactions from disk and accounts that have only infeasible transactions from that pending map? 

 this must be a problem, if there is no eviction and no limit on the number of accounts

countvonzero:
> is there any eviction policy for infeasible transactions from disk and accounts that have only infeasible transactions from that pending map?
> 
> this must be a problem, if there is no eviction and no limit on the number of accounts

i thought i had previously filed an issue for this but i cannot find it now. in any case, i filed https://github.com/spacemeshos/go-spacemesh/issues/3668 with more detailed description and implementation proposals. please review. thanks

txs/cache.go
countvonzero: don't want to call it cache (clash with the struct name) so use it as it sounds the same. anyhow, renamed to `accCache`

countvonzero:
bors merge

countvonzero:
bors merge

PR #3665, opened 2022-10-17
[Merged by Bors] - Workaround for TimeClock ticking too soon

### LINKED ISSUE #3617: `TimeClock` sometimes ticks too early
## Description
The code in https://github.com/spacemeshos/go-spacemesh/blob/75ea165d6fb450160a36679861ebb5d00d1a0b9f/timesync/clock.go#L58 enters a loop in which it calls https://github.com/spacemeshos/go-spacemesh/blob/75ea165d6fb450160a36679861ebb5d00d1a0b9f/timesync/clock.go#L86 every start of the layer. It is supposed to fit within a 500ms margin from the start time of the layer.

It turns out, that it sometimes ticks too soon (around 1ms early). The check in `Notify()`: https://github.com/spacemeshos/go-spacemesh/blob/3b2941c84f1ef830afd47f4bccbbb74f1bd4dbc7/timesync/ticker.go#L121-L127 treats this situation as a very late tick (still in the previous layer) and aborts. As a result, the tick is missed.

A fragment of logs showing this case:
```
{"L":"ERROR","T":"2022-10-04T04:32:43.999-0400","N":"d4709.clock","M":"skipping tick since we missed the time of the tick by more than the allowed threshold","current_layer":20,"timeSinceStart":"14.999953798s","layerStart":"2022-10-04T04:32:29.000-0400","threshold":"500ms","name":"clock"}
```

```
{"L":"INFO","T":"2022-10-04T04:32:14.089-0400","N":"d4709.clock","M":"clock notifying subscribers of new layer tick","subscriber_count":3,"layer_id":19,"name":"clock"}
{"L":"INFO","T":"2022-10-04T04:32:29.090-0400","N":"d4709.clock","M":"clock notifying subscribers of new layer tick","subscriber_count":3,"layer_id":20,"name":"clock"}
{"L":"INFO","T":"2022-10-04T04:32:43.999-0400","N":"d4709.clock","M":"clock notifying subscribers of new layer tick","subscriber_count":3,"layer_id":20,"name":"clock"}
```

We can see, that it's been logged twice for layer 20. The second one is too early - it should happen in layer 21. Also, mark the timestamp. Each layer is 15s long in these tests. The clock should therefore tick every 15s. The faulty tick should happen at **44**.000+. Instead it happened at **43.999**.

## Steps to reproduce
Run `TestSmeshing` systest. It's reproducible even in local run (minikube) with a small cluster (size=5)

## Actual Behavior
See description

## Expected Behavior
The clock should tick correctly **after** the new layer started.

## Environment
 - OS: Windows, WSL2, minikube

poszu:
Closes #3617 

## Changes
- use raw `time.After(time.Duration)` instead of `time.Ticker`
- retry waiting for the remaining time in case `time.After()` returned too early
- moved a log into `Notify()`, avoiding an extra Lock just for logging

## Test Plan
ut, systests

## DevOps Notes


poszu:
bors try

fasmat:
This PR looks ok to me. Any particular reason as tho why the loop in `startClock` can't rely on `time.Timer`?:

```golang
func (t *TimeClock) startClock() error {
	t.log.Info("starting global clock now=%v genesis=%v %p", t.clock.Now(), t.genesis, t)

	// possibly some delay and initial call to t.Notify() to get in sync with a layer

	ticker := time.NewTicker(layerDuration)
	for {
		select {
		case <-ticker.C:
			// should fire every layerDuration, so if the ticker was created right when a new layer started this should never be too early
		case <-t.stop:
			t.log.Info("stopping global clock %p", t)
			ticker.Stop()
			return nil
		}
		if missed, err := t.Notify(); err != nil {
			t.log.With().Warning("could not notify all subscribers",
				log.Err(err),
				log.Int("missed", missed))
		}
	}
}
```

poszu:
> This PR looks ok to me. Any particular reason as tho why the loop in `startClock` can't rely on `time.Timer`?:

I assume you meant `time.Ticker`.

Yes, primarily the fact that it cannot be scheduled to tick every `duration` starting at a specific **time**, only **every duration**. It would need some extra code to start the timer at the start of a layer to ensure that the `Timer` actually ticks at the beginning of the layer (and not x time after it).

poszu:
> > This PR looks ok to me. Any particular reason as tho why the loop in `startClock` can't rely on `time.Timer`?:
> 
> Yes, primarily the fact that it cannot be scheduled to tick every `duration` starting at a specific **time**, only **every duration**. It would need some extra code to start the timer at the start of a layer to ensure that the `Timer` actually ticks at the beginning of the layer (and not x time after it).

I forgot the second reason. `Timer` can silently drop ticks if

poszu:
> > This PR looks ok to me. Any particular reason as tho why the loop in `startClock` can't rely on `time.Timer`?:
> 
> I assume you meant `time.Ticker`.
> 
> Yes, primarily the fact that it cannot be scheduled to tick every `duration` starting at a specific **time**, only **every duration**. It would need some extra code to start the timer at the start of a layer to ensure that the `Timer` actually ticks at the beginning of the layer (and not x time after it).

I forgot the second reason. According to [docs](https://pkg.go.dev/time#NewTicker):
> The ticker will adjust the time interval or drop ticks to make up for slow receivers.

I didn't want dropping or adjusting the time interval to be out of our hands.

fasmat:
Yes I meant `time.Ticker` 🙂.

> I didn't want dropping or adjusting the time interval to be out of our hands.

It will always be out of our hands, if the system time changes (sync with time server or user just changing time) we will run into issues.

According to https://pkg.go.dev/time time calculations "assume a Gregorian calendar, with no leap seconds". Which means that our calculations for layer starts and ends could be off by a second or more when the next hits.

timesync/clock.go
```diff
@@ -59,39 +59,35 @@ func (t *TimeClock) startClock() error {
...
-			t.log.With().Info("clock notifying subscribers of new layer tick",
-				log.Int("subscriber_count", subscriberCount),
-				t.TimeToLayer(t.clock.Now()))
-
-			// notify subscribers
-			if missed, err := t.Notify(); err != nil {
-				t.log.With().Warning("could not notify all subscribers",
-					log.Err(err),
-					log.Int("missed", missed))
+			log.FieldNamed("next_layer", nextLayer),
```
countvonzero: log.Stringer. Dmitry wanted to deprecate FieldNamed

timesync/clock.go
```diff
@@ -59,39 +59,35 @@ func (t *TimeClock) startClock() error {
...
+			log.Time("next_tick_time", nextTickTime),
+		)
+
+		for {
+			// `time.After` sometimes unblocks bit too soon. In this case - wait again.
+			// See https://github.com/spacemeshos/go-spacemesh/issues/3617.
```
countvonzero, line 76: this comment seems inconsistent with the issue. the issue is about the timer, not `time.After`.
are they the same thing?

if not, why not wrap the original code with this for loop with if check so the issue is on point?

poszu: `time.After` and `time.Timer` share the same issue. The mentions `time.After` because this is used in the code right now.

I changed `time.Timer` to `time.After` because the code is simpler - no need to remember to `.Stop()` it manually. Is there any benefit of using `time.Timer` that I'm not aware of?

countvonzero:
```
// After waits for the duration to elapse and then sends the current time
// on the returned channel.
// It is equivalent to NewTimer(d).C.
// The underlying Timer is not recovered by the garbage collector
// until the timer fires. If efficiency is a concern, use NewTimer
// instead and call Timer.Stop if the timer is no longer needed.
func After(d Duration) <-chan Time {
	return NewTimer(d).C
}
```

countvonzero (approved):

fasmat (approved):

timesync/clock.go
poszu: Thanks, I wasn't aware :)

poszu:
bors merge

poszu:
bors merge

PR #3669, opened 2022-10-18
[Merged by Bors] - use opinion hash for verifying tortoise

### LINKED ISSUE #3637: use opinion hash for verifying tortoise
depends on #3635 

tortoise uses a special concept of ballot goodness to determine if ballot should be counted in the verifying mode. it creates a significant amount of complexity that can be completely removed, instead ballot in verifying mode should be counted only if:
- beacon matches local node beacon
- ballot opinion matches local node opinion for last votable layer

## implementation

1. compute local opinion hash. algorithm should follow same rules as specified in the #3635 
- within hdist when opinion on hare output changes
- outside hdist when weight for the layer crosses global threshold
- avoid recomputing local opinion hash due to weak coin or local threshold, as at that point we  are using full tortoise either way
2. replace code in verifying mode to count ballot when ballot beacon matches local beacon and ballot opinion hash matches local opinion hash for last votable layer (ballot layer - 1)
3. always try to verify a layer using verifying mode. if it works disable full mode. this simplifies logic around switching between the two

### affected code

- all code related to ballot goodness
https://github.com/spacemeshos/go-spacemesh/blob/develop/tortoise/verifying.go#L66-L79
- markGoodCut code can be removed
https://github.com/spacemeshos/go-spacemesh/blob/develop/tortoise/tortoise.go#L530-L543

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3637

verifying tortoise counts ballots that vote consistently with local opinion. this pr changes previous approach to a simpler version, that compares local opinion hash with ballot opinion hash, if hashes match each other - ballot will be counted.

local opinion hash is recomputed under the following conditions:
- hare output changes within hdist, including terminating a layer because it falls out of zdist distance
- layer outside hdist terminated with result that doesn't match hare output

every time when local opinion is recomputed tortoise will also re-scan ballots in verifying mode to re-enable it if it wasn't working.

### DIFF
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -60,9 +60,7 @@ type (
 		evicted types.LayerID
 
 		changedOpinion struct {
-			// sector of layers on which opinion is different from hdist
-			// once max falls out of hdist distance - opinion must
-			// be recomputed starting from min layer
+			// sector of layers where opinion is different from previously computed opinion
 			min, max types.LayerID
 		}
 
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -40,8 +40,8 @@ type (
 		margin weight
 
 		validity sign
-		// if validity requires persisting it to the disk
-		dirty bool
+		// persisted validity
+		persisted sign
 	}
 
 	state struct {
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -61,8 +61,8 @@ type (
 
 		changedOpinion struct {
 			// sector of layers on which opinion is different from hdist
-			// once max fall out of hdist distance - opinion needs
-			// can be recomputed starting from min layer
+			// once max falls out of hdist distance - opinion must
+			// be recomputed starting from min layer
 			min, max types.LayerID
 		}
 
--- a/tortoise/state.go
+++ b/tortoise/state.go
@@ -21,7 +21,6 @@ type (
 		// 10 and above, therefore its weight needs to be added to goodUncounted
 		// for layers 10 and above
 		goodUncounted   weight
-		abstained       weight
 		referenceHeight uint64
 	}
 
@@ -31,15 +30,6 @@ type (
 		weight uint64
 		// median height from atxs
 		height uint64
-	}
-
-	layerInfo struct {
-		lid            types.LayerID
-		empty          weight
-		hareTerminated bool
-		blocks         []*blockInfo
-		ballots        []*ballotInfo
-		verifying      verifyingInfo
 	}
 
 	blockInfo struct {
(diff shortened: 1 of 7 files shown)


tortoise/state.go
```diff
@@ -69,6 +59,13 @@ type (
 		// last evicted layer
 		evicted types.LayerID
 
+		changedOpinion struct {
+			// sector of layers on which opinion is different from hdist
+			// once max fall out of hdist distance - opinion needs
+			// can be recomputed starting from min layer
```
countvonzero: can you rewrite the comment? the last two sentences don't quite make sense.

tortoise/state.go
```diff
@@ -145,20 +142,53 @@ func (s *state) updateRefHeight(layer *layerInfo, block *blockInfo) error {
...
+	opinion types.Hash32
+	// a pointer to the value stored on the previous layerInfo object
+	// it is stored as a pointer so that when previous layerInfo is evicted
+	// we still have access in case we need to recompute opinion for this layer
+	prevOpinion *types.Hash32
+}
+
+func (l *layerInfo) computeOpinion(hdist uint32, last types.LayerID) {
+	hasher := hash.New()
+	if l.prevOpinion != nil {
```
countvonzero, line 160: can this ever be false?

tortoise/tortoise.go
```diff
@@ -457,11 +470,32 @@ func (t *turtle) verifyLayers() error {
...
+			if block.validity != block.hare {
+				if target.After(t.changedOpinion.max) {
+					t.changedOpinion.max = target
+				}
+				if t.changedOpinion.min.Value == 0 || target.Before(t.changedOpinion.min) {
+					t.changedOpinion.min = target
+				}
+			}
+			if block.validity == abstain {
+				logger.With().Panic("bug: layer should not be verified if there is an undecided block", target, block.id)
```
countvonzero: nit: maybe use fatal without the stack trace

tortoise/tortoise.go
```diff
@@ -457,11 +470,32 @@ func (t *turtle) verifyLayers() error {
 			break
 		}
 		verified = target
+		for _, block := range t.layers[target].blocks {
+			if !block.dirty {
+				continue
+			}
+			// record range of layers where opinion has changed.
+			// once those layers fall out of hdist window - opinion can be recomputed
+			if block.validity != block.hare {
```
countvonzero: what happens at startup regarding t.changedOpinion? say the following events take place

t0: hare changed opinion (received a hare certificate)
t1: app shutdown
t2: app startup

tortoise won't be able to detect that opinions has changed, as block wouldn't be dirty.

tortoise/state.go
dshulyak: first layer in the chain won't have prev opinion, currently it is genesis layer

tortoise/tortoise.go
dshulyak: if hare changed opinion and memory state was lost it doesn't matter, since hash will be recomputed on startup using new opinion

there could be a bug if opinion wasn't according to hare and that state got lost, will add a test

tortoise/state.go
dshulyak: updated

tortoise/tortoise.go
dshulyak: found couple of bugs, must be robust now

countvonzero (approved):

dshulyak:
bors try

dshulyak:
bors merge

PR #3670, opened 2022-10-18
[Merged by Bors] - Fix race in `genesisLayer` initialization

poszu:
## Motivation
```go
func GenesisLayer() *Layer
```
is racy.

## Changes
Wrap the genesis layer in a struct with `sync.Once` and use it to initialize the genesis layer only once.

## Test Plan
unit tests

## DevOps Notes

### DIFF
--- a/common/types/layer.go
+++ b/common/types/layer.go
@@ -4,6 +4,7 @@ package types
 import (
 	"fmt"
 	"strconv"
+	"sync"
 	"sync/atomic"
 
 	"github.com/spacemeshos/go-spacemesh/common/util"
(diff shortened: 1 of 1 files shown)


fasmat:
Imho it would be OK to have the oncer as package variable and use it in `GenesisLayer()` to protect the call to `InitGenesisData` / `genesisDate` and drop the wrapper construct.

dshulyak:
btw GenesisLayer is supposed to be removed, it was added for convenience long time ago and doesn't serve any real purpose (e.g not point to look for a better solution)

countvonzero (approved):

poszu:
bors merge

PR #3673, opened 2022-10-19
[Merged by Bors] - syncer/fetch: add feature to find opinion fork with peers

countvonzero:
## Motivation

part of #2507

## Changes

- add fork finder to find the layer at which the node's mesh hash diverge with the peer's. for each fork-finding session:
  - number of hashes requested is capped at N
  - uses a modified binary search algorithm that request N hashes at a time, with the boundaries hashes (from and to) re-queried each time to make sure the peer has not changed during the session

- fetch: request mesh hashes in batch from a specified peer


syncer/find_fork.go
```diff
@@ -0,0 +1,318 @@
...
+
+func (b *boundary) setTo(lh *layerHash) {
+	b[1] = lh
+}
+
+func (b *boundary) fromLayer() types.LayerID {
+	return b[0].lid
+}
+
+func (b *boundary) fomHash() types.Hash32 {
```
dshulyak: fromHash

syncer/find_fork.go
```diff
@@ -0,0 +1,318 @@
...
+		lid:  lid,
+		hash: hash,
+	}
+}
+
+// boundary is used to define the to and from layers in the hash mesh requests to peers.
+// the hashes of the boundary layers are known to the node and are used to double-check that
+// the peer has not changed its opinions on those layers.
+// if the boundary hashes change during the fork-finding session, the session is aborted.
+type boundary [2]*layerHash
```
dshulyak: i think if it will be defined as 
```go
type boundary struct {
   from, to layerHash
}
```
there will be no need for additional setters/getters

dshulyak (commented):
i haven't looked at all code, but do you expect this code to guarantee that missing ballots are downloaded? in case of missing ballots there won't be a difference in the opinion, but instead node with missing ballots won't be able to form an opinion on the layer

countvonzero:
> i haven't looked at all code, but do you expect this code to guarantee that missing ballots are downloaded? in case of missing ballots there won't be a difference in the opinion, but instead node with missing ballots won't be able to form an opinion on the layer

thanks. the actually syncing ballots/blocks will take place in the next PR. 
this PR only contain code that finds the layer of convergence, say N. 

in the next PR i'll add code to syncer 
- figure out which peer to resolve mesh hash difference with (the most weighted one with valid ATXs)
- find the layer N when mesh hash start to differ 
- fetch layer data from layer N from the specific peer.

fetch/mesh_data_test.go
```diff
@@ -707,3 +707,83 @@ func TestFetch_GetEpochATXIDs(t *testing.T) {
...
+	tt := []struct {
+		name     string
+		params   [4]uint32 // from, to, delta, steps
+		expected []uint32
+		err      error
+	}{
+		{
+			name:     "success",
+			params:   [4]uint32{7, 23, 5, 4},
+			expected: []uint32{7, 12, 17, 22, 23},
```
fasmat: Would it make sense for `expected` to be of type `[]types.LayerID`? Then we wouldn't need the function `toLayerIDs` to convert them in the test.

syncer/find_fork.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	uniquePeers := util.SliceToSetStringer(ff.fetcher.GetPeers())
+	for p, lh := range ff.agreedPeers {
+		if _, ok := uniquePeers[p.String()]; !ok {
+			if time.Since(lh.created) >= ff.maxStaleDuration {
+				delete(ff.agreedPeers, p)
+			}
+		}
+	}
```
fasmat: Can `ff.fetcher.GetPeers()` return a slice of peers that contains duplicates? If not this would work as well:

```suggestion
	for _, p := range ff.fetcher.GetPeers() {
		if lh, ok := ff.agreedPeers[p.String()]; !ok {
			if time.Since(lh.created) >= ff.maxStaleDuration {
				delete(ff.agreedPeers, p)
			}
		}
	}
```

This could also be faster, if we expect the slice returned by `ff.fetch.GetPeers()` to have fewer elements than the map `ff.agreedPeers`.

common/util/slices.go
```diff
@@ -14,3 +14,11 @@ func UniqueSliceStringer[T fmt.Stringer](s []T) []T {
 	}
 	return result
 }
+
+func SliceToSetStringer[T fmt.Stringer](s []T) map[string]struct{} {
```
fasmat: This can be removed if we iterate over the slice instead of the map (see other comment).

fasmat (approved):
lgtm, just added some small suggestions for discussion

syncer/find_fork.go
fasmat: Even if it returns duplicates, it's still at least as fast as first converting it to a set and then iterating over the set would be.

syncer/find_fork.go
dshulyak: it will also get rid from newBoundary function, that has no purpose

syncer/find_fork.go
```diff
@@ -0,0 +1,318 @@
...
+
+var ErrPeerMeshChangedMidSession = errors.New("peer mesh changed mid session")
+
+type layerHash struct {
+	lid     types.LayerID
+	hash    types.Hash32
+	created time.Time
+}
+
+func newLayerHash(lid types.LayerID, hash types.Hash32) *layerHash {
```
dshulyak: this function has no purpose

syncer/find_fork.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	if len(toPurge) > 0 {
+		for _, p := range toPurge {
+			delete(ff.agreedPeers, p)
+		}
+		return
+	}
+
+	uniquePeers := util.SliceToSetStringer(ff.fetcher.GetPeers())
```
dshulyak: it doesn't make sense to do it like that. peer.String() encodes Peer (which is string) using base58.  p2p.Peer can be used as a key

syncer/find_fork.go
```diff
@@ -0,0 +1,318 @@
...
+	for p, lh := range ff.agreedPeers {
+		if _, ok := uniquePeers[p.String()]; !ok {
```
dshulyak: every time String is called additional memory is allocated

syncer/find_fork.go
countvonzero: it is a little different. here i want to remove items in `agreedPeers` that's not in `fetcher.GetPeers()`.
i ended up created a temp map here. (also part of Dmitry's feedback)

syncer/find_fork.go
countvonzero: removed.

countvonzero:
bors try

countvonzero:
bors merge

PR #3678, opened 2022-10-20
[Merged by Bors] - Use context instead of timing out on WaitGroup.Wait()

fasmat:
## Motivation
Not waiting for `WaitGroup.Wait()` to return leaks go routines. This change tries to ensure that `Wait()` always returns.

## Changes
- Remove util function to timeout on `sync.WaitGroup.Wait()`
- Instead of `context.Background()` use `context.WithTimeout()` or `context.WithDeadline()` for long running tasks
   - This way we can ensure that spawned go routines finish and `Wait()` unblocks
   - Additionally the returned errors give insight on where inside the go routine the timeout happened.

## Test Plan
Changed tests still pass

## TODO

## DevOps Notes


api/grpcserver/grpcserver_test.go
```diff
@@ -118,22 +117,6 @@ var (
 	stateRoot = types.HexToHash32("11111")
 )
 
-func dialGrpc(t testing.TB, cfg config.Config) *grpc.ClientConn {
```
countvonzero, line 121: this is new code from @poszu where he factored out the duplicate code.
can you keep it? it doesn't seem to interfere with your change.

countvonzero (approved):

api/grpcserver/grpcserver_test.go
poszu: @fasmat I don't understand why you basically reverted it. IMO, it's worth having this helper as it hides details unrelated to the test meat and makes sure that grpc connection handling is correct in every test. Duplicating it yields no benefit. 

You can still pass context and use `grpc.DialContext` here.

api/grpcserver/grpcserver_test.go
```diff
@@ -1628,28 +1643,30 @@ func TestTransactionServiceSubmitUnsync(t *testing.T) {
...
-	conn := dialGrpc(t, cfg)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	conn, err := grpc.DialContext(ctx,
+		"localhost:"+strconv.Itoa(cfg.GrpcServerPort),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithBlock(),
+	)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, conn.Close()) })
```
poszu: IMO, grpc connection should have a separate timeout. 10s for connecting makes no sense.
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
	conn, err := grpc.DialContext(ctx,
		"localhost:"+strconv.Itoa(cfg.GrpcServerPort),
		grpc.WithTransportCredentials(insecure.NewCredentials()),
		grpc.WithBlock(),
		grpc.WithTimeout(time.Second),
	)
	require.NoError(t, err)
	t.Cleanup(func() { require.NoError(t, conn.Close()) })
```

poszu (changes requested):

api/grpcserver/grpcserver_test.go
fasmat: `grpc.WithTimeout` is functionally the same as `context.WithTimeout`. The functional option is marked as deprecated and reported by staticcheck if used.

api/grpcserver/grpcserver_test.go
fasmat: Originally I had two `dialGrpc` functions, one with `context` and one without. That didn't feel right to me, since the function is just calling one other function (`grpc.DialContext`) and asserting that this function didn't return an error.

I can revert the change, although I'm generally opposed to `util` and `helper` methods 🙂

api/grpcserver/grpcserver_test.go
fasmat: > 10s for connecting makes no sense.

Why not? The timeout isn't only for `grpc.DialContext` but for the whole test. If connecting takes 5 seconds and the remaining test finishes in under 5 seconds is that not just as valid as if connecting takes 1 second?

poszu: Because connecting shouldn't take 10s. I'd use a separate ctx for the test with timeout 10s and a child ctx for connection with timeout 1s.

api/grpcserver/grpcserver_test.go
```diff
@@ -118,19 +117,14 @@ var (
 	stateRoot = types.HexToHash32("11111")
 )
 
-func dialGrpc(t testing.TB, cfg config.Config) *grpc.ClientConn {
-	t.Helper()
```
poszu, line 122: Why did you remove it?

poszu (approved):

api/grpcserver/grpcserver_test.go
fasmat: Didn't remove it intentionally

fasmat:
bors merge

fasmat:
bors merge

PR #3680, opened 2022-10-21
[Merged by Bors] - remove genesis layer

### LINKED ISSUE #3642: remove dependency on the genesis block and ballot
this is a technical debt, historically tortoise use genesis ballot as a base ballot for the first batch of created ballots. 

## implementation

- remove both genesis ballot and genesis block
- instead of genesis ballot first base should be empty (e.g 20 empty bytes), and base layer should be equal to types.GetEffectiveGenesis
- most of the change will be adjusting tests

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3642

- empty 20 bytes are used as a base ballot if there are no other choices
- genesis layer is considered to be empty, and aggregated hash (opinion) is set accordingly

### DIFF
--- a/proposals/handler.go
+++ b/proposals/handler.go
@@ -27,7 +27,6 @@ var (
 	errMalformedData         = errors.New("malformed data")
 	errInitialize            = errors.New("failed to initialize")
 	errInvalidATXID          = errors.New("ballot has invalid ATXID")
-	errMissingBaseBallot     = errors.New("base ballot is missing")
 	errMissingEpochData      = errors.New("epoch data is missing in ref ballot")
 	errUnexpectedEpochData   = errors.New("non-ref ballot declares epoch data")
 	errEmptyActiveSet        = errors.New("ref ballot declares empty active set")
--- a/proposals/handler.go
+++ b/proposals/handler.go
@@ -346,10 +346,6 @@ func (h *Handler) checkBallotDataIntegrity(b *types.Ballot) error {
 		return errInvalidATXID
 	}
 
-	if b.Votes.Base == types.EmptyBallotID {
-		return errMissingBaseBallot
-	}
-
 	if b.RefBallot == types.EmptyBallotID {
 		// this is the smesher's first Ballot in this epoch, should contain EpochData
 		if b.EpochData == nil {
@@ -474,7 +470,10 @@ func ballotBlockView(b *types.Ballot) []types.BlockID {
 }
 
 func (h *Handler) checkBallotDataAvailability(ctx context.Context, b *types.Ballot) error {
-	blts := []types.BallotID{b.Votes.Base}
+	blts := []types.BallotID{}
+	if b.Votes.Base != types.EmptyBallotID {
+		blts = append(blts, b.Votes.Base)
+	}
 	if b.RefBallot != types.EmptyBallotID {
 		blts = append(blts, b.RefBallot)
 	}
--- a/tortoise/tortoise.go
+++ b/tortoise/tortoise.go
@@ -31,7 +31,7 @@ type turtle struct {
 	beacons system.BeaconGetter
 	updater blockValidityUpdater
 
-	state
+	*state
 
 	verifying *verifying
 
@@ -49,7 +49,7 @@ func newTurtle(
 ) *turtle {
 	t := &turtle{
 		Config:  config,
-		state:   *newState(),
+		state:   newState(),
 		logger:  logger,
 		cdb:     cdb,
 		beacons: beacons,
@@ -68,11 +68,12 @@ func newTurtle(
 		empty:          util.WeightFromUint64(0),
 		hareTerminated: true,
 	}
+	t.verifying = newVerifying(config, t.state)
+	t.full = newFullTortoise(config, t.state)
+	t.full.counted = genesis
+
 	gen := t.layer(genesis)
 	gen.computeOpinion(t.Hdist, t.last)
-	t.verifying = newVerifying(config, &t.state)
-	t.full = newFullTortoise(config, &t.state)
-	t.full.counted = genesis
 	return t
 }
 
--- a/tortoise/tortoise.go
+++ b/tortoise/tortoise.go
@@ -55,14 +55,12 @@ func newTurtle(
 		beacons: beacons,
 		updater: updater,
 	}
-	t.verifying = newVerifying(config, &t.state)
-	t.full = newFullTortoise(config, &t.state)
-	return t
-}
+	genesis := types.GetEffectiveGenesis()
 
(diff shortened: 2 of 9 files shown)


dshulyak:
bors try

tortoise/tortoise.go
```diff
@@ -55,48 +55,25 @@ func newTurtle(
...
+	t.processed = genesis
+	t.verified = genesis
+	t.evicted = genesis.Sub(1)
 
 	t.epochs[genesis.GetEpoch()] = &epochInfo{}
 	t.layers[genesis] = &layerInfo{
 		lid:            genesis,
 		empty:          util.WeightFromUint64(0),
 		hareTerminated: true,
 	}
```
poszu, line 70: These assignments could be part of the `t = &turtle {...}` block to avoid a halfway initialized instance.

tortoise/tortoise.go
```diff
@@ -55,48 +55,25 @@ func newTurtle(
...
-
-	t.last = genesis
-	t.processed = genesis
-	t.verified = genesis
-	t.evicted = genesis.Sub(1)
+	gen := t.layer(genesis)
+	gen.computeOpinion(t.Hdist, t.last)
+	t.verifying = newVerifying(config, &t.state)
+	t.full = newFullTortoise(config, &t.state)
 	t.full.counted = genesis
```
poszu: Could the order of `t.layer(genesis)` and the assignments below be inverted? `t.layer()` is called on a halfway initialized object `t`. Basically:
```go
t = &turtle {
    ...
}
t.verifying = newVerifying(config, &t.state)
t.full = newFullTortoise(config, &t.state)
// t is fully initialized now
gen := t.layer(genesis)
gen.computeOpinion(t.Hdist, t.last)
return t
```

dshulyak:
bors try

tortoise/tortoise.go
```diff
@@ -55,48 +55,25 @@ func newTurtle(
...
-	t.layer(genesis).computeOpinion(t.Hdist, genesis)
-
-	t.last = genesis
-	t.processed = genesis
-	t.verified = genesis
-	t.evicted = genesis.Sub(1)
+	gen := t.layer(genesis)
+	gen.computeOpinion(t.Hdist, t.last)
+	t.verifying = newVerifying(config, &t.state)
+	t.full = newFullTortoise(config, &t.state)
```
poszu: `type turtle struct` should probably be made non-copyable because it references itself in `verifying` and `full` (they have a pointer to `state`). Copying an instance of a `turtle` will lead to terrible bugs.

I saw there is no good way in Go that would be detectable in compile time. There is a `vet` check, details are described here: https://stackoverflow.com/a/52495303/2806736

The idea is to put a `Mutex` or an instance of ZST `noCopy`:
```go
type noCopy struct{}
func (*noCopy) Lock()   {}
func (*noCopy) Unlock() {}
```

But the best would be to avoid self-referencing. For example, `turtle` could keep a pointer to `type state struct` instead.

tortoise/tortoise_test.go
```diff
@@ -2146,23 +2140,22 @@ func TestOnBallotComputeOpinion(t *testing.T) {
...
-		hasher.Write(types.GenesisBlockID[:])
 		for i := 0; i < distance-1; i++ {
 			buf := hasher.Sum(nil)
 			hasher.Reset()
 			hasher.Write(buf)
 		}
 		require.Equal(t, hasher.Sum(nil), info.opinion().Bytes())
 	})
-	t.Run("support abstain support", func(t *testing.T) {
+	t.Run("against abstain support", func(t *testing.T) {
```
poszu, line 2158: NIT, It took me a while to understand this title. How about:
```suggestion
	t.Run("against → abstain → support", func(t *testing.T) {
```

tortoise/tortoise.go
dshulyak: i don't quite understand where is a problem, i pass the pointer to state

dshulyak: i will change to tortoise to keep pointer to state

tortoise/tortoise.go
dshulyak: they are modifying state struct

tortoise/tortoise.go
dshulyak: changed

tortoise/tortoise_test.go
dshulyak: i am not sure that i like additional ->

dshulyak:
bors try

proposals/handler.go
```diff
@@ -346,10 +346,6 @@ func (h *Handler) checkBallotDataIntegrity(b *types.Ballot) error {
 		return errInvalidATXID
 	}
 
-	if b.Votes.Base == types.EmptyBallotID {
```
countvonzero, line 349: in stead of being removed, this should be changed to 
`if b.Votes.Base == types.EmptyBallotID && b.LayerIndex != genesis layer`
otherwise how does one declare a ballot syntactically invalid when it doesn't declare the base ballot?

dshulyak: it is syntactically valid to have no base ballot. for example if at layer genesis + 3, it is valid to use not base ballot to add support votes for blocks in layers genesis + 1 and + 2

countvonzero: i see. thanks for the explanation

countvonzero (approved):

proposals/handler.go
countvonzero: is a ballot considered malicious if it doesnt vote for all historical blocks?

dshulyak: i don't see why. if ballot doesn't vote for a block it can simply drop it from the support list
we still need to enforce max distance to the base ballot though

countvonzero: what confuses me is, say all ballots dont vote for historical blocks and are still considered valid, how does any block cross the global threshold? isnt this an attack against the tortoise protocol?

countvonzero: sure when a node dont have a block locally, it doesnt have the block in the support list. i thought thats then considered as an against vote in the full tortoise mode.

dshulyak: > how does any block cross the global threshold? 

it means that all ballots votes against such block and it will cross negative threshold

> isnt this an attack against the tortoise protocol?

tortoise counts votes. i don't see an attack here

> i thought thats then considered as an against vote in the full tortoise mode.

against votes are counted by verifying mode as well

countvonzero: ok. thanks. the protocol depends on 2/3 honest actors and honest miners are incentivized to vote for historical blocks.

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #3684, opened 2022-10-24
[Merged by Bors] - within hdist should not include layer on a lower boundary

### LINKED ISSUE #3646: tortoise: refactor hdist and zdist usage
off by one error for hdist and zdist computation.

example: current layer is 100, zdist 2, layers 100 and 99 should within zdist. 98 should not be within the zdist

## implementation 

fix withinDistance method to return true only if layer is after `last - distance`

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3646

in this change hdist window is equal to (current-hdist, current].
for example if hdist 10, current 100 then hdist window will include exactly 10 layers, 91...100.
same was fixed for zdist

### DIFF
--- a/proposals/handler.go
+++ b/proposals/handler.go
@@ -385,10 +385,6 @@ func (h *Handler) setBallotMalicious(ctx context.Context, b *types.Ballot) error
 
 func (h *Handler) checkVotesConsistency(ctx context.Context, b *types.Ballot) error {
 	exceptions := map[types.BlockID]struct{}{}
-	cutoff := types.LayerID{}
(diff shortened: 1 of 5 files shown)


countvonzero (approved):

countvonzero:
i got it wrong here also. 

https://github.com/spacemeshos/go-spacemesh/blob/4a388cda7fb77e532e337a5eb189cce579b25cea/syncer/state_syncer.go#L114

dshulyak:
bors merge

PR #3686, opened 2022-10-24
[Merged by Bors] - extend votes with layer id and height

### LINKED ISSUE #3607: ballots: modify ballot.votes so that block body can be pruned later

- modify votes in the Support array to include layer id and tick height
- validate that layer id and tick height recorded in the vote matches values stored in the actual block

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3607
see https://community.spacemesh.io/t/deferred-block-fetching-a-k-a-dangling-pointers/309

- added layer/height to support/against votes. in this change they are simply validated that the vote is consistent with received block.
- opinion hash was updated so that support votes now also commit height into the hash. there is a helper for that that is used both in tortoise and mesh modules. it is kept in tortoise module, because it is likely that we can compute it always in tortoise and give out to mesh for writing to db, but i kept that for future
- bulk of the changes is refactoring fixture code in tests of the proposals module


dshulyak:
bors try

tortoise/opinionhash/opinion_hasher.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// OpinionHasher is a utility for computing opinion hash.
+type OpinionHasher struct {
+	enc *scale.Encoder
+	h   hash.Hash
+}
+
+// WritePrevious aggregated hash.
+func (h *OpinionHasher) WritePrevious(hash types.Hash32) {
```
countvonzero, line 25: questions about the APIs. does the order of calls to add previous/abstain/support matter?
does write of previous/abstain/support vs previous/support/abstain yield the same hash? if not, i think the code should enforce it.

dshulyak: yes, all of that matters. order of supported blocks also matters

dshulyak: i am not sure that it is really needed, the code would be much more complicated
that order of calls can be enforced with tests as it was done previously

countvonzero: the previous code is simpler (prev hash + hash of sorted block ids) tho. in any case, it's fine as long as you've considered it.

countvonzero (approved):

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3690, opened 2022-10-27
[Merged by Bors] - tortoise: recompute opinion if tally votes was called before hare output was saved

dshulyak:
opinion wasn't recomputed if tally votes was called before hare output was saved


countvonzero (approved):

dshulyak:
bors merge

PR #3693, opened 2022-10-28
[Merged by Bors] - vm: emit account update even after tx commit

### LINKED ISSUE #3687: GRPC GlobalStateService.AccountDataStream skips some updates
## Description
Smapp requests the current balance via `GlobalStateService.AccountDataQuery` and subscribes to account balance updates via GRPC Endpoint `GlobalStateService.AccountDataStream` (flag 4).
Meanwhile, this stream does not send any updates when some funds are received there and updates it only after the second transaction. But the message contains the previous state.

## Steps to reproduce
1. Create an account and subscribe to this endpoint
2. Send 100 smidges to the account
3. Wait until the transaction will be processed. You'll find out that there are no updates on the stream.
4. Send 200 more smidges to the account
5. Wait. And you'll find out that Stream shows up 100 smidges in current & projected balances
6. If you send a request to `GlobalStateService.AccountDataQuery` you'll see that the account has 300 smidges on the balance.

Waiting for some more layers applied after the transaction does not work.

## Actual Behavior
The endpoint sends the previous state all the time.

## Expected Behavior
It should send all the latest changes as soon as possible.

## Environment
 - Node Version: [5279493](https://github.com/spacemeshos/go-spacemesh/tree/5279493f46a831e26293719013ad86a079387732)
 - Smapp version: [from PR](https://github.com/spacemeshos/smapp/pull/975)

## Additional Resources 
image
If I restart Smapp, then it will request the `AccountDataQuery` endpoint and balance will be fine.
Tested it on devnet-300.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3687

added test that would fail previously if executed many times:
`go test ./api/grpcserver/ -run=TestVMAccountUpdates -v -count=1000 -cpu=1

### DIFF
--- a/api/grpcserver/globalstate_service.go
+++ b/api/grpcserver/globalstate_service.go
@@ -6,6 +6,7 @@ import (
 
 	pb "github.com/spacemeshos/api/release/go/spacemesh/v1"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
 	"github.com/spacemeshos/go-spacemesh/api"
@@ -234,6 +235,9 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 		if rewardsSubscription := events.SubscribeRewards(); rewardsSubscription != nil {
 			rewardsCh, rewardsBufFull = consumeEvents(stream.Context(), rewardsSubscription)
 		}
+	}
+	if err := stream.SendHeader(metadata.MD{}); err != nil {
+		return err
 	}
 
 	for {
--- a/api/grpcserver/grpcserver_test.go
+++ b/api/grpcserver/grpcserver_test.go
@@ -14,6 +14,7 @@ import (
 	"net"
 	"net/http"
 	"os"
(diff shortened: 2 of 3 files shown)


api/grpcserver/globalstate_service.go
```diff
@@ -235,6 +236,9 @@ func (s GlobalStateService) AccountDataStream(in *pb.AccountDataStreamRequest, s
 			rewardsCh, rewardsBufFull = consumeEvents(stream.Context(), rewardsSubscription)
 		}
 	}
+	if err := stream.SendHeader(metadata.MD{}); err != nil {
```
countvonzero, line 239: curious what difference does this make?

countvonzero (approved):

api/grpcserver/globalstate_service.go
dshulyak: it allows to wait for a confirmation that stream subscribed for events

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3695, opened 2022-10-31
[Merged by Bors] - mesh: consider block validity update when it's first updated

countvonzero:
## Motivation
when a block was optimistically applied, but the validity was never set before, the validity update didn't trigger a check on possible revert.

## Changes

- improve/reduce logging
- added test that failed before but passed after fix
- decouple state hash reporting from revert


miner/proposal_builder.go
```diff
@@ -338,9 +338,7 @@ func (pb *ProposalBuilder) createProposalLoop(ctx context.Context) {
 
 		case layerID := <-pb.layerTimer:
 			lyrCtx := log.WithNewSessionID(ctx)
-			if err := pb.handleLayer(lyrCtx, layerID); err != nil {
-				pb.logger.WithContext(lyrCtx).With().Warning("failed to create proposal", log.Err(err))
-			}
+			_ = pb.handleLayer(lyrCtx, layerID) // err logged in `handleLayer`
```
dshulyak: it is better practice to log error where it is handled instead of where it was created

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

Issue #3444, opened 2022-08-14
fetch: blocks on a channel forever for a hash result


countvonzero:
## Description

during longevity test, we continuously destroy node/start new nodes. occasionally a new node will get stuck syncing ATXs at the beginning. it's blocking on a channel for a hash result.

https://github.com/spacemeshos/go-spacemesh/blob/2e271a0bd577e7473ad8ad49b3463f6fd49dfc09/fetch/layers.go#L420 

## Steps to reproduce
1. make run test_name=TestScheduleBasic size=50 bootstrap=5m keep=true

countvonzero:
review comments from https://github.com/spacemeshos/go-spacemesh/pull/3696

https://github.com/spacemeshos/go-spacemesh/pull/3696#pullrequestreview-1161318004
```
could be also leaking requests, definitely worth investigating
it should be possible to drop batching altogether, and use libp2p stream per request. streams are not expensive to create
```

countvonzero:
more info on testing with gke-dev cluster.
whenever fetch timeout, around the same time p2p module will have many logs like these

```
2022-11-03 08:19:43 | {"L":"WARN","T":"2022-11-03T11:19:43.177-0400","N":"6cefe.p2p          ","M":"failed to connect to peer","node_id":"6cefea20165a3a2ec6273f7e56425fadbc291f505c85a8ee5b05ac203a52d1ac","module":"p2p","peer":"12D3KooWBwZRfasW9dzp5oDtYHdR45JRRKovj4ew2NbF7J2zwuFG","errmsg":"failed to dial 12D3KooWBwZRfasW9dzp5oDtYHdR45JRRKovj4ew2NbF7J2zwuFG:\n  * [/ip4/127.0.0.1/tcp/7513] failed to negotiate security protocol: peer id mismatch: expected 12D3KooWBwZRfasW9dzp5oDtYHdR45JRRKovj4ew2NbF7J2zwuFG, but remote key matches 12D3KooWJSXvhauSJoXDHQV1JJw5N7WMFZ5K94fgvfjt8FQhbUGN\n  * [/ip4/10.20.6.102/tcp/7513] dial tcp4 10.20.6.102:7513: i/o timeout","name":"p2p"}
```
and
```
[...]

PR #3696, opened 2022-10-31
[Merged by Bors] - syncer: add timeout to layer polling

countvonzero:
## Motivation

related to #3444
during network partition test, occasionally observed that `syncLayer` hangs and never return.

## Changes

- add layer polling timeout


syncer/data_fetch.go
```diff
@@ -13,6 +14,8 @@ import (
 	"github.com/spacemeshos/go-spacemesh/rand"
 )
 
+const pollTimeOut = time.Second * 30
```
dshulyak: could you please use 30 * time.Second, it is common in golang to write number before duration type

dshulyak:
do you know what hangs in the stack? every p2p server is configured with RequestTimeout in the fetcher, default value is 10s. so it could be that this timeout won't help as well

countvonzero:
> do you know what hangs in the stack? every p2p server is configured with RequestTimeout in the fetcher, default value is 10s. so it could be that this timeout won't help as well

there are 2 levels of waiting on top of that
- polling across N peers for each layer
- multiple requests to p2p server for each layer
  - polling layer data (ballot isd and block ids)
  - fetch those ids
 
most likely the bug is in the last one (fetching ids, see #3444). the batching of hash requests were not updated since i joined and i didn't invest time to debug there. 

my motivation for this PR are
- clients of the fetch library enforce a timeout so data syncing can progress
- print a warning when fetching times out for further investigation

dshulyak (approved):
could be also leaking requests, definitely worth investigating

it should be possible to drop batching altogether, and use libp2p stream per request. streams are not expensive to create

countvonzero:
bors merge

PR #3697, opened 2022-10-31
[Merged by Bors] - syncer: re-sync data from peers with diverged mesh hash

### LINKED ISSUE #2507: mesh hash resolution with peers

### LINKED ISSUE #2527: implement syncer code that drives hash resolution
## Description
part of #2507

implement syncer code that drives hash resolution as described (in pink) in the following chart

sync-cfg

specifically, the box `resolve hash difference` is expanded to the diagram below:

hash-resolution-cfg

countvonzero:
## Motivation

Closes #2507
Closes #2527

## Changes

- syncer: remove code syncing block validity. this is done to simplified genesis logic, as full tortoise is feasible.

- syncer: when detecting peers has a different mesh hash, taking the following measure
  - for each peer, find fork layer with that peer and resync from that peer [fork+1, latest]
  - avoid resync from the same peer within the same run
  - avoid resync based on the same layer hash divergence
  - after all data is collected, trigger ProcessLayer to allow tortoise to tally votes

- fetch: use EpochInfo to capture weight/beacon/active set when peers exchange epoch data. this is used 
  - to sync epoch ATXs
  - to avoid malicious peer forcing the node to do endless fork-finding session, make sure the peer's active set is valid prior to entering a session

- add systests for 30/70 and 50/50 splits. the success condition is that all nodes agreed on the state hash for every layer


systest/tests/partition_test.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	require.NoError(t, finalErr)
+	require.True(t, pass)
+	_ = txeg.Wait()
+}
+
+func TestPartition_30_70(t *testing.T) {
+	t.Parallel()
+
+	tctx := testcontext.New(t, testcontext.Labels("sanity"))
```
dshulyak: we may have to remove it from sanity group. and create a separate job for lets say destructive group. or use more resources and parallize tests better

there were some problems if tests were running for more than 60m

countvonzero: > create a separate job

do you mean to have two systest jobs running concurrently?

dshulyak: one by one. for concurrent we need to provision more clusters at the same time

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/actions/runs/3388705783/jobs/5632226985)

the usual TestAddNodes failure

tortoise/tortoise.go
```diff
@@ -486,9 +483,15 @@ func (t *turtle) verifyLayers() error {
...
-			if err != nil {
-				return fmt.Errorf("saving validity for %s: %w", block.id, err)
+			valid := block.validity == support
+			if valid {
+				if err := blocks.SetValid(t.cdb, block.id); err != nil {
+					return fmt.Errorf("setting %s valid: %w", block.id, err)
+				}
+			} else {
+				if err := blocks.SetInvalid(t.cdb, block.id); err != nil {
+					return fmt.Errorf("setting block %s invalid: %w", block.id, err)
```
dshulyak: this wrapping is completely unnecessary, it doesn't add any new information

but i think this change will have to be reverted

dshulyak (commented):
revert decision should be based on tortoise counting
if tortoise terminates with X, then all nodes change the vote to Y (note that there is no partition) - they all will remain with X applied

so in order to detect that opinion changed to Y - layer needs to be reverted based on tortoise counting, not on what peer is saying

systest/tests/partition_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"golang.org/x/sync/errgroup"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/systest/chaos"
+	"github.com/spacemeshos/go-spacemesh/systest/cluster"
+	"github.com/spacemeshos/go-spacemesh/systest/testcontext"
+)
+
+func sendTransactions(tctx *testcontext.Context, cl *cluster.Cluster, stop uint32) *errgroup.Group {
```
dshulyak: is it possible to reuse previous code for sending transactions?

fetch/wire_types.go
```diff
@@ -52,6 +52,12 @@ type MeshHashes struct {
 	Hashes []types.Hash32
 }
 
+type EpochData struct {
+	Beacon types.Beacon
+	Weight uint64
```
dshulyak: can we please drop this two fields? both of them should have zero impact on downloading atxs. if they do then something is clearly wrong

dshulyak:
> to determine whether to trust a peer's data prior to entering a fork-finding session with the peer. i.e. if the node fails to fetch the active set from the peer, it will not trust the peer's mesh hash.

i don't understand what is it for. syncer doesn't need to trust peer's mesh hash, it needs to download missing data. if data is valid it will be used by tortoise and thats it

> gather the minFork layer from all peers, and re-sync layer data from them from minFork+1 layer

it makes more sense to individually download diff from every peer based on the fork with that peer

> syncer: prior to adopting a peer's opinions, do some basic validity checking on the opinions

i would consider dropping peers opinions for genesis, and maybe restore them in some distant future. it is an optimization that introduces even more complexity in the syncer

syncer/data_fetch.go
```diff
@@ -51,23 +52,25 @@ type (
...
-		fetcher: fetch,
 	}
 }
 
 // PollLayerData polls all peers for data in the specified layer.
-func (d *DataFetch) PollLayerData(ctx context.Context, lid types.LayerID) error {
-	peers := d.fetcher.GetPeers()
+func (d *DataFetch) PollLayerData(ctx context.Context, lid types.LayerID, peers ...p2p.Peer) error {
+	peers = append(peers, d.fetcher.GetPeers()...)
+	peers = util.UniqueSliceStringer(peers)
```
dshulyak: that function should not be used for anything beside logging. peer.ID is already a string, String method copies that data as a []byte buffer and encodes it with b58

syncer/data_fetch.go
```diff
@@ -51,23 +52,25 @@ type (
...
 		msh:     msh,
-		fetcher: fetch,
 	}
 }
 
 // PollLayerData polls all peers for data in the specified layer.
-func (d *DataFetch) PollLayerData(ctx context.Context, lid types.LayerID) error {
-	peers := d.fetcher.GetPeers()
+func (d *DataFetch) PollLayerData(ctx context.Context, lid types.LayerID, peers ...p2p.Peer) error {
+	peers = append(peers, d.fetcher.GetPeers()...)
```
dshulyak: where fetcher gets peers? it is from currently open connections? where do you get peers that are sent to this method?

dshulyak: they are also from fetcher.. this doesn't makes sense then

dshulyak:
what happens if data was download from peer but local opinion didn't change, it can happen atleast for two reasons:
- ballots with different beacon counted are delayed
- local opinion can be more "correct"

is there something that will prevent querying peer with different opinion again?

cmd/node/node_test.go
```diff
@@ -817,6 +817,12 @@ func TestInitialize_BadTortoiseParams(t *testing.T) {
 	app = New(WithLog(logtest.New(t)), WithConfig(&conf))
 	err := app.Initialize()
 	assert.EqualError(t, err, "incompatible tortoise hare params")
+
+	conf, err = presets.Get("fastnet")
```
dshulyak: this seems out of palce

countvonzero:
> > syncer: prior to adopting a peer's opinions, do some basic validity checking on the opinions
> 
> i would consider dropping peers opinions for genesis, and maybe restore them in some distant future. it is an optimization that introduces even more complexity in the syncer

are you saying that a new node doesn't need block validity from peers to bootstrap verifying tortoise local opinion?
see 2nd point on https://github.com/spacemeshos/go-spacemesh/issues/3003

countvonzero:
> what happens if data was download from peer but local opinion didn't change, it can happen atleast for two reasons:
> 
> * ballots with different beacon counted are delayed
> * local opinion can be more "correct"
> 
> is there something that will prevent querying peer with different opinion again?

in the current code in develop, if the block validity is set already it will not query peer for block validity again.
but there is a race condition, hence adding `blocks.SetValidIfNotSet`

countvonzero:
> revert decision should be based on tortoise counting if tortoise terminates with X, then all nodes change the vote to Y (note that there is no partition) - they all will remain with X applied
> 
> so in order to detect that opinion changed to Y - layer needs to be reverted based on tortoise counting, not on what peer is saying

i know this idea is conceptually incorrect, as the mesh content should be based on tortoise's vote counting decision.
but i did it for efficiency, as 

- blocks are optimistically applied. i.e. the validity is always decided after a block is applied that can be a possible revert
- even without partition, nodes can disagree on mesh hash. some node can fail locally with hare and didn't apply anything. 
in this case, it also rely on syncer to detect that there is a hash difference with peers and try to download data from them.

ultimately the fork detection in syncer doesn't care whether there is a partition or not. it just cares that there is disagreement from peers. 

the risk is that when all nodes agree on the mesh content (based on hare output), then nothing will drive the network to the correct tortoise output. i checked the logs on other systests (TestAddNodes and TestFailedNodes), the syncer fork detection occurred many times as well.

in any case, i am happy to restore to the old behavior. tho the simplicity of checking for revert appeal to me. @dshulyak

countvonzero:
> > to determine whether to trust a peer's data prior to entering a fork-finding session with the peer. i.e. if the node fails to fetch the active set from the peer, it will not trust the peer's mesh hash.
> 
> i don't understand what is it for. syncer doesn't need to trust peer's mesh hash, it needs to download missing data. if data is valid it will be used by tortoise and thats it

but how else are we going to prevent malicious peers trying to get the node endlessly trying to sync non-existing data?
the only way i can think of, is to double check if the peer has all the ATXs it claimed it has. we had a discussion in slack a while back, you also asked the same question.

beacon is indeed unnecessary/unused. i only logged the difference now. 

UPD: the approach (verifying peers' active set) is also suggested by Tal when discussing mesh resolution. that was also 1 yr back.

dshulyak:
> are you saying that a new node doesn't need block validity from peers to bootstrap verifying tortoise local opinion?

verifying tortoise is an optimization. we don't have to use it for syncing a node. that was a mistake due to miscommunication  with research

> in any case, i am happy to restore to the old behavior. tho the simplicity of checking for revert appeal to me. 

i think there are just too many problems with that:
1. timing. syncer may download ballots with different beacon that will be delayed. and opinion will change long after syncer reverted the layer where opinion hash was different. e.g. it run revert at point X, 10 layer after tortoise counted more delayed ballots, opinion actually changed, but we didn't revert
2. peers do not commit to the provided layer hash. so they may say that opinion is X, but counting votes will result in Y
3. we currently don't update layer hash every time when tortoise changes vote. so it is not guaranteed that layer hash in mesh will be updated if originally vote was X and then became Y

dshulyak:
> in the current code in develop, if the block validity is set already it will not query peer for block validity again.
but there is a race condition, hence adding blocks.SetValidIfNotSet

my question was about not trying to run fork finder with peer if we already downloaded data from that peer, but opinion didn't change

> but how else are we going to prevent malicious peers trying to get the node endlessly trying to sync non-existing data?

maybe i misunderstood the comment then. clearly if node can't download something from a peer it shouldn't try to download more stuff

countvonzero:
> what happens if data was download from peer but local opinion didn't change, it can happen atleast for two reasons:
> 
> * ballots with different beacon counted are delayed
> * local opinion can be more "correct"
> 
> is there something that will prevent querying peer with different opinion again?

not in the current PR. tho i'll need to remember the disagreed hash across syncer runs.
the thing is that the same peer can also be changing its opinions, so its mesh content can change across syncer runs.
but i can cache the peer+hash combo to avoid another fork-finding session.

fetch/wire_types.go
countvonzero: i ended up dropping beacon and kept weight for logging the amount of ignored weight from the missing atxs

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
@dshulyak can you take another look? thank you

.github/workflows/ci.yml
```diff
@@ -263,7 +263,7 @@ jobs:
     needs:
       - filter-changes
       - dockerpush
-    timeout-minutes: 60
+    timeout-minutes: 90
```
dshulyak, line 266: do we still need to increase it, if there is another group?

fetch/wire_types.go
dshulyak: but why to send it? it is not like this number can be trusted

dshulyak (approved):

.github/workflows/ci.yml
countvonzero: my understanding is that `systest I` (timeout 50 min) and `systest II` (timeout 30 min) will run sequentially,
hence increasing the overall timeout from 60 -> 90. 
do i understand this incorrectly?

countvonzero: actually based on data in https://github.com/spacemeshos/go-spacemesh/actions/runs/3401176463/jobs/5656190035, the two tests are running sequentially.

countvonzero:
bors try

countvonzero:
bors merge

PR #3701, opened 2022-11-02
[Merged by Bors] - systest: parametrize tests

### LINKED ISSUE #3699: system tests (especially longevity) should be parametrizable
current setup runs short epochs (4 layers/epoch), where every layer is 15s. this is not optimal for long-running networks and for introducing short disruptions in network connectivity and other non-critical failure modes.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3699

Tests are parametrized using configmap that must be created in the same namespace
where test pod is running. Name of the configmap by default will match name of the pod where
tests are running.

Developer can replace configmap that is used in tests by creating a custom one, and updating `configname` variable.

```bash
make run configname=<custom configmap> test_name=TestStepCreate
```

Alternatively it is possible to replace smesher/poet config and add more values to the config map.

Command below will update configmap to use longer epochs:

```bash
make run smesher_config=parameters/longfast/smesher.json poet_config=parameters/longfast/poet.conf test_name=TestStepCreate
```

Command below will add additional variable to configmap from env file:
```bash
echo "layers-to-check=20" >> properties.env
make run properties=properties.env test_name=TestStepCreate
```


dshulyak:
bors try

dshulyak:
bors try

countvonzero (approved):

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

PR #3715, opened 2022-11-07
[Merged by Bors] - Use updated Initializer API from PoST

fasmat:
## Motivation
Integrates changes of https://github.com/spacemeshos/post/issues/78 into go-spacemesh, only merge after the former has been merged.

## Changes
The asynchronous `Initializer::SessionNumLabelsWrittenChan()` is removed in favor of the synchronous `Initializer::SessionNumLabelsWritten()` with the changes in https://github.com/spacemeshos/post/pull/84/

This updates go-spacemesh to not use the removed method any more.

The GRPC server now sends PoST status updates to a client in 1 second intervals instead of relying on PoST to report its status in (possibly) irregular intervals. This also means that a client now has to cancel a request or it will receive status updates about PoST indefinitely.

## TODO

## DevOps Notes

### DIFF
--- a/activation/post.go
+++ b/activation/post.go
@@ -225,7 +225,7 @@ func (mgr *PostSetupManager) StartSession(opts atypes.PostSetupOpts, commitmentA
 			defer mgr.mu.Unlock()
 
 			if errors.Is(err, context.Canceled) {
-				mgr.logger.Info("post setup session stopped")
+				mgr.logger.Info("post setup session was stopped")
 				mgr.state = atypes.PostSetupStateNotStarted
 			} else {
 				mgr.state = atypes.PostSetupStateError
--- a/activation/post.go
+++ b/activation/post.go
@@ -1,6 +1,7 @@
 package activation
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sync"
@@ -62,9 +63,10 @@ type PostSetupManager struct {
 	lastOpts *atypes.PostSetupOpts
 	lastErr  error
 
-	// startedChan indicates whether a data creation session has started.
-	// The channel instance is replaced in the end of the session.
-	startedChan chan struct{}
+	// ctx and cancel are the context and cancel function of the initialization
+	// they can be used to stop and monitor a started session
+	ctx    context.Context
+	cancel context.CancelFunc
 
 	// doneChan indicates whether the current data creation session has finished.
 	// The channel instance is replaced in the beginning of the session.
@@ -81,7 +83,6 @@ func NewPostSetupManager(id types.NodeID, cfg atypes.PostConfig, logger log.Log,
 		goldenATXID:       goldenATXID,
 		state:             atypes.PostSetupStateNotStarted,
 		initCompletedChan: make(chan struct{}),
-		startedChan:       make(chan struct{}),
 	}
 
 	return mgr, nil
@@ -186,7 +187,12 @@ func (mgr *PostSetupManager) StartSession(opts atypes.PostSetupOpts, commitmentA
 	}
 
 	commitment := GetCommitmentBytes(mgr.id, commitmentAtx)
-	newInit, err := initialization.NewInitializer(config.Config(mgr.cfg), config.InitOpts(opts), commitment)
+	newInit, err := initialization.NewInitializer(
+		initialization.WithCommitment(commitment),
+		initialization.WithConfig(config.Config(mgr.cfg)),
+		initialization.WithInitOpts(config.InitOpts(opts)),
+		initialization.WithLogger(mgr.logger),
+	)
 	if err != nil {
 		mgr.mu.Lock()
 		mgr.state = atypes.PostSetupStateError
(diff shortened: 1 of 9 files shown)


activation/post.go
```diff
@@ -232,20 +201,19 @@ func (mgr *PostSetupManager) StartSession(opts atypes.PostSetupOpts, commitmentA
...
 	}
 
-	newInit.SetLogger(mgr.logger)
-
 	mgr.mu.Lock()
 	mgr.init = newInit
 	mgr.lastOpts = &opts
 	mgr.lastErr = nil
-	close(mgr.startedChan)
+	mgr.ctx, mgr.cancel = context.WithCancel(context.TODO())
```
poszu:
```suggestion
	mgr.ctx, mgr.cancel = context.WithCancel(context.Background())
```

activation/post.go
```diff
@@ -63,9 +63,10 @@ type PostSetupManager struct {
 	lastOpts *atypes.PostSetupOpts
 	lastErr  error
 
-	// startedChan indicates whether a data creation session has started.
-	// The channel instance is replaced in the end of the session.
-	startedChan chan struct{}
+	// ctx and cancel are the context and cancel function of the initialization
+	// they can be used to stop and monitor a started session
+	ctx    context.Context
+	cancel context.CancelFunc
```
poszu: What's the point of storing the context? Isn't it enough to story the `cancel` function only?
```suggestion
	cancel context.CancelFunc
```

activation/post.go
```diff
@@ -258,11 +226,11 @@ func (mgr *PostSetupManager) StartSession(opts atypes.PostSetupOpts, commitmentA
...
 		)
 
-		if err := newInit.Initialize(); err != nil {
+		if err := newInit.Initialize(mgr.ctx); err != nil {
 			mgr.mu.Lock()
 			defer mgr.mu.Unlock()
 
-			if errors.Is(err, initialization.ErrStopped) {
+			if errors.Is(err, context.Canceled) {
 				mgr.logger.Info("post setup session stopped")
```
poszu: How about changing the log to say "session was stopped" to reflect that it was interrupted by an external event and not just finished?
```suggestion
				mgr.logger.Info("post setup session was stopped")
```

activation/post.go
```diff
@@ -298,9 +266,7 @@ func (mgr *PostSetupManager) StopSession(deleteFiles bool) error {
 	mgr.mu.Unlock()
 
 	if state == atypes.PostSetupStateInProgress {
-		if err := init.Stop(); err != nil {
-			return fmt.Errorf("stop: %w", err)
-		}
+		mgr.cancel()
 
 		// Block until the current data creation session will be finished.
 		<-doneChan
```
poszu, line 265: There is a possible deadlock here. Consider events, A and B depict two goroutines:
1. A calls `StartSession`. `state = PostSetupStateInProgress`. `doneChan` is still nil
2. B calls `StopSession`. `state` is already `InProgress` but the chan is `nil` thus it hangs forever.

`doneChan` and `state` must be set in a single critical section.

activation/post_test.go
```diff
@@ -34,32 +35,42 @@ func TestPostSetupManager(t *testing.T) {
...
+		defer timer.Stop()
 
-			lastStatus = status
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-timer.C:
+				status := mgr.Status()
+				req.True(status.NumLabelsWritten >= lastStatus.NumLabelsWritten)
```
poszu: How about using `GreaterOrEqual`? It would give better error messages in case of failure.
```suggestion
				req.GreaterOrEqual(status.NumLabelsWritten, lastStatus.NumLabelsWritten)
```

api/grpcserver/smesher_service.go
```diff
@@ -183,13 +183,13 @@ func (s SmesherService) PostSetupStatus(context.Context, *empty.Empty) (*pb.Post
 func (s SmesherService) PostSetupStatusStream(_ *empty.Empty, stream pb.SmesherService_PostSetupStatusStreamServer) error {
 	log.Info("GRPC SmesherService.PostSetupStatusStream")
 
-	statusChan := s.postSetupProvider.StatusChan()
+	timer := time.NewTicker(time.Second)
+	defer timer.Stop()
```
poszu: The tick period could be configurable. It could be passed in the request data or it could be a property of the `SmesherService`. What do you think?

api/grpcserver/grpcserver_test.go
```diff
@@ -988,112 +987,118 @@ func TestSmesherService(t *testing.T) {
...
+
+	t.Run("PostSetupStatusStream", func(t *testing.T) {
+		logtest.SetupGlobal(t)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		stream, err := c.PostSetupStatusStream(ctx, &empty.Empty{})
+		require.NoError(t, err)
+
+		// Expecting the stream to return updates before closing.
+		for i := 0; i < 3; i++ {
```
poszu, line 1093: This loop will take 3s making this test very slow 😢 . It could be speeded up if the ticking time was configurable.

poszu (commented):
Great! All comments are nits, except for the comment about deadlock - IMO a bug issue should be created about it.

activation/post.go
fasmat: Added a clarifying comment

activation/post.go
fasmat: I'm not sure what you mean. This message is logged when the caller calls `StopSession()`. "was stopped" to me implies it might have been unintentional, while this (at least as it is implemented at the moment) is always intentional.

activation/post.go
fasmat: There is an even bigger problem in `StartSession`: `mgr.state` is checked for "InProgress" and if it is not, `StartSession` does some things and then sets the state to "InProgress", essentially allowing multiple go routines to call `StartSession` simultaneously.

I'll see how I can improve this; I might have to do similar changes to `PostSetupManager` as I did to `Initializer` to get rid of deadlocks and races 😐

fasmat: I updated `StartSession` to not unlock the mutex before both the status is set and the channel is created, so this specific deadlock should be resolved.

activation/post.go
poszu: For me, it's the opposite :)
- something stopped - possibly on its own ("a car stopped")
- something **was** stopped - it was stopped by someone ("a car was stopped by the driver")

api/grpcserver/smesher_service.go
fasmat: Made timeout configurable

poszu (approved):

fasmat:
bors merge

fasmat:
bors merge

poszu:
bors merge

PR #3727, opened 2022-11-09
[Merged by Bors] - don't cancel identity because of syntactically invalid ballots

dshulyak:
identity should not be cancelled if the data is not kept on mesh and can't be provided to other peers, to ensure that they see things the same way. syntactically invalid data is simply discarded, therefore should not be able to cancel identity.


countvonzero (approved):

dshulyak:
bors merge

PR #3728, opened 2022-11-09
[Merged by Bors] - tortoise: load all stored ballots

### LINKED ISSUE #3719: tortoise: load all saved data, not limited by processed layer
on startup tortoise needs to load data that was stored before node crashed, currently it loads data until processed layer.
processed layer can lag behind actually fetched data. 

for example on my environment node received ballots from layer 1042, but tortoise recovered data only for 1041. so when after failure sync started to recover more data - it couldn't fetch anything due to missing state in tortoise.

## solution

find max ballots layer and use it in tortoise for recovery.
makes sense to rename MeshProcessed parameter to something more appropriate.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3719

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -509,15 +509,9 @@ func (app *App) initServices(ctx context.Context,
 		return fmt.Errorf("failed to create mesh: %w", err)
 	}
 
-	processed := msh.ProcessedLayer()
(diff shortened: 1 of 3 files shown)


countvonzero (approved):

dshulyak:
bors merge

PR #3730, opened 2022-11-09
[Merged by Bors] - Run systests in Job

poszu:
## Motivation
System tests in CI fail sporadically because `kubectl logs -f <systest pod>` exits prematurely (while tests are still running). An example of a CI run failed in this way: https://github.com/spacemeshos/go-spacemesh/actions/runs/3421691184/jobs/5698440170.

I suspect this is because of log rotation, which `kubectl` doesn't support properly (see https://github.com/kubernetes/kubernetes/issues/59902 and https://github.com/kubernetes/kubernetes/issues/28369).

## Changes
The fix is to wait for the systests to actually finish. It turns out that it's not possible when using a Pod because `kubectl wait --for=<>` doesn't provide a way to wait for Pod to "complete". Kubernetes has a different kind of object for this - a [Job](https://kubernetes.io/docs/concepts/workloads/controllers/job/).

Hence in this PR, I changed the deployment of systests to run in Jobs and I added a script to wait for the tests to finish.

`make run` now returns the exit code properly depending if the tests passed or failed. Thanks to it the CI step running tests is automatically marked as failed if tests failed - I could remove additional steps to check tests result.

Finally, in order to spawn a Job with restart policy = Never and additional labels, I had to create it via `kubectl apply -f` because `kubectl create job` is too limited. In order to create a Job deployment yaml file (to be used with `kubectl apply -f`), I use a template file (`systest_job.yml.tmpl`) which is rendered using [gomplate](https://gomplate.ca).

## Test Plan
CI passing

## DevOps Notes


poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

dshulyak:
would it be simpler to wait on container status?
something like `kubectl wait --for=jsonpath='{.status.containerStatuses[].state.terminated.reason}'=Completed pod/systest-e1d1431f`

poszu:
> would it be simpler to wait on container status? something like `kubectl wait --for=jsonpath='{.status.containerStatuses[].state.terminated.reason}'=Completed pod/systest-e1d1431f`

No, it would not work. Kubectl would return an error because it couldn't find the status:
```bash
❯ kubectl wait --for=jsonpath='{.status.containerStatuses[].state.terminated.reason}'=Completed pod -l testid=systest-4f681d22
pod/systest-4f681d22-1668067634-qrkhg condition met # <-- WORKS when pod finished
error: terminated is not found                      # <-- FAILS when pod is executing
```

poszu:
bors try

.github/workflows/ci.yml
```diff
@@ -292,43 +292,61 @@ jobs:
...
 
       - name: System test result
-        # SUCCESS will capture all the pods exit status separated by space. adding up all the numbers.
         run: |
-          SUCCESS=$(kubectl get pod --selector testid=systest-${{ steps.vars.outputs.sha_short }} -o json | jq .items[].status.containerStatuses[].state.terminated.exitCode)
-          SUCCESS=$(echo $SUCCESS | tr ' ' '+' | bc -l)
-          echo "::set-output name=content::$SUCCESS"
+          PASSED_TESTS=$(kubectl logs --selector testid=systest-${{ steps.vars.outputs.sha_short }} --tail=1 | grep PASS | wc -l)
+          echo "$PASSED_TESTS tests passed"
+          echo "passed_tests=$PASSED_TESTS" >> $GITHUB_OUTPUT
```
fasmat: The advantage of this previous approach was that if the number of executed tests changes the job doesn't have to be updated. Can the new approach be updated to also just check if all tests print `PASS` instead of counting the number of jobs that printed it?

fasmat (approved):

.github/workflows/ci.yml
poszu: I know. I originally had it check if none test printed "FAIL" at the end. But It wouldn't work if the tests crashed (it would print neither PASS nor FAIL). Do you have any ideas how to improve it?

fasmat: The exit code of the container started by the job is overwritten with `|| exit 0`. Is this required? The test binary returns a non-0 exit code when at least one test failed.

dshulyak (approved):

poszu:
bors try

fasmat (approved):

poszu:
bors merge

PR #3736, opened 2022-11-10
[Merged by Bors] - block certifyer: handle multiple certificates and equivocation

### LINKED ISSUE #3467: tortoise: vote for an empty layer if multiple certificate were signed by the same committee
## Description
it's possible for the network to have multiple certificates for the same layer when there is a network split.

but if the same committee signs more than one certificate, it should be considered malicious and the node should vote for an empty layer.

@dshulyak

countvonzero:
## Motivation

Closes #3467

## Changes

- sql: add  table `certificates` to hold certificates. the reason why the certificate is not a property of a block is that the network will certify a empty layer (using `types.EmptyBlockID` in certificate) when the hare output an empty set.

- mesh, tortoise: while multiple valid certificates are stored in the table, when requesting the block to apply/vote for, `certificates.GetHareOutput` will return `types.EmptyBlockID` as hare output when there are multiple valid certificates

- sync, fetcher: only asked peers for certificate when the nodes doesn't have any. only serve a certificate to peers when there is just one


countvonzero:
bors try

sql/certificates/certs.go
```diff
@@ -0,0 +1,121 @@
...
+		return fmt.Errorf("add wo cert %s: %w", lid, err)
+	}
+	return nil
+}
+
+// GetHareOutput returns the block that's valid as hare output for the spcified layer.
+// if there are more than one valid blocks, return types.EmptyBlockID.
+func GetHareOutput(db sql.Executor, lid types.LayerID) (types.BlockID, error) {
+	var (
+		result []types.BlockID
```
dshulyak: result can be just a types.BlockID
if rows > 1 then it can return types.EmptyBlockID

mesh/mesh.go
```diff
@@ -458,12 +459,12 @@ func layerValidBlocks(logger log.Log, cdb *datastore.CachedDB, layerID, latestVe
...
 	if layerID.After(latestVerified) {
 		// tortoise has not verified this layer yet, simply apply the block that hare certified
-		bid, err := layers.GetHareOutput(cdb, layerID)
+		bid, err := certificates.GetHareOutput(cdb, layerID)
 		if err != nil {
 			logger.With().Warning("failed to get hare output", layerID, log.Err(err))
 			return nil, fmt.Errorf("%w: get hare output %v", errMissingHareOutput, err.Error())
 		}
-		// hare output an empty layer
+		// hare output an empty layer, or the network equivocated with multiple valid certificate
```
dshulyak: turns out there is a case when multiple valid certificates can exist even without equivocation, so maybe equivocation can be omitted from this description

800 is an expected number of eligibility (so it can be 802 for example), and then both sides of 50/50 split can succeed in forming valid certs

fetch/handler.go
```diff
@@ -95,11 +96,27 @@ func (h *handler) handleLayerOpinionsReq(ctx context.Context, req []byte) ([]byt
...
+			if cert.Valid {
+				numValids++
+				if lo.Cert == nil && cert.Cert != nil {
+					lo.Cert = cert.Cert
+				}
+			}
+		}
+		if numValids > 1 {
+			lo.Cert = nil
+		}
```
fasmat: This for loop can be modified to terminate earlier if it finds more than one valid cert:

```suggestion
		var validCert *types.Certificate
		for _, cert := range certs {
			if !cert.Valid {
				continue
			}
			if validCert != nil {
				validCert = nil
				break
			}
			validCert = cert.Cert
		}
		lo.Cert = validCert
```

fasmat (commented):
LGTM, just added a comment for a possible performance improvement.

sql/certificates/certs.go
```diff
@@ -0,0 +1,121 @@
...
+		func(stmt *sql.Statement) {
+			stmt.BindInt64(1, int64(lid.Value))
+			stmt.BindBytes(2, bid[:])
+		}, nil); err != nil {
+		return fmt.Errorf("add wo cert %s: %w", lid, err)
+	}
+	return nil
+}
+
+// GetHareOutput returns the block that's valid as hare output for the spcified layer.
```
fasmat: Typo:

```suggestion
// GetHareOutput returns the block that's valid as hare output for the specified layer.
```

fasmat (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #3741, opened 2022-11-11
[Merged by Bors] - optimistic filtering: add two more conditions

### LINKED ISSUE #3378: block: correct conditions for optimistic filtering
## Description
currently we only check if there is consensus on mesh hash to enable optimistic filtering.
we need do two more
- We have confidence about every layer i < N-hdist (that is, for layers which we check using tortoise, we've passed the confidence threshold)
- We have hare output for every layer i such that N-hdist <= i <= N

### full quote from Tal in Slack:
we only do optimistic filtering at layer N if all of the following conditions are met:
- We have confidence about every layer i < N-hdist (that is, for layers which we check using tortoise, we've passed the confidence threshold)
- We have hare output for every layer i such that N-hdist <= i <= N
- The proposals in the output of  layer N have a large-enough majority that agrees that the first two conditions are met and on the cumulative layer hash for layer N (I don't remember the threshold, but let's say 9/10 of the proposals need to agree)

If our security assumptions hold, then once honest parties receive hare output it's guaranteed that the tortoise will confirm the blocks generated by the hare. Also, if the honest parties will all agree on the hare output, so either all honest parties will do optimistic filtering or none will.

If our security assumptions don't hold, then it's possible that some honest parties will do optimistic filtering and some won't (e.g., if they disagree on the hare output). It's also possible that blocks they were confident about it the past suddenly change (which is the case I think you're describing), and the layer hash they have for N is different from what's in the hare output.

Here's what I think the behavior should be in the broken security assumption cases (this isn't formally proved, it's a heuristic whose  purpose is to prioritize return to consensus):
- If block certifier detects that there's a problem of this type (there is hare output for layer N that indicates optimistic filtering should be performed, but the computed cumulative layer hash for N is different than the proposed filtering hash), then it should:
  - check with peers to see if it's missing data that would change its opinion of the correct layer hash for N
  - If the peers are not convincing (or agree that the layer hash is incorrect), it should certify an empty layer.
- If a node sees a valid certificate for two different blocks in layer N, it should vote for the empty layer (this should never happen if there's an honest majority)
- If a node detects that no valid certificate is possible in layer N (e.g., it sees more than 1/3 of the certifier weight supporting block A, and more than 1/3 supporting block B, for two different blocks A and B), it votes for the empty layer.
- If a node doesn't receive any valid certificate until zdist , this is treated like a hare failure (and the node votes for an empty layer).

countvonzero:
## Motivation

Closes #3378

## Changes

add two more conditions for enabling optimistic filtering for layer N
- node is confidence about every layer i < N-hdist
- node has hare output for every layer i such that N-hdist <= i <= N

these conditions are added to miner prior to recording its mesh hash for layer N-1 in its proposal.

note: this is done at proposal creation time instead of block creation time to prevent nodes from making different decisions on whether to do optimistic filtering and causing multiple blocks created for the same layer.


dshulyak (approved):

countvonzero:
bors merge

PR #3743, opened 2022-11-11
[Merged by Bors] - prevent double close panic

### LINKED ISSUE #3735: hare: panic close of closed channel
image

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3735

this closer should be dropped from the codebase, but i want a quick fix to prevent panics

### DIFF
--- a/common/util/closer.go
+++ b/common/util/closer.go
@@ -8,12 +8,12 @@ type Closer struct {
 }
 
 // NewCloser creates a new (not closed) closer.
+// DEPRECATED: use context.WithCancel.
 func NewCloser() Closer {
 	return Closer{make(chan struct{})}
 }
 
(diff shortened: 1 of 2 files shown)


common/util/closer.go
```diff
@@ -15,7 +15,11 @@ func NewCloser() Closer {
 // Close signals all listening instances to close.
 // Note: should be called only once.
```
fasmat: This comment can probably be removed now.

fasmat (commented):
The fix looks ok to me.

I think most uses of `Closer` can be replaced by a `context`. Should we mark this component as Deprecated and add a comment to use `context` instead?

fasmat (approved):

dshulyak:
bors merge

PR #3758, opened 2022-11-16
[Merged by Bors] - add block to tortoise before persisting it

### LINKED ISSUE #3750: computed opinion hash doesn't match signed
image

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3750

- tortoise.OnBlock is called before block is saved to the database
- tortoise will fail with error if ballot is trying to be decoded with missing block in the state (also missing block can be inserted from votes)

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -667,10 +667,12 @@ func (msh *Mesh) AddBlockWithTXs(ctx context.Context, block *types.Block) error 
 	msh.setLatestLayer(logger, block.LayerIndex)
 	logger.Debug("associated txs to block")
 
+	// add block to the tortoise before storing it
+	// otherwise fetcher will not wait until data is stored in the tortoise
+	msh.trtl.OnBlock(block)
 	if err := blocks.Add(msh.cdb, block); err != nil && !errors.Is(err, sql.ErrObjectExists) {
 		return err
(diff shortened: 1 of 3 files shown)


mesh/mesh.go
```diff
@@ -667,10 +667,12 @@ func (msh *Mesh) AddBlockWithTXs(ctx context.Context, block *types.Block) error
 	msh.setLatestLayer(logger, block.LayerIndex)
 	logger.Debug("associated txs to block")
 
+	// add block to the tortoise before storing it
+	// otherwise fetcher will not wait until data is stored in the tortoise
```
countvonzero, line 671: the comment `fetcher will not wait` isn't clear to me. can you elaborate a little bit more please? thanks

countvonzero (approved):

mesh/mesh.go
dshulyak: the assumption was that fetcher returns without error either:
- block is already available in the local database
- after block is downloaded from peer and passes handler without error

if block is saved before it is added to the tortoise then there is a race condition and tortoise.DecodeBallot can be called after block was saved to db but before OnBlock was called

dshulyak:
bors merge

PR #3762, opened 2022-11-17
[Merged by Bors] - vm: don't hold write lock until it is time to write accounts updates

dshulyak:
vm will take non-negligible cpu time to execute with increasing tx load. we are using sqlite that has a single write lock, so every other writer will have to wait.

it contributes to performance issues on the network

### DIFF
--- a/genvm/vm.go
+++ b/genvm/vm.go
@@ -89,7 +89,7 @@ type VM struct {
 func (v *VM) Validation(raw types.RawTx) system.ValidationRequest {
 	return &Request{
 		vm:      v,
-		cache:   core.NewStagedCache(v.db),
+		cache:   core.NewStagedCache(core.DBLoader{Executor: v.db}),
 		decoder: scale.NewDecoder(bytes.NewReader(raw.Raw)),
 		raw:     raw,
 	}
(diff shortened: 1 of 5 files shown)


dshulyak:
bors try

genvm/vm.go
```diff
@@ -223,6 +220,19 @@ func (v *VM) Apply(lctx ApplyContext, txs []types.Transaction, blockRewards []ty
...
+	}
+	defer tx.Release()
+
+	for _, reward := range rewardsResult {
+		if err := rewards.Add(tx, &reward); err != nil {
+			return nil, nil, fmt.Errorf("%w: %s", core.ErrInternal, err.Error())
+		}
+	}
+
 	ss.IterateChanged(func(account *core.Account) bool {
```
countvonzero, line 236: this change trigger a concern that probably exists before.

the fact that the `StagedCache` holds on to a db executor concerns me a little. here the code is calling `ss.IterateChanged` that execute sql stmts inside a db transaction, with the possibility that inside `ss.ItrateChanged` it could potentially use it's own `sql.Executor` to do something.

would it be ok to pass a `sql.Executor` to `StagedCache.Get` so that the staged cache doesn't need a member of `sql.Exectuor`?

countvonzero (approved):

genvm/vm.go
dshulyak: i passed interface to the cache

dshulyak:
bors merge

PR #3766, opened 2022-11-19
[Merged by Bors] - replace writes with reads in conservative cache

### LINKED ISSUE #3761: transaction mempool needs to be optimized
i am running workload for 10 tx/s

image

transaction mempool executes 600 queries per second, two of them are writes. we can't use database this way

image

### LINKED ISSUE #3765: improve performance of the conservative cache after applying layer
it has unreasonable number of db writes after applying transactions.
executing transactions and applying state is negligible comparing to the rest of applying layer processing time

image

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3765
close: https://github.com/spacemeshos/go-spacemesh/issues/3761

this pr removes UpdateIfBetter, SetNextLayer, DiscardNonceBelow and DiscardByAcctNonce functions.
all of them were a problem either when transactions are accepted into mempool, or after executing a layer. writes are in general expensive, all 4 of those writes are conditional and complex. if more information is required for API it needs to be implemented in a way that doesn't harm consensus critical path.

instead of UpdateIfBetter/SetNextLayer there are two additional queries TransactionInProposal/TransactionInBlock.

DiscardNonceBelow/DiscardByAcctNonce are not necessary if `applied` field is not maintained in transactions table. also i dropped DISCARDED/PROPOSAL/BLOCK transaction states, will open an issue to bring them back if necessary.


dshulyak:
image

sql/transactions/transactions.go
```diff
@@ -262,43 +153,36 @@ func decodeTransaction(id types.TransactionID, applied int, stmt *sql.Statement)
...
+	} else if parsed.TxHeader != nil {
+		state = types.MEMPOOL
 	}
+	var bid types.BlockID
+	stmt.ColumnBytes(3, bid[:])
+
 	return &types.MeshTransaction{
 		Transaction: parsed,
-		LayerID:     lid,
+		LayerID:     types.NewLayerID(uint32(stmt.ColumnInt64(2))),
```
countvonzero: nit: capture the layer id in line 159?

sql/transactions/transactions.go
```diff
@@ -262,43 +153,36 @@ func decodeTransaction(id types.TransactionID, applied int, stmt *sql.Statement)
...
 		BlockID:     bid,
 		Received:    time.Unix(0, stmt.ColumnInt64(4)),
 		State:       state,
 	}, nil
 }
 
 // Get gets a transaction from database.
+// Layer and Block fields are if transaction was applied.
```
countvonzero: :s/are/are set/

sql/transactions/transactions.go
```diff
@@ -449,3 +340,47 @@ func AddResult(db *sql.Tx, id types.TransactionID, rst *types.TransactionResult)
 	}
 	return nil
 }
+
+// TransactionInProposal returns lowest layer of the proposal where tx is included after the specified layer.
+func TransactionInProposal(db sql.Executor, id types.TransactionID, after types.LayerID) (types.LayerID, error) {
+	var rst types.LayerID
+	rows, err := db.Exec("select layer from proposal_transactions where tid = ?1 and layer > ?2 order by layer asc limit 1",
```
countvonzero, line 346: does min(layer) work here?

countvonzero (approved):
thanks for fixing it. much appreciated.

i didn't review the changes in systests/genvm/hare. i assume those are for different PRs.

countvonzero (commented):
just want to say that it's so great that tests stay the same. thanks

sql/transactions/transactions.go
dshulyak: i guess it does, but i wanted it to be consistent with query in TransactionInBlock. and there i would have to use group by block, which would be more complex if compared with current version

sql/transactions/transactions.go
```diff
@@ -369,37 +246,51 @@ func GetByAddress(db sql.Executor, from, to types.LayerID, address types.Address
...
-// GetAllPending get all transactions that are not yet applied.
-func GetAllPending(db sql.Executor) ([]*types.MeshTransaction, error) {
-	return queryPending(db, `
-		select tx, header, layer, block, timestamp, id from transactions
-		where applied = ?1 and header is not null order by timestamp asc`,
-		func(stmt *sql.Statement) {
-			stmt.BindInt64(1, statePending)
-		}, "get all pending")
+// AddressesWithPendingTransactions returns list of addresses with pending transactions.
+// Query is expensive, meant to be used only on startup.
```
countvonzero, line 250: note that this will also be called when a revert occurs.

countvonzero:
i'd like to use code in this PR to fix https://github.com/spacemeshos/go-spacemesh/issues/3805

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #3782, opened 2022-11-22
[Merged by Bors] - hare: make beacon syncedness criteria for hare participation and msg validation

### LINKED ISSUE #2937: hare: `panic: tried to update a previous layer`
## Description

I had been running a synced node connected to the @brusherru's devnet 1 on `dshulyak/fix-organizer-bug` at `c8d509bebf029a3b1be49ce3007684af0f6691a9` (the branch is even with `develop` at the time of creating the issue) for 4.5 hours, then it crashed with the following panic:
```
panic: tried to update a previous layer

goroutine 1200 [running]:
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc0001b0840, {0xc0034f66c0, 0x3, 0x3})
	/Users/nkryuchkov/go/pkg/mod/go.uber.org/zap@v1.19.0/zapcore/entry.go:232 +0x446
go.uber.org/zap.(*Logger).Panic(0xc00d36ddc0, {0x4d6276a, 0x0}, {0xc0034f66c0, 0x3, 0x3})
	/Users/nkryuchkov/go/pkg/mod/go.uber.org/zap@v1.19.0/logger.go:230 +0x59
github.com/spacemeshos/go-spacemesh/log.FieldLogger.Panic({0xc00a26e000, {0x4d3d947, 0x4ff1c20}}, {0x4d6276a, 0x20}, {0xc00326f6e8, 0x2, 0x2})
	/Users/nkryuchkov/GolandProjects/go-spacemesh/log/zap.go:258 +0x1b9
github.com/spacemeshos/go-spacemesh/hare.(*Broker).updateLatestLayer(0xc00041f100, {0x500e0b8, 0xc0085ae390}, {0x0})
	/Users/nkryuchkov/GolandProjects/go-spacemesh/hare/broker.go:341 +0x272
github.com/spacemeshos/go-spacemesh/hare.(*Broker).Register.func1()
	/Users/nkryuchkov/GolandProjects/go-spacemesh/hare/broker.go:405 +0x8c
github.com/spacemeshos/go-spacemesh/hare.(*Broker).eventLoop(0xc00041f100, {0x500e0b8, 0xc009723dd0})
	/Users/nkryuchkov/GolandProjects/go-spacemesh/hare/broker.go:332 +0x4fc
created by github.com/spacemeshos/go-spacemesh/hare.(*Broker).Start
	/Users/nkryuchkov/GolandProjects/go-spacemesh/hare/broker.go:97 +0x131
```

Here are the last few minutes of logs:
[panic1.log](https://github.com/spacemeshos/go-spacemesh/files/7530839/panic1.log)

### LINKED ISSUE #3776: hare: do not validate hare messages before synced
### tldr
only start processing hare messages after a node is synced

### context
```
kimmy
quick question on hare.
when a node is not sync'ed and therefore cannot participate in hare, should it process any hare messages at all?

Noam Nelke
I don’t see why not passively listen. But we may not be able to validate the eligibility if we don’t have ATXs. 
You mean after we have ATXs, but not other objects?

kimmy
when the node doesn't have beacon, it cannot validate any roles
even tho it has ATXs
what good does "passively listen" do?

Noam Nelke
Yeah, that was my point. We should strive to be able to validate Hare messages as quickly as possible, so 
prioritize ATXs and beacon messages (and fraud proofs when we have those). Then we can passively listen 
(run Hare but not participate- because we don’t necessarily have proposals).

kimmy
i see. we want to help the consensus converge faster even if the node didn't participate at the beginning 
of the hare

Wojciech Klaudiusz Zaborowski
I see the ‚passive particaption’ as unnecessary complicationn of code (and spec). It rather feels right to 
keep things simple in this case. Finish sync, then participate.
The ability to run eligibility validation is a prerequisite to meaningful participation.

Tal Moran
I agree with @wojtek in this instance, at least until we have real-world data that shows this is a significant 
bottleneck. 
I also agree with @noam that we should try to be able to validate gossip messages as soon as possible 
by prioritizing our sync; I foresee that this will also be very helpful in potential "recovery playbook" --- 
for example, it means that we can spin up new gossip nodes quickly if there are network connectivity problems.
Note that in order for the prioritization to help, we need to have a granular sync state (i.e., a node should be 
able to check if it can validate a message even if it is not fully synced)

Dmitry Shulyak
eligibility validation depends on the ballots and activations specified in active sets of the ballots. so it is not 
only activations and beacons. in the existing implementation node has to be fully synced

kimmy 
@Dmitry Shulyak do you mean ballots in the context of beacon sync?
i cannot remember what data we need from ballots for hare participation other than the beacon

Dmitry Shulyak
func (o *Oracle) actives(ctx context.Context, targetLayer types.LayerID

kimmy
right. we need ballots for the safe range layers relative to the layer roles are being verified.
i will disable verifying hare msgs for not-synced nodes for now.
```

countvonzero:
## Motivation

Closes #2937
Closes #3776

## Changes

- remove unnecessary turbo hare
- serialize hare onTick() for each layer
- remove caching for syncedness for each layer. there is really no need for this cache
- add beacon syncedness as a criteria for hare participation and message validation


countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #3784, opened 2022-11-23
[Merged by Bors] - tortoise: store ballots separately from layers

### LINKED ISSUE #3779: tortoise: stored ballots are not properly evicted
image

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3779

each ballot stores votes on X previous layers. ballots referenced in the layer store votes for another X layers and so on. this is clearly recursive and can't be used like this.


tortoise/state.go
```diff
@@ -62,8 +62,9 @@ type (
 			min, max types.LayerID
 		}
 
-		epochs map[types.EpochID]*epochInfo
-		layers map[types.LayerID]*layerInfo
+		epochs  map[types.EpochID]*epochInfo
+		layers  map[types.LayerID]*layerInfo
+		ballots map[types.LayerID][]*ballotInfo
```
countvonzero, line 71: it's definitely worth adding a comments about this recursive behavior.
i think it'll be very tempting for future dev to do the same thing without realizing the subtleties.

countvonzero (approved):

tortoise/state.go
dshulyak: definitely, added a comment

dshulyak:
memory consistently around 300-400 mb. seems a bit high, but that can be explored separately

image

dshulyak:
bors merge

dshulyak:
bors merge

Issue #3458, opened 2022-08-18
poet: allow one submission of PoST per miner per round per poet service


countvonzero:
## Description

condensing Noam's comments in slack:

To enforce this, the node will need to submit the entire “poet challenge” (NIPostChallenge) to the PoET and it will calculate the hash itself (could be good to return the hash in the response)

This part is not trivial, I think we may want the PoET to ask a node (it already has access to the “PoET gateway” - maybe we should rename it) to validate this data structure.

If this is a smesher’s first ATX, then the NIPostChallenge includes the initial PoST proof - which needs to be validated.

If it’s a subsequent ATX, the NIPostChallenge only includes a reference to the PrevATXID - that previous ATX includes the PoST proof that came after the previous PoET run (we only do one PoST proof each epoch). So validation here must be contextual - the challenge only tells the PoET which smesher it is, but to validate that they still expend resources, we must actually inspect the mesh. This is trivial for the node to do, but not for the context-less PoET.

## Tasks:

- [x] #3706
- [x] #3707
- [x] #3708

moshababo:
## Design proposal

### Upon registration, the node should provide:
* `PositioningATX`
* `PubLayerID`
* `InitialPost` + its metadata (for an initial ATX registration)
* `PrevATXID` (for a non-initial ATX registration)
* Signature for a msg containing all other fields

### PoET service steps:
* Verify signature and extract pubkey (as `NodeID`). The reason for this origin check is that although the utility of the PoET results is bound to the specified `NodeID`, a non-honest user can register for a given round on behalf of some other node, while providing request data that will eventually yield a non-valid ATX, and so practically preventing the node from using the PoET service for a given round.  
* Verify `NodeID` to be unregistered for the upcoming round. 
* For an initial ATX
	* Verify the validity of `InitialPost` with respect to the metadata and `NodeID`. [^1]
	* Verify the metadata with respect to consensus rules. [^2]
* For a non-initial ATX
	* Fetch `PrevATXID` ATX. [^3]
	* Verify `NodeID` to match the prev ATX's `NodeID`. 
	* Calculate `Sequence` according to the prev ATX's `Sequence`. 
* Calculate the challenge hash and apply it for the upcoming round
	* For an initial ATX: `sha256( NodeID ++ PubLayerID ++ PositioningATX ++ InitialPostIndices )` [^4]
[...]

PR #3789, opened 2022-11-23
[Merged by Bors] - Support VerifyChallenge in GatewayService

poszu:
## Motivation
Implemented service for verifying poet challenges by its trusted gateway. Poet is supposed to relay submitted challenges to its gateway for verification and hashing.
Part of #3458 

## Test Plan
- unit tests
- tested e2e (systests) with updated poet POC using this service

## DevOps Notes


poszu:
bors try

poszu:
bors try

activation/challenge_verifier.go
```diff
@@ -0,0 +1,181 @@
...
+		signatureVerifier: signatureVerifier,
+		cfg:               cfg,
+		goldenATXID:       goldenATX,
+	}
+}
+
+func (v *challengeVerifier) Verify(ctx context.Context, challengeBytes, signature []byte) (*ChallengeVerificationResult, error) {
+	pubkey, err := v.signatureVerifier.Extract(challengeBytes, signature)
+	nodeID := types.BytesToNodeID(pubkey.Bytes())
+	if err != nil {
```
countvonzero, line 67: should check this err before converting to node ID

activation/challenge_verifier.go
```diff
@@ -0,0 +1,181 @@
+package activation
```
countvonzero, line 1: i am not sure this should be in activation package. i think it should probably stay in grpcserver pkg as this is a generic gateway service

activation/challenge_verifier.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (v *challengeVerifier) verifyChallenge(ctx context.Context, challenge *types.PoetChallenge, nodeID types.NodeID) error {
+	log.With().Info("verifying challenge", log.Object("challenge", challenge))
+
+	if err := validateNumUnits(&v.cfg, challenge.NumUnits); err != nil {
+		return fmt.Errorf("%w: %v", ErrChallengeInvalid, err)
+	}
+
+	// Verify against positioning ATX
```
countvonzero: the verification logic should share code with the atx handler. there should only be some small part that verifies that verifies node's own id that's different

activation/challenge_verifier.go
poszu: It's not generic. It's nipost-challenge-specific. IMO, `grpcserver` package should only contain grpc-related code (parsing input, turning the result into a grpc status etc.) without any business logic. The code in `challenge_verifier.go` is business logic, it needs to know what are the constraints on the nipost challenge etc.

activation/challenge_verifier.go
poszu: Factored out nipost challenge validation.

activation/challenge_verifier.go
```diff
@@ -0,0 +1,155 @@
...
+	log.With().Info("verifying challenge", log.Object("challenge", challenge))
+
+	if err := validateNumUnits(&v.cfg, challenge.NumUnits); err != nil {
+		return fmt.Errorf("%w: %v", ErrChallengeInvalid, err)
+	}
+
+	if err := validatePositioningAtx(&challenge.PositioningATX, v.atxDB, v.goldenATXID, challenge.PubLayerID, v.layersPerEpoch); err != nil {
+		switch err.(type) {
+		case *AtxNotFoundError:
+			return &CouldntVerifyError{source: err}
```
countvonzero, line 101: i don't understand why use a special error in the case ATX is not found. how does poet server react to two different errors?

poszu: I want to differentiate between:
- could not verify the challenge,
- the challenge is invalid.

In the former case, Poet will retry to verify the challenge with a different gateway (or the same one, after some time).

The ChallengeVerifier GRPC service logic doesn't know the verification logic and thus is not able to distinguish the errors. For its sake, I wrap every _retryable_ error in `CouldntVerifyError` so that it knows how to respond to the Poet.

countvonzero: ideally you can signal retryable error only when the node is not ATX synced. a node is expected to have all the ATXs in the network when it is ATX synced (barring the case of network split).

but i can compromise if you limit the retry attempts in poet server. i just want to avoid this become a dos vector.

activation/challenge_verifier.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/signing"
+)
+
+var (
+	ErrSignatureInvalid = errors.New("signature is invalid")
+	ErrChallengeInvalid = errors.New("challenge is invalid")
+)
+
+type CouldntVerifyError struct {
```
countvonzero, line 20: it should suffice to use fmt.errorf for this and ATXNotFound too.
i don't see the need to create two custom errors.

countvonzero (approved):

activation/challenge_verifier.go
poszu: Poet retries are limited. 

As discussed offline - let's consider distinguishing between synced/not synced and returning invalid/retry status respectively an optimization for the future.

poszu:
bors merge

PR #3791, opened 2022-11-24
[Merged by Bors] - mesh: query block validity updates from tortoise

### LINKED ISSUE #3744: mesh: block validity updater is not robust
### current approach:
- mesh will apply the hare output `A` for layer `N` to the mesh optimistically when hare terminates
- tortoise verifies layer `N` (earliest at layer `N+1`) and voted against A, supported B, it triggers mesh to set `minUpdatedLayer` to `N`
- when mesh is processing layer `N+2`, it sees the `minUpdatedLayer` is `N`, it discovers that it applied the wrong block, reverts the state to layer `N-1`, and applies `B` for layer `N`.

### problem

1.  updating `minUpdatedLayer` and saving validity to db aren't atomic. if the node is checking for revert between this window, it'll be incorrect. (`UpdateBlockValidity` doesn't hold a lock)
2. if the node exits right after the block validity updated in db, upon restart, mesh will not re-examined if the applied block is correct
3. if tortoise changed its opinion about the same layer within a narrow window of resetting `minUpdatedLayer`, then that 2nd opinion will be ignored

a memory-only solution will not work for 2.

### proposed fix
- add a boolean field `dirty` to blocks table.
- when tortoise verifies layer `N`, it immediately saves block validity for `A` and `B` into database and mark their `dirty` field to true
- when mesh process layer `N+2`
  1. query all the dirty blocks from db and found `A` is invalid and `B` is valid
  2. re-examines the applied block in `N`, reverts to layer `N-1`, and applies `B`
  3. sets `A`'s `dirty` to false iff `A`'s `valid` is still false, and sets `B`'s `dirty` to false iff `B`'s `valid` is still true

countvonzero:
## Motivation

Closes #3744

## Changes

- tortoise: keep a `[]types.BlockContextualValidity` in memory until being queried
- mesh: pull block validity updates instead of providing a callback, process the updates to check for possible state revert, and persist the validity to db


countvonzero:
bors try

tortoise/tortoise.go
```diff
@@ -24,13 +24,18 @@ var (
 	errstrTooManyExceptions = "too many exceptions to base ballot vote"
 )
 
+type blockValidityUpdater interface {
```
dshulyak: does it need to use updater? i thought that we can either:
- drop interface and use db directly in tortoise
- drop writes in tortoise (more changes in tests i guess)

countvonzero: > does it need to use updater? 

i keep it mostly to keep the current tests unchanged.

> i thought that we can either:
> 
> * drop interface and use db directly in tortoise
> * drop writes in tortoise (more changes in tests i guess)

i can write the validity updates to db after `Updates()` is queried. that will allow dropping the interface and not changing the tests. in mesh.go i currently write to db in a defer function anyway.

countvonzero:
bors try

countvonzero:
bors try

tortoise/tortoise.go
countvonzero: @dshulyak 
i ended up dropping writes in tortoise and changed the tests. please review. thank you

common/types/block.go
```diff
@@ -175,6 +175,7 @@ func ToBlockIDs(blocks []*Block) []BlockID {
 // BlockContextualValidity tuple with block id and contextual validity.
```
dshulyak: comment

common/types/block_scale.go
```diff
@@ -178,6 +178,13 @@ func (t *BlockContextualValidity) EncodeScale(enc *scale.Encoder) (total int, er
 		}
 		total += n
 	}
+	{
```
dshulyak: weird that we generate scale for this type, can you please add a whitelist for Block and what else is needed?

tortoise/algorithm.go
```diff
@@ -147,6 +146,20 @@ func (t *Tortoise) LatestComplete() types.LayerID {
 	return t.trtl.verified
 }
 
+func (t *Tortoise) Updates() (types.LayerID, []types.BlockContextualValidity) {
```
dshulyak, line 150: why it needs to return verified layer? can mesh work just based on updates? empty layer will have empty list

tortoise/algorithm.go
```diff
@@ -147,6 +146,20 @@ func (t *Tortoise) LatestComplete() types.LayerID {
...
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	res := make([]types.BlockContextualValidity, len(t.trtl.updated))
+	copy(res, t.trtl.updated)
+	return t.trtl.verified, res
+}
+
+func (t *Tortoise) UpdatesPersisted(persisted []types.BlockContextualValidity) {
```
dshulyak: i think this interface is not necessary. tortoise created an update event, and it won't create it again unless opinion changes. it doesn't need to know if that event was persisted

if you want persisted can be renamed to emitted

tortoise/tortoise.go
```diff
@@ -54,7 +53,7 @@ func newTurtle(
 		logger:  logger,
 		cdb:     cdb,
 		beacons: beacons,
-		updater: updater,
+		updated: make([]types.BlockContextualValidity, 0),
```
dshulyak: updated can be nil

mesh/mesh.go
```diff
@@ -281,30 +225,40 @@ func (msh *Mesh) setProcessedLayer(layerID types.LayerID) error {
...
+			}
+			return nil
+		})))
 
-	msh.resetMinUpdatedLayer(minUpdated)
-	if minUpdated.After(msh.LatestLayerInState()) {
-		// nothing to do
-		return nil
+	inState := msh.LatestLayerInState()
+	byLayer := make(map[types.LayerID][]types.BlockContextualValidity)
```
dshulyak: it is also possible to maintain a map of updates in the tortoise if you need it instead of list

in general i would try to avoid transformations and prepare a data structure that works for intended use case

common/types/block.go
```diff
@@ -175,6 +175,7 @@ func ToBlockIDs(blocks []*Block) []BlockID {
...
 type BlockContextualValidity struct {
```
dshulyak, line 176: tortoise can return map[LayerId][]BlockContextualValidity

dshulyak (commented):
> note that if any errors occur when processing validity, say state failed to revert, block validity will still be persisted. in that case the node require peer hash resolution or a node restart to attempt state revert again

i am not sure if this is robust. sync should not be on the consensus path

tortoise/tortoise.go
dshulyak: also in general make with 0 doesn't make sense, it is unfortunate that this syntax is allowed

tortoise/algorithm.go
dshulyak: it definitely correct to work on the list of updates, originally i proposed this type to track empty layers. but map[Layer][]Block also works

```go
type Event struct {
    Layer uint32
    []BlockHeader
}
```

layerInfo can be extended with emittedEmpty bool if layer blocks are empty and layer is verified

countvonzero: what i want to achieve is to get a consistent view of tortoise, as now there are two separate threads that can cause verified layer to change (proposal builder and mesh).

so here verified layer and the list of updates are atomic, the information also feeds to `getLayerValidBlocks` to pick the block to apply.

countvonzero:
> > note that if any errors occur when processing validity, say state failed to revert, block validity will still be persisted. in that case the node require peer hash resolution or a node restart to attempt state revert again
> 
> i am not sure if this is robust. sync should not be on the consensus path

this is no longer true for the current implementation. if state failed to revert, block validity will not be persisted.

tortoise/algorithm.go
countvonzero: ok. but there is one corner case:

when mesh process the updates, and it fails to revert the state accordingly, those block validity will not be persisted to database. in that case, tortoise also won't emit those updates again. in that case, mesh will not try to check those layers again, until syncer realizes it has different mesh hash.

i could accumulate the updates in mesh.go if the failure happen and remove this interface.

mesh/mesh.go
countvonzero: for the corner case stated here https://github.com/spacemeshos/go-spacemesh/pull/3791#discussion_r1032649571
i ended up keeping a map in mesh.go to remember update failures. so i keep tortoise interface as a slice as it's simpler.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/actions/runs/3550608057/jobs/5964766995)

this first run finished in 19 minutes. (normal time to run system test II)
logs inspection didn't find problems. don't know why the exit code is not 0.

retry succeeded
https://github.com/spacemeshos/go-spacemesh/actions/runs/3550608057

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/actions/runs/3551265312/jobs/5965340415)
```
=== Failed
=== FAIL: p2p/peerexchange TestHost_ReceiveAddressesOnCheck (4.01s)
    bookchecker_test.go:179: 
        	Error Trace:	/home/runner/work/go-spacemesh/go-spacemesh/p2p/peerexchange/bookchecker_test.go:179
        	Error:      	Condition never satisfied
        	Test:       	TestHost_ReceiveAddressesOnCheck
        	Messages:   	nodeA and nodeB should have only 1 address

DONE 1944 tests, 6 skipped, 1 failure in 297.264s
make: *** [Makefile:126: test] Error 1
Error: Process completed with exit code 2.
```

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
@dshulyak tested with the latest develop. extended the wait time in 50/50 split by 1 more epoch to reduce flakiness.

tortoise/algorithm.go
dshulyak: my point was that there is no need to read verified layer at all. the least of updates should be enough for mesh

dshulyak (approved):
> extended the wait time in 50/50 split by 1 more epoch to reduce flakiness.

i think it means that something doesn't work as it should

proposals/handler_test.go
```diff
@@ -384,7 +384,7 @@ func TestBallot_BallotDoubleVotedOutsideHdist(t *testing.T) {
 			require.Equal(t, b.ID(), ballot.ID())
 			return true, nil
 		})
-	th.mm.EXPECT().AddBallot(b).Return(nil)
+	th.mm.EXPECT().AddBallot(context.TODO(), b).Return(nil)
```
poszu: `context.Background()` is meant to be used as the top-level parent context (in main, tests etc.), see: https://pkg.go.dev/context#Background

>Background returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.
```suggestion
	th.mm.EXPECT().AddBallot(context.Background(), b).Return(nil)
```

proposals/handler_test.go
```diff
@@ -783,8 +783,8 @@ func TestProposal_DuplicateTXs(t *testing.T) {
 			require.Equal(t, p.Ballot.ID(), ballot.ID())
 			return true, nil
 		})
-	th.mm.EXPECT().AddBallot(&p.Ballot).Return(nil).DoAndReturn(
-		func(got *types.Ballot) error {
+	th.mm.EXPECT().AddBallot(context.TODO(), &p.Ballot).Return(nil).DoAndReturn(
```
poszu: What's the meaning of `.Return(nil).DoAndReturn(...)`? Does it return `nil` the first time or is `Return(nil)` simply overwritten? If the latter, then:
```suggestion
	th.mm.EXPECT().AddBallot(context.TODO(), &p.Ballot).DoAndReturn(
```

tortoise/verifying.go
```diff
@@ -88,12 +88,12 @@ func (v *verifying) verify(logger log.Log, lid types.LayerID) bool {
...
 	logger = logger.WithFields(
 		log.String("verifier", "verifying"),
-		log.Stringer("candidate layer", lid),
+		log.Stringer("candidate_layer", lid),
 		log.Stringer("margin", margin),
 		log.Stringer("uncounted", uncounted),
 		log.Stringer("total good weight", v.totalGoodWeight),
 		log.Stringer("good uncounted", layer.verifying.goodUncounted),
-		log.Stringer("global threshold", threshold),
+		log.Stringer("global_threshold", threshold),
```
poszu: Field naming is inconsistent now - some use `_` and some - ` `.

poszu (commented):
Some nitpicks

tortoise/verifying.go
dshulyak: thanks didn't notice it. not sure why Kimmy changed it, probably should be reverted

countvonzero: changed it so it is consistent with the full mode when inspecting logs

tortoise/algorithm.go
countvonzero: yes. i can usually work out the verified layer from the list of updates. but for empty layer that is tricky.
say
t0: at layer 112 hare failed (network flakiness or whatever)
t1: at layer 113 tortoise verified layer 112 to be an empty layer, but the list of updates would be empty

the mesh won't be able to determine that it needs to apply an empty layer without knowing that tortoise actually has verified to 112.
(until later synced a certificate for the EmptyBlock ID, but this is out-of-protocol)

countvonzero:
> > extended the wait time in 50/50 split by 1 more epoch to reduce flakiness.
> 
> i think it means that something doesn't work as it should

few considerations:
- this test is already flaky before this change
- during testing, whenever i found a failure, the network recovery worked as expected, but the state queries in test happened at the time some nodes were reverting.

i'll test the shorter recovery time more to be sure. but i think it's worth extending the time to reduce flakiness if nothing else seems wrong.

tortoise/algorithm.go
dshulyak: i don't understand what is the problem
there will be a list: [{111, [{a, true}]} {112, []}]
111 is not empty, block a is valid
112 is empty, no known blocks

also i am not sure how current variant works
verified layer can move back if expected weight increased, but actual weight didn't

countvonzero: oh. i had thought that {112, []} will _not_ be in the updates from one of our previous discussion.
ok. in that case i will drop the verified layer from the API. thanks.

countvonzero: > also i am not sure how current variant works
verified layer can move back if expected weight increased, but actual weight didn't

@dshulyak 
say verified layers moved to 100 when we process layer 113, the updates will be still [{111, [{a, true}]} {112, []}] ?
if so, in that case, i expect tortoise will output updates btwn 100 and 112 (if its opinions changed) at a later time. and mesh can check revert accordingly.

proposals/handler_test.go
countvonzero: changed for all instances in this file

proposals/handler_test.go
countvonzero: not intentional. by the fact that tests were passing, i'd say it was overwritten. fixed.

tortoise/verifying.go
countvonzero: so the goal here is that the field name is consistent btwn verifying.go and full.go
i've updated the fields here to use "_". will leave those "-" untouched in this pkg.

tortoise/algorithm.go
countvonzero: i just realized, while changing the code, that i need to know the latest verified layer anyway if tortoise didn't make any new progress (i.e. empty updates)
previously this is accomplished by calling `tortoise.LatestComplete()`. here i condense it in one call `tortoise.Updates()`.
do you object to leaving the code as is? @dshulyak

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

tortoise/algorithm.go
countvonzero: anyhow, i'll merge as is. please file an issue if you disagree.

countvonzero:
bors merge

countvonzero:
bors merge

tortoise/algorithm.go
dshulyak: sorry i was away when you posted last messages

> here i condense it in one call tortoise.Updates().
do you object to leaving the code as is?

i still think that this is not needed. tortoise can be tweaked to emit event when layer is terminated as empty. i also wrote it above

Issue #3748, opened 2022-11-14
smesher voted twice in the same layer


dshulyak:
this is super critical, and should never be allowed
the reason behind is unclear, but it might be due to receiving several signals from the same layer. or because operator accidentally spawned several processes

![image](https://user-images.githubusercontent.com/1964576/201616938-172ab5bd-7905-4e3e-8ee2-a101efaf02fd.png)

## implementation

1. verify that ballot in this layer wasn't created before creating a ballot 
2. add a mutex for creating a ballot (if it doesn't exist)
3. log error if ballot initializer received signal for the layer several time
4. add flock at /DATADIR/LOCK to guarantee that operator didn't accidentally spawned two processes

brusherru:
Probably it also causes super slow node syncing.
Or at least it reruns tortoise really often and consumes a lot of CPU for it (up to 400% aka 4 cores entirely).
<img width="986" alt="image" src="https://user-images.githubusercontent.com/1897530/202241440-f3606292-6a7b-4ad3-8e71-69a845837bfd.png">

Tortoise keeps finding malicious ballots in each layer :-p
```
...
2022-11-16T16:37:05.031Z	INFO	6a1f7.trtl         	switching tortoise mode	{"node_id": "6a1f7a84dfa6d53428aa48a0a2a3ae62574743911b15fa479f285afd8e5cdd85", "module": "trtl", "rerun": true, "layer_id": 9118, "last_layer": "9155", "processed_layer": "9118", "verified_layer": "8116", "from_mode": "rerun_full", "to_mode": "rerun_verifying", "name": "trtl"}
2022-11-16T16:37:05.033Z	WARN	6a1f7.trtl         	observed malicious ballot	{"node_id": "6a1f7a84dfa6d53428aa48a0a2a3ae62574743911b15fa479f285afd8e5cdd85", "module": "trtl", "rerun": true, "ballot_id": "d3f77533a2", "layer_id": 9119, "name": "trtl"}
2022-11-16T16:37:05.034Z	INFO	6a1f7.trtl         	switching tortoise mode	{"node_id": "6a1f7a84dfa6d53428aa48a0a2a3ae62574743911b15fa479f285afd8e5cdd85", "module": "trtl", "rerun": true, "layer_id": 9119, "last_layer": "9155", "processed_layer": "9119", "verified_layer": "8117", "from_mode": "rerun_verifying", "to_mode": "rerun_full", "name": "trtl"}
[...]

PR #3792, opened 2022-11-24
[Merged by Bors] - run only one spacemesh instance

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/3748

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -12,6 +12,7 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/gofrs/flock"
 	grpcmiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpczap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
 	grpcctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
@@ -61,6 +62,7 @@ import (
 const (
 	edKeyFileName   = "key.bin"
 	genesisFileName = "genesis.json"
+	lockFile        = "LOCK"
 )
 
 // Logger names.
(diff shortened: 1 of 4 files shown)


pigmej (approved):
Ideawise Lgtm.

cmd/node/node.go
```diff
@@ -297,6 +300,15 @@ func (app *App) Initialize() (err error) {
 	if err := os.MkdirAll(app.Config.DataDir(), 0o700); err != nil {
 		return fmt.Errorf("ensure folders exist: %w", err)
 	}
+	lockName := filepath.Join(app.Config.DataDir(), lockFile)
+	fl := flock.New(lockName)
+	locked, err := fl.TryLock()
```
countvonzero, line 305: curious about the difference btwn using flock package and using os package directly to create an empty file

pigmej: @countvonzero it's not only about creation of empty file. It's atomic file lock. (on some platforms it will create directory instead because of API differences).

That should prohibit running two instances even in racy-ish conditions.

countvonzero: oh. i guess i am only thinking about linux OS. thanks

poszu (commented):
What if the lock file is not unlocked before quitting, for example because of panic or SIGKILL. The app will fail to start again, right? Is there any recovery?

poszu:
> What if the lock file is not unlocked before quitting, for example because of panic or SIGKILL. The app will fail to start again, right? Is there any recovery?

Answering myself - at least on Linux, according to https://linux.die.net/man/2/flock, the file lock will be released when all file descriptors are closed:
> Furthermore, the lock is released either by an explicit LOCK_UN operation on any of these duplicate descriptors, or when all such descriptors have been closed.

dshulyak:
bors merge

PR #3793, opened 2022-11-24
[Merged by Bors] - check if smesher already casted a vote in the handled layer

### LINKED ISSUE #3748: smesher voted twice in the same layer

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3748

### DIFF
--- a/miner/proposal_builder.go
+++ b/miner/proposal_builder.go
@@ -32,9 +32,10 @@ const (
 )
 
 var (
-	errGenesis   = errors.New("not building proposals: genesis")
-	errNotSynced = errors.New("not building proposals: node not synced")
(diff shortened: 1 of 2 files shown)


countvonzero (approved):

dshulyak:
bors merge

Issue #3459, opened 2022-08-18
p2p: gossip traffic prioritization


countvonzero:
## Description

for genesis, we are not implementing phased poet servers. so it's very likely that ATXs will be completed/published at the same time.
in p2p, we need to prioritize different gossip messages. for example. hare messages are time-sensitive so should have the highest prioritization.

@noamnelke @dshulyak please elaborate the exact prioritization of our message and maybe give more detailed implementation instructions if possible.

dshulyak:
this will need to be implemented in the gossipsub fork if we will do it before genesis

dshulyak:
before trying to implement priority queuing (it would require forking gossipsub, or custom streaming protocol), we should try to use two (or three) separate routers:
- fast router for hare and proposals (uses flood and eager push)
- slow router for atxs and beacon (only lazy pull)
- maybe separate txs into another slow router (also only lazy pull)

the goal is to reduce duplicate atxs and beacon messages in the gossip, and also don't share queues that are used for protocol and transactions.

PR #3798, opened 2022-11-25
[Merged by Bors] - track gossip payload/messages by protocol

### LINKED ISSUE #3305: p2p: capture metrics for per-protocol gossip
missing information on how many messages/bytes were processed by every sub-protocol in the gossip router

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/3459
close: https://github.com/spacemeshos/go-spacemesh/issues/3305

### DIFF
--- a/p2p/metrics/gossipsub.go
+++ b/p2p/metrics/gossipsub.go
@@ -16,13 +16,39 @@ var (
 	peersPerProtocol = metrics.NewGauge("peers_per_protocol", subsystem, "Number of peers per protocol", []string{"protocol"})
 )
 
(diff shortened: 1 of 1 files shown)


pigmej (approved):
Huge +1 from my side to be able to asses better gossip.

One question though, would it be possible to add hare per type also? Like preroud, status, proposal etc?

dshulyak:
> One question though, would it be possible to add hare per type also? Like preroud, status, proposal etc?

due to the implementation it is not easy. but should be possible though with some additional hacks

dshulyak:
bors merge

PR #3799, opened 2022-11-25
Remove Context from testcontext.Context

poszu:
## Motivation
Keeping a `context.Context` inside of a struct is not recommended and leads to problems, for example with creating ad-hoc contexts for cancellation of short term operations (i.e timing out).

## Changes
Removed `context.Context` field from `testcontext.Context` and pass it as a separate argument all over systest.

Additionally, renamed all uses of `testcontext.Context` to `tctx`.

## Test Plan
- systests still passing

## DevOps Notes


dshulyak (commented):
would it be better to add a method Context() context.Context to testcontext.Context object that will return context with timeout?

poszu:
> would it be better to add a method Context() context.Context to testcontext.Context object that will return context with timeout?

It wouldn't help. Even now it is possible to create a context with timeout from `testcontext.Context` by calling `context.WithTimeout(tctx,...)`. The problem is to pass it down the call stack because the functions don't accept a context.

dshulyak:
i understand that part. adding Context method will remove initialization code from every test

```
ctx, cancel := context.WithTimeout(context.Background(), *testcontext.TestTimeout)
t.Cleanup(cancel)
```

and won't require exporting TestTimeout

poszu:
> i understand that part. adding Context method will remove initialization code from every test
> 
> ```
> ctx, cancel := context.WithTimeout(context.Background(), *testcontext.TestTimeout)
> t.Cleanup(cancel)
> ```
> 
> and won't require exporting TestTimeout

I don't think it's a good idea, because then there will be two `context.Context` available in each method that takes both the standard `ctx context.Context` and `tctx testcontext.Context`: `ctx` and `tctx.Context()`. It is already confusing to have to `Context`s. It will be even more confusing if the `tctx` had its own `context.Context` inside.

btw, what do you think about renaming the` testcontext.Context` struct to avoid confusion caused by passing two `*.Context` arguments?

fasmat:
> btw, what do you think about renaming the` testcontext.Context` struct to avoid confusion caused by passing two `*.Context` arguments?

I would support that. Something like `testSession` or `testParams` instead of `testContext`.

poszu:
@dshulyak I'm stuck. What is it going to be?

dshulyak (approved):

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors merge

poszu:
bors merge

PR #3813, opened 2022-11-30
[Merged by Bors] - node: graceful shutdown on SIGTERM

### LINKED ISSUE #2003: Handle SIGTERM gracefully
I noticed that sending a SIGTERM (using the `kill` command) causes the node to immediately exit, without even printing any message to the log. Can we trap this and handle it the same way we handle a SIGINT (ctrl-c)?

countvonzero:
## Motivation

Closes #2003

also see discussion in https://github.com/spacemeshos/go-spacemesh/pull/2440

## Changes

handle sigterm

## Testing
verified on CI grafana that shutdown routines were executed

Screenshot from 2022-11-30 10-37-17

### DIFF
--- a/api/grpcserver/http_server.go
+++ b/api/grpcserver/http_server.go
@@ -9,21 +9,21 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	pb "github.com/spacemeshos/api/release/go/spacemesh/v1"
 
-	"github.com/spacemeshos/go-spacemesh/cmd"
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
 // JSONHTTPServer is a JSON http server providing the Spacemesh API.
 // It is implemented using a grpc-gateway. See https://github.com/grpc-ecosystem/grpc-gateway .
 type JSONHTTPServer struct {
+	ctx    context.Context
 	mu     sync.RWMutex
 	port   int
 	server *http.Server
 }
 
 // NewJSONHTTPServer creates a new json http server.
-func NewJSONHTTPServer(port int) *JSONHTTPServer {
-	return &JSONHTTPServer{port: port}
+func NewJSONHTTPServer(ctx context.Context, port int) *JSONHTTPServer {
+	return &JSONHTTPServer{ctx: ctx, port: port}
 }
 
 // Close stops the server.
@@ -31,9 +31,7 @@ func (s *JSONHTTPServer) Close() error {
 	log.Debug("stopping new json-http service...")
 	server := s.getServer()
 	if server != nil {
-		ctx := cmd.Ctx()
-
-		if err := server.Shutdown(ctx); err != nil {
+		if err := server.Shutdown(s.ctx); err != nil {
 			return fmt.Errorf("shutdown: %w", err)
 		}
 	}
--- a/api/grpcserver/http_server.go
+++ b/api/grpcserver/http_server.go
@@ -15,23 +15,22 @@ import (
 // JSONHTTPServer is a JSON http server providing the Spacemesh API.
 // It is implemented using a grpc-gateway. See https://github.com/grpc-ecosystem/grpc-gateway .
 type JSONHTTPServer struct {
-	ctx    context.Context
 	mu     sync.RWMutex
 	port   int
 	server *http.Server
 }
 
 // NewJSONHTTPServer creates a new json http server.
-func NewJSONHTTPServer(ctx context.Context, port int) *JSONHTTPServer {
-	return &JSONHTTPServer{ctx: ctx, port: port}
+func NewJSONHTTPServer(port int) *JSONHTTPServer {
+	return &JSONHTTPServer{port: port}
 }
 
-// Close stops the server.
-func (s *JSONHTTPServer) Close() error {
+// Shutdown stops the server.
+func (s *JSONHTTPServer) Shutdown(ctx context.Context) error {
 	log.Debug("stopping new json-http service...")
 	server := s.getServer()
 	if server != nil {
-		if err := server.Shutdown(s.ctx); err != nil {
+		if err := server.Shutdown(ctx); err != nil {
 			return fmt.Errorf("shutdown: %w", err)
 		}
 	}
--- a/cmd/integration/harness.go
+++ b/cmd/integration/harness.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"syscall"
 
 	"github.com/spacemeshos/go-spacemesh/log"
 )
@@ -85,7 +86,8 @@ func main() {
 	log.With().Info("integration: harness is listening on a blocking dummy channel")
 
 	interruptChannel := make(chan os.Signal, 1)
-	signal.Notify(interruptChannel, os.Interrupt)
+	// os.Interrupt for all systems, especially windows, syscall.SIGTERM is mainly for docker.
+	signal.Notify(interruptChannel, os.Interrupt, syscall.SIGTERM)
 
 	func() {
 		for {
--- a/cmd/integration/harness.go
+++ b/cmd/integration/harness.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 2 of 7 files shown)


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

fasmat:
There is a function in the signal package [signal.NotifiyContext](https://pkg.go.dev/os/signal#NotifyContext) it automatically cancels the context returned by it if a signal is captured.

This would allow to get rid of the `globalContext` and `globalCancel` variables and the associated functions / methods.

Imo a `context` should never be "gotten" from a function / method, but rather passed in as argument. The caller should decide which `context` a function is supposed to use, not the function getting the `context` it wants to use. This would require some additional refactoring though.

countvonzero:
> There is a function in the signal package [signal.NotifiyContext](https://pkg.go.dev/os/signal#NotifyContext) it automatically cancels the context returned by it if a signal is captured.
> 
> This would allow to get rid of the `globalContext` and `globalCancel` variables and the associated functions / methods.
> 
> Imo a `context` should never be "gotten" from a function / method, but rather passed in as argument. The caller should decide which `context` a function is supposed to use, not the function getting the `context` it wants to use. This would require some additional refactoring though.

thanks for the suggestion. done. @fasmat

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

cmd/node/node_test.go
```diff
@@ -566,6 +546,8 @@ func TestSpacemeshApp_NodeService(t *testing.T) {
 		poetHarness.HTTPPoetClient, clock, h, edSgn)
 	require.NoError(t, err)
 
+	appCtx, appCancel := context.WithTimeout(context.Background(), 10*time.Second)
```
poszu: I'd suggest not using an explicit short timeout as it tends to create flaky tests on slower runners. The test timeout is handled by `go test -timeout <value>` externally. IMO there's no need to duplicate it in test.
```suggestion
	appCtx, appCancel := context.WithCancel(context.Background())
```

api/grpcserver/http_server.go
```diff
@@ -9,31 +9,29 @@ import (
...
 
 // Close stops the server.
 func (s *JSONHTTPServer) Close() error {
 	log.Debug("stopping new json-http service...")
 	server := s.getServer()
 	if server != nil {
-		ctx := cmd.Ctx()
-
-		if err := server.Shutdown(ctx); err != nil {
+		if err := server.Shutdown(s.ctx); err != nil {
```
poszu: The forwarded context is usually the context of the current call stack. I think it should not be saved in `JSONHTTPServer` and reused here.

It'd require a slightly more refactor:
- accept `ctx context.Context` in `JSONHTTPServer::Close`, potentially renaming to `Shutdown` as it no longer implements `io::Closer` interface,
- accept `ctx context.Context` in `App::Cleanup` and `App::StopServices`,
- remove `ctx context.Context` from `JSONHTTPServer struct`.

💡 the `server.Shutdown` will not wait for the ongoing connections to be closed if the passed context expires.

What do you think?

poszu (commented):
Great change, simplifying the code A LOT! ❤️

api/grpcserver/http_server.go
countvonzero: i think it's great. done as suggested. thank you.

cmd/node/node.go
```diff
@@ -118,12 +118,24 @@ func GetCommand() *cobra.Command {
...
+				done := make(chan struct{}, 1)
+				go func() {
+					app.Cleanup(ctx2)
+					close(done)
+				}()
+				select {
+				case <-done:
+				case <-ctx2.Done():
+					log.With().Error("app failed to clean up in time")
+				}
```
poszu: NIT
```suggestion
				cleanupCtx, cancelCleanup := context.WithTimeout(context.Background(), 30*time.Second)
				defer cancelCleanup ()
				done := make(chan struct{}, 1)
				go func() {
					app.Cleanup(cleanupCtx)
					close(done)
				}()
				select {
				case <-done:
				case <-cleanupCtx.Done():
					log.With().Error("app failed to clean up in time")
				}
```

poszu (approved):

cmd/integration/harness.go
```diff
@@ -85,7 +86,8 @@ func main() {
 	log.With().Info("integration: harness is listening on a blocking dummy channel")
 
 	interruptChannel := make(chan os.Signal, 1)
-	signal.Notify(interruptChannel, os.Interrupt)
+	// os.Interrupt for all systems, especially windows, syscall.SIGTERM is mainly for docker.
+	signal.Notify(interruptChannel, os.Interrupt, syscall.SIGTERM)
```
fasmat: This could also be done with `signal.NotifyContext` 🙂

cmd/node/node.go
```diff
@@ -115,20 +116,32 @@ func GetCommand() *cobra.Command {
...
+					close(done)
+				}()
+				select {
+				case <-done:
+				case <-ctx2.Done():
+					log.With().Error("app failed to clean up in time")
 				}
-
-				return nil
+				return err
```
fasmat: I think this is more complicated than it needs to be. Can `app.Cleanup()` not react to the timeout itself?

```suggestion
				err = app.Start()
				
				ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
				defer cancel()
				app.Cleanup(ctx)
				return err
```

cmd/node/node_test.go
```diff
@@ -566,8 +546,10 @@ func TestSpacemeshApp_NodeService(t *testing.T) {
 		poetHarness.HTTPPoetClient, clock, h, edSgn)
 	require.NoError(t, err)
 
+	appCtx, appCancel := context.WithCancel(context.Background())
```
fasmat, line 549: In a test I usually just immediately call `defer appCancel()` after creating a `context`. To avoid it not being cleaned up if the test fails.

fasmat (commented):
How timed app shutdown is handled confuses me, besides that the changes LGTM!

cmd/node/node.go
countvonzero: app.Cleanup will hang in this case. as the shutdown routines do not check the context when shutting down. 
refactoring that is more time than i want to spend on this.

fasmat: But then `app.Cleanup()` will also hang in the unchanged version of the code and `"app failed to clean up in time"` will be logged after 30 seconds before this function returns?

If this is the case can we add a `// FIXME` here and create an issue for it? `app.Cleanup` should definitely return when done cleaning up or when the `context` is canceled.

countvonzero: my intention in the current code is to give up on cleaning up after 30 sec.
in this case, app will exit with the original app.Start() error. is that consistent with your reading?

i can create an issue for this yes.

countvonzero: tho, even if we changed our codepath in the shutdown routine, the library we use may not support that. so i suspect the code here will remain unchanged even after the timeout ctx is plumbed through in our code.

UPD:
i guess alternatively every component's shutdown will respect this timeout and kill whatever underlying library if they hang

fasmat: I'm just wondering why `app.Cleanup()` shouldn't be able to finish within 30 seconds. It unlocks via the `.LOCK` file and then tries to stop all our services. This would mean one of our services doesn't return when its `Stop()`/`Close()` method is called.

If we skip `app.Cleanup()` this could possibly lead to an inconsistent state.

I agree that this is out of scope for this issue but should be investigated.

fasmat: All `Close()` / `Stop()` calls are currently done sequentially, these could be done in parallel, e.g. with a `WaitGroup`; then every `Stop()` would have (about) the same timeout.

If one of the libraries we have doesn't have the option to pass in a `context` we should instead call the `Stop()` method of that component with something like this:

```golang
// within `app.stopServices(context.Context)`

var eg errgroup.Group
eg.Go(func() error {
	// component that respects context
	return component1.Stop(ctx)
})
eg.Go(func() error {
	errChan := make(chan error)
	go func() {
		err := component2.Stop()
		if err != nil {
			errChan <- fmt.Errorf("failed to stop component 2: %w", err)
		}
		close(errChan)
	}()
	
	select {
	case err := <-errCh:
		return err
	case <-ctx.Done():
		// possibly log here that component2.Stop() failed to return before context timed out
		return fmt.Errorf("failed to stop component 2: %w", ctx.Err())
	}
	return nil
})
// more components
return eg.Wait()
```

This way at least we can figure out which components cause the issues.

cmd/node/node_test.go
countvonzero: done

cmd/node/node.go
countvonzero: filed https://github.com/spacemeshos/go-spacemesh/issues/3830 and added FIXME

cmd/integration/harness.go
countvonzero: done

cmd/node/node.go
countvonzero: done

fasmat (approved):

countvonzero:
bors try

countvonzero:
bors merge

PR #3825, opened 2022-12-03
[Merged by Bors] - mesh: apply empty layer to cache and vm

### LINKED ISSUE #3805: conservative state: apply empty layer
## Description

currently conservative state/cache does not apply empty layers.
this cause a problem when getting txs out of mempool, as the code will skip txs in those empty layers

for example:
t0: tx_A100 and tx_B200 are packed in proposal P1 in layer 10.
t1: hare output an empty block for layer 10.
t2: when miner builds proposal P2 for layer 11, it skipped txA_100 and tx_B200 (as they are still marked as P1) and selected tx_B201 instead.
t3: hare output for layer 11 included P2
t4: tx_B201 got executed (as nonce gap is allowed)
t5: account B got reset in cache, tx_B200 is no longer applicable and dropped from mempool. account A still has tx_A100 in mempool, but still won't be selected for the next proposal.

countvonzero:
## Motivation

Closes #3805

## Changes

- mesh: applies a nil block to conservative state and vm for an empty layer.
- conservative state: reset unapplied txs in the empty layer


countvonzero:
bors try

mesh/mesh.go
```diff
@@ -518,12 +518,13 @@ func (msh *Mesh) applyState(ctx context.Context, logger log.Log, lid types.Layer
...
-		err := msh.conState.ApplyLayer(ctx, block)
-		if err != nil {
-			logger.With().Error("failed to apply transactions", log.Err(err))
-			return fmt.Errorf("apply layer: %w", err)
-		}
 	}
+
+	if err := msh.conState.ApplyLayer(ctx, lid, block); err != nil {
+		logger.With().Error("failed to apply transactions", log.Err(err))
+		return fmt.Errorf("apply layer: %w", err)
```
poszu: I know the code was just moved, but it's a great moment to improve :) 

IMO errors should be logged when they are handled because otherwise they might be logged multiple times (every time they are bubbled up). The attached context should contain all extra information required to debug.
```suggestion
		return fmt.Errorf("failed to apply transactions for layer %d: %w", lid.Value, err)
```

txs/cache_test.go
```diff
@@ -1021,41 +1069,37 @@ func TestCache_LinkTXsWithProposal(t *testing.T) {
...
-		if len(mtxs) > 1 {
-			checkTX(t, tc.cache, mtxs[1])
+		checkTX(t, tc.cache, mtxs[0].ID, lid0, types.EmptyBlockID)
+		cnt := len(mtxs)
+		if cnt > 1 {
+			checkTX(t, tc.cache, mtxs[1].ID, lid1, types.EmptyBlockID)
+		} else if cnt > 2 {
+			for _, mtx := range mtxs[2:] {
+				checkTX(t, tc.cache, mtx.ID, types.LayerID{}, types.EmptyBlockID)
+			}
```
poszu: The `else if cnt > 2` branch will never be taken.

txs/cache_test.go
```diff
@@ -1097,42 +1138,35 @@ func TestCache_LinkTXsWithBlock(t *testing.T) {
...
 	}
 	require.NoError(t, tc.LinkTXsWithBlock(tc.db, lid1, bid1, tids1))
 	for _, mtxs := range mtxsByAccount {
-		if len(mtxs) > 1 {
-			checkTX(t, tc.cache, mtxs[1])
+		checkTX(t, tc.cache, mtxs[0].ID, lid0, bid0)
+		cnt := len(mtxs)
+		if cnt > 1 {
+			checkTX(t, tc.cache, mtxs[1].ID, lid1, bid1)
+		} else if cnt > 2 {
```
poszu: ditto

txs/conservative_state.go
```diff
@@ -184,6 +188,17 @@ func (cs *ConservativeState) ApplyLayer(ctx context.Context, block *types.Block)
...
 
+func (cs *ConservativeState) applyEmptyLayer(ctx context.Context, lid types.LayerID) error {
+	_, _, err := cs.vmState.Apply(vm.ApplyContext{Layer: lid}, nil, nil)
+	if err != nil {
+		return fmt.Errorf("apply empty layer: %w", err)
+	}
+	if err = cs.cache.ApplyLayer(ctx, cs.db, lid, types.EmptyBlockID, nil, nil); err != nil {
+		return err
+	}
+	return nil
```
poszu:
```suggestion
	return cs.cache.ApplyLayer(ctx, cs.db, lid, types.EmptyBlockID, nil, nil)
```

txs/conservative_state_test.go
```diff
@@ -670,6 +670,13 @@ func TestApplyLayer(t *testing.T) {
 	}
 }
 
+func TestApplyEmptyLayer(t *testing.T) {
+	tcs := createConservativeState(t)
+	lid := types.NewLayerID(1)
+	tcs.mvm.EXPECT().Apply(vm.ApplyContext{Layer: lid, Block: types.EmptyBlockID}, nil, nil).Return(nil, nil, nil)
+	require.NoError(t, tcs.ApplyLayer(context.TODO(), lid, nil))
+}
```
poszu: Shouldn't the test verify if the layer was applied to the cache too? If the following lines run:
```go
if err = cs.cache.ApplyLayer(ctx, cs.db, lid, types.EmptyBlockID, nil, nil); err != nil {
	return err
}
```

poszu (changes requested):

mesh/mesh.go
dshulyak: > IMO errors should be logged when they are handled because otherwise they might be logged multiple times (every time they are bubbled up). The attached context should contain all extra information required to debug.

+1. they are logged multiple times and code looks messy when logs are written in this style

additional information can be included by inmplemeting structured errors that support custom MarshalLogObject method

txs/cache_test.go
countvonzero: why? each account has N txs in the pool, N is a random number bwtn 1 and 10.

mesh/mesh.go
countvonzero: done for mesh.go

txs/conservative_state_test.go
countvonzero: added test that this `cache.ApplyLayer()` is called by releasing the txs

txs/cache_test.go
dshulyak: first branch is always true if second branch is true, so it will always go through the first

countvonzero: ahhh... my bad. thank you! fixed

txs/cache_test.go
countvonzero: fixed. thank you

poszu (approved):

countvonzero:
bors try

countvonzero:
bors merge

PR #3832, opened 2022-12-06
[Merged by Bors] - objects stored on mesh should be decodable into different types based on layer

### LINKED ISSUE #3211: types: ensure that non-ephemeral types are upgradable
there are two kinds of types in the go-spacemesh:

1. ephemeral types

they are not stored on mesh, and need to remain decodable only within the context of some protocol. they will be versioned together with a sub-protocol itself. 

for example hare is identified in the gossipsub with `hr1`, next version of the hare will have any other unique identifier, for example `hr2`. `hr2` may use any types in the protocol, not necessarily `hr1`.

other ephemeral types and protocols:
- beacon
- proposals
- block certificate

2. types that are stored on mesh

such types should remain decodable by all future versions of the go-spacemesh. to make upgrade path future-proof we need to have some version as a first field, so that the rest of the payload can be decoded based on the value of the first field. 

- block - layer id is a first field.
- ballot - layer id should be made a first field.
- activation - pub layer id should be made as a first field (within nipost challenge)

layer id will be used for protocol forks, when the fork is made go-spacemesh will be released with update that will decode specified objects (depending on the nature of the update) starting from layer X differently.

as an example consider ballot type.

currently votes are decoded into:

```go
type Votes struct {
	// Base ballot.
	Base BallotID
	// Support and Against blocks that base ballot votes differently.
	Support, Against []BlockID
	// Abstain on layers until they are terminated.
	Abstain []LayerID
}
``` 

we could change this structure and add one more array with empty layers id (if it will make sense) 
```go
type Votes1100 struct {
	Base BallotID
	Support, Against []BlockID
	Abstain []LayerID 
        Empty []LayerID 
}
```

Starting from layer 1100 all ballots will be decoded using Votes1100.

3. special case:

- poet protocol
the type itself doesn't need a version. as it can be decode either based on atx version, when we verify that challenge is included into the protocol. or based on protocol version, when object is accepted into the node.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3211

### DIFF
--- a/common/types/activation_test.go
+++ b/common/types/activation_test.go
@@ -2,6 +2,10 @@ package types
 
 import (
 	"testing"
+
+	fuzz "github.com/google/gofuzz"
+	"github.com/spacemeshos/go-spacemesh/codec"
+	"github.com/stretchr/testify/require"
 )
 
 var result uint64
@@ -35,4 +39,23 @@ func BenchmarkMul(b *testing.B) {
 func unsafeMul(x, y uint64) uint64 {
 	result = x // avoid compiler optimizations
 	return x * y
+}
+
+func layerTester(tb testing.TB, encodable codec.Encodable) LayerID {
+	tb.Helper()
+	f := fuzz.NewWithSeed(1001)
+	f.Fuzz(encodable)
+	buf, err := codec.Encode(encodable)
+	require.NoError(tb, err)
+	var lid LayerID
+	require.NoError(tb, codec.Decode(buf, &lid))
+	return lid
+}
+
+func TestActivationEncoding(t *testing.T) {
+	t.Run("layer is first", func(t *testing.T) {
+		atx := ActivationTx{}
+		lid := layerTester(t, &atx)
+		require.Equal(t, atx.PubLayerID, lid)
+	})
 }
--- a/common/types/activation_test.go
+++ b/common/types/activation_test.go
@@ -4,8 +4,9 @@ import (
 	"testing"
 
 	fuzz "github.com/google/gofuzz"
-	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/stretchr/testify/require"
+
+	"github.com/spacemeshos/go-spacemesh/codec"
 )
 
 var result uint64
--- a/common/types/activation.go
+++ b/common/types/activation.go
@@ -98,11 +98,11 @@ var EmptyATXID = &ATXID{}
 // the intended publication layer ID, the PoET's start and end ticks, the positioning ATX's ID and for
 // the first ATX in the sequence also the commitment Merkle root.
 type NIPostChallenge struct {
+	PubLayerID LayerID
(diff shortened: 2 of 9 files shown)


countvonzero (approved):
can you also add a test for proposal as well?
i know it's ephemeral but it's also important when we are forking live.

dshulyak:
> i know it's ephemeral but it's also important when we are forking live.

i will add, but i think it won't be needed for proposal. for ephemeral objects updating the version of the protocol should be enough

what does it mean - forking live?

dshulyak:
actually after second thought - i don't think that we should enforce something if there is no use for it. so please explain why do you expect it to be needed

countvonzero:
> > i know it's ephemeral but it's also important when we are forking live.
> 
> i will add, but i think it won't be needed for proposal. for ephemeral objects updating the version of the protocol should be enough
> 
> what does it mean - forking live?

there are two protocols that deal with proposals, gossip and fetch. 

i was thinking if a node upgraded its sub protocol version in the middle of a hare protocol, and ended up having to fetch a proposal at the end of hare, it will need to accept the proposal from the old version. in that case, it seems easier to handle proposals the same way as ballots

dshulyak:
i don't think that it would make sense to do it. bumping protocol version (for gossip it is super simple, and i guess can be done for fetcher as well) would be easier. btw i also think that it wasn't necessary to add proposals to the fetcher, it would work without it the same way

dshulyak:
bors merge

common/types/activation_test.go
```diff
@@ -36,3 +41,22 @@ func unsafeMul(x, y uint64) uint64 {
...
+	var lid LayerID
+	require.NoError(tb, codec.Decode(buf, &lid))
+	return lid
+}
+
+func TestActivationEncoding(t *testing.T) {
+	t.Run("layer is first", func(t *testing.T) {
+		atx := ActivationTx{}
+		lid := layerTester(t, &atx)
+		require.Equal(t, atx.PubLayerID, lid)
```
poszu, line 60: IMO, the `atx.PubLayerID` should be set to be certain:
```suggestion
		atx := ActivationTx{PubLayerID: LayerID{Value: 123}}
		lid := layerTester(t, &atx)
		require.Equal(t, atx.PubLayerID, lid)
```

Same in other tests

countvonzero:
> i don't think that it would make sense to do it. bumping protocol version (for gossip it is super simple, and i guess can be done for fetcher as well) would be easier. btw i also think that it wasn't necessary to add proposals to the fetcher, it would work without it the same way

so your opinion is, if hare terminates with proposals not available locally, it should simply abort and wait for syncing the block?

dshulyak:
> so your opinion is, if hare terminates with proposals not available locally, it should simply abort and wait for syncing the block?

yes, i don't expect that to fail if hare works. we already expect that messages will be propagated on time for hare to terminate. and it is just simpler to have one recovery (e.g downloading block). but thats not very important i guess

PR #3833, opened 2022-12-06
[Merged by Bors] - Move activation/mocks into activation package

fasmat:
## Motivation
The packages `activation/mocks` and `activation` are inherently dependent on each other, since the `activation` package uses the mocks and the mocks use types from the `activation` package. This leads to the necessity of moving types that would belong in the `activation` package to `activation/types`.

This practice creates a lot of small packages that all have the same generic name (`types`, `mocks`, `config`, `util` etc.) that makes code that uses multiple of them confusing. `config.Config` says considerably less about what it is compared to `activation.Config` or `hare.Config`. Another example is `mocks.NewBuilderMock` vs. `activation.NewBuilderMock`.

Some handwritten mocks exist purely for the reason that generating them in a dedicated `mocks` package would introduce a cyclic dependency that isn't allowed in Go.

This PR only changes the `activation` package in this regard but IMO we should slowly move away from packages that are split based their content (`types`, `mocks`, `config`, `util`) towards packages that are split on their intended use (`activation`, `hare`, `crypto`, `grpc`, etc.).

This also allows us to reduce the surface of a package by making types and functions private that currently are not, just because of the imo unnecessary package split.

## Changes
Merge package `activation/mocks` into package `activation`.

## Test Plan
no functionality was changed, just code re-arranged.

## TODO

- [X] Explain motivation or link existing issue(s)
- [X] Test changes and document test plan
- [X] Update documentation as needed

## DevOps Notes


countvonzero:
> IMO we should slowly move away from packages that are split based their content (types, mocks, config, util) towards packages that are split on their intended use (activation, hare, crypto, grpc, etc.).

not sure i agree. i think our top level pkgs are already split on their intended use. `config` here means top level app config. inside the config pkg it does refer to tortoise.Config, hare.Config etc. i assume you are referring to `types` and `util` inside common pkg? that just means data that's shared across packages.

the code organization inside each package is indeed a bit messy now. i agree that we dont need sub packages in most cases, def a good idea to remove mocks packages.

activation/activation.go
```diff
@@ -38,26 +38,26 @@ func DefaultPoetConfig() PoetConfig {
 
 const defaultPoetRetryInterval = 5 * time.Second
 
-type nipostBuilder interface {
+type NipostBuilder interface {
```
countvonzero: is it possible to move these interface inside interface.go?

btw, why do they need to be public interface?

fasmat:
> not sure i agree. i think our top level pkgs are already split on their intended use. config here means top level app config. inside the config pkg it does refer to tortoise.Config, hare.Config etc. i assume you are referring to types and util inside common pkg? that just means data that's shared across packages.

Take a look at `config/config.go` its inside a `config` package and it imports another 4 packages called `config`. All of them contain a function called `DefaultConfig()` so `config.DefaultConfig()` instead of `eligibility.DefaultConfig()`, `hare.DefaultConfig()`. Some packages there do not have this "config" split: `tortoise`, `beacon` and `vm` don't have a sub-package called `config`. Why do the others?

This is not intended as criticism to anyone. I just think this is a bad practice that we should move away from.

activation/activation.go
fasmat: Good point. These interfaces should be private.

Regarding moving them to `interface.go`: I'd prefer getting rid of `interface.go` and move the interfaces to where they are used. The only code that depends on that interface is inside this file. Yes, the actual implementation of this interface is in `nipost.go` (or for the mock in `activation_mocks.go`), but these don't depend on the interface, they just implement it.

countvonzero:
> > not sure i agree. i think our top level pkgs are already split on their intended use. config here means top level app config. inside the config pkg it does refer to tortoise.Config, hare.Config etc. i assume you are referring to types and util inside common pkg? that just means data that's shared across packages.
> 
> Take a look at `config/config.go` its inside a `config` package and it imports another 4 packages called `config`. All of them contain a function called `DefaultConfig()` so `config.DefaultConfig()` instead of `eligibility.DefaultConfig()`, `hare.DefaultConfig()`. Some packages there do not have this "config" split: `tortoise`, `beacon` and `vm` don't have a sub-package called `config`. Why do the others?

yes. i agree that there shouldn't be config sub pkg within each package. i don't like sub packages in general.
 
> This is not intended as criticism to anyone. I just think this is a bad practice that we should move away from.

lol. it would be a criticism for me anyway hahah.... like i said, i agree we should do without sub packages within packages.

activation/activation.go
countvonzero: for me, within each package, the interface is already residing in the place they are used, either in interface.go or in each xxxxx.go file. the two advantage i see:
- one go:generate directive to generate the mock
- sometimes multiple go files use the same interface

countvonzero:
> it would be a criticism for me anyway hahah

i mean it would NOT be :P

activation/activation.go
fasmat: I moved (most) interfaces to `interface.go` and made them private.

genvm/core/gomock_reflect_1275411437/prog.go
```diff
@@ -0,0 +1,66 @@
+
+package main
```
countvonzero: what's this file about?

countvonzero (approved):

genvm/core/gomock_reflect_1275411437/prog.go
fasmat: just noticed, seems like these are leftovers from `gomock`? I'll delete them.

fasmat:
bors merge

fasmat:
bors merge

PR #3835, opened 2022-12-07
block: execute transactions in situ when optimistic filtering is ON 

### LINKED ISSUE #3297: block construction: execute txs in situ in the optimistic case

countvonzero:
## Motivation

Closes #3297

## Changes

- generate block while selecting txs from proposals
- pass an extra boolean param to ProcessLayerPerHareoutput to indicate whether the generated block is executed


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-stage2](https://github.com/spacemeshos/go-spacemesh/actions/runs/3643103533/jobs/6151883742)

succeeded on retry

https://github.com/spacemeshos/go-spacemesh/actions/runs/3643103533

dshulyak:
i don't understand this code structure, and the role of conservative cache in it. can i clarify what happens here before review? this is what i thought about this problem
- select all txs from proposals in hare, and sort them - without doing anything in conservative cache
- execute transactions (if previous layer is executed) - not related to conservative cache
- check consensus hash and if it matches between proposals - don't add ineffective transactions to the block. not related to conservative cache
- at this point the block is saved, with the references to the transactions
- conservative cache is notified that certain txs are applied, so that it can refresh its state

is it different from the implementation? if this is the case why conservative cache wraps vm and has a method for GenerateBlock?

mesh/mesh.go
```diff
@@ -341,6 +339,20 @@ func (msh *Mesh) ProcessLayer(ctx context.Context, layerID types.LayerID) error
 
 	newVerified, updated := msh.trtl.Updates()
 	logger = logger.WithFields(log.Stringer("verified", newVerified))
+
+	if len(executed) == 1 {
```
dshulyak, line 343: i think it needs to be heavily refactored

countvonzero:
> i don't understand this code structure, and the role of conservative cache in it. can i clarify what happens here before review? this is what i thought about this problem
> 
> * select all txs from proposals in hare, and sort them - without doing anything in conservative cache
> * execute transactions if previous layer is executed - not related to conservative cache
> * check consensus hash, if it matches between proposal - don't add ineffective transactions to the block
> * at this point the block is saved, with the references to the transactions
> * conservative cache is notified that certain txs are applied, so that it can update its state
> 
> is it different from the implementation? if this is the case why conservative cache wraps vm and has a method for GenerateBlock?

the steps you described are correct and implemented.

our current code flow for block execution are 
- non-optimistic:
hare->block generator->mesh->conservative cache->vm 
- optimistic
hare->block generator->vm->conservative cache->mesh

my goal is for the block execution to use the same code for executing txs in both optimistic and non-optimistic cases (execute in vm, update cache, update DB).

executing txs in place means block generator needs access to vm after gathering/sorting the proposal txs, and access to conservative cache to update the cache. but in the non-optimistic case, it seems odd to do it there. hence the choice. 

what do you prefer? in blocks package (adding reference to vm), or a new package all together, or as is?

i don't see it very strange to have the code in conservative cache (tunnel vision most likely). maybe renaming could help too?

dshulyak:
i don't know what i prefer. i need to think longer about it. not sure if it is helpful, but conservative cache is clearly overloaded and some code here doesn't make sense (like that if condition with variadic argument)

conservative cache is suppose to validate transactions and select subset of them for proposal, it was doing too much even before this pr

countvonzero:
you are def right about conservative cache doing too much. i will try to move the execution code out to the mesh pkg and separate mempool and execution. conservative cache just gets notified once txs are executed.

countvonzero:
alternative implementation merged in https://github.com/spacemeshos/go-spacemesh/pull/3841

PR #3841, opened 2022-12-09
[Merged by Bors] - block: execute transactions in situ

### LINKED ISSUE #3297: block construction: execute txs in situ in the optimistic case

countvonzero:
## Motivation

Closes #3297 

## Changes

- block generator process  hare output in layer order
- move txs execution code out of conservative state/cache to mesh pkg (`executor.go`)
- move utility functions that organize transactions in a block from txs pkg to blocks pkg
- flatten txs pkg to avoid sub pkg and circular dependency
- block generator execute the transactions in situ when generating if proposals agree on the state from previous layer

with this PR, execution path for transaction execution are
- optimistic filtering: hare->block generator->executor(vm+conservative cache)->mesh
- non-optimistic: hare->block generator->mesh->executor(vm+conservative cache)


dshulyak:
i think that block should not be created until optimistic filtering is applied, if protocol works this way. in current version block object is still created and passed around, there are some conditions that block.ID() is empty, and then there is a condition to check if block was executed at some point in the past. it is very error prone, this block object can be accidentally saved, passed to the tortoise, registered for certification. 

also premature creation of block object hides a significant design problem. transaction result has a block id field which is set inside the vm, so if context.ApplyContext has a wrong block id - transaction result will have wrong block id as well. i don't know how to fix it though. patching block id outside vm is an obvious solution, but it is just a mess. also results are meant to be saved by vm, and in this case we can't do it. 

optimistic filtering thing means that block id can't be a part of vm context, which is a separate question, if thats the case it should be removed from it

if we look at what needs to be done it is also very different from the normal flow. what are the pro's of trying to keep merge them together?

```go
ExecuteOptimistic(proposals []types.Proposal) {
    txs := getTxs(proposals)
    mersenne.Seed(proposals).Sort(txs)
    rewards := getRewards(proposals)

    results, ineffective := vm.Apply(txs, rewards)
    if optimistic(proposals) {
        txs = results
    }
    // block is created only at this point, so there is no room for misusing invalid block
    block := types.Block{Txs: txs}
    
    // the rest of the stuff
    saveBlock(block)
    updateResults(results) // if there is no better option
    saveResults(results)

    tortoise.OnHareOutput(block)
    cerifier.Certify(block)
    txpool.UpdateCache(...)
}

ExecuteValidBlock(block types.Block) {
     txs := getTxs(block)
     rewards := getRewards(rewards)
     results, ineffective := vm.Apply(txs, rewards)
     
     saveResults(results)

     txpool.UpdateCache(...)
}
```

genvm/vm.go
```diff
@@ -375,7 +374,6 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.Transa
 
 		rst := types.TransactionWithResult{}
```
countvonzero, line 375: @dshulyak should i generate types.TransactionWithResult outside of vm and change `Apply()` to return `[]types.Transaction` for the executed transactions?

countvonzero (commented):
@dshulyak please review.

i have mostly followed your suggestion on ExecuteOptimistic and ExecuteValidBlock. one question i have, also commented in the code, is maybe we can simply create TransactionWithResult outside VM.

genvm/vm.go
dshulyak: doesn't seem possible if you check what tx result contains

dshulyak (approved):
tx result needs to be fixed somehow

countvonzero:
> tx result needs to be fixed somehow

filed https://github.com/spacemeshos/go-spacemesh/issues/3894

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #3850, opened 2022-12-12
[Merged by Bors] - Inverse dependency of common/types and signing packages

fasmat:
## Motivation
Pre-requesite for https://github.com/spacemeshos/pm/issues/201

## Changes
This PR inverses the dependency of `common/types` on `signing`, so that `signing` depends on `common/types`.

Other smaller changes:
- Use `NodeID` instead of `[]byte`/`signing.PublicKey` where appropriate
- `VRFPostIndex` has been added to `types.ActivationTxHeader`: this is for the next PR that follows this one
- There is no global `signing.DefaultVerifier` anymore. It is injected into the services that need it instead
- Use `go-scale` for generating code instead of handwritten scale encoding
- Tests should use `context.Background()` instead of `context.TODO()`
- The controller returned by `gomock.New(t)` doesn't need to be finalized at the end of the test with `ctrl.Finish()` any more. `gomock.New(t)` does this by itself with `t.Cleanup()`. See also: https://pkg.go.dev/github.com/golang/mock/gomock#Controller.Finish

## Test Plan
Existing tests pass.

## TODO

- [x Update documentation as needed

## DevOps Notes


dshulyak:
i would prefer to use nonce only in beacon. for other types it doesn't have to be included explicitly, since other are using beacon one way or the other. not adding it to other types allows to replace beacon protocol in future and drop nonce completely

cmd/node/node.go
```diff
@@ -945,20 +945,20 @@ func (app *App) LoadOrCreateEdSigner() (*signing.EdSigner, error) {
...
+		err = os.WriteFile(filename, edSgn.Bytes(), 0o600)
 		if err != nil {
 			return nil, fmt.Errorf("failed to write identity file: %w", err)
 		}
 
 		log.With().Info("created new identity", edSgn.PublicKey())
 		return edSgn, nil
 	}
-	edSgn, err := signing.NewEdSignerFromBuffer(data, signing.WithSignerPrefix(app.Config.Genesis.GenesisID().Bytes()))
+	edSgn, err := signing.NewEdSignerFromBuffer(data, signing.WithPrefix(app.Config.Genesis.GenesisID().Bytes()))
```
dshulyak: WithSignerPrefix tells which option is used

signing/vrfsigner.go
```diff
@@ -1,21 +1,25 @@
...
 	pub        *PublicKey
+
+	nonce uint64
 }
 
 // Sign signs a message for VRF purposes.
 func (s VRFSigner) Sign(msg []byte) []byte {
-	return ecvrf.Prove(ed25519.PrivateKey(s.privateKey), msg)
+	nonceMsg := make([]byte, 8)
+	binary.LittleEndian.PutUint64(nonceMsg, s.nonce)
```
dshulyak: i think everything should be encoded with scale, but i would rather not even do this change like this

fasmat: I cannot pass in anything besides `[]bytes` because the `common/types` package depends on `signing` so using any concrete type instead of `[]bytes` will introduce a cyclic dependency. The proper fix imo for this would be to not have `common/types` depend on `signing` but have it the other way round, but this requires a lot of changes.

I could pass in `scale.Encodable`, encode it with scale and then sign it, but we don't use scale in beacon weakcoin. Proposals are e.g. encoded like this before signing them: https://github.com/spacemeshos/go-spacemesh/blob/vrf-nonce-invokations/beacon/weakcoin/weak_coin.go#L341-L358

I feel like changing all of that to `scale` encoding is out of scope for this PR. What would do you say?

cmd/node/node.go
fasmat: I find the name `WithSignerPrefix` confusing. The message is prefixed not the signer? Also the name is repetitive: `signing.WithSignerPrefix` contains the hint that it is used to sign something twice.

fasmat:
> i would prefer to use nonce only in beacon. for other types it doesn't have to be included explicitly, since other are using beacon one way or the other. not adding it to other types allows to replace beacon protocol in future and drop nonce completely

So should I create a `BeaconVRFSigner` that is only used in the `beacon` package and leave all other uses of VRFSigner untouched, e.g. [here](https://github.com/spacemeshos/go-spacemesh/blob/55bdd15427f8a63a21b948f6540b7e700798be07/hare/eligibility/oracle.go#L259-L270) in `hare` (this one uses scale for encoding)?

dshulyak:
> So should I create a BeaconVRFSigner that is only used in the beacon package and leave all other uses of VRFSigner untouched, e.g. [here](https://github.com/spacemeshos/go-spacemesh/blob/55bdd15427f8a63a21b948f6540b7e700798be07/hare/eligibility/oracle.go#L259-L270) in hare (this one uses scale for encoding)?

i think the simpler change would be to add vrf explicitly where it is used (e.g beacon proposal vrf and beacon weakcoin vrf)

signing/vrfsigner.go
dshulyak: i think this is irrelevant if more isolated approach will be used, lets get back to this if there will be disagreement about that

fasmat:
> i think the simpler change would be to add vrf explicitly where it is used (e.g beacon proposal vrf and beacon weakcoin vrf)

I'm not sure I understand. VRF is used in several places. My questions was if I should create a second VRFSigner specifically for `beacon` that uses the POPS-VRF nonce and leave the exist `VRFSigner` and its uses in other packages as they are?

cmd/node/node.go
dshulyak: understood, the idea though is that there is more than one type in that module, so option with a similar purpose can exist for them as well

dshulyak:
> I'm not sure I understand. VRF is used in several places. My questions was if I should create a second VRFSigner specifically for beacon that uses the POPS-VRF nonce and leave the exist VRFSigner and its uses in other packages as they are?

i meant to use nonce explicitly for two vrf inputs (beacn proposals and beacon weakcoin). but anyway thats irrelevant because Tal wants to use it everywhere

signing/vrfsigner.go
```diff
@@ -3,43 +3,50 @@ package signing
...
 // VRFSigner is a signer for VRF purposes.
 type VRFSigner struct {
+	// cdb *datastore.CachedDB
+
 	privateKey []byte
-	pub        *PublicKey
+	nodeID     types.NodeID
 }
 
+// func (s VRFSigner) getVRFNonce(nodeId types.NodeID) (*types.VRFPostIndex, error) {
```
countvonzero: rm

common/types/activation_test.go
```diff
@@ -42,21 +42,15 @@ func unsafeMul(x, y uint64) uint64 {
 	return x * y
 }
 
-func layerTester(tb testing.TB, encodable codec.Encodable) LayerID {
```
countvonzero, line 45: this is not sustainable. 
i know you don't like helper functions. but other teammates do.
and we get in the cycle of removing/creating helper functions endlessly.
i think we should keep this.

common/types/layer.go
```diff
@@ -144,6 +146,15 @@ func BytesToNodeID(buf []byte) (id NodeID) {
 	return id
 }
 
+// ExtractNodeIDFromSig extracts the NodeID from a signature.
+func ExtractNodeIDFromSig(msg, sig []byte) (NodeID, error) {
```
countvonzero: doesn't seem intuitive to live here. how about creating a new file nodeid.go and move everything NodeID there?

fasmat:
bors try

fasmat:
> i meant to use nonce explicitly for two vrf inputs (beacn proposals and beacon weakcoin). but anyway thats irrelevant because Tal wants to use it everywhere

to accomplish this I refactored the code to allow me to use `types.NodeID` to fetch the nonce required for the creation of signature and verification of it directly from within the signing package.

This PR now only contains the refactoring changes. The functional change will be in a follow up PR that I'm currently working on

cmd/node/node.go
```diff
@@ -568,7 +571,7 @@ func (app *App) initServices(ctx context.Context,
 
 	txHandler := txs.NewTxHandler(app.conState, app.addLogger(TxHandlerLogger, lg))
 
-	hOracle := eligibility.New(beaconProtocol, cdb, signing.VRFVerify, vrfSigner, app.Config.LayersPerEpoch, app.Config.HareEligibility, app.addLogger(HareOracleLogger, lg))
+	hOracle := eligibility.New(beaconProtocol, cdb, signing.VRFVerifier{}, vrfSigner, app.Config.LayersPerEpoch, app.Config.HareEligibility, app.addLogger(HareOracleLogger, lg))
```
countvonzero, line 575: not sure why we don't pass app.edVerify here?

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	github.com/oasisprotocol/curve25519-voi v0.0.0-20221003100820-41fad3beba17
 	github.com/prometheus/client_golang v1.14.0
 	github.com/prometheus/common v0.39.0
-	github.com/pyroscope-io/pyroscope v0.35.1
+	github.com/pyroscope-io/pyroscope v0.36.0
```
countvonzero: please remove go.mod / go.sum and make a separate PR for this. this change is big enough on its own. i'd like to separate causes of failures in different PRs.

hare/preroundtracker_test.go
```diff
@@ -23,7 +23,30 @@ const (
 )
 
 func getProposalID(i int) types.ProposalID {
-	return types.GenLayerProposal(types.NewLayerID(1), nil).ID()
+	return genLayerProposal(types.NewLayerID(1), nil).ID()
+}
+
+func genLayerProposal(layerID types.LayerID, txs []types.TransactionID) *types.Proposal {
```
countvonzero, line 29: i don't get why removing this from common/testutils? i can support moving the testutils to a different pkg maybe.

countvonzero (commented):
great change.

signing/vrfsigner.go
fasmat: 👍

common/types/activation_test.go
fasmat: The only place where this function would be used now is for the test right below it.

The other 3 places are in different packages now. I would now need to create a new package that contains this function, just so it can be re-used.

At the same time I'm actually of the opinion to get rid of those tests. All 4 tests do the same thing: check if go-scale encodes and decodes correctly, once with an ATX, once with a Proposal, once with a Ballot and once with a Block. Why should this fail with one of those types and not with the others?

common/types/layer.go
fasmat: I put it here because the `NodeID` type, its methods and `BytesToNodeID` are also defined here.

I can move all of those to a new file if you want?

common/types/activation_test.go
countvonzero: because those types can be changed separately later. isn't that the idea of unit-testing?

common/types/layer.go
countvonzero: yes please. those didn't make sense in the first place.

cmd/node/node.go
fasmat: `app.edVerify` is an `signing.EDVerifier` not a `signing.VRFVerifier`; `singing.EDVerifier` (contrary to its name) extracts public keys from signatures, while a `signing.VRFVerifier` verifies signatures for correctness.

I can change the name of `signing.EDVerifier` to make that clearer if you want.

go.mod
fasmat: 👍

cmd/node/node.go
countvonzero: yes. please. since the whole point of this PR is making code better, go for it!

hare/preroundtracker_test.go
fasmat: We already have a `common/utils` package that I cannot move this to as otherwise I'd introduce a cyclic dependency. Creating a new `common/testutils` package for this creates another point of friction that will cause issues of cyclic dependencies in the future.

This is btw. another good example of why I think this "do not repeat yourself" with helpers is well intentioned but usually more trouble than what it's worth: This function is called in 4 places.

- `hare/preroundtracker_test.go` function `getProposalID` which can just return a random ID instead.
- `hare/hare_test.go` function `randomProposal` which overwrites half the fields in the proposal and creates a new signature, at that point why call `genLayerProposal` in the first place?
- `hare/hare_rounds_test.go`; the only legitimate use of this helper
- `hare/flows_test.go`; which overwrites the epoch data of the generated proposal rendering its signature invalid

Especially in tests I feel like we are re-using too much code between tests leading to weird and hard to debug errors and race conditions. Imo if we would reduce code re-usability we would have more self contained and thereby stable tests.

go.mod
fasmat: 👍

common/types/activation_test.go
fasmat: I think I was able to create something that can be re-used using generics. See my latest commit.

common/types/layer.go
fasmat: code moved to `nodeid.go`

cmd/node/node.go
fasmat: Renamed to `signing.PubKeyExtractor`.

common/types/activation_tx_header.go
```diff
@@ -10,6 +10,9 @@ type ActivationTxHeader struct {
 	// current epoch; a unit represents a configurable amount of data for PoST
 	NumUnits uint32
 
+	// VRFNonce is the nonce found during PoST initialization
+	VRFNonce *VRFPostIndex
```
countvonzero: rm

datastore/store.go
```diff
@@ -192,6 +192,7 @@ func getHeader(vatx *types.VerifiedActivationTx) *types.ActivationTxHeader {
 		NIPostChallenge: vatx.NIPostChallenge,
 		Coinbase:        vatx.Coinbase,
 		NumUnits:        vatx.NumUnits,
+		VRFNonce:        vatx.VRFNonce,
```
countvonzero: rm

hare/eligibility/oracle.go
```diff
@@ -68,7 +68,7 @@ type Oracle struct {
 	beacons        system.BeaconGetter
 	cdb            *datastore.CachedDB
 	vrfSigner      *signing.VRFSigner
-	vrfVerifier    verifierFunc
+	vrfVerifier    signing.Verifier
```
countvonzero, line 68: remove def of verifierFunc on line 63

countvonzero (approved):

fasmat:
bors merge

fasmat:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

fasmat:
bors merge

andres-spacemesh:
bors try

andres-spacemesh:
bors try

andres-spacemesh:
bors try-

andres-spacemesh:
bors try

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

cmd/node/node.go
```diff
@@ -523,9 +523,7 @@ func (app *App) initServices(ctx context.Context,
 		return errors.New("invalid golden atx id")
 	}
 
-	app.keyExtractor = signing.NewPubKeyExtractor(
-		signing.WithVerifierPrefix(app.Config.Genesis.GenesisID().Bytes()),
-	)
+	app.keyExtractor = signing.NewPubKeyExtractor()
```
countvonzero: i don't understand this change. in the old code, the default verifier does have the genesis id for prefix

fasmat: I was just trying if this is a possible cause, it is not; the change will be reverted with the next commit. I have a few more spots that could be a cause for issues in systests

fasmat: Undid this change

signing/verifier.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (e PubKeyExtractor) ExtractNodeID(m, sig []byte) (types.NodeID, error) {
+	msg := make([]byte, len(m)+len(e.prefix))
+	copy(msg, e.prefix)
+	copy(msg[len(e.prefix):], m)
+	pub, err := ed25519.ExtractPublicKey(msg, sig)
+	if err != nil {
+		return *types.EmptyNodeID, err
+	}
+	return types.BytesToNodeID(pub), nil
```
countvonzero: looks like you can do 
```
pub, err := e.Extract()
return types.ByteToNodeID(pub.PublicKey)
```

countvonzero (approved):

fasmat:
bors merge

Issue #3763, opened 2022-11-18
use https://github.com/spacemeshos/fixed for floating numbers consistently


dshulyak:
[library](https://github.com/spacemeshos/fixed) is much more efficient (both memory and cpu-wise) compared to alternatives that are based on rational number or golang big.Int implementation 

use cases:
- ~distributing rewards based on space units~
- [x] threshold and weight computation in tortoise
- threshold computation in tortoise beacon

dshulyak:
I assigned myself to work on tortoise/rewards part

dshulyak:
I am using it in tortoise, but i think it should not be used for rewards, as it is sufficient to perform math on integers with single round down. Not sure if this library is relevant for beacon, issue can be closed if it is not

PR #3852, opened 2022-12-14
[Merged by Bors] - replace util weight with fixed float lib in tortoise

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/3763

- ballot weight is computed with rational numbers (using bit.Rat), so that space unit can be encoded without loss of precision
- tortoise get space unit in big.Rat and uses spacemeshos/fixed library for math and comparison with threshold. i think it is safe to consider this as an implementation detail and can be safely swapped to anything more efficient in future
- util.Weight is removed


tortoise/verifying.go
```diff
@@ -76,13 +75,13 @@ func (v *verifying) verify(logger log.Log, lid types.LayerID) bool {
...
 
-	margin := util.WeightFromUint64(0).
-		Add(v.totalGoodWeight).
+	margin := v.totalGoodWeight.
 		Sub(layer.verifying.goodUncounted)
 	uncounted := v.expectedWeight(v.Config, lid).
 		Sub(margin)
-	if uncounted.Sign() > 0 {
-		margin.Sub(uncounted)
+	// GreaterThan(zero) returns true even if value with negative sign
```
countvonzero, line 82: curious. is this a bug in fixed64 library?

tortoise/tortoise.go
```diff
@@ -694,12 +695,16 @@ func (t *turtle) onAtx(atx *types.ActivationTxHeader) {
 			log.Uint64("weight", atx.GetWeight()),
 		)
 		epoch.atxs[atx.ID] = atx.GetWeight()
-		epoch.weight += atx.GetWeight()
+		if atx.GetWeight() > math.MaxInt64 {
+			// atx weight is not expected to overflow int64
+			panic("fixme: atx size overflows int64 " + strconv.FormatUint(atx.GetWeight(), 10))
```
countvonzero: log.Fatal

tortoise/tortoise.go
```diff
@@ -694,12 +695,16 @@ func (t *turtle) onAtx(atx *types.ActivationTxHeader) {
...
+		}
+		epoch.weight = epoch.weight.Add(fixed.New64(int64(atx.GetWeight())))
```
countvonzero, line 701: it is also expected that total epoch weight will not exceed uint64. does `fixed.New64` satisfy that?

countvonzero (approved):

tortoise/tortoise.go
dshulyak: yes, it checks overflow on math operations

tortoise/verifying.go
dshulyak: i want to understand that as well)

dshulyak:
bors merge

PR #3854, opened 2022-12-14
[Merged by Bors] - add base gas that reflects verify/parse and storage cost

### LINKED ISSUE #3611: vm: changes in the gas estimation
more info: https://community.spacemesh.io/t/nonce-in-self-spawn-never-ending-battle/304/4

```
val storageCost = tx.binarySize * STORAGE_COST_FACTOR
val baseCost = FIXED_GAS(tx.template, "Verify")
val intrinsicGas = baseCost + storageCost
val computationCost = FIXED_GAS(tx.template, tx.method)
val totalGas = intrinsicGas + computationCost
```

## implementation

- add constant BaseGas for every template
- extend ParseOutput with BaseGas field
- header.MaxGas should be equal to BaseGas + len(tx) * STORAGE_COST + FixedGas
- change the following condition `ctx.PrincipalAccount.Balance < ctx.ParseOutput.FixedGas*ctx.ParseOutput.GasPrice`
to `ctx.PrincipalAccount.Balance < BaseGas + len(tx) * STORAGE_COST`

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3611


genvm/vm.go
```diff
@@ -313,11 +313,11 @@ func (v *VM) execute(lctx ApplyContext, ss *core.StagedCache, txs []types.Transa
 		}
 		ctx := req.ctx
 		args := req.args
-		if ctx.PrincipalAccount.Balance < ctx.ParseOutput.FixedGas*ctx.ParseOutput.GasPrice {
-			logger.With().Warning("ineffective transaction. fixed gas not covered",
+		if intrinsic := core.ComputeIntrinsicGasCost(ctx.ParseOutput.BaseGas, tx.GetRaw().Raw, v.cfg.StorageCostFactor); ctx.PrincipalAccount.Balance < intrinsic {
```
countvonzero, line 326: i am sure i asked this before.

why not check if `ctx.PrincipalAccount.Balance > ctx.Header.MaxGas` here since the first thing done in the execution flow is `ctx.Consume(ctx.Header.MaxGas)`?

dshulyak: i think if transaction was included into the block we don't want to store it for free so we will try to get as much from it as possible. but if it can't cover this intrinsic gas (parse, verify and storage cost) and then we don't want to even try to execute it

countvonzero: but this feels like we are coding for future when we have metered gas.
the code as is, doesn't make sense in the fixed gas model. consuming max gas is literally the first thing the code does.

countvonzero (approved):

genvm/vm.go
dshulyak: i don't understand what doesn't make sense.
code consumes max gas and if max gas is not available it will consume as much as it can, but more than intrinsic gas

dshulyak:
bors merge

genvm/vm.go
countvonzero: ah. now i remembered. it will clean up the balance if max gas is not available. thanks

dshulyak:
bors merge

PR #3862, opened 2022-12-15
[Merged by Bors] - simplify base ballot encoding

### LINKED ISSUE #3643: simplify base ballot selection
depends on #3637 

mutable ballots allows to simplify votes encoding. even if original encoding is not optimal in size, it can be re-encoded later and will not cause a mesh growth.

## implementation

- pick latest ballot that is consistent with local opinion (within 10 000 layers there must be such ballot), otherwise pick any ballot from the latest layer
- remove code that selects ballots based on priority

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3643

picks non-malicious ballot starting from the last layer and encode the difference in the opinion, previous "optimizations"
are not relevant because of ballot mutability, and because verifying tortoise is using consensus (history) hash to determine "good" ballots.


tortoise/tortoise.go
```diff
@@ -135,103 +135,49 @@ func (t *turtle) evict(ctx context.Context) {
...
+		} else {
+			choices = t.ballots[lid]
 		}
-		for _, block := range lvote.blocks {
-			vote, _, err := t.getFullVote(t.verified, current, block)
-			if err != nil {
-				return types.LayerID{}, err
+		for _, base := range choices {
+			if base.malicious {
+				// skim them as they are candidates for pruning
```
countvonzero: :s/skim/skip/

tortoise/tortoise.go
```diff
@@ -135,103 +135,49 @@ func (t *turtle) evict(ctx context.Context) {
...
+		if lid == types.GetEffectiveGenesis() {
+			choices = []*ballotInfo{{layer: types.GetEffectiveGenesis()}}
+		} else {
+			choices = t.ballots[lid]
 		}
-		for _, block := range lvote.blocks {
-			vote, _, err := t.getFullVote(t.verified, current, block)
-			if err != nil {
-				return types.LayerID{}, err
+		for _, base := range choices {
```
countvonzero, line 148: how about bad beacon? why wouldn't we want to prioritize ballots with the same beacon (to minimize data in votes)?

dshulyak: data in votes is ephemeral, it can be reorganized later by receiver if it wants to minimize amount of stored data. i would rather have simple code instead of meaningless optimizations

the only reason i kept malicious filter is that we may decide to keep only first two malicious ballots, so it may break encoding

countvonzero: > data in votes is ephemeral, it can be reorganized later by receiver if it wants to minimize amount of stored data. i would rather have simple code instead of meaningless optimizations

i agree with this approach in general. but in this case, the default node implementation doesn't support reorganizing votes yet, and this optimization seems rather cost-less. but probably not important enough.

countvonzero (approved):

dshulyak:
bors merge

fasmat:
bors merge

countvonzero:
bors merge

PR #3865, opened 2022-12-15
[Merged by Bors] - Query poet proofs instead of relying on broadcasting

### LINKED ISSUE #3746: Remove RPC API support for broadcasting PoET proofs.

### LINKED ISSUE #3814: NIPost builder should query poet for proof
ATX builder: follow-up on the PoET service to query the PoET proof once expected to be available. Then store it in local db to become available for other peers to fetch. This should replace code paths which subscribe to incoming PoET proofs.

follows: https://github.com/spacemeshos/poet/issues/169

poszu:
## Motivation
Part of https://github.com/spacemeshos/pm/issues/173

Closes #3746 
Closes #3814 

## Changes
- removed broadcasting method from `GatewayService`
- removed p2p listeners for broadcasted poet proofs
- changed `NIPostBuilder` to query poets for proofs after the rounds end

## Test Plan
- added a system test in which nodes use different poets to verify if poet proofs are properly propagated between nodes

## TODO
- [ ] Bump poet to a released version in go.mod after https://github.com/spacemeshos/poet/pull/187 is merged

## DevOps Notes
- [ ] ~This PR does not affect public APIs~ Proof broadcasting was removed
- [ ] ~This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)~ - It relies on a new Poet version
- [ ] ~This PR does not make changes to log messages (which monitoring infrastructure may rely on)~


poszu:
bors try

poszu:
bors try

activation/activation.go
```diff
@@ -594,7 +594,7 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 		return fmt.Errorf("broadcast: %w", err)
 	}
 
-	logger.Event().Info("atx published", log.Inline(atx), log.Int("size", size))
+	logger.Event().Info("atx published", log.Inline(atx), log.Int("size", size), atx.ID().Field())
```
countvonzero: the inlined log `atx.MarshalLogObject` contains id. maybe change the field name to atx_id?

activation/nipost.go
```diff
@@ -195,26 +192,21 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.PoetC
...
-		log.String("poet_id", hex.EncodeToString(poetServiceID)))
+	logger := nb.log.WithFields(log.String("poet_id", hex.EncodeToString(poetServiceID)))
+	logger.Debug("submitting challenge to poet proving service")
 
 	round, err := poet.Submit(ctx, challenge, signature)
 	if err != nil {
-		logger.With().Error("failed to submit challenge to poet proving service",
-			log.String("poet_id", hex.EncodeToString(poetServiceID)),
-			log.Err(err))
+		logger.With().Error("failed to submit challenge to poet proving service", log.Err(err))
```
countvonzero: log only where the error is handled.

activation/nipost.go
```diff
@@ -149,17 +150,13 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.PoetC
...
 	if nb.state.PoetProofRef == nil {
-		select {
-		case <-time.After(time.Until(poetProofDeadline)):
-		case <-ctx.Done():
-			return nil, 0, ctx.Err()
-		}
-		poetProofRef := nb.getBestProof(ctx, challengeHash)
+		getProofsCtx, cancel := context.WithDeadline(ctx, poetProofDeadline)
+		defer cancel()
+		poetProofRef := nb.getBestProof(getProofsCtx, challengeHash)
```
countvonzero: the name challengeHash is getting confusing for me.
sometimes it's Hash32, sometimes it's []byte (in the struct)
why not simply use []byte everywhere?

activation/nipost.go
```diff
@@ -247,46 +239,109 @@ func (nb *NIPostBuilder) submitPoetChallenges(ctx context.Context, challenge []b
...
-			nb.log.With().Debug("poet proof membership doesn't contain the challenge", log.Binary("challenge", challenge[:]))
+	}
+}
+
+func (nb *NIPostBuilder) getBestProof(ctx context.Context, challenge *types.Hash32) types.PoetProofRef {
+	proofs := make(chan *types.PoetProofMessage, len(nb.state.PoetRequests))
+
+	var eg errgroup.Group
+	for _, r := range nb.state.PoetRequests {
+		logger := nb.log.WithFields(log.String("poet-id", hex.EncodeToString(r.PoetServiceID)), log.String("round", r.PoetRound.ID))
```
countvonzero: consistent field name "poet_id"

activation/nipost.go
```diff
@@ -247,46 +239,109 @@ func (nb *NIPostBuilder) submitPoetChallenges(ctx context.Context, challenge []b
...
+				return ctx.Err()
+			case <-time.After(waitTime):
+			}
+			proof, err := nb.getProofWithRetry(ctx, client, round, time.Second)
+			if err != nil {
+				logger.With().Warning("Failed to get proof from Poet", log.Err(err))
+				return nil
+			}
+
+			if err := nb.poetDB.ValidateAndStore(ctx, proof); err != nil && !errors.Is(err, ErrObjectExists) {
```
countvonzero, line 310: under the current model where you have one goroutine to query proof from each service, do you still expect  ErrObjectExists to be a normal condition?

activation/nipost.go
```diff
@@ -247,46 +239,109 @@ func (nb *NIPostBuilder) submitPoetChallenges(ctx context.Context, challenge []b
...
+				return nil
+			}
 
-		if bestProof == nil || bestProof.leafCount < proof.LeafCount {
-			bestProof = &poetProof{
-				ref:       ref,
-				leafCount: proof.LeafCount,
+			// We are interested only in proofs that we are members of
+			if !membersContain(proof.Members, challenge) {
+				logger.With().Debug("poet proof membership doesn't contain the challenge", log.Binary("challenge", challenge[:]))
```
countvonzero: does logging the binary format or hash more informative?

activation/nipost.go
```diff
@@ -247,46 +239,109 @@ func (nb *NIPostBuilder) submitPoetChallenges(ctx context.Context, challenge []b
...
+				logger.With().Warning("Failed to validate and store proof", log.Err(err), log.Object("proof", proof))
+				return nil
+			}
 
-		if bestProof == nil || bestProof.leafCount < proof.LeafCount {
-			bestProof = &poetProof{
-				ref:       ref,
-				leafCount: proof.LeafCount,
+			// We are interested only in proofs that we are members of
+			if !membersContain(proof.Members, challenge) {
```
countvonzero, line 316: does this warrant a warning log? 
the challenge is expected to be in the members if we have successfully submit it to the service id. right?

activation/nipost.go
```diff
@@ -247,46 +239,109 @@ func (nb *NIPostBuilder) submitPoetChallenges(ctx context.Context, challenge []b
...
+				logger.With().Debug("poet proof membership doesn't contain the challenge", log.Binary("challenge", challenge[:]))
+				return nil
 			}
+
+			proofs <- proof
+			return nil
+		})
+	}
+	eg.Wait()
```
countvonzero: should check for err here. the app may be shutting down.

activation/nipost.go
```diff
@@ -247,46 +239,109 @@ func (nb *NIPostBuilder) submitPoetChallenges(ctx context.Context, challenge []b
...
 		}
-		// We are interested only in proofs that we are members of
-		membership, err := nb.poetDB.GetMembershipMap(ref)
-		if err != nil {
-			nb.log.With().Panic("failed to fetch membership for poet proof", log.Binary("challenge", challenge[:]))
+	}
+	return false
+}
+
+func (nb *NIPostBuilder) getProofWithRetry(ctx context.Context, client PoetProvingServiceClient, roundID string, retryInterval time.Duration) (*types.PoetProofMessage, error) {
```
countvonzero, line 261: seems the info log here is more debugging purpose.

activation/poet.go
```diff
@@ -108,19 +126,52 @@ func (c *HTTPPoetClient) Submit(ctx context.Context, challenge []byte, signature
 	if err := c.req(ctx, "POST", "/submit", &request, &resBody); err != nil {
 		return nil, err
 	}
-
-	return &types.PoetRound{ID: resBody.RoundId, ChallengeHash: resBody.Hash}, nil
+	roundEnd := time.Time{}
+	if resBody.RoundEnd != nil {
+		roundEnd = time.Now().Add(resBody.RoundEnd.AsDuration())
```
countvonzero, line 131: curious about the decision to communicate round end as duration and not absolute time?

activation/poet.go
```diff
@@ -108,19 +126,52 @@ func (c *HTTPPoetClient) Submit(ctx context.Context, challenge []byte, signature
...
+		PoetProof: types.PoetProof{
+			MerkleProof: shared.MerkleProof{
+				Root:         resBody.Proof.GetProof().GetRoot(),
+				ProvenLeaves: resBody.Proof.GetProof().GetProvenLeaves(),
+				ProofNodes:   resBody.Proof.GetProof().GetProofNodes(),
+			},
+			Members:   resBody.Proof.GetMembers(),
+			LeafCount: resBody.Proof.GetLeaves(),
+		},
+		PoetServiceID: resBody.Pubkey,
```
countvonzero, line 175: maybe also set c.poetServiceID if not set too.

common/types/activation.go
```diff
@@ -357,6 +387,18 @@ type PoetProofMessage struct {
 	Signature     []byte
 }
 
+func (p *PoetProofMessage) MarshalLogObject(encoder log.ObjectEncoder) error {
```
countvonzero, line 388: this struct now have inconsistent receiver.

common/types/activation.go
```diff
@@ -369,10 +411,40 @@ func (proofMessage PoetProofMessage) Ref() (PoetProofRef, error) {
 	return h.Bytes(), nil
 }
 
+type RoundEnd time.Time
+
+func (re RoundEnd) IntoTime() time.Time {
```
countvonzero: inconsistent receiver

activation/activation.go
poszu: I didn't notice because of the different keys and because it is a "short" version.

activation/nipost.go
poszu: I don't have any opinion as long as it is consistent. I changed it to use the default encoding for Hash32.

activation/nipost.go
poszu: Yes. Hitting this condition would mean that something is wrong with the poet. Changed to a warning log. Thanks!

activation/nipost.go
poszu: Yes. The proofs database is concurrently populated from the ATX handler. It might happen that this node has already received an ATX that references the same proof and it has been fetched from another node (and thus - it exists in the DB already).

activation/poet.go
poszu: I figured that it will be simpler to pass duration as it is an absolute value (no clock drift, no timezones etc.). The downside is that the calculated round end will be off (late) by an unknown value due to the time it takes for the GRPC response to arrive. We don't need an exact round end time. An estimation of when the proof _should_ be available is enough.

activation/nipost.go
poszu: You are right, it should be unified. I'd go with a `Hash32` as it is more specific.

poszu:
bors try

activation/nipost.go
countvonzero: but the database key contains a PoetServiceID. so this would mean we receive the same proof from the same poet server multiple times?

activation/poet.go
countvonzero: i pointed this out because using duration, the rpc response is not informational at a later time, i.e. no good for debugging, or if the data is stored somewhere and then relayed. i think it better to use absolute time (utc) in req/rep in general.
either way we will have clock drift.

countvonzero: for later consideration. no need to change now.

countvonzero (approved):

activation/nipost.go
poszu: There are two situations where a poet's proof is stored, happening concurrently:
- when building NIPoST - the proof is queried directly from the Poet service,
- when an ATX handler receives a new ATX - it validates the ATX, which includes checking if the proof is valid. The handler will fetch the poet proof from other nodes if it is not already in the database.
  It happens here: https://github.com/spacemeshos/go-spacemesh/blob/6dcadfdf7d66beb5dc3f0ebb67040e8ef0eaa0e2/activation/handler.go#L369-L372

countvonzero: ah, i see. so a node could potentially receive the proof it needs via processing ATXs from other miners.

poszu:
bors try

andres-spacemesh:
bors try

poszu:
bors merge

poszu:
bors -merge

poszu:
bors try

poszu:
bors merge-

poszu:
bors try

poszu:
bors try

poszu:
bors try-

poszu:
bors try

poszu:
bors try

fetch/mesh_data.go
```diff
@@ -126,14 +127,19 @@ func (f *Fetch) GetPoetProof(ctx context.Context, id types.Hash32) error {
...
-				log.String("hint", string(datastore.POETDB)),
-				log.Stringer("hash", id),
-				log.Err(pm.err))
-		}
 	}
-	return pm.err
+	switch {
+	case pm.err == nil:
+		return nil
+	case errors.Is(pm.err, activation.ErrObjectExists):
```
countvonzero, line 134: please add comment on why this could happen. it's not a normal condition in the node implementation.
for other objects, we don't have fetching mechanism other than the hash fetcher.

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
Bors merge

countvonzero:
bors merge

PR #3866, opened 2022-12-16
[Merged by Bors] - beacon sync: use weight unit and ballot weight for beacon sync

### LINKED ISSUE #3174: beacon sync: threshold for accepting beacon for epoch is incorrect
## Description

from Tal
I think the abstract protocol should be:
- Look at the last cluster of layers (a cluster is the smallest number of consecutive layers for which the expected number of weight units (in ballots) is at least 800; if we have expected 50 weight units per layer, then a cluster is 16 layers).
- For each ballot in the cluster, determine which beacon it supports, and add its weight to that beacon value
- Take the majority value (if there's no majority, use the plurality value, but in this case there should be a strong warning to the user)

The ballot weight is the associated ATX weight divided by the number of ballot slots this ATX committed to (there's a lower bound on the number of slots --- which depends on how much weight is 'visible' to the ATX ---  but not an upper bound).
Because we're sampling, we can't just take the entire ATX weight (otherwise, a "whale" ATX would dominate the voting).
Unless we support miners changing beacons mid-epoch (we talked about doing this to allow faster network rejoins), the way to decide which beacon a ballot supports is to take it from the first ballot from the same id in that epoch.

for more detailed discussion see
https://spacemesh.slack.com/archives/CG91F13C7/p1649770485044299

## Actual Behavior

- absolute number of ballot is used for the threshold, instead of weight unit.
- ballot's number of eligibility is not counted correctly
- ATX weight is used instead of ballot weight

## Expected Behavior

see above

countvonzero:
## Motivation

Closes #3174

## Changes

- use number of weight units (eligibilities) instead of number of ballots for beacon sync condition
- since beacon sync samples ballots for the correct beacon value of the epoch, use ballot weight instead of the whole atx weight for adding up total weight behind a beacon value to avoid undue influence on ATX whale 
- once enough weight unit is accumulated, take the beacon with majority weight for the epoch. if majority weight not available, take the plurality weight.


countvonzero:
bors try

proposals/eligibility_validator.go
```diff
@@ -140,7 +142,8 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot)
 		beacon,
 	)
 
-	v.beacons.ReportBeaconFromBallot(epoch, ballot.ID(), beacon, weight)
+	weightPer := fixed.DivUint64(atxWeight, uint64(numEligibleSlots))
```
dshulyak, line 145: this number was already computed somewhere in big.Rat, as it is basically weight / slots

proposals/eligibility_validator.go
```diff
@@ -140,7 +142,8 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot)
...
+	v.beacons.ReportBeaconFromBallot(epoch, ballot, beacon, weightPer)
```
dshulyak, line 146: what if ballots is with several eligibilities?

beacon/beacon.go
```diff
@@ -233,86 +235,111 @@ func (pd *ProtocolDriver) isClosed() bool {
...
 }
 
-func (pd *ProtocolDriver) recordBeacon(epochID types.EpochID, bid types.BallotID, beacon types.Beacon, weight uint64) {
+func (pd *ProtocolDriver) recordBeacon(epochID types.EpochID, ballot *types.Ballot, beacon types.Beacon, weightPer fixed.Fixed) {
 	pd.mu.Lock()
 	defer pd.mu.Unlock()
 
-	if _, ok := pd.beaconsFromBallots[epochID]; !ok {
-		pd.beaconsFromBallots[epochID] = make(map[types.Beacon]*ballotWeight)
+	// using ballot weight here because we're sampling miners for the beacon value recorded
```
dshulyak, line 257: i don't see sampling below

beacon/beacon.go
```diff
@@ -233,86 +235,111 @@ func (pd *ProtocolDriver) isClosed() bool {
...
+	// in the ballots. we can't just take the entire ATX weight because otherwise, a "whale"
+	// ATX would dominate the voting.
+	ballotWeight := weightPer.Mul(fixed.New(len(ballot.EligibilityProofs)))
+	weightUinit := len(ballot.EligibilityProofs)
```
dshulyak: weightUnit

proposals/eligibility_validator.go
countvonzero: are you talking about here?

https://github.com/spacemeshos/go-spacemesh/blob/e9d1a564bda11964b9a49f3d820757cd7989a140/proposals/util/util.go#L45

i decided to do the calculation here (instead of calling the library) because all the info (atx weight, active set and numEligibleSlots) was already calculated earlier in this function.

proposals/eligibility_validator.go
countvonzero: this number is multiplied by number of eligibilities here https://github.com/spacemeshos/go-spacemesh/blob/e9d1a564bda11964b9a49f3d820757cd7989a140/beacon/beacon.go#L260

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

fasmat:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #3867, opened 2022-12-16
[Merged by Bors] - p2p: set address that will be advertised to peers

### LINKED ISSUE #3610: p2p: allow to specify address that will be advertised by a node
in the implementation remote address is learned from the connection. for example consider two nodes N1 and N2. N1 technically is a bootnode and configured with a dns address. N2 knows that address and uses it to connect with N1, but also N2 is in the same cloud (or k8s cluster) as N1, so when N2 connects with N1 it will learn address from the private network. Later N3 joins the network and N1 shares that private address with N3, N3 is not running in the same cloud and can't connect with N2 this way. 

Also, assume that N2 was initially configured with a dns address for N1, however after direct connection was established it will overwrite that dns address in the addressbook with a private address.

## implementation
- add advertise-address to p2p config (address is anything that is resolvable by libp2p (e.g dns4, dns6, tcp4) including port, but without p2p identity)
- include this address when peerExchange.Request is executed
```go
type Request struct {
      // address should include only network information (ip4, dns4 namespace)
     Address string
     // Port is used if Address is empty, otherwise it should be encoded into the Address field
     Port uint16
}
``` 
- if advertise-address is not configured - Address field should be empty and peer should continue using ip from a connection

as an option:
- consider configuring multiple addresses, i think that addressbook doesn't accept more than one

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3610

added option to set advertise-address. it can be set to any valid libp2p address, that doesn't include p2p identity,
such as /dns4/bootnode.spacemesh.io/tcp/5003. peer will receive this address and add identity that will be known from handshake.

existing discovery behavior is kept backward compatible. if advertise-address is not provided node will try to pick either locally listened port or the port that was mapped by nat device. and advertise it as single-component multiaddr - /tcp/9090. and receiver will add ip address from the peer socket if it receives only port.

can be configured using command line flag:
```
--advertise-address=/dns4/bootnode.spacemesh.io/tcp/5003
```
or config option:
```json
{
    "p2p": {"advertise-address": "/dns4/bootnode.spacemesh.io/tcp/5003"}
}
```

### DIFF
--- a/p2p/peerexchange/discovery.go
+++ b/p2p/peerexchange/discovery.go
@@ -29,6 +29,7 @@ const (
 type Config struct {
 	Bootnodes            []string
 	DataDir              string
+	AdvertiseAddress     string             // Address to advertise to a peers.
 	CheckInterval        time.Duration      // Interval to check for dead|alive peers in the book.
 	CheckTimeout         time.Duration      // Timeout to connect while node check for dead|alive peers in the book.
 	CheckPeersNumber     int                // Number of peers to check for dead|alive peers in the book.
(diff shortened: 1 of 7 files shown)


p2p/peerexchange/discovery.go
```diff
@@ -75,32 +76,59 @@ func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
...
+			return nil, fmt.Errorf("address to advertise (%s) is invalid: %w", config.AdvertiseAddress, err)
+		}
+		for _, proto := range advertise.Protocols() {
+			if proto.Code == ma.P_P2P {
+				return nil, fmt.Errorf("address to advertise (%s) includes p2p identity", advertise.String())
+			}
+		}
+	} else {
+		var err error
+		advertise, err = ma.NewComponent("tcp", strconv.Itoa(int(portFromHost(logger, h))))
```
fasmat, line 94: any particular reason as to why `portFromHost` returns an `uint16` instead of just an `int`? Everywhere it is used the return value is cast to `int` anyway.

p2p/peerexchange/discovery.go
```diff
@@ -75,32 +76,59 @@ func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
...
+				case <-sub.Out():
+					port := portFromHost(logger, h)
+					if port != 0 {
+						advertise, err := ma.NewComponent("tcp", strconv.Itoa(int(portFromHost(logger, h))))
+						if err != nil {
+							logger.With().Error("failed to create tcp multiaddr", log.Err(err))
+						} else {
+							protocol.UpdateAdvertisedAddress(advertise)
+						}
+					}
```
fasmat, line 125: can advertise from line 94 reused here?

```suggestion
				case <-sub.Out():
					protocol.UpdateAdvertisedAddress(advertise)
```

it has already been checked to be valid as well.

fasmat (approved):
Just some minor suggestions for code readability improvements, besides that LGTM

dshulyak:
bors merge

Issue #3517, opened 2022-08-31
Use generated mocks over custom implemented ones where possible


fasmat:
## Description
Currently a mix of generated and handwritten mocks is used in various tests, mostly in the `activation` package. We should replace the handwritten mocks with generated ones.

## Affected code
- in `activation`
  - `LayerClockMock`: There is already `activation.NewMocklayerClock` that can be used instead of the handwritten mock
  - `NetMock` and `FaultyNetMock`: There is already `pubsub.NewMockPublisher`
  - `postSetupProviderMock` and `PostAPIMock`: There is already `activation.NewMockpostSetupProvider`
  - `NIPostBuilderMock`: There is already `activation.NewMocknipostBuilder`
- in `api/grpcserver`
  - various mocks that could be generated, some of them are already available in the `api/grpcserver/mocks` package
  - move used interfaces from `api/node.go` to `api/grpcserver/interfaces.go` and generate mocks for them in the same package to be used for tests

PR #3876, opened 2022-12-19
[Merged by Bors] - Fix data race in TestBuilder_NIPostPublishRecovery

poszu:
## Motivation
There is a data race when accessing mocked structure fields.
## Changes
Lock the structure's mutex for the accesses.

### DIFF
--- a/activation/activation_test.go
+++ b/activation/activation_test.go
@@ -1240,7 +1240,9 @@ func TestBuilder_NIPostPublishRecovery(t *testing.T) {
 	challengeHash, err := challenge.Hash()
 	assert.NoError(t, err)
 	npst2 := newNIPostWithChallenge(challengeHash, poetRef)
+	layerClockMock.mu.Lock()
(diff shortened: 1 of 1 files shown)


fasmat (approved):

fasmat:
I would prefer to have a proper fix by not using the same instance of `layerClock` between all tests and instead use a separate `gomock` of the `layerClock` interface for every test.

I don't know how much effort that is, but if it isn't too much could you do that instead?

poszu:
bors merge

poszu:
> I would prefer to have a proper fix by not using the same instance of `layerClock` between all tests and instead use a separate `gomock` of the `layerClock` interface for every test.
> 
> I don't know how much effort that is, but if it isn't too much could you do that instead?

I'd prefer it too 😉 . But it's not a priority right now. Do you mind creating a tech-debt issue for it?

fasmat:
@poszu https://github.com/spacemeshos/go-spacemesh/issues/3517

PR #3877, opened 2022-12-20
[Merged by Bors] - Update VRFSign and -Verify to include nonces from PoST

### LINKED ISSUE #201: Apply VRF nonce as a prefix in every VRF invocation for a given identity.
With https://github.com/spacemeshos/pm/issues/175 the nonce found during PoST initialization was added to the initial ATX of a node.

This nonce must now be used as input to every VRF invocation used in the beacon weak protocol.

For this the message that is used as an input to the VRF is prefixed with the nonce from the ATX.

fasmat:
## Motivation
Closes https://github.com/spacemeshos/pm/issues/201

## Changes
- The signing package has been completely rewritten with testing in mind
  - The services can be set up to fetch the required data from the db automatically by using `WithNonceFromDB`, alternatively for tests they can use fixed nonces with `WithNonceForNode` and fixed keys with `WithKeyFromRand` and `WithPrivateKey`.
  - Unit tests and fuzzing tests have been added for all functionality of the signing package
  - It is recommended to use mocked signers and verifiers in tests when possible
- Hard dependencies on the signing package have been reduced wherever possible and replaced by interfaces, which also simplifies testing with mocks
- Use domain types when possible and appropriate; e.g. instead of `PublicKeys` use `NodeID` if applicable
- Tests that were modified also had the following changes applied if applicable:
  - Use `context.Background()` over `context.TODO()` in tests
  - Use `require` when a failing assert should end the test and `assert` when not (e.g. during cleanup)
  - Use `t.Cleanup()` over `defer` in tests for cleanup steps
  - Replace naked go routines with `errgroup.Group`
- VRFPostIndex has been added to the ATXHeader so it can more easily be fetched from the DB and is cached.

## Test Plan
The rewritten signing package has received new tests. Existing tests have been updated to work with the new API.

## TODO

## DevOps Notes


fasmat:
bors try

beacon/beacon.go
```diff
@@ -969,13 +971,17 @@ func atxThreshold(kappa int, q *big.Rat, numATXs int) *big.Int {
 	return threshold
 }
 
-func buildSignedProposal(ctx context.Context, signer signing.Signer, epoch types.EpochID, logger log.Log) []byte {
+func buildSignedProposal(ctx context.Context, signer vrfSigner, epoch types.EpochID, logger log.Log) []byte {
 	p := buildProposal(epoch, logger)
-	signature := signer.Sign(p)
+	signature, err := signer.Sign(p)
+	if err != nil {
+		logger.With().Panic("failed to sign proposal", log.Err(err))
```
countvonzero:
```suggestion
		logger.With().Fatal("failed to sign proposal", log.Err(err))
```

beacon/handlers.go
```diff
@@ -26,15 +26,15 @@ const (
...
-	errProposalDoesntPassThreshold = errors.New("proposal doesn't pass threshold")
-	errAlreadyProposed             = errors.New("already proposed")
-	errAlreadyVoted                = errors.New("already voted")
-	errMinerATXNotFound            = errors.New("miner ATX not found in previous epoch")
-	errProtocolNotRunning          = errors.New("beacon protocol not running")
-	errEpochNotActive              = errors.New("epoch not active")
-	errMalformedMessage            = errors.New("malformed msg")
-	errUntimelyMessage             = errors.New("untimely msg")
+	errVRFNotVerified         = errors.New("proposal failed vrf verification")
+	errProposalBelowThreshold = errors.New("proposal below threshold")
```
countvonzero:
```suggestion
	errProposalNotEligible = errors.New("proposal not eligible")
```

countvonzero: actually the criteria is the opposite, proposal value is above threshold. in any case, lets make it more general

beacon/handlers.go
```diff
@@ -182,20 +182,20 @@ func (pd *ProtocolDriver) verifyProposalMessage(logger log.Log, m ProposalMessag
...
 	}
 
+	vrfPK := signing.NewPublicKey(m.NodeID.Bytes())
 	if err := pd.registerProposed(logger, m.EpochID, vrfPK); err != nil {
 		return types.ATXID{}, fmt.Errorf("[proposal] failed to register proposal (miner ID %v): %w", minerID, err)
 	}
 
 	if !pd.checkProposalEligibility(logger, m.EpochID, m.VRFSignature) {
-		return types.ATXID{}, fmt.Errorf("[proposal] not eligible (miner ID %v): %w", minerID, errProposalDoesntPassThreshold)
+		return types.ATXID{}, fmt.Errorf("[proposal] not eligible (miner ID %v): %w", minerID, errProposalBelowThreshold)
```
countvonzero:
```suggestion
		return types.ATXID{}, fmt.Errorf("[proposal] not eligible (miner ID %v): %w", minerID, errProposalNotEligible)
```

beacon/interface.go
```diff
@@ -8,10 +8,11 @@ import (
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/p2p"
 	"github.com/spacemeshos/go-spacemesh/p2p/pubsub"
+	"github.com/spacemeshos/go-spacemesh/signing"
 	"github.com/spacemeshos/go-spacemesh/timesync"
 )
 
-//go:generate mockgen -package=beacon -destination=./mocks.go -source=./interface.go
+//go:generate mockgen -package=mocks -destination=./mocks/mocks.go -source=./interface.go
```
countvonzero: i thought in general you want to flatten the pkg and leave mocks in the same layer as the interface?

signing/keyExtractor.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	return &PublicKey{PublicKey: pub}, nil
+}
+
+// ExtractNodeID from a signature.
+func (e PubKeyExtractor) ExtractNodeID(m, sig []byte) (types.NodeID, error) {
+	msg := make([]byte, len(m)+len(e.prefix))
+	copy(msg, e.prefix)
+	copy(msg[len(e.prefix):], m)
+	pub, err := ed25519.ExtractPublicKey(msg, sig)
```
countvonzero: would this cause error if pub is `nil`?

i think it's better to reuse code from Extract()

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 require (
 	crawshaw.io/sqlite v0.3.3-0.20211227050848-2cdb5c1a86a1
 	github.com/ALTree/bigfloat v0.0.0-20220102081255-38c8b72a9924
-	github.com/chaos-mesh/chaos-mesh/api v0.0.0-20221213145053-386291e73746
```
countvonzero: i'd really prefer to remove unnecessary files from this PR. go.mod and go.sum

signing/signer.go
```diff
@@ -11,70 +12,90 @@ import (
...
+	return sig, nil
 }
 
 // Sign signs the provided message.
 func (es *EdSigner) Sign(m []byte) []byte {
-	if es.prefix != nil {
-		m = append(es.prefix, m...)
-	}
-	return ed25519.Sign2(es.priv, m)
+	msg := make([]byte, len(m)+len(es.prefix))
```
countvonzero, line 95: check if prefix is nil?

cmd/node/node.go
```diff
@@ -960,7 +973,10 @@ func (app *App) LoadOrCreateEdSigner() (*signing.EdSigner, error) {
 		log.With().Info("created new identity", edSgn.PublicKey())
 		return edSgn, nil
 	}
-	edSgn, err := signing.NewEdSignerFromKey(data, signing.WithPrefix(app.Config.Genesis.GenesisID().Bytes()))
+	edSgn, err := signing.NewEdSigner(
+		signing.WithPrivateKey(data),
```
countvonzero, line 977: it's strange that private key is passed as option. shouldn't it be a required field?

signing/keyExtractor.go
```diff
@@ -0,0 +1,62 @@
...
+	extractor := &PubKeyExtractor{
+		prefix: cfg.prefix,
+	}
+	return extractor, nil
+}
+
+// Extract public key from a signature.
+func (e PubKeyExtractor) Extract(m, sig []byte) (*PublicKey, error) {
+	msg := make([]byte, len(m)+len(e.prefix))
+	copy(msg, e.prefix)
```
countvonzero, line 46: check if prefix is nil?

signing/vrf.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Sign signs a message for VRF purposes.
+func (s VRFSigner) Sign(msg []byte) ([]byte, error) {
+	nonce, err := s.fetcher.NonceForNode(s.nodeID)
+	if err != nil {
+		return nil, err
+	}
+
+	buf := make([]byte, 8)
+	binary.LittleEndian.PutUint64(buf, uint64(nonce))
```
countvonzero, line 37: i always get nervous when the code does this instead of using the encoding library. why not put in a private struct and use scale?
@dshulyak

signing/vrf.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Verify that signature matches public key.
+func (v VRFVerifier) Verify(nodeID types.NodeID, msg, sig []byte) bool {
+	nonce, err := v.fetcher.NonceForNode(nodeID)
+	if err != nil {
+		return false
+	}
+
+	buf := make([]byte, 8)
+	binary.LittleEndian.PutUint64(buf, uint64(nonce))
```
countvonzero, line 155: same here.

signing/keyExtractor.go
fasmat: not needed: https://go.dev/play/p/TPej-2ISuli

`len(nil)` is 0 and `copy(msg, nil)` is a no-op

cmd/node/node.go
fasmat: this is the only place in the whole codebase where `WithPrivateKey` is used. Everywhere else the signer is instantiated with `signing.NewEdSigner()` and will generate a random key.

Before there were 2 functions to create a signer: `NewEdSigner` and `NewEdSignerFromKey`, the later one has been replaced with this functional option.

signing/signer.go
fasmat: not needed, see my comment here: https://github.com/spacemeshos/go-spacemesh/pull/3877#discussion_r1056031159

signing/keyExtractor.go
fasmat: if `pub` is `nil` `types.BytesToNodeID` will return an empty `[32]byte` which is the base for the `NodeId` type. This is safe to do and will also result in a `NodeID` that equals `types.EmptyNodeID`.

re-using `Extract` here requires me to check for an error twice (once in `Extract` and once in `ExtractNodeID` or I'll get a panic due to a `nil` pointer dereference in the case of an error:

```golang
func (e PubKeyExtractor) ExtractNodeID(m, sig []byte) (types.NodeID, error) {
	pub, err := e.Extract(m, sig)
	if err != nil {
		return types.EmptyNodeID, err
	}
	return types.BytesToNodeID(pub.PublicKey), nil
}
```

Would you prefer this over the current version?

signing/vrf.go
fasmat: I would like to avoid `go-scale` here. `msg` is already a serialized object I'm only prefixing it with 8 bytes before I sign it. The resulting `buf` would be the same doing it like this or with `go-scale` anyway, but with go-scale I have to create another type and generate code for it to be encoded. That's a lot of extra code to just prefix a few bytes...

go.mod
fasmat: The versions used to build `go-spacemesh` are the ones in this `go.mod` file even if I do not update it. Dependencies are overwritten if a dependent package uses a newer version of a package listed here.

These are updates from `go work sync` the only thing it does is to synchronize dependencies between our repositories, it does not update.

beacon/interface.go
fasmat: I still think so, this must be a copy & paste mistake, I'll change it back if you want 🙂

fasmat:
bors try

fasmat:
bors try

beacon/interface.go
countvonzero: yes please.

go.mod
countvonzero: i still don't understand the relevance. are you talking about indirect dependencies?
which other repository does this PR require to sync with?

signing/keyExtractor.go
countvonzero: checking error twice is trivial. using the same code to extract is more important to me.

signing/vrf.go
countvonzero: the amount of code doesn't bother me. consistency across different node implementation is my main concern.
node a rust implementation cannot just rely on scale-encoding. the rust dev now has to care about what order you prefix the nonce?

having opaque []byte in the struct shouldn't be an issue for encoding.

signing/signer.go
```diff
@@ -98,9 +119,22 @@ func (es *EdSigner) LittleEndian() bool {
 }
```
countvonzero: does EdSigner needs this interface?

countvonzero (approved):
LG in general.

for the remaining comments, feel free to file issues and do it later if you agree.

signing/keyExtractor.go
fasmat: Updated code for ExtractNodeID

go.mod
fasmat: what I mean is we are already using newer versions of dependencies in other packages. I only did `go work sync` to align the versions of dependencies we use between all of our codebase.

When compiling the code it will not necessarily use what's in `go.mod` but might overwrite that dependency when another package uses a newer version of it. `go work sync` just makes visible what actually is compiled taking all (also indirect) dependencies into account.

I reverted this file now and will update again in a new PR 🙂

signing/signer.go
fasmat: Before my changes it did, but not any more (only `VRFSigner` still needs it). I'll remove it

fasmat:
bors merge

signing/vrf.go
fasmat: for now I will merge the code as is. I created an issue to come back to the encoding here: https://github.com/spacemeshos/go-spacemesh/issues/3902

signing/vrf.go
fasmat: responded here: https://github.com/spacemeshos/go-spacemesh/pull/3877#discussion_r1056546232

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #3880, opened 2022-12-20
[Merged by Bors] - hare: restart nofify counter in every iteration

### LINKED ISSUE #3783: Hare: bug in termination condition
https://github.com/spacemeshos/go-spacemesh/blob/b2da54055571b1d8e6724dc96e75aac17befc92f/hare/notifytracker.go#L42

On page 64 of the latest Hare specification, the termination condition for Hare is described as:

```
If at any point in the protocol, Pi party gathers f+1 notication headers (S; notify; k) from distinct parties on a set S, it terminates.
```

In this context `k` is a bound variable referencing the current iteration of the protocol. Therefore, the termination condition requires achieving at least f+1 notification messages in the CURRENT ITERATION. In other words, the counter of notifications should be restarted at the beginning of every iteration.

In the current implementation this is not the case. The counter is initialized once and the accumulation of notifications runs for the whole Hare run. This is a bug.

Caution:  This bug was not discovered before because it would only manifest during an equivocation-based attack. Using this hole, a colluding minority of malicious nodes could trick the protocol into accepting "wrong" consensus result..

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3783

### DIFF
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -176,7 +176,6 @@ type consensusProcess struct {
 	notifyTracker     *notifyTracker
 	cfg               config.Config
 	pending           map[string]*Msg // buffer for early messages that are pending process
-	notifySent        bool            // flag to set in case a notification had already been sent by this instance
 	mTracker          *msgsTracker    // tracks valid messages
 	terminating       bool
 	eligibilityCount  uint16
@@ -199,7 +198,6 @@ func newConsensusProcess(cfg config.Config, instanceID types.LayerID, s *Set, or
 		nid:               nid,
 		publisher:         p2p,
 		preRoundTracker:   newPreRoundTracker(cfg.F+1, cfg.N, logger),
-		notifyTracker:     newNotifyTracker(cfg.N),
 		cfg:               cfg,
 		terminationReport: terminationReport,
 		pending:           make(map[string]*Msg, cfg.N),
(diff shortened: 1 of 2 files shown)


hare/algorithm.go
```diff
@@ -739,6 +730,11 @@ func (proc *consensusProcess) processCommitMsg(ctx context.Context, msg *Msg) {
 }
 
 func (proc *consensusProcess) processNotifyMsg(ctx context.Context, msg *Msg) {
+	if proc.notifyTracker == nil {
+		proc.WithContext(ctx).With().Error("notify tracker is nil")
```
poszu: `With()` is redundant here.
```suggestion
		proc.WithContext(ctx).Error("notify tracker is nil")
```

poszu (approved):

hare/algorithm.go
```diff
@@ -739,6 +730,11 @@ func (proc *consensusProcess) processCommitMsg(ctx context.Context, msg *Msg) {
 }
 
 func (proc *consensusProcess) processNotifyMsg(ctx context.Context, msg *Msg) {
+	if proc.notifyTracker == nil {
```
countvonzero, line 733: does it make sense to assign one whichever (beginNotifyRound or processNotifyMsg) happens first?

dshulyak:
bors merge

PR #3881, opened 2022-12-20
[Merged by Bors] - Use multierror pkg for returning mutliple errors as `error`

poszu:
## Motivation
The [multierror](https://github.com/hashicorp/go-multierror) pkg greatly simplifies returning a collection of errors (`[]error`). It provides an `Error` type that supports appending many errors and implements `Error` interfaces and `Unwrap()`.

It also provides a `Group` that works similarly to `errgroup.Group`, but aggregates returned errors. It's useful when we want all goroutines to continue in spite of one of them failing **and** collect all the errors.

## Changes
- use `multierror.Error` where `[]error` was returned,
- use `multierror.Group` where errors from goroutines need to be aggregated.


fasmat:
This will be part of the core language in 2 months: https://tip.golang.org/doc/go1.20#errors making the `go-multierr` package obsolete.

fasmat:
> It also provides a `Group` that works similarly to `errgroup.Group`, but aggregates returned errors. It's useful when we want all goroutines to continue in spite of one of them failing and collect all the errors.

The go routines of `errgroup.Group` also continue to run if one of the other routines returns an `error` and `Wait` only returns after all goroutines have finished. The only way to exit a goroutine early is if it uses the `context` returned from `errgroup.WithContext` to return early: https://go.dev/play/p/bf2QJpOZGR6

poszu:
> > It also provides a `Group` that works similarly to `errgroup.Group`, but aggregates returned errors. It's useful when we want all goroutines to continue in spite of one of them failing and collect all the errors.
> 
> The go routines of `errgroup.Group` also continue to run if one of the other routines returns an `error` and `Wait` only returns after all goroutines have finished. The only way to exit a goroutine early is if it uses the `context` returned from `errgroup.WithContext` to return early: https://go.dev/play/p/bf2QJpOZGR6

What `errgroup.Group` provides is entirely different. `errgroup.Group::Wait` doesn't return aggregated errors. It returns the first error only. See https://go.dev/play/p/2xucP8f4rdD. The essence of `multierror.Group` is that it returns all the errors that happened in the spawned goroutines.

fasmat:
> What errgroup.Group provides is entirely different. errgroup.Group::Wait doesn't return aggregated errors. It returns the first error only. See https://go.dev/play/p/2xucP8f4rdD. The essence of multierror.Group is that it returns all the errors that happened in the spawned goroutines.

I am aware 🙂. I was just referring to

> we want all goroutines to continue in spite of one of them failing

which is already the case.

countvonzero (approved):
nice change! thanks

poszu:
bors merge

PR #3886, opened 2022-12-21
Run systests in CI with Go's race detector

poszu:
## Motivation
Go supports running binaries with a race detector. As we approach genesis, we could leverage this to see if there are any race conditions in "live" conditions.

## Changes
Build go-spacemesh docker image for system tests with `-race` flag.


poszu:
bors try

dshulyak:
i am not sure about merging it right now, enabling race detector is not free. lets make sure that tests are completely stable first

poszu:
> i am not sure about merging it right now, enabling race detector is not free. lets make sure that tests are completely stable first

I totally agree, I don't want to degrade CI stability any further.

countvonzero:
in any case, if we choose to do it, we probably should build two binaries, one with race detector and one without for release.

poszu:
Closing as postponed until we have a truly stable CI

PR #3888, opened 2022-12-21
[Merged by Bors] - data-driven tortoise

dshulyak:
all interactions with tortoise are through the data-driven interface, with the following goals:
- improve reasoning about tortoise dependencies and timings
- predictable performance (for example loading many objects from disk is not predictable)
- enable tracing and further simplify testing

specifically for tracing the plan is to dump all passed data inputs to disk on the subset of nodes in the cloud, and then rerun tortoise in exact observed order. this will be an upcoming change.

other changes:
- state from disk is loaded by using Recover method, same approach is reused for tests as they were writing generated state to disk, and refactoring all of them is a long process
- additionally weak coin and beacon have to be reported externally


beacon/beacon_test.go
```diff
@@ -267,6 +268,16 @@ func TestBeacon_NoProposals(t *testing.T) {
...
 }
 
+func getNoWait(tb testing.TB, results <-chan result.Beacon) result.Beacon {
+	select {
+	case rst := <-results:
+		return rst
+	default:
+	}
+	require.Fail(tb, "beacon is not available")
+	panic("unreachable")
```
countvonzero: why is this needed?

beacon/beacon.go
```diff
@@ -262,9 +266,26 @@ func (pd *ProtocolDriver) Close() {
...
+	default:
+		pd.logger.With().Error("results queue is congested",
+			log.Uint32("epoch_id", epoch.Uint32()),
+			log.Stringer("beacon", beacon),
+		)
+	}
+}
+
+// Results notifies waiter that beacon for a target epoch has completed.
+func (pd *ProtocolDriver) Results() <-chan result.Beacon {
```
countvonzero, line 285: i worry the next dev will misuse this interface. 

for example, if proposal builder wants to get notified by the new beacon, and share with the network immediately (a form of early beacon sync), then we get into a race condition between tortoise and proposal builder.

common/types/result/beacon.go
```diff
@@ -0,0 +1,8 @@
+package result
+
+import "github.com/spacemeshos/go-spacemesh/common/types"
+
+type Beacon struct {
```
countvonzero, line 5: i don't understand the rationale behind this package structure. are there more data types planned for this package? why not under types directly?

tortoise/algorithm.go
```diff
@@ -197,13 +198,22 @@ func (t *Tortoise) OnAtx(atx *types.ActivationTxHeader) {
...
 }
 
-// OnBlock should be called every time new block is received.
-func (t *Tortoise) OnBlock(block *types.Block) {
+// OnBlock updates tortoise with information that data is available locally.
+func (t *Tortoise) OnBlock(header types.BlockHeader) {
+	start := time.Now()
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	waitBlockDuration.Observe(float64(time.Since(start).Nanoseconds()))
```
countvonzero, line 206: this should be after onBlock?

tortoise/algorithm.go
```diff
@@ -197,13 +198,22 @@ func (t *Tortoise) OnAtx(atx *types.ActivationTxHeader) {
...
+	t.trtl.onBlock(header, true, false)
+}
+
+// OnValidBlock is used for loading stored historical data or possibly opinion from peers. It lets tortoise know that data is available locally, and block is supported by local opinion (required for verifying tortoise).
```
countvonzero: nit: seems this line is very long

countvonzero (approved):

tortoise/algorithm.go
dshulyak: > waitBlockDuration

this is to track how long it waited on mutex.

common/types/result/beacon.go
dshulyak: i am not sure if it is well considered, i just needed values that are grouped together and named similarly to existing values.

so e.g. there is a Layer, Block and now Beacon. i could make it is a sub-package in a package where this is created if it seems better, e.g tortoise/result or beacon/result. what would be better options?

beacon/beacon.go
dshulyak: i imagined that such notifications will be done in same place, for example

```
func watchBeacons(results <-chan Beacon) error {
     for beacon := results {
           tortoise.OnBeacon(beacon)
           builder.OnBeacon(beacon)
           hare.OnBeacon(beacon)
           beacons.Add(beacon) // persist
     }
}
```

maybe once every 6 month there will be new component that needs to watch beacon, so it will be just added to this loop, but this is still very easy to digest

beacon/beacon_test.go
dshulyak: otherwise i need to write return, just to satisfy compiler

dshulyak:
bors try

beacon/beacon_test.go
dshulyak: changed to return empty struct

common/types/result/beacon.go
dshulyak: let me know if you prefer another approach, i will then change in another pr

dshulyak:
bors merge

Issue #3729, opened 2022-11-09
dependencies should be fully processed before processing dependant object


dshulyak:
protocol objects have many dependencies, that must be stored in db and related memory structures (such as tortoise state), examples:
- refballot -> ballot dependency for reading active set and beacon
- base ballot -> ballot dependency for decoding votes

it looks like that there is a code path that allows concurrent execution, yet it is unclear in what cases. in the example below ballot was passed to the tortoise, before base ballot and failed to decode as expected:
 
![Screenshot from 2022-11-08 09-22-52](https://user-images.githubusercontent.com/1964576/200742737-aa3cf287-4ba3-4df2-80ad-c557833f46ab.png)

dshulyak:
i think to fix this problem all dependencies have to be saved into database after they are loaded into tortoise.

dshulyak:
there are two more possible problems in this area:
- [x] persisting atxs before adding them to tortoise may result in undefined behavior https://github.com/spacemeshos/go-spacemesh/pull/3904
- [ ] persisting ballots before storing decoded ballot may result in undefined behavior

countvonzero:
> it looks like that there is a code path that allows concurrent execution, yet it is unclear in what cases.

i think this is fixed by https://github.com/spacemeshos/go-spacemesh/pull/3778

> i think to fix this problem all dependencies have to be saved into database after they are loaded into tortoise.

can you elaborate more on the reasoning behind this conclusion? it is not obvious to me.

dshulyak:
every handler checks if dependency exists locally before fetching it. for example base ballot -> ballot dependency.
if it is stored in the database before loading in the tortoise then another thread may process ballot before first one loaded base ballot into tortoise and decoding will fail.

example with two threads t1 and t2 and ballots b1 -> b2 

1. t1 receives b1
2. t2 receives b2
3. t1 adds b1 to the database
4. t2 checks that b1 in the database, adds b2 to the database, runs tortoise decoding step and it fails
5. t1 adds b1 to the tortoise 

i realized that fetcher pr slightly improves this. but it only affects objects that are requested by a node. any two objects can be processed concurrently from broadcast. leaving it like this is just not robust

countvonzero:
thank you for the explanation. that makes sense. this behavior remains because
- concurrent gossip processing
- recursive data availability check

@dshulyak do you think we could add a callback for database save? most of the time it's a noop. but for atx/ballot/block we add the data to tortoise after they are saved in db.

[...]

PR #3904, opened 2022-12-26
pass atx to tortoise before storing it

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/3729


poszu (approved):

countvonzero:
as discussed in https://github.com/spacemeshos/go-spacemesh/issues/3729
i feel we need a more generic solution to only updating state after data is persisted.

PR #3913, opened 2022-12-28
[Merged by Bors] - Split unittests to linux and non-linux to speedup CI execution

pigmej:
linux is one group and windows & mac are another. Systests will wait only for Linux execution.


poszu (approved):

.github/workflows/ci.yml
```diff
@@ -145,15 +145,9 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
```
poszu, line 153: NIT

The ubuntu-only job could be cleaned up from all `matrix.os` and `matrix.allow-failure` uses.

pigmej: After some thinking I'd like to keep it here for clarity in the job overview etc. Then all unitests are visible in the same way etc.

pigmej:
bors try

pigmej:
bors try

.github/workflows/ci.yml
fasmat: I think we don't need a matrix definition here at all.

The job now only runs on `ubuntu-latest` with `allow-failure` always being true; not using a matrix would allow to simplify the job definition.

.github/workflows/ci.yml
```diff
@@ -162,6 +156,58 @@ jobs:
...
+        uses: mikepenz/action-junit-report@v3
+        # always run even if the previous step fails
+        if: always()
+        with:
+          report_paths: "**/unit-tests.xml"
+          annotate_only: true
+          fail_on_failure: ${{ matrix.allow-failure == 'false' }}
+
+
+  unittests-slower:
```
fasmat, line 201: I think `unittests-non-linux` or similar would be a more expressive name 🙂 or alternatively call this job `unittests` and the other `unittests-linux`

pigmej: Well there could be m1 also coming etc. I kinda left that open "just in case".

pigmej:
bors merge

PR #3914, opened 2022-12-28
[Merged by Bors] - Run systests in CI on PRs

poszu:
## Motivation
The current approach to test changes that PR brings is to run `bors try`. It has, however, a drawback that **only one** systest workflow can run at the same time, effectively serializing runs for `bors try`. Given a single try takes approximately 2 hours, a developer might need to check a few hours (2, 4, 6 etc.) to see the results of tests. It is unacceptable and slows down development a lot.

## Changes
Run systests after unit tests right in the PR workflow.

💡 One can still use `bors try` to verify if tests would pass after merge with `develop` branch.


dshulyak:
i think the idea of using bors try is to limit who can spawn workload in the cloud to the people from organization

pigmej:
@dshulyak we can control that on org level. We can make sure that *any* external contributor requires explicit approval before pipeline is triggered.

dshulyak (approved):
is there a way to run trigger systest job on demand using github api? i would be a bit worried that cloud can be overwhelmed if many people will open prs simultaneously. lets say 5 prs, 4 clusters in each 50 nodes, 1000 nodes, each 0.5 cpu, atleast 500 cores. thats ~15 vms atleast

poszu:
> is there a way to run trigger systest job on demand using github api? i would be a bit worried that cloud can be overwhelmed if many people will open prs simultaneously. lets say 5 prs, 4 clusters in each 50 nodes, 1000 nodes, each 0.5 cpu, atleast 500 cores. thats ~15 vms atleast

I think that the number of concurrent runs of a job can be limited with job _concurrency_, right @pigmej?

Edit: we could also run "small" systests automatically on PR and "big" systests on `bors try` and `bors merge`.

pigmej:
The concurrency IS NOT a number, it's more like a group.

One of the possibilities would be applying k8s limits, and then jobs would be pending on k8s scheduler (but we would risk being stuck if half-half would schedule instead of one big).

There is no real "number" there. We could do something with `Expressions` and do some artificial queue or schedule.

On the other hand, I still believe that it's cheaper to have 15 nodes cluster for 2 hours than run 5 nodes and wait too long for execution.

pigmej:
Answering @dshulyak and on-demand question would be, yes it's possible. We could utilize https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ and find some way to trigger it per PR (not all our prs are FROM the spacemeshos org)

poszu:
bors try

poszu:
bors cancel

poszu:
bors try

poszu:
bors try

poszu:
bors try

pigmej:
Bors try

poszu:
bors merge

fasmat:
If I understand the .yml correctly then with this change the systests will be executed on every push on every branch? Won't this lead to issues when people are working on e.g. draft PRs?

Additionally I think the systests would be executed 3 times per PR: once on the push, once via `bors merge` and then again after the merge when `develop` is updated?

poszu:
> If I understand the .yml correctly then with this change the systests will be executed on every push on every branch? Won't this lead to issues when people are working on e.g. draft PRs?
> 
> Additionally I think the systests would be executed 3 times per PR: once on the push, once via `bors merge` and then again after the merge when `develop` is updated?

Why do you think they are executed on every push? They are executed on:
- push to branch _trying_ (`bors try`),
- push to branch _staging_ (`bors merge`),
- on-demand byuser triggering it in GitHub Actions page.

fasmat:
I was missing something like `if: github.event_name == 'push' && github.branch_name == 'trying'` on the `systest`/`dockerpush` job definition but I just saw that there is:

```yaml
on:
  workflow_dispatch:
  push:
    branches:
      - staging
      - trying
```

at the top of the file, so its probably fine 👍

poszu:
bors try

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
Bors merge

poszu:
Bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

poszu:
bors merge

PR #3924, opened 2022-12-30
[Merged by Bors] - ATX Builder early detection of stale challenges.

### LINKED ISSUE #3917: ATX builder tries to build nipost with a stale challenge
## Description
TLDR; A node fails to build an ATX after smapp has been down for over an hour (more than an epoch) because it tries to use a stale nipost challenge.

```logs
2022-12-26T21:06:01.499-0500	INFO	ad66d.atxBuilder   	ready to build first atx	{"node_id": "ad66d4f5cd9f1e55ed856df70bb7855fcb6c65d1ef04ae6ae9f0b760582be759", "module": "atxBuilder", "current_layer": "3783", "current_epoch": "63", "name": "atxBuilder"}
2022-12-26T21:06:01.501-0500	INFO	ad66d.atxBuilder   	using existing atx challenge	{"node_id": "ad66d4f5cd9f1e55ed856df70bb7855fcb6c65d1ef04ae6ae9f0b760582be759", "module": "atxBuilder", "sessionId": "8c5da117-e062-4413-bbd8-25305b794d1f", "current_epoch": "63", "name": "atxBuilder"}
2022-12-26T21:06:01.501-0500	INFO	ad66d.atxBuilder   	new atx challenge is ready	{"node_id": "ad66d4f5cd9f1e55ed856df70bb7855fcb6c65d1ef04ae6ae9f0b760582be759", "module": "atxBuilder", "sessionId": "8c5da117-e062-4413-bbd8-25305b794d1f", "current_epoch": "63", "publish_epoch": "63", "target_epoch": "64", "name": "atxBuilder"}
2022-12-26T21:06:01.501-0500	INFO	ad66d.atxBuilder   	challenge ready	{"node_id": "ad66d4f5cd9f1e55ed856df70bb7855fcb6c65d1ef04ae6ae9f0b760582be759", "module": "atxBuilder", "name": "atxBuilder"}
2022-12-26T21:06:01.502-0500	INFO	ad66d.atxBuilder   	building NIPost	{"node_id": "ad66d4f5cd9f1e55ed856df70bb7855fcb6c65d1ef04ae6ae9f0b760582be759", "module": "atxBuilder", "poet round start": "2022-12-26 21:10:00.498 -0500 EST", "pub_epoch": "63", "deadline time": "2022-12-26T21:05:30.498-0500", "name": "atxBuilder"}
2022-12-26T21:06:01.516-0500	ERROR	ad66d.atxBuilder   	error attempting to publish atx	{"node_id": "ad66d4f5cd9f1e55ed856df70bb7855fcb6c65d1ef04ae6ae9f0b760582be759", "module": "atxBuilder", "sessionId": "8c5da117-e062-4413-bbd8-25305b794d1f", "layer_id": 3783, "epoch_id": 63, "errmsg": "create ATX: failed to build NIPost: builder: didn't receive any poet proof", "name": "atxBuilder"}
```

It doesn't detect that the challenge is stale and it will not be able to build a valid nipost with it. Especially:
- the current time is `2022-12-26T21:06:01`
- "current_epoch": "63"
- "publish_epoch": "63"
- "target_epoch": "64"
- (next) "poet round start": "2022-12-26 21:10:00" - poet round (that would give proof on time to publish in epoch 63) has already started. (The "poet round start" in the logs is for the next round).
- "deadline time": "2022-12-26T21:05:30.498" - the deadline to wait for a poet proof is in the past and hence it terminates immediately.

## Actual Behavior
`atxBuilder` picks up a stale nipost challenge.

## Expected Behavior
`atxBuilder` should determine that the challenge is stale and create a new one.

## Environment
 - OS: 
 
 - Node Version: 
 
 - Smapp version (if applicable): 
 

## Additional Resources 
[spacemesh-log-55dc4325e8d7f8f9824a47d646599420a2cf4516 (1).zip](https://github.com/spacemeshos/go-spacemesh/files/10318520/spacemesh-log-55dc4325e8d7f8f9824a47d646599420a2cf4516.1.zip)

poszu:
## Motivation
When the ATX Builder tries to publish an ATX, it first creates a challenge that commits to a certain publication layer. It might happen that the challenge is not technically stale _yet_ (the `PubLayerID` is in the future), but it's impossible to create a poet proof on time as the respective poet rounds have already started. In such a case, it makes no sense to submit that challenge to poets. The builder should detect it and create a fresh challenge targeting the next epoch.

See #3917 where it happens.

## Changes
Builder verifies if the poet round, which would yield proof on time to be able to publish ATX in the requested layer (by `challenge.PubLayerID`), has not started yet.

It allows discarding stale challenges early, before prompting Poets.

Additionally, removed `challenge` from Builder's state. It's now only keeped as a local var and persisted in DB for recovery.

## Test Plan
- unit tests


poszu:
bors try

poszu:
bors try

poszu:
bors try

activation/activation.go
```diff
@@ -410,7 +404,7 @@ func (b *Builder) loop(ctx context.Context) {
 				case <-poetRetryTimer.C:
 				}
 			default:
-				b.log.WithContext(ctx).Warning("Unknown error")
+				b.log.WithContext(ctx).Warning("Unknown error", log.Err(err))
```
countvonzero: With()

activation/activation.go
```diff
@@ -555,27 +548,41 @@ func (b *Builder) getCommitmentAtx(ctx context.Context) (*types.ATXID, error) {
 
 // PublishActivationTx attempts to publish an atx, it returns an error if an atx cannot be created.
 func (b *Builder) PublishActivationTx(ctx context.Context) error {
-	b.discardChallengeIfStale()
 	logger := b.log.WithContext(ctx)
 
-	if b.challenge != nil {
+	challenge, err := b.loadChallenge()
+	if err != nil {
+		b.log.With().Info("challenge not loaded", log.Err(err))
```
countvonzero: use logger (already has context inside)

activation/activation.go
```diff
@@ -555,27 +548,41 @@ func (b *Builder) getCommitmentAtx(ctx context.Context) (*types.ATXID, error) {
...
+	}
+	if challenge != nil && isStale(challenge, b.currentEpoch()) {
+		b.log.With().Info("atx challenge is stale - discarding it",
```
countvonzero: ditto

activation/activation.go
```diff
@@ -555,27 +548,41 @@ func (b *Builder) getCommitmentAtx(ctx context.Context) (*types.ATXID, error) {
...
 		log.Stringer("current_epoch", b.currentEpoch()),
-		log.Stringer("publish_epoch", b.challenge.PublishEpoch()),
-		log.Stringer("target_epoch", b.challenge.TargetEpoch()))
+		log.Stringer("publish_epoch", challenge.PublishEpoch()),
+		log.Stringer("target_epoch", challenge.TargetEpoch()))
 
 	if b.pendingATX == nil {
 		var err error
-		b.pendingATX, err = b.createAtx(ctx)
+		b.pendingATX, err = b.createAtx(ctx, challenge)
```
countvonzero, line 576: nice. will make testing easier

activation/activation.go
```diff
@@ -639,28 +644,38 @@ func (b *Builder) createAtx(ctx context.Context) (*types.ActivationTx, error) {
...
+		log.Time("poet proof deadline", poetProofDeadline),
+		log.FieldNamed("current epoch", b.currentEpoch()),
+		log.FieldNamed("publish epoch", pubEpoch),
+		log.FieldNamed("target epoch", challenge.TargetEpoch()),
 	)
 
-	challenge := types.PoetChallenge{
-		NIPostChallenge: b.challenge,
+	now := time.Now()
+	if poetRoundStart.Before(now) {
```
countvonzero, line 650: why not check it right away? the extra info log above seems wasted.

activation/activation.go
```diff
@@ -639,28 +644,38 @@ func (b *Builder) createAtx(ctx context.Context) (*types.ActivationTx, error) {
...
 	poetProofDeadline := poetRoundEnd.Add(b.poetCfg.GracePeriod)
 
 	b.log.With().Info("building NIPost",
-		log.Stringer("poet round start", nextPoetRoundStart),
-		log.Stringer("pub_epoch", pubEpoch),
-		log.Time("deadline time", poetProofDeadline),
+		log.Time("poet round start", poetRoundStart),
+		log.Time("next poet round start", nextPoetRoundStart),
+		log.Time("poet proof deadline", poetProofDeadline),
+		log.FieldNamed("current epoch", b.currentEpoch()),
```
countvonzero, line 644: :s/FieldNamed/Stringer

activation/activation.go
```diff
@@ -775,3 +779,7 @@ func SignAtx(signer signer, atx *types.ActivationTx) error {
 	atx.Sig = signer.Sign(bts)
 	return nil
 }
+
+func isStale(challenge *types.NIPostChallenge, currentEpoch types.EpochID) bool {
+	return challenge.TargetEpoch() < currentEpoch
```
fasmat: Is this correct? Lets say `PublishEpoch` is , `TargetEpoch` is 5 and `currentEpoch` is also 5 can the ATX still be published? I think this should be:

```suggestion
	return challenge.PublishEpoch() < currentEpoch
```

wdyt?

activation/activation.go
```diff
@@ -775,3 +779,7 @@ func SignAtx(signer signer, atx *types.ActivationTx) error {
 	atx.Sig = signer.Sign(bts)
 	return nil
 }
+
+func isStale(challenge *types.NIPostChallenge, currentEpoch types.EpochID) bool {
```
fasmat: can we make this a method of the builder; or even just inline it? The context of "this is a check for staleness" is given by the log messages already.

activation/activation.go
```diff
@@ -555,27 +548,41 @@ func (b *Builder) getCommitmentAtx(ctx context.Context) (*types.ATXID, error) {
...
 func (b *Builder) PublishActivationTx(ctx context.Context) error {
-	b.discardChallengeIfStale()
 	logger := b.log.WithContext(ctx)
 
-	if b.challenge != nil {
+	challenge, err := b.loadChallenge()
+	if err != nil {
+		b.log.With().Info("challenge not loaded", log.Err(err))
+	}
+	if challenge != nil && isStale(challenge, b.currentEpoch()) {
```
fasmat: imo `loadChallenge` should already check for staleness and return `nil` if the challenge is stale and needs to be rebuild.

activation/activation.go
poszu: Yes, it is OK (not ideal but technically still correct) to publish an ATX in the target epoch. Please see the [discussion](https://spacemesh.slack.com/archives/C89QJJY3G/p1672396924995409) (sorry for private link). Copying over the most important part:

> target epoch is when atx is actually used. there are multiple downsides if someone will do that
for smesher the downside is that it already missed some amount of rewards
for the network - average number of proposals/ballots will be higher than expected (though i am not fully confident if everything is correct here)
i think the codebase should optimize for not doing this. but if smesher finds itself in such condition it probably should still publish it, otherwise it won't get any rewards for the whole epoch

activation/activation.go
poszu: IMO, the log above gives a piece of important contextual information for debugging.

activation/activation.go
poszu: It would result in `"epoch_id"` instead of `"current epoch"`. I prefer the latter to differentiate between 3 epochs logged here.

poszu: What's the benefit? `EpochID` implements `log.LoggableField`. Using `log.FieldNamed` I get the same serialization as in other logs, except with a different _key_. By using `Stringer` it could have a different serialization.

activation/activation.go
poszu: I removed it completely

activation/activation.go
```diff
@@ -494,13 +488,21 @@ func (b *Builder) Coinbase() types.Address {
...
-	b.challenge = nipost
-	return nil
+	if nipost != nil && nipost.TargetEpoch() < b.currentEpoch() {
+		b.log.With().Info("atx nipost challenge is stale - discarding it",
+			log.FieldNamed("target_epoch", nipost.TargetEpoch()),
+			log.FieldNamed("publish_epoch", nipost.PublishEpoch()),
+			log.FieldNamed("current_epoch", b.currentEpoch()),
+		)
+		b.discardChallenge()
+		return nil, nil
```
fasmat: If we return an `error` here instead of `nil, nil` the code in `PublishActivationTx` could be simplified to:

```golang
	if err != nil {
		logger.With().Info("building new atx challenge", log.Stringer("current_epoch", b.currentEpoch()))
		if err := b.buildNIPostChallenge(ctx); err != nil {
		ch, err := b.buildNIPostChallenge(ctx)
		if err != nil {
			return fmt.Errorf("build new atx challenge: %w", err)
		}
		challenge = ch
	}

	logger.With().Info("atx challenge is ready",
		log.Stringer("current_epoch", b.currentEpoch()),
		log.Stringer("publish_epoch", b.challenge.PublishEpoch()),
		log.Stringer("target_epoch", b.challenge.TargetEpoch()))
		log.Stringer("publish_epoch", challenge.PublishEpoch()),
		log.Stringer("target_epoch", challenge.TargetEpoch()),
	)
```

specifically the `nil`-check for `challenge` becomes unnecessary.

poszu: Done. Thanks for the suggestion!

activation/activation.go
```diff
@@ -494,13 +488,21 @@ func (b *Builder) Coinbase() types.Address {
...
-func (b *Builder) loadChallenge() error {
+func (b *Builder) loadChallenge() (*types.NIPostChallenge, error) {
 	nipost, err := kvstore.GetNIPostChallenge(b.cdb)
 	if err != nil {
-		return err
+		return nil, fmt.Errorf("failed to load nipost challenge from DB: %w", err)
 	}
-	b.challenge = nipost
-	return nil
+	if nipost != nil && nipost.TargetEpoch() < b.currentEpoch() {
```
fasmat: NIT: if `err == nil` then `nipost != nil` will always be `true`.

activation/activation.go
countvonzero: the point is to get a consistent serialization. 
the Stringer interface is much more intuitive. the object just needs to implement a String() method.
at one point the plan is to replace all FiledNamed to Stringer but that's not a priority.

activation/activation.go
countvonzero: yeah, but at this point what's there to debug? the challenge is stale... anyhow, not important. just an attempt to reduce logged data.

countvonzero (approved):

activation/activation.go
poszu: I checked if this is really true. It is. Fun fact, `kvstore.AddNIPostBuilderState()` panics on nil pointer dereference if it gets `nil`.

activation/activation.go
poszu: But `Field()` and `String()` might return different textual representations. The former is dedicated to logging and IMO this one should be used.

activation/activation.go
fasmat: At least I would expect it to be true 🙂, but good that you double checked!

poszu:
bors merge

activation/activation.go
countvonzero: i.cannot think of good use of String() other than logging....

poszu:
bors merge

PR #3937, opened 2023-01-09
[Merged by Bors] - tortoise: count ballots when beacon is available

### LINKED ISSUE #3933: tortoise: failed to add ballot to tortoise state from database
> 2023-01-04T19:06:53.298+0100	ERROR	f015f.trtl         	failed to add ballot to the state	{"node_id": "f015fa5c6a27d2860b3a908226bcf689b931c9760d95a5812e585fe2d9af5e24", "module": "trtl", "errmsg": "beacon is not calculated for this epoch", "ballot_id": "3d5af6e9a0", "layer_id": 3900, "epoch_id": 65, "smesher": "1df4eb09484a182c2c6b666010743cb0b3c346af27518c79f2cf6985cc9f4eb4", "opinion hash": "0xa67e122555f8a4246198f7aa8ccfec8a3cd1786966cb66a0b267d948f306e198", "base_ballot": "0df842d534", "support": 1, "against": 0, "abstain": 0, "atx_id": "e29e3b691d", "eligibilities": [{"j": 139, "sig": "9059765ddacf2c7a5701935513d8082807ff46e6c032dc481c789128d8eac95ae967d0b92faba17ed169ed793bba38a00cc6cf6414d314173f85f3ff7548bf81c1eb3fba601c91714cfbef1b29e01c05"}, {"j": 175, "sig": "48d34c39cd5817030ffcd9f7df40f143c88743eaf6338e4f4ecbf43fb7c8bc85e0a573b2b494c4f68be71c9e6fdfb6d04a75c34fb64a3d76277916a0e140bb5feeff61babc0a9e1ef8ee0f520fa5940b"}], "ref_ballot": "0000000000", "active_set_size": 13, "beacon": "8dce7ed8", "votes": {"base": "0df842d534", "support": [{"id": "a4ae9d98b8", "layer": 3899, "height": 460340446}], "against": [], "abstain": []}, "name": "trtl"}

ballot is saved into the state while we don't have a beacon as expected. when state is recovered from disk tortoise tries to count ballot, and fails as beacon is not available and tortoise doesn't know when it should count the ballot.

dshulyak:
this change makes it so that ballot is never dropped if beacon is not yet available.
i want to have something more straighforward, but i will do that after thinking more in https://github.com/spacemeshos/go-spacemesh/pull/3888

closes: https://github.com/spacemeshos/go-spacemesh/issues/3933

### DIFF
--- a/tortoise/tortoise.go
+++ b/tortoise/tortoise.go
@@ -1,6 +1,7 @@
 package tortoise
 
 import (
+	"container/list"
 	"context"
 	"errors"
 	"fmt"
@@ -22,6 +23,8 @@ import (
 	"github.com/spacemeshos/go-spacemesh/tortoise/metrics"
 )
 
+var errBeaconUnavailable = errors.New("beacon unavailable")
+
 type turtle struct {
 	Config
 	logger log.Log
@@ -31,6 +34,12 @@ type turtle struct {
 	updated []types.BlockContextualValidity
 
 	*state
+
+	// a linked list with retriable ballots
+	// the purpose is to add ballot to the state even
+	// if beacon is not available locally, as tortoise
+	// can't count ballots without knowing local beacon
+	retriable list.List
 
 	verifying *verifying
 
@@ -129,6 +138,9 @@ func (t *turtle) evict(ctx context.Context) {
 
 // EncodeVotes by choosing base ballot and explicit votes.
 func (t *turtle) EncodeVotes(ctx context.Context, conf *encodeConf) (*types.Opinion, error) {
+	if err := t.checkDrained(); err != nil {
+		return nil, err
+	}
 	var (
 		logger = t.logger.WithContext(ctx)
 		err    error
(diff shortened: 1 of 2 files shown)


tortoise/tortoise.go
```diff
@@ -129,6 +138,9 @@ func (t *turtle) evict(ctx context.Context) {
 
 // EncodeVotes by choosing base ballot and explicit votes.
 func (t *turtle) EncodeVotes(ctx context.Context, conf *encodeConf) (*types.Opinion, error) {
+	if err := t.checkDrained(); err != nil {
```
countvonzero, line 141: for me this interface is more straightforward to either
- `checkDrained` return boolean indicating drained or not, and return the formatted error here
- or just drain the the list right here. as if we are creating ballot, the beacon is available.

countvonzero (approved):

tortoise/tortoise.go
dshulyak: i wrote it this way as it seems easier to remove in future, and i also want to call tally votes when encode votes is called, but current code doesn't do it

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

Issue #3919, opened 2022-12-29
malfeasance: generate proof for malfeasance behavior


countvonzero:
## Description

2nd part of https://github.com/spacemeshos/go-spacemesh/issues/3915

generate `MalfeasanceProof` for the following malicious behavior:
- miners that publish multiple ATXs targeting the same epoch
- miners that publish multiple ballots for the same layer
- miners that equivocates in the hare protocol

once generated, also gossip `MalfeasanceProof` with an optional `HareEligibility`

PR #3942, opened 2023-01-10
[Merged by Bors] - generate MalfeasanceProof for ATXs and Ballots

countvonzero:
## Motivation

1 of 2 PRs for #3919

## Changes

- create 3 types of MalfeasanceProof: ATX, ballots, and hare equivocation
- MalfeasanceProof gossip handler
- generate ATX MalfeasanceProof when receiving multiple ATX for the same epoch
- generate Ballot MalfeasanceProof when receiving multiple ballots for the same epoch

note: malicious ATX/proposals/ballots are all saved for now, pending research conclusions. only identities are marked as malicious.


common/types/malfeasance.go
```diff
@@ -0,0 +1,201 @@
...
+
+type MalfeasanceProof struct {
+	// for network upgrade
+	Layer LayerID
+	Type  MalfeasanceType
+	// conflicting messages signed by the same NodeID
+	Messages []any
+}
+
+func (t *MalfeasanceProof) EncodeScale(enc *scale.Encoder) (int, error) {
```
countvonzero: @dshulyak please advise if improvements can be made in the custom encode/decode. thank you

countvonzero:
bors try

common/types/malfeasance.go
dshulyak: i would use a following approach:

1. define different malfeasence proofs. all of them can be autogenerated

```go
type AtxProof struct {
    [2]ConcreteAtxMessage
}

type HareProof struct {
   [2]ConcreteHareMessage
}
```

2. define enum for them. this is the only object that needs custom encoding, as the library doesn't support autogenerating it and it is not straighforward, but in general should be possible to autogenerate it with some work

```go

type ProofEnum struct {
    Type uint8
    Proof scale.Type
}

func (e *ProofEnum) EncodeScale(e *scale.Encode) {
    // not compact, as scale spec uses "full" uint8 for enums
    scale.EncodeByte(e, byte(e.Type))
   e.Proof.EncodeScale(e)
}

func (e *ProofEnum) DecodeScale(d *scale.Decoder) {
   typ := scale.DecodeByte(d)
   switch typ {
      // handle concrete type
   }
}
```  

3. higher level malfeasence proof for versioning. autogenerated

```go
type MalfeasenceProof struct {
    Layer types.LayerID
    Enum ProofEnum
}

```

countvonzero:
bors try

common/types/malfeasance.go
countvonzero: @dshulyak done as suggested with some name changes. thanks for helping.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #3947, opened 2023-01-11
[Merged by Bors] - Fix GKE Auth Plugin

andres-spacemesh:
## Changes
Fix the CI Job when authenticating on GKE with the Auth Plugin

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)

PR #3948, opened 2023-01-11
partition test: check state up to one epoch before test stops

countvonzero:
## Motivation

an attempt to stablise partition systest.

theoretically in the cloud the nodes should continue to agree on the state after healing from partition.
but it seems the cloud environment is not optimal right now and is slowing down dev process

## Changes

check states one epoch before the `stop` layer


fasmat (approved):
lgtm

countvonzero:
merged as part of https://github.com/spacemeshos/go-spacemesh/pull/3947

Issue #3936, opened 2023-01-07
Investigate the issue with blocks with invalid rewards


maparr:
## Description
We have an issue with our user. As asked @countvonzero , need to add additional logs, we need to add additional logs why we have such an error `invalid rewards: reward with invalid (zeroed) weight (0/1)`, need to add more logs to debug this.

## Steps to reproduce
1. Use the application
2. start to sync the node
3. Start to smeshing

## Actual Behavior
we are getting an error, more [logs](https://sentry.io/organizations/spacemesh/issues/3853995497/events/c118d9e0865447d5b862608d13b6ff59/?project=6324919#attachments) from our client.
`2023-01-07T01:28:00.087+0200	WARN	cddf1.fetcher      	failed to get hash	{"node_id": "cddf1eb79b3ef142d82c82faa01ba3126c1c5859d30ea6b3f7f76cdfb05399a0", "module": "fetcher", "sessionId": "8b151420-7b78-4302-b7a1-21a08ab979d3", "hint": "ballotDB", "hash": "0x24362d182a4b9961871bd32152115a61bfdb1d7e000000000000000000000000", "errmsg": "fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch ballots: fetch blocks: invalid rewards: reward with invalid (zeroed) weight (0/1) included into the block for [0 0 0 0 120 71 32 140 122 134 128 32 172 68 134 187 127 63 39 218 10 207 148 36]", "name": "fetcher"}`

## Expected Behavior
- the smashing process is starting, node is not stuck 

## Environment
Please complete the following information:
 - OS:  Windows
 - Node Version: 0.20
 - Smapp version (if applicable):  0.2.10

countvonzero:
@maparr can you clarify the OS this bug occurred to? did you also see this on Linux and MacOS? i had thought the original bug report was from windows.

dshulyak:
- [x] validate that ballot atx is included into the active set. otherwise ballot is syntactically invalid and should not be accepted https://github.com/spacemeshos/go-spacemesh/pull/3943
[...]

PR #3949, opened 2023-01-11
ATX builder should discard challenge if target epoch is current epoch

fasmat:
## Motivation
See also https://github.com/spacemeshos/go-spacemesh/issues/3936#issuecomment-1377219636

Prevents possible double publication of an ATX in the same epoch by the same smesher

## Changes
Instead of discarding the challenge when the target epoch is before the current epoch it should discard it if is before or the same as the current epoch.

## Test Plan
Existing tests pass, new tests added

## TODO

## DevOps Notes

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -302,6 +302,8 @@ func (b *Builder) generateProof(ctx context.Context) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


fasmat:
Closed in favor of https://github.com/spacemeshos/go-spacemesh/pull/3924

PR #3950, opened 2023-01-12
[Merged by Bors] - hare: generate MalfeasanceProof upon equivocation

### LINKED ISSUE #3919: malfeasance: generate proof for malfeasance behavior

countvonzero:
## Motivation

Closes #3919

## Changes
- for each round, detect equivocation and generate MalfeasanceProof
- added a goroutine in hare service to process MalfeasanceGossip from each round tracker
- a semi-integrated tests for pre-round equivocation


common/types/activation.go
```diff
@@ -291,6 +292,11 @@ func (atx *ActivationTx) CalcAndSetID() error {
 	if atx.Signature == nil {
 		return fmt.Errorf("cannot calculate ATX ID: sig is nil")
 	}
+
+	if atx.MsgHash != BytesToHash(atx.HashInnerBytes()) {
+		return fmt.Errorf("bad message hash")
+	}
```
fasmat: This would require the caller of `CalcAndSetID` to always call `SetMetadata` first. Would it make sense to do this instead of checking that the hash matches?

activation/handler.go
```diff
@@ -280,12 +287,48 @@ func (h *Handler) StoreAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 	h.Lock()
 	defer h.Unlock()
 
-	if err := atxs.Add(h.cdb, atx, time.Now()); err != nil {
-		if errors.Is(err, sql.ErrObjectExists) {
-			return nil
+	malicious, err := identities.IsMalicious(h.cdb, atx.NodeID())
+	if err != nil {
+		return fmt.Errorf("store atx: %w", err)
+	}
```
fasmat: would it make sense to also include this in the transaction below?

common/types/ballot.go
```diff
@@ -206,6 +206,10 @@ func (b *Ballot) Initialize() error {
 		return fmt.Errorf("cannot calculate Ballot ID: signature is nil")
 	}
 
+	if b.MsgHash != BytesToHash(b.HashInnerBytes()) {
+		return fmt.Errorf("bad message hash")
+	}
```
fasmat: Similar to ATXs: would it make sense to just set the hash here by calling `SetMetadata` in `Initialize`?

common/types/malfeasance.go
```diff
@@ -0,0 +1,159 @@
...
+	Layer     LayerID
+	ProofData TypedProof
+}
+
+type TypedProof struct {
+	// MultipleATXs | MultipleBallots | HareEquivocation
+	Type uint8
+	// AtxProof | BallotProof | HareProof
+	Proof scale.Type
+}
```
fasmat: Just a suggestion for a different naming and using a custom type for the enum:

```suggestion
type ProofType byte

const (
	MultipleATXs ProofType = iota + 1
	MultipleBallots
	HareEquivocation
)

type MalfeasanceProof struct {
	// for network upgrade
	Layer LayerID
	Proof Proof
}

type Proof struct {
	// MultipleATXs | MultipleBallots | HareEquivocation
	Type ProofType
	// AtxProof | BallotProof | HareProof
	Data scale.Type
}
```

sql/atxs/atxs.go
```diff
@@ -141,6 +142,51 @@ func GetIDByEpochAndNodeID(db sql.Executor, epoch types.EpochID, nodeID types.No
 	return id, err
 }
 
+// GetByEpochAndNodeID gets any ATX by the specified NodeID in the given epoch.
+func GetByEpochAndNodeID(db sql.Executor, epoch types.EpochID, nodeID types.NodeID) (*types.VerifiedActivationTx, error) {
```
fasmat: Do we need this function when we can also do:

```golang
atxid, _ := GetIDByEpochAndNodeID(db, epoch, nodeId)
atx, _ := Get(db, id)
```

Is there a reason to have a dedicated method that does both in one function call?

countvonzero:
bors try

dshulyak (approved):

hare/algorithm.go
```diff
@@ -658,7 +669,10 @@ func (proc *consensusProcess) beginNotifyRound(ctx context.Context) {
 // passes all pending messages to the inbox of the process so they will be handled.
 func (proc *consensusProcess) handlePending(pending map[string]*Msg) {
 	for _, m := range pending {
-		proc.inbox <- m
+		select {
+		case <-proc.ctx.Done():
```
fasmat, line 673: I assume when the context is `Done` the for loop should break here?

fasmat (approved):
Just a few minor things, besides that lgtm

common/types/activation.go
countvonzero: no it doesn't. 

SetMetadata is only called when ATX needs to be signed. i.e. it should only be called in the ATX building flow (see `SignAndFinalizeAtx`).

for gossiped/synced ATXs, Metadata should be already set correctly. otherwise it is syntactically invalid. hence the check here. Metadata is the actual data that gets signed.

activation/handler.go
countvonzero: it's not critical, as `SetMalicious()` will not overwrite the previous value. in general i want to minimize code inside a database transaction.

common/types/malfeasance.go
countvonzero: i did have a custom type, but in the code i have to constantly cast it back and forth.

sql/atxs/atxs.go
countvonzero: but what you suggested takes 2 db read to accomplished. this new method just needs 1 db read. that's a worthwhile efficiency IMO.

hare/algorithm.go
countvonzero: you assume correctly. thanks

common/types/malfeasance.go
countvonzero: renamed as suggested, with the exception of the custom type

common/types/ballot.go
countvonzero: same as https://github.com/spacemeshos/go-spacemesh/pull/3950#discussion_r1069566816

sql/atxs/atxs.go
fasmat: It could still be done in a single transaction with

```golang
db.WithTx(ctx, func(tx *sql.Tx) error {
		atxid, _ := atxs.GetIDByEpochAndNodeID(tx, epoch, nodeId)
		atx, _ := atxs.Get(tx, atxid)
}
```

or would this be less efficient?

countvonzero: it's disk reads at the db level. disregarding disk reads for database index, isn't it 2 disk read vs 1 disk read in terms of the actual data?

countvonzero:
bors merge

PR #3953, opened 2023-01-13
[Merged by Bors] - hare: applying f plus one minus k rule

### LINKED ISSUE #3921: hare: handle hare equivocation with f+1-k rule
## Description

4th and final part of https://github.com/spacemeshos/go-spacemesh/issues/3915

### hare
below are the places in hare where f+1 rule is applied

- preround: determine whether each value has at least f+1 support.
- status round: determine whether there are at least f+1 status messages to build a safe value proof (SVP)
- commit round: determine whether there are at least f+1 commit messages to to build a commit certificate
- notify round: validate the commit messages in the received commit certificate
- notify round: determine whether there are at least f+1 notify messages to terminate the hare instance
the new f+1-k rule will be used wherever f+1 rule is applied currently.

note: the k malicious identity must also be eligible for the given layer and round. this is why the `HareEligibility` needs to be gossiped along with the `MalfeasanceProof` in live hare rounds, hence `MalfeasanceGossip`. the `HareEligibility` messages also need to be kept for the duration of a hare instance, as the node will need to verify earlier round messages when validating the commit messages in the commit certificate.

### block certificate
- generating: f+1 valid `CertifyMessage`
wait for f+1 valid `CertifyMessage` before generating a block certificate. all `CeritfyMessage` need to be
from eligible and non-malicious identity.
- validating a synced certificate: f+1-k valid `CertifyMessage`
accept a certificate if there are f+1 valid `CeritfyMessage` from eligible identities, of which the maximum
number of malicious identities is k.

countvonzero:
## Motivation

Closes #3921

### behavior changes 
- malfeasance gossip handler:
if any incoming malfeasance proof contains hare eligibility, verify and relay the eligibility to hare
- hare gossip handler (aka broker.go)
when receiving a hare message, check if the NodeID is known malicious. if so
  - do not relay such gossip
  - replace hare message with its malfeasance proof + hare eligibility and broadcast it to the network
  - pass on the eligibility to its running hare consensus process
- consensus process (algorithm.go)
  - track eligibility for all rounds in this instance/layer. this is needed as later rounds will be verifying messages from earlier rounds in the case of svp and certificate.
  - replace f+1 rule with f+1-k rules whenever applied  
  
### f+1-k rule
since the network will replace a malicious hare message with MalfeasanceProof+HareEligibility for a known equivocator (KE), most nodes will not receive HareMessage from the KEs. so the criteria for the node to apply f+1-k rule is
- x honest message, x >= 1
- k known equivocators without messages
- consensus on the round that previously required f + 1 rule now requires x + k >= f + 1

### hare information flow
each hare consensus process can potentially discover malicious identities in 3 ways
1. the consensus process detects equivocations: 
    - output MalfeasanceGossip to a channel that's processed in the hare background loop
2. broker checking malicious identity from local MalfeasanceProof store when processing hare gossip message: 
    - output MalfeasanceGossip to the same channel as in 1
    - relay HareEligibility to the consensus process's inbox channel
3. from MalfeasanceGossip received from the network
    - relay HareEligibility to the consensus process's inbox channel


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

malfeasance/handler.go
```diff
@@ -92,6 +114,38 @@ func updateMetrics(tp types.Proof) {
...
+		logger.Fatal("invalid input")
+	}
+
+	emsg := gossip.Eligibility
+	eNodeID := types.BytesToNodeID(emsg.PubKey)
+	if nodeID != eNodeID {
+		logger.With().Warning("mismatch node id", log.Stringer("eligibility", eNodeID))
+		return fmt.Errorf("mismatch node id")
+	}
+	eligible, err := h.validator.Validate(ctx, emsg.Layer, emsg.Round, h.committeeSize, nodeID, emsg.Eligibility.Proof, emsg.Eligibility.Count)
```
dshulyak: i just noticed that this is quite expensive and i don't understand why it needs to verify eligibility

it will have to load atxs in the past, do you know if it caches them? if so, when does it drop them from cache?

i thought that it is sufficient for this method to check if two vrf signatures are from the same identity and it can be done without loading lots of data

dshulyak: i am quite confident that checking full eligibility is a bug

imagine that hare active set depends on the tortoise. lets say online nodes are using active set X and generate proofs that will pass this validation. later tortoise reaches different consensus (e.g self-healing) and active set is now Y, now nodes that received proof won't be able to verify eligibility. as a consequence identity will be malicious on some nodes, and not malicious on other nodes

countvonzero: thank you. hare oracle currently caches the active set for 5 epochs.

the intention here is for live hare rounds. when malicious identity sends a hare message, the rule is for a node to replace the hare message with malfeasance proof, but attach the hare eligibility so the network will still complete the hare instance with the f+1-k rule. 

as such, there should be the same amount of validation before and after this PR. the honest hare eligibilities will be validated in hare/broker.go gossip handler, the dishonest/know equivocator's hare eligibilities will be validated here.

but you are correct that this is a bug, as i didn't check for whether the hare instance for the gossiped eligibility is running.

what do you think that i just unconditionally passed this to the hare, and let hare broker decide whether it should check for full eligibility? @dshulyak

dshulyak: i realized that i misunderstood what is going on, but sending msg to hare over channel in the initializer is indeed looks questionable. i would try to avoid such code

dshulyak (approved):

malfeasance/handler.go
countvonzero: ok. thanks. i will pass the hare's gossip handler to the malfeasance handler instead.

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

PR #3957, opened 2023-01-13
[Merged by Bors] - hare: more simplification

countvonzero:
## Motivation

part of #3921 

## Changes

- simplify broker locking scheme
broker has complexity in both actor-pattern (where locking should be none or minimal) and mutex world. 
choose the mutex, as the broker handles incoming gossip messages and concurrency is important.
this allows for the removal of event loop and goroutine management 
- simplify consensus process (algorithm.go) locking scheme
- use sync.Once for Start() on broker and consensus 
- simplify hare set implementation

### DIFF
--- a/hare/algorithm_test.go
+++ b/hare/algorithm_test.go
@@ -22,8 +22,6 @@ import (
 	smocks "github.com/spacemeshos/go-spacemesh/system/mocks"
 )
 
-var cfg = config.Config{N: 10, F: 5, RoundDuration: 2, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 1000}
-
 func newRoundClockFromCfg(logger log.Log, cfg config.Config) *SimpleRoundClock {
 	return NewSimpleRoundClock(time.Now(),
 		time.Duration(cfg.WakeupDelta)*time.Second,
@@ -150,6 +148,7 @@ type testBroker struct {
 }
 
 func buildBroker(tb testing.TB, testName string) *testBroker {
+	cfg := config.Config{N: 10, F: 5, RoundDuration: 2, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 1000, Hdist: 20}
 	return buildBrokerWithLimit(tb, testName, cfg.LimitIterations)
 }
 
@@ -197,7 +196,7 @@ func TestConsensusProcess_Start(t *testing.T) {
 }
 
 func TestConsensusProcess_TerminationLimit(t *testing.T) {
-	c := cfg
+	c := config.Config{N: 10, F: 5, RoundDuration: 2, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 1000, Hdist: 20}
 	c.LimitConcurrent = 1
 	c.RoundDuration = 1
 	p := generateConsensusProcessWithConfig(t, c)
@@ -218,7 +217,7 @@ func TestConsensusProcess_eventLoop(t *testing.T) {
 	broker.mockSyncS.EXPECT().IsSynced(gomock.Any()).Return(true).AnyTimes()
 	broker.mockSyncS.EXPECT().IsBeaconSynced(gomock.Any()).Return(true).AnyTimes()
 	require.NoError(t, broker.Start(context.Background()))
-	c := cfg
+	c := config.Config{N: 10, F: 5, RoundDuration: 2, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 1000, Hdist: 20}
 	c.F = 2
 	proc := generateConsensusProcessWithConfig(t, c)
 	proc.publisher = net
@@ -305,6 +304,7 @@ func TestConsensusProcess_nextRound(t *testing.T) {
 }
 
 func generateConsensusProcess(t *testing.T) *consensusProcess {
+	cfg := config.Config{N: 10, F: 5, RoundDuration: 2, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 1000, Hdist: 20}
 	return generateConsensusProcessWithConfig(t, cfg)
 }
 
@@ -524,6 +524,7 @@ func TestConsensusProcess_Termination(t *testing.T) {
 	proc.advanceToNextRound(context.Background())
 	s := NewSetFromValues(value1)
 
+	cfg := config.Config{N: 10, F: 5, RoundDuration: 2, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 1000, Hdist: 20}
 	for i := 0; i < cfg.F+1; i++ {
 		signer, err := signing.NewEdSigner()
 		require.NoError(t, err)
--- a/hare/algorithm_test.go
+++ b/hare/algorithm_test.go
@@ -182,7 +182,7 @@ func TestConsensusProcess_Start(t *testing.T) {
 	broker.mockSyncS.EXPECT().IsSynced(gomock.Any()).Return(true).AnyTimes()
 	broker.mockSyncS.EXPECT().IsBeaconSynced(gomock.Any()).Return(true).AnyTimes()
 	broker.Start(context.Background())
-	defer broker.Close()
+	t.Cleanup(broker.Close)
 	proc := generateConsensusProcess(t)
 	inbox, err := broker.Register(context.Background(), proc.ID())
 	require.NoError(t, err)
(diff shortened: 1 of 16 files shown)


countvonzero:
bors try

hare/broker_test.go
```diff
@@ -50,35 +50,21 @@ func createMessage(tb testing.TB, instanceID types.LayerID) []byte {
...
-
 // test that a InnerMsg to a specific set ID is delivered by the broker.
 func TestBroker_Received(t *testing.T) {
 	broker := buildBroker(t, t.Name())
 	broker.mockSyncS.EXPECT().IsSynced(gomock.Any()).Return(true).AnyTimes()
 	broker.mockSyncS.EXPECT().IsBeaconSynced(gomock.Any()).Return(true).AnyTimes()
 	broker.mockStateQ.EXPECT().IsIdentityActiveOnConsensusView(gomock.Any(), gomock.Any(), gomock.Any()).Return(true, nil).AnyTimes()
-	require.NoError(t, broker.Start(context.Background()))
+	broker.Start(context.Background())
+	defer broker.Close()
```
fasmat:
```suggestion
	t.Cleanup(broker.Close())
```

NIT: `defer` isn't executed when a panic happens, `t.Cleanup()` is.

hare/algorithm_test.go
```diff
@@ -217,8 +219,9 @@ func TestConsensusProcess_eventLoop(t *testing.T) {
 	broker := buildBroker(t, t.Name())
 	broker.mockSyncS.EXPECT().IsSynced(gomock.Any()).Return(true).AnyTimes()
 	broker.mockSyncS.EXPECT().IsBeaconSynced(gomock.Any()).Return(true).AnyTimes()
-	require.NoError(t, broker.Start(context.Background()))
-	c := cfg
+	broker.Start(context.Background())
+	defer broker.Close()
```
fasmat:
```suggestion
	t.Cleanup(broker.Close())
```

hare/consensus_test.go
```diff
@@ -112,7 +112,7 @@ func (his *HareSuite) checkResult(t *testing.T) {
 	}
 
 	// check that the output has no intersection with the complement of the union of honest
-	for _, v := range his.outputs[0].elements() {
+	for _, v := range his.outputs[0].ToSlice() {
```
fasmat, line 115: This iterates twice! Once over all elements of `outputs` to get the values and once again over the values.

hare/hare_test.go
```diff
@@ -536,7 +537,8 @@ func TestHare_onTick_NoBeacon(t *testing.T) {
 	mockBeacons := smocks.NewMockBeaconGetter(gomock.NewController(t))
 	h.beacons = mockBeacons
 	mockBeacons.EXPECT().GetBeacon(lyr.GetEpoch()).Return(types.EmptyBeacon, errors.New("whatever")).Times(1)
-	require.NoError(t, h.broker.Start(context.Background()))
+	h.broker.Start(context.Background())
+	defer h.broker.Close()
```
fasmat, line 541:
```suggestion
	t.Cleanup(h.broker.Close())
```

hare/haretypes.go
```diff
@@ -212,38 +186,43 @@ func (s *Set) updateID() {
...
-// ID returns the ObjectID of the set.
-func (s *Set) ID() types.Hash32 {
-	if !s.isIDValid {
-		s.updateID()
+func (s *Set) sortedLocked() []types.ProposalID {
+	result := make([]types.ProposalID, 0, len(s.values))
+
+	for id := range s.values {
+		result = append(result, id)
 	}
```
fasmat:
```suggestion
// import "golang.org/x/exp/maps"

	result := maps.Keys(s)
```

hare/messagevalidation.go
```diff
@@ -450,8 +450,8 @@ func (v *syntaxContextValidator) validateSVPTypeA(ctx context.Context, m *Msg) b
 	for _, status := range m.InnerMsg.Svp.Messages {
 		statusSet := NewSet(status.InnerMsg.Values)
 		// build union
-		for _, bid := range statusSet.elements() {
-			unionSet.Add(bid) // assuming add is unique
+		for _, pid := range statusSet.ToSlice() {
```
fasmat, line 453: This also iterates twice, once to create the slice and then again for the for loop here.

hare/preroundtracker.go
```diff
@@ -92,7 +92,7 @@ func (pre *preRoundTracker) OnPreRound(ctx context.Context, msg *Msg) {
 	}
 
 	// record Values
-	for _, v := range sToTrack.elements() {
+	for _, v := range sToTrack.ToSlice() {
```
fasmat, line 95: Double iteration

hare/preroundtracker.go
```diff
@@ -120,7 +120,7 @@ func (pre *preRoundTracker) CanProveValue(value types.ProposalID) bool {
 // a set is said to be provable if all his values are provable.
 func (pre *preRoundTracker) CanProveSet(set *Set) bool {
 	// a set is provable iff all its Values are provable
-	for _, bid := range set.elements() {
+	for _, bid := range set.ToSlice() {
```
fasmat, line 123: Double iteration

hare/preroundtracker.go
```diff
@@ -131,9 +131,9 @@ func (pre *preRoundTracker) CanProveSet(set *Set) bool {
 
 // FilterSet filters out non-provable values from the given set.
 func (pre *preRoundTracker) FilterSet(set *Set) {
-	for _, bid := range set.elements() {
-		if !pre.CanProveValue(bid) { // not enough witnesses
-			set.Remove(bid)
+	for _, v := range set.ToSlice() {
```
fasmat, line 134: Double iteration

hare/statustracker.go
```diff
@@ -100,8 +100,8 @@ func (st *statusTracker) ProposalSet(expectedSize int) *Set {
 func (st *statusTracker) buildUnionSet(expectedSize int) *Set {
 	unionSet := NewEmptySet(expectedSize)
 	for _, m := range st.statuses {
-		for _, bid := range NewSet(m.InnerMsg.Values).elements() {
-			unionSet.Add(bid) // assuming add is unique
+		for _, v := range NewSet(m.InnerMsg.Values).ToSlice() {
```
fasmat, line 103: Double iteration

fasmat (changes requested):
To avoid the double iteration that happens with the `ToSlice` method I would probably do something like:

```golang
(s *Set) ForEachElement(apply func(types.ProposalID)) {

    for k := range s.values {
        apply(k)
    }
}
```

or similar to avoid one of the two for loops.

hare/preroundtracker.go
```diff
@@ -131,9 +131,9 @@ func (pre *preRoundTracker) CanProveSet(set *Set) bool {
...
+		if !pre.CanProveValue(v) { // not enough witnesses
+			set.Remove(v)
 		}
```
fasmat, line 137: Double iteration, with the `ForEachElement` closure this would then be:

```suggestion
	set.ForEachElement(func(v types.ProposalID) {
		if !pre.CanProveValue(v) { // not enough witnesses
			set.Remove(v)
		}
	}
```

countvonzero:
> To avoid the double iteration that happens with the `ToSlice` method I would probably do something like:
> 
> ```go
> (s *Set) ForEachElement(apply func(types.ProposalID)) {
> 
>     for k := range s.values {
>         apply(k)
>     }
> }
> ```
> 
> or similar to avoid one of the two for loops.

so the order of the set size is max 50. the math in proposal eligibility ensures that.
your suggestion is definite generic and more efficient, unfortunately in this case i'll need to keep the code as is, because there can be mutation (and mutex locking) in the `apply` function. for example, removing an element from its set while iterating.

making your suggestion work will take more labor. so i'm going to punt on that.

hare/broker_test.go
countvonzero: done.

hare/algorithm_test.go
countvonzero: done

hare/haretypes.go
countvonzero: done.

fasmat:
> because there can be mutation (and mutex locking) in the `apply` function. for example, removing an element from its set while iterating.

Removing elements from a map while iterating over them is safe to do with go maps: https://go.dev/doc/effective_go#for at least I haven't found any source that would say it's not safe to do (assuming it happens on the same goroutine).

Regarding using locks - the `ForEachElement` method would need to lock before iterating over the map and executing the `apply` function just in case. It could be that this is less efficient than the current code and without benchmarks it might even be hard to tell which is better - so I'm OK with leaving it as it is.

fasmat (approved):

countvonzero:
> > because there can be mutation (and mutex locking) in the `apply` function. for example, removing an element from its set while iterating.
> 
> Removing elements from a map while iterating over them is safe to do with go maps: https://go.dev/doc/effective_go#for at least I haven't found any source that would say it's not safe to do (assuming it happens on the same goroutine).
> 
> Regarding using locks - the `ForEachElement` method would need to lock before iterating over the map and executing the apply function just in case. It could be that this is less efficient than the current code and without benchmarks it might even be hard to tell which is better.

yes. it's safe in general for golang. 
but in safe in the current hare set implementation. in `FilterSet()` it will need to call `s.Remove()` (which takes the set lock too.)

UPDATE: i mean "not safe in current hare set implementaion"

countvonzero:
bors merge

PR #3960, opened 2023-01-14
[Merged by Bors] - use cache for identity check to reduce db reads

countvonzero:
## Motivation

part of  #3921

## Changes

use CacheDB to cache both ATXs and malicious proof.

for each atx, proposal, and hare gossip message, we now query if an identity is known malicious.
atx and proposals are less time sensitive than hare. 

for each hare consensus instance, a different committee of N (800) is selected for each round (1preround+4 rounds per iteration).

### DIFF
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -18,7 +18,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/p2p/pubsub"
 	"github.com/spacemeshos/go-spacemesh/sql"
 	"github.com/spacemeshos/go-spacemesh/sql/atxs"
-	"github.com/spacemeshos/go-spacemesh/sql/identities"
 	"github.com/spacemeshos/go-spacemesh/system"
 )
 
@@ -296,14 +295,14 @@ func (h *Handler) StoreAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 	h.Lock()
 	defer h.Unlock()
 
-	malicious, err := identities.IsMalicious(h.cdb, atx.NodeID())
+	malicious, err := h.cdb.IsMalicious(atx.NodeID())
 	if err != nil {
 		return fmt.Errorf("store atx: %w", err)
 	}
 	var proof *types.MalfeasanceProof
-	h.cdb.WithTx(ctx, func(tx *sql.Tx) error {
+	h.cdb.WithTx(ctx, func(dbtx *sql.Tx) error {
 		if !malicious {
-			prev, err := atxs.GetByEpochAndNodeID(tx, atx.PublishEpoch(), atx.NodeID())
+			prev, err := atxs.GetByEpochAndNodeID(dbtx, atx.PublishEpoch(), atx.NodeID())
 			if err != nil && !errors.Is(err, sql.ErrNotFound) {
 				return err
 			}
@@ -322,9 +321,7 @@ func (h *Handler) StoreAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 						Data: &atxProof,
 					},
 				}
-				if proofBytes, err := codec.Encode(proof); err != nil {
-					return err
-				} else if err = identities.SetMalicious(tx, atx.NodeID(), proofBytes); err != nil {
+				if err = h.cdb.AddMalfeasanceProof(atx.NodeID(), proof, dbtx); err != nil {
 					return err
 				}
 				h.log.With().Warning("smesher produced more than one atx in the same epoch",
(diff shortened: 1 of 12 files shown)


countvonzero:
bors try

countvonzero:
bors try

activation/handler.go
```diff
@@ -296,14 +295,14 @@ func (h *Handler) StoreAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 	h.Lock()
 	defer h.Unlock()
 
-	malicious, err := identities.IsMalicious(h.cdb, atx.NodeID())
+	malicious, err := h.cdb.IsMalicious(atx.NodeID())
 	if err != nil {
 		return fmt.Errorf("store atx: %w", err)
```
poszu, line 300: IMO this error would be more helpful as:
```suggestion
		return fmt.Errorf("checking if node is malicious: %w", err)
```

activation/handler.go
```diff
@@ -322,9 +321,7 @@ func (h *Handler) StoreAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 						Data: &atxProof,
 					},
 				}
-				if proofBytes, err := codec.Encode(proof); err != nil {
-					return err
-				} else if err = identities.SetMalicious(tx, atx.NodeID(), proofBytes); err != nil {
+				if err = h.cdb.AddMalfeasanceProof(atx.NodeID(), proof, dbtx); err != nil {
 					return err
```
poszu, line 325: Please, consider adding some context so we know what operation failed:
```suggestion
					return fmt.Errorf("adding malfeasense proof: %w", err)
```

cmd/node/node.go
```diff
@@ -459,48 +458,29 @@ func (app *App) SetLogLevel(name, loglevel string) error {
...
-	layerSize uint32,
 	poetClients []activation.PoetProvingServiceClient,
 	vrfSigner *signing.VRFSigner,
-	layersPerEpoch uint32, clock TickProvider,
+	clock TickProvider,
 ) error {
-	app.nodeID = nodeID
-
+	nodeID := sgn.NodeID()
+	layerSize := uint32(app.Config.LayerAvgSize)
```
poszu, line 469: How about changing the definition in the Config structure to uint32? Here: https://github.com/spacemeshos/go-spacemesh/blob/baf7df04c4a9187fde46490ff226573ea3844190/config/config.go#L87

It doesn't make sense to accept a signed integer in config in the first place.

fasmat (approved):
LGTM

datastore/cache.go
```diff
@@ -7,34 +7,40 @@ import (
...
+	return NewCache[types.ATXID, types.ActivationTxHeader](size)
 }
 
-// NewAtxCache creates a new cache for activation transaction headers.
-func NewAtxCache(size int) AtxCache {
-	cache, err := lru.New(size)
+// NewMalfeasanceCache creates a cache of MalfeasanceProof of recent malicious identities.
+func NewMalfeasanceCache(size int) *Cache[types.NodeID, types.MalfeasanceProof] {
+	return NewCache[types.NodeID, types.MalfeasanceProof](size)
+}
```
poszu, line 18: I'd avoid getters for specialized Cache and keep only the generic `Cache` implementation here.

datastore/store.go
```diff
@@ -11,29 +12,114 @@ import (
...
+
+// IsMalicious returns true if the NodeID is malicious.
+func (db *CachedDB) IsMalicious(id types.NodeID) (bool, error) {
+	if id == types.EmptyNodeID {
+		return false, errors.New("cache db: get empty node id")
+	}
+
+	db.mu.Lock()
+	defer db.mu.Unlock()
+	if proof, ok := db.malfeasanceCache.Get(id); ok {
```
poszu, line 59: NIT: Using a `lru::Contains` would be much faster if you don't care about moving `id` to the front of LRU's eviction list.

countvonzero:
bors merge

datastore/store.go
```diff
@@ -11,29 +12,114 @@ import (
...
+
+func (db *CachedDB) MalfeasanceCacheSize() int {
+	return db.malfeasanceCache.lru.Len()
+}
+
+// IsMalicious returns true if the NodeID is malicious.
+func (db *CachedDB) IsMalicious(id types.NodeID) (bool, error) {
+	if id == types.EmptyNodeID {
+		return false, errors.New("cache db: get empty node id")
+	}
```
poszu, line 55: Why special-case it?

datastore/cache.go
countvonzero: why? my intention is to make it clear that what's cached application wise. 
ppl who want to instantiate a generic cache can still do it...

datastore/store.go
countvonzero: i kinda do in the identity case. i want it to have recency bias.

datastore/store.go
countvonzero: this is a programming error. maybe i should log.fatal this...

datastore/store.go
countvonzero: btw, i tried and found that Contains will return false even if there is a entry (key, nil). this is less consistent with my use case.

PR #3965, opened 2023-01-16
[Merged by Bors] - Request 1Gi of memory for poet in systests

poszu:
## Motivation
16Gi is excessive for "normal" systests where the epoch is 1 minute. It is problematic when running a local minikube cluster.

## Changes
Request 1Gi for poet in systests


dshulyak (approved):

poszu:
bors merge

Issue #3963, opened 2023-01-16
Allow nodes to increase their claimed PoST size.


fasmat:
## Description
Nodes may increase their PoST size in the future and we should support this on a protocol level.

## Acceptance criteria
- Incoming ATXs are allowed to increase the claimed `numUnits` again.
- When an ATX is received that has a higher `NumUnits` value than the previous ATX of the same identity it may also contain a VRF Nonce that satisfies the increased difficulty.
  - Alternatively if it doesn't contain a new nonce the current needs to be re-validated to satisfy the increased difficulty
- The changed `numUnits` value becomes effective in the `TargetEpoch` of the ATX, when `numUnits` was lowered, but when `numUnits` is increased it becomes effective one epoch after `TargetEpoch`.
  - A changed VRFNonce always becomes effective in `TargetEpoch`.
  - see also these comments: https://github.com/spacemeshos/pm/issues/209#issuecomment-1377810006, https://github.com/spacemeshos/go-spacemesh/issues/3773#issuecomment-1339597893, https://github.com/spacemeshos/go-spacemesh/issues/3773#issuecomment-1374781236

## Implementation hints
- Introduce a new `EffectiveNumUnits` property for `VerifiedActivationTx` that returns the smaller `numUnits` value of the ATX and its `prevATX`.
  - This property might be persisted in the local DB for efficiency but should not be serialised when transmitting an ATX over the network.
- Use `EffectiveNumUnits` as input for `GetWeight` calculation of `VerifiedActivationTx`.

PR #3966, opened 2023-01-16
Handle updated VRF Nonces correctly

### LINKED ISSUE #3962: Allow nodes to update their VRF Nonce
### Description
Nodes must be able to update their VRF Nonce in case they increase their allocated PoST size.

### Acceptance criteria
- Any incoming ATX (instead of just initial ATXs) may contain a VRFNonce that shall be used for the identity starting in the `TargetEpoch` of the ATX.
- New VRFNonces need to be verified to satisfy the required difficulty as set by the claimed PoST size.

### Implementation hints
- Instead of fetching the VRF Nonce from the initial ATX of an identity, VRF Nonces should be stored along with the identity they belong to and the epoch from which on they become valid in the nodes database
   - When verifying a VRF signature the epoch in which it was produced needs to be used to fetch the right VRF Nonce for the given identity
   - previous VRF Nonces should not be removed from the local node state until some cohesive strategy for pruning nodes has been specified

fasmat:
## Motivation
Closes #3962

## Changes
VRF Nonces are now stored by Identity in the database. Incoming ATXs are checked for new (valid) VRF Nonces which are used starting from the TargetEpoch of the ATX.

## Test Plan
Tests were added that new VRF Nonces are handled correctly. Logic to update ones VRF Nonce is not yet implemented and comes with a future change.

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try-

fasmat:
bors try

fasmat:
bors try

activation/activation.go
```diff
@@ -660,9 +659,7 @@ func (b *Builder) createAtx(ctx context.Context, challenge *types.NIPostChalleng
 		poetChallenge.InitialPost = b.initialPost
 		poetChallenge.InitialPostMetadata = b.initialPostMeta
 	}
-	buildingNipostCtx, cancel := context.WithDeadline(ctx, nextPoetRoundStart)
-	defer cancel()
```
countvonzero: why remove this?

activation/handler.go
```diff
@@ -121,23 +122,38 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.VerifiedActivationT
 	if existingATX != nil { // Already processed
 		return nil
 	}
-	epoch := atx.PublishEpoch()
 	h.log.WithContext(ctx).With().Info("processing atx",
-		atx.ID(),
-		epoch,
+		log.Stringer("atx_id", atx.ID()),
```
countvonzero: curious why you changed here and elsewhere. the filed name is exactly "atx_id" with just atx.ID().

sql/migrations/0001_initial.sql
```diff
@@ -113,6 +113,15 @@ CREATE TABLE atxs
 CREATE INDEX atxs_by_smesher_by_epoch_desc ON atxs (smesher, epoch desc);
 CREATE INDEX atxs_by_epoch_by_pubkey ON atxs (epoch, smesher);
 
+CREATE TABLE vrf_nonces
```
countvonzero: what do you think of adding vrf_nonces field in the atxs table? 
conceptually they should be a property of ATX, instead of having a separate life-cycle.

when you query, just pick the nonce in atxs table with the highest epoch.

activation/handler.go
```diff
@@ -153,12 +169,24 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.VerifiedActivationT
...
 	if atx.PrevATXID == *types.EmptyATXID {
 		if err := h.validateInitialAtx(ctx, atx); err != nil {
 			return nil, err
 		}
+		commitmentATX = atx.CommitmentATX
 	} else {
-		if err := h.validateNonInitialAtx(ctx, atx); err != nil {
+		commitmentATX, err = h.getCommitmentAtx(atx)
+		if err != nil {
+			return nil, fmt.Errorf("validation failed: commitment atx for %s not found: %w", atx.NodeID(), err)
```
countvonzero: i think you are using error for logging purpose...
maybe shorten it a little?

activation/handler.go
```diff
@@ -153,12 +169,24 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.VerifiedActivationT
...
+	var (
+		commitmentATX *types.ATXID
+		err           error
+	)
+
 	if atx.PrevATXID == *types.EmptyATXID {
 		if err := h.validateInitialAtx(ctx, atx); err != nil {
 			return nil, err
 		}
+		commitmentATX = atx.CommitmentATX
```
countvonzero: where is atx.CommitmentATX checked for nil in this case? i can't seem to find it in the code...

activation/handler.go
```diff
@@ -232,6 +255,13 @@ func (h *Handler) validateNonInitialAtx(ctx context.Context, atx *types.Activati
 		return fmt.Errorf("num units %d is greater than previous atx num units %d", atx.NumUnits, prevAtx.NumUnits)
 	}
 
+	if atx.VRFNonce != nil {
```
countvonzero, line 244: what is the protocol rule here? atx can arbitrarily update its vrf nonce? or the PoST size have to increase in this case? if so, does it need to be verified?

signing/vrf.go
```diff
@@ -10,26 +10,32 @@ import (
...
-	nodeID types.NodeID
+	privateKey []byte
+	nodeID     types.NodeID
 }
 
 // Sign signs a message for VRF purposes.
-func (s VRFSigner) Sign(msg []byte) ([]byte, error) {
-	nonce, err := s.fetcher.NonceForNode(s.nodeID)
+func (s VRFSigner) Sign(msg []byte, epoch types.EpochID) ([]byte, error) {
+	nonce, err := s.fetcher.NonceForNode(s.nodeID, epoch)
```
countvonzero, line 33: this API is confusing.

it is more intuitive to pass the nonce as a parameter to Sign, instead of passing all the parameters needed to find the nonce.

in terms of code dependency, it is strange that we need a database to create a vrf signer. it should be just a nonce, regardless of where the nonce is stored.

besides, and it's prolly sounding old now, that i still think the code will be easier to read if we just add nonce to be part of the vrf message.

sql/migrations/0001_initial.sql
dshulyak: there is an option to store the last one in the identities table

activation/activation.go
fasmat: it serves no purpose?

`buildingNipostCtx` is created from `ctx` and canceled with `defer` after `createAtx` returns. The only function it is passed into is `b.nipostBuilder.BuildNIPost`. This function is synchronous, when it returns it is done.

fasmat: Late night changes... I read it as `context.WithCancel` instead of `context.WithDeadline`. Will revert

activation/handler.go
fasmat: I updated some log lines because I was bug hunting and I was confused by log lines which had multiple fields with the same name but different values, i.e. 3 `node_id` fields, one of which was different from the others.

I've reverted some of the log changes.

sql/migrations/0001_initial.sql
fasmat: I decided on a dedicated table for VRF Nonces because of the following:

- most ATXs published by a node will not contain a VRF Nonce, it is only necessary that the initial ATX contains one
- if a node publishes a new VRF Nonce (lets say in epoch 50) it becomes effective the epoch after. any messages published before epoch 51 use the initial VRF Nonce and any messages after epoch 50 will use the new one

> what do you think of adding vrf_nonces field in the atxs table?
> conceptually they should be a property of ATX, instead of having a separate life-cycle.

VRF nonces have their own life cycle, the majority of ATXs will not contain a VRF Nonce. A significant amount of smeshers will probably never update their initially published VRF Nonce. Only smeshers that decide to increase their PoST at some point after they join the network will start using a different VRF Nonce starting from the epoch they use the increased PoST.

> when you query, just pick the nonce in atxs table with the highest epoch.

That would be wrong, since it depends on which epoch the message was published in which VRF Nonce has to be queried from the table. Lets say Smesher A started with 2 `numUnits` of PoST size and smeshed for a few months. Then in epoch 10 they decide to increase their PoST size to 4 `numUnits` and publish a new VRF Nonce that they use starting from epoch 11.

Now a new Smesher B comes online and syncs with the network. To be able to verify messages published by Smesher A in epoch 1 - 10 they have to use the initial VRF Nonce published by Smesher A, but starting from epoch 11 they have to use the new value.

activation/handler.go
fasmat: `h.validateInitialAtx` returns an `error` if `atx.CommitmentATX` does not reference an existing valid ATX.

added clarifying comment.

activation/handler.go
fasmat: I changed it to: `fmt.Errorf("commitment atx for %s not found: %w", atx.NodeID(), err)`

activation/handler.go
fasmat: > atx can arbitrarily update its vrf nonce?

Yes. A smesher can publish a new VRF Nonce any time, when they do it becomes effective with `atx.TargetEpoch`.

> or the PoST size have to increase in this case?

When PoST size increases the node has 2 options: not publish a new VRFNonce (if the existing one satisfies the difficulty required for the increased PoST size) or publish a new VRFNonce that satisfies the higher difficulty.

> if so, does it need to be verified?

Yes in any case the existing (or new) VRFNonce has to be validated again for the higher difficulty when the PoST size increases. This is not part of this PR/Issue (https://github.com/spacemeshos/go-spacemesh/issues/3962) but will be done in a future PR: https://github.com/spacemeshos/go-spacemesh/issues/3963

signing/vrf.go
fasmat: I chose this design for the following reason: Everywhere `VRFSigner::Sign` is used (and similarly for `VRFVerifier::Verify`), the epoch is known and can be passed into the function.

If instead `Sign` (or `Verify`) would take the Nonce as parameter all places where they are invoked need to be updated to fetch the correct nonce first using the `NodeID` (that in the places where `Sign` is used, isn't always readily available) and `epoch` just to then pass it into this function.

> besides, and it's prolly sounding old now, that i still think the code will be easier to read if we just add nonce to be part of the vrf message.

If you mean that we should change the `VRFSigner` and `VRFVerifier` to be "dumb" and only sign / verify a payload / signature, I see some issues with this approach:

This would mean that we need to extend multiple places (`hare oracle`, `proposal eligibility_verifier`, `miner oracle`, `weak_coin`, `beacon` and `beacon handler`) to all do the same thing:

1. fetch the correct nonce for the given identity from the database using the `NodeID` and `epoch` (and sometimes the `NodeID` would first be needed to be passed in from further up the call chain)
2. serialize that nonce into the message or use a dedicated type for a VRFMessage like:

```golang
type VRFMessage {
    Nonce   types.VRFPostIndex
    Message []byte
}
```

and use that to encode / decode the byte array with scale before it is signed or verified against. It is an option but it would repeat a lot of the same code in multiple places where they are not directly part of the logic that the service should care about.

Also if we make any change to the scheme we are using for VRFMessages (like this PR) instead of just changing it in one place we have to change it in multiple places increasing the likelihood that something is overlooked or inconsistent between different protocols. No matter if `beacon`, `weak_coin` or `hare` use a VRF Signature imho it should always everywhere be the same, so it makes sense to put the logic for fetching the correct nonce and serializing the message into the service that signs the payload.

Additionally tests for those services would then need to be updated to be able to fetch the nonce from the In-Memory SQL Database instead of just mocking the `VRFSigner` / `VRFVerifier`.

fasmat:
bors try

sql/migrations/0001_initial.sql
fasmat: After discussing the topic in more detail with @countvonzero I changed the implementation such that the VRF Nonce is now a (nullable) column in the atxs table and updated the code for persisting and fetching the nonce accordingly.

This allows in the future to more easily prune local state by just considering ATXs, and keep at least the newest ATX of an identity that contains a VRF Nonce instead of having to manage two tables.

dshulyak: > This allows in the future to more easily prune local state by just considering ATXs, and keep at least the newest ATX of an identity that contains a VRF Nonce instead of having to manage two tables.

but this is irrelevant. if identity is cancelled it doesn't matter what is a previous vrf nonce. 

actually i realized that you need to store all of them, and fetch a correct one when validating vrf's, so my idea is broken

signing/vrf.go
dshulyak: this is a huge mistake in my opinion, it just accidentally adds fetcher and database to the hare codepath, it changes latency expectations by quite a lot

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

signing/vrf.go
countvonzero: for the record, this is not what i suggested. take beacon vrf message for example, it would result in the following
```
type beaconVRF struct {
  beacon_eligibility
  nonce
  epoch  
}
```
that way vrf verifier does not need to take any parameter at all. it's up to the app logic to construct the expected vrf message and verify the proof (vrf signature) against it.

fasmat:
Closed after merge of https://github.com/spacemeshos/go-spacemesh/pull/3971

PR #3971, opened 2023-01-18
[Merged by Bors] - Handle updated VRF Nonces correctly without VRFSigner service

### LINKED ISSUE #3902: Re-assess nonce encoding in VRF signatures
## Description

## Affected code

### LINKED ISSUE #3962: Allow nodes to update their VRF Nonce

fasmat:
## Motivation
Alternative approach for https://github.com/spacemeshos/go-spacemesh/pull/3966

Closes https://github.com/spacemeshos/go-spacemesh/issues/3902 and closes https://github.com/spacemeshos/go-spacemesh/issues/3962

## Changes
Instead of using a service to fetch nonces and sign VRF messages, put the logic into the services themselves and make the VRF Signer/Verifier only sign messages

## Test Plan
n/a

## TODO

## DevOps Notes


fasmat:
bors try

signing/signer.go
```diff
@@ -114,22 +115,9 @@ func (es *EdSigner) PrivateKey() PrivateKey {
...
-	}
-
-	if err := cfg.validate(); err != nil {
-		return nil, err
-	}
-
+func (es *EdSigner) VRFSigner() (*VRFSigner, error) {
 	return &VRFSigner{
-		privateKey: es.priv,
+		privateKey: oasis.PrivateKey(es.priv),
```
countvonzero, line 120: i am confused by this cast. why is this needed?

proposals/handler_test.go
```diff
@@ -61,21 +61,22 @@ type testHandler struct {
...
+		mNonce: NewMocknonceFetcher(ctrl),
 	}
 }
 
 func createTestHandler(t *testing.T) *testHandler {
 	types.SetLayersPerEpoch(layersPerEpoch)
 	ms := fullMockSet(t)
 	return &testHandler{
-		Handler: NewHandler(datastore.NewCachedDB(sql.InMemory(), logtest.New(t)), ms.mpub, ms.mf, ms.mbc, ms.mm, ms.md,
+		Handler: NewHandler(datastore.NewCachedDB(sql.InMemory(), logtest.New(t)), ms.mpub, ms.mf, ms.mbc, ms.mm, ms.md, ms.mvrf, ms.mNonce,
```
countvonzero: looks like its not needed in this test.

beacon/message.go
```diff
@@ -14,6 +14,7 @@ import (
 type ProposalMessage struct {
 	EpochID      types.EpochID
 	NodeID       types.NodeID
+	Nonce        types.VRFPostIndex
```
countvonzero: this is incorrect. we don't need Nonce here. this is not the VRF message being signed. this is just the msg sent over the wire during the beacon protocol.

beacon/weakcoin/weak_coin.go
```diff
@@ -222,6 +235,7 @@ func (wc *WeakCoin) prepareProposal(epoch types.EpochID, round types.RoundID) ([
 				Round:     round,
 				Unit:      unit,
 				MinerPK:   wc.signer.PublicKey().Bytes(),
+				Nonce:     nonce,
```
countvonzero: same here. this is the gossiped message. not the vrf message. don't add nonce here.

countvonzero (commented):
you can also close https://github.com/spacemeshos/go-spacemesh/issues/3902 with this PR

signing/signer.go
fasmat: `es.priv` is a `crypto.PrivateKey` defined in the standard library. The but the signer uses `"github.com/oasisprotocol/curve25519-voi/primitives/ed25519"` which for some reason defined its own type of `PrivateKey`. Both are just `[]byte` but there are no implicit casts in go (at least not from non-basic type to non-basic type), so this explicit cast is necessary.

proposals/handler_test.go
fasmat: Not in this file, but `createTestHandler` is also used in `elegibility_validator_test.go` and there I set expectations on those mocks.

cmd/node/node.go
```diff
@@ -458,6 +459,14 @@ func (app *App) SetLogLevel(name, loglevel string) error {
 	return nil
 }
 
+type nonceFetcherAdapter struct {
+	cdb *datastore.CachedDB
+}
+
+func (nfa nonceFetcherAdapter) VRFNonce(nodeID types.NodeID, epoch types.EpochID) (types.VRFPostIndex, error) {
+	return atxs.VRFNonce(nfa.cdb, nodeID, epoch)
+}
```
fasmat: This is probably not the right place for this; but I'm not sure in which package to put it? `datastore`?

countvonzero: i think i will prefer creating a option for each component's `NewXXXX` function.
that option will default to a function that executes `txs.VRFNonce(cdb, nodeID, epoch)` (because cdb is already passed to most components), and gets overwritten in tests

countvonzero: similar to this https://github.com/spacemeshos/go-spacemesh/blob/3c2f1837675840e69e3f5bb5cabbd9a9e54807ce/proposals/handler.go#L79

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try-

fasmat:
bors -try

fasmat:
bors try

fasmat:
bors try

cmd/node/node.go
fasmat: I added a few functional options to inject a mocked `nonceFetcher` in tests. If not used a standard implementation is used that reads the nonce from the ATX table using `sql/atxs`

beacon/beacon.go
```diff
@@ -669,16 +688,22 @@ func (pd *ProtocolDriver) sendProposal(ctx context.Context, epoch types.EpochID)
 	}
 
 	logger := pd.logger.WithContext(ctx).WithFields(epoch)
-	proposedSignature := buildSignedProposal(ctx, pd.vrfSigner, epoch, pd.logger)
+	nonce, err := pd.nonceFetcher.VRFNonce(pd.nodeID, epoch)
+	if err != nil {
+		logger.With().Panic("failed to get VRF nonce", log.Err(err))
```
countvonzero: prefer `logger.Fatal()`
stack trace is usually useless.

beacon/weakcoin/weak_coin.go
```diff
@@ -78,17 +79,23 @@ func WithNextRoundBufferSize(size int) OptionFunc {
...
 	wc := &WeakCoin{
-		logger:    log.NewNop(),
-		config:    defaultConfig(),
-		signer:    signer,
-		publisher: publisher,
-		coins:     make(map[types.RoundID]bool),
-		verifier:  verifier,
+		logger:       log.NewNop(),
+		config:       defaultConfig(),
+		cdb:          cdb,
```
countvonzero: not needed. the nonceFetcher is sufficient.

beacon/weakcoin/weak_coin_test.go
```diff
@@ -337,6 +364,12 @@ func TestWeakCoinNextRoundBufferOverflow(t *testing.T) {
 	require.True(t, wc.Get(context.Background(), epoch, nextRound))
 }
 
+type vrfVerifierAdapter struct{}
+
+func (m vrfVerifierAdapter) Verify(nodeID types.NodeID, msg, sig []byte) bool {
```
countvonzero: why is this needed?

beacon/weakcoin/weak_coin_test.go
```diff
@@ -358,19 +391,25 @@ func TestWeakCoinEncodingRegression(t *testing.T) {
...
+		types.RandomNodeID(),
+		vrfSig,
+		vrfVerifierAdapter{},
+		nonceFetcher(t, ctrl),
+		weakcoin.WithThreshold([]byte{0xff}),
+	)
 	instance.StartEpoch(context.Background(), epoch, allowances)
 	require.NoError(t, instance.StartRound(context.Background(), round))
 
+	// TODO(mafa): figure out why sig is now different on every test run
```
countvonzero: but it should change. nonce is an extra data to the vrf message now.

countvonzero (approved):

beacon/weakcoin/weak_coin_test.go
fasmat: When I wrote that code there VRFVerifier was just a function so I needed to wrap it into an object to satisfy the interface. Since then I created a `VRFVerifier` type in the `signing` package and this is not needed any more.

Removed.

beacon/weakcoin/weak_coin_test.go
fasmat: The problem was that every time I ran the test the signature would be different (without changing code). I've fixed the issue so this comment is obsolete.

Removed.

sql/atxs/atxs.go
```diff
@@ -208,6 +208,30 @@ func GetIDsByEpoch(db sql.Executor, epoch types.EpochID) (ids []types.ATXID, err
...
+		stmt.BindInt64(2, int64(epoch))
+	}
+	dec := func(stmt *sql.Statement) bool {
+		nonce = types.VRFPostIndex(stmt.ColumnInt64(0))
+		return true
+	}
+
+	if rows, err := db.Exec(`
+		select nonce from atxs
+		where smesher = ?1 and epoch <= ?2 and nonce is not null
```
countvonzero: the recorded epoch is publish epoch. so here i think you want `<` instead of `<=`

countvonzero: never mind. i am wrong.

countvonzero: actually you do want `<` here. the epoch you passed in is target epoch

fasmat: You are probably right: when I wrote that query I was persisting the nonce based on `TargetEpoch`, but `epoch` here refers to `PublishEpoch`. Changed `<=` to `<`.

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

hare/algorithm.go
```diff
@@ -239,6 +240,7 @@ func newConsensusProcess(
 	stateQuerier stateQuerier,
 	signing Signer,
 	nid types.NodeID,
+	nonce types.VRFPostIndex,
```
countvonzero, line 243: this needs to be assigned.

hare/interfaces.go
```diff
@@ -17,7 +17,7 @@ type layerPatrol interface {
 type Rolacle interface {
 	Validate(context.Context, types.LayerID, uint32, int, types.NodeID, []byte, uint16) (bool, error)
 	CalcEligibility(context.Context, types.LayerID, uint32, int, types.NodeID, []byte) (uint16, error)
```
countvonzero: this needs to add nonce too. this is only used by own hare generating proof.

hare/eligibility/oracle.go
```diff
@@ -167,15 +185,16 @@ func (o *Oracle) getBeaconValue(ctx context.Context, epochID types.EpochID) (uin
...
-		log.Uint32("beacon_val", value))
+		log.Uint32("beacon_val", value),
+	)
 	return value, nil
 }
 
 // buildVRFMessage builds the VRF message used as input for the BLS (msg=Beacon##Layer##Round).
-func (o *Oracle) buildVRFMessage(ctx context.Context, layer types.LayerID, round uint32) ([]byte, error) {
+func (o *Oracle) buildVRFMessage(ctx context.Context, nonce types.VRFPostIndex, layer types.LayerID, round uint32) ([]byte, error) {
 	key := buildKey(layer, round)
```
countvonzero: need to get rid of the vrfMsgCache. this is what causing the vrf signature to be invalid. once the node sign its own vrf message, it is validating all other nodes' vrf signature with the cached msg

fasmat: I removed the cache and fixed all unit tests that broke because of it; running a new `bors try` now

fasmat:
bors try

fasmat:
bors try

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

PR #3976, opened 2023-01-19
[Merged by Bors] - Using PubKeyExtractor as injected service instead of as a global function

### LINKED ISSUE #3902: Re-assess nonce encoding in VRF signatures

fasmat:
## Motivation
This is a followup to #3971 and closes #3902. Merge after #3971 

## Changes
Instead of using the global variable `ExtractNodeIDFromSig` in `common/types` inject a service into the places where pubkey-extraction is needed. This still allows to use different prefixes between tests and actually running the node but avoids using the global function.

## Test Plan
All existing tests still pass

## TODO

## DevOps Notes


countvonzero:
i am not super sure about this. why does component need a custom function for extracting public key from signature?
it seems like a perfect use case for a global static function.

fasmat:
> i am not super sure about this. why does component need a custom function for extracting public key from signature?
it seems like a perfect use case for a global static function.

It is not static but changes during startup of the node based on the configuration (network prefix) that is loaded from the config. The global function is defined without prefix for tests [here](https://github.com/spacemeshos/go-spacemesh/blob/32b67e2c2d561b991bdf2d7808de4f9c656e5a90/common/types/nodeid.go#L17-L27) and is overwritten during startup [here](https://github.com/spacemeshos/go-spacemesh/blob/32b67e2c2d561b991bdf2d7808de4f9c656e5a90/cmd/node/node.go#L508-L516).

countvonzero:
> > i am not super sure about this. why does component need a custom function for extracting public key from signature?
> > it seems like a perfect use case for a global static function.
> 
> It is not static but changes during startup of the node based on the configuration (network prefix) that is loaded from the config. The global function is defined without prefix for tests [here](https://github.com/spacemeshos/go-spacemesh/blob/32b67e2c2d561b991bdf2d7808de4f9c656e5a90/common/types/nodeid.go#L17-L27) and is overwritten during startup [here](https://github.com/spacemeshos/go-spacemesh/blob/32b67e2c2d561b991bdf2d7808de4f9c656e5a90/cmd/node/node.go#L508-L516).

yes. but this function is not going to be used before the app initialization is finished. why not just do the following
```
var ExtractNodeIDFromSig = func(msg, sig []byte) (types.NodeID, error) {
	log.Fatal("uninitialized")
	return types.NodeID{}, errors.New("extractor not initialized")
}
```
and do the initialization in node.go as early as possible.

countvonzero:
i simply don't understand the need having to pass key extractor to all components when a global static function would do.

fasmat:
> why not just do the following ... and do the initialization in node.go as early as possible.

Then we would also need to set this function in many tests - at least tests that use ATXs, Ballots, or Proposals - or those tests would fail.

Why not make it injectable? Then it can be mocked (currently not a requirement, but might be interesting in the future). 

> i simply don't understand the need having to pass key extractor to all components when a global static function would do.

It is weird to me that `signing.EdSigner` is used via injection, but `signing.KeyExtractor` is not. The former is instantiated during startup and passed into all services, the later is also instantiated during startup and then set as value to a global variable?

Another argument against making it a global function is that we then have a tightly coupled dependency between the package where the function is defined (`common/types`) and any other package that sets or uses that function. Injecting loosens up that coupling.

The primary reason this function exists at the moment is to break up a cyclic dependency. `signing` depends on `common/types` but `ExtractNodeIDFromSig` (or at least the implementation that is set during node startup) is dependent on the `signing` package.

countvonzero:
> Why not make it injectable? Then it can be mocked (currently not a requirement, but might be interesting in the future).

i don't believe in injecting crypto.... the only tangible reason would be speeding up the tests.

> It is weird to me that signing.EdSigner is used via injection, but signing.KeyExtractor is not. The former is instantiated during startup and passed into all services, the later is also instantiated during startup and then set as value to a global variable?

i don't see why symmetry here is important. signing is for the node itself, so instead of reading the private key multiple times, you read it once and pass it to components that signs. 

key extractor is a generic functions that extract public key from signatures. a global function that gets initialized once and all components just use it.

> Another argument against making it a global function is that we then have a tightly coupled dependency between the package where the function is defined (common/types) and any other package that sets or uses that function. Injecting loosens up that coupling.

LOL. but which package now doesn't have a dependency on common/types?

> The primary reason this function exists at the moment is to break up a cyclic dependency. signing depends on common/types but ExtractNodeIDFromSig (or at least the implementation that is set during node startup) is dependent on the signing package.

could we move all things signing to common/types package then?

in any case, injecting signer and key extractor for me is more theoretical than practical. but i don't care that much.

countvonzero (approved):

fasmat:
> but which package now doesn't have a dependency on common/types?

That's not what I meant. Yes, almost every package is dependent on `common/types` but it doesn't work without another package first defining what it does.

It is a hidden cyclic dependency. `cmd/node` depends on `common/types` as it should. But `common/types` depends on `cmd/node` to define what it does.

The reason when have to set this function during runtime and can't just implement it normally is because the complier would complain. Yes merging `signing` into `common/types` would be another way to resolve that issue. One that we both wouldn't be happy about I think 😉

Imo the underlying issue is the `common/types` package itself. In my opinion it shouldn't exist, but this is a whole other discussion and I don't want to open that can of worms just yet. 😕

fasmat:
bors merge

fasmat:
bors merge

Issue #3970, opened 2023-01-18
p2p server: set limits for request and response


dshulyak:
implementation of the p2p server that is used for data requests from a peer doesn't set limits for received objects. it allows to trivially use node's resources, potentially causing oom.

the problem is that some objects can be large, and i lost track how large they can be. ballots may encode many votes, blocks - transactions. perhaps the limit of 1-2mb is good enough. number of concurrent requests also needs to be limited.

PR #3982, opened 2023-01-20
[Merged by Bors] - syncer: sync malfeasance proof when the node is not atx synced

### LINKED ISSUE #3920: malfeasance: sync MalfeasanceProof from peers
## Description

3rd part of https://github.com/spacemeshos/go-spacemesh/issues/3915

`MalfeasanceProof` will be treated like ATXs during sync flow - once the node is atxSynced, it will start listening to 1MalfeasanceGossip1 and save network's `MalfeasanceProof` locally.

note: a not-synced node is capable of validating MalfeasanceProof without the actual conflicting atxs/ballots/hare messages.

countvonzero:
## Motivation

Closes #3920

## Changes

syncer syncs MalfeasanceProof after ATXs are synced. it has to be synced after ATXs to avoid being spammed by unknown identities.

it is synced as follows.
- poll peers for all the malicious NodeIDs
- if those NodeIDs exists, fetch the malfeasance proofs associated with the NodeID hash in batches

this implementation is sub-optimal (see https://github.com/spacemeshos/go-spacemesh/issues/3987) and is intended to be a genesis compromise


dshulyak:
> poll peers for all the malicious NodeIDs

but why does it poll for node ids and not proofs themself from the start?
do you save node ids or just keep in memory? i guess if they are saved then they need to be cleaned up if there is no proof for them afterwards. but i really don't understand why is it done like this

> sync malfeasance proof at the top of every iteration

they can't be synced first as you don't know if such identity has anything in the state. basically it it the same spam as noted before. one peer can generate any number of identities and proofs for them, then poison its peers  and those peers will poison the rest of the network.

i doubt that there is an option to download proofs before atxs

countvonzero:
> > poll peers for all the malicious NodeIDs
> 
> but why does it poll for node ids and not proofs themself from the start? do you save node ids or just keep in memory? i guess if they are saved then they need to be cleaned up if there is no proof for them afterwards. but i really don't understand why is it done like this

proofs are saved in the identity table and one identity 
has at most one proof. my intention was to limit the data size transmitted. so fetching malicious IDs first and use NodeID as hash to fetch the proof data in batches (like how ballots and blocks are fetched by hash) will help limit the data size in malfeasance proof.

> 
> > sync malfeasance proof at the top of every iteration
> 
> they can't be synced first as you don't know if such identity has anything in the state. basically it it the same spam as noted before. one peer can generate any number of identities and proofs for them, then poison its peers and those peers will poison the rest of the network.
> 
> i doubt that there is an option to download proofs before atxs

you are right. this is subject to the identity spams. what do you think about syncing atxs first, then the malicious NodeIDs, then fetch the actual proof data via hash batches? @dshulyak 

i am planning to remove the server request/response design and read streamed bytes directly in p2p after features are done. then i can get rid of the 2 staged fetch in the current code for this and other data types too.

countvonzero:
on the other hand, getting the IDs first allows the node to only fetch data that doesnt exist locally.

countvonzero:
@dshulyak 
i changed the code such that 
- it syncs malfeasance proof after ATXs are synced
- it only syncs malfeasance proof when the node is not ATX synced.
- it starts listening to malfeasance gossip after the node is ATX synced.
- it only fetches the proof from peers if the identity exists.

also filed https://github.com/spacemeshos/go-spacemesh/issues/3987 for post-genesis improvement.
i'm planning to work on that along with https://github.com/spacemeshos/go-spacemesh/issues/3970 once go-spacemesh is stable for genesis.

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #3986, opened 2023-01-21
Refactor VRF Signer

### LINKED ISSUE #3902: Re-assess nonce encoding in VRF signatures

fasmat:
## Motivation
Closes #3902 

## Changes
Only refactors VRF Signer and Verifier to not fetch nonces from the DB. Only refactors existing code in a functional identical way; does not implement #3962 or other features!

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
bors try

fasmat:
Superseded by https://github.com/spacemeshos/go-spacemesh/pull/3971

Issue #3991, opened 2023-01-23
potential memory leak in newer hare versions


dshulyak:
there seems to be a memory leak regression, i launched network on friday and it consistently hits the limit of 1gb and gets restarted
see first screenshot with container_memory_rss (it is basically an anonymous memory), on my previous run this metric didn't go above ~200mb
second screenshot is a dump of the heap from pprof

![image(1)](https://user-images.githubusercontent.com/1964576/214074616-61f2e49d-dabf-41a8-91d2-a32dbc1739a9.png)
![image(2)](https://user-images.githubusercontent.com/1964576/214074694-a6519e76-deb0-4334-b35c-c6d88a5c00d9.png)

dshulyak:
pr missed close <issue number>

PR #3994, opened 2023-01-24
[Merged by Bors] - hare: fix memory leak

countvonzero:
## Motivation

fix hare memory leak #3991

## Changes

- change hare time-related configuration to using Duration to facilitate testing
- create a test that fails under old code (stuck at `consensusProcess.terminate()`)
- separate cleanup and terminate logic
  - move waiting for errgroup into a separate function `Stop()`
  - keep created consensus process inside hare, and invoke `Stop()` after unregistering from message broker


countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):

hare/algorithm_test.go
```diff
@@ -182,48 +178,59 @@ func (mev *mockEligibilityValidator) ValidateEligibilityGossip(context.Context,
...
+	mo.EXPECT().Proof(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()
+	mo.EXPECT().CalcEligibility(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), proc.nid, gomock.Any()).Return(uint16(1), nil).AnyTimes()
+	proc.oracle = mo
+
+	proc.value = NewSetFromValues(types.ProposalID{1}, types.ProposalID{2})
+	proc.Start()
+	require.Eventually(t, func() bool {
+		proc.Stop()
+		return true
+	}, 500*time.Millisecond, 100*time.Millisecond)
```
poszu, line 234: It could be simplified to:
```suggestion
	proc.Stop()
```

The test will timeout if `proc.Stop()` never returns.

hare/hare.go
```diff
@@ -174,8 +173,8 @@ func (h *Hare) HandleEligibility(ctx context.Context, emsg *types.HareEligibilit
 }
 
 func (h *Hare) getLastLayer() types.LayerID {
-	h.layerLock.RLock()
-	defer h.layerLock.RUnlock()
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
poszu, line 210: Why did you change the `RWMutex` to a `Mutex`? It made sense to lock for reading only here.

hare/hare.go
```diff
@@ -206,9 +205,15 @@ func (h *Hare) outOfBufferRange(id types.LayerID) bool {
 }
 
 func (h *Hare) oldestResultInBuffer() types.LayerID {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	return h.oldestResultInBufferLocked()
```
poszu, line 243: It could benefit from an `RWMutex` too

hare/hare.go
```diff
@@ -342,14 +344,43 @@ func (h *Hare) onTick(ctx context.Context, id types.LayerID) (bool, error) {
...
+	defer h.mu.Unlock()
+	h.cps[cp.ID()] = cp
+	logger.With().Info("number of consensus processes (after register)",
+		log.Int("count", len(h.cps)))
+}
+
+func (h *Hare) getCP(lid types.LayerID) Consensus {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	return h.cps[lid]
```
poszu, line 403: It could benefit from `RWMutex` too if the `value, ok := h.cps[lid]` notation was used.

poszu (approved):
LGTM.

I'd bring back the `RWMutex` though.

hare/algorithm_test.go
countvonzero: i don't see why test timing out is better. when i do `go test` it will just hang.

hare/hare.go
countvonzero: mostly because i prefer code simplicity over premature optimization. 

- Mutex is more straightforward to use
- RWMutex.Lock is slower than Mutex.Lock
- in hare module, the only place one needs to worry about thread-safety is the gossip handler. the rest should not have lock contention

i'd use RWMutex only when the component is write-once and read-many-times. otherwise i prefer simplicity.

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #3996, opened 2023-01-24
[Merged by Bors] - validate that network remains in sync and has the same consensus/state

### LINKED ISSUE #3974: systests: validation for longevity tests
basic validation that nodes are in consensus and has the same state.
need to tolerate nodes failures (chaos mesh) and temporary difference in consensus/state while node recovers from chaos.

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/3974


systest/Makefile
```diff
@@ -41,22 +42,26 @@ docker:
 push:
 	docker push $(image_name)
 
+.PHONY: template
+template:
+	@echo smesher=`cat $(smesher_config)` >> $(tmpfile)
+	@echo poet=`cat $(poet_config)` >> $(tmpfile)
```
poszu: There is a risk of getting an invalid file if it just appends, when the file already exists. Consider writing on the first `echo`, or removing the file:
```suggestion
.PHONY: template
template:
	@echo smesher=`cat $(smesher_config)` > $(tmpfile)
	@echo poet=`cat $(poet_config)` >> $(tmpfile)
```

systest/cluster/nodes.go
```diff
@@ -38,16 +39,46 @@ var (
...
+				apiv1.ResourceCPU:    resource.MustParse("1"),
+				apiv1.ResourceMemory: resource.MustParse("1Gi"),
+			},
+			Limits: apiv1.ResourceList{
+				apiv1.ResourceCPU:    resource.MustParse("2"),
+				apiv1.ResourceMemory: resource.MustParse("1Gi"),
+			},
+		},
+		toResources,
+	)
```
poszu: Why give poet so much CPU? It will take all it gets, basically burning $ for generating bigger POSW. We should be OK with a slower, cheaper poet for systests.

systest/cluster/nodes.go
```diff
@@ -38,16 +39,46 @@ var (
...
 )
 
+func toResources(value string) (*apiv1.ResourceRequirements, error) {
+	var rst apiv1.ResourceRequirements
+	if err := json.Unmarshal([]byte(value), &rst); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal %s into apiv1.ResourceRequirements: %w", value, err)
```
poszu: NIT: describe only the context, leave the "failed" keyword for logging
```suggestion
		return nil, fmt.Errorf("unmarshaling %s into apiv1.ResourceRequirements: %w", value, err)
```

to avoid logs like: "Failed <something>: failed to <x>: failed to <y> ..."

systest/parameters/longfast/resources.env
```diff
@@ -0,0 +1,2 @@
+poet_resources={"requests": {"cpu": "1", "memory": "16Gi"}, "limits": {"cpu": "2", "memory": "16Gi"}}
```
poszu, line 1: NIT: Poet can be configured to keep more layers of the tree on disk with `--memory=N` with a small N like even `0`. Perhaps we could save some $ in CI  by keeping the tree on disk and requesting less memory for poet Pods. The downside is that generation of the poet proof (Merkle proof, not the sequential work itself) will take longer.

systest/validation/consensus.go
```diff
@@ -0,0 +1,165 @@
...
+	for i, n := range c.failures {
+		if n == prev[i] {
+			c.failures[i] = 0
+		}
+	}
+	for i, rst := range c.failures {
+		if rst > c.tolerate {
+			return fmt.Errorf("node %d failed to reach consensus in %d period(s)",
+				i, rst,
+			)
```
poszu, line 136: NIT: It could improve debuggability to have the name of the node's pod in the log.

systest/validation/consensus_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConsensusValidation(t *testing.T) {
+	type item struct {
+		consensus, state string
+	}
```
poszu, line 13: IMO the unit tests should run with `make test` but they are currently excluded in https://github.com/spacemeshos/go-spacemesh/blob/2ae0df137fe2fcec2ec172a2d040b0f2ac81e2fd/Makefile#L7

systest/Makefile
dshulyak: i am removing a file after reading it in kubectl, see config subcommand
this is the best i could come up with. do you know how it can be done better?

dshulyak: actually didnt notice, changing the first >> to > is neat. will use

systest/validation/consensus.go
```diff
@@ -0,0 +1,165 @@
...
+		EndLayer:   &pb.LayerNumber{Number: uint32(target)},
+	})
+	if err != nil {
+		return nil
+	}
+	if len(ls.Layer) != 1 {
+		return nil
+	}
+	layer := ls.Layer[0]
+	return &ConsensusData{Consensus: layer.Hash, State: layer.RootStateHash}
```
countvonzero, line 47: i removed hash in https://github.com/spacemeshos/go-spacemesh/pull/4026.
the reason being that layer hash wasn't properly updated when block validity changes, and it's not used anywhere.
do you think you need it, or can i use aggregated hash instead? i think that represent consensus better.

systest/parameters/longfast/resources.env
dshulyak: thanks, i will skip that

systest/validation/consensus.go
dshulyak: will think how to pass it nicely in future changes

systest/validation/consensus_test.go
dshulyak: changed to systest/tests

systest/validation/consensus.go
dshulyak: image

i don't see aggregated hash

systest/cluster/nodes.go
dshulyak: ok will set to 0.5

dshulyak:
pushed updates, please review

systest/validation/consensus.go
countvonzero: didn't get your response in time. it's now left as empty.
will do a followup PR. will leave the filed name as Hash but populate it with the aggregated hash.

countvonzero: https://github.com/spacemeshos/go-spacemesh/pull/4046

countvonzero (approved):

dshulyak:
bors merge

PR #3999, opened 2023-01-24
[Merged by Bors] - Proper poet challenge submit timeout

poszu:
## Motivation
Currently, there is a hardcoded 10s timeout for all requests to the Poet service. It's limiting, we might want to wait longer or shorter depending on the situation. The user should set the timeout. 

## Changes
Removed the context factory from the HTTP Poet client. The user of the client decides about the timeout on a per-request basis.

## Test Plan
- added unit tests verifying if a hanged challenge submit times out properly.


poszu:
bors try

cmd/node/node.go
```diff
@@ -679,11 +684,7 @@ func (app *App) initServices(
 	atxBuilder := activation.NewBuilder(builderConfig, nodeID, sgn, app.cachedDB, atxHandler, app.host, nipostBuilder,
 		postSetupMgr, clock, newSyncer, app.addLogger("atxBuilder", lg),
 		activation.WithContext(ctx),
-		activation.WithPoetConfig(activation.PoetConfig{
-			PhaseShift:  app.Config.POET.PhaseShift,
-			CycleGap:    app.Config.POET.CycleGap,
-			GracePeriod: app.Config.POET.GracePeriod,
-		}))
+		activation.WithPoetConfig(poetCfg))
```
fasmat: NIT: end the last line with:

```suggestion
		activation.WithPoetConfig(poetCfg),
	)
```

This way if a line is added or removed from the function call the diff is easier to read 🙂

activation/nipost_test.go
```diff
@@ -527,27 +626,34 @@ func TestNIPostBuilder_Close(t *testing.T) {
...
 	postProver := &postSetupProviderMock{}
-	challenge := types.PoetChallenge{NIPostChallenge: &types.NIPostChallenge{}}
+	challenge := types.PoetChallenge{NIPostChallenge: &types.NIPostChallenge{
+		PubLayerID: (postGenesisEpoch + 1).FirstLayer(),
+	}}
+	poetCfg := PoetConfig{PhaseShift: layerDuration}
 
 	poetDb := NewMockpoetDbAPI(gomock.NewController(t))
-	poetDb.EXPECT().ValidateAndStore(gomock.Any(), gomock.Any()).Return(nil)
+	poetDb.EXPECT().ValidateAndStore(gomock.Any(), gomock.Any()).AnyTimes().Return(nil)
```
fasmat: I think we should specify how often we expect a method of a mock to be called when relevant. Do we expect this call to happen once per sub-test? Is it even called in most sub-tests?

Also: the mock is bound to `t` of the parent test, not to `t` of the sub tests, so if any one of the subtests fail the parent test will be marked as failing without indicating which sub-test was the issue; The error message is then something like "subtest called `t.FailNow()` on parent test".

If possible use a dedicated instance of the mock per sub-test to avoid that issue.

fasmat (approved):
Just some nitpicks

activation/nipost_test.go
poszu: I totally agree.

> The error message is then something like "subtest called t.FailNow() on parent test".

I was wondering why I see these weird failures sometimes. Thanks!

poszu:
bors merge

poszu:
bors merge

PR #4000, opened 2023-01-24
[Merged by Bors] - Update ATX handler to allow increasing numUnits again

### LINKED ISSUE #3963: Allow nodes to increase their claimed PoST size.

### LINKED ISSUE #3992: Add VRF nonce caching
### Description
At the moment hare queries the VRF nonce for any identity on every round. Since an identity cannot change their nonce within an epoch that is unnecessary and should be cached.

### Affected code
Everywhere nonces are fetched from the DB those queries should be cached based on the `NodeID` and `Epoch` such that a nonce for any identity only needs to be fetched at most once per epoch.

- `hare/eligibility/oracle.go` and other places where nonces are fetched from the DB.

### Implementation hints:
- The nonce fetcher already uses `datastore.CachedDB`, which could be extended to cache nonces for identities on a per epoch basis.
- Consider re-using the malfeasance cache since it holds `nil` for all valid identities and could be used to store their VRF nonce.

fasmat:
## Motivation
Closes #3963 and #3992

## Changes
- Incoming ATXs with higher `numUnits` than their `prevAtx` are allowed again
- If `numUnits` is increased from one ATX to the next, re-validate the VRF for the identity
- VRFNonces are now cached when fetched from database: the nonce for an identity is fetched at most once per epoch from the DB and then served from cache
- `EffectiveNumUnits` has been introduced as a property of `ActivationTx` and `ActivationTxHeader`
  - it is the minimum of the `NumUnits` of the current and the previous ATX
  - it is persisted in a new column in the ATX table to not impact performance (otherwise we would always need to fetch 2 ATXs to determine its value)
  - it replaces `NumUnits` in the determination of the weight

## Test Plan
- Tests were added to check if increases in NumUnits are correctly validated by the ATX handler.
- Existing tests were updated to use set EffecitveNumUnits where relevant

## TODO

## DevOps Notes


proposals/eligibility_validator_test.go
```diff
@@ -293,48 +296,6 @@ func TestCheckEligibility_KeyMismatch(t *testing.T) {
 	require.False(t, eligible)
 }
 
-func TestCheckEligibility_ZeroTotalWeight(t *testing.T) {
```
fasmat, line 296: A node cannot have 0 weight an ATX with `NumUnits` 0 is not accepted as a valid ATX.

miner/oracle_test.go
```diff
@@ -197,29 +198,3 @@ func TestOracle_OwnATXNotFound(t *testing.T) {
 	assert.Len(t, activeSet, 0)
 	assert.Len(t, proofs, 0)
 }
-
-func TestOracle_ZeroEpochWeight(t *testing.T) {
```
fasmat, line 201: An ATX with `NumUnits` 0 is not a valid ATX, it will be refused by the ATX handler.

common/types/activation_tx_header.go
```diff
@@ -10,6 +10,12 @@ type ActivationTxHeader struct {
 	// current epoch; a unit represents a configurable amount of data for PoST
 	NumUnits uint32
 
+	// EffectiveNumUnits is the minimum of this ATX's NumUnits and the previous ATX's NumUnits
+	// NumUnit decreases become effective immediately, while NumUnit increases become effective one epoch later
+	// This is because the increased PoST size only becomes effective after a PoET proof has
+	// been generated for it, which is published with the next epoch's ATX.
+	EffectiveNumUnits uint32
```
noamnelke, line 17: Isn't `ActivationTxHeader` the type that gets gossiped/synced? I don't think we should send this value to others, they should calculate it themselves.

common/types/nodeid.go
```diff
@@ -9,8 +9,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
-//go:generate scalegen
-
```
noamnelke: Seems unrelated 🤔

noamnelke (approved):

common/types/nodeid.go
fasmat: It breaks `make generate`; I usually run it before I push something for review. Maybe we should add it as a step to the CI?

It is not required, the only type in this file `NodeID` has a handwritten encode and decode method, but if we leave it `make generate` fails.

noamnelke: Then keep the removal, but I agree that we should add this as a step to CI.

common/types/activation_tx_header.go
fasmat: `ActivationTxHeader` is not broadcasted to the network. It is only used by the node to not need to fetch and deserialize the whole ATX from the database and for caching.

fasmat:
bors try

common/types/activation_tx_header.go
noamnelke: Ah, awesome then 👍🏻

fasmat:
bors merge

common/types/activation_tx_header.go
```diff
@@ -10,6 +10,12 @@ type ActivationTxHeader struct {
 	// current epoch; a unit represents a configurable amount of data for PoST
 	NumUnits uint32
```
countvonzero, line 11: why do the code still need this then?

countvonzero: NVM. dumb question.

activation/handler.go
```diff
@@ -252,6 +263,11 @@ func (h *Handler) validateNonInitialAtx(ctx context.Context, atx *types.Activati
 		return fmt.Errorf("prevATX declared, but initial Post is included")
 	}
 
+	if prevAtx.NumUnits < atx.NumUnits {
+		atx.SetEffectiveNumUnits(prevAtx.NumUnits)
+	} else {
+		atx.SetEffectiveNumUnits(atx.NumUnits)
+	}
```
countvonzero, line 270: issue #3963 stated
> The changed numUnits value becomes effective in the TargetEpoch of the ATX, when numUnits was lowered, but when numUnits is increased it becomes effective one epoch after TargetEpoch.
A changed VRFNonce always becomes effective in TargetEpoch.

how is this achieved?

fasmat: The `EffectiveNumUnits` value is always the minimum of the `NumUnit` value of the ATX and its previous ATX. This `if` sets the value to the minimum of the two. The VRF Nonce becomes effective in the `TargetEpoch` because if `numUnits` was decreased the old VRF Nonce is still valid (still satisfies difficulty threshold) and if `numUnits` was increased the node will provide a new Nonce that is valid for both the lower and the higher `numUnits` value.

Example 1 - smesher decreases `numUnits`:

- Epoch 100: ATX published (`NumUnits` = 5, `VRFNonce` = 100)
- Epoch 101: ATX published (`NumUnits` = 3), used values `NumUnits` = 5, `VRFNonce` = 100
- Epoch 102: ATX published (`NumUnits` = 3), used values `NumUnits` = 3, `VRFNonce` = 100

The smesher decreases its `NumUnits` in epoch 101 from 5 to 3 which becomes effective immediately in the epoch after.

Example 2 - smesher increases `numUnits`:

- Epoch 100: ATX published (`NumUnits` = 5, `VRFNonce` = 100)
- Epoch 101: ATX published (`NumUnits` = 8, `VRFNonce` = 333), used values `NumUnits` = 5, `VRFNonce` = 100
- Epoch 102: ATX published (`NumUnits` = 8), used values `NumUnits` = 5, `VRFNonce` = 333
- Epoch 103: ATX published (`NumUnits` = 8), used values `NumUnits` = 8, `VRFNonce` = 333

The changed `VRFNonce` becomes effective immediately in the epoch after the ATX was published, but the increased `NumUnits` value becomes effective one epoch later. So that PoET can confirm the increase.

fasmat:
bors merge

PR #4009, opened 2023-01-25
[Merged by Bors] - Fix flaky nipost unit tests

poszu:
## Motivation
The test `TestNIPostBuilder_Close` is flaky. The code sometimes takes a different route through a `select` statement and doesn't return the proper error.

Also tests using the Poet integration harness sometimes fail w/o a clear explanation of why. The tests with harness are generally difficult to debug as the poet runs in a separate process and its output is not propagated correctly.

## Changes
- fixed error propagation from an error group in nipost builder,
- brought the removed ctx.Err() check after challenges are submitted back,
- removed uses of poet integration harnesses. Now the Poet is spawned in the same process as the test using its public API

Not using the harness is the first step to parallelizing tests using poet. The next one is to use a dynamically allocated port (port 9), but this requires a change on the poet side.


dshulyak (approved):

poszu:
bors merge

Issue #4010, opened 2023-01-25
Design Proposal: remove pub/sub pattern from internal clock


fasmat:
## Description
go-spacemesh has an internal clock that is used for multiple purposes, but most importantly to signal to other parts of the codebase when layers start and end and when a certain layer has been reached.

The problem with the current implementation is the `Subscribe` / `Unsubscribe` methods of the clock. These cause avoidable issues if instead the code would only use `AwaitLayer` and `GetCurrentLayer`.

### Affected code
- `TimeClock`
- `Ticker`
- `LayerConverter`
- and possible other classes / interfaces in the `timesync` package

### Issues with the current approach

The current approach has a few drawbacks:
- The Pub / Sub design causes the clock to take longer to signal a layer change the more listeners subscribe to this signal. This can cause performance issues if many parts of the code subscribe to the clock or worse: do not unsubscribe if they are not interested in the signal any more.
- If a listener subscribes and then isn't immediately ready to read the update via the channel it received from the subscription the signal it will cause the clock to skip sending the signal. This will cause the subscriber to wait reading on a channel that is never written to or closed leading to a deadlock.

### Proposal

go-spacemesh needs the following features across its codebase that are time-related:

- `CurrentLayer` this method needs to return the current layer of the network based on the system time. This might in some cases be not monotonic (i.e. when the system time was changed via NTP).
- `AwaitLayer` this method returns a channel that is closed when the layer in question is reached. This again uses wall time (the system time of the computer) to determine which is the current layer. The returned channel is closed when the layer has been reached OR surpassed (e.g. the system was hibernating for multiple layers - `AwaitLayer` on all those layers will signal at the same time when the system wakes up). Code that should be executed in a specific layer and uses `AwaitLayer` should check again with `CurrentLayer` if the layer in question might have passed already.

### Advantages and disadvantages with the proposed approach

- `AwaitLayer` scales to an arbitrary number of listeners without causing performance issues: Only one `close` operation per layer is performed by the clock, independently of how many listeners await it.
- `AwaitLayer` signals exactly once to all listeners, when the layer is reached or has been passed, The signal can be processed by listeners immediately or later without affecting the clock. `Subscribe` on the other hand tries to signal to multiple listeners at the same time. In the case of time jumps possibly multiple layers are signalled to multiple listeners. The clock has to skip signalling if a listener doesn't react to the signal immediately or it gets stuck, causing "missed tick" errors.
[...]

PR #4012, opened 2023-01-26
[Merged by Bors] - Use AwaitLayer instead of subscription for proposalbuilder

fasmat:
## Motivation
Part of #4010. Refactor code to not subscribe to layer changes but rather await new layers.

## Changes
Use `AwaitLayer` in `ProposalBuilder` instead of a layer timer and make sure that the correct layer is processed by `ProposalBuilder`:
- `AwaitLayer` might return late (e.g. if the node was hibernating while it was waiting for the next layer) in this case `ProposalBuilder` will continue with the current layer rather than the one it was awaiting
- `GetCurrentLayer` might return a layer lower than the one awaited (if a time sync happens right between awaiting and getting the current layer). In this case wait a little longer.

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/miner/proposal_builder.go
+++ b/miner/proposal_builder.go
@@ -403,14 +403,21 @@ func (pb *ProposalBuilder) handleLayer(ctx context.Context, layerID types.LayerI
 }
 
 func (pb *ProposalBuilder) createProposalLoop(ctx context.Context) {
+	current := pb.clock.GetCurrentLayer()
+	next := current.Add(1)
 	for {
-		next := pb.clock.GetCurrentLayer().Add(1)
 		select {
 		case <-pb.ctx.Done():
 			return
 		case <-pb.clock.AwaitLayer(next):
+			current = pb.clock.GetCurrentLayer()
+			if current.Before(next) {
+				pb.logger.Info("proposal creation triggered early for layer %v, wait again", next)
+				continue
+			}
+			next = current.Add(1)
 			lyrCtx := log.WithNewSessionID(ctx)
-			_ = pb.handleLayer(lyrCtx, next)
+			_ = pb.handleLayer(lyrCtx, current)
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

miner/proposal_builder.go
```diff
@@ -405,13 +404,13 @@ func (pb *ProposalBuilder) handleLayer(ctx context.Context, layerID types.LayerI
...
+		next := pb.clock.GetCurrentLayer().Add(1)
 		select {
 		case <-pb.ctx.Done():
 			return
-
-		case layerID := <-pb.layerTimer:
+		case <-pb.clock.AwaitLayer(next):
 			lyrCtx := log.WithNewSessionID(ctx)
-			_ = pb.handleLayer(lyrCtx, layerID)
+			_ = pb.handleLayer(lyrCtx, next)
```
poszu: Is `pb.handleLayer(lyrCtx, next)` idempotent? What if it is called with same `next` many times?

My understanding is that in this loop, `pb.clock.GetCurrentLayer().Add(1)`  can return the same value many times (if time is changed), right?

fasmat: Yeah, I added checks to make sure the same layer isn't processed twice.

I am not 100% sure how the layer subscription would have behaved in this case but this is how it behaves now:

- `pb.handleLayer` is never called twice on the same layer
- jumps backwards in time just cause the loop to wait again for the correct layer
- jumps forwards in time skip layers

This should correctly handle small time deviations backwards and any time deviations forwards. Large time deviations backwards (e.g. moving 100 layers back) will cause the `ProposalBuilder` to stop working until the layer it is expecting has arrived.

This might be not totally correct, but I think would mimic the behaviour as it was before with `Subscribe` when time moved backwards. Time moving forwards here would now be handled correct imo (when before it would cause the loop to trigger on every missed layer).

poszu (approved):
LGTM

noamnelke (approved):
🚢 it

noamnelke:
This is relevant here, too.
https://github.com/spacemeshos/go-spacemesh/pull/4013#discussion_r1089899203

fasmat:
bors merge

fasmat:
bors merge

PR #4013, opened 2023-01-26
[Merged by Bors] - Use AwaitLayer instead of subscription in beacon

fasmat:
## Motivation
Part of https://github.com/spacemeshos/go-spacemesh/issues/4010. Refactor code to not subscribe to layer changes but rather await new layers.

## Changes
Use `AwaitLayer` in `Beacon` instead of a layer timer and make sure that the correct layer is processed by `ProtocolDriver`:
- `AwaitLayer` might return late (e.g. if the node was hibernating while it was waiting for the next epoch) in this case `ProtocolDriver` will continue with the current epoch rather than the one it was awaiting.
- `GetCurrentLayer` might return a layer lower than the one awaited (if a time sync happens right between awaiting and getting the current layer). In this case wait a little longer.
- `listenEpochs` now fires once per epoch instead of once per layer and then checking if the layer is the first in an epoch. This should make the code a bit more efficient.

## Test Plan
n/a

## TODO

## DevOps Notes


beacon/beacon.go
```diff
@@ -530,24 +527,29 @@ func (pd *ProtocolDriver) cleanupEpoch(epoch types.EpochID) {
...
-				pd.eg.Go(func() error {
-					_ = pd.onNewEpoch(ctx, epoch)
-					return nil
-				})
+		case <-pd.clock.AwaitLayer(layer):
+			current := pd.clock.GetCurrentLayer()
+			if current.Before(layer) {
+				pd.logger.With().Info("time sync detected, realigning Beacon")
+				continue
 			}
```
noamnelke, line 543: I slightly dislike this check. I feel that it's arbitrary.

In theory, time sync can happen anywhere, so why do we only check for it here? Why not after every row in this function? My point is that just like we're willing to tolerate that this happens right after we check, we should be willing to tolerate that it happens between closing the channel on the timer side and waking up here.

OTOH, there's a different check missing, IMO. If the computer sleeps and wakes up after a while, we might be a long time after the start of the layer (epoch, in this case). I'm guessing there should be a timeout, after which we don't send the proposal because it's too late and everyone else already moved on.

noamnelke (approved):

beacon/beacon.go
fasmat: I had the following train of thought when I wrote that code:

1. `AwaitLayer` waits until *at least* `layer`, but possibly longer (e.g. hibernation, etc.) so when it returns we have to check in which layer we are actually in.
2. `GetCurrentLayer` returns the current layer which is expected to be the one we waited for, but could be a later one, thats why we use `current` and not `layer` to avoid processing an already passed layer/epoch.
3. In *rare* cases `current` is before `layer` (when exactly between `AwaitLayer` and `GetCurrentLayer` the time was adjusted backwards via NTP or by the user). In this case we must prevent beacon to process a layer/epoch that it possibly already handled.

This `if` should on most systems rarely if ever trigger but if it does it prevents a node from being flagged as malicious for broadcasting 2 messages in a layer/epoch where it should only have broadcasted one.

beacon/beacon.go
```diff
@@ -530,24 +527,29 @@ func (pd *ProtocolDriver) cleanupEpoch(epoch types.EpochID) {
...
+			layer = current.GetEpoch().Add(1).FirstLayer()
+			epoch := current.GetEpoch()
```
countvonzero, line 547: current may not be the first layer of this epoch in the case of delayed tick. i don't think the node should participate in the beacon protocol when it's late.

fasmat: I updated the code to skip the epoch if the tick is late.

beacon/beacon.go
fasmat: I added a check to verify we are in the first layer of an epoch. If we aren't then the epoch will be skipped by the beacon. This should mimic the behaviour as it was before the changes.

fasmat:
bors try

fasmat:
bors merge

fasmat:
bors merge

PR #4014, opened 2023-01-26
[Merged by Bors] - Use AwaitLayer instead of subscribing to layer updates for status updates

fasmat:
## Motivation
Part of https://github.com/spacemeshos/go-spacemesh/issues/4010. Refactor code to not subscribe to layer changes but rather await new layers.

## Changes
Use `AwaitLayer` in `SubscribeToLayers` instead of a layer timer and make sure that the correct layer is reported / logged. Before if the node was hibernating or busy for some time it might report on multiple layers in a short period. Now it will skip reporting status updates to the current layer to avoid spamming clients with (outdated) updates.

## Test Plan
n/a

## TODO

## DevOps Notes


events/reporter.go
```diff
@@ -301,28 +301,37 @@ func SubscribeRewards() Subscription {
...
+		for {
+			mu.RLock()
+			stopChan := reporter.stopChan
+			mu.RUnlock()
+
+			select {
+			case <-ticker.AwaitLayer(next):
+				current := ticker.GetCurrentLayer()
+				if current.Before(next) {
+					log.Info("time sync detected, realigning ProposalBuilder")
```
noamnelke, line 324: This applies here, too.
https://github.com/spacemeshos/go-spacemesh/pull/4013#discussion_r1089899203

noamnelke (approved):

events/reporter.go
fasmat: I had the following train of thought when I wrote that code:

1. `AwaitLayer` waits until at least `next`, but possibly longer (e.g. hibernation, etc.) so when it returns we have to check in which layer we are actually in.
2. `GetCurrentLayer` returns the current layer which is expected to be the one we waited for, but could be a later one, thats why we use `current` and not `layer` to avoid reporting an already passed layer/epoch.
3. In rare cases `current` is before `layer` (when exactly between `AwaitLayer` and `GetCurrentLayer` the time was adjusted backwards via NTP or by the user). In this case we must prevent reporting the same layer again.

fasmat:
bors merge

PR #4017, opened 2023-01-27
[Merged by Bors] - Use a single channel to signal if ATXs are synced

fasmat:
## Motivation
The current implementation of `RegisterForATXSynced` uses one channel per listener.

The code can be simplified to only use one channel for all listeners.

## Changes
- Instead of one channel per listener use a single channel
- Remove unneeded mutex

## Test Plan
N/A

## TODO

## DevOps Notes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -133,7 +133,6 @@ type Syncer struct {
 
 	// awaitATXSyncedCh is the list of subscribers' channels to notify when this node enters ATX synced state
 	awaitATXSyncedCh chan struct{}
-	muATXSyncedCh    sync.RWMutex // protects the channel above from concurrent access
 
 	eg errgroup.Group
 
@@ -195,13 +194,6 @@ func (s *Syncer) Close() {
 
 // RegisterForATXSynced returns a channel for notification when the node enters ATX synced state.
 func (s *Syncer) RegisterForATXSynced() chan struct{} {
-	if s.ListenToATXGossip() {
-		ch := make(chan struct{})
-		close(ch)
-		return ch
-	}
-	s.muATXSyncedCh.RLock()
-	defer s.muATXSyncedCh.RUnlock()
 	return s.awaitATXSyncedCh
 }
 
@@ -268,12 +260,11 @@ func (s *Syncer) Start(ctx context.Context) {
 
 func (s *Syncer) setATXSynced() {
 	s.atxSyncState.Store(synced)
-
-	s.muATXSyncedCh.Lock()
-	defer s.muATXSyncedCh.Unlock()
-
-	close(s.awaitATXSyncedCh)
-	s.awaitATXSyncedCh = make(chan struct{})
+	select {
+	case <-s.awaitATXSyncedCh:
+	default:
+		close(s.awaitATXSyncedCh)
+	}
 }
 
 func (s *Syncer) getATXSyncState() syncState {
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -132,8 +132,8 @@ type Syncer struct {
 	lastATXsSynced    atomic.Value
 
 	// awaitATXSyncedCh is the list of subscribers' channels to notify when this node enters ATX synced state
-	awaitATXSyncedCh []chan struct{}
-	muATXSyncedCh    sync.Mutex // protects the slice above from concurrent access
+	awaitATXSyncedCh chan struct{}
+	muATXSyncedCh    sync.RWMutex // protects the slice above from concurrent access
 
 	eg errgroup.Group
 
@@ -161,7 +161,7 @@ func NewSyncer(
 		mesh:             mesh,
 		certHandler:      ch,
 		patrol:           patrol,
-		awaitATXSyncedCh: make([]chan struct{}, 0),
+		awaitATXSyncedCh: make(chan struct{}),
 	}
 	for _, opt := range opts {
 		opt(s)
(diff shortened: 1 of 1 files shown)


fasmat:
bors try

fasmat:
bors try

syncer/syncer.go
```diff
@@ -132,8 +132,8 @@ type Syncer struct {
 	lastATXsSynced    atomic.Value
 
 	// awaitATXSyncedCh is the list of subscribers' channels to notify when this node enters ATX synced state
-	awaitATXSyncedCh []chan struct{}
-	muATXSyncedCh    sync.Mutex // protects the slice above from concurrent access
+	awaitATXSyncedCh chan struct{}
+	muATXSyncedCh    sync.RWMutex // protects the channel above from concurrent access
```
countvonzero: are you observing race condition in this code?

first of all we don't need a RWMutex here. i prefer Mutex when there is not race condition. it's faster and and simpler to use.
secondly, if you are using single channel, why not initialize it the constructor and there is no need to protect it.

countvonzero (commented):
this PR surprises me. it's basically constant time. what is the problem here?

fasmat:
> this PR surprises me. it's basically constant time. what is the problem here?

Neither `RegisterForATXSynced`, nor `setATXSynced` are constant time.

Register has to append a channel to a slice which requires the runtime to create a new backing array copy over all existing channels and then add a new one.

setsynced has to iterate over a slice and close all channels which takes longer the more clients registered for the signal.

With the changes all calls to register return the same channel (or a closed one if already synced) and setsynced only ever needs to close one channel.

Before it was linear now it is constant.

countvonzero:
> Register has to append a channel to a slice which requires the runtime to create a new backing array copy over all existing channels and then add a new one.

but this should only happens the first time it's created. the default size is way bigger than the expected registrants.
 
> setsynced has to iterate over a slice and close all channels which takes longer the more clients registered for the signal.

yes. but we are talking about less than 5 clients here? to me that's constant time.

syncer/syncer.go
fasmat: The RWMutex allows for a more fine-grained access restriction to the field that holds the channel.

Since RegisterATXSynced only reads the field it can use a `RLock()` allowing multiple concurrent calls without mutual exclusion. Only when `setAtxSynced()` is called all access to the field needs to be blocked until the lock is freed again.

countvonzero:
note that i dont object to this PR. just surprised that it drew your attention.

syncer/syncer.go
countvonzero: whether we agreed on whether RWMutex is necessary is beyond the point.
we don't need a lock here. if the channel is created and never changed within syncer, why need lock to protect it?

fasmat:
I am initialing the channel in the constructor. But a channel can only be closed once!

`setAtxSynced` creates a new channel and replaces the existing one with it. This resets the state to "not synced" when `ListenForGossip` returns false

This is exactly the same behavior as before.

countvonzero:
> I am initialing the channel in the constructor. But a channel can only be closed once!
> 
> `setAtxSynced` creates a new channel and replaces the existing one with it. This resets the state to "not synced" when `ListenForGossip` returns false
> 
> This is exactly the same behavior as before.

what i am saying is a node can only ever wait for atx synced once in its life time. there is no need to create another channel in setAtxSynced. simply close it once there.

fasmat:
> what i am saying is a node can only ever wait for atx synced once in its life time. there is no need to create another channel in setAtxSynced. simply close it once there.

I had the same thought 🙂 and actually implemented it like you suggested in an earlier commit (just go back one or two commits in this PR)

That caused system tests to fail with `bors try` so I changed the behavior back to how it was before. Are you sure that the ATX synched state is reached and never goes back to not synched?

fasmat:
> > Register has to append a channel to a slice which requires the runtime to create a new backing array copy over all existing channels and then add a new one.
> 
> but this should only happens the first time it's created. the default size is way bigger than the expected registrants.

It starts with size 0 and appends one element on each call. How the runtime resizes a slice is not specified but the current behavior is to double for small slices and reduce the factor to 1.25 for bigger slices:
 https://forum.golangbridge.org/t/slice-append-what-is-the-actual-resize-factor/15654

Meaning if the function is called 10 calls are made it has to be resized 5 times.

> > setsynced has to iterate over a slice and close all channels which takes longer the more clients registered for the signal.
> 
> yes. but we are talking about less than 5 clients here? to me that's constant time.

I didn't check how many calls are made in our codebase but I think it's a bit more than that 🙂

syncer/syncer.go
fasmat: It is changed; `setATXsynced` assigns the field a new value. Without a mutex I'm getting warnings from the race detector.

I just changed the Mutex to an RWMutex that does the same thing but allows multiple concurrent reads.

The field needs to be protected by a Mutex and RWMutex just allows for a more fine grained access control.

countvonzero:
> It starts with size 0 and appends one element on each call.

you are right. 

> I didn't check how many calls are made in our codebase but I think it's a bit more than that

it's exactly in 4 places. and realistically only once (buildNIPostChallenge) in the lifetime of a node, because once ATX synced, we simply return a closed channel.

> Are you sure that the ATX synched state is reached and never goes back to not synched?

you can double-check the code. but the code only has code to set it to `synced`. no code path set it to otherwise.

fasmat:
bors try

fasmat:
bors try

fasmat:
@countvonzero looks like you were right. The earlier failed CI run was probably caused by some flakiness and not the fact that ATX synced state is only reached once.

I updated the code with your suggestions

countvonzero (approved):
i would remove the claim of performance improvement from the PR description.

fasmat:
bors merge

PR #4019, opened 2023-01-27
[Merged by Bors] - prevents syncer races with block generation

countvonzero:
## Motivation

layer patrol was in place to prevent races between syncer and hare to advance the mesh.
but since https://github.com/spacemeshos/go-spacemesh/issues/3297 is implemented, 
the race period extended to block generation, as blocks are now generated serially in layer order.

this is one of the reasons why systest is flaky, as when the layer N is processed in the mesh, syncer
already saved the certificate sync'ed from peers for that layer, causing the node to think there were
multiple certificates for the same layer, which then apply empty block for that layer.

## Changes

- move layer patrol to block generator so it can properly semaphore btwn syncer and block generation
- in case processing layer was delayed by more than patrol's wait limit, make sure mesh doesn't trump the certificate that's already saved

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -591,9 +591,14 @@ func (msh *Mesh) saveHareOutput(ctx context.Context, logger log.Log, lid types.L
 	}); err != nil {
 		return err
 	}
-	if len(certs) == 0 {
+	switch len(certs) {
+	case 0:
 		msh.trtl.OnHareOutput(lid, bid)
-	} else if len(certs) > 1 {
+	case 1:
+		logger.With().Info("already synced certificate",
+			log.Stringer("cert_block_id", certs[0].Block),
+			log.Bool("cert_valid", certs[0].Valid))
+	default: // more than 1 cert
 		logger.With().Warning("multiple valid certs found in network",
 			log.Object("certificates", log.ObjectMarshallerFunc(func(encoder zapcore.ObjectEncoder) error {
 				for _, cert := range certs {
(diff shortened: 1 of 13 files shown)


mesh/mesh.go
```diff
@@ -550,6 +551,65 @@ func (msh *Mesh) revertState(ctx context.Context, logger log.Log, revertTo types
...
+					encoder.AddString("block_id", cert.Block.String())
+					encoder.AddBool("valid", cert.Valid)
+				}
+				return nil
+			})))
+	} else {
+		logger.With().Info("already synced certificate",
+			log.Stringer("cert_block_id", certs[0].Block),
+			log.Bool("cert_valid", certs[0].Valid))
+	}
```
fasmat: NIT: i think switch would help with readability here

```suggestion
	switch len(certs) {
	case 0:
		msh.trtl.OnHareOutput(lid, bid)
	case 1:
		logger.With().Info("already synced certificate",
			log.Stringer("cert_block_id", certs[0].Block),
			log.Bool("cert_valid", certs[0].Valid))
	default: // more than 1 cert
		logger.With().Warning("multiple valid certs found in network",
			log.Object("certificates", log.ObjectMarshallerFunc(func(encoder zapcore.ObjectEncoder) error {
				for _, cert := range certs {
					encoder.AddString("block_id", cert.Block.String())
					encoder.AddBool("valid", cert.Valid)
				}
				return nil
			})))
	}
```

fasmat (approved):

countvonzero:
bors merge

PR #4023, opened 2023-01-29
[Merged by Bors] - block certificate: allow wider layer range of certify message

countvonzero:
## Motivation

https://github.com/spacemeshos/go-spacemesh/issues/3297 enforces blocks to be executed in layer order.
but the original window to accept certify message gossip is too narrow when nodes make different progress.
this is causing the nodes in the systest to reject certify messages and failed to generate block certificate.

## Changes

- extend the acceptance window to 2 * `zdist` where `zdist` is the number of layers tortoise will wait for hare to terminate for a given layer.
- prune ineffective messages after 2 * `zdist` layers has passed.


countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

PR #4025, opened 2023-01-30
[Merged by Bors] - Remove Subscription mechanism from internal clock

### LINKED ISSUE #4010: Design Proposal: remove pub/sub pattern from internal clock

fasmat:
## Motivation
Closes #4010, this is a follow up to #4012, #4013, #4014 and removes the now obsolete pub/sub mechanism of the internal clock.

## Changes
The core of the change is the `timesync` package, the rest of the PR is adapting the existing code to the new interface of the clock. `AwaitLayer` and `Subscribe` have the same purpose, but behave differently. This PR removes the duplicated code and simplifies the clock:

- `Subscribe`/`Unsubscribe` and all logic associated with these two methods has been removed
- `Ticker` and `TimeClock` have been merged into `NodeClock`
- Methods `GetGenesisTime` and `GetCurrentLayer` have been renamed to `GenesisTime` and `CurrentLayer`.
- The clock itself is now ticking once at the beginning of a layer and adjusts the waiting period until the next layer based on when the tick was issued.

## Test Plan
Tests for `NodeClock` have been written from scratch, with emphasis on safe concurrent use and stability against time shifts that might be caused by NTP or the node hibernating.

For this I added a dependency on http://github.com/benbjohnson/clock (was already an indirect dependency). It allows to mock time more easily in tests than it would be possible with `gomock`.

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

countvonzero (approved):

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
Bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
Bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors try

fasmat:
bors merge

fasmat:
bors try

fasmat:
bors merge-

fasmat:
bors try

fasmat:
bors try

fasmat:
bors merge

PR #4026, opened 2023-01-30
[Merged by Bors] - mesh: remove dependency on verified layer

### LINKED ISSUE #3744: mesh: block validity updater is not robust

countvonzero:
## Motivation

Closes #3744

## Changes

- mesh can determine which block to apply without knowing the latest verified layer with the following heuristic:
1. use validity from tortoise updates
2. if 1 fails, get block validity from database
3. if there are no valid blocks, apply empty block for the layer
4. if the block validity is not determined, use the hare output

- remove layer hash that captured all valid blocks in the layer. this field is unused. removing it so we don't need to add code to update layer hash when block validity changes.


countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #4036, opened 2023-02-01
[Merged by Bors] - Build go-spacemesh binaries for arm64

### LINKED ISSUE #4035: Provide arm64 linux and apple silicon builds for go-spacemesh
As in the title, we have self-hosted runners for that, so getting HW is not an issue.

brusherru:
## Motivation
It closes #4035 

## Changes
Added two more build targets and tweaked GHA workflow a bit.

## How to test
For testing purposes I've pushed tag [`v0.0.0-test-arm64`](https://github.com/spacemeshos/go-spacemesh/releases/tag/v0.0.0-test-arm64) — there are links to archives.
1. Download the archive corresponding to your os and arch
2. Unzip it
3. Go to extracted dir
4. Set chmod (on Linux & macOS): `chmod 755 ./*`
5. On MacOS turn off quarantine: `sudo xattr -rd com.apple.quarantine ./*`
6. Run go-spacemesh as usually (aka `./go-spacemesh -d ~/node-data -c ~/config.json`)
7. Ensure it has a full list of PoS providers. Call:
    ```
    grpcurl -d "{\"benchmark\": true}" -plaintext localhost:9092 spacemesh.v1.SmesherService.PostSetupComputeProviders
    ```
    The list depends on your OS, but hopefully, you'll see some GPU beside the CPU. :)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,8 +15,21 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        include:
+          - os: macos-latest
+          - os: ubuntu-latest
+          - os: windows-latest
+          - os: [self-hosted, macos, arm64]
+          - os: [self-hosted, linux, arm64]
     steps:
+      - shell: bash
+        run: |
+          if [[ ${{ runner.arch }} == "ARM64" ]]; then
+            echo "OUTNAME=${{ runner.os }}_${{ runner.arch }}" >> $GITHUB_ENV
+          else
+            echo "OUTNAME=${{ runner.os }}" >> $GITHUB_ENV
+          fi
+
       - name: Setup Go
         uses: actions/setup-go@v3
         with:
@@ -54,6 +67,8 @@ jobs:
       - name: Get gpu-post release
         shell: bash
         run: |
+          echo "OUTNAME = ${{ env.OUTNAME }} \n"
+
           mkdir build_${{ runner.os }}
           unzip libgpu-setup-*.zip -d build_${{ runner.os }}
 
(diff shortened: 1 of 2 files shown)


brusherru:
bors try

version.txt
```diff
@@ -1 +1 @@
-v0.0.0-unreleased
\ No newline at end of file
+v0.0.0-test-arm64
```
fasmat, line 1: Is this change needed or can it be reverted?

fasmat (approved):

fasmat (commented):
Generally looks fine to me, but I'm not sure if the two self hosted runners are intended for other things? @pigmej @andres-spacemesh

version.txt
brusherru: I believe it SHOULD be reverted :) I'll do it

brusherru: Done ;)

brusherru:
Reverted the latest commit which was made to test the workflow.
However, I keep the tag for a while, so you can still download artifacts and check them out.

I've tested both new macOS and Linux artifacts (inside VM). Both working well.

image

andres-spacemesh (approved):
I run the pipeline with https://github.com/nektos/act and dependencies looks ok on arm64

pigmej (approved):

pigmej (commented):
We somehow include smrepl for darwin_amd64 in darwin_arm64 release.

I think we shall in separate pr remove smrepl at all from the release process.

.github/workflows/release.yml
```diff
@@ -106,6 +121,8 @@ jobs:
             ## Zip Files
             - Windows: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/Windows.zip
             - macOS: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/macOS.zip
+            - macOS: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/macOS_ARM64.zip
```
pigmej: we could add here that this is arm64

.github/workflows/release.yml
```diff
@@ -106,6 +121,8 @@ jobs:
...
             - Linux: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/Linux.zip
+            - Linux: https://storage.googleapis.com/${{ secrets.GCP_BUCKET }}/${{ steps.version.outputs.content }}/Linux_ARM64.zip
```
pigmej: we could add here that this is arm64

brusherru:
Ok, got rid of `smrepl` and tweak the release message a bit.
So I've pushed the "test" commit and tag one more time, after a successful build I'll revert it.

brusherru:
bors merge

PR #4037, opened 2023-02-02
[Merged by Bors] - atx: add received time in atx and atx header

countvonzero:
## Motivation

preparing for implementations that require atx timing
- beacon
- hare active set

## Changes

add received time as a field in atx and atx header


common/types/activation.go
```diff
@@ -385,6 +386,14 @@ func (atx *ActivationTx) SetEffectiveNumUnits(numUnits uint32) {
 	atx.effectiveNumUnits = numUnits
 }
 
+func (atx *ActivationTx) SetReceived(received time.Time) {
+	atx.received = received
+}
+
+func (atx *ActivationTx) Received() time.Time {
```
fasmat, line 393: I assume `received` has to be set for an ATX to be valid? Should we add a check in `Verify` or add it as a parameter to `types.NewActivationTx` so it is always set?

sql/atxs/atxs_test.go
```diff
@@ -24,6 +24,12 @@ func TestMain(m *testing.M) {
 	os.Exit(res)
 }
 
+func checkAtxEqual(t *testing.T, this, that *types.VerifiedActivationTx) {
+	require.Equal(t, this.Received().UnixNano(), that.Received().UnixNano())
+	that.SetReceived(this.Received())
+	require.Equal(t, this, that)
+}
```
fasmat: Is this needed because of different time zones in the time stamp? If yes imo we should parse it with `time.Local` when reading from DB. The internal clock of the node also uses `time.Local`.

common/types/activation.go
countvonzero: no. received time is determined by the receiving node, not the generating node. this is used to determine whether the receiving node will consider such ATX eligible in the node's view of active set, or in the beacon case, eligible for making beacon proposals.

sql/atxs/atxs_test.go
countvonzero: yes. it's due to the time zone. 
if we special case this one (parsing it to time.Local), that means we need to do that for all the time output by clock. specifically LayerToTime(). and i think that's too big a change for the purpose. 

only during testing do we compare two ATXs to be equal.

countvonzero: > The internal clock of the node also uses time.Local.

right. i was wrong. ok. i will parse it to local time.

common/types/activation.go
fasmat: I understand. I mean that when we persist an ATX we should check if `received` is set (by the receiving node) to avoid storing an ATX with the received time `January 1, year 1, 00:00:00.000000000 UTC`.

`Verify` would be imo be a good place to do that, but we could also do it in the `atx.Add` function.

fasmat (approved):

common/types/activation.go
countvonzero: missed this earlier. done. thanks

common/types/activation.go
```diff
@@ -400,6 +409,9 @@ func (atx *ActivationTx) Verify(baseTickHeight, tickCount uint64) (*VerifiedActi
 	if atx.effectiveNumUnits == 0 {
 		return nil, fmt.Errorf("effective num units not set")
 	}
+	if atx.received == (time.Time{}) {
```
fasmat: NIT

```suggestion
	if atx.received.IsZero() {
```

fasmat (approved):

countvonzero:
bors merge

countvonzero:
bors merge

PR #4045, opened 2023-02-06
[Merged by Bors] - systests: make bootnodes non-smeshing

### LINKED ISSUE #4042: systests: change bootnodes to non-smeshing
## Description

mainnet bootnodes will not be smeshing. systest bootnodes should do the same.

countvonzero:
## Motivation

Closes #4042 

## Changes

- beacon
  - allow non-smeshers to be passive participants in the beacon protocol
- tune some logs for non-smeshing nodes (error->info/no log)
- activation
  - after atx is published, wait for network propagation delta before building the next challenge.
- hare
  - add session id for the same consensus process for debugging
  - terminate hare right away when iteration limit is reached
  - check beacon after waking up from delta to improve chance of participating in hare
- fetch
  - register poet proof hash with the peer that gossiped the atx
- systests
  - set smeshing-start to false for boot nodes

debugging note:
partition tests became flaky with the change because now that bootnodes are not smeshing, they are not directly getting proof from poet servers and instead rely on peer fetching mechanism when the proof is referenced in the atx.
however, poet proof fetching was not optimized to fetch from source of the referencing atx. instead, it's fetching from the other sides of the partition (randomly) and failed. this happens in the first layer of the epoch after the partition starts.

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -373,44 +373,54 @@ func (b *Builder) loop(ctx context.Context) {
 		}
 
 		ctx := log.WithNewSessionID(ctx)
-		if err := b.PublishActivationTx(ctx); err != nil {
-			if errors.Is(err, context.Canceled) {
+		err := b.PublishActivationTx(ctx)
+		if err == nil {
+			b.log.WithContext(ctx).With().Info("waiting for atx to propagate before building the next challenge",
+				log.Duration("wait", b.poetRetryInterval))
+			select {
+			case <-ctx.Done():
 				return
+			case <-time.After(b.poetRetryInterval):
 			}
+			continue
+		}
 
-			b.log.WithContext(ctx).With().Error("error attempting to publish atx",
-				b.layerClock.CurrentLayer(),
-				b.currentEpoch(),
-				log.Err(err),
-			)
+		if errors.Is(err, context.Canceled) {
+			return
+		}
 
-			switch {
-			case errors.Is(err, ErrATXChallengeExpired):
-				b.log.WithContext(ctx).Debug("retrying with new challenge after waiting for a layer")
-				b.discardChallenge()
-				// give node some time to sync in case selecting the positioning ATX caused the challenge to expire
-				currentLayer := b.layerClock.CurrentLayer()
-				select {
-				case <-ctx.Done():
-					return
-				case <-b.layerClock.AwaitLayer(currentLayer.Add(1)):
-				}
-			case errors.Is(err, ErrPoetServiceUnstable):
-				b.log.WithContext(ctx).Debug("retrying after poet retry interval")
-				select {
-				case <-ctx.Done():
-					return
-				case <-time.After(b.poetRetryInterval):
-				}
-			default:
-				b.log.WithContext(ctx).With().Warning("unknown error", log.Err(err))
-				// other failures are related to in-process software. we may as well panic here
-				currentLayer := b.layerClock.CurrentLayer()
-				select {
-				case <-ctx.Done():
-					return
-				case <-b.layerClock.AwaitLayer(currentLayer.Add(1)):
-				}
+		b.log.WithContext(ctx).With().Error("error attempting to publish atx",
+			b.layerClock.CurrentLayer(),
+			b.currentEpoch(),
+			log.Err(err),
+		)
+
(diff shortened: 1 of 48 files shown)


countvonzero:
bors try

dshulyak:
> bootnodes may not be able to run hare and approve proposals at the 1st layer of the epoch as it may not have beacon

why? there should be no such limitation simply because bootnode doesn't smesh

countvonzero:
> > bootnodes may not be able to run hare and approve proposals at the 1st layer of the epoch as it may not have beacon
> 
> why? there should be no such limitation simply because bootnode doesn't smesh

because they don't participate in the beacon protocol and thus rely on beacon sync at the beginning of the epoch.
so in the first layer of the epoch, they may not participate in hare (active or passive) due to lack of beacon.
thats why i also reduced beacon sync threshold, and move beacon check after the wakeup delta.

dshulyak:
> because they don't participate in the beacon protocol and thus rely on beacon sync at the beginning of the epoch.

this is not correct. they should be able to listen passively

countvonzero:
> > because they don't participate in the beacon protocol and thus rely on beacon sync at the beginning of the epoch.
> 
> this is not correct. they should be able to listen passively

but how? they cannot verify any hare messages.

dshulyak:
> but how? they cannot verify any hare messages.

what it has to do with hare? you said that it can't participate in beacon

non-smeshing nodes should be capable of verifying all hare and beacon protocol messages, and see that nodes reached consensus. the only thing that they can't do is to participate in consensus (like proposals, voting), as they don't have any weight in the network

moreover this is also **very** important condition for network to work, as the ability to verify a message is a prerequisite for propagating it futher

perhaps something was introduced in the latest changes that prevent them from doing so, in such case thats a bug

countvonzero:
> what it has to do with hare? you said that it can't participate in beacon

all hare messages comes with hare eligibility, which is the vrf signature of the vrf message.
all vrf messages contain beacon. therefore, a node without beacon value cannot verify hare messages.

countvonzero:
> non-smeshing nodes should be capable of verifying all hare and beacon protocol messages

this is also not true for beacon protocol at the moment. but can be fixed.

dshulyak:
> all hare messages comes with hare eligibility, which is the vrf signature of the vrf message.
all vrf messages contain beacon. therefore, a node without beacon value cannot verify hare messages.

ok i see what you mean. the problem is that bootnodes should be able to listen to a beacon protocol and compute beacon from observed messages without beacon sync

dshulyak:
> this is also not true for beacon protocol at the moment. but can be fixed.

so the consequence of that is much bigger than the hare not working in the couple of layers. in the network where majority of nodes are not smeshing - it will be very unlikely for beacon to make progress, as messages will be dropped unpredictably. for instance on a devnet's beacon won't work almost 100% of the time

countvonzero:
> in the network where majority of nodes are not smeshing - it will be very unlikely for beacon to make progress, as messages will be dropped unpredictably. for instance on a devnet's beacon won't work almost 100% of the time

the beacon majority rule is applied on the smeshers, not nodes.

nevertheless, it does reduce the message propagation. i will change so that nodes can listen in to beacon protocol and calculate beacon itself for non-smeshing nodes.

countvonzero:
bors try

countvonzero:
@dshulyak changed the non-smeshing nodes to participate in beacon protocol, and changed the systests back to treat bootnodes like smeshing nodes.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> ## try
> Build failed:
> 
> * [ci-status](https://github.com/spacemeshos/go-spacemesh/actions/runs/4132050384/jobs/7140471080)

systest passed

https://github.com/spacemeshos/go-spacemesh/actions/runs/4132050382

dshulyak (approved):

countvonzero:
bors try

countvonzero:
bors merge

PR #4047, opened 2023-02-06
[Merged by Bors] - Fix incorrect deadlock check in clock

fasmat:
## Motivation
The deadlock check in `AwaitLayer` incorrectly updates `t.firstAwaitedLayer` always to one before the awaited. This fixes the check and only updates the variable if a layer is awaited that otherwise would never be signalled.

## Changes
- Update the incorrect check in `AwaitLayer`
- Add more logs

## Test Plan
I added a test that would fail if the check for the possible deadlock in `AwaitLayer` wouldn't be there.

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

dshulyak (changes requested):
this doesn't need to be a parameter, please set any reasonable default value

config/presets/testnet.go
```diff
@@ -65,11 +66,11 @@ func testnet() config.Config {
 	conf.Beacon.FirstVotingRoundDuration = 3 * time.Minute
 	conf.Beacon.GracePeriodDuration = 10 * time.Second
 	conf.Beacon.ProposalDuration = 30 * time.Second
+	conf.Beacon.VotingRoundDuration = 50 * time.Second
```
dshulyak: why do you move stuff around?

dshulyak (changes requested):

config/presets/testnet.go
fasmat: I diffed `testnet.go` and `fastnet.go` and to check the differences and adapted `testnet.go` to the order of parameters in `fastnet.go`.

I noticed that `fastnet.go` is missing a few parameters that are set in `testnet.go`.

I can revert these changes if you want.

fasmat:
> this doesn't need to be a parameter, please set any reasonable default value

I had it at 1 second before, which seems reasonable for a clock. I'll revert most of the changes in the PR.

timesync/clock.go
```diff
@@ -116,16 +117,23 @@ func (t *NodeClock) tick() {
...
+		t.log.With().Info("clock ticked back in time",
+			log.Stringer("layer", layer),
+			log.Stringer("last_ticked_layer", t.firstAwaitedLayer),
+		)
+		return
+	}
 
 	// close await channel for prev layers
-	for l := t.lastTickedLayer; !l.After(layer); l = l.Add(1) {
+	for l := t.firstAwaitedLayer; !l.After(layer); l = l.Add(1) {
```
countvonzero: i think as a service, it should keep two variables, `lastTicked` and `minAwaited`.
`lastTicked` to detect skipping a tick
`minAwaited` to decide from which layer to signal channels.

fasmat: Re-added `lastTicked` and logs that are printed when the clock moves backwards in time or skips layers.

fasmat:
bors try

timesync/clock.go
```diff
@@ -116,16 +118,29 @@ func (t *NodeClock) tick() {
 	}
 
 	layer := t.TimeToLayer(t.clock.Now())
+	switch {
+	case layer.Before(t.lastTicked):
+		t.log.With().Info("clock ticked back in time",
+			log.Stringer("layer", layer),
+			log.Stringer("last_ticked_layer", t.lastTicked),
+		)
+	case layer.Difference(t.lastTicked) > 1:
```
countvonzero, line 139: do you want to simply add @dshulyak 's PR that add metrics here? seems more efficient. we are seeing in systest that node failed to log during high CPU. hoping metrics can survive that.

timesync/clock.go
```diff
@@ -116,16 +118,29 @@ func (t *NodeClock) tick() {
...
+		t.log.With().Warning("clock skipped layers",
```
countvonzero, line 140: on 2nd thought, maybe this should be an error?

countvonzero (approved):

timesync/clock.go
fasmat: I would say no, I was even thinking of just making it a `Info` instead of `Warning`.

This will regularly happen with nodes running on desktops or notebooks. Every time one of those hibernates and wakes up again we will skip layers. Imo a node should be able to handle this and not treat it as an unexpected event. Layer X might not be succeeded by Layer X+1 but by any Layer X+n.

timesync/clock.go
fasmat: Added metrics from https://github.com/spacemeshos/go-spacemesh/pull/4050

fasmat:
bors merge

dshulyak (approved):

fasmat:
bors merge

PR #4050, opened 2023-02-07
add a warning and metric to track problems with local clock

dshulyak:


dshulyak:
bors try

timesync/clock.go
```diff
@@ -117,6 +127,15 @@ func (t *NodeClock) tick() {
 
 	layer := t.TimeToLayer(t.clock.Now())
 
+	d := layer.Difference(t.lastTickedLayer)
```
countvonzero: do you know why lastTickedLayer is updated here?
https://github.com/spacemeshos/go-spacemesh/blob/d59cadce8a6c80721b2cbebd0a7f5872ec9b65e2/timesync/clock.go#L176
@fasmat

countvonzero (approved):

dshulyak:
bors try

timesync/clock.go
fasmat: Yes, to avoid deadlocks.

Lets say the last ticked layer is 100. The clock just ticked, everyone awaiting layer 100 was signalled that the layer started and now a new client also calls `AwaitLayer(100)`. Due to a time correction (by NTP, user, etc.) `AwaitLayer` determines we are in layer 99 and returns an open channel. If I wouldn't update `t.lastTickedLayer` that channel would never be closed.

`t.lastTickedLayer` is probably incorrectly named; it should be something like `t.earliestAwaitedLayer`. It doesn't keep track of which layer was ticked at last, but rather which layer is the earliest someone is currently awaiting.

timesync/clock.go
```diff
@@ -114,9 +124,21 @@ func (t *NodeClock) tick() {
...
+		if d > 1 {
+			t.log.Warning("tick is too late",
+				log.Stringer("current", layer),
+				log.Stringer("last", t.lastTickedLayer),
+			)
+		}
+		if d != 0 {
+			tickDistance.Observe(float64(d))
+		}
+	}
```
fasmat, line 141: I am confused by the log. If `layer` is before `t.lastTickedLayer` the clock moved backwards. The tick isn't to late (a tick is once every second) but rather the clock was 12:00:00 at some point, starting a new layer and then moved back to e.g. 11:59:00 and is now 2 layers prior the one it already was in.

The tick isn't to late, the clock just moved backwards in time.

EDIT: nvm I misread the code, it checks for skipped layers, my bad.

timesync/clock.go
fasmat: But now that you mention it, the check is incorrect. It should update `lastTickedLayer` if the layer awaited is before it not after. I fixed it in this PR: https://github.com/spacemeshos/go-spacemesh/pull/4047

dshulyak:
bors try

PR #4051, opened 2023-02-07
[Merged by Bors] - Wait between failing attempts to publish an ATX instead of retrying immediately

### LINKED ISSUE #4020: High CPU load by node when creating / publishing ATX fails
### Description
The ATX builder seems to be stuck in an endless loop if errors occur during building and publishing ATXs

### Steps to reproduce
In tests nodes sometimes end up using high CPU load while printing the following logs hundrets of times per second:

```
2023-01-27T20:48:51.564+0100	INFO	fbdc7.app.fbdc7.atxBuilder	challenge not loaded	{"node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "app", "node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "atxBuilder", "sessionId": "f1d7ab40-35e3-40cf-9d14-0b3789b6e988", "errmsg": "failed to load nipost challenge from DB: failed to get value: database: not found", "name": "atxBuilder"}
2023-01-27T20:48:51.564+0100	INFO	fbdc7.app.fbdc7.atxBuilder	building new atx challenge	{"node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "app", "node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "atxBuilder", "sessionId": "f1d7ab40-35e3-40cf-9d14-0b3789b6e988", "current_epoch": "2", "name": "atxBuilder"}
2023-01-27T20:48:51.565+0100	INFO	fbdc7.app.fbdc7.atxBuilder	atx challenge is ready	{"node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "app", "node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "atxBuilder", "sessionId": "f1d7ab40-35e3-40cf-9d14-0b3789b6e988", "current_epoch": 2, "publish_epoch": 2, "target_epoch": 3, "name": "atxBuilder"}
2023-01-27T20:48:51.565+0100	ERROR	fbdc7.app.fbdc7.atxBuilder	error attempting to publish atx	{"node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "app", "node_id": "fbdc703ddaa6ccc1d96303e01c3135c46925c8331f922ed1c75152a628b1b379", "module": "atxBuilder", "sessionId": "f1d7ab40-35e3-40cf-9d14-0b3789b6e988", "layer_id": 174, "epoch_id": 2, "errmsg": "create ATX: builder: atx expired: poet round has already started at 2023-01-27 18:30:00.498 +0100 CET (now: 2023-01-27 20:48:51.565028 +0100 CET m=+20881.013172626)", "name": "atxBuilder"} 
```

It looks like if publishing an ATX fails repeatedly the node enters an endless loop. Here in this specific case it was caused by the builder trying to submit a challenge to poet when the round for that challenge had already passed.

### Expected Behaviour

- The node waits (with some jittered or exponential backoff) before creating a new ATX or submitting the challenge to PoET to avoid high load on the CPU and on PoET.
- The new challenge should target the correct PoET round

fasmat:
## Motivation
Closes #4020 

## Changes
If creating an ATX fails due to an expired challenge the builder used to try immediately again. I added a wait until the next layer to give the node time to sync more ATXs in case the reason the challenge is outdated because the selected Positioning ATX is outdated and no new ATXs have been synced yet.

This will also fix the high CPU usage when PoST is initialized before genesis, because the node will wait to create a new ATX until layer 1.

## Test Plan
A test was added to check for the changed behaviour.

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
bors try

activation/activation.go
```diff
@@ -594,15 +605,8 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 	case <-atxReceived:
 		logger.With().Info("received atx in db", atx.ID())
 	case <-b.layerClock.AwaitLayer((atx.TargetEpoch() + 1).FirstLayer()):
-		select {
-		case <-atxReceived:
-			logger.With().Info("received atx in db (in the last moment)", atx.ID())
```
countvonzero, line 599: why is this removed? select can pick random case right?

activation/activation.go
```diff
@@ -720,7 +724,7 @@ func (b *Builder) GetPositioningAtxInfo() (types.ATXID, types.LayerID, error) {
 	if err != nil {
 		if errors.Is(err, sql.ErrNotFound) {
 			b.log.With().Info("using golden atx as positioning atx", b.goldenATXID)
-			return b.goldenATXID, types.LayerID{}, nil
+			return b.goldenATXID, types.NewLayerID(0), nil
```
countvonzero, line 727: what's the diff btwn types.LayerID{} and types.NewLayerID(0)?

countvonzero (commented):
will let @poszu review this as i am not very good at poet math around timing here.

activation/activation.go
```diff
@@ -386,9 +386,15 @@ func (b *Builder) loop(ctx context.Context) {
 
 			switch {
 			case errors.Is(err, ErrATXChallengeExpired):
-				b.log.WithContext(ctx).Debug("discarding challenge")
+				b.log.WithContext(ctx).Debug("retrying with new challenge after waiting for a layer")
```
countvonzero, line 389: does it make sense to wait for time-based instead of layer based? atx can arrive any time.

activation/activation.go
poszu: `select` will pick a random out of all ready cases. I am not sure if this is a problem. It depends if "receiving" ATX happening AFTER the target epoch is still OK. I think it's basically useless at this point, isn't it?

activation/activation.go
poszu: I think that it would be most robust to wait until we receive a positioning ATX with PubLayerID = current challenge's PositioninigATX.PubLayerID + 1. 

For this, the `AtxHandler` would need a new method like `AwaitAtxWithPublayerID(lid types.LayerID)`.

activation/activation.go
fasmat: `select` will pick the case that is satisfied. If more than one case is satisfied simultaneously it will pick a random case. Here the check was the following:

If the layer was reached it would check again if the ATX was received. This again has a chance to not be picked since `RegisterForATXSync` would always be satisfied. So if the ATX was received right when the layer switched it might still not be handled but instead marked as `ErrATXChallengeExpired`.

Essentially for this case to be selected the following has to happen:

1. ATX must be received within a very short time period (few ms) before `TargetEpoch()` ends
2. The `AwaitLayer` signal must be picked over the `atxReceived` by the `select` since both are satisfied (50:50 chance)
3. Then the next `select` must choose `atxReceived` when it also could always choose `RegisterATXSynced` (another 50:50 chance)

This whole code block is there to detect an ATX that is "in the last moment" with 25% chance. In the last 2 days not one system test contained logs that it happened: [Grafana](https://grafana.spacemesh.dev/explore?orgId=1&left=%7B%22datasource%22:%22loki-gke-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-gke-dev%22%7D,%22editorMode%22:%22builder%22,%22expr%22:%22%7Bcontainer%3D%5C%22smesher%5C%22%7D%20%7C%3D%20%60in%20the%20last%20moment%60%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D%7D)

It would even be impossible to write a test that check if this works, because every test that we would write would have a 50% chance of failing randomly.

activation/activation.go
fasmat: Nothing, it is the same. The difference is here it means layer 0, while below it means n/a.

activation/activation.go
fasmat: > I think that it would be most robust to wait until we receive a positioning ATX with PubLayerID = current challenge's PositioninigATX.PubLayerID + 1.

There might not be an ATX in that exact layer, imo it would need to be an ATX with `PubLayerID` >= `currentlayer.GetEpoch().FirstLayer()`. But wouldn't this be the same as `RegisterForATXSynced()`?

The reason this happens is because the node is not in sync with the network but believes it is. This is also why restarting the node fixes the issue. A proper fix imo would update `RegisterForATXSynced()` to not return a closed channel if the node is not in sync with the network, but this would be a much bigger change.

The "good enough for now" fix is to just wait between attempts to give the node time and resources to sync new ATXs to hopefully have one in the current epoch on the next try.

I chose to wait for the next layer because it scales with the network. It would only wait 15 seconds in system tests and 5 minutes on mainnet. A fixed time would probably be too long for system tests or not long enough for mainnet and this imo shouldn't be configurable.

countvonzero: not sure why `RegisterForATXSynced()` matters. atx synced only happens once in a node's lifetime. once the first time is done, every call afterwards will return immediately.

activation/activation.go
countvonzero: i find this confusing if the end result is the same. differentiating here means you should differentiate it everywhere else.

activation/activation.go
fasmat: I know, but this error happens when a node is not ATX synced (meaning it hasn't seen any ATXs that were published in the current epoch) and still returns a closed channel via `RegisterForATXSynced()`.

A node can be ATX synced and then lose this state - by being offline for some time or hibernating, or other causes.

activation/activation.go
fasmat: There is no way to return "no layer" since 0 is a valid layer and `-1` would panic. I can make the returned type a pointer and return `nil` instead below.

Imo it was confusing before: why does it return the same here as below in the case of an error?

poszu (approved):

fasmat:
bors merge

activation/activation.go
countvonzero: i agree its confusing before. but only changing it here make it more confusing. the code has many other place that does this. that was my point  : P

fasmat:
bors merge

PR #4059, opened 2023-02-13
[Merged by Bors] - beacon: reduce the likelihood of network disagreeing on beacon proposals

### LINKED ISSUE #3903: beacon: multiple beacon values in the network
## Description

when @poszu is testing https://github.com/spacemeshos/go-spacemesh/pull/3865, a race condition between hash fetcher and direct queries to poet servers caused some ATXs to be rejected (with error stating poet proof already existed). 

while this is easily fixed, it exposed a condition that when nodes have different view of ATXs in the last epoch, they will accept different set of beacon proposals and end up calculated different beacon values from each other.

another (minor) issue is that when the beacon protocol terminates when zero hashes, it should error out and wait for beacon sync.

countvonzero:
## Motivation
Closes #3903

## Changes
- make two-tiered atx threshold
  - `thresholdStrict` is calculated from number of atxs received δ before the end of the previous epoch, used to gate valid proposals
  - `threshold` is calculated from number of atxs received before the end of the previous epoch, used to gate potentially valid proposals

- allow beacon to be notified as atxs arrive.
- weakcoin: do not relay weakcoin message that contains higher value proposals


dshulyak (commented):
beacon is a perfect example where tests and implementation can be completely data driven, i think current testing approach is deeply flawed

dshulyak (approved):

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
> beacon is a perfect example where tests and implementation can be completely data driven, i think current testing approach is deeply flawed

completely agree. https://github.com/spacemeshos/go-spacemesh/issues/4061
will aim to create one before launch.

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

PR #4066, opened 2023-02-16
[Merged by Bors] - README: Fix instructions to run go outside of make

piersy:
## Motivation
The readme instructions were broken

## Changes
Readme instructions updated to be inline with the Makefile. Changed `print-env` and `print-test-env` to print the variable assignments on one line, this allows them to be easily incorporated into shell commands.

Since the print output could now easily be passed to `go env`, removed the `go-env` and `go-test-env` targets to reduce duplication.

I'm unsure whether i should make the same changes to `Makefile-svm.Inc`, since it already seemed to be somewhat out of sync with `Makefile-gpu.Inc`?

## Test Plan
Manually checked that the instructions in the readme work for building and testing.

## DevOps Notes

  - This changes the output format of some make commands and removes some make commands, so it could impact infrastructure config.

### DIFF
--- a/Makefile-gpu.Inc
+++ b/Makefile-gpu.Inc
@@ -107,22 +107,10 @@ print-test-targets:
 	done
 .PHONY: $(SUBDIRS_ALL) $(SUBDIRS_LVL2) $(SUBDIRS_ONLY) print-test-targets
 
-go-env: get-gpu-setup
-	go env -w CGO_CFLAGS="$(CGO_CFLAGS)"
-	go env -w CGO_LDFLAGS="$(CGO_LDFLAGS)"
-.PHONY: go-env
-
-go-env-test: get-gpu-setup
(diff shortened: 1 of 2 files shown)


dshulyak (approved):

piersy:
@dshulyak What's the situation with `Makefile-svm.Inc` should that be updated to follow `Makefile-gpu.Inc`?

dshulyak:
@piersy i think that file is not needed. it is a legacy lib, which is likely won't be relevant anymore

piersy:
> @piersy i think that file is not needed. it is a legacy lib, which is likely won't be relevant anymore

Thanks, I'll merge this for now and double check in our dev channel before removing `Makefile-svm.Inc`.

piersy:
bors merge

piersy:
bors merge

Issue #4067, opened 2023-02-17
investigate malfeasance proofs in devnet-323-short


countvonzero:
## Description

metrics for malfeasance proofs in devnet-323-short running binary [v0.2.21-beta.0](https://github.com/spacemeshos/go-spacemesh/releases/tag/v0.2.21-beta.0)

![Screenshot from 2023-02-17 12-53-18](https://user-images.githubusercontent.com/30611210/219552596-03244978-98c4-4165-aee3-7d7b4d8f8483.png)

things that are wrong:
- metrics over-reported the proofs: in DB there are [atxs:1 ballots:72 hare:3] malfeasance proofs (total 76). metrics reported 121 total.
- ballots
  - a bug was discovered and fixed in https://github.com/spacemeshos/go-spacemesh/pull/4026/files#diff-352cd57ffb5b4a49d2aa14b595f10c044d2f89abfd5dbfb92362ff4e67d6c61dR696 but didn't make it into the binary
- there are not enough information to verify whether the malfeasance proofs are legit. for example, objects used for proofs were not recorded.
- no metrics to indicate the number of malicious identity, which should match the total of malfeasance proof metrics.

the network also reported 3 instances of hare equivocation. but it's not clear yet whether it's legitimate malfeasance or a software bug.

![Screenshot from 2023-02-17 13-07-48](https://user-images.githubusercontent.com/30611210/219554464-8ab72734-017e-41ff-955a-1f3b3655b2f3.png)

![Screenshot from 2023-02-17 13-08-12](https://user-images.githubusercontent.com/30611210/219554506-3935c713-337c-46c9-a86f-75788845b9bb.png)

![Screenshot from 2023-02-17 13-08-29](https://user-images.githubusercontent.com/30611210/219554522-9c47ce37-85b2-4b39-80a9-9def13f87b02.png)

countvonzero:
to summarize the state of this issue.
- ballots equivocations was most likely a bug that was fixed in https://github.com/spacemeshos/go-spacemesh/pull/4026/files#diff-352cd57ffb5b4a49d2aa14b595f10c044d2f89abfd5dbfb92362ff4e67d6c61dR696
- atx and hare equivocations are still unclear. loggings were added to improve understanding in #4068

see #4085 for proposals

lrettig:
anything pending here? Can this be closed?

countvonzero:
mystery still unsolved. no code change is made.

countvonzero:
[...]

PR #4068, opened 2023-02-17
[Merged by Bors] - fix malfeasance proof metrics and add logging

countvonzero:
## Motivation

part of #4067

## Changes

- malfeasance proof: only update metrics when save is successful
- add more logging to malfeasance proof upon generation and reception
- make atx handler uses just 1 mutex, not 2
- rename all miner_id, atx_node_id, sender_Id logging field to just smesher
- fix benchmark tests in atx handler

### DIFF
--- a/Makefile-gpu.Inc
+++ b/Makefile-gpu.Inc
@@ -73,6 +73,10 @@ $(BINDIR_GPU_SETUP_LIBS): $(PROJ_DIR)$(GPU_SETUP_ZIP)
 $(PROJ_DIR)$(GPU_SETUP_ZIP):
 	curl -L $(GPU_SETUP_URL_ZIP) -o $(PROJ_DIR)$(GPU_SETUP_ZIP)
 
+go-env-test: get-gpu-setup
+	go env -w CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)"
+.PHONY: go-env-test
+
 get-gpu-setup: $(PROJ_DIR)$(GPU_SETUP_ZIP) $(BINDIR_GPU_SETUP_LIBS)
 .PHONY: get-gpu-setup
 
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -135,13 +135,13 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.VerifiedActivationT
 	h.log.WithContext(ctx).With().Info("processing atx",
 		atx.ID(),
 		atx.PublishEpoch(),
-		log.FieldNamed("atx_node_id", atx.NodeID()),
+		log.FieldNamed("smesher", atx.NodeID()),
 		atx.PubLayerID,
 	)
 	if err := h.ContextuallyValidateAtx(atx); err != nil {
 		h.log.WithContext(ctx).With().Warning("atx failed contextual validation",
 			atx.ID(),
-			log.FieldNamed("atx_node_id", atx.NodeID()),
+			log.FieldNamed("smesher", atx.NodeID()),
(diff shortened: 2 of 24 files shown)


activation/handler.go
```diff
@@ -517,7 +517,7 @@ func (h *Handler) handleAtxData(ctx context.Context, peer p2p.Peer, data []byte)
 		r.OnAtx(header)
 	}
 	events.ReportNewActivation(vAtx)
-	logger.With().Info("new atx", log.Inline(atx), log.Int("size", len(data)))
+	logger.With().Info("new atx", log.Inline(vAtx), log.Int("size", len(data)))
```
countvonzero, line 517: @dshulyak is this sufficient? this will log base tick height, tick count and weight

activation/handler.go
```diff
@@ -371,7 +367,8 @@ func (h *Handler) StoreAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 				}
 				h.log.With().Warning("smesher produced more than one atx in the same epoch",
 					log.Stringer("smesher", atx.NodeID()),
-					log.Inline(atx),
+					log.Object("prev", prev),
+					log.Object("curr", atx),
```
fasmat, line 371: Would logging the ID be sufficient?

fasmat (approved):

activation/handler.go
countvonzero: thanks. the same argument can be made about ballots and i have considered that too.
at this stage i prefer to log the whole thing for two reasons:
- there may be bugs in the code that cause honest miners to generate multiple atxs/ballots in an epoch/layer. the extra information will be helpful for us
- if everything works as expected, this occasion should be rare and the extra log won't matter.

i'm open to reduce the log once we have more confidence with the system.

countvonzero:
bors try

Makefile-gpu.Inc
```diff
@@ -73,6 +73,10 @@ $(BINDIR_GPU_SETUP_LIBS): $(PROJ_DIR)$(GPU_SETUP_ZIP)
 $(PROJ_DIR)$(GPU_SETUP_ZIP):
 	curl -L $(GPU_SETUP_URL_ZIP) -o $(PROJ_DIR)$(GPU_SETUP_ZIP)
 
+go-env-test: get-gpu-setup
+	go env -w CGO_LDFLAGS="$(CGO_TEST_LDFLAGS)"
+.PHONY: go-env-test
+
```
countvonzero, line 79: @piersy @fasmat 
moving this target from -svm to -gpu as goland requires it to be in the same location as get-gpu-setup. 
(and i didn't want to spend time to investigate goland configuration further)

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

Makefile-gpu.Inc
piersy: Hey @countvonzero, I just removed that target (#4066 ) from `Makefile-gpu.Inc` because I thought that the `print-env` and `print-env-test` targets should be sufficient. Is `go-env-test` a special target that is depended upon by goland?

countvonzero: @piersy you are right. i just verified that `print-test-env` is equivalent to `go-env-test` for goland.

i think files under [`https://github.com/spacemeshos/go-spacemesh/tree/develop/.run`](https://github.com/spacemeshos/go-spacemesh/tree/develop/.run) defines the targets for goland, but to be honest i don't think we need .run directory. it seems IDE specific and checked in by accident.

PR #4069, opened 2023-02-17
Revert "build(deps): Bump github.com/pyroscope-io/pyroscope from 0.37…

piersy:
## Motivation
It seems that commit 783d34035a6faefbfcd79d103bafc62c81e2a38d broke CI by breaking the windows build. This pr reverts it.

## Test Plan
The windows build is not failing in CI for this branch.

## DevOps Notes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/oasisprotocol/curve25519-voi v0.0.0-20230110094441-db37f07504ce
 	github.com/prometheus/client_golang v1.14.0
 	github.com/prometheus/common v0.39.0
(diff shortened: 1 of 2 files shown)


piersy:
Actually, I think this was just a random failure in the windows build, not caused by the pyroscope update

fasmat:
> Actually, I think this was just a random failure in the windows build, not caused by the pyroscope update

We do have some flaky tests, especially in Windows and macOS builds. That's why at the moment failing tests on either of those two systems don't fail the CI pipeline.

If you had some insight in one of those tests and now an idea on how to make it better go for it 🙂

piersy:
@fasmat It'll probably be a while till I have some insight there :)

Issue #4091, opened 2023-02-26
add integrated testing for malfeasance proof


countvonzero:
## Description

currently there isn't a way to end-to-end test malfeasance proof code path.

this require 
- starting a set of honest nodes N1
- starting a set of dishonest nodes N2 that equivocates in ATX/ballots/hare messages
- starting a set of new nodes N3 to sync from the network

at the end of testing, all nodes should have the same set of malfeasance proofs, one for each node in N2.

countvonzero:
see also https://github.com/spacemeshos/go-spacemesh/issues/3988

PR #4092, opened 2023-02-26
[Merged by Bors] - provide data validator for malfeasance proof

countvonzero:
## Motivation

fixing user report https://github.com/spacemeshos/smapp/issues/1168

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x7ff6b92b40e2]

goroutine 328 [running]:
github.com/spacemeshos/go-spacemesh/fetch.(*Fetch).GetMalfeasanceProofs(0xc002206000, {0x7ff6ba040f18, 0xc003a88ab0}, {0xc0031ca100, 0x1, 0xc00058f168?})
	D:/a/go-spacemesh/go-spacemesh/fetch/mesh_data.go:72 +0x302
github.com/spacemeshos/go-spacemesh/syncer.fetchMalfeasanceProof({0x7ff6ba040f18, 0xc003a88ab0}, {0xc0022b4a80?, {0x7ff6b9b92e2f?, 0x0?}}, {0x7ff6ba030880?, 0xc0017aea40?}, {0x7ff6ba04c2b0, 0xc002206000}, 0xc003eb43c0, ...)
	D:/a/go-spacemesh/go-spacemesh/syncer/data_fetch.go:271 +0x222
github.com/spacemeshos/go-spacemesh/syncer.(*DataFetch).PollMaliciousProofs(0xc0017a5770, {0x7ff6ba040f18?, 0xc003a88ab0})
	D:/a/go-spacemesh/go-spacemesh/syncer/data_fetch.go:106 +0x53e
github.com/spacemeshos/go-spacemesh/syncer.(*Syncer).syncMalfeasance(0xc0038dec40?, {0x7ff6ba040f18?, 0xc003a88ab0?})
	D:/a/go-spacemesh/go-spacemesh/syncer/syncer.go:497 +0x32
github.com/spacemeshos/go-spacemesh/syncer.(*Syncer).synchronize.func1()
	D:/a/go-spacemesh/go-spacemesh/syncer/syncer.go:384 +0x329
github.com/spacemeshos/go-spacemesh/syncer.(*Syncer).synchronize(0xc002206180, {0x7ff6ba0425d8?, 0xc001628140?})
	D:/a/go-spacemesh/go-spacemesh/syncer/syncer.go:422 +0xa13
github.com/spacemeshos/go-spacemesh/syncer.(*Syncer).Start.func1.1()
	D:/a/go-spacemesh/go-spacemesh/syncer/syncer.go:242 +0x107
	C:/Users/runneradmin/go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:75 +0x64
	C:/Users/runneradmin/go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:72 +0xa5
```

## Changes
- initialize data validator for malfeasance proof in fetcher
- due to dependency mess in node.go, data validators can only be added when all handlers are initialized. as a result, a check for validators is added in `fetcher.Start()`

see https://github.com/spacemeshos/go-spacemesh/issues/4091 for adding tests.

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -31,8 +31,12 @@ const (
 	cacheSize = 1000
 )
 
-// errExceedMaxRetries is returned when MaxRetriesForRequest attempts has been made to fetch data for a hash and failed.
-var errExceedMaxRetries = errors.New("fetch failed after max retries for request")
+var (
+	// errExceedMaxRetries is returned when MaxRetriesForRequest attempts has been made to fetch data for a hash and failed.
+	errExceedMaxRetries = errors.New("fetch failed after max retries for request")
+
+	errValidatorsNotSet = errors.New("validators not set")
+)
 
 // request contains all relevant Data for a single request for a specified hash.
 type request struct {
@@ -95,7 +99,7 @@ func DefaultConfig() Config {
 // randomPeer returns a random peer from current peer list.
 func randomPeer(peers []p2p.Peer) p2p.Peer {
 	if len(peers) == 0 {
-		log.Panic("cannot send fetch: no peers found")
+		log.Fatal("cannot send fetch: no peers found")
 	}
 	return peers[rand.Intn(len(peers))]
 }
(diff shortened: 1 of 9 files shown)


countvonzero:
bors try

fetch/interface.go
```diff
@@ -13,32 +13,32 @@ type requester interface {
...
-type txHandler interface {
+type TxValidator interface {
 	HandleBlockTransaction(context.Context, p2p.Peer, []byte) error
 	HandleProposalTransaction(context.Context, p2p.Peer, []byte) error
 }
 
-type poetHandler interface {
+type PoetValidator interface {
 	ValidateAndStoreMsg(context.Context, p2p.Peer, []byte) error
 }
```
poszu, line 43: Do you know what is the reason that these all are not just a single interface like:
```go
type P2PHandler interface {
    Handle(context.Context, p2p.Peer, []byte) error
}
```
?

fetch/fetch.go
```diff
@@ -252,14 +201,49 @@ func NewFetch(cdb *datastore.CachedDB, msh meshProvider, b system.BeaconGetter,
...
+		malfeasance: mal,
+	}
+}
+
 // Start starts handling fetch requests.
-func (f *Fetch) Start() {
+func (f *Fetch) Start() error {
+	if f.validators == nil {
+		return errValidatorsNotSet
+	}
```
poszu, line 239: `f.validators == nil` means a bug in the code. Perhaps instead of checking this at runtime, better prevent it altogether by requiring the user to pass all validators during the construction of `Fetch`? Similar to what was in the original code but actually to require it rather than making it an option.

fetch/interface.go
countvonzero: if you take a look at node.go, these interface are provided by different instances in different packages with different validation logic. i cannot think of a quick way to resolve this in a clean way.

fetch/fetch.go
countvonzero: the dependency in node.go is messed up. one needs to wait for all the handlers to be initialized before setting them in fetcher. but fetcher is needed earlier than one can finish initializing all handlers...

fetch/interface.go
poszu: Rename the methods or wrap them :)

poszu (approved):

fetch/fetch.go
poszu: Then perhaps handlers should be "registered" as they are booted up?

fetch/interface.go
countvonzero: renaming methods is not an option without big change, as the same handler can handle multiple data (see TxValidator)
wrapping them is possible, but that basically is a layer of indirection.

i.e. you will have an 
```go
type dataValidator interface {
   validateA()
   validateB()
   validateC()
}
```
then fills the following struct with the gossip handlers for A, B, and C
```go
type validatorImpls struct {
   ValidatorAImpl
   ValidatorBImpl
   ValidatorCImpl
}
```
and pass this validatorImpls to fetcher. it's not clear to me that this approach is better. let me know if you think differently and i will follow up with another PR

countvonzero:
bors merge

PR #4096, opened 2023-02-27
include domain in the signed data

### LINKED ISSUE #3967: malfeasance proof: prevent cross-data types misuse of eligibility
## Description
currently ATX, ballot and hare message uses different metadata for signing. and the code only validates that the same identity signed two different message hash.

since ATX and ballot metadata are structurally equivalent, a malicious actor can stitch up one metadata from ballot and one from atx for an honest identity H and convince the network that H is malicious.

the research team concluded that there should be shared metadata across all roles (beacon, ATX, ballots, hare for now, possibly validator... for future). such metadata has the structure below and is fixed-sized.

```
Metadata {
  Role: 1 byte
  Layer/Epoch: 4 bytes (Layer for ballots/hare roles, Epoch for beacon and ATX roles)
  Round: 2 bytes (for hare only, ignored for others. for hare it also infers)
  MessageHash: 32 bytes (the hash of the beacon/ATX/ballot/hare message body)
}
```

there were discussion on whether to include protocol versions into the metadata as well. the decision is nay.

the main idea here is not allowing for protocol versions enforce the rule:
one eligibility can be used for one role at most once, even across protocol versions.

for example, if spacemesh were to upgrade ATX protocol version from A->B. if a miner generates one ATX for version A and one ATX for version B in the same epoch, it is considered to be malicious.

countvonzero:
## Motivation
Closes #3967

## Changes
- add domain in signer and key extractor
- in node.go, generate signer/extractor pairs for atx, beacon, ballots and hare
- when creating proposals, instead of using `Initialize(PubKeyExtractor)` like the gossip codepath, explicitly calculates and sets ID (like atx builder), since it doesn't make sense to pass a key extractor to the miner component.
- creates some test only methods to facilitate creating atx and ballots.
- (unrelated) use eligibility type for beacon vrf msg


malfeasance/handler_test.go
```diff
@@ -728,3 +729,57 @@ func TestHandler_HandleMalfeasanceProof_validateHare(t *testing.T) {
...
+					Messages: [2]types.BallotProofMsg{
+						{InnerMsg: m1, Signature: signer.Sign(final1)},
+						{InnerMsg: *(*types.BallotMetadata)(unsafe.Pointer(&m2)), Signature: signer.Sign(final2)},
+					},
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, pubsub.ValidationIgnore, h.HandleMalfeasanceProof(context.Background(), "", msg))
```
countvonzero: @dshulyak this is essentially the test you supplied (thank you!)
it's not rejected because the code failed looking up the node ID after extracting the public key. so ignore instead

dshulyak:
i think this is not a good approach

we are using signatures in multiple domains, e.g atx - 1, ballot - 2, hare - 3
each mal. proof contains 2 messages that are signed in the specific domain.

i also think that there should be explicit API for signing with domain, such as
`Signer.SignerWithDomain(Domain, []byte)`
and then replace previous non-domain signature everywhere

countvonzero:
bors try

common/types/ballot.go
```diff
@@ -204,7 +205,11 @@ type EpochData struct {
 
 // Initialize calculates and sets the Ballot's cached ballotID and smesherID.
 // this should be called once all the other fields of the Ballot are set.
-func (b *Ballot) Initialize() error {
+func (b *Ballot) Initialize(extractor keyExtractor) error {
```
dshulyak, line 208: but why it needs to be passed like that? domain is already known in this function, in my opinion single line change would be much nicer

```go
ExtractNodeID(BallotDomain, data, sig)
``` 

or if you want to encapsulate BallotDomain in the object

```go
signing.Ballots.ExtractNodeID(data, sig)
```

dshulyak:
how much of this code is essential for the change? it is important to get right, but i can't really review it

dshulyak (dismissed):

signing/signer.go
```diff
@@ -13,9 +13,20 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
+type Domain byte
+
+const (
+	Default Domain = iota
```
dshulyak, line 19: default domain? this is for stuff that is not refactored? there should be no such thing

dshulyak:
i looked up codebase, there are about 200 calls to Sign (177) and Extract (8). Doing `Sign(domain, data)` and `Extract(domain, data, sig)` would produce much nicer change. and also remove the need for default domain, as it won't be possible to avoid passing some meaningful domain to sign function

the difference between genesis id and domain, is that genesis id is known only at runtime (unless we compile software for a specific network). domain is known at compile time, and every function that uses sign already knows the domain

dshulyak:
nice tweak to prevent sending custom domains by mistake is to make a private type like that, and define exhaustive list of domains

image

countvonzero:
> i looked up codebase, there are about 200 calls to Sign (177) and Extract (8). Doing `Sign(domain, data)` and `Extract(domain, data, sig)` would produce much nicer change. and also remove the need for default domain, as it won't be possible to avoid passing some meaningful domain to sign function

the original intent is also breaking the cyclic dependency btwn common/types and signing, as we extract the public in Initialize().
workaround like https://github.com/spacemeshos/go-spacemesh/blob/6271e39ddafa202c8061fe5143911eb0a0dff419/cmd/node/node.go#L517 was done for that purpose.

but i agree that this is closer to a declarative approach i prefer. will adopt your suggestion.

common/types/ballot.go
countvonzero: the original intent was to break the cyclic dependency btwn common/types and signing. i think i want to move ExtractKey outside of common/types pkg.

PR #4104, opened 2023-02-28
Target in HTTPClient can be now https

### LINKED ISSUE #4102: Poet URL have to accept https endpoints too
```
func NewHTTPPoetClient(target string) *HTTPPoetClient {
	return &HTTPPoetClient{baseURL: fmt.Sprintf("http://%s/v1", target)}
```

It needs to accept https endpoints too.

pigmej:
Closes #4102

### DIFF
--- a/activation/poet.go
+++ b/activation/poet.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
@@ -127,8 +128,13 @@ func NewHTTPPoetClient(target string, cfg PoetConfig) *HTTPPoetClient {
 		return false, nil
 	}
 
+	// if no prefix given assuming http://
+	if !strings.HasPrefix(target, "http://") && !strings.HasPrefix(target, "https://") {
+		target = fmt.Sprintf("http://%s/v1", target)
+	}
+
 	return &HTTPPoetClient{
-		baseURL: fmt.Sprintf("http://%s/v1", target),
+		baseURL: target,
 		client:  client,
 	}
 }
--- a/activation/poet.go
+++ b/activation/poet.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
@@ -126,11 +125,6 @@ func NewHTTPPoetClient(target string, cfg PoetConfig) *HTTPPoetClient {
 			return true, nil
 		}
 		return false, nil
-	}
-
-	// if no prefix given assuming http://
-	if !strings.HasPrefix(target, "http://") && !strings.HasPrefix(target, "https://") {
-		target = fmt.Sprintf("http://%s/v1", target)
 	}
 
 	return &HTTPPoetClient{
--- a/config/config.go
+++ b/config/config.go
@@ -160,7 +160,7 @@ func defaultBaseConfig() BaseConfig {
 		OracleServerWorldID: 0,
 		LayerDuration:       30 * time.Second,
 		LayersPerEpoch:      3,
-		PoETServers:         []string{"127.0.0.1"},
+		PoETServers:         []string{"http://127.0.0.1/v1"},
 		SyncRequestTimeout:  2000,
 		SyncInterval:        10,
 		TxsPerProposal:      100,


activation/poet.go
```diff
@@ -127,8 +128,13 @@ func NewHTTPPoetClient(target string, cfg PoetConfig) *HTTPPoetClient {
 		return false, nil
 	}
 
+	// if no prefix given assuming http://
+	if !strings.HasPrefix(target, "http://") && !strings.HasPrefix(target, "https://") {
+		target = fmt.Sprintf("http://%s/v1", target)
+	}
```
fasmat: Go has URL parsing. Without testing the code, something like this should avoid manual string parsing and conserve the scheme defined (or set it to http if not defined).

```suggestion
	baseUrl, err := url.Parse(target)
	if err != nil {
		// not valid URL
	}
	baseUrl = baseUrl.JoinPath("v1")
```

countvonzero: i think we should simply remove this block

countvonzero: why not pass the whole string without any parsing? what's the benefit of guessing here?

fasmat: The goal is to add the `/v1` path. A more complete example: https://go.dev/play/p/MiaqecTyyL3

imo the manual setting of `URL.Scheme` is probably not necessary.

countvonzero: why is adding `v1` a goal? and what happens if we need to change to `v2`? or talk to `v1` and `v2` at the same time?

fasmat: Then we need to explicitly handle that 😉 

Look at the current code, it also adds the `v1` path, but using `fmt.Sprintf` instead of `url.Parse()` / `(*URL) JoinPath()`.

countvonzero: so we explicitly set it. i don't see anything wrong with that. 
i find it strange that we have to segment the url and via config and hardcoded string, when we can just set the whole string in config.

countvonzero: i am saying that the old code doesn't make sense to me.

fasmat: It's not unusual to specify an endpoint to connect to and let the client decide which version of the protocol it wants to talk to that endpoint. In our case we differentiate incompatible versions of the API via path, `PoetHttpClient` understands `v1` of the protocol so it "selects" it by adding the path to the URL.

I would find it odd to have to define `http://mydomain.com/v1` as the endpoint in a config file instead of `http://mydomain.com`.

countvonzero: i don't understand why it's odd. what happens if v1 got deprecated by the resource owner that's not us?
it simply doesn't make sense to hardcode v1

fasmat: At least that's how I'm used to it with openAPI generated servers and clients, where one usually also just defines the domain but not the specific version of the protocol to use.

countvonzero: the pros are:
- it's flexible, we control the exact url in config
- it's less code

what are the cons?

fasmat: > i don't understand why it's odd. what happens if v1 got deprecated by the resource owner that's not us? it simply doesn't make sense to hardcode v1

Then our client breaks anyway. `v2` is by definition not backwards compatible and our client doesn't understand it.

fasmat: Updating the URL from `http://mydomain.com/v1` to `http://mydomain.com/v2` in a config will not make our client understand a different version of the protocol. We have to update code anyway. Why leave room for a user error?

countvonzero: > Then our client breaks anyway. v2 is by definition not backwards compatible and our client doesn't understand it.

lets say we are using a limited portion of API, and v1 and v2 both support them. but the resource owner is going to deprecated v1. we have to then change the code to v2. but we could just change the config to v2. 

i still don't understand why we want less flexibility? and i also feel like we are spending too much time/effort arguing this.

fasmat: > lets say we are using a limited portion of API, and v1 and v2 both support them. but the resource owner is going to deprecated v1. we have to then change the code to v2. but we could just change the config to v2.

This is an extremely specific use case that I've never seen...

> i still don't understand why we want less flexibility? and i also feel like we are spending too much time/effort arguing this.

I agree, I say we handle it like pretty much like any other HTTP API handles it and how it is currently implemented. I'm not advocating for changing it 🙂.

countvonzero: ok. i promise this is the last post. 

fundamentally i don't want to hardcode anything. i don't know why you want to argue against that.
take a look of our networks.json. you will see lots of url string passed.

pigmej: Ok. I'll remove that completely.

Basically anything will be passed further. 

Network config will need to have that then added Including the one for systests.
I wanted to avoid changes there but maybe it's ok...

countvonzero: no worries. just submit as is is fine. no need to expand the change

countvonzero (approved):

activation/poet.go
piersy: This block is confusing to me and I think it will be for users too.

IIUC for a poet server at `http://mydomain.com` as a user I could provide the following in the config
* `http://mydomain.com/v1`
* `mydomain.com`
but not
* `http://mydomain.com`

I'd rather not have 2 different valid strings for the address of a server, it makes it difficult to compare configs and tends to confuse users because they don't know what is the "right" way to config.

And the fact that `http://mydomain.com/v1` &  `mydomain.com` are accepted but `http://mydomain.com` will not be, I think would confuse users even more.

pigmej: Yes, I'll remove that completely. ANd then fix tests.

PR #4105, opened 2023-02-28
[Merged by Bors] - block: use ATX IDs to express rewards

### LINKED ISSUE #4097: block: use ATX IDs instead of coinbase addresses for rewards
## Description

use ATX IDs in block for 
- rewards distribution
- facilitate hare active set impl per https://community.spacemesh.io/t/grading-atxs-for-the-active-set/335
- compatible with future incentive mechanism

countvonzero:
## Motivation
Closes #4097

## Changes
- replace Address (coinbase) with ATXID in AnyAward
- use a new type CoinbaseAward to interface with VM.
- remove unused field Layer from BlockContextualValidity


dshulyak (commented):
could you please keep vm code as is but provide different types from upper level?

countvonzero:
> could you please keep vm code as is but provide different types from upper level?

i thought thats what i did. i replaced AnyReward with CoinbaseReward and didnt change any code inside.

i moved the verifying utility out because the logic no longer applies there. we dont need to ensure unique coinbase anymore

countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):
thanks, i thought that ValidateRewards was moved somehwere

piersy:
Hi @countvonzero I'm not understanding the motivation for changing the coinbase to be the ATX ID

countvonzero:
> Hi @countvonzero I'm not understanding the motivation for changing the coinbase to be the ATX ID

@piersy 
most important reason is here https://community.spacemesh.io/t/grading-atxs-for-the-active-set/335
secondary reason is that we want to leave room for future rewards as incentives for hare participation, running poet servers ...etc. the consensus within research team is that it's easier to record ATXID so that we can encode/infer those reward recipients from the ATX itself.

countvonzero:
> thanks, i thought that ValidateRewards was moved somehwere

the code was moved to the block handler to validate a block. but the logic was changed to verify unique ATXIDs.
since we don't want to leak atx logic into vm. and i think there is no longer needs to check for uniqueness of coinbase addresses. if you disagree, i can add it back. @dshulyak

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

Issue #4060, opened 2023-02-13
Sometimes we're advertisting the private IP addreses over the p2p network


pigmej:
Sometimes in the log, we can see failures to connect to the private IP addresses, that possibly mean that we're sometimes discovering wrong IP addresses and then announcing them as node IPs.

```
{"L":"WARN","T":"2023-02-13T06:17:06.625-0500","N":"fc000.p2p","M":"failed to connect to peer","node_id":"fc000227bf4b78cb8110a059b34a6dfcf73abee8133f9c72ba114837be13c7ca","module":"p2p","peer":"12D3KooWCX5G9fdxsnxeJVoh6ckEc33Xa2yXXAcv8jPrKuEKJEv9","errmsg":"failed to dial 12D3KooWCX5G9fdxsnxeJVoh6ckEc33Xa2yXXAcv8jPrKuEKJEv9:\n  * [/ip4/10.8.0.50/tcp/17098] dial tcp4 10.8.0.50:17098: connect: connection refused","name":"p2p"}
{"L":"WARN","T":"2023-02-13T06:17:03.933-0500","N":"6e60a.p2p","M":"failed to connect to peer","node_id":"6e60a37ab7d9672927086f386bfedce6081477ecbd836911aeec96d330bcd352","module":"p2p","peer":"12D3KooWKr4toMwWzP5j9S6EhfG2pNBbrFvCDn9RbW7DpeB55SeE","errmsg":"failed to dial 12D3KooWKr4toMwWzP5j9S6EhfG2pNBbrFvCDn9RbW7DpeB55SeE:\n  * [/ip4/10.8.0.53/tcp/7513] dial tcp4 10.8.0.53:7513: connect: connection refused","name":"p2p"}
[...]

PR #4108, opened 2023-03-01
[Merged by Bors] - rework discovery

dshulyak:
the main change in this pr is a new module in p2p directory: book. it is a replacement for legacy addressbook and a bunch of workarounds that were made around it.

new book targets the following behavior:
- internally it mains two buckets - private and public. private addresses with be shared over private connections. public - over public. any dns name is assumed to be public
- when peer learns new address it is classified as `learned` and placed into the queue of addresses to try. if it fails - we place it immediately into `stale`. `stale` addresses are not shared with peers. if it succeeds it will be placed into `stable`, the reason for this promotion is to make such address a bit stickier, and don't drop it on first failure.
- address are shared randomly, book filters them based on the bucket and class of the requestor

concurrency code for bootstrapper was simplified. it will make N concurrent requests every period, based on the order maintained in the book. checker (component that suppose to maintain quality of the local book) is removed, instead it will rely on the regular crawling code, but number of concurrent requests and period are adjusted.

how it should play out in our deployment model:
- bootsrappers will connect to each other over private cloud network and advertise dns addresses to the world.
- any node outside the cloud connects to the bootsrapper, bootstrapper learns its public address and a port. and shares it with other nodes with public addresses. if address turns out to be not dialable - it will be moved to a `stale` class, and will remain there until it is time to delete it.

deployment changes:
`--min-peers` replaces `--target-outbound`. once node established connections with min-peers it will switch into slower mode.

metrics to track the state of the discovery were added to spacemesh_discovery namespace: total_addresses, connected_addresses, addresses_in_bucket{private, public}, addresses_in_class{stale, learned, stable}.


countvonzero:
> address are shared randomly, book filters them based on the bucket and class of the requestor

what are the buckets and classes of requestor?

> how it should play out in our deployment model:

does it mean that our deployment is dependent on having a private network?

dshulyak:
> what are the buckets and classes of requestor?

buckets are private and public, classes are stale/learned/stable

> does it mean that our deployment is dependent on having a private network?

no it doesn't

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+
+type Address = ma.Multiaddr
+
+type class uint32
+
+const (
+	deleted class = iota
+	stale
+	learned
+	stable
```
countvonzero, line 32: comment on the definition of each class plz

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+)
+
+func classify(addr *addressInfo) class {
+	switch addr.Class {
+	case stale:
+		if addr.success == 2 {
```
countvonzero, line 38: is it possible that this would be > 2? here and below

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+	shareable []*addressInfo
+}
+
+func (b *Book) Add(src, id ID, raw Address) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	addr := b.known[id]
+	bucket := bucketize(raw)
+	if src != SELF {
+		if source := b.known[src]; source == nil || source.bucket != bucket {
```
countvonzero, line 161: does this fix https://github.com/spacemeshos/go-spacemesh/issues/4060?

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+			stats.Learned++
+		case stable:
+			stats.Stable++
+		}
+	}
+	return stats
+}
+
+func (b *Book) iterShareable(src ID, bucket bucket) iterator {
+	b.rng.Shuffle(len(b.shareable), func(i, j int) {
```
countvonzero, line 336: what is the intuition for shuffling here?

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+		return
+	} else if addr == nil {
+		addr = &addressInfo{
+			Raw:       jsonAddress{raw},
+			Class:     learned,
+			ID:        id,
+			bucket:    bucket,
+			shareable: true,
+		}
+		b.shareable = append(b.shareable, addr)
```
countvonzero, line 176: so the combination of `addressInfo.shareable` and the book's `shareable` array appear to be a map.
in the past you have shared that map has higher overhead. i am curious what is your consideration here.
is the number of shareable addresses going to be high?

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+			if !addr.shareable {
+				addr.shareable = true
+				b.shareable = append(b.shareable, addr)
+			}
+		case Connected:
+			addr.Connected = true
+		case Disconnected:
+			addr.Connected = false
+		case Success, Fail:
+			b.queue.PushBack(addr)
```
countvonzero, line 216: why put the addr back onto the queue when a crawl attempt had failed previously?

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+
+func recover(known map[ID]*addressInfo, r io.Reader) error {
+	checksum := crc64.New(crc64.MakeTable(crc64.ISO))
+	scanner := bufio.NewScanner(r)
+	stored := uint64(0)
+	for scanner.Scan() {
+		if len(scanner.Bytes()) == 0 {
+			return fmt.Errorf("corrupted data: empty lines are not expected")
+		}
+		if scanner.Bytes()[0] != '{' {
```
countvonzero, line 419: the old code assumes the first 4 bytes is the CRC value. that seems cleaner to me.
what's the reason to take a different approach here?

p2p/book/book.go
```diff
@@ -0,0 +1,434 @@
...
+		}
+		addr := &addressInfo{}
+		if err := json.Unmarshal(scanner.Bytes(), addr); err != nil {
+			return fmt.Errorf("unmarshal %s: %w", scanner.Text(), err)
+		}
+		checksum.Write(scanner.Bytes())
+		checksum.Write([]byte{'\n'})
+		known[addr.ID] = addr
+	}
+	if stored != checksum.Sum64() {
```
countvonzero, line 435: reset `known` to a empty map?

p2p/book/book.go
dshulyak: classify called after every update

p2p/book/book.go
dshulyak: i guess this line is important for that

p2p/book/book.go
dshulyak: it is dropped from a queue only when address is deleted from a book

p2p/book/book.go
dshulyak: to return random results

p2p/book/book.go
dshulyak: ok, i expected this error to result in the os.exit though

p2p/book/book.go
dshulyak: i overlooked it when reviewing the change, or maybe suggested that by mistake

appending checksum doesn't require to keep content in memory, so it is either better to add it as a suffix or separate file

p2p/book/book.go
dshulyak: i don't know what useful i can add, it is private classes. developers are expected to read code to understand what they are about

p2p/book/book.go
dshulyak: map won't be deterministic

countvonzero: it's not deterministic but also not random enough so you have to do the shuffling before sharing :P

p2p/book/book.go
countvonzero: it would be so in the code now. but the code path may change in the future. better on the safe side.

p2p/book/book.go
countvonzero: the condition to put an address in the class would help. but i don't insist.

implementation comments often helps reader to understand faster. yes i will eventually understand it but it'll take longer. i guess this is a natural entry barrier :D

countvonzero (approved):

p2p/book/book.go
dshulyak: i think there is no qualitative difference in rng used for maps and math/rand. i can't write tests in a good way if order is not deterministic

dshulyak:
bors merge

dshulyak:
image

dshulyak:
bors merge

p2p/book/book.go
piersy: I'd also be in favour of commenting these for two reasons.

* As @countvonzero said, it takes time to understand what these definitions mean, there are 40 different references to these classes throughout the diff, so understanding really what they do is not quick or easy. If a quick comment can save all that time for every future reader of the code then as a team it's much more efficient to comment than to not comment.
* The idea that you just need to read the code to understand what it does assumes that the code is correct. If there is a bug and the written code does not actually follow the author's original intent, the readers may never know because there was no comment to let them understand the original intent of the code.

dshulyak:
bors try

dshulyak:
bors try

p2p/book/book.go
dshulyak: the intent is expressed in tests

dshulyak:
bors try

dshulyak:
bors cancel

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

PR #4116, opened 2023-03-01
[Merged by Bors] - Allow any scheme for http poet client and default to http

### LINKED ISSUE #4102: Poet URL have to accept https endpoints too

fasmat:
## Motivation
Closes #4102

## Changes
This changes HTTPPoetClient to only default to http as scheme if it isn't set in the address configuration. This allows to the node to connect to poet via https if it is set as part of the address.

## Test Plan
- Existing tests pass
- Added test to check if client defaults to http if not set

## TODO

## DevOps Notes


activation/activation.go
```diff
@@ -477,14 +477,17 @@ func (b *Builder) UpdatePoETServers(ctx context.Context, endpoints []string) err
 
 	clients := make([]PoetProvingServiceClient, 0, len(endpoints))
 	for _, endpoint := range endpoints {
-		client := b.poetClientInitializer(endpoint, b.poetCfg)
+		client, err := b.poetClientInitializer(endpoint, b.poetCfg)
+		if err != nil {
+			return &PoetSvcUnstableError{source: fmt.Errorf("failed to initialize poet client for '%s': %w", endpoint, err)}
```
countvonzero: nit: i would shorten this. maybe just "init poet client %s: %w", here and other places.

countvonzero (approved):

pigmej (approved):

activation/poet.go
```diff
@@ -84,7 +84,13 @@ func NewHTTPPoetHarness(ctx context.Context, poetdir string, opts ...HTTPPoetOpt
 		return nil, err
 	}
 
-	client, err := NewHTTPPoetClient(cfg.RawRESTListener, PoetConfig{
+	// NewHTTPPoetClient takes an URL as connection string. Server speaks HTTP.
+	url := &url.URL{
+		Scheme: "http",
+		Host:   poet.GrpcRestProxyAddr().String(),
```
countvonzero: i am confused.... isn't the whole point being able to configure the scheme?

fasmat: The `HTTPPoetClient` can take any URL now and does not assume `http` any more. 

The server we are starting with `NewHTTPHarness` exposes an address. Since we are connecting via localhost and are not using `https` in tests we can hardcode the scheme here.

The code here is only used in tests. The node can still be configured to connect to other poets via `https` now.

countvonzero: oh! this is super confusing. can you move this function to a test then?

fasmat: I moved the `HTTPPoetHarness` to its own file and renamed it to `HTTPPoetTestHarness` with a clarifying comment that it is only to be used in tests.

@poszu maybe it makes sense to move `HTTPPoetClient` and `HTTPPoetTestHarness` into the poet repository? Maybe to `poet/client` and `poet/integration` respectively? The only thing from go-spacemesh that the two depend on is the type `type.PoetServiceID` which is just an alias for `[]byte`.

fasmat:
bors try

activation/poet.go
fasmat: I created a ticket in the poet repository: https://github.com/spacemeshos/poet/issues/253

There are 2 implementations of the PoetHarness. I think it would be beneficial to merge those two into the poet repository and only use the harness here, instead of also having the implementation in `go-spacemesh`.

fasmat:
bors merge

poszu (approved):

fasmat:
bors merge

fasmat:
bors merge

PR #4100, opened 2023-02-28
Piersy/speedup ci tests trying to break it

piersy:
Draft for CI

## Motivation

Closes #

## TODO

- [ ] Explain motivation or link existing issue(s)
- [ ] Test changes and document test plan
- [ ] Update documentation as needed

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #4121, opened 2023-03-02
[Merged by Bors] - hare: do not vote for future proposals

### LINKED ISSUE #4084: mitigate slightly faster poet servers
## description

research discussion https://community.spacemesh.io/t/fast-poet-could-prevent-voting-by-honest-smeshers/339

specifically [Preventing Voting for Future Proposals in Hare](https://community.spacemesh.io/t/fast-poet-could-prevent-voting-by-honest-smeshers/339#preventing-voting-for-future-proposals-in-hare-3) will be implemented.

## implementation plan
- before:
hare preround participants voted for proposals that have the same beacon value as the participant
- after:
hare preround participants voted for proposals that
  - carry the same beacon value as the participant
  - reference ATXs that have lower tick height than the participant's own ATX

@noamnelke @pigmej

### LINKED ISSUE #4120: hare: configure N is sufficient to infer F
## Description
current configuration require setting N and F. it is sufficient to simply set N and infer that F is N/2.

countvonzero:
## Motivation
Closes #4084
Closes #4120

## Changes
- do not vote for future proposals. i.e. only put proposals with base tick height lower than node's own tick height in the hare's initial set.
- remove F from hare config and infer it from N
- make clock log level configurable. currently it's logging at DEBUG (every second) on devnet
- test improvements
  - mock database in tests for all data (atx header, proposals, ballot, malfeasance proof)
  - remove unnecessary clock setup for Test_multipleCPs and Test_multipleCPsAndIterations
  - fixed Test_multipleCPsAndIterations and disrupt gossip traffic at a layer after genesis 
  - reduce proposals from 200 to 20 in Test_multipleCPs and Test_multipleCPsAndIterations. getting proposals is causing ~2s delay in tests and cause many nodes to finish preround with empty set.


hare/flows_test.go
```diff
@@ -61,7 +56,7 @@ func (his *HareWrapper) waitForTermination() {
 			break
 		}
 
-		time.Sleep(300 * time.Millisecond)
```
piersy: Why this change?

countvonzero: its just checking whether outputs are there. since all the params are in seconds, i want it do less busy work.

why 300 millisecond in the first place i'd ask :D

hare/flows_test.go
```diff
@@ -462,10 +314,11 @@ func Test_multipleCPs(t *testing.T) {
 		}
 		return true
 	}, 5*time.Second, 10*time.Millisecond)
+	layerDuration := 3 * time.Second
```
piersy: This is quite a significant change from the previous value of 250ms, and I suspect its causing the test to not test what it intended to test which is multiple consensus processes running concurrently.

piersy: Also why change this just for this test and not for `Test_multipleCPsAndIterations`

hare/flows_test.go
```diff
@@ -405,35 +251,42 @@ func Test_multipleCPs(t *testing.T) {
...
 	test.initialSets = make([]*Set, totalNodes)
 
-	dbs := make([]*sql.Database, 0, totalNodes)
-	for i := 0; i < totalNodes; i++ {
-		dbs = append(dbs, sql.InMemory())
-	}
-	pList := make(map[types.LayerID][]types.ProposalID)
+	pList := make(map[types.LayerID][]*types.Proposal)
 	for j := types.GetEffectiveGenesis().Add(1); !j.After(finalLyr); j = j.Add(1) {
-		for i := uint64(0); i < 200; i++ {
```
piersy, line 414: If I run either of the `Test_multiple...` tests with your changes but the old value of 200 here they timeout

hare/flows_test.go
```diff
@@ -277,116 +232,6 @@ func (m *mockClock) advanceLayer() {
 	m.currentLayer = m.currentLayer.Add(1)
 }
 
-type SharedRoundClock struct {
```
piersy, line 280: What was the purpose of this clock?

hare/flows_test.go
```diff
@@ -127,48 +122,12 @@ func (m *p2pManipulator) Publish(ctx context.Context, protocol string, payload [
 	return nil
 }
 
-// Test - runs a single CP for more than one iteration.
-func Test_consensusIterations(t *testing.T) {
```
piersy, line 131: Why move this test?

hare/flows_test.go
countvonzero: no idea. i don't see it valuable as it cause the test not to exercise the actual code.

hare/flows_test.go
countvonzero: yes. like i explained in the PR description, assembling 200 proposals from the mock and get them ready caused ~2sec delay and cause some of the nodes to end preround with empty set, because by the time they get to preround, preround is already over.

cmd/node/node.go
```diff
@@ -358,6 +359,10 @@ func (app *App) Initialize() (err error) {
 		}
 	}
 
+	if app.Config.HARE.N%2 != 0 {
```
piersy: Whats the reason for enforcing even numbers?

hare/flows_test.go
countvonzero: > and I suspect its causing the test to not test what it intended to test which is multiple consensus processes running concurrently.

nope. it still runs multiple CPs concurrently. 3 second of layer duration is still less than 1 sec of wakeup delta + 1 preround + 4 hare rounds.

countvonzero: > Also why change this just for this test and not for Test_multipleCPsAndIterations

that was a mistake. will fix

hare/flows_test.go
countvonzero: i moved it to a consensus_test.go and renamed it to `TestConsensus_MultipleIterations`. tests there are more similar to each other.

cmd/node/node.go
countvonzero: so that in other places of the code i can freely use `app.Config.Hare.N/2` without worries.

the idea is if N = 801. what is the majority (the original f+1)? 401? i guess. 402? maybe better.
this is arguably unnecessary. if you feel strongly i will remove.

hare/flows_test.go
countvonzero: i.e. single iteration. but i can also move it back if you prefer.

hare/flows_test.go
countvonzero: the reason why i increased it is to lessen the artificial pressure cramping all the clock tick together.
i don't see why we should test round duration in the seconds but layer time in the 1/4 second.

hare/interfaces.go
```diff
@@ -27,6 +27,14 @@ type stateQuerier interface {
 	IsIdentityActiveOnConsensusView(context.Context, types.NodeID, types.LayerID) (bool, error)
 }
 
-type nonceFetcher interface {
+type mesh interface {
```
piersy, line 31: Whats the purpose of this interface? My understanding is that is wraps a database and exposes some of the functions, but I don't see the reason to use this instead of just using the database?

hare/flows_test.go
piersy: I don't believe the amount of work being done here in the check is significant compared to the rest of the work being done in these tests, so I'd be in favour of checking more often because that's going to make the tests complete faster.

hare/interfaces.go
countvonzero: testing with a real database is a PITA.

i tolerated it before because before this PR it only needs to generate/save proposals into database.

but now that i have to create the FULL atx for testing, while all i need is basetickcount and tickcount in the header.
@fasmat will sympathize i am sure.

besides, with the real database you don't get to test errors other than `not found` error. tho that's a very small point.

hare/flows_test.go
piersy: I'm not really understanding what you mean by
>artificial pressure cramping all the clock tick together

With the previous layer time this test would have been running the three consensus processes concurrently for most of the test,  with the new time that overlap of processes running concurrently has been reduced, so the second layer would overlap with the first but by the time the third layer starts the first will have finished. So although it may still be testing concurrent consensus processes I'm not sure it's testing them to the same degree as before.

cmd/node/node.go
piersy: I'm not fully understanding. In what case would you use `app.Config.Hare.N/2` ?

countvonzero: the hare protocol tolerates 1/2 of its participants to be malicious or with flaky connection.
in code previous to this PR, we rely on another param F. which should be set to 1/2N.
and in the hare code, we use F+1 as the threshold for a successful termination.

removing param F and use 1/2N+1 in the code remove the opportunity to mis-configure. the devnet now is set to 399 incorrectly.

hare/interfaces.go
piersy: So is the reason that you need to generate a full atx for testing because you need to control the tick heights?

Why not just mock the CacheDB? This mesh construction seems a bit odd to me, its like a little addition to the db, it calls itself mesh, but it seems like its just a db

countvonzero: what is the advantage of mocking CacheDB over this?
- i still need to create a (minimal) interface on the hare (client) side for testing.
- CacheDB is mostly used to cache ATXs/nonce/maliciousness of identities, which will be query a lot during hare. it doesn't have interface for proposals and ballots.

mesh is a general term for mesh data and a abstraction of access to data. it could be that we need to fetch data from peers. hare proper should not care.

hare/flows_test.go
piersy: :+1: sorry!

hare/flows_test.go
piersy: From my understanding it was used to
A) Force round progression only when all participants were ready for the next round
B) Make rounds progress as quick as possible by removing the need to actually wait for time to pass for the round to complete.

The `Test_multiple...` tests are now about 2.5s slower with this PR and I think this could be the reason. However I'm still in favour of the clock removal, it was quite a lot of code and confusion for not much value.

cmd/node/node.go
piersy: What is the number of participants in the devnet?

hare/flows_test.go
countvonzero: yes, but thats just not realistic testing. i dont know why it doesnt used fixed round durations. it also make test failures difficult to reproduce. just cannot see the value here, hence the removal

cmd/node/node.go
countvonzero: dunno. depends on how widely we invite the community. usually on the order of 100 or less i think.

number of nodes are not the same as the number N here. N is the total number of eligibility count, which is proportional to the miner's atx weight. in a network of 50 nodes, like in the systest, we still use 800 for N, and as soon as the eligibility count in the notify round reache s 401, hare terminates.

hare/flows_test.go
countvonzero: you are right. i wasn't considering all 3 processes running at the same time. will fix.

cmd/node/node.go
piersy: Ok so with eligibility of 800, then the majority would be 401 and  F would be 399, right?

hare/interfaces.go
piersy: I wouldn't say there is an advantage to mocking CacheDB over the mesh. But I don't think (with my limited knowledge of the codebase) that the mesh stands alone as a particularly useful construct. If the way that you want to access the database is the current mesh interface, why not just make the database have that interface? I don't see any reason to split the tiny amount of logic that the mesh contains from the database.

Having the mesh now requires implementing several methods that provide no value because they are simply delegating to the database. E.g.
```
func (m defaultMesh) Header(id types.ATXID) (*types.ActivationTxHeader, error) {
	return m.cdb.GetAtxHeader(id)
}
```

cmd/node/node.go
countvonzero: > Ok so with eligibility of 800, then the majority would be 401 and F would be 399, right?

the majority would be 401 yes. and F would be 400.
we want termination condition to be 401 so the counts has to reach F+1.

i think i lost the line of questions here :P it started from why do i want N to be even. do you object?

hare/interfaces.go
countvonzero: make sense to get rid of the delegation. done.

cmd/node/node.go
piersy: I'm just trying to understand how this works.

I think there could be a mistake here though. If you have an eligibility of 800 and your majority is 401 then I think F needs to be 399. Because otherwise if F were 400 and they all abstained from participating then they could permanently halt the protocol, since only 400 would remain and you wouldn't be able to reach the majority of 401.

piersy: Ok so I think the detail that you were lacking here is that F must be _less_ than half the participants.

I got this from the whitepaper:
>In the following we assume that at most f parties are not honest and at least f + 1 parties are honest. In
particular, f is less than half of all active parties (in each round)

countvonzero: > Because otherwise if F were 400 and they all abstained from participating then they could permanently halt the protocol, since only 400 would remain and you wouldn't be able to reach the majority of 401.

hare will fail for this layer if the 400 continue to abstain through iterations until it reached the max iteration limit.
but they will not permanently halt the protocol. the next layer would be already running.

fasmat: For me "majority" is always `N/2 + 1` even if N happens to be odd, but maybe I'm not fully understanding the issue. Is it a problem that `N/2 + 1` and `F` add or do not add up to `N`?

cmd/node/node_test.go
```diff
@@ -981,8 +981,7 @@ func getTestDefaultConfig() *config.Config {
 	// note: these need to be set sufficiently low enough that turbohare finishes well before the LayerDurationSec
 	cfg.HARE.RoundDuration = 2
 	cfg.HARE.WakeupDelta = 1
-	cfg.HARE.N = 5
-	cfg.HARE.F = 2
+	cfg.HARE.N = 6
```
fasmat: To add to the discussion: here majority was 3 (`N/2 + 1`) and `F` was 2.
With `N = 6` majority is now 4 - which seems correct to me.

cmd/node/node.go
piersy: Yes that is a problem, the way I've seen this defined before is  `M=N/2 + 1` and `F=N - M` that leaves no room for error. And then it doesn't matter if N is odd or even (if we're dealing with ints).

countvonzero: for me it's a slight advantage to debug, as i am a little math-challenged :P and easier to divide a round number by 2.
if `N` is odd, the majority is still `N/2 + 1` yes.

this is causing too much confusion. i am removing the check.

piersy: Either way `N/2+1` is a majority :D

countvonzero: removed. sorry for causing such big confusion.

hare/interfaces.go
fasmat: It goes into ownership of the interface. IMO the interface should be designed with the needs of its user in mind, not with the API of its provider.

Just because the API of the database changes we don't want to have to update Hare, that's why it often makes sense to introduce a custom interface and a small adapter.

The adapter is cheap and if the database wasn't our code but a dependency we are using imo the way to go. But because we have full control over Hare and the DB API it depends on our anticipation of what will be more of a PITA in the future:

a) having to update hare in multiple places because we had to update the API of the DB and updating all mocks / tests because the implementation drives the interface or
b) having to update the adapter because the user of the interface drives it design and the implementation has to be adapted to it?

hare/hare_test.go
```diff
@@ -98,6 +94,11 @@ func noopPubSub(tb testing.TB) pubsub.PublishSubsciber {
 	return publisher
 }
 
+func newMockMesh(tb testing.TB) *mocks.Mockmesh {
+	tb.Helper()
+	return mocks.NewMockmesh(gomock.NewController(tb))
```
fasmat, line 99: I think `gomock.NewController` should only be called once per test. So it would be better to pass in the controller here, but at that point we can just completely inline this function.

hare/interfaces.go
countvonzero: agreed. what you said is also what i was alluding to here:

> mesh is a general term for mesh data and a abstraction of access to data. it could be that we need to fetch data from peers. hare proper should not care.

but the effort of edit in the future is not that bad for me, so i adopted @piersy 's recommendation :)

hare/hare_test.go
countvonzero: what's the reason for such restriction tho?

fasmat: gomock isn't super clear in its documentation:

> A Controller represents the top-level control of a mock ecosystem. It defines the scope and lifetime of mock objects, as well as their expectations. It is safe to call Controller's methods from multiple goroutines. Each test should create a new Controller and invoke Finish via defer.

It's also outdated, `Finish` doesn't need to be called at all any more since the introduction of `t.Cleanup`.

I never tried to use multiple controllers so idk what happens in that case. I assumed then possibly missing or unexpected calls would be reported twice?

fasmat: But thanks for making me look it up ❤️ so I saw this: 

https://pkg.go.dev/github.com/golang/mock/gomock#WithContext

I did not know that this exists, this should be super helpful in tests doing concurrency with mocks (otherwise they often get stuck). Maybe this also helps you with `Test_multipleCPs` @piersy ?

countvonzero: 1. the reason i am resisting here is that the hare test setup is convoluted. if i tried to make the same ctrl that will be a huge change.
2. i think the idea of not having to call `Finish` explicitly implies that gomock is tracking the ctrl underneath the cover. in any case, the tests failed if i don't set expectations so it definitely is checking....

fasmat: `NewController` calls `t.Cleanup` with `Finish` if it is supported: https://github.com/golang/mock/blob/aba2ff9a6844d5e3289e8472d3217d5b3090f083/gomock/controller.go#L136-L141

But yeah, looks like its safe to have multiple controllers... 👍

hare/flows_test.go
```diff
@@ -396,7 +241,8 @@ func Test_multipleCPs(t *testing.T) {
 	finalLyr := types.GetEffectiveGenesis().Add(totalCp)
 	test := newHareWrapper(totalCp)
 	totalNodes := 10
-	cfg := config.Config{N: totalNodes, F: totalNodes / 2, WakeupDelta: time.Second, RoundDuration: 5 * time.Second, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 100, Hdist: 20}
+	networkDelay := time.Second
+	cfg := config.Config{N: totalNodes, WakeupDelta: networkDelay, RoundDuration: networkDelay, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 100, Hdist: 20}
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
```
fasmat, line 248: If `cancel` is never explicitly called, `context.WithCancel` is probably not needed.

Maybe here `gomock.WithContext` would be a good fit. The context it returns is canceled when any assert fails.

Further down the test when ranging over channels we can add a `select` that checks if the context is `Done` and if yes returns instead of getting stuck on a channel read that will never happen.

This could fix the issues we have with timeouts on this test, if that's the cause? At least then they would report an error instead of timing out after 5 minutes.

hare/flows_test.go
```diff
@@ -491,7 +348,8 @@ func Test_multipleCPsAndIterations(t *testing.T) {
 	finalLyr := types.GetEffectiveGenesis().Add(totalCp)
 	test := newHareWrapper(totalCp)
 	totalNodes := 10
-	cfg := config.Config{N: totalNodes, F: totalNodes/2 - 1, WakeupDelta: time.Second, RoundDuration: 5 * time.Second, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 100, Hdist: 20}
+	networkDelay := time.Second
+	cfg := config.Config{N: totalNodes, WakeupDelta: networkDelay, RoundDuration: networkDelay, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 100, Hdist: 20}
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
```
fasmat, line 357: see my comment above

hare/flows_test.go
countvonzero: @fasmat do you mind if i leave this for the person (hopefully @piersy) attempting to fix this test?

i don't want to do a half-ass job if i am doing it, but this is just too much change to do it right.
because if you want to do a select later when waiting for termination, that means you have to plug the controller through all mocks. and that affect other tests too because they use the same harness.

fasmat: sure 🙂 I just thought of it when I saw the `WithContext` function in the gomock documentation. I think this could be the issue: some unexpected mock call is made on a go routine that is not the main routine -> test gets stuck but doesn't print why.

countvonzero:
bors try

hare/flows_test.go
fasmat: Refactoring this test is out of scope for this PR. There is also another PR here that is already working on it: https://github.com/spacemeshos/go-spacemesh/pull/4100

hare/flows_test.go
```diff
@@ -573,7 +440,10 @@ func Test_multipleCPsAndIterations(t *testing.T) {
 	for _, out := range outputs {
 		for lid := types.GetEffectiveGenesis().Add(1); !lid.After(finalLyr); lid = lid.Add(1) {
 			require.NotNil(t, out[lid])
-			require.ElementsMatch(t, pList[lid], out[lid].Proposals)
+			require.ElementsMatch(t, types.ToProposalIDs(pList[lid]), out[lid].Proposals)
 		}
 	}
+	for _, h := range test.hare {
+		h.Close()
+	}
```
fasmat: Does this need to happen independently of if the test completes or not? If yes we could use a `t.Cleanup` or `defer` for `h.Close()`.

fasmat (approved):
lgtm

hare/flows_test.go
piersy: No preference just wondering

dshulyak:
is it tested that such proposal id won't be in the hare output? for example if any single node includes it in to the status message, would it be included into svp and eventually get certified?

countvonzero:
bors merge

PR #4124, opened 2023-03-05
TEST ONLY: add public key to hare messages

countvonzero:
## Motivation

- add public key to hare message to measure the effect of not using public key extraction
- add specific metrics to track the public key size

### DIFF
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -549,6 +549,15 @@ func (proc *consensusProcess) sendMessage(ctx context.Context, msg *Msg) bool {
 		return false
 	}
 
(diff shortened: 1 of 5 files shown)


dshulyak:
it is better to add 32 bytes array to every message. []byte slice is encoded with length prefix

countvonzero:
> it is better to add 32 bytes array to every message. []byte slice is encoded with length prefix

done. 

maybe we should do that to SIgnature too?

dshulyak:
> maybe we should do that to SIgnature too?

yes, better to use fixed-size types

PR #4127, opened 2023-03-07
[Merged by Bors] - Bump post v0.5.0

poszu:
## Changes
Bumps post repo to v0.5.0 which includes implementation of post generation in Rust.

## Test Plan
All existing tests should pass


dshulyak (approved):

config/presets/fastnet.go
```diff
@@ -54,11 +54,11 @@ func fastnet() config.Config {
 	conf.POST.BitsPerLabel = 8
 	conf.POST.K1 = 12
 	conf.POST.K2 = 4
-	conf.POST.LabelsPerUnit = 32 // bytes
+	conf.POST.LabelsPerUnit = 128 // bytes
 	conf.POST.MaxNumUnits = 4
 	conf.POST.MinNumUnits = 2
 	conf.POST.N = 32
-	conf.POST.B = 2
+	conf.POST.B = 16
```
fasmat, line 61: Since `B` will now always be 16, should we get rid of it as a parameter?

go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	github.com/spacemeshos/go-scale v1.1.4
 	github.com/spacemeshos/merkle-tree v0.2.1
 	github.com/spacemeshos/poet v0.6.4
-	github.com/spacemeshos/post v0.4.1
+	github.com/spacemeshos/post v0.4.2-0.20230306123131-b29abb1875e8
```
fasmat: Is there a version v0.5.0 of post?

```suggestion
	github.com/spacemeshos/post v0.5.0
```

fasmat (commented):
looks OK, but the referenced version of PoST is not v0.5.0

go.mod
poszu: There was not yet. I wanted first to ensure that the systest pass and only then make a release and bump. Sorry I will leave a note in the description next time.

It's v0.5.0 now.

config/presets/fastnet.go
poszu: The post-proving and verification code technically still supports B other than 16 so I'm leaving it for now. I will remove it once we finish work on the post and stabilize the algo.

poszu:
bors merge

Issue #4095, opened 2023-02-27
revisit collections limits and apply sane max values


dshulyak:
some time ago we added feature to go-scale that allows to limit memory allocation size for collections. the default is set to 1 << 20, so that scale can be used for any types in go-spacemesh. this is arguable a security hole and needs to be addressed.

https://github.com/spacemeshos/go-scale/blob/master/examples/bytes.go#L21-L23

fasmat:
should we move this to the `go-scale` repository?

dshulyak:
we will have to add `scale:"max=XX"` annotations in this repository

PR #4129, opened 2023-03-07
[Merged by Bors] - add max size for public keys in multisig spawn tx according to storage limits

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4095

the only place in vm that requires dyn sized collection, more than 10 keys are not going to be supported.

### DIFF
--- a/genvm/templates/multisig/types.go
+++ b/genvm/templates/multisig/types.go
@@ -9,7 +9,7 @@ import (
 
 // SpawnArguments contains a collection with PublicKeys.
 type SpawnArguments struct {
-	PublicKeys []core.PublicKey
+	PublicKeys []core.PublicKey `scale:"max=10"` // update StorageLimit if it changes.
 }
 
 // Signatures is a collections of parts that must satisfy multisig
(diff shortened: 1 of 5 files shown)


genvm/templates/multisig/types.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // SpawnArguments contains a collection with PublicKeys.
 type SpawnArguments struct {
-	PublicKeys []core.PublicKey
+	PublicKeys []core.PublicKey `scale:"max=10"` // update StorageLimit if it changes.
```
countvonzero, line 12: does the same restriction need to be in `MultiSig` as well?

countvonzero (approved):

genvm/templates/multisig/types.go
dshulyak: this is what is being decoded from user, that stuff is internal state

dshulyak: actually we may enforce it in scalegen, then every collection will have to specify this

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

fasmat (approved):

Issue #2925, opened 2021-11-09
Run unit tests for each package in parallel


nkryuchkov:
## Description

Currently `make test` uses the `-p 1` flag which doesn't allow executing more than one test in parallel. Therefore, unit tests take too long to execute. By default, tests for all packages are run in parallel, where tests in each package are run one by one. This would reduce our test runs by a lot. However, if we remove the flag and test packages in parallel, some tests start failing with these errors:

```
=== RUN   TestValidator_Validate
    nipost_test.go:223: 
        	Error Trace:	nipost_test.go:223
        	            				nipost_test.go:347
        	Error:      	Received unexpected error:
        	            	builder: poet service is unstable: failed to get PoET service ID: response status code: 500, body: {"error":"service not started","code":2,"message":"service not started"}
        	Test:       	TestValidator_Validate
    nipost_test.go:203: 
        	Error Trace:	nipost_test.go:203
        	            				panic.go:642
        	            				testing.go:756
        	            				nipost_test.go:223
        	            				nipost_test.go:347
        	Error:      	Received unexpected error:
        	            	failed to kill process: os: process already finished
        	Test:       	TestValidator_Validate
--- FAIL: TestValidator_Validate (1.39s)
```

acud:
closing due to #6214

PR #4130, opened 2023-03-07
Refactor broker remove race conditions

piersy:
## Motivation
The original motivation for these changes was to be able to improve the speed of our CI, by solving the flakyness in the `Test_multipleCPs...` tests, but that is turning out to be a difficult task. In the meantime I was able to identify and resolve 2 race conditions in the broker and further simplify the broker. Fixing the race conditions in the broker definitely had a positive effect on the flakyness of the `Test_multipleCPs...` tests (I am now unable to reproduce a failure locally) but they still fail on CI and given the length of the feedback loop it will take some time to determine what is now causing those tests to fail.

Relates to #2925

## Changes
Fixed 2 race conditions present in the broker that would cause messages to be lost.  Both of these would occur if a consensus process was registering at the same time as a message was being delivered. In the first case the broker would reject the message mistakenly thinking that registration had failed when in fact registration had simply not fully completed. In the second case messages would be classified as early and put into a map that would be overwritten as the consensus process registered.

Removed knowledge of consensus process limits from the broker, the limit is now enforced by hare.

Made the broker panic if there is an attempt to register an old layer since that would indicate some coding error.

Removed sync check on broker register since this is checked in hare before calling register.

## Test Plan
All the tests seem to be passing

## DevOps Notes


hare/algorithm.go
```diff
@@ -499,7 +499,7 @@ func (proc *consensusProcess) handleMessage(ctx context.Context, m *Msg) {
 
 // process the message by its type.
 func (proc *consensusProcess) processMsg(ctx context.Context, m *Msg) {
-	proc.WithContext(ctx).With().Debug("processing message",
+	proc.WithContext(ctx).With().Warning("processing message",
```
countvonzero: definitely too verbose. please restore back to debug.

hare/broker.go
```diff
@@ -155,21 +112,22 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
 	logger.Debug("broker received hare message")
 
 	msgLayer := hareMsg.Layer
-	if !b.Synced(ctx, msgLayer) {
-		return errNotSynced
-	}
```
countvonzero, line 159: why is this check removed? we don't want to process gossip message when the node is not sync'ed.

hare/broker.go
```diff
@@ -155,21 +112,22 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
+		// We consider messages for the next layer to be early and we process
+		// them, all other messages we ignore. We don't impose the outbox
+		// capacity at this point since this is an externally initiated event
+		// and at most it can cause us to exceed our capacity by one slot.
+		if msgLayer == b.latestLayer.Add(1) {
+			out = make(chan any, inboxCapacity)
+			b.outbox[msgLayer.Uint32()] = out
+		} else {
+			// This error is never inspected except to determine if it is not nil.
+			return errors.New("")
```
countvonzero: i don't think this is a good practice. upstream callers can change and be surprised by the lack of msg in this error.
it should at least contain the layer/round of the message to provide context.

hare/broker.go
```diff
@@ -155,21 +112,22 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
-	isEarly := false
-	if err = b.validateTiming(ctx, &hareMsg); err != nil {
-		if !errors.Is(err, errEarlyMsg) {
-			// not early, validation failed
-			logger.With().Debug("broker received a message to a consensus process that is not registered",
-				log.Err(err))
-			return err
+	out, exist := b.outbox[msgLayer.Uint32()]
+	if !exist {
+		logger.Debug("broker received a message to a consensus process that is not registered, latestLayer: %v, messageLayer: %v, messageType: %v", hareMsg.InnerMsg.Type)
```
countvonzero: the rest of the codebase uses zap's loggable field. it allows for loki to fitler by the loggable field name.

hare/broker.go
```diff
@@ -79,51 +71,16 @@ func (b *Broker) Start(ctx context.Context) {
...
-
-	// future msg
-	return fmt.Errorf("%w: msg %v, latest %v", errFutureMsg, msgInstID, latestLayer)
-}
+var errClosed = errors.New("closed")
 
 // HandleMessage separate listener routine that receives gossip messages and adds them to the priority queue.
 func (b *Broker) HandleMessage(ctx context.Context, _ p2p.Peer, msg []byte) pubsub.ValidationResult {
+	b.mu.Lock()
+	defer b.mu.Unlock()
```
countvonzero: i am not comfortable with locking for the whole duration of processing the gossip msg.

for example, if we are under spam attack, among other things, `case out <- em:` can block.

hare/broker.go
```diff
@@ -62,15 +58,11 @@ func newBroker(
...
 		stateQuerier:    stateQuerier,
-		nodeSyncState:   syncState,
 		mchOut:          mch,
 		outbox:          make(map[uint32]chan any),
-		pending:         make(map[uint32][]any),
 		latestLayer:     types.GetEffectiveGenesis(),
 		limit:           limit,
-		minDeleted:      types.GetEffectiveGenesis(),
 	}
-	b.ctx, b.cancel = context.WithCancel(context.Background())
```
countvonzero, line 72: why remove this?

it is meant for it to be safe to call Close() without calling Start()

hare/broker.go
```diff
@@ -327,124 +267,61 @@ func (b *Broker) HandleEligibility(ctx context.Context, em *types.HareEligibilit
...
+	error,
+) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	layerNum := layer.Uint32()
-	if _, exist := b.pending[layerNum]; !exist { // create buffer if first msg
-		b.pending[layerNum] = make([]any, 0, inboxCapacity)
+	if !id.After(b.latestLayer) { // should expect to update only newer layers
+		b.WithContext(ctx).With().Error("tried to update a previous layer",
+			log.Stringer("this_layer", id),
```
countvonzero: i think it's better to simply use the default field name (i.e just put `id` here)

when debugging, it's useful to search all the loki log where layer_id == blah

hare/broker_test.go
```diff
@@ -708,12 +620,74 @@ func TestBroker_Flow(t *testing.T) {
...
 	b.HandleMessage(context.Background(), "", mustEncode(t, m.Message))
 	<-ch1
 
 	b.Unregister(context.Background(), instanceID1)
-	r.Equal(instanceID2, b.minDeleted)
+	r.Nil(b.outbox[instanceID1.Uint32()])
+}
+
+// Checks that the broker correctly enforces the limit on concurrently registered processes.
+func TestBroker_Limit1(t *testing.T) {
```
countvonzero: can you put a more descriptive name other than Limit1? what's the diff from Limit2

hare/broker.go
countvonzero: ideally if this happen we should log.Fatal this to signal a programming error. 
at the very least it should not hide this programming error and move on.
should log.Error AND return an error.

best to add test for this case.

hare/broker.go
```diff
@@ -327,124 +267,61 @@ func (b *Broker) HandleEligibility(ctx context.Context, em *types.HareEligibilit
...
-		delete(b.outbox, instance.Uint32())
-		b.minDeleted = instance
-		b.With().Info("unregistered layer due to maximum concurrent processes", instance)
-	}
-	outboxCh := make(chan any, inboxCapacity)
-	b.outbox[id.Uint32()] = outboxCh
-	for _, mOut := range b.pending[id.Uint32()] {
-		outboxCh <- mOut
+func sortedKeys(m map[uint32]chan any) []uint32 {
+	keys := make([]uint32, 0, len(m))
```
countvonzero: use `maps.Keys`

hare/broker.go
```diff
@@ -327,124 +267,61 @@ func (b *Broker) HandleEligibility(ctx context.Context, em *types.HareEligibilit
...
-}
 
-// Register a layer to receive messages
-// Note: the registering instance is assumed to be started and accepting messages.
-func (b *Broker) Register(ctx context.Context, id types.LayerID) (chan any, error) {
-	b.setLatestLayer(ctx, id)
+	b.latestLayer = id
 
-	// check to see if the node is still synced
-	if !b.Synced(ctx, id) {
```
countvonzero, line 380: i think this is where i mentioned redundant during our offline discussion. not the gossip handler path.

hare/broker.go
```diff
@@ -327,124 +267,61 @@ func (b *Broker) HandleEligibility(ctx context.Context, em *types.HareEligibilit
...
-	for i := b.minDeleted.Add(1); i.Before(b.latestLayer); i = i.Add(1) {
-		_, exist := b.outbox[i.Uint32()]
-
-		if !exist { // unregistered
-			b.minDeleted = b.minDeleted.Add(1)
-		} else { // encountered first still running layer
-			break
+	oversize := len(b.outbox) - b.limit
+	if oversize >= 0 {
+		// Remove earliest layers to make space for this layer
```
countvonzero: i would leave a TODO here to clean up the consensus process that owns the earliest layer properly.

in essence, the oldest outbox removed will cause confusing error msgs for the ongoing consensus process. it is still running, but doesn't have a outbox to receive msgs.

countvonzero (commented):
very much liked the removal of pending queue for early message.

hare/algorithm.go
piersy: Well spotted, that's some leftover debug assistance, there may be some more that slipped my net.

hare/broker.go
piersy: Right, I wondered about this. So we actually don't want any limit imposed by the broker?

hare/broker.go
piersy: I don't have a clear understanding yet of what it means to be synced, but when a consensus process registers with the broker it checks that it is synced before registering. So it seemed unnecessary to further check this inside the broker, and also seemed like it would be confusing to the consensus process if after registering, messages just stopped being relayed by the broker (which is what would happen if this synced check fails). Also it seems like (I'm guessing here though) if there was a necessity to be synced (i.e. some data needs to be retrieved that can only be retrieved by a synced process) then the consensus process is going to run into that necessity and at that point will understand that it is not synced and that it can't further process messages and maybe Unregister as a result. That seems like a better flow than the broker just silently discarding all the messages for a consensus process.

hare/broker.go
piersy: Agreed :+1:

hare/broker.go
countvonzero: the state of synced means that it has sync'ed all ATXs, ballots, and blocks since genesis
this check is important for a new node syncing from genesis. without this check, a new node will keep seeing errors around ATXs because it doesn't know the identity that send this gossip message.

specifically, a syncing node will not register for a consensus process, but will keep receiving hare gossip messages.

> That seems like a better than the broker just silently discarding all the messages for a consensus process.

this is only true for a node that is already sync'ed and online. if a node were to go from synced -> notSynced state, then the node will silently stopped processing message. but this state transition is detectable during debugging. alternatively, you can detect this (already registered but not sync'ed anymore) and terminate the consensus process with error like `losing synced state`. 

btw, 

https://github.com/spacemeshos/go-spacemesh/blob/6271e39ddafa202c8061fe5143911eb0a0dff419/cmd/node/node.go#L730

already ensures that it is in a synced state before receiving hare gossip message. but it doesn't guarantee the beacon syncedness, which is different by epoch. that's why it's not straight-foward to add there.

hare/broker.go
countvonzero: > So we actually don't want any limit imposed by the broker?

not sure what this mean.

current state is that the limit of concurrent consensus processes is ONLY imposed by the broker, and in a non-obvious way (by removing the inbox and hence dropping all communication).

i think we should do the following in described order)
- call `proc.terminate()` on some error
- remove its inbox

the logic should reside in `onTick()` in `hare.go` where it register new consensus process. if there are too many, remove and ask broker to remove inbox. the whole prune logic isn't even necessary in broker.

hare/broker.go
piersy: By `case out <- em:` are you referring to the point where the message is forwarded to the consensus process?

https://github.com/spacemeshos/go-spacemesh/blob/6271e39ddafa202c8061fe5143911eb0a0dff419/hare/broker.go#L225-L233

Could you go into more detail about how this spam attack would work and what the outcome would be?

hare/broker.go
piersy: So in this case I was maintaining the previous logic that existed in `setLatestLayer`, if we were to return an error here that would mean that once a a process has registered for a layer, you can only register for later layers. Is this the case, or are there situations where you could end up wanting to register a lower layer? What about if tortoise re-orgs and the new chain is at a lower layer?

hare/broker.go
piersy: Sorry, I'm not understanding. What are you suggesting?

hare/broker.go
piersy: Ok got you, in that case I think it would be clearer to only register with the broker if we are below some concurrent consensus process limit. And then we can free the broker from any knowledge of limits or pruning.

hare/broker.go
piersy: My thinking was that it's a mistake to:
* create a broker and then close it without calling Start since it won't have done any work
* create a broker and handle messages before calling start
So it's better to disallow those flows. Now we'll get a panic if we try to close the broker or handle messages before calling Start, which will allow us to address the problem

countvonzero: > So it's better to disallow those flows. Now we'll get a panic if we try to close the broker or handle messages before calling Start, which will allow us to address the problem

i don't see this change as "better". the problems you described are quite unimportant and would be a bug in the startup flow. the fact that `Close()` is called means the app is shutting down. having a panic during the shutdown flow is .... distracting from the actual cause that trigger it to shutdown in the first place.

i think a better change would be initializing broker with a real context obj. 

in any case, i am also fine with this change.

hare/broker.go
countvonzero: say current layer is 10. adversaries start spamming nodes with hare messages of layer 11 and the channel gets filled up. now nodes are not able to process anything due to the lock. but it could be any layers really.

the old code seems to very careful with locking, as hare is performance critical in the code. i don't have a good reason to undo that.

hare/broker.go
countvonzero: no. hare only run one layer at most once. so it SHOULD never go back in time to register an older layer.
the only possibility is that the `Register` is called on a separate goroutine, but that's not the case now.

hare/broker.go
countvonzero: when we discussed offline, i did mention that checking for syncedness here is redundant. i thought that is why you removed the syncedness check everywhere

hare/broker.go
piersy: Ok got you.

The good reason to undo that would be that the old locking code exhibited race conditions which lead to messages being lost.

I think that we could solve that by forwarding the message in a new goroutine, E.G.:

```go
go func() {
    select { 
     case out <- iMsg: 
     case <-ctx.Done(): 
     	    return ctx.Err() 
     case <-b.ctx.Done(): 
 	    return errClosed 
     }
}()
```

What do you think?

hare/broker.go
piersy: Ok cool, will update to return an error here :+1:

hare/broker.go
piersy: Actually maybe better we could use an anonymous function to wrap up to just before the call to the `pubKeyExtractor`, and then defer the unlock in the anonymous function.

piersy: Actually maybe that wouldn't work, because we need to lock in order to safely reference `b.ctx` in `HandleMessage`

countvonzero: i am sympathetic. i also prefer simpler locking scheme. i am a forever opponent for premature optimization.

maybe we should give it a try? i'd go for using a separate goroutine. 
regarding b.ctx, you just need to use errgroup (see hare.go) to manage goroutines.
and when shutting down broker, make sure all outstanding goroutines are terminated.

piersy: I've updated it in [20fb45c](https://github.com/spacemeshos/go-spacemesh/pull/4130/commits/20fb45c3d7f4cf55bf3695b41059ee9844d25003) I think it's still safe, but would appreciate your scrutiny.

hare/broker.go
```diff
@@ -79,51 +75,18 @@ func (b *Broker) Start(ctx context.Context) {
...
-	return fmt.Errorf("%w: msg %v, latest %v", errFutureMsg, msgInstID, latestLayer)
-}
+var errClosed = errors.New("closed")
 
 // HandleMessage separate listener routine that receives gossip messages and adds them to the priority queue.
 func (b *Broker) HandleMessage(ctx context.Context, _ p2p.Peer, msg []byte) pubsub.ValidationResult {
+	// We lock and unlock here to make sure we see b.ctx if it has been set.
+	// See - https://go.dev/ref/mem#locks
+	b.mu.Lock()
+	b.mu.Unlock()
```
countvonzero: i failed to understand the relationship btwn lock/unlock and b.ctx. can you elaborate?

hare/broker.go
```diff
@@ -155,22 +118,31 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
-	}
 
-	isEarly := false
-	if err = b.validateTiming(ctx, &hareMsg); err != nil {
-		if !errors.Is(err, errEarlyMsg) {
-			// not early, validation failed
-			logger.With().Debug("broker received a message to a consensus process that is not registered",
-				log.Err(err))
-			return err
+	if !(b.nodeSyncState.IsSynced(ctx) && b.nodeSyncState.IsBeaconSynced(msgLayer.GetEpoch())) {
```
countvonzero: plz restore prior `b.Synced()` that is also called by hare.go, or something similar. it's better this is shared code by hare and broker

hare/broker.go
```diff
@@ -155,22 +118,31 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
+		err := fmt.Errorf("broker rejecting message because not synced")
+		logger.With().Debug("rejecting message", log.Err(err))
+		return err
```
countvonzero, line 132: i think using the previous `errNotSynced`, instead of creating a new one, is better since there is no extra context provided.

hare/broker.go
piersy: I've re-instated this check for syncedness.

hare/broker.go
piersy: I think using a "real" context object would be an improvement on this change but I still think this change is good on its own. It will highlight areas where we are misusing objects, which in turn will allow us to clarify those code areas. For now I'm going to leave as is since using a "real" context at startup looks like it would be quite a large change.

hare/broker.go
piersy: Actually I'm going to go for a panic in this case because it seems like this would be a coding error if it occurred.

hare/broker_test.go
piersy: I've removed the limit in broker

hare/broker.go
piersy: So we now only check syncedness when handling messages.

hare/broker.go
```diff
@@ -138,7 +100,9 @@ func (b *Broker) HandleMessage(ctx context.Context, _ p2p.Peer, msg []byte) pubs
 
 func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
 	h := types.CalcMessageHash12(msg, pubsub.HareProtocol)
-	logger := b.WithContext(ctx).WithFields(log.Stringer("latest_layer", b.getLatestLayer()), h)
+	b.mu.RLock()
+	logger := b.WithContext(ctx).WithFields(log.Stringer("latest_layer", b.latestLayer), h)
+	b.mu.RUnlock()
```
countvonzero: doing this is error prone. i'd much rather restore getLatestLayer with the pattern
```
lock
defer unlock
```
whenever possible.

hare/broker.go
```diff
@@ -154,22 +118,40 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
-			return err
+		// We consider messages for the next layer to be early and we process
+		// them, all other messages we ignore. We don't impose the outbox
+		// capacity at this point since this is an externally initiated event
+		// and at most it can cause us to exceed our capacity by one slot.
+		if msgLayer == b.latestLayer.Add(1) {
+			out = make(chan any, inboxCapacity)
+			b.outbox[msgLayer.Uint32()] = out
+		} else {
+			b.mu.Unlock()
```
countvonzero: ditto. this is error-prone. please find a way such that locking/unlocking is done safely.
maybe move this block of code out to a separate method so you can access outbox and latestLayer under the same lock.

hare/broker.go
```diff
@@ -154,22 +118,40 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
+			// This error is never inspected except to determine if it is not nil.
+			return fmt.Errorf(
+				"ignoring message to unregistered consensus process, latestLayer: %v, messageLayer: %v, messageType: %v",
```
countvonzero: please try to shorten error msg here, should upstream caller decide to print it :)
maybe just `msg layer not registered`. and it's more consistent with the rest of the code to print round number instead of msg type (can be inferred by by round number)

hare/broker.go
```diff
@@ -326,124 +290,40 @@ func (b *Broker) HandleEligibility(ctx context.Context, em *types.HareEligibilit
...
 
-	if len(b.outbox) >= b.limit {
-		// unregister the earliest layer to make space for the new layer
-		// cannot call unregister here because unregister blocks and this would cause a deadlock
-		instance := b.minDeleted.Add(1)
-		delete(b.outbox, instance.Uint32())
-		b.minDeleted = instance
-		b.With().Info("unregistered layer due to maximum concurrent processes", instance)
+	if !id.After(b.latestLayer) { // should expect to update only newer layers, its a coding error if this happens
+		panic(fmt.Sprintf("broker register called with old layer, current layer: %v, given layer: %v", b.latestLayer, id))
```
countvonzero, line 317: use `b.logger.With().Fatal()` to print the fields and avoid unnecessary stacktrace.

hare/hare.go
```diff
@@ -368,6 +370,13 @@ func (h *Hare) onTick(ctx context.Context, lid types.LayerID) (bool, error) {
 		nonce = &nnc
 	}
 
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	if len(h.cps) >= h.config.LimitConcurrent {
+		err := fmt.Errorf("failed to start cp for layer %q hare reached concurrent cp limit %q", lid.Uint32(), h.config.LimitConcurrent)
+		logger.With().Warning("hare failure", log.Err(err))
```
countvonzero: log the details (layer and limit) here and only propagate the error with minimal content.
errors are usually wrapped many levels by the time it propagates up. too long of error msg will clogged up upstream logging.

hare/hare.go
```diff
@@ -368,6 +370,13 @@ func (h *Hare) onTick(ctx context.Context, lid types.LayerID) (bool, error) {
...
+		return false, err
```
countvonzero: this is incorrect. what we want is to kill the oldest consensus process and make room for the new one. later consensus is more important than old ones. if the old one fails locally, it's likely we can already get a block certificate from the network if other nodes succeed. 

i'd suggest leaving that as a TODO earlier. i think it's better doing that in a separate PR.

hare/hare.go
```diff
@@ -108,6 +108,7 @@ type Hare struct {
 	broker     *Broker
 	sign       *signing.EdSigner
 	blockGenCh chan LayerOutput
+	syncState  system.SyncStateProvider // provider function to check if the node is currently synced
```
countvonzero, line 113: it's less error prone to delegate the synced'ness check to broker to ensure they check for the the same condition.

hare/broker.go
piersy: I've expanded the comment in [3ed0ad4](https://github.com/spacemeshos/go-spacemesh/pull/4130/commits/3ed0ad49fdc77894f827f44bcdefc51b1bb87562)

hare/broker.go
piersy: The stack trace is useful to understand the context of the failure, it lets you know at least what component called this code, that's why I didn't use log.Fatal and why I would suggest not using log.Fatal except in cases where the stack is very obvious, e.g. parsing commandline args.

hare/hare.go
piersy: Yep this is wrong, but trying to re-instate the old logic, which was also broken would be quite a lot of work and I'm not seeing that it's worth doing that just to have a separate PR, when the fix here would be quite simple. What do you think?

hare/broker.go
piersy: I don't see a reason to create an error value for things that are used only once, also I have found that error values can deter people from adding context to errors, which can be detrimental to understanding problems when they occur.

hare/hare.go
piersy: I've extracted what previously constituted the broker's  `Synced` to a single method of the `syncState`, so I think the current approach is pretty much equivalent to what you are suggesting now.

hare/hare.go
piersy: The problem I see with returning simple errors is that you lose the context of the error, so when it reaches the top level and you get the error `hare reached concurrent limit` you're in the dark as to what actually happened. Logs can help but if you have multiple processes or threads you may be unable to determine what caused the error you are looking at. Also for testing we often don't have logs enabled so it can be really helpful in determining why a test fails when the error contains all the context.

I'm not saying that there couldn't be a problem with upstream logging being clogged, but I think there is a balance to be met between adding contextual information to returned errors and deciding what to log.

Just having a blanket policy of returning simple errors and logging details can lead to logs that don't let you understand what actually went wrong.

In this case the result of `onTick` is ignored except in tests so I don't see a longer error message here being a problem.

hare/broker.go
countvonzero: i still don't understand why this guarantee that `b.ctx` to be set.
it would be easier and intuitive to simply return/do nothing if b.ctx == nil

hare/broker.go
countvonzero: what do you mean used only once? 
this is on the gossip path. it will be called many times before a node eventually become synced.

hare/broker.go
countvonzero: i disagree. there is only one caller of Register. i can tell exactly the stack trace in this case. 
stack trace just filled up the log volume without extra information.

hare/hare.go
countvonzero: sure if you feel comfortable doing so.

i just want to caution that changing too much behavior in one PR is not rollback friendly. but up to you.

hare/broker.go
piersy: I'm not understanding the problem with having a long message here? As a caller if something goes wrong here the context of why it went wrong is useful to deciphering the problem. As a node operator or a dev debugging something I'd much rather have a long message that let me understand better what went wrong than a short message that doesn't help me understand the source of the problem.

piersy: E.g: 
```
ignoring message to unregistered consensus process, latestLayer: 0, messageLayer: 1237643, round: 0
```
would immediately let me know something was badly wrong with the sync or startup of the node. As opposed to

```
ignoring message to unregistered consensus process
```

which would leave me in a position where I need to start re-creating the issue, adding extra debug logs, maybe debugging ... etc. Before reaching the same understanding that was immediately apparent from the error.

hare/broker.go
piersy: I've wrapped this in a function call

hare/broker.go
piersy: Also wrapped in a function call

hare/hare.go
countvonzero: your argument seems to contradicts itself...

if you want more context for debugging, then you should log the context instead of packing it in the error object, since the caller of onTick() ignore the error.

for me there is no hard rule in how much you log and pack in the error. the responsibility of logging should rest on the entity that handles errors. if the upstream caller handles the error, then i support packing the context in the error itself. but here it's not the case.

hare/broker.go
piersy: I've added a new function to the `nodeSyncState` for this.

hare/broker.go
```diff
@@ -154,21 +136,47 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
+	var out chan any
+	err = func() error {
+		b.mu.Lock()
+		defer b.mu.Unlock()
+		var exist bool
+		out, exist = b.outbox[msgLayer.Uint32()]
+		if !exist {
+			logger.With().Debug(
+				"broker received a message to a consensus process that is not registered",
+				msgLayer.Field(),
```
countvonzero, line 144: just `msgLayer` is enough

hare/broker.go
```diff
@@ -154,21 +136,47 @@ func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
...
+				log.Stringer("msg_type", hareMsg.InnerMsg.Type),
```
countvonzero: can you log round instead, here and below?

hare/broker.go
countvonzero: you can always grep for the exact text when you are debugging.... i don't see why long msg is different from short msg. the key here is the extra context (layer and round)

hare/broker.go
piersy: My original thinking was that trying to handle a message before the broker has been started is an error, or at best a waste of resources. The sequence should be

1 . Make a new broker
2.  Start the broker
3.  Handle messages with the broker

So I thought it would be an improvement to fail if the broker starts to handle messages before it has been started. Then we'd be aware of the failure to correctly start up the broker.

So this code does not guarantee that the ctx has been set it may not have been, all it guaranteed was that if the context had been set, that change would be visible here.

However looking into the code that we have in node it looks like this problem currently exists, it seems we register the broker handler with libp2p before Start is called. So given that I think your suggestion is probably the best option for now, longer term it could be nice to actually start things up in an order that makes sense.

hare/broker.go
piersy: I just meant that there is only one place in the code that would use this error, so there's not a benefit to extracting it into a variable.

hare/broker.go
piersy: I count 26 references to `Register` in our tests and if this were to happen in a test just calling log.Fatal might leave you wondering where the problem occurred, with a stack trace you'd know the line in the test that caused the problem.

Given that the node will be shutting down at this point, the addition of a few more lines to the log seems like an insignificant addition, not something to worry about.

Maybe I'm misunderstanding something about how our logging works?

hare/hare.go
piersy: What's different here as opposed to [this comment](https://github.com/spacemeshos/go-spacemesh/pull/4130#discussion_r1128497406), where I thought you were saying that because the caller can change it's good practice to include all the context in the error irrespective of what the current caller is doing with it?

hare/broker.go
countvonzero: > Given that the node will be shutting down at this point, the addition of a few more lines to the log seems like an insignificant addition, not something to worry about.

by that same argument, i'd say reducing the log is also not something to worry about.

tests are not really my consideration. 
my concerns are about log volumes in grafana when we actually debug. or from the user submitted log.
the smaller the better. we are doing a concerted effort to reduce log volume and this does not help.

hare/hare.go
countvonzero: like i explained. the diff is who gets to log it.

if you are logging it anyway, then you log the context and/or pack the context in the error
if you are passing the error upstream and not log it, then you pack the context in the error.

this discussion seems longer than it needs to be. lets add the contexts (layer/limit) in the log message also. i don't really care (at this point) what's in the error.

hare/broker.go
```diff
@@ -138,7 +103,13 @@ func (b *Broker) HandleMessage(ctx context.Context, _ p2p.Peer, msg []byte) pubs
 
 func (b *Broker) handleMessage(ctx context.Context, msg []byte) error {
 	h := types.CalcMessageHash12(msg, pubsub.HareProtocol)
-	logger := b.WithContext(ctx).WithFields(log.Stringer("latest_layer", b.getLatestLayer()), h)
+	var logger log.Log
+	func() {
+		b.mu.RLock()
+		defer b.mu.RUnlock()
+		logger = b.WithContext(ctx).WithFields(log.Stringer("latest_layer", b.latestLayer), h)
+	}()
```
countvonzero, line 111: this is def. clever.
but i must over-using anonymous func. makes the code harder to read.

hare/broker.go
countvonzero: it's mostly because we don't check the error during tests at the moment. if one is testing this method, then specific error would be desired. in any case, not important.

piersy:
Closing this, to address these issues in smaller PRs

Issue #4112, opened 2023-03-01
investigate partition test failure 


countvonzero:
## Description

this failure https://github.com/spacemeshos/go-spacemesh/actions/runs/4300663720/jobs/7497080392 
is from this PR https://github.com/spacemeshos/go-spacemesh/pull/4105
namespace `test-havd`

the PR should not impact tortoise. same partition-50-50 test was run in systemtest nodepool 5 times in a row and passed.
tortoise from two sides disagreed on layer 16. i checked boot-0-0 and boot-1-0, they received the same amount of ballots (869 of them)

dshulyak:
so far i don't see any bug in the tortoise. maybe it will be worth talking about beacon<>self-healing interaction, it doesn't seem to be nice.

block was generated at layer 16 on one side of the partition (boot-1-0 side):
![image](https://user-images.githubusercontent.com/1964576/222693349-3babd9b7-95df-4253-b05a-2cf242a6f05d.png)

that side voted for it and confirmed it early
![image](https://user-images.githubusercontent.com/1964576/222693468-409836da-5ec4-483c-8c3d-983ab21c1a59.png)

based on the beacon value partition lasted starting at from layer 20 till layer 30. logs are a bit verbose, but here they are [ballot has different beacon logs](https://grafana.spacemesh.dev/explore?orgId=1&left=%7B%22datasource%22:%22loki-gke-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-gke-dev%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bnamespace%3D%5C%22test-havd%5C%22,%20pod%3D%5C%22boot-1-0%5C%22%7D%20%7C%3D%20%5C%22trtl%5C%22%20%7C%3D%20%5C%22ballot%20has%20different%5C%22%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221677631495491%22,%22to%22:%221677677594254%22%7D%7D)

we delay counting ballots with different beacons to prevent concentration of the weight, therefore nodes on the side of the boot-0-0 are voting against the block, during the time while partition lasts.

later when there is no difference in beacon partition on the side of boot-0-0 seems to terminate the same way as boot-1-0 side.

[...]

PR #4131, opened 2023-03-07
[Merged by Bors] - systest: reduce ballot delay in systest for diff beacon

countvonzero:
## Motivation

related to #4112

speed up systest partition test recovery

- make the delay configurable with default value of 1 epoch if not set
- set value in the preset configs
- reduce the partition duration from 2 to 1 epoch in test

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -525,7 +525,6 @@ func (app *App) initServices(
 
 	trtlCfg := app.Config.Tortoise
 	trtlCfg.LayerSize = layerSize
-	trtlCfg.BadBeaconVoteDelayLayers = app.Config.LayersPerEpoch
 	trtl := tortoise.New(app.cachedDB, beaconProtocol,
 		tortoise.WithContext(ctx),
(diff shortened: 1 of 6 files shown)


countvonzero:
@dshulyak hopefully this will reduce the failure rate for 50-50 test.

dshulyak:
it looks like that it is unconditionally overwritten to number of layers in the epoch

image

countvonzero:
@dshulyak thanks! can you review again. i have
- remove the overwrite of value and make it configurable
- set value in the preset configs
- reduce the partition duration from 2 to 1 epoch in test

dshulyak (approved):
@countvonzero what do you think about setting it to epoch length if value is zero (and make it zero by default)? it is unlikely will be anything else than epoch length in testnets

countvonzero:
> @countvonzero what do you think about setting it to epoch length if value is zero (and make it zero by default)? it is unlikely will be anything else than epoch length in testnets

sure. make sense. done

countvonzero:
bors merge

PR #4132, opened 2023-03-08
[Merged by Bors] - Remove commitment ATX from DB and use post_metadata.json

### LINKED ISSUE #3843: Use commitment ATX from post_metadata.json instead of from DB
## Description

The commitment ATX for PoST initialization is stored in the metadata contained in `postdata_metadata.json`. A node should use the following order of priorities in determining it's own commitment ATX:

1. From their own initial ATX. (PoST setup was previously completed)
2. From the metadata available in the PoST `dataDir`. (PoST setup is currently ongoing)
3. Chose an ATX with the highest weight and use it to initialize PoST. (PoST not yet available)

At the moment a node reads the commitment ATX from the database when PoST setup has started but not yet completed. This can lead to an inconsistent state and importing PoST data generated on a different machine impossible (or at least more difficult, since the state database has to be manually updated to contain the same value for the commitment ATX as `post_metadata.json`).

## Affected code

- Update code in the `activation` package.
- During startup when starting the `atxBuilder` Session: `StartSmeshing` doesn't proceed until all ATXs are synched. We should add an additional check that when a commitment ATX is already available in the PoST metadata this check also passes.
   - otherwise an already initialized node that was offline for some time will hang during startup waiting for ATXs to be synced when it's not actually needed.

fasmat:
## Motivation
Closes #3843 

## Changes
Instead of the ATX `Builder` selecting the commitment ATX, the `PostSetupManager` should take the responsibility. This simplifies several things:

- Instead of needing to store the selected commitment ATX in the DB between selecting and publishing it, the source of truth is now `post_metadata.json` or (if missing) the initial ATX of the node. As before the ATX with the highest weight will be selected if the node has never published an initial ATX before. This removes the need to store the commitment ATX in two places.
- to determine the Commitment ATX `PostSetupManager::StartSession` now has to be called, which selects the ATX with the highest weight in case the node has not previously published any ATX and hasn't started initialisation.

## Test Plan
- All existing tests still pass
- Tests for the logic to select the correct commitmentATX have been moved / updated.

## TODO

## DevOps Notes


activation/post_test.go
```diff
@@ -256,10 +256,76 @@ func TestPostSetupManager_Stop_WhileInProgress(t *testing.T) {
...
-	req.NoError(mgr.StartSession(context.Background(), opts, goldenATXID))
+	req.NoError(mgr.StartSession(context.Background(), opts))
 
 	// Verify status.
 	status = mgr.Status()
 	req.Equal(PostSetupStateComplete, status.State)
 	req.Equal(uint64(opts.NumUnits)*cfg.LabelsPerUnit, status.NumLabelsWritten)
 }
+
+// func TestBuilder_findCommitmentAtx_UsesLatestAtx(t *testing.T) {
```
fasmat, line 267: TODO: update tests after moving commitment ATX selection logic from builder to postsetupmanager

activation/nipost_test.go
```diff
@@ -53,7 +53,7 @@ func (p *postSetupProviderMock) Benchmark(PostSetupComputeProvider) (int, error)
 	return 0, nil
 }
 
-func (p *postSetupProviderMock) StartSession(_ context.Context, _ PostSetupOpts, _ types.ATXID) error {
+func (p *postSetupProviderMock) StartSession(_ context.Context, _ PostSetupOpts) error {
```
fasmat, line 56: TODO: consider replacing this mock with a generated one

fasmat:
bors try

fasmat:
bors try

countvonzero:
need some clarification in the PR description.

> or (if missing) the initial ATX of the node. 

if missing it should either get it from the initial ATX of the node. failing that, select the atx with highest tick height?

> an initial ATX can now not be published any more unless a node also starts smeshing. It didn't make sense before to do this but was possible.

i don't understand this sentence. do you mean the current code path will publish an ATX regardless if the node is smeshing?

fasmat:
@countvonzero 

>> or (if missing) the initial ATX of the node.
>
> if missing it should either get it from the initial ATX of the node. failing that, select the atx with highest tick height?

This is correct. A node still selects the ATX with the highest weight as commitment ATX if it hasn't published an initial ATX before. Updated description

>> an initial ATX can now not be published any more unless a node also starts smeshing. It didn't make sense before to do this but was possible.
>
> i don't understand this sentence. do you mean the current code path will publish an ATX regardless if the node is smeshing?

Before the ATX builder could select the commitment ATX before `StartSmeshing` was called, possibly committing to an ATX long before the node starts initialising. This would also allow to the node to publish an initial ATX without starting initialisation. Since now the commitment ATX is selected in `StartSmeshing` this is not possible any more.

Its the difference between a) do I publish my initial ATX first and then start initialising or b) start initialising and then publish my initial ATX.

activation/post.go
```diff
@@ -181,13 +186,8 @@ func (mgr *PostSetupManager) StartSession(ctx context.Context, opts PostSetupOpt
...
 	var err error
-	mgr.commitmentAtxId, err = mgr.getCommitmentAtx(opts.DataDir)
-
-	switch {
-	case errors.Is(err, initialization.ErrStateMetadataFileMissing):
-		// TODO(mafa): commitmentAtx shouldn't be passed in as an argument. See: https://github.com/spacemeshos/go-spacemesh/issues/3843
-		mgr.commitmentAtxId = commitmentAtx
-	case err != nil:
+	mgr.commitmentAtxId, err = mgr.commitmentAtx(opts.DataDir)
+	if err != nil {
```
countvonzero, line 233: the locking scheme in this method is just baddddddd.

countvonzero:
> Its the difference between a) do I publish my initial ATX first and then start initialising or b) start initialising and then publish my initial ATX.

i cannot find the codepath where the node would publish an initial ATX without smeshing. can you point me to it?

fasmat:
Looks like you are correct. 🙂 

`Builder::getCommitmentAtx` used to lock in the choice of the Commitment ATX, but it was only called called in `Builder::StartSmeshing` and `Builder::buildNIPostChallenge`, which was also only (indirectly) called by `Builder::StartSmeshing`.

Now `PostSetupManager::StartSession` commits to an ATX for initialisation (instead of having it passed in from the builder).

countvonzero:
> Looks like you are correct. slightly_smiling_face
> 
> `Builder::getCommitmentAtx` used to lock in the choice of the Commitment ATX, but it was only called called in `Builder::StartSmeshing` and `Builder::buildNIPostChallenge`, which was also only (indirectly) called by `Builder::StartSmeshing`.
> 
> Now `PostSetupManager::StartSession` commits to an ATX for initialisation (instead of having it passed in from the builder).

thanks for confirming. can you update the PR description for the record? thanks

fasmat:
bors try

activation/post.go
```diff
@@ -251,20 +253,62 @@ func (mgr *PostSetupManager) StartSession(ctx context.Context, opts PostSetupOpt
...
 
+// findCommitmentAtx determines the best commitment ATX to use for the node.
+// It will use the ATX with the highest height seen by the node and defaults to the goldenATX,
+// when no ATXs have yet been published.
+func (mgr *PostSetupManager) findCommitmentAtx(ctx context.Context) (types.ATXID, error) {
+	// wait for ATX to be synced before selecting commitment ATX
+	select {
+	case <-ctx.Done():
+		return *types.EmptyATXID, ctx.Err()
+	case <-mgr.syncer.RegisterForATXSynced():
```
countvonzero: i think it's better to check atx synced before `StartSession` is invoked. it's leaking unrelated app logic into PostSetupManager.

activation/post.go
fasmat: I tried to improve it

fasmat:
bors try

activation/post.go
```diff
@@ -158,21 +163,21 @@ func (mgr *PostSetupManager) Benchmark(p PostSetupComputeProvider) (int, error)
...
-// It supports resuming a previously started session, as well as changing the Post setup options (e.g., number of units)
-// after initial setup.
-func (mgr *PostSetupManager) StartSession(ctx context.Context, opts PostSetupOpts, commitmentAtx types.ATXID) error {
+// StartSession starts (or continues) a PoST session. It supports resuming a previously started
+// session, and will return an error if a session is already in progress.
+//
+// Ensure that before calling this method, the node is ATX synced.
+func (mgr *PostSetupManager) StartSession(ctx context.Context, opts PostSetupOpts) error {
 	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
```
countvonzero, line 215: i think it would be better to break StartSession into 2 smaller methods.
```go
StartSession() {
  prepare()
  initialize()
}
```

countvonzero (approved):

fasmat:
bors merge

PR #4134, opened 2023-03-08
[Merged by Bors] - add domain to signature

### LINKED ISSUE #3967: malfeasance proof: prevent cross-data types misuse of eligibility

countvonzero:
## Motivation
Closes #3967

## Changes
- add domain in signer and key extractor
- move key extraction out of types package to break circular deps btwn signing and types
- remove some (un-mocked) signer interface so that we can use this trick https://github.com/spacemeshos/go-spacemesh/pull/4096#issuecomment-1458269265
- (unrelated) use eligibility type for beacon vrf msg

note: this is an alternative impl of https://github.com/spacemeshos/go-spacemesh/pull/4096


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
@dshulyak the change is indeed smaller than the original PR. thanks

countvonzero:
bors try

signing/keyExtractor.go
```diff
@@ -41,10 +41,11 @@ func NewPubKeyExtractor(opts ...ExtractorOptionFunc) (*PubKeyExtractor, error) {
...
 
 // Extract public key from a signature.
-func (e PubKeyExtractor) Extract(m, sig []byte) (*PublicKey, error) {
-	msg := make([]byte, len(m)+len(e.prefix))
+func (e PubKeyExtractor) Extract(d domain, m, sig []byte) (*PublicKey, error) {
+	msg := make([]byte, len(m)+len(e.prefix)+1)
 	copy(msg, e.prefix)
-	copy(msg[len(e.prefix):], m)
+	msg[len(e.prefix)] = byte(d)
+	copy(msg[len(e.prefix)+1:], m)
```
dshulyak: with append will be easier to see that it is correct

```
msg := make([]byte, 0, len(es.prefix)+1+len(m))
msg = append(msg, es.prefix...)
msg = append(msg, byte(d))
msg = append(msg, m...)
```

signing/signer.go
```diff
@@ -13,6 +13,15 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
+type domain byte
+
+const (
+	ATX domain = iota
+	BEACON
+	BALLOT
+	HARE
```
dshulyak, line 22: @poszu do we need to add domain for poet? there was something about validation

dshulyak (approved):

signing/signer.go
countvonzero: i think poet forward the whole opaque blob for a node to verify the challenge. and it's done in `ChallengeVerifier`.
is that what you are referring to?

countvonzero:
bors merge

signing/signer.go
dshulyak: i remember there was a discussion about signing something for the poet using smesher key, if so we need to use domain for that signature as well

activation/nipost.go
```diff
@@ -159,7 +160,7 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.PoetC
 		if err != nil {
 			return nil, 0, err
 		}
-		signature := nb.signer.Sign(challenge)
+		signature := nb.signer.Sign(signing.ATX, challenge)
```
dshulyak: @countvonzero this is what i was talking about. it is safer to use different domain here, otherwise we need to carefully consider what is the structure of this challenge and ensure that it can't be decoded as a valid atx metadata

countvonzero: ok. understood. will change to be safe.

countvonzero:
bors try

countvonzero:
bors merge

Issue #3781, opened 2022-11-22
flaky test: TestForkFinder_FindFork_Permutation


countvonzero:
## Description

https://github.com/spacemeshos/go-spacemesh/actions/runs/3520291379/jobs/5901072490

not able to reproduce :(

```
=== FAIL: syncer TestForkFinder_FindFork_Permutation (0.01s)
    find_fork_test.go:146: 
        	Error Trace:	/home/runner/work/go-spacemesh/go-spacemesh/syncer/find_fork_test.go:146
        	Error:      	Received unexpected error:
        	            	peer mesh changed mid session
        	Test:       	TestForkFinder_FindFork_Permutation
```

countvonzero:
no longer observed

PR #4138, opened 2023-03-09
[Merged by Bors] - TestForkFinder: make constant layer hash for a given layer

countvonzero:
## Motivation
attempt for #3781

## Changes
make each layer hash constant:
- peer hashes are Hash32 of "good/"+layerNum
- node hashes are Hash32 of "good/"+layerNum before the diverging layer, and Hash of "bad/"+layerNum at and after the diverging layer

### DIFF
--- a/syncer/find_fork_test.go
+++ b/syncer/find_fork_test.go
@@ -74,35 +74,27 @@ func TestForkFinder_Purge(t *testing.T) {
 	require.Equal(t, 0, tf.NumPeersCached())
 }
 
-func createPeerHashes(max uint32) []types.Hash32 {
-	peerHashes := make([]types.Hash32, max+1)
-	gLid := types.GetEffectiveGenesis()
-	for i := uint32(0); i <= max; i++ {
-		lid := types.NewLayerID(i)
-		if lid.Before(gLid) {
-			peerHashes[i] = types.Hash32{}
-		} else {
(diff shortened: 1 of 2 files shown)


syncer/find_fork_test.go
```diff
@@ -74,48 +74,41 @@ func TestForkFinder_Purge(t *testing.T) {
...
-			peerHashes[i] = types.Hash32{}
-		} else {
-			peerHashes[i] = types.RandomHash()
-		}
+func layerHash(layer int, good bool) types.Hash32 {
+	if good {
+		return types.BytesToHash([]byte(strconv.Itoa(layer)))
 	}
-	return peerHashes
+	return types.BytesToHash([]byte(strconv.Itoa(2 * layer)))
```
dshulyak: do you want to use even as bad, and odd as good? is it ok that good=2, will be equal to bad=1?

maybe prefixing them will be more obvious 
  
```
var h2 types.Hash32
h := h2[:0]
h = append(h, "good/"...)
binary.LittleEndian.AppendUint32(h, 72)
```

countvonzero: thanks for sharing. done.

good here implies that the peer always has the correct layer hash, and local node has the incorrect one.

dshulyak (approved):

countvonzero:
bors merge

PR #4144, opened 2023-03-10
hare: only process msg that is 2 level deep

countvonzero:
## Motivation
disallow further processing messages that contain msg more than 2 layers

## Changes
- don't process hare msg that has 2 level of nested msg

note: the node still suffer from having to decode such msg :(

### DIFF
--- a/hare/builder.go
+++ b/hare/builder.go
@@ -2,6 +2,7 @@ package hare
 
 import (
 	"encoding/hex"
+	"errors"
 	"fmt"
 
 	"github.com/spacemeshos/go-spacemesh/codec"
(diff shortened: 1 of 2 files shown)


dshulyak (approved):
i don't have issue with merging it, but it is not useful on its own (it really doesn't do anything)

decoder either needs to limit depth and then this check is redundant, or message type have to be refactored and then there is no way to write this check

countvonzero:
> i don't have issue with merging it, but it is not useful on its own (it really doesn't do anything)
> 
> decoder either needs to limit depth and then this check is redundant, or message type have to be refactored and then there is no way to write this check

yeah... i also noted it in the PR description.... i started working on the msg type refactoring but it's just not trivial...

Issue #4001, opened 2023-01-24
Flaky Test_multipleCPsAndIterations UT


poszu:
## Description
Unit test `Test_multipleCPsAndIterations` is flaky

Example CI run: https://github.com/spacemeshos/go-spacemesh/actions/runs/3997966599/jobs/6859990376

fasmat:
Test_multipleCPs is also flaky, probably for the same reasons.

piersy:
AFAICS these haven't been particularly flaky on CI for a while now, but they become very flaky if trying to run tests without
the `-p 1` flag

PR #4145, opened 2023-03-10
[Merged by Bors] - Mitigate Test_multipleCPs.. timeout

piersy:
## Motivation
The `Test_multipleCPs..` tests seem to be timing out on develop, I believe this is as a result of removing the `SharedRoundClock` in #4121.

## Changes
Increased the timeouts for these tests.

## Test Plan
See that CI passes

## DevOps Notes

### DIFF
--- a/hare/flows_test.go
+++ b/hare/flows_test.go
@@ -241,7 +241,7 @@ func Test_multipleCPs(t *testing.T) {
 	finalLyr := types.GetEffectiveGenesis().Add(totalCp)
 	test := newHareWrapper(totalCp)
 	totalNodes := 10
-	networkDelay := time.Second
+	networkDelay := time.Second * 4
 	cfg := config.Config{N: totalNodes, WakeupDelta: networkDelay, RoundDuration: networkDelay, ExpectedLeaders: 5, LimitIterations: 1000, LimitConcurrent: 100, Hdist: 20}
 
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


countvonzero:
> I believe this is as a result of removing the SharedRoundClock in https://github.com/spacemeshos/go-spacemesh/pull/4121.

how do you determine this? 
https://github.com/spacemeshos/go-spacemesh/issues/4001 already documented such timeout.

piersy:
@countvonzero 

The test is now failing fairly regularly whereas before I had not witnessed it failing when running tests with `-p 1`. See these recent jobs:
https://github.com/spacemeshos/go-spacemesh/actions/runs/4388352031/jobs/7684748852
https://github.com/spacemeshos/go-spacemesh/actions/runs/4387428276/jobs/7682715479
https://github.com/spacemeshos/go-spacemesh/actions/runs/4387765368/jobs/7683451294

https://github.com/spacemeshos/go-spacemesh/pull/4121 is the most recent PR to change that test.

hare/flows_test.go
```diff
@@ -434,7 +435,8 @@ func Test_multipleCPsAndIterations(t *testing.T) {
 		}
 	}()
 
-	test.WaitForTimedTermination(t, 100*time.Second)
+	// There are 5 rounds per layer and totoalCPs layers and we double for good measure.
+	test.WaitForTimedTermination(t, 2*networkDelay*5*time.Duration(totalCp))
```
countvonzero: shouldn't this be longer if we expect one of the round will fail, hence causing the 2nd iteration?

countvonzero (approved):

hare/flows_test.go
piersy: Yes I think it should, would we want to double it again then, to account for each consensus process running twice?

piersy: Actually I see that its just one layer that runs more than a single iteration, so I think we can just increase the timeout for one more layer.

piersy:
bors merge

PR #4147, opened 2023-03-13
Add pre-commit git hook

piersy:
## Motivation

I found this pre-commit hook useful, it saved me from waiting for needless CI runs. It's not exhaustive, but it only takes a few seconds on my machine.

## DevOps Notes

### DIFF
--- a/githooks/pre-commit
+++ b/githooks/pre-commit
@@ -1,0 +1,7 @@
+#!/bin/sh
+set -e
+
+make test-fmt
+make test-tidy
+make lint
+


dshulyak (dismissed):
i would prefer if this was made optional.
i am making commits very frequently when working on something non-trivial, waiting couple of seconds doen't seem good

piersy:
@dshulyak It is optional, to enable the hook you would need to symlink to the added file in the `.git/hooks` directory.  I.e. run `ln -s ../../githooks/pre-commit` from inside `.git/hooks`

countvonzero:
> @dshulyak It is optional, to enable the hook you would need to symlink to the added file in the `.git/hooks` directory. I.e. run `ln -s ../../githooks/pre-commit` from inside `.git/hooks`

hmmm.... don't you then need to add .git/hooks to the ignored files? why not simply do that and you can do whatever in .git/hooks?

fasmat:
@countvonzero the `.git/hooks` directory is not added to the index by default.

I'm not a huge fan of hooks running part of the CI locally, it makes working with git much slower. I feel like `make test-fmt` or `make test-tidy` can become quit annoying on every commit, since I like to amend and rebase commits before I push them and these checks would interfere with my workflow 🙁 .

Since the hook is currently not enforced, I don't mind adding it. But at least it should then be mentioned in the `README.md`.

pigmej:
Few cents from my side, most of the IDEs or even editors have the possibility to run hooks on save / modify / focus out etc. I don't think adding such to pre-commit is that valuable nowadays, honestly.

piersy:
@pigmej yes you can do this just locally through your IDE, but I thought it could be useful for people to just have the hook pre-defined. Otherwise it takes some time to figure out what would be a good hook. It's also useful for people who are using the hook to have it managed with the repository.

That said it sounds like no-one else who has looked at this PR is going to find this useful, so I'm closing it.

Issue #4089, opened 2023-02-26
implement hare active set


countvonzero:
## background

prior to unified content block, the hare active set was derived from contextually valid blocks verified by tortoise, hence are in consensus.
after unified content block, the hare active set was derived from first ballot of epoch from smeshers. since nodes do not have consensus on ballots, this active set is not in consensus

## proposal

- implement https://community.spacemesh.io/t/grading-atxs-for-the-active-set/335
- hare active set for epoch E will be derived from the first block in epoch E in the following way:
  - take the proposals produced by the ATXIDs recorded in the block
  - the union of the active set referenced by said proposals is used as the hare active set for epoch E (plus safe layer range)
- newly synced node (not online last epoch to receive ATXs real time) will use the first block of current epoch and for the active set in its ballot

## bootstrapping and recovery
a spacemesh signed active set (a list of ATX IDs) will b used for
- the first safe layer range in epoch 2.
- when hare fails to reach consensus after N layers for any given epoch

## implementation plan
- [ ] https://github.com/spacemeshos/go-spacemesh/pull/4148 use atx grading scheme in reference ballot and hare
- [x] https://github.com/spacemeshos/go-spacemesh/pull/4231 binary that publishes bootstrap and fallback values for beacon and active set
- [x] https://github.com/spacemeshos/go-spacemesh/pull/4241 bootstrap and use first consensus block for hare active set

PR #4148, opened 2023-03-14
hare activeset: employ ATX grading scheme

countvonzero:
## Motivation
part of #4089

## Changes
- grade atxs as described in https://community.spacemesh.io/t/grading-atxs-for-the-active-set/335, let s be the start of the epoch, and δ the network propagation time.
  - grade 0/evil: ATX was received at time t >= s-3δ, or an equivocation proof was received by time s-δ.
  - grade 1/acceptable: ATX was received at time t < s-3δ before the start of the epoch, and no equivocation proof was received by time s-δ.
  - grade 2/good: ATX was received at time t < s-4δ, and no equivocation proof was received for that id until time s.
 
- miner only include grade 2/good ATXs in its active set for its first ballot of the epoch
- hare only vote for proposals with grade 1/acceptable and 2/good atxs in the activeset
- newly sync'ed node uses the first block of epoch for active set in ref ballot, since it doesn't have accurate received timestamp for atxs or malfeasance proof

### DIFF
--- a/hare/hare.go
+++ b/hare/hare.go
@@ -16,6 +16,7 @@ import (
 	"github.com/spacemeshos/go-spacemesh/hare/config"
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/malfeasance"
+	"github.com/spacemeshos/go-spacemesh/miner"
 	"github.com/spacemeshos/go-spacemesh/p2p/pubsub"
 	"github.com/spacemeshos/go-spacemesh/signing"
 	"github.com/spacemeshos/go-spacemesh/sql"
@@ -388,7 +389,7 @@ func (h *Hare) onTick(ctx context.Context, lid types.LayerID) (bool, error) {
 		report: h.outputChan,
 		wc:     h.wcChan,
 	}
-	props := goodProposals(ctx, h.Log, h.msh, h.nodeID, lid, beacon)
+	props := goodProposals(ctx, h.Log, h.msh, h.nodeID, lid, beacon, h.layerClock.LayerToTime(lid.GetEpoch().FirstLayer()), h.config.WakeupDelta)
 	preNumProposals.Add(float64(len(props)))
 	set := NewSet(props)
 	cp := h.factory(ctx, h.config, lid, set, h.rolacle, et, h.sign, h.publisher, comm, clock)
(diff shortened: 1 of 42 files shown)


countvonzero:
bors try

dshulyak:
so synced node votes against everyone? or it downloads timestamp from a peer?

countvonzero:
> so synced node votes against everyone? or it downloads timestamp from a peer?

timestamps are not synced. so yes. a newly synced node will vote against everyone and end up with an empty active set.

this means
1. a new node (synced from genesis) won't be able to make a block proposal in the same epoch that it join (empty active set)
2. a node that are mostly offline before `s-4δ` won't be able to make a block proposal in the following epoch (empty active set)
3. a node that publishes its atx within `s-4δ` window, won't be able to make a block proposal in the following epoch (the node itself is not in the active set)

in 1 and 2, it also won't vote for any proposals in the hare preround, but this should not affect hare consensus

in essence, this incentivize miners to stay online as much as possible.

dshulyak:
i understand that this is an attempt to solve some problems, but it almost guaranteed to introduce new problems. i don't see this complexity as a good protocol extension, maybe we can find a reference in working implementations that are doing something similar. otherwise i hope someone else can approve it

countvonzero:
> i understand that this is an attempt to solve some problems, but it almost guaranteed to introduce new problems. i don't see this complexity as a good protocol extension, maybe we can find a reference in working implementations that are doing something similar. otherwise i hope someone else can approve it

it's not about approving this PR.... it's more about flushing out concerns you may have.

> maybe we can find a reference in working implementations that are doing something similar

by working implementations do you mean other blockchain projects?

dshulyak:
yes, blockchain or other software (not necessarily byzantine) that works. i reviewed quite a few projects in the last years, and i don't recall conditions that assume so much.

dshulyak:
what do you think about next criteria for making such changes? there must be a test that clearly shows sequence of steps that must result in consensus failure. if the change is so obscure (or misunderstood) that it can't be tested - then there is no reason to make the change. 

i tried to understand Tal explanation, but it literally feels like a random change to me, that may fix something or may not fix anything

hare/hare.go
```diff
@@ -451,12 +460,32 @@ func goodProposals(logger log.Log, msh mesh, nodeID types.NodeID, lid types.Laye
 		ownTickHeight = ownHdr.TickHeight()
 	}
 	for _, p := range props {
+		hdr, err := msh.GetAtxHeader(p.AtxID)
```
countvonzero: should look at p's active set, not p's own atxid

countvonzero:
merged separately in
https://github.com/spacemeshos/go-spacemesh/pull/4717
https://github.com/spacemeshos/go-spacemesh/pull/4718

PR #4151, opened 2023-03-14
[Merged by Bors] - Fix logging of Post type

fasmat:
## Motivation
Log indices of Post as string instead of as value type. Indices are stored as compressed byte array. To decompress them we would need to know `NumUnits` and `K2` that were used to generate the proof. Instead we just log the first few bytes.

## TODO

## DevOps Notes


common/types/activation.go
```diff
@@ -616,22 +613,14 @@ func (p *Post) MarshalLogObject(encoder log.ObjectEncoder) error {
 		return nil
 	}
 	encoder.AddUint32("Nonce", p.Nonce)
-	encoder.AddString("Indicies", hex.EncodeToString(p.Indices))
+	encoder.AddString("Indices", hex.EncodeToString(p.Indices))
```
countvonzero: nit: lower case Nonce and Indices to be consistent?

countvonzero (approved):

fasmat:
bors merge

PR #4154, opened 2023-03-15
[Merged by Bors] - Set max size for variable size fields in types encoded with scale

### LINKED ISSUE #4095: revisit collections limits and apply sane max values

fasmat:
## Motivation
Closes https://github.com/spacemeshos/go-spacemesh/issues/4095

## Changes
Use newest go-scale that fails code generation if the "max" tag isn't set for variable size fields in types encoded with go-scale and fix code generation errors caused by missing tags.

- Some types had to be restructured (specifically named) to make tagging possible
- Some types were changed from slices to arrays, since their size is constant
- Here are all the types I changed, some of which I am unsure about. Feedback is appreciated:

Test
====

| Type | Field name | Max. Size | Description |
| ----- | ----------- | --------- | ------------ |
| `beacon.Proposal` | - | `[4]bytes` | Used to be just `[]byte`, since a proposal is always 4 bytes; converted to an array and added `DecodeScale` / `EncodeScale` methods. Additionally added a `String` method and a `ProposalFromString` function since these are often converted from and to strings (usually to be used as key in a `map`) |
| `beacon.ProposalMessage` | `VRFSignature` | 80 | VRF Signature of size 80 |
| `beacon.FirstVotingMessageBody` | `ValidProposals` | 1000 | The number of valid proposals should be lower than 100, max size of 1000 as a safety limit |
| `beacon.FirstVotingMessageBody` | `PotentiallyValidProposals` | 1000 | The number of valid proposals should be lower than 100, max size of 1000 as a safety limit |
| `beacon.FirstVotingMessage` | `Signature` | 64 | ED25519 Signature of size 64 |
| `beacon.FollowingVotingMessageBody` | `VotesBitVector` | 128 | We expect less than 1000 votes -> 1000/8 = 125 bytes; 128 bytes should be sufficient for encoding |
| `beacon.FollowingVotingMessage` | `Signature` | 64 | ED25519 Signature of size 64 |
| `weakcoin.Message` | `MinerPK` | 32 | ED25519 Public key of size 32 |
| | `VrfSignature` | 80 | VRF Signature of size 80 |
| `type.Account` | `State` | 10_000 | ??? (m: Template for State should be much smaller than 10k ?) |
| `type.NIPostChallenge` | `InitialPostIndices` | 8000 | Should be K2 * 8, some buffer for safety |
| `type.ActivationTx` | `Signature` | 64 | ED25519 Signature of size 64 |
| `type.PoetProof` | `Members` | 100_000 | Depends on the number of Smeshers we expect committing a challenge to a PoET |
| `type.PoetProofMessage` | `PoetServiceID` | 32 | ED25519 Public key for the PoetService |
| | `RoundID` | 32 | Identifier for a poet round. string representation of round count since genesis |
| | `Signature` | 64 | ED25519 Signature of size 64 |
| `type.PoetRound` | `ID` | 32 | Identifier for a poet round. string representation of round count since genesis |
| `type.Post` | `Indices` | 8000 | Should be K2 * 8, some buffer for safety |
| `type.PostMetadata` | `Challenge` | 32 | Poet challenge of size 32 byte |
| `type.ProcessingError` | `Err` | n/a | **TODO(mafa)** replace with Error codes |
| `type.Ballot` | `Signature` | 64 | ED25519 Signature of size 64 |
| | `EligibilityProofs` | 500 | number of expected proposals per layer is 50, 10X for safety margin |
| `type.InnerBallot` | `Support` | 10_000 | tortoise sliding window 10K layers |
| | `Against` | 10_000 | tortoise sliding window 10K layers |
| | `Abstain` | 10_000 | tortoise sliding window 10K layers |
| `type.EpochData` | `ActiveSet` | 100_000 | Depends on the number of Smeshers we expect publishing an ATX in an epoch |
| `type.InnerBlock` | `Rewards` | 500 | number of expected proposals per layer is 50, 10X for safety margin |
| | `TxIDs` | 100_000 | The upper limit (with buffer) on the number of transactions we expect in a block |
| `type.Certificate` | `Signatures` | 1000 | Depends on the Hare Parameter `N` + some buffer |
| `type.CertifyMessage` | `Signature` | 64 | ED25519 Signature of size 64 |
| `type.CertifyContent` | `Proof` | 80 | VRF Signature of size 80 |
| `type.HareEligibilityGossip` | `PubKey` | 32 | ED25519 Public key of size 32 |
| `type.HareEligibility` | `Proof` | 80 | VRF Signature of size 80 |
| `type.VotingEligibility` | `Sig` | 80 | VRF Signature of size 80 |
| `type.AtxProofMsg` | `Signature` | 64 | ED25519 Signature of size 64 |
| `type.BallotProofMsg` | `Signature` | 64 | ED25519 Signature of size 64 |
| `type.HareProofMsg` | `Signature` | 64 | ED25519 Signature of size 64 |
| `type.PoetServiceID` | `ServiceID` | 32 | ED25519 Public key for the PoetService |
| `type.NiPostBuilderState` | `PoetRequests` | 100 | 100 should be a safe limit for number of PoETs used by a single node |
| `type.PoetProofRef` | - | `[32]byte` | Used to be just `[]byte`, since always 32 bytes converted to an array |
| `type.Proposal` | `Signature` | 64 | ED25519 Signature of size 64 |
| `type.InnerProposal` | `TxIDs` | 100_000 | Depends on `NumTXsPerProposal`, which is defined by the app config `TxsPerProposal` with a lot of extra margin |
| `type.TransactionResult` | `Message` | 1024 | human readable error. **TODO(mafa)** convert from string to enum |
| | `Addresses` | 10 | Depending on type of transaction 1-3 with some safety buffer |
| `type.RawTx` | `Raw` | 4000 | A transaction is expected to be less than 4 KiB in size |
| `fetch.RequestMessage` | `Hint` | ? (currently 256) | **TODO(mafa)** convert from string to enum |
| `fetch.ResponseMessage` | `Data` | 3_200_000 | 100K of ATXIDs |
| `fetch.RequestBatch` | `Requests` | 1000 | configurable via BatchSize |
| `fetch.ResponseBatch` | `Responses` | 1000 | configurable via BatchSize |
| `fetch.MeshHashes` | `Layers` | 1000 | configurable via `MaxHashesInReq` |
| | `Hashes` | 1000 | configurable via `MaxHashesInReq` |
| `fetch.MaliciousIDs` | `NodeIDs` | 100_000 | Max. number of malicious nodes we expect in the network |
| `fetch.EpochData` | `AtxIDs` | 100_000) | max number of atx in a epoch |
| `fetch.LayerData` | `Ballots` | 500 | number of expected ballots per layer is 50, 10X for safety margin |
| | `Blocks` | 100 | expected 1 per layer |
| `multisig.MultiSig` | `PublicKeys` | 10 | hardcoded for genesis VM |
| `multisig.SpawnArguments` | `PublicKeys` | 10 | hardcoded for genesis VM |
| `hare.Message` | `Signature` | 64 | ED25519 Signature of size 64 |
| `hare.Certificate` | `Values` | 500 | number of expected proposals per layer is 50, 10X for safety margin |
| `hare.AggregatedMessages` | `Messages` | 1000 | hare param N plus safety margin |
| `hare.InnerMessage` | `Values` | 1000 | hare param N plus safety margin |
| `handshake.HandshakeAck` | `Error` | n/a | **TODO(mafa)** replace with Error codes |
| `server.Response` | `Data` | 10485760 (= 10 MiB) | Generic server response type |
| | `Error` | n/a | **TODO(mafa)** replace with Error codes |

## Test Plan
n/a

## TODO

## DevOps Notes


beacon/message.go
```diff
@@ -17,31 +17,35 @@ type ProposalVrfMessage struct {
...
-	PotentiallyValidProposals [][]byte
+	ValidProposals            []Proposal `scale:"max=32"` // TODO(mafa): check if max. value is correct
+	PotentiallyValidProposals []Proposal `scale:"max=32"` // TODO(mafa): check if max. value is correct
 }
 
 // FirstVotingMessage is a message type which is used when sending first voting messages.
 type FirstVotingMessage struct {
 	FirstVotingMessageBody
-	Signature []byte
+	Signature []byte `scale:"max=32"`
```
dshulyak: i think signature is 64 bytes

beacon/message.go
```diff
@@ -17,31 +17,35 @@ type ProposalVrfMessage struct {
...
+	Value []byte `scale:"max=32"` // TODO(mafa): check if max. value is correct
 }
 
 // FirstVotingMessageBody is FirstVotingMessage without a signature.
 type FirstVotingMessageBody struct {
 	EpochID                   types.EpochID
-	ValidProposals            [][]byte
-	PotentiallyValidProposals [][]byte
+	ValidProposals            []Proposal `scale:"max=32"` // TODO(mafa): check if max. value is correct
+	PotentiallyValidProposals []Proposal `scale:"max=32"` // TODO(mafa): check if max. value is correct
```
dshulyak: maybe it makes sense to change Proposal to byte array (e.g [32]byte ) ? it is more compact, easier to decode and safer

@countvonzero

beacon/weakcoin/weak_coin.go
```diff
@@ -41,8 +41,8 @@ type Message struct {
 	Epoch        types.EpochID
 	Round        types.RoundID
 	Unit         uint32
-	MinerPK      []byte
-	VrfSignature []byte
+	MinerPK      []byte `scale:"max=32"`
+	VrfSignature []byte `scale:"max=32"`
```
dshulyak: vrf signature is 80 bytes

activation/poet.go
```diff
@@ -118,41 +118,64 @@ func (c *HTTPPoetClient) Submit(ctx context.Context, challenge []byte, signature
...
 func (c *HTTPPoetClient) Proof(ctx context.Context, roundID string) (*types.PoetProofMessage, error) {
-	resBody := rpcapi.GetProofResponse{}
+	resBody := rpcapi.ProofResponse{}
 	if err := c.req(ctx, http.MethodGet, fmt.Sprintf("/v1/proofs/%s", roundID), nil, &resBody); err != nil {
 		return nil, fmt.Errorf("getting proof: %w", err)
 	}
 
+	p := resBody.Proof.GetProof()
+	root := p.GetRoot()
+	leaves := make([]shared.Leaf, 0, len(p.ProvenLeaves))
```
poszu: The original code used getters like `p.GetProvenLeaves()`. The reason is that in protobuf every field is _optional_ and the getters have nil-checks. It is possible that both `err` and `p` are `nil` here - the data from the poet cannot be trusted. Direct access can lead to a segfault.

beacon/message.go
fasmat: In some places (like signatures) this might work, but the size of `ValidProposals` and `PotentiallyValidProposals` is afaik a configuration parameter.

Arrays cannot be dynamic in size, that being said the `scale:"max=xxx"` tag also doesn't consider configuration...

activation/poet.go
fasmat: Fixed

beacon/message.go
countvonzero: i think @dshulyak  is talking about the element `Proposal` itself. it is always 4 byte (defined as types.BeaconSize). so you can safely define it as a byte array

dshulyak: yes i was talking about that

we need to set some upper bound on the total number of proposals as well, maybe 10000 is ok ?

beacon/message.go
```diff
@@ -17,31 +17,35 @@ type ProposalVrfMessage struct {
...
-	Signature []byte
+	Signature []byte `scale:"max=64"`
 }
 
 // FollowingVotingMessageBody is FollowingVotingMessage without a signature.
 type FollowingVotingMessageBody struct {
 	EpochID        types.EpochID
 	RoundID        types.RoundID
-	VotesBitVector []byte
+	VotesBitVector []byte `scale:"max=32"` // TODO(mafa): check if max. value is correct
```
countvonzero: the protocol guarantees that there are very low probability that there will be over 100 proposals.
so a bit vector of 100 bits is 13 bytes. i think 16 byte may suffice.

beacon/message.go
```diff
@@ -17,31 +17,35 @@ type ProposalVrfMessage struct {
 type ProposalMessage struct {
 	EpochID      types.EpochID
 	NodeID       types.NodeID
-	VRFSignature []byte
+	VRFSignature []byte `scale:"max=80"`
+}
+
+type Proposal struct {
+	Value []byte `scale:"max=32"` // TODO(mafa): check if max. value is correct
```
countvonzero: 4 byte exactly. (types.BeaconSize)

common/types/account.go
```diff
@@ -14,7 +14,7 @@ type Account struct {
 	NextNonce       uint64
 	Balance         uint64
 	TemplateAddress *Address
-	State           []byte
+	State           []byte `scale:"max=1024"` // TODO(mafa): check if max. value is correct
```
countvonzero: how do you come up with 1024 byte here?

common/types/ballot.go
```diff
@@ -135,11 +135,11 @@ type Votes struct {
...
 	Base BallotID
 	// Support block id at a particular layer and height.
-	Support []Vote
+	Support []Vote `scale:"max=800"` // TODO(mafa): check if this is the right max size
 	// Against previously supported block.
-	Against []Vote
+	Against []Vote `scale:"max=800"` // TODO(mafa): check if this is the right max size
 	// Abstain on layers until they are terminated.
-	Abstain []LayerID
+	Abstain []LayerID `scale:"max=800"` // TODO(mafa): check if this is the right max size
```
countvonzero: can @dshulyak @fasmat tell me the rough math of how these max values are derived?

countvonzero: i mean 800 sounds excessive for VotingEligibility (considering we are aiming total 50 proposals per layer)
for Support/Against/Abstain this means we will not find a reasonable base ballot for 800 layers?

common/types/ballot.go
```diff
@@ -197,7 +197,7 @@ func (o *Opinion) MarshalLogObject(encoder log.ObjectEncoder) error {
 // EpochData contains information that cannot be changed mid-epoch.
 type EpochData struct {
 	// from the smesher's view, the set of ATXs eligible to vote and propose block content in this epoch
-	ActiveSet []ATXID
+	ActiveSet []ATXID `scale:"max=800"` // TODO(mafa): check if this is the right max size
```
countvonzero: this mean we will have max 800 ATXs in the active set. this is def. too small

countvonzero (commented):
we should work out the exact max number for each of the protocol msgs instead of leaving them as TODOs at this point.

beacon/message.go
countvonzero: here the total number of proposals, with high probabilities, would be lower than 100. so i think 200 is enough.

common/types/ballot.go
fasmat: whenever I didn't know what to use as max value I just put a value there that I thought was reasonable and added a `TODO`. I can lower the values to `50` each if that is a better maximum value?

common/types/ballot.go
fasmat: What would be a better max. size in your opinion?

beacon/message.go
fasmat: Made Proposal a `[4]byte` instead of `[]byte` and added a max of 200 to `ValidProposals` and `PotentiallyValidProposals`

common/types/account.go
fasmat: I just made up a number, what is a better size?

countvonzero:
@fasmat 
what do you think of making a table that list all the max sizes decision that are made in this PR?
then we can go through them one-by-one on the dev sync meeting.
this would be great protocol documentation as well.

fasmat:
@countvonzero @dshulyak @poszu 

I added all fields of all types that I added size specifications for to the PR description. I don't think this is a good base for a spec, since all the context for those types are missing, but maybe it helps in finishing this PR by agreeing on the sizes of those fields.

There is also at least one type `type.PoetRequest` that contains itself. So theoretically even with max. sizes it can be infinitely big. Does it have to be recursive?

fasmat:
bors try

countvonzero:
> @countvonzero @dshulyak @poszu
> 
> I added all fields of all types that I added size specifications for to the PR description. I don't think this is a good base for a spec, since all the context for those types are missing, but maybe it helps in finishing this PR by agreeing on the sizes of those fields.

the PR itself is a documentation. (not spec). i think it's better to put a table in the PR description that lists all the max size enforced. that way, people just need to read the PR description instead of reading through the changes.

my preference would be using that table to do a team discussion. it's a good opportunity to share the parameters amongst the team members so we are on the same page.

countvonzero:
> the PR itself is a documentation. (not spec). i think it's better to put a table in the PR description that lists all the max size enforced. that way, people just need to read the PR description instead of reading through the changes.

just saw that you already did it! ❤️

countvonzero:
i updated the table in the PR description and added my thoughts. 
some of the max values should correspond to a config, but should not be exactly the configured value, since we may want to change the configured value without doing a software update

fasmat:
I updated some of the sizes and the PR description based on feedback from @countvonzero ❤️

fasmat:
bors try

beacon/message.go
```diff
@@ -17,31 +19,53 @@ type ProposalVrfMessage struct {
...
 	NodeID       types.NodeID
-	VRFSignature []byte
+	VRFSignature []byte `scale:"max=80"`
+}
+
+type Proposal [4]byte
+
+func ProposalFromString(s string) Proposal {
+	var p Proposal
+	copy(p[:], []byte(s))
```
countvonzero: for copy i think you can do just copy(p[:], s) without casting as

beacon/message.go
```diff
@@ -17,31 +19,53 @@ type ProposalVrfMessage struct {
...
+	return p
+}
+
+func (p Proposal) String() string {
```
countvonzero: i don't think i like this name. this can be easily used in the log.Stringer() interface and the output would be garbage binary.

beacon/message.go
```diff
@@ -17,31 +19,53 @@ type ProposalVrfMessage struct {
...
+func (p *Proposal) DecodeScale(d *scale.Decoder) (int, error) {
+	return scale.DecodeByteArray(d, p[:])
 }
 
 // FirstVotingMessageBody is FirstVotingMessage without a signature.
 type FirstVotingMessageBody struct {
 	EpochID                   types.EpochID
-	ValidProposals            [][]byte
-	PotentiallyValidProposals [][]byte
+	ValidProposals            []Proposal `scale:"max=1000"`
```
countvonzero: add comment:

// number of proposals is under 100 with a high probability. 10X for safety margin.

common/fixture/transaction_results.go
```diff
@@ -71,7 +71,8 @@ func (g *TransactionResultGenerator) Next() *types.TransactionWithResult {
 	tx.Block = g.Blocks[g.rng.Intn(len(g.Blocks))]
 	tx.Layer = g.Layers[g.rng.Intn(len(g.Layers))]
 	if lth := g.rng.Intn(len(g.Addrs)); lth > 0 {
-		tx.Addresses = make([]types.Address, lth)
+		// tx.Addresses will only be 1 or 2 addresses
```
countvonzero: rm

common/fixture/transaction_results.go
```diff
@@ -71,7 +71,8 @@ func (g *TransactionResultGenerator) Next() *types.TransactionWithResult {
...
+		tx.Addresses = make([]types.Address, lth%2+1)
```
countvonzero: use 10 instead

common/types/activation.go
```diff
@@ -466,9 +466,9 @@ func (p *PoetProof) MarshalLogObject(encoder log.ObjectEncoder) error {
 // PoetProofMessage is the envelope which includes the PoetProof, service ID, round ID and signature.
 type PoetProofMessage struct {
 	PoetProof
-	PoetServiceID []byte
-	RoundID       string
-	Signature     []byte
+	PoetServiceID []byte `scale:"max=32"` // public key of the PoET service
+	RoundID       string `scale:"max=32"`
```
countvonzero: add TODO to convert to int

common/types/ballot.go
```diff
@@ -35,18 +35,18 @@ func (id *BallotID) DecodeScale(d *scale.Decoder) (int, error) {
...
 	BallotMetadata
 	InnerBallot
-	// smesher's signature on InnerBallot
-	Signature []byte
+	// smeshers signature on InnerBallot
+	Signature []byte `scale:"max=64"`
 	// Votes field is not signed.
-	Votes Votes
-	// the proof of the smesher's eligibility to vote and propose block content in this epoch.
+	Votes Votes `scale:"max=10000"` // TODO(mafa): check again why this is the right max size
```
countvonzero: you don't need limit here. `Votes` is just a plain struct

common/types/ballot.go
```diff
@@ -35,18 +35,18 @@ func (id *BallotID) DecodeScale(d *scale.Decoder) (int, error) {
...
+	// the proof of the smeshers eligibility to vote and propose block content in this epoch.
 	// Eligibilities must be produced in the ascending order.
-	EligibilityProofs []VotingEligibility
+	EligibilityProofs []VotingEligibility `scale:"max=500"` // TODO(mafa): check again why this is the right max size
```
countvonzero: the protocol param is ~50 proposals per layer. 10X for safety margin.

common/types/ballot.go
```diff
@@ -135,11 +135,11 @@ type Votes struct {
 	// Base ballot.
 	Base BallotID
 	// Support block id at a particular layer and height.
-	Support []Vote
+	Support []Vote `scale:"max=10000"` // TODO(mafa): check again why this is the right max size
```
countvonzero: for Support and Against, the max number of votes would be the tortoise sliding window, which is 10K layers, assuming 1 block per layer.

for Abstain, similar except the vote is one per layer (as opposed to one per block for Support/Against)

common/types/transaction_result.go
```diff
@@ -44,14 +44,13 @@ func (t TransactionStatus) DecodeScale(d *scale.Decoder) (uint8, int, error) {
 // TransactionResult is created after consuming transaction.
 type TransactionResult struct {
 	Status  TransactionStatus
-	Message string
+	Message string `scale:"max=1024"` // TODO(mafa): check again why this is the right max size
```
countvonzero: this is just error message for UI/explorer to explain why the transaction failed (if it failed). so it's arbitrary human readable text. 1K is reasonable.

fetch/wire_types.go
```diff
@@ -11,27 +11,27 @@ import (
...
-	Hint datastore.Hint
+	Hint datastore.Hint `scale:"max=256"` // TODO(mafa): covert to an enum
 	Hash types.Hash32
 }
 
 // ResponseMessage is sent to the node as a response.
 type ResponseMessage struct {
 	Hash types.Hash32
-	Data []byte
+	Data []byte `scale:"max=1000000"` // TODO(mafa): check if this is the right max size
```
countvonzero: i think the max number would be 100K ATXID. s 4MB

fetch/wire_types.go
```diff
@@ -11,27 +11,27 @@ import (
...
 }
 
 // RequestBatch is a batch of requests and a hash of all requests as ID.
 type RequestBatch struct {
 	ID       types.Hash32
-	Requests []RequestMessage
+	Requests []RequestMessage `scale:"max=20"` // depends on fetch config `BatchSize` which defaults to 20
```
countvonzero: make it 1000 instead. we want to be able to configure this number without release a binary.

fetch/wire_types.go
```diff
@@ -48,22 +48,22 @@ func (r *MeshHashRequest) MarshalLogObject(encoder log.ObjectEncoder) error {
 }
 
 type MeshHashes struct {
-	Layers []types.LayerID
-	Hashes []types.Hash32
+	Layers []types.LayerID `scale:"max=100"` // same as syncer Config `MaxHashesInReq`
+	Hashes []types.Hash32  `scale:"max=100"` // same as syncer Config `MaxHashesInReq`
```
countvonzero: 1K for both too

fetch/wire_types.go
```diff
@@ -11,27 +11,27 @@ import (
...
-	Requests []RequestMessage
+	Requests []RequestMessage `scale:"max=20"` // depends on fetch config `BatchSize` which defaults to 20
 }
 
 // ResponseBatch is the response struct send for a RequestBatch. the ResponseBatch ID must be the same
 // as stated in RequestBatch even if not all Data is present.
 type ResponseBatch struct {
 	ID        types.Hash32
-	Responses []ResponseMessage
+	Responses []ResponseMessage `scale:"max=20"` // depends on fetch config `BatchSize` which defaults to 20
```
countvonzero: 1000, same as RequestBatch.Requests

fetch/wire_types.go
```diff
@@ -48,22 +48,22 @@ func (r *MeshHashRequest) MarshalLogObject(encoder log.ObjectEncoder) error {
...
 type MeshHashes struct {
-	Layers []types.LayerID
-	Hashes []types.Hash32
+	Layers []types.LayerID `scale:"max=100"` // same as syncer Config `MaxHashesInReq`
+	Hashes []types.Hash32  `scale:"max=100"` // same as syncer Config `MaxHashesInReq`
 }
 
 type MaliciousIDs struct {
-	NodeIDs []types.NodeID
+	NodeIDs []types.NodeID `scale:"max=100000"` // maximum expected number of nodes in the network
```
countvonzero: :s/nodes/ATXs

fetch/wire_types.go
```diff
@@ -48,22 +48,22 @@ func (r *MeshHashRequest) MarshalLogObject(encoder log.ObjectEncoder) error {
...
-	AtxIDs []types.ATXID
+	AtxIDs []types.ATXID `scale:"max=1000000"`
 }
 
 // LayerData is the data response for a given layer ID.
 type LayerData struct {
-	Ballots []types.BallotID
-	Blocks  []types.BlockID
+	Ballots []types.BallotID `scale:"max=800"`
+	Blocks  []types.BlockID  `scale:"max=800"`
```
countvonzero: this is different from ballots. each layer we expect to have 1 block. so maybe 100 would be a good safety margin.

hare/builder.go
```diff
@@ -68,32 +68,32 @@ func (m *Message) SignedBytes() []byte {
 // Certificate is a collection of messages and the set of values.
 // Typically used as a collection of commit messages.
 type Certificate struct {
-	Values  []types.ProposalID // the committed set S
+	Values  []types.ProposalID `scale:"max=800"` // the committed set S // TODO(mafa): check if this is the right max size
```
countvonzero: 500. this is the same as number of expected proposals per layer with 10X margin

hare/builder.go
```diff
@@ -68,32 +68,32 @@ func (m *Message) SignedBytes() []byte {
...
 	AggMsgs *AggregatedMessages
 }
 
 // AggregatedMessages is a collection of messages.
 type AggregatedMessages struct {
-	Messages []Message
+	Messages []Message `scale:"max=800"` // TODO(mafa): check if this is the right max size
```
countvonzero: should be consistent with N with safety margin -> 1000

hare/builder.go
```diff
@@ -68,32 +68,32 @@ func (m *Message) SignedBytes() []byte {
...
 }
 
 // InnerMessage is the actual set of fields that describe a message in the Hare protocol.
 type InnerMessage struct {
 	Type           MessageType
 	CommittedRound uint32              // the round Values (S) is committed (Ki)
-	Values         []types.ProposalID  // the set S. optional for commit InnerMsg in a certificate
+	Values         []types.ProposalID  `scale:"max=800"` // the set S. optional for commit InnerMsg in a certificate // TODO(mafa): check if this is the right max size
```
countvonzero: same as Certificate.Values

hare/flows_test.go
```diff
@@ -289,21 +290,29 @@ func Test_multipleCPs(t *testing.T) {
...
+				case <-ctx.Done():
+					return
+				case out, ok := <-h.blockGenCh:
+					if !ok {
+						return
+					}
+					if outputs[idx] == nil {
+						outputs[idx] = make(map[types.LayerID]LayerOutput)
+					}
+					outputs[idx][out.Layer] = out
```
countvonzero, line 310: if you change this, you should also change Test_multipleCPsAndIterations too.

countvonzero (commented):
for PR description:
- proposalList remove from table
- ballot.Votes remove from table

question:
for pubkey, ed signature and vrf signature, why didn't you go with fixed-size array [32]byte, [64]byte and [80]byte?

beacon/message.go
fasmat: I completely got rid of the conversion of `Proposal` from and to `string` so this method isn't necessary any more.

common/types/transaction_result.go
fasmat: added TODO that it should be converted to an error code instead of a message

fasmat:
bors try

fetch/wire_types.go
```diff
@@ -48,22 +48,22 @@ func (r *MeshHashRequest) MarshalLogObject(encoder log.ObjectEncoder) error {
...
 type EpochData struct {
-	AtxIDs []types.ATXID
+	AtxIDs []types.ATXID `scale:"max=1000000"`
 }
 
 // LayerData is the data response for a given layer ID.
 type LayerData struct {
-	Ballots []types.BallotID
-	Blocks  []types.BlockID
+	Ballots []types.BallotID `scale:"max=800"`
```
countvonzero: i'll set this to 500 just to be consistent with number of proposals expected in a layer

fetch/wire_types.go
```diff
@@ -48,22 +48,22 @@ func (r *MeshHashRequest) MarshalLogObject(encoder log.ObjectEncoder) error {
...
 }
 
 type MaliciousIDs struct {
-	NodeIDs []types.NodeID
+	NodeIDs []types.NodeID `scale:"max=100000"` // max. expected number of ATXs per epoch is 100_000
 }
 
 type EpochData struct {
-	AtxIDs []types.ATXID
+	AtxIDs []types.ATXID `scale:"max=1000000"`
```
countvonzero: i think we agreed on 100K, the max atx ids in an epoch.

countvonzero (approved):
i updated the table to fill out some missing comments.

fetch/wire_types.go
fasmat: looks like a typo - fixed

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

PR #4161, opened 2023-03-16
[Merged by Bors] - node: crash if specified config can't be loaded

### LINKED ISSUE #4160: Node should crash when the specified config is invalid
When the `go-spacemesh` receives wrong config it logs
``` failed loading config from /home/XXX/services/node_config.json trying ./config.toml. error While parsing config: invalid character ']' looking for beginning of value```

And then it basically loads defaults. 

I think this behavior is really dangerous. The wrong config should result with a failed start.

### PR DESCRIPTION
**Pull request by pigmej**
Closes #4160

changes:

* if config file is specified we don't fallback to anything and instead load that config file
* otherwise load config based on the specified preset or regular default config

### CODE REVIEW

File: config/config.go [hunk 1]
```diff
@@ -195,6 +196,10 @@ func LoadConfig(fileLocation string, vip *viper.Viper) (err error) {
 	err = vip.ReadInConfig()
```
countvonzero: just do `return vip.ReadInConfig()` here. the upstream caller does log.fatal() on error already.

in essence, we simply don't fall back to do default config.

pigmej: @countvonzero yeah, but the problem is that this logic is needed in tests, and we pretty much somehow assume kinda everywhere that this is tried to be ALWAYS read.

That logic is fatally flawed and should be reworked completely.

Nevertheless, that fix is meant to fix the problem faced in the real production application now.

dshulyak: i refactored to return fatal error on any failure to load specified config file.
if config file is not specified we will try to load default config file if it exists

dshulyak (approved):

dshulyak:
@countvonzero @pigmej please check if what i wrote in description sounds good, i simplified config handling a bit

dshulyak:
bors try

pigmej:
bors try

config/config.go
```diff
@@ -186,26 +185,20 @@ func defaultTestConfig() BaseConfig {
...
-			log.Warning("failed loading config from %v trying %v. error %v", fileLocation, defaultConfigFileName, err)
-			vip.SetConfigFile(defaultConfigFileName)
-			err = vip.ReadInConfig()
-		}
-		// we change err so check again
-		if err != nil {
-			return fmt.Errorf("failed to read config file %v", err)
+func LoadConfig(config string, vip *viper.Viper) error {
+	if config == defaultConfigFileName || config == "" {
+		if _, err := os.Stat(defaultConfigFileName); err != nil {
```
countvonzero: i honestly don't understand the purpose of `defaultConfigFileName`, which is set to `config.toml`. this file was deleted a while back i think.

the code would be simpler without this variable, and move on (return nil) `config == ""`

countvonzero (approved):

config/config.go
dshulyak: i thought that it is kind of a ux thing, but you are right there is no need for it

dshulyak:
bors merge

dshulyak:
bors merge

fasmat (approved):

Issue #4070, opened 2023-02-17
Observability to understand network propagation times


piersy:
In order for the spacemesh protocol to operate effectively we need to ensure that for rounds of hare we provide a sufficiently long timeout such that the vast majority of the network receive all the messages they need to be able to participate.

Some possible approaches:
* Collect metrics from all nodes (or all opt in nodes) to some central server and process them there, in this way we can track a message's progress through the network.
* Set up some geographically distributed nodes that we own disable gossiping for these nodes and use them to send "tracer messages" through the network. That way we can understand how long it takes to go across the network between our nodes and our nodes are not skewing the reading because they do not participate in gossiping.
* Somehow manage this locally, if messages are timestamped then receiving nodes can tell how long a message took to reach them. This doesn't account for clock drift, but maybe there is a way to handle that (nodes could exchange clock time during handshake).
* Do we even need to actually need a concrete pre-defined time, or could we just adjust round timing depending on the participation level?

piersy:
So it looks like a promising approach here is to record message arrival time relative to the time a message should be sent in the protocol. For example we know the instant at which a layer is started, and certain messages are sent at layer start, so by comparing their arrival time to the start time of the layer we can understand how long a message took to arrive. This doesn't account for clock drift, but nodes run a protocol to keep their clocks in sync, so that shouldn't pose a problem.Recording this information in a prometheus histogram provides an efficient way to record this information.

PR #4162, opened 2023-03-16
[Merged by Bors] - Track hare message delay with prometheus histogram

piersy:
## Motivation

#4070

## Changes
Added 2 prometheus histograms per hare message type, one to track how late a message is and one to track how early a a message is relative to the time a round starts.

## Test Plan
We need to look at this in grafana with a more realistic network than the one generated by systest to see if the metrics will be useful.

I plan to follow up with an alternative approach that analyses the logs of message sending and receipt from multiple machines to see if the results correlate with what we get from the histograms.

The accuracy of the histogram approach will be correlated with the level of synchronisation of nodes' clocks, so if nodes tend to be quite out of sync then the histogram approach will not provide reliable data.

## DevOps Notes

### DIFF
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -503,7 +503,7 @@ func (proc *consensusProcess) processMsg(ctx context.Context, m *Msg) {
 		log.String("msg_type", m.InnerMsg.Type.String()),
 		log.Int("num_values", len(m.InnerMsg.Values)))
 
-	msgDelayTracker.trackDelay(m.InnerMsg.Type, proc.clock)
+	reportLatency(m.InnerMsg.Type, m.Round, proc.clock)
 	switch m.InnerMsg.Type {
 	case pre:
 		proc.processPreRoundMsg(ctx, m)
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -213,6 +213,9 @@ type consensusProcess struct {
 	eligibilityCount uint16
 	clock            RoundClock
 	once             sync.Once
+	// layerTime is used to calculate the difference between expected message
+	// arrival times and actual message arrival times
+	layerTime time.Time
 }
 
 // newConsensusProcess creates a new consensus process instance.
@@ -232,6 +235,7 @@ func newConsensusProcess(
 	ev roleValidator,
 	clock RoundClock,
 	logger log.Log,
+	layerTime time.Time,
 ) *consensusProcess {
 	proc := &consensusProcess{
 		State: State{
@@ -252,6 +256,7 @@ func newConsensusProcess(
 		mTracker:  newMsgsTracker(),
 		eTracker:  NewEligibilityTracker(cfg.N),
 		clock:     clock,
+		layerTime: layerTime,
 	}
 	proc.ctx, proc.cancel = context.WithCancel(ctx)
 	proc.preRoundTracker = newPreRoundTracker(logger.WithContext(proc.ctx).WithFields(proc.layer), comm.mchOut, proc.eTracker, cfg.N/2+1, cfg.N)
@@ -498,6 +503,7 @@ func (proc *consensusProcess) processMsg(ctx context.Context, m *Msg) {
 		log.String("msg_type", m.InnerMsg.Type.String()),
 		log.Int("num_values", len(m.InnerMsg.Values)))
 
+	msgDelayTracker.trackDelay(m.InnerMsg.Type, proc.clock)
 	switch m.InnerMsg.Type {
 	case pre:
 		proc.processPreRoundMsg(ctx, m)
--- a/hare/algorithm.go
+++ b/hare/algorithm.go
@@ -213,9 +213,6 @@ type consensusProcess struct {
 	eligibilityCount uint16
 	clock            RoundClock
 	once             sync.Once
-	// layerTime is used to calculate the difference between expected message
-	// arrival times and actual message arrival times
-	layerTime time.Time
 }
 
 // newConsensusProcess creates a new consensus process instance.
@@ -235,7 +232,6 @@ func newConsensusProcess(
 	ev roleValidator,
 	clock RoundClock,
 	logger log.Log,
-	layerTime time.Time,
 ) *consensusProcess {
 	proc := &consensusProcess{
 		State: State{
@@ -256,7 +252,6 @@ func newConsensusProcess(
 		mTracker:  newMsgsTracker(),
 		eTracker:  NewEligibilityTracker(cfg.N),
 		clock:     clock,
-		layerTime: layerTime,
 	}
 	proc.ctx, proc.cancel = context.WithCancel(ctx)
 	proc.preRoundTracker = newPreRoundTracker(logger.WithContext(proc.ctx).WithFields(proc.layer), comm.mchOut, proc.eTracker, cfg.N/2+1, cfg.N)
--- a/hare/metrics.go
+++ b/hare/metrics.go
@@ -1,6 +1,10 @@
 package hare
 
 import (
+	"fmt"
+	"strings"
+	"time"
+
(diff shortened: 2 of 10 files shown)


hare/metrics.go
```diff
@@ -45,4 +49,55 @@ var (
...
+	metrics map[MessageType]map[bool]prometheus.Observer
+}
+
+func (m messageDelayTracker) trackDelay(t MessageType, clock RoundClock) {
+	subMap, ok := m.metrics[t]
+	if !ok {
+		// Untracked message type
+		return
+	}
+	seconds := time.Since(clock.RoundEnd(t.round() - 1)).Seconds()
```
countvonzero: there is a difference btwn message type value and round number.

for example. status round (val 0) can have round number of 0/4/8/12. the param to `clock.RoundEnd()` expect the actual round number, which is recorded in `msg.Round()`

countvonzero (commented):
will leave it to @dshulyak to review/approve.

dshulyak:
> The accuracy of the histogram approach will be correlated with the level of synchronisation of nodes' clocks, so if nodes tend to be quite out of sync then the histogram approach will not provide reliable data.

this is expected, but node clocks must remain in sync in order for protocol to work. there is a sub-protocol that checks that peers clock is in sync. i assume that most of the good nodes will be running ntp, small deviation is irrelevant and big outliers can be filtered out using statistical tools

hare/metrics.go
```diff
@@ -45,4 +49,55 @@ var (
 		[]string{},
 		prometheus.ExponentialBuckets(4, 2, 3),
 	).WithLabelValues()
+
+	msgDelayTracker = newMessageDelayTracker()
 )
+
+func newMessageDelayTracker() *messageDelayTracker {
```
dshulyak: i think it will be easier to navigate one histogram with two labels (sign is optional for me though, we can understand if time is misconfigurated by looking at outliers). it is arguably also cleaner code, something like below

```go
var hareLatency = metrics.NewHistogramWithBuckets(
	"hare_latency_ms",
	subsystem,
	"Observed latency for hare message",
	[]string{"type", "sign"},
	prometheus.ExponentialBuckets(50, 2, 20),
)

func reportLatency(msgtype string, value time.Duration) {
	sign := "pos"
	if value < 0 {
		sign = "neg"
	}
	hareLatency.WithLabelValues(msgtype, sign).Observe(float64(value.Milliseconds()))
}
```

hare/metrics.go
dshulyak: i think the current implementation works correctly only for the first iteration (which is a bug)

hare/metrics.go
```diff
@@ -45,4 +49,55 @@ var (
...
+			if !pos {
+				sign = "neg"
+			}
+
+			m[t][pos] = metrics.NewHistogramWithBuckets(
+				fmt.Sprintf("%v_%v_delay_seconds", strings.ToLower(t.String()), sign),
+				namespace,
+				fmt.Sprintf("delta between expected time of receipt of %v message and actual receipt time", t),
+				[]string{},
+				prometheus.ExponentialBuckets(0.2, 2, 10),
```
dshulyak: i opted for ms for similar thing, do you think that working with floats is better?

hare/algorithm.go
```diff
@@ -213,6 +213,9 @@ type consensusProcess struct {
 	eligibilityCount uint16
 	clock            RoundClock
 	once             sync.Once
+	// layerTime is used to calculate the difference between expected message
+	// arrival times and actual message arrival times
+	layerTime time.Time
```
dshulyak: expected hare start time can be always computed. 

hare_layer_time = genesis + layer*layer_duration + hare_wake_up

round_time = hare_layer_time + (round*iteration)*round_duration

layer and (round*iteration) is already encoded in the message, so time can be always computed without additional context

hare/metrics.go
piersy: Label sounds good. The reason I decided to track negative values was that from my understanding of [prometheus histograms](https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations) negative values are not supported, and if we were to simply discard them then we'd miss out on those observations.

hare/metrics.go
piersy: I was thinking of using ms too but then I read in the [prometheus label naming practices](https://prometheus.io/docs/practices/naming/) that you should use base units, which seemed a bit arbitrary to me but I switched :) I don't have any preference for float or non float, but in the end `Observe` expects a float, so I don't see any problem with using floats. Is there a problem?

piersy: Happy to switch back to ms if that's going to be more inline with our current metrics

hare/algorithm.go
piersy: Great!

hare/metrics.go
piersy: Are you referring to the round clock or my use of it here?

piersy: Ok, thanks! Figuring out a fix.

hare/metrics.go
dshulyak: i guess seconds might be somewhat easier to consume, with milliseconds it looks like this

image

hare/metrics.go
piersy: Actually, now understanding labels a bit better, it seems like making the message type a label would also be beneficial?

hare/algorithm.go
piersy: Actually, I wasn't even using it, since the round clock included everything that was required.

piersy:
@dshulyak Updated, I think I've addressed all points

hare/metrics.go
```diff
@@ -45,4 +47,32 @@ var (
...
+// arrive after the round has started. In a properly functioning system all
+// messages should be late. Early messages indicate a lack of time
+// synchronization between nodes.
+func reportLatency(t MessageType, round uint32, clock RoundClock) {
+	seconds := time.Since(clock.RoundEnd(round - 1)).Seconds()
+	sign := "pos"
+	if seconds < 0 {
+		sign = "neg"
+		// If the observation is negative make it positive.
+		seconds -= seconds
```
dshulyak: i think this will be always zero, as it translates to `seconds = seconds - seconds`

should probably be `seconds = -seconds`

hare/metrics.go
```diff
@@ -45,4 +47,32 @@ var (
...
+	}
+	hareLatency.WithLabelValues(t.String(), sign).Observe(seconds)
```
dshulyak: t.String() is in lowercase right?

dshulyak (approved):
please fix comment about neg

hare/metrics.go
piersy: Haha, yes, well spotted!

hare/metrics.go
piersy: That should probably be lowercased. Thanks

piersy:
bors merge

hare/metrics.go
```diff
@@ -45,4 +48,32 @@ var (
...
+// synchronization between nodes.
+func reportLatency(t MessageType, round uint32, clock RoundClock) {
+	seconds := time.Since(clock.RoundEnd(round - 1)).Seconds()
+	sign := "pos"
+	if seconds < 0 {
+		sign = "neg"
+		// If the observation is negative make it positive.
+		seconds = -seconds
+	}
+	hareLatency.WithLabelValues(strings.ToLower(t.String()), sign).Observe(seconds)
```
countvonzero, line 78: i think it would've been fine if you change the impl of t.String() to use lower case.

piersy: I didn't change that since it would have affected a number of logs, but if its helpful then I can always change it in another PR.

PR #4163, opened 2023-03-17
[Merged by Bors] - atx: remove commitment check

dshulyak:
this condition can be relaxed without any security concerns. for genesis we can assume that honest nodes are incentivized to pick higher commitment as it will prevent non-honest nodes from influencing beacon. long term - protocol will be extended with material incentive for selecting higher commitment atx.


fasmat (approved):
There should also be some tests that need to updated with this check removed.

dshulyak:
i run activation tests and they passed, lets see if ci fails)

dshulyak:
bors merge

countvonzero:
btw, prioritizing more recent commitment atx in beacon is not done.

PR #4169, opened 2023-03-20
[Merged by Bors] - hare active set: create bootstrap data updater

countvonzero:
## Motivation
part of #4089

## Changes
bootstrap data updater that 
- load the latest bootstrap data from disk at instantiation time
- after app started, periodically query an URL for json update
  - validate data against json schema
  - validate data with app logic that cannot be enforced by schema
  - persist the latest update on disk and prune old ones
  - notify subscribers of a new update

note:
- test with a memory filesystem from github.com/spf13/afero


bootstrap/updater.go
```diff
@@ -0,0 +1,406 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("bootstrap read resonse: %w", err)
+	}
+	return data, nil
+}
+
+type Config struct {
+	URI             string `mapstructure:"bootstrap-uri"`
+	Version         string `mapstructure:"bootstrap-version"`
+	SpacemeshPubKey string `mapstructure:"spacemesh-pubkey"`
```
dshulyak: i think the agreement is that we won't be doing it like this, at least there is no reason to do it. signature comes from PKI and HTTPS, so it will be verified when TLS connection is establishedd

countvonzero: you are right. it was a previous version where i wanted to merge recovery use case with bootstrapping. 
removed.

bootstrap/mocks.go
```diff
@@ -0,0 +1,86 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: ./interface.go
+
+// Package bootstrap is a generated GoMock package.
```
piersy: Probably we don't want this package comment, you can avoid writing it by using `-write_package_comment false` when generating mocks

bootstrap/updater.go
```diff
@@ -0,0 +1,374 @@
...
+	format     = "2006-01-02T15-04-05"
+)
+
+var (
+	ErrEpochOutOfOrder = errors.New("epoch out of order")
+	ErrWrongVersion    = errors.New("wrong schema version")
+	ErrInvalidBeacon   = errors.New("invalid beacon")
+)
+
+type realClient struct{}
```
piersy: This is a strange name for a type, what about just client?

bootstrap/updater_test.go
```diff
@@ -0,0 +1,418 @@
...
+	require.Len(t, got.Data[1].ActiveSet, 3)
+	require.Equal(t, types.HexToHash32("65af4350d28f3d953c6c6660e37954698839125fbda7aac3edcef469b2ad9e64"), got.Data[1].ActiveSet[0].Hash32())
+	require.Equal(t, types.HexToHash32("e46b23d64140357b16d18eace600b28ab767bfd7b51c8e9977a342b71c3a23dd"), got.Data[1].ActiveSet[1].Hash32())
+	require.Equal(t, types.HexToHash32("85de8823d6a0cd251aa62ce9315459302ea31ce9701531d3677ac8ba548a4210"), got.Data[1].ActiveSet[2].Hash32())
+}
+
+type checkFunc func(*testing.T, *bootstrap.VerifiedUpdate)
+
+func TestNew(t *testing.T) {
+	tcs := []struct {
```
piersy, line 149: Using the test configs approach makes debugging tests laborious when the config you want to debug is not the first config.

Making multiple calls to `t.Run` and extracting a function to do the main work of the test, solves that problem and also tends to result in less code.

E.g.:
```go
t.Run("recovery one", func(t *testing.T) {
	testNew(t, map[string]string{"00001-2023-03-18t22-26-13": update1}, checkUpdate1)
})

```

piersy:
Hey @countvonzero it would be useful for me if there was some package doc explaining what the purpose of the bootstrap package is, how its expected to be used and how it functions at a high level.

bootstrap/updater_test.go
dshulyak: can you elaborate please how does it make it more difficult? you can run any test case from this list

piersy: I should have clarified, I'm referring to debugging with vscode, a code lens provides a single click mechanism to debug a test, which runs the whole test.

If you want to debug a specific subtest you can either place your breakpoint and skip past all the subtests you are not interested in, manually make a debug config for the specific subtest and use that or comment out the entries for the other subtests. They are all more work than placing your breakpoint on the test you are interested in and clicking on the code lens.

countvonzero: do you mean that in vscode you can not invoke a subtests independently?

in goland if i can just place a breakpoint in the main shared code and invoke the subtest to debug.

my main objection to the approach you are suggesting is that there can be code in btwn multiple invocation of `t.Run()` that makes the tests not hermetic. and my experience, different from yours, is that this approach almost always result in more code.

bootstrap/mocks.go
countvonzero: thanks for the tip. done

bootstrap/updater_test.go
piersy: @countvonzero yes, invoking individual table driven subtests  is [not currently supported for vscode](https://github.com/golang/vscode-go/issues/1602)

With regard to the hermeticness, I guess that is up to the test writer, maybe its useful to run some code between invocations, in this case it doesn't seem necessary to run code between invocations.

With regard to code size I thought I'd just see what happens with this test you can see the result [here](https://github.com/spacemeshos/go-spacemesh/blob/a10abcb13d4790a1d028e553bae6c8e798b6a3a3/bootstrap/updater_test.go#L146-L195) in this case it seems to have reduced test size by about 15% I also find the test slightly easier to read like this, but maybe that is just me.

bootstrap/updater.go
```diff
@@ -1,3 +1,17 @@
+// Package bootstrap checks for the bootstrap/fallback data update from the
+// spacemesh administrator. This is intended as a short-term solution at the
```
piersy: What is the spacemesh administrator?

bootstrap/updater.go
```diff
@@ -1,3 +1,17 @@
...
+// beginning of the network deployment and should be removed once the network
+// is stable.
+//
+// The updater periodically checks for the latest update from a URL provided
+// by the spacemesh administrator, verifies the data, persists on disk and
+// notifies subscribers of a new update.
+//
+// Subscribers need to implement the following interface:
```
piersy: What components would subscribe to this?

bootstrap/updater.go
countvonzero: us. for genesis launch spacemesh will maintain a url that provides such network-wide update data.

bootstrap/updater.go
countvonzero: beacon (for the fallback beacon value) and hare oracle (for the active set).

bootstrap/updater_test.go
countvonzero: thanks for offering the code.

is this a deal breaker for you? if not i'd rather keep the test as hermetic as possible.

bootstrap/interface.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"context"
+	"net/url"
+)
+
+//go:generate mockgen -package=bootstrap -write_package_comment=false -destination=./mocks.go -source=./interface.go
+
+type Receiver interface {
+	OnBoostrapUpdate(*VerifiedUpdate)
```
dshulyak: my preference is to avoid dependency injection. this code can be pushed outside by providing a way to subscribe for updates, and then add a goroutine that listens for updates and integrates with go-spacemesh

```go

bootstrap := ...
bootstrap.Listen()
for update := bootstrap.Updates() {
     oracle.UpdateActiveSet(update.Epoch, update.ActiveSet)
     beacon.UpdateBeacon(update.Epoch, update.Beacon)
}
```

this is sometimes called dependency rejection (https://blog.ploeh.dk/2017/02/02/dependency-rejection/)

dshulyak (approved):

bootstrap/updater_test.go
piersy: Not a dealbreaker, but I prefer the version I linked due to (smaller size, easier debug for vscode and I think easier to read), so I don't see a reason not to use it, except for the hermeticness which I don't feel I've really understood. Could you give an example or point out in the [linked code](https://github.com/spacemeshos/go-spacemesh/blob/a10abcb13d4790a1d028e553bae6c8e798b6a3a3/bootstrap/updater_test.go#L146-L195) what are the problems that it poses with regards to hermeticness?

bootstrap/interface.go
piersy: @dshulyak I have my own thoughts on this but could you elaborate on what you see as the concrete benefits of the approach you proposed? ( I read the article but I'm not sure it maps very well to our codebase and the current situation)

bootstrap/schema.json
```diff
@@ -0,0 +1,55 @@
...
+  "description": "epoch data for bootstrapping and fallback",
+  "type": "object",
+  "required": ["version", "data"],
+  "properties": {
+    "version": {
+      "description": "version of the checkpoint file. should be compatible schema's $id",
+      "type": "string"
+    },
+    "data": {
+      "description": "the signed portion of this file",
```
piersy: Where is the signature stored/calculated/verified ?

bootstrap/schema.json
```diff
@@ -0,0 +1,55 @@
...
+      "type": "object",
+      "required": ["id", "epochs"],
+      "properties": {
+        "id": {
+          "type": "integer",
+          "exclusiveMinimum": 0
+        },
+        "epochs": {
```
piersy, line 21: Is this an array of all epochs since genesis?

bootstrap/updater.go
```diff
@@ -0,0 +1,388 @@
+// Package bootstrap checks for the bootstrap/fallback data update from the
+// spacemesh administrator. This is intended as a short-term solution at the
+// beginning of the network deployment and should be removed once the network
+// is stable.
```
piersy:
```suggestion
// Package bootstrap checks for the bootstrap/fallback data update from the
// spacemesh administrator (a centralised entity controlled by the spacemesh
// team). This is intended as a short-term solution at the beginning of the
// network deployment to simplify the process of recovering from a network
// failure and should be removed once the network is stable.
```

bootstrap/updater_test.go
countvonzero: in your code (quoted below), you could insert code for the next t.Run(), but then had to clean it for the second t.Run()
now only the `persisted` map matter, but maintenance or future changes, it may not be the case.

basically before every invocation one needs to worry about whether the state is set correctly. in the current code, each subtest is completely separate. i feel going forward it's less error-prone.

for me ease of maintenance is more important than curtailing the code to specific IDE needs.

```
func TestNew(t *testing.T) {
	t.Run("no recovery", func(t *testing.T) { testNew(t, nil, nil) })

	persisted := map[string]string{"00001-2023-03-18T22-26-13": update1}
	t.Run("recovery one", func(t *testing.T) { testNew(t, persisted, checkUpdate1) })

	persisted = map[string]string{
		"00014-2023-04-18T22-26-11": update3,
		"00001-2023-03-18T22-26-13": update1,
		"00002-2023-04-18T22-26-13": update2,
	}
	t.Run("recovery latest", func(t *testing.T) { testNew(t, persisted, checkUpdate3) })
}
```

bootstrap/interface.go
countvonzero: @dshulyak 
i had two considerations for the current design decisions
- keeping the code as simple as possible. currently the code doesn't require locking (single-threaded). if allowing to subscribe outside, then it requires locking. and i decide to keep it simple by passing fixed subscribers in the init flow

- in your model, which i did consider doing, 
```
bootstrap := ...
bootstrap.Listen()
for update := bootstrap.Updates() {
     oracle.UpdateActiveSet(update.Epoch, update.ActiveSet)
     beacon.UpdateBeacon(update.Epoch, update.Beacon)
}
```
the activeset data is potentially big, and since this data isn't immediately used to oracle or beacon because both components need to wait for X period of time to fall back to the value provided in the update, i don't want to keep multiple copies around in memory. in my prototype, beacon and oracle will each keep a reference to the latest update (the same copy as the updater) and use it as it needs.

bootstrap/schema.json
countvonzero: most likely we need one epoch for bootstrap fallback purposes. keeping it an array for now to leave options open.

UPD: to clarify, the one upcoming epoch .

bootstrap/schema.json
countvonzero: the signature is removed. so i'll update this comment. thanks

bootstrap/interface.go
dshulyak: > if allowing to subscribe outside, then it requires locking. and i decide to keep it simple by passing fixed subscribers in the init flow

by doing dependency injection this way you are multiplying complexity. nested code path is always more complex than flat code path. what i shared is flat.

i know that this is a very simple app, but just consider what should be done to understand how active set is updated. in my example i can tell immediately how it is updated. in your - i will have to find how this bootstrap is initialized and then how does it call this thing.

also you will need to have oracle.UpdateActiveSet (or alternative method) to handle changes in the consensus in a good way 

> the activeset data is potentially big, and since this data isn't immediately used to oracle or beacon because both components need to wait for X period of time to fall back to the value provided in the update, i don't want to keep multiple copies around in memory

i think there is some confusion about how many pointers needs to be around for long time. what i shared doesn't force runtime to keep multiple copies of data for long time, or copy any data... also memory concern is in general doesn't seem very relevant if you consider how this data compares to other data.

countvonzero: > nested code path is always more complex than flat code path. what i shared is flat.

this i agree with.

> i think there is some confusion about how many pointers needs to be around for long time. 

yes. i realized this later as well. oracle will keep the ptr to the backing array of the atx ids, which is the same ptr updater will hold onto until the next update.

ok. i'll change to the approach you described.

bootstrap/updater.go
countvonzero: modified a little. done

bootstrap/updater_test.go
piersy: Thanks for the explanation @countvonzero I think I understand. I have come across the problem you mention in the past, generally when re-using complex test setups to test multiple cases. In this scenario given the simplicity of the the test parameters I don't think it's a significant concern. In terms of hermeticness I think I would rate the two variants equally, in the variant I provided state could be messed up between calls to t.Run, but in the current version it seems to me that the same could happen by reusing values across multiple iterations of the for loop. Anyway I'm happy to leave it as it is.

bootstrap/updater.go
```diff
@@ -0,0 +1,407 @@
...
+func (u *Updater) latestUpdateId() uint32 {
+	u.mu.Lock()
+	defer u.mu.Unlock()
+	if u.latest != nil {
+		return u.latest.ID
+	}
+	return 0
+}
+
+func (u *Updater) DoIt(ctx context.Context) error {
```
piersy, line 187: Should this be exported (also `Load`)? From what I understand about the updater it seems like the public API should be:
```
Start
Subscribe
Close
```

I see it's used in the test but the test could be moved to this package.

bootstrap/updater.go
```diff
@@ -0,0 +1,407 @@
...
+	defer u.mu.Unlock()
+	for _, ch := range u.subscribers {
+		close(ch)
+	}
+	_ = u.eg.Wait()
+}
+
+func (u *Updater) latestUpdateId() uint32 {
+	u.mu.Lock()
+	defer u.mu.Unlock()
```
piersy, line 180: I don't think we need to lock here. `u.latest` seems like its only accessed by the single internal go routine. It looks like the lock is only required to lock overs subscribers, if so I think calling it `subscribersMu` would help with readability.

bootstrap/interface.go
dshulyak: > ok. i'll change to the approach you described.

btw, i wasn't trying to force a change, if there is no time and/or what you have is enough for it to work, it is certainly ok for me

bootstrap/updater.go
```diff
@@ -0,0 +1,407 @@
...
+		opt(u)
+	}
+	return u
+}
+
+func (u *Updater) Subscribe() chan *VerifiedUpdate {
+	u.mu.Lock()
+	defer u.mu.Unlock()
+	ch := make(chan *VerifiedUpdate, 10)
+	u.subscribers = append(u.subscribers, ch)
```
piersy, line 129: Do we need to push the latest update to the newly subscribed channel?

bootstrap/updater.go
countvonzero: it's a deliberate act on my part. in my view it is more important to put tests in XXX_test pkg here.
there are so many places in go-spacemesh where tests access private members directly and cause race issues.

`Load` and `DoIt` makes testing easier. i'll keep this in mind in my next iteration (there are prolly 2 more PRs) and see how i can reduce the exported methods.

bootstrap/updater.go
countvonzero: i don't think one should conditionally lock mutable data. 
the only data i didn't protect are those that are assigned in `New` and never get changed in the object's life cycle.

this module is not performance critical. for those that are changed outside of `New`, i think it's safer/future-proof to protect always.

bootstrap/updater.go
countvonzero: this is a design decision. there is no need for now. but i'll keep that in mind in the upcoming iterations.

bootstrap/updater.go
piersy: I would prefer keeping the public API clean.

I don't think putting the tests in a different package is helping if we then export all the private things needed for the test. In this case it looks like the exported methods could cause race issues if called concurrently, so they're exhibiting the very problem that this approach was trying to avoid.

In my opinion the way to deal with race issues in tests is to structure the tests correctly, and that requires an understanding of the underlying code, and that becomes harder if the public API is not clear about what should and should not be called.

bootstrap/updater.go
piersy: Well, we disagree on this point :)

I find locking of objects that do not need to be synchronised makes the code much harder to understand, since it implies to me that those objects are going to be accessed by multiple go-routines.

I'm also not convinced that it makes the code any safer since adding locked sections also raises the risk of deadlocks.

Also fine grained locking can lead to bugs such as the race conditions in the [broker](https://github.com/spacemeshos/go-spacemesh/pull/4130/commits/2ad05a19e1d0b357a13523314801a0ae1d89b6c7) that would have been much easier to track down if the parts had not been locked, since they would have triggered a data race. Having the fine grained locking technically made them safe from the point of view of the go race detector, but actually that hid the real race conditions.

I think if we actually followed the approach of locking over all mutable fields the code-base would become exceedingly complex.

bootstrap/updater.go
piersy: Is there some doc explaining the design decision?

bootstrap/updater_test.go
countvonzero: my guiding principal is: with little cost, make code future proof whenever i can. so the two approaches are actually different for me.

if you had done
```go
 func TestNew(t *testing.T) {
	t.Run("no recovery", func(t *testing.T) { testNoRecovery(t)})
	t.Run("recovery", func(t *testing.T) { testRecovery(t)})
}
```
where there are no dependencies btwn sub tests, then i'll definitely oblige. like i said, this approach usually resulted in more code.

bootstrap/updater.go
countvonzero: yes your preference is clear to me.

if i were developing low level libraries, i'd agree with you. and even there, you will get disagreement across the team.
my consideration is that we are trying to hit genesis testnet next week. so there is time crunch. i don't feel spending more time perfecting this PR is worthwhile.

bootstrap/updater.go
countvonzero: ok. we disagreed.

bootstrap/updater.go
countvonzero: nope. if in my follow up PRs it looks problematic to you, then i can change then.

countvonzero:
bors merge

bootstrap/updater.go
piersy: Lol :D

PR #4180, opened 2023-03-22
[Merged by Bors] - p2p/book: do not overwrite protected address

dshulyak:
preserves address of the bootnode, even if you got another seemingly valid address for it. 

there are two concerns:
- ip address may change, so we want to use dns for protected addresses
- it can be abused to feed invalid addresses


fasmat (approved):

dshulyak:
bors merge

Issue #4181, opened 2023-03-22
fetcher: atx request can cause node to crash


dshulyak:
![image](https://user-images.githubusercontent.com/1964576/226920050-88efcc3e-ec0c-4114-a6c1-bfe2a870408d.png)

there are two problems:
- any can abuse this to crash node by sending less than 4 bytes, when handler expects atleast 4 bytes (to decode uint32)
- in network this seems to be caused by honest node

solution is to either check boundaries, or to use scale for the every message, and avoid doing manual uint32 decoding

PR #4182, opened 2023-03-22
[Merged by Bors] - p2p/server: dont treat eof as success

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4181

binary.ReadUvarint was misused. if connection is terminated concurrently with the read - binary.ReadUvarint will treat it as success, varint.ReadUvarint does the right thing and will fail with UnexpectedEOF


countvonzero (approved):

dshulyak:
bors merge

Issue #4164, opened 2023-03-17
cache: reduce amount of data stored in atxs cache and store all such headers for current and next epoch


dshulyak:
we store whole atx header in the cache, however the most of it we don't need after validating atxs. such things as indices are not needed, but we still keep them in memory.

list of critical things, that are needed in other protocols: tick height, base tick height, weight, public key, atx id (as a key in the cache)

the reason for minimizing the size of the in-memory atx, is to keep at least one full epoch of atxs. for example 10 000 would be nice. the estimated difference would be around 10MB with indices (given that indices is atleast 1KB), and ~1MB if we store only critical parts

PR #4184, opened 2023-03-23
[Merged by Bors] - hare: cache tortoise active set

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4164

without cache hare doesn't work with large number of activations, as it takes too long if eligibility verification triggers load of 1000s of activations for every single message

### DIFF
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -59,9 +59,8 @@ type Oracle struct {
 	vrfVerifier    vrfVerifier
 	nonceFetcher   nonceFetcher
 	layersPerEpoch uint32
-	// map from cacheKey to map[types.NodeID]uint64
-	activesCache cache
-	cfg          config.Config
+	activesCache   cache
+	cfg            config.Config
 	log.Log
 }
 
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -59,8 +59,9 @@ type Oracle struct {
 	vrfVerifier    vrfVerifier
 	nonceFetcher   nonceFetcher
 	layersPerEpoch uint32
-	activesCache   cache
-	cfg            config.Config
+	// map from cacheKey to map[types.NodeID]uint64
+	activesCache cache
+	cfg          config.Config
 	log.Log
 }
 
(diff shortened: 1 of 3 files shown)


hare/eligibility/oracle.go
```diff
@@ -59,8 +59,9 @@ type Oracle struct {
 	vrfVerifier    vrfVerifier
 	nonceFetcher   nonceFetcher
 	layersPerEpoch uint32
-	activesCache   cache
-	cfg            config.Config
+	// map from cacheKey to map[types.NodeID]uint64
+	activesCache cache
```
fasmat: lru has released v2.0 of their library a while ago and now offers a generic interface to their cache.

Maybe this would be a fitting place to update our codebase to use the new version of the library? Its not a super trivial change:

- gomock afaik doesn't understand generics yet, so we cannot expose a generic interface of lru with the intent of mocking it.
- at the moment I think we are caching the pointers to the objects rather than the objects themselves, which doesn't work this way any more with lru v2. This might even be an oversight in the current implementation

What are your thoughts?

hare/eligibility/oracle_test.go
```diff
@@ -232,8 +232,7 @@ func TestCalcEligibility_EligibleFromTortoiseActiveSet(t *testing.T) {
 
 	numMiners := 5
 	o.mBeacon.EXPECT().GetBeacon(layer.GetEpoch()).Return(beacon, nil).Times(numMiners)
-	// there is no cache for tortoise active set. so each signature will cause 2 calls to GetBeacon()
-	o.mBeacon.EXPECT().GetBeacon(start.GetEpoch()).Return(beacon, nil).Times(2 * numMiners)
+	o.mBeacon.EXPECT().GetBeacon(start.GetEpoch()).Return(beacon, nil).Times(1)
```
fasmat, line 235: would it be correct to expect `numMiners + 1` calls here?

dshulyak: they are using different parameters for GetBeacon

fasmat: of course, then ignore my comment 🙂

countvonzero (approved):

hare/eligibility/oracle.go
dshulyak: map[NodeID]uint64 is cached, i think it should work as expected with lru that uses generic. i would like to do it, if it wasn't for changing test code.

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #4185, opened 2023-03-23
[Merged by Bors] - Ignore systest for coverage and use -coverpkg for better accuracy

poszu:
## Motivation
The `systest` directory contains tests so we don't want to include it in test coverage report (otherwise we will be checking if test utilities are tested too).

With the `-coverpkg` flag, go will also include packages without _any_ tests and include coverage _across packages_ (for example, if test for package `X` executes some code from package `Y` - then w/o this flag it would not count as covering code in `Y`).

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -14,3 +14,4 @@ ignore:
   - "cmd"
   - "**/*_scale.go"
   - "**/*mock*.go"
+  - systest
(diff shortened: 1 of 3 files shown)


dshulyak (approved):
i realized that unit tests for systests utilities are useful, as code there turned out to be not that simple. but i guess thats optional

poszu:
> i realized that unit tests for systests utilities are useful, as code there turned out to be not that simple. but i guess thats optional

It could be useful to write tests for them too. Anyway, the coverage report for that utility code is lying as it counts using it in a test as being tested, which is not true...

poszu:
bors merge

PR #4186, opened 2023-03-23
[Merged by Bors] - systest: improvements after tests on larger scale

dshulyak:
changes:
- use deployment with 1 replica to manage apps. we absolutely have to redeploy apps (poet, spacemesh) if spot vm is deleted. pod will not reschedule by itself, so it is not a good fit
- deploy services only for poet and bootnode. i run into limits for number of services
- use local storage instead of pvc. local storage is significantly faster, also provisioning time is significantly faster. state won't recover if vm get deleted, but thats acceptable for testing
- scale number of bootnodes. 2 bootnodes for every 1000 of smeshers
- new configuration in systest/parameters/bignet

### DIFF
--- a/systest/cluster/cluster.go
+++ b/systest/cluster/cluster.go
@@ -27,22 +27,21 @@ var errNotInitialized = errors.New("cluster: not initialized")
 
 const (
 	defaultExtraData = "systest"
-	poetSvc          = "poet"
-	bootnodesPrefix  = "boot"
-	smesherPrefix    = "smesher"
+	poetApp          = "poet"
+	bootnodeApp      = "boot"
+	smesherApp       = "smesher"
 	poetPort         = 80
-	defaultBootnodes = 2
 
 	poetFlags    = "poetflags"
 	smesherFlags = "smesherflags"
 )
 
-func setName(podname string) string {
-	return podname[:len(podname)-2]
-}
-
-func headlessSvc(setname string) string {
-	return setname + "-headless"
+func defaultBootnodes(size int) int {
+	bsize := (size / 1000) * 2
+	if bsize == 0 {
+		return 2
+	}
+	return bsize
 }
 
 // MakePoetEndpoint generate a poet endpoint for the ith instance.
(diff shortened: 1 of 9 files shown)


systest/cluster/cluster.go
```diff
@@ -599,7 +599,7 @@ func minPeers(size int) int {
 	}
 	other := size / 3
 	if other > 100 {
-		return 100
+		return 50
```
countvonzero, line 602: what is the rationale here?

systest/cluster/nodes.go
```diff
@@ -142,45 +132,67 @@ func deployPoetPod(ctx *testcontext.Context, id string, flags ...DeploymentFlag)
...
+						WithConfigMap(corev1.ConfigMapVolumeSource().WithName(poetConfigMapName)),
 					).
-					WithResources(corev1.ResourceRequirements().
-						WithRequests(poetResources.Get(ctx.Parameters).Requests).
-						WithLimits(poetResources.Get(ctx.Parameters).Limits),
+					WithContainers(corev1.Container().
+						WithName("poet").
+						WithImage(ctx.PoetImage).
+						WithArgs(args...).
+						WithPorts(corev1.ContainerPort().WithName("rest").WithProtocol("TCP").WithContainerPort(poetPort)).
```
countvonzero: odd formatting here

countvonzero (approved):

systest/cluster/cluster.go
dshulyak: we set high-peers to 100, and at that point node prunes number of peers to low-peers. i had to set min-peers to be less than high-peers. actually > 100 then 50 looks a bit weird :) will reconsider how to set default value

systest/cluster/nodes.go
dshulyak: can you elaborate? i tried to fix it, but not sure whats odd actually)

dshulyak:
bors try

dshulyak:
bors try

systest/cluster/nodes.go
countvonzero: sorry. i had thought that WithName/Protocol/ContainerPort() should be at the same indentation as WithPorts(). 
my bad. it's good.

dshulyak:
bors try

poszu (approved):

dshulyak:
bors try

dshulyak:
bors merge

PR #4188, opened 2023-03-23
[Merged by Bors] - datastore: use tuple instead of fmt.Sprintf

dshulyak:
i was looking at profiles from scale test, and noticed that this gets on heap very often.
i think tuple is a cleaner approach, also nodeID.String encodes to hex


dshulyak:
image

fasmat (approved):
lgtm, again we should consider upgrading to lru v2 to maybe not need our `Cache[K comparable, V any]` wrapper around the cache any more 🙂

dshulyak:
i would like to switch in other change. it is just a small cleanup that i noticed :)

dshulyak:
bors merge

Issue #4189, opened 2023-03-23
hare: optimize code path for message handler 


dshulyak:
![image](https://user-images.githubusercontent.com/1964576/227221655-ff34ec39-32d7-425a-bf75-09b465361b33.png)

attached 10m of cpu profile.

[pprof.go-spacemesh.samples.cpu.016.pb.gz](https://github.com/spacemeshos/go-spacemesh/files/11051416/pprof.go-spacemesh.samples.cpu.016.pb.gz)

most of the time spent is ecvrf and extract node id. we can halve the time for extract by using canonical signature (~by 2x). but ecvrf will probably stay the way it is.

encoding/decoding is likely will stay the way it is. 

we spend quite a bit on initializing loggers, even if we don't write things. this can be optimized.

and eligibility oracle should be possible to optimize as well: 
![image](https://user-images.githubusercontent.com/1964576/227223051-371260be-0282-4e4f-8120-636e01048e0d.png)

PR #4191, opened 2023-03-24
[Merged by Bors] - hare/oracle: optimize oracle validation

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4189

- don't count total weight on every request
- initialize logger only when oracle computes active set


countvonzero (approved):

dshulyak:
bors merge

PR #4195, opened 2023-03-24
[Merged by Bors] - Made the max msg sizes in grpc adjustable

pigmej:
We are already observing errors in Explorer that say:
``` received message larger than max (14117712 vs. 4194304)```

Basically, if there is a lot of happening the default size is not enough for sending layers over GRPC correctly.

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -20,6 +20,8 @@ const (
 	defaultStartSmesherService     = false
 	defaultStartTransactionService = false
 	defaultStartActivationService  = false
+	defaultGrpcSendBufferSize      = 1024 * 1024 * 10
+	defaultGrpcRecvBufferSize      = 1024 * 1024 * 10
 
 	defaultSmesherStreamInterval = 1 * time.Second
 )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: api/config/config.go [hunk 1]
```diff
@@ -29,8 +31,12 @@ type Config struct {
 	StartGrpcServices   []string `mapstructure:"grpc"`
 	GrpcServerPort      int      `mapstructure:"grpc-port"`
 	GrpcServerInterface string   `mapstructure:"grpc-interface"`
-	StartJSONServer     bool     `mapstructure:"json-server"`
-	JSONServerPort      int      `mapstructure:"json-port"`
+	// GRPC send and receive buffer size
+	GrpcSendBufferSize int `mapstructure:"grpc-send-buffer-size"`
```
dshulyak: but it is not exactly about buffer, i think options should be max-send-msg-size, max-recv-msg-size

pigmej: Yeah makes sense. Addressed.

dshulyak (approved):

dshulyak:
bors merge

Issue #4139, opened 2023-03-10
crypto: use canonical ed25519 signature in all domains except HARE


dshulyak:
domains were recently added and specified here https://github.com/spacemeshos/go-spacemesh/blob/develop/signing/signer.go#L19-L23

all domains except hare will avoid public key extraction and can use canonical ed25519 (both curve-voi and stdlib are good). my preference is towards curve-voi for performance, and stricter verification. additionally we may use batch verification in the future during sync.

#### BEACON

ProposalMessage includes NodeID. FirstVotingMessage and FollowingVotingMessage have to be extended with PublicKey.

#### ATX

Initial ATX will have to include public key in the body of atx, to guarantee that initial atx is unique (not sure if this is necessary, if we are using strict signature). Public key can be always found by following the chain of references.

But in order to enable validation without downloading full chain of atxs, public key can be also attached to every message.

#### BALLOT    
 
Similarly to ATXs, ballot can recover public key by following chain of references. However we would like to enable more efficient validation as well, so that signature can be validated before doing any disk reads or requests from peers. 

#### POET

Public key needs to be sent as a part of the message.

#### Malfeasance proofs

It is not a separate domain, but the proofs will have to be extended with public key as well.

#### Summary & Changes

With several exceptions (hare, and messages that include only output from vrf) we can encode messages that are sent over the wire using next pattern:

message + signature + public key

it is equivalent to the next golang type

```go
type SignedMessage[T scale.Type] struct {
     Message  T
     Signature [64]byte
     PublicKey [32]byte
}
```

countvonzero:
regarding ATX

> But in order to enable validation without downloading full chain of atxs, public key can be also attached to every message.

with the prev_atx field, the public key can be looked up in database.

dshulyak:
> with the prev_atx field, the public key can be looked up in database.

it can be done both with atxs and ballots (if we preserve reference to atx id at least for reference ballot). but i think that it will be better to gossip public key together with the message. in order to validate signature as a first step in the validation pipeline

PR #4197, opened 2023-03-24
[Merged by Bors] - Make VrfSignature fixed size

fasmat:
## Motivation
Part of #4139: migrate all VRF signatures from `[]byte` to `types.VrfSignature` (`[80]byte`) and update related code.

## Changes
- Introduce type `VrfSignature` that is a `[80]byte` to be used for VRF signatures
- Some tests needed updating since they used invalid VRF signatures before
  - e.g. `[]byte{1}` is not a valid sig, but `types.VrfSignature{1}` is. They will have different hashes, so I had to adapt tests that change because of this
- I simplified finding the smallest VRF signature in weak coin

## Test Plan
Update existing tests.

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
bors try

beacon/message.go
```diff
@@ -19,10 +19,15 @@ type ProposalVrfMessage struct {
...
 	EpochID      types.EpochID
 	NodeID       types.NodeID
-	VRFSignature []byte `scale:"max=80"`
+	VRFSignature types.VrfSignature
 }
 
-type Proposal [4]byte
+const (
+	// ProposalSize in bytes.
+	ProposalSize = 4
```
countvonzero: this is deliberate to be the same as BeaconSize. please do not define a separate const for this.
the idea is to use the same entropy.

beacon/weakcoin/weak_coin.go
```diff
@@ -236,20 +240,20 @@ func (wc *WeakCoin) updateProposal(ctx context.Context, message Message) error {
...
-func (wc *WeakCoin) prepareProposal(epoch types.EpochID, nonce types.VRFPostIndex, round types.RoundID) ([]byte, []byte) {
+func (wc *WeakCoin) prepareProposal(epoch types.EpochID, nonce types.VRFPostIndex, round types.RoundID) ([]byte, types.VrfSignature) {
 	minerAllowance := wc.allowance.MinerAllowance(wc.epoch, wc.signer.NodeID())
 	if minerAllowance == 0 {
-		return nil, nil
+		return nil, types.EmptyVrfSignature
 	}
 	var broadcast []byte
-	var smallest []byte
+	smallest := new(big.Int).Lsh(big.NewInt(1), types.VrfSignatureSize*8)
```
countvonzero: i don't get this. shouldn't this be the largest possible value, 11111...111111? 

the code is more readable to just use []byte here, instead of converting back and forth from big.Int
how about smallest := bytes.Repeat([]byte{1}, types.VrfSignatureSize*8)?

beacon/weakcoin/weak_coin.go
```diff
@@ -236,20 +240,20 @@ func (wc *WeakCoin) updateProposal(ctx context.Context, message Message) error {
...
 	for unit := uint32(0); unit < minerAllowance; unit++ {
 		proposal := wc.encodeProposal(epoch, nonce, round, unit)
 		signature := wc.signer.Sign(proposal)
 		if wc.aboveThreshold(signature) {
 			continue
 		}
-		if smallest == nil || bytes.Compare(signature, smallest) == -1 {
+		if new(big.Int).SetBytes(signature[:]).Cmp(smallest) == -1 {
```
countvonzero: i think with above, we can do a simple bytes.Compare(signature[:], smallest)?

beacon/weakcoin/weak_coin.go
```diff
@@ -21,6 +21,9 @@ var (
 	errNotSmallest  = errors.New("proposal not smallest")
 )
 
+// biggest is 1 bigger than the largest value that can be represented by a VRF signature.
+var biggest = new(big.Int).Lsh(big.NewInt(1), types.VrfSignatureSize*8)
```
countvonzero: why use a big.Int to do this? why not a byte array of 1s?

countvonzero (commented):
can you also double-check how hare coin flip is derived. it's in preroundtracker.go coinflip

beacon/weakcoin/weak_coin.go
fasmat: Yes the biggest value would be:

```golang
[80]byte{0xff, ..., 0xff} // ... = 78x 0xff

new(big.Int).Lsh(big.NewInt(1), types.VrfSignatureSize*8) == [81]byte{0x01, 0x00, ..., 0x00} // ... = 78x 0x00, so exactly 1 bigger than the biggest size

// this is not the biggest size
bytes.Repeat([]byte{1}, types.VrfSignatureSize) == []byte{0x01, ..., 0x01} // ... = 78x 0x01
```

I wanted to have `smallest` initalized with a value outside of the range so I can check against it (basically the same as before `== nil`), but this doesn't seem to be necessary so I can just use `0xff ... ff`.

beacon/weakcoin/weak_coin.go
fasmat: Probably, I could get rid of `big.Int` and replace everything with `[]byte` or `[80]byte`, I just did it because I thought

```golang
smallest := new(big.Int).Lsh(big.NewInt(1), types.VrfSignatureSize*8)
```

is cleaner than

```golang
var smallest types.VrfSignature
for i := range smallest {
    smallest[i] = 0xff
}
```

but you are right the comparison would be simpler:

```golang
bytes.Compare(signature.Bytes(), smallest.Bytes()) == -1
```

vs

```golang
new(big.Int).SetBytes(signature[:]).Cmp(smallest) == -1
```

I'll switch back to arrays over `big.Int`

beacon/weakcoin/weak_coin.go
fasmat: that would be incorrect, see my other comment, but I'll try to make it work with arrays 🙂

hare/preroundtracker.go
```diff
@@ -47,7 +47,7 @@ func (pre *preRoundTracker) OnPreRound(ctx context.Context, msg *Msg) {
 	logger := pre.logger.WithContext(ctx)
 
 	// check for winning VRF
-	vrfHash := hash.Sum(msg.Eligibility.Proof)
+	vrfHash := hash.Sum(msg.Eligibility.Proof.Bytes())
 	vrfHashVal := binary.LittleEndian.Uint32(vrfHash[:4])
```
countvonzero: this is what i meant by inconsistent. we have 2 ways of comparing vrf values. 
not caused by your change of course.

signing/vrf.go
```diff
@@ -14,8 +14,10 @@ type VRFSigner struct {
 }
 
 // Sign signs a message for VRF purposes.
-func (s VRFSigner) Sign(msg []byte) []byte {
-	return ecvrf.Prove(s.privateKey, msg)
+func (s VRFSigner) Sign(msg []byte) types.VrfSignature {
+	var sig types.VrfSignature
+	copy(sig[:], ecvrf.Prove(s.privateKey, msg))
```
countvonzero: can you do
```
vrf := (*[80]byte)(sig)
return *vrf
```
to avoid the copy?

signing/vrf_test.go
```diff
@@ -76,9 +76,9 @@ func Test_VRFVerifier(t *testing.T) {
 	ok = VRFVerify(signer.NodeID(), []byte("different message"), sig)
 	require.False(t, ok, "VRF signature should not be verified")
 
-	sig2 := make([]byte, len(sig))
-	copy(sig2, sig)
-	sig2[0] = ^sig2[0] // flip all bits of first byte in the signature
+	var sig2 types.VrfSignature
+	copy(sig2[:], sig[:])
+	sig2[0] = 0
```
countvonzero: is this the same thing as flipping all bits?

signing/vrf.go
fasmat: yes that should be possible here, will update

signing/vrf_test.go
fasmat: No, this must have been a copy paste error, reverted the change 🙂

signing/signer.go
```diff
@@ -107,9 +107,7 @@ func (es *EdSigner) Sign(d domain, m []byte) types.EdSignature {
 	msg = append(msg, es.prefix...)
 	msg = append(msg, byte(d))
 	msg = append(msg, m...)
-	var sig types.EdSignature
-	copy(sig[:], ed25519.Sign(es.priv, msg))
-	return sig
+	return *(*[64]byte)(ed25519.Sign(es.priv, msg))
```
countvonzero: [EdSignatureSize]

signing/vrf.go
```diff
@@ -14,8 +14,8 @@ type VRFSigner struct {
 }
 
 // Sign signs a message for VRF purposes.
-func (s VRFSigner) Sign(msg []byte) []byte {
-	return ecvrf.Prove(s.privateKey, msg)
+func (s VRFSigner) Sign(msg []byte) types.VrfSignature {
+	return *(*[80]byte)(ecvrf.Prove(s.privateKey, msg))
```
countvonzero: [VrfSignatureSize]

countvonzero (approved):

fasmat:
bors try

hare/preroundtracker.go
fasmat: I updated the relevant code. Both in beacon and in hare the smallest VRF signature is now selected without hashing it first (but by comparing them litte instead of big endian) and for the coinflip the LSB of the signature is used.

common/types/signatures.go
```diff
@@ -27,11 +28,63 @@ func (s *EdSignature) DecodeScale(decoder *scale.Decoder) (int, error) {
...
+		}
+		if s[i] > x[i] {
+			return 1
+		}
+	}
+	return 0
+}
+
+// LSB returns the least significant byte of the signature.
+func (s *VrfSignature) LSB() byte {
```
countvonzero, line 88: since we are relying on this for coinflip. is it worth writing a sanity check test that verifies that out of 10K signatures, we at least get a 30/70 or 40/60 of different result?

countvonzero (commented):
LGTM

common/types/signatures.go
fasmat: sure, should be simple enough 🙂

fasmat: Added a test and ran it 100 times. Random signatures were 50:50 in their last bit with a deviation of less than 5% in every testrun.

fasmat:
bors merge

PR #4200, opened 2023-03-25
[Merged by Bors] - Network latency histogram

### LINKED ISSUE #4070: Observability to understand network propagation times

piersy:
## Motivation

We would like to understand how long messages are taking to propagate across the network, so that we may set protocol parameters accordingly.

Closes #4070

## Changes
Adds a prometheus histogram with buckets in the metrics package and records message latencies relative to the time messages should be sent.

The protocols for which we record message latency are:
Hare
Atx
Proposal
Beacon weak coin
Beacon proposal
Beacon first votes
Beacon following votes

The histogram has 3 labels, protocol, message type and sign. Message type is only useful when reporting latency for hare messages, since the hare protocol has multiple different message types, for protocols with only one message type, the protocol label is used to populate the message type label.

Removes old hare message latency histogram in order to consolidate all message latencies under a single histogram.
Removes `node.NodeClock` interface since it was introducing circular dependencies when trying to pass it to services constructed in the `node` package. Replacing references to it with the concrete clock didn't pose any problem.

## Test Plan
CI is passing.

## DevOps Notes


hare/algorithm.go
```diff
@@ -498,7 +500,9 @@ func (proc *consensusProcess) processMsg(ctx context.Context, m *Msg) {
 		log.String("msg_type", m.InnerMsg.Type.String()),
 		log.Int("num_values", len(m.InnerMsg.Values)))
 
-	reportLatency(m.InnerMsg.Type, m.Round, proc.clock)
+	// Report the latency since the beginning of the round
+	latency := time.Since(proc.clock.RoundEnd(m.Round - 1))
+	metrics.ReportMessageLatency(pubsub.HareProtocol, strings.ToLower(m.InnerMsg.Type.String()), latency)
```
dshulyak: could you change m.InnerMsg.Type.String() to return already lowercased string? ToLower involves string copy and allocation on heap, it is notthing huge, but there is also no reason for doing so

beacon/handlers.go
```diff
@@ -71,6 +72,9 @@ func (pd *ProtocolDriver) handleProposal(ctx context.Context, peer p2p.Peer, msg
 		return errMalformedMessage
 	}
 
+	latency := receivedTime.Sub(pd.bClock.proposalSendTime(m.EpochID))
+	metrics.ReportMessageLatency(pubsub.BeaconProposalProtocol, pubsub.BeaconProposalProtocol, latency)
```
dshulyak: i think it would be useful to have Beacon as a protocol, and Proposal/First/Followup/Weakcoin as a message type (the same as in hare). it will not be consistent with handlers (maybe we can change that too btw), but overall it will be easier to use

beacon/weakcoin/interface.go
```diff
@@ -24,3 +26,12 @@ type nonceFetcher interface {
 type allowance interface {
 	MinerAllowance(types.EpochID, []byte) uint32
 }
+
+// weakCoinClock interface exists so that we can pass an object from the beacon
+// package to the weakCoinPackage (as does allowance), this is indicatave of a
+// circular dependency, probably the weak coin should be merged with the beacon
+// package.
+// Issue: https://github.com/spacemeshos/go-spacemesh/issues/4199
+type weakCoinClock interface {
```
dshulyak: maybe skip this when generating mocks? it seems you are using your own implementation to mock it

dshulyak (approved):
please take at comments before merging

beacon/weakcoin/interface.go
piersy: I was unsure about what to do here:
* On the one hand it's an interface and following the convention it should live in `interface.go`.
* On the other hand putting it in `interface.go` creates a mock that we don't need.

Is there some other way to avoid generating the mock or are you just suggesting moving the interface definition to some other file?

beacon/handlers.go
piersy: Ok, yeah it did feel a bit odd to me to have 4 protocols for beacon, if we want to have all messages in the same protocol then we'll need to manage all messages with a single handler. I'll take a look and see what that change would entail.

hare/algorithm.go
piersy: Just to be clear, the reason I didn't make this change is because it will change a number of logs (difficult to say which ones exactly without doing a fair bit of investigation), so as long as that's not a problem I'm happy to make the change.

dshulyak: yes, i think it wont be a problem for anyone

beacon/handlers.go
dshulyak: i think those can be separate changes though. we don't need to change handlers in order to have metrics

beacon/weakcoin/interface.go
dshulyak: was pretty sure that there is an api for that

https://github.com/spacemeshos/go-spacemesh/blob/develop/genvm/core/types.go#L39-L42

beacon/handlers.go
piersy: Sure, it looks like there's a fair bit of work to merge the handlers so I'll add a ticket for that.

beacon/weakcoin/interface.go
piersy: Unfortunately it's not possible to use that approach for unexported interfaces, you need to use source mode.

```
Source mode generates mock interfaces from a source file.
It is enabled by using the -source flag. Other flags that
may be useful in this mode are -imports and -aux_files.
Example:
	mockgen -source=foo.go [other options]
```

So it looks like we're back to the 2 options I outlined above, or we could export all these interfaces, but I like that option less than the other two.

dshulyak: i think whatever you prefer is ok. i personally would make interfaces public in order for them to work with that source mode, and likely wouldn't put them in interfaces file at all

beacon/beacon.go
```diff
@@ -1139,3 +1145,32 @@ func (pd *ProtocolDriver) gatherMetricsData() ([]*metrics.BeaconStats, *metrics.
 
 	return observed, calculated
 }
+
+// beaconClock provides methods to determine the intended send time of messages
+// spceific to the beacon protocol.
+type beaconClock struct {
```
countvonzero: maybe rename this to something like messageTimer or something?

in this codebase, using the name `clock` implies it's ticking and driving something (as the layer is time-based).
but here you are using it to derive msg times.

piersy: I'm not sure about using timer because go has a timer and that is an object which does drive something, but clock is also not great, let me think about it :thinking:

piersy: Ok renamed to `beacon.messageTimes`

beacon/weakcoin/interface.go
piersy: I just moved the weakCoinClock interface out of interface.go.

Given that all these interfaces exist to either allow objects to be injected into the package or to allow mocking (i.e. none of the interfaces represented functionality from within the package), making them public didn't seem like a good option.

piersy:
bors merge

piersy:
bors merge

beacon/beacon.go
countvonzero: perfect.

PR #4201, opened 2023-03-27
[Merged by Bors] - p2p: disable flood and lower number of peers in eager mesh

dshulyak:
lowers amount of traffic in the gossip. based on 2000 test


countvonzero (approved):

dshulyak:
bors merge

PR #4202, opened 2023-03-27
[Merged by Bors] - db: parametrize number of connections and latency metering

dshulyak:
latency metering consumes a lot of data if enabled on every node. however it is useful to quickly identify database misuse. additionally, it will be useful to configure higher number of connections for doing read requests in parallel (e.g for api).

### DIFF
--- a/sql/database.go
+++ b/sql/database.go
@@ -8,6 +8,7 @@ import (
 
 	"crawshaw.io/sqlite"
 	"crawshaw.io/sqlite/sqlitex"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 var (
@@ -53,9 +54,10 @@ func defaultConf() *conf {
 }
 
 type conf struct {
-	flags       sqlite.OpenFlags
-	connections int
-	migrations  Migrations
+	flags         sqlite.OpenFlags
+	connections   int
+	migrations    Migrations
+	enableLatency bool
 }
 
 // WithConnections overwrites number of pooled connections.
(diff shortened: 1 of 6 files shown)


sql/database.go
```diff
@@ -72,6 +74,15 @@ func WithMigrations(migrations Migrations) Opt {
...
 
+// WithLatencyMetering enables metric that track latency for every database query.
+// Note that it will be a significant amount of data, and should not be enabled on
+// multiple nodes by default.
+func WithLatencyMetering(enable bool) Opt {
+	return func(c *conf) {
+		c.enableLatency = enable
+	}
+}
+
```
poszu, line 85: How about measuring every n-th query instead of all, like measuring only 1 in 5 queries? It should still give valuable data while being less resource-demanding.

For it:
1. convert `conf::enableLatency` to an integer. A `0` value - disabled. `1` - every guery, `N` - 1 in `N` queries,
2. count queries with an atomic counter.
3. measure when `(counter % conf::enableLatency) == 0`

poszu (approved):

sql/database.go
dshulyak: for example there are 2000 different queries - 2000 histograms. if we collect every query atleast once - all of them will be created. i think reducing number of observations, will not lower the amount of data in a noticeable way. 

it may lower it slightly if we will skip observations that are added to a specific bucket. but this is not very relevant

poszu: I meant that it is not needed to measure every query to get a good picture. If overall query/s is high and only every, say, 100th query is measured, there's a high chance it will measure every kind of query with 100x less data. 

There is still an option to disable it entirely and enable it for every query like with a bool. IMO it's a more flexible approach.

dshulyak: >  there's a high chance it will measure every kind of query with 100x less data.

it is not the case. it will be almost same amount of data, regardless of how often you call observe for a query

data will be represented as:

query_1_bucket_0_count=10
query_1_bucket_1_count=30
query_2_bucket_0_count=45

and so on, if i will have 1 instead of 10 - total amount of collected data won't change by much.

dshulyak:
bors merge

PR #4203, opened 2023-03-27
[Merged by Bors] - proposals: check atx availability in tortoise state

### LINKED ISSUE #4177: optimize proposals handler
it takes an unreasonably long time to process proposal message that includes active set.
image

there might be other easy optimizations, it makes sense to get the profile for a handler and optimize just by looking at top time spenders.

ideal solution will use tortoise state to check what do we have in it. since we are relying on tortoise to decode ballot - everything relevant must be available, otherwise this is critical undefined behavior. if something is not found tortoise can return next error:

```go
type ErrorMissingData struct {
     Activations []ID
     Ballots []ID
}
```

there is also a simpler alternative. add lru cache, but it should be large enough to fit all atxs in the epoch, so it is most likely will require https://github.com/spacemeshos/go-spacemesh/issues/4164 . and generally it will be less clean compared to using tortoise state.

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/4177

this change introduces a fast path for verifying atxs availability for receiving proposals. tortoise always has to keep activations from the latest number of epochs that fit into sliding window, in order to efficiently decode ballots and compute thresholds.

this change intentionally doesn't use cache introduced in datastore, as that cache has a significantly worse caching policy (it will drop things out of memory due to lru). in contrast tortoise drops things only after sliding window. this results in more predictable computation cost to verify proposal. and also prevents ddos-ing proposals handler by making it drop relevant atxs from the cache.

### DIFF
--- a/proposals/handler.go
+++ b/proposals/handler.go
@@ -528,7 +528,7 @@ func (h *Handler) fetchReferencedATXs(ctx context.Context, b *types.Ballot) erro
 	if b.EpochData != nil {
 		atxs = append(atxs, b.EpochData.ActiveSet...)
 	}
-	if err := h.fetcher.GetAtxs(ctx, atxs); err != nil {
+	if err := h.fetcher.GetAtxs(ctx, h.decoder.GetMissingActiveSet(b.Layer.GetEpoch(), atxs)); err != nil {
 		return fmt.Errorf("proposal get ATXs: %w", err)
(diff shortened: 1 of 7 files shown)


countvonzero (approved):
nice.

dshulyak:
bors merge

PR #4211, opened 2023-03-28
[Merged by Bors] - fetcher: codec error in handleEpochInfoReq

### LINKED ISSUE #4181: fetcher: atx request can cause node to crash

dshulyak:
close: https://github.com/spacemeshos/go-spacemesh/issues/4181

turns out original problem was already fixed by @fasmat , i just noticed that we don't bubble up the error


countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #4213, opened 2023-03-28
[Merged by Bors] - node, api: split out private api and rework api configuration

### LINKED ISSUE #3883: Extract Node Management commands from public APIs to IPC
## Description
Currently, we have only public API endpoints (without any authentication).
So it makes it possible to operate the remote Node which exposes API. So it is a big security issue _(let's say someone exposes his API to the community just to support the network reliability — then anyone is able to turn off his node, change poet, start smeshing)_.
Meanwhile, we still need these handles to operate the Node without restarting it.

## Solution
As we discussed in the dev meeting, we want to extract these endpoints from the public GRPC API into a separate IPC API.
For consistency and speed of implementation, it still can use GRPC but over IPC sockets (UDS, NamedPipes) instead of TCP or just listen to another TCP port.

Let's call it **Admin GRPC API**
- Names and implementation of services and methods can be kept.
- It should not contain duplicates of public API endpoints.
- Should listen for UDS / NamedPipe / not exposed port.

## Endpoints
1. [ ] `NodeService.Shutdown`
    Not sure that we need it in the new API.
    - Already "deprecated" in https://github.com/spacemeshos/go-spacemesh/commit/36d1a996d804f242e8776409e16ad6a12832d27a#diff-b18dd59d4dba594837ef768659a50f3a371bf963f928b48ca9a03972969a85ccR120-R122,
    - but JSON API still supports the endpoint,
2. [ ] `NodeService.UpdatePoetServer`
3. [ ] `SmesherService.StartSmeshing`
4. [ ] `SmesherService.StopSmeshing`
5. [ ] `SmesherService.SetCoinbase`
6. [ ] `SmesherService.SetMinGas`
     It was not implemented yet. So I'm not sure that we need to copy that (adding endpoints is always easier than removing them).

## Acceptance criteria
- [ ] All endpoints listed above should be handled by Admin GRPC API
   - [ ] Do not forget to bring back the implementation of `NodeService.Shutdown` in the new API
- [ ] All listed endpoints should not be presented in Public API (GRPC & JSON)
- [ ] IPC API  should work well on Unix / Windows (via Named Pipes)

## Affected API consumers
Since nothing except Smapp (and outdated Smrepl) operates the Node, then we'll need to make changes only in Smapp:
1. [x] already using `SIGTERM` instead of calling `NodeService.Shutdown`: https://github.com/spacemeshos/smapp/pull/1047/commits/7d822ee9040ac6ce44c7e333d2a929a4e3888b9d -- and I don't think we need to switch to IPC for this case
2. [x] does not use `NodeService.UpdatePoetServer` (but in the future may need it)
3. [ ] uses `SmesherService.StartSmeshing`,
4. [ ] uses `SmesherService.StopSmeshing`,
5. [x] does not use `SmesherService.SetCoinbase`, but has some stale code for it
6. [x] does not use `SmesherService.SetMinGas`, since it was never implemented

❗ Please, mention Smapp team members (@brusherru @maparr) in the PR for this issue.

## Environment
 - Node Version: develop https://github.com/spacemeshos/go-spacemesh/commit/4894b2d39f7cb745a6bd4a49b6c8a1a285b8b4a5

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3883
needs: https://github.com/spacemeshos/api/pull/226

```
      --grpc-json-listener string                     Socket for the grpc gateway for the list of services in grpc-public-services. If left empty - grpc gateway won't be enabled.
      --grpc-private-listener string                  Socket for the list of services specified in grpc-private-services. (default "127.0.0.1:9093")
      --grpc-private-services strings                 List of services that must be kept private or exposed only in secure environments. (default [smesher])
      --grpc-public-listener string                   Socket for the list of services specified in grpc-public-services. (default "0.0.0.0:9092")
      --grpc-public-services strings                  List of services that are safe to open for the network. (default [debug,gateway,global,mesh,transaction,node])
      --grpc-recv-msg-size int                        GRPC api recv message size (default 10485760)
      --grpc-send-msg-size int                        GRPC api send message size (default 10485760)
```

we support two different sockets where services can be enabled. public (default to 0.0.0.0:9092) and private (defaults to 127.0.0.1:9093). by default we run smesher on private, and everything else on public. this can be changed by option above.

---

### <samp>🤖 Generated by Copilot at 2abafcf</samp>

This pull request refactors the API configuration and implementation to support multiple GRPC servers and services, and adds a new feature to allow smeshers to update and switch between PoET servers. It also updates the tests, logging, and presets to use the new configuration and services. It modifies the following files: `api/config/config.go`, `api/grpcserver/grpc.go`, `api/grpcserver/grpcserver_test.go`, `api/grpcserver/http_server.go`, `api/grpcserver/node_service.go`, `api/grpcserver/smesher_service.go`, `api/grpcserver/transaction_service_test.go`, `cmd/base.go`, `cmd/node/node_test.go`, `cmd/node/node.go`, `activation/interface.go`, `activation/mocks.go`, `api/grpcserver/smesher_service_test.go`, `api/node.go`, `cmd/bootstrapper/generator_test.go`, `cmd/root.go`, `config/presets/fastnet.go`, `config/presets/testnet.go`, `go.mod`, and `systest/parameters/fastnet/smesher.json`.


brusherru:
One important moment, `NodeService` contains some endpoints that should work well on a remote API: `Status`, `Build`, `Version`, `Echo`, `StatusStream`. However, the rest endpoints should be private or completely removed: `UpdatePoetServers`, `ErrorStream`, `SyncStart` (remove?), `Shutdown` (already deprecated as I know).

Instead of doing some complicated separation of endpoints within the Service, I propose to move them to another Service (e.g. `AdminService`). What do you think?

dshulyak:
> Instead of doing some complicated separation of endpoints within the Service, I propose to move them to another Service (e.g. AdminService). What do you think?

sure, if thats what you prefer i will move them

dshulyak:
@brusherru actually does it makes sense if i remove Shutdown/StartSync completely, and move UpdatePoetServers to Smeshing api. and move Node to the list of public services?

brusherru:
@dshulyak yes, sounds good to me! And much better than one more service! :)

dshulyak:
@brusherru i moved stuff around the way we discussedd

api/config/config.go
```diff
@@ -2,132 +2,53 @@
...
-		StartSmesherService:     defaultStartSmesherService,
-		StartTransactionService: defaultStartTransactionService,
-		StartActivationService:  defaultStartActivationService,
-		GrpcSendMsgSize:         defaultGrpcSendMsgSize,
-		GrpcRecvMsgSize:         defaultGrpcRecvMsgSize,
-
-		SmesherStreamInterval: defaultSmesherStreamInterval,
+		PublicServices:        []Service{Debug, Gateway, GlobalState, Mesh, Transaction, Node},
+		PublicListener:        "0.0.0.0:9092",
+		PrivateServices:       []Service{Smesher},
```
countvonzero, line 37: it almost feels like we should call this Admin

countvonzero (approved):
please update the PR description. thanks

brusherru (approved):
LGTM

pigmej (approved):

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

Issue #4174, opened 2023-03-21
Logging configuration


piersy:
## Description
Currently there is no way to configure a global log level, logging must be configured on a per module basis. Also the AppLogger acts as a sort of gate to the log levels set for each module. If a modules log is set to a more detailed level than that of the AppLogger then it will have its level set to the default of info. Thus it's impossible to set the log level of any module to a certain detail level unless the AppLogger is at least that level.

It should be possible to set a global default log level and it should also be possible to configure the level of each module independently.

We will likely want to use `zap.AtomicLevel` to achieve this.

## Affected code
Log config code

This issue appears in commit hash: a57b3305070b11c9f9fb4e427308224791302172

dshulyak:
@piersy do you want to submit fix for original problem? i can do that if you plan to work on bigger refactoring longer, but i think we need fix asap

countvonzero:
> This issue appears in commit hash: https://github.com/spacemeshos/go-spacemesh/commit/a57b3305070b11c9f9fb4e427308224791302172

i cannot put two and two together. what is wrong with this commit logging wise? @piersy

PR #4218, opened 2023-03-28
[Merged by Bors] - Fix module level logging

piersy:
## Motivation
Setting module log levels was not working properly.

The app log was erroneously being re-assigned using `app.addLogger`, which if no specific level was provided for the given module would default to info. So in this case the app log was being set to an info level log, even though it was hardcoded to debug at construction time.

The result was that no module level log could set levels lower than info.

Relates to #4174

## Changes
This change simply deletes the re-assignment, and deletes the now unused `AppLogger` constant.

## Test Plan
Manually tested

## DevOps Notes

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -69,7 +69,6 @@ const (
 
 // Logger names.
 const (
-	AppLogger              = "app"
 	ClockLogger            = "clock"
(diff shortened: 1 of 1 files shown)


dshulyak (approved):
thanks

piersy:
bors merge

PR #4222, opened 2023-03-29
[Merged by Bors] - implement epoch atxs API

countvonzero:
## Motivation
part of #4089

## Changes
- implement stream API for query epoch atx ids

### DIFF
--- a/api/grpcserver/mesh_service.go
+++ b/api/grpcserver/mesh_service.go
@@ -15,8 +15,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
-const pageSize = 1000
-
 // MeshService exposes mesh data such as accounts, blocks, and transactions.
 type MeshService struct {
 	mesh           api.MeshAPI // Mesh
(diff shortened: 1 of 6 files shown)


api/grpcserver/mesh_service.go
```diff
@@ -567,3 +569,30 @@ func convertLayerStatus(in int) pb.Layer_LayerStatus {
...
 	}
 }
+
+func (s MeshService) EpochStream(req *pb.EpochStreamRequest, stream pb.MeshService_EpochStreamServer) error {
+	epoch := types.EpochID(req.Epoch)
+	atxids, err := s.mesh.EpochAtxs(epoch)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+	for i := 0; i < len(atxids); i += pageSize {
```
countvonzero: @dshulyak 
i realized that pagination + stream is unusual. 
my original goal was to serve large set of data over RPC. but maybe this is an overkill.
if you'd like me to do a simple stream without repeated field in EpochStreamResponse, i'll change the api.

dshulyak: yes, i think tech-wise pagination with streaming doesn't make much sense. the way TransactionResult works is that it load objects iteratively and writes it one by one to a stream, so max amount of memory that can be allocated by a stream is bounded by stream buffer.

in this change all atxs are loaded into memory and then paginated, not sure if there is anything good about that

countvonzero: done to stream atx one by one. thanks for the help.

dshulyak (approved):

countvonzero:
bors merge

PR #4230, opened 2023-03-30
[Merged by Bors] - Change Signature of ATXs from spacemesh to canonical ed25519

fasmat:
## Motivation
Part of #4139 

## Changes
- New type `signing.EdVerifier`
  - Is intended to be used instead of `PubKeyExtractor` for domain ATX
  - If used in other domains will panic (temporarily until other domains are migrated as well)
  - `PubKeyExtractor` now panics if used in ATX domain
- `EdSigner` now uses canonical ed25519 for ATX domain, spacemesh ed25519 for all other domains (until migrated)
- `InnerActivationTx` now has a field `NodeID` that must be set for initial ATXs and must not be set for non-initial ATXs
- `ActivationTx` now has a field `SmesherID` that is not part of the message body (not encoded into the signature) but can be used to verify the signature.
- `NodeID` and `ID` are now value types instead of pointers to be in line with other structures like `Ballot` and `Block`
- `CalcAndSetID` is now `Initialize` to be in line with other structures
- Move all that's not ATX from `types/activation.go` to `types/epoch.go` and `types/poet.go`.

## Test Plan
- existing tests pass
- new tests for `InnerActivationTx.NodeID` being present in initial ATXs and not present in non-initial

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
bors try

common/types/activation.go
```diff
@@ -191,11 +132,12 @@ type InnerActivationTx struct {
...
 	NIPost      *NIPost
 	InitialPost *Post
+	NodeID      *NodeID
 	VRFNonce    *VRFPostIndex
 
 	// the following fields are kept private and from being serialized
-	id                *ATXID    // non-exported cache of the ATXID
-	nodeID            *NodeID   // the id of the Node that created the ATX (public key)
+	id                ATXID     // non-exported cache of the ATXID
+	nodeID            NodeID    // non-exported cache of the NodeID that created the ATX
```
countvonzero: why do we still need `nodeID` when `ActivationTx.SmesherID` is always set?

activation/activation.go
```diff
@@ -666,5 +669,6 @@ func (b *Builder) GetPositioningAtxInfo() (types.ATXID, types.LayerID, error) {
 // SignAndFinalizeAtx signs the atx with specified signer and calculates the ID of the ATX.
 func SignAndFinalizeAtx(signer *signing.EdSigner, atx *types.ActivationTx) error {
 	atx.Signature = signer.Sign(signing.ATX, atx.SignedBytes())
-	return atx.CalcAndSetID()
+	atx.SmesherID = signer.NodeID()
```
countvonzero, line 671: this is not necessary if you set SmesherID in `NewActivationTx()`

activation/handler.go
```diff
@@ -487,13 +496,10 @@ func (h *Handler) handleAtxData(ctx context.Context, peer p2p.Peer, data []byte)
 
 	atx.SetReceived(receivedTime.Local())
 
-	nodeID, err := h.extractor.ExtractNodeID(signing.ATX, atx.SignedBytes(), atx.Signature)
-	if err != nil {
-		return fmt.Errorf("failed to derive Node ID from ATX with sig %v: %w", atx.Signature, err)
-	}
-	atx.SetNodeID(&nodeID)
+	h.edVerifier.Verify(signing.ATX, atx.SmesherID, atx.SignedBytes(), atx.Signature)
```
countvonzero: check result

common/types/activation.go
```diff
@@ -299,18 +241,28 @@ func (atx *ActivationTx) MarshalLogObject(encoder log.ObjectEncoder) error {
 	return nil
 }
 
-// CalcAndSetID calculates and sets the cached ID field. This field must be set before calling the ID() method.
-func (atx *ActivationTx) CalcAndSetID() error {
+// Initialize calculates and sets the cached ID field. This field must be set before calling the ID() method.
+func (atx *ActivationTx) Initialize() error {
+	if atx.ID() != EmptyATXID {
+		return fmt.Errorf("ATX already initialized")
```
countvonzero, line 244: either panic or return nil.
this is a programming error.

common/types/activation.go
```diff
@@ -299,18 +241,28 @@ func (atx *ActivationTx) MarshalLogObject(encoder log.ObjectEncoder) error {
...
 		return fmt.Errorf("bad message hash")
 	}
 
-	id := ATXID(CalcObjectHash32(atx))
-	atx.id = &id
+	h := hash.New()
+	if _, err := h.Write(atx.MsgHash[:]); err != nil {
+		return fmt.Errorf("failed to write to hash")
+	}
+	if _, err := scale.EncodeByteSlice(scale.NewEncoder(h), atx.Signature[:]); err != nil {
```
countvonzero: i don't understand this change. i thought the only change we want to do is to exclude signature from hashing ID.

common/types/activation.go
```diff
@@ -299,18 +241,28 @@ func (atx *ActivationTx) MarshalLogObject(encoder log.ObjectEncoder) error {
...
-func (atx *ActivationTx) CalcAndSetID() error {
+// Initialize calculates and sets the cached ID field. This field must be set before calling the ID() method.
+func (atx *ActivationTx) Initialize() error {
+	if atx.ID() != EmptyATXID {
+		return fmt.Errorf("ATX already initialized")
+	}
+
 	if atx.Signature == EmptyEdSignature {
-		return fmt.Errorf("cannot calculate ATX ID: sig is nil")
+		return fmt.Errorf("ATX must be signed before initialization")
```
countvonzero: this comment makes less sense to me than the one one.
`Initialize()` is called on the gossip handler path as well as miner generating atx. the msg implied it's only on the miner path.
same for ballot/propsal `Initialize()`

common/types/ballot.go
```diff
@@ -230,21 +230,19 @@ func (b *Ballot) Initialize() error {
...
-	if err != nil {
-		return fmt.Errorf("failed to encode inner ballot for hashing")
+	if _, err := h.Write(b.MsgHash[:]); err != nil {
+		return fmt.Errorf("failed to write to hash")
 	}
-	_, err = scale.EncodeByteSlice(scale.NewEncoder(h), b.Signature[:])
-	if err != nil {
-		return fmt.Errorf("failed to encode byte slice")
+	if _, err := scale.EncodeByteSlice(scale.NewEncoder(h), b.Signature[:]); err != nil {
+		return fmt.Errorf("failed to encode signature")
```
countvonzero: don't change anything for ballot in this PR.

activation/activation.go
fasmat: I find it more intuitive this way. It's the identity that signed the ATX and the ATX is signed here.

common/types/activation.go
fasmat: Ballot and blocks also behave this way? ATX is the only odd one out that panics instead of returning an error if initialize is called twice.

activation/activation.go
countvonzero: NewActivationTx() is only called by the miner.
it doesn't make sense to do two-staged value assignment. one there and one here.

common/types/activation.go
countvonzero: for ballot/proposal/block, Initialize() is also called on the gossip handler path. ok. i see now you also use this in atx gossip handler path.

activation/activation.go
countvonzero: the goal i want to achieve here is having all atx data filled in one place in the ATX build codepath.

fasmat: But signature is also assigned here? `types.NewActivationTx` sets only fields of `InnerActivationTx`, `SignAndFinalize` signs the transaction and sets the remaining fields of the `ActivationTx`?

common/types/ballot.go
fasmat: I can move this to its own separate PR. I just noticed that a ballot is hashed twice when calling `Initialize`

fasmat: PR: https://github.com/spacemeshos/go-spacemesh/pull/new/ballot-initialize-double-hash

common/types/activation.go
fasmat: I don't read that from #4139. The only change is to use canonical ED signatures instead of our custom library.

I know the addition of the `NodeID` field in the initial ATX was proposed to make this possible, but I would rather that field again, since we are broadcasting the Node ID for signature validation anyway and the data would be redundant.

The alternative would be to not include signature in ATX ID calculation, but should this change then also be applied to all other messages with signatures (e.g. Ballots)? Also we would then need to enforce strict signatures, which might not be supported by some implementations of ed25519.

fasmat: Regarding the actual change here: `CalcObjectHash32` encodes and hashes the atx. We already have the hash in `atx.MsgHash` and verified it already against `atx.HashInnerBytes`.

So instead I just take the `atx.MsgHash` and hash it together with `atx.Signature` (like `Ballot` does).

countvonzero: > The alternative would be to not include signature in ATX ID calculation, 

this is the conclusion of the research meeting, and recorded by you https://spacemesh.slack.com/archives/C89QJJY3G/p1678395548603799

> but should this change then also be applied to all other messages with signatures (e.g. Ballots)? 

yes. but it can be done in separate PR.

> Also we would then need to enforce strict signatures, which might not be supported by some implementations of ed25519.

what dose strict signatures mean?

countvonzero: > So instead I just take the atx.MsgHash and hash it together with atx.Signature (like Ballot does).

the old ID calculation include the whole content of InnerActivationTx, ATXMetadata (not just the hash) and the Signature. we just need to remove Signature here.

fasmat: I don't understand why we have `ATXMetadata` as part of the ATX.

Target Epoch is derived from the positioning ATX and doesn't need to be included into the ATX. The MsgHash will be identical to the ATXID when we remove the signature from the ID calculation.

Shall this type be removed completely?

activation/activation.go
countvonzero: bottom line. i don't think there should be separate nodeID in InnerActivationTx and SmesherID in ActivationTx. why not just keep the latter only?

common/types/activation.go
fasmat: > what dose strict signatures mean?

Depending on the signature scheme there might be multiple valid signatures for the same message and key. Strict means there are some checks in place that only one of those signatures is accepted as valid.

Now that I think about it, this might not actually be an issue. We just need to check what happens if the same (valid) ATX is broadcasted twice with different (valid) signatures. Now those ATX would have different IDs and the node would be flagged as malicious. Afterwards both ATX would be valid.

countvonzero: maybe it's worth adding a sanity check on ID calculation so there is no ambiguity.

activation/activation.go
fasmat: I haven't checked yet if removing `nodeId` from the ATX breaks anything 😉, that's also why I haven't responded to the other comment about it yet.

common/types/activation.go
countvonzero: ultimately i think just writing atx.MsgHash[:] is sufficient for unique ID?

countvonzero: > I don't understand why we have ATXMetadata as part of the ATX.

it's there for generating malfeasance proof. the signature signs that hash and epoch. if two different msg for the same epoch appear in the network, the signer is considered malicious.

fasmat: > it's there for generating malfeasance proof. the signature signs that hash and epoch. if two different msg for the same epoch appear in the network, the signer is considered malicious.

It's also encoded and broadcasted with every ATX although the data inside can be derived by the node and doesn't necessarily need to be broadcasted.

Regarding malicious behaviour detection: If we make the hash of `InnerActivationTx` the new ID it could replace `MsgHash` I'm not sure about `TargetEpoch`. I can't find any place where this value is validated to be `atx.PubLayerID.GetEpoch() + 1`. What if a node just uses a different number here?

fasmat: I think `TargetEpoch` might need to be replaced with `atx.PubLayerID`. Malicious ATXs would then be two that are less than `layersPerEpoch` (or some slightly smaller value) apart.

countvonzero: > It's also encoded and broadcasted with every ATX although the data inside can be derived by the node and doesn't necessarily need to be broadcasted.

yes. Dmitry mentioned that optimization for hare too.

> I can't find any place where this value is validated to be atx.PubLayerID.GetEpoch() + 1. What if a node just uses a different number here?

true that it wasn't checked in my implementation. but target epoch is part of the signed data.

dshulyak: i am changing atx to store PublishEpoch instead of PublayerID. Target epoch is always +1 to publish, both for equivocation purposes

common/types/activation.go
fasmat: Updated to remove `nodeId`

common/types/activation.go
fasmat: So how shall I proceed with signature creation for ATXs? If its OK for you I would do the following:

- Scale encode `InnerActivationTx`
- Hash of encoded data -> `ATXID`
- Signature over encoded data -> `atx.Signature`

This would a) remove the signature from ID calculation and b) remove the need to have both `MsgHash` and `ATXID`. This would also exclude `ATXMetadata.TargetEpoch` from both the ID and signature calculation, since it will be replaced by @dshulyak changes in the near future anyway?

Opinions? @dshulyak @countvonzero

activation/activation.go
fasmat: the private nodeId field was removed and replaced with `atx.SmesherId` where needed

common/types/activation.go
dshulyak: for signature input should be as small as possible, that satisfies equivocation: `PublishEpoch/TargetEpoch, MsgHash`. as this is the only data that needs to be kept around if there is an actual equivocation

both PublishEpoch or TargetEpoch don't make any practical difference for equivocation

Using hash of InnerActivationTx as ATXID and MsgHash sounds good to me

countvonzero: publish/target epoch does play a role in equivocation verification.

if signer only signs `MsgHash` but not `TargetEpoch`, someone could put together an atx from epoch N and one from N+X in the malfeasance proof.

i also responded in another thread that using `MsgHash` solely for ID is good.

activation/activation.go
```diff
@@ -611,13 +613,13 @@ func (b *Builder) createAtx(ctx context.Context, challenge *types.NIPostChalleng
 
 	atx := types.NewActivationTx(
 		*challenge,
-		&b.nodeID,
```
countvonzero, line 614: if you had assigned SmesherID inside this function you won't need to change other files....

countvonzero: not a request for change. just quibbling :P

common/types/activation.go
dshulyak: > both PublishEpoch or TargetEpoch don't make any practical difference for equivocation

i probably wrote it in a strange way, i meant that using any of those two for signature is ok

countvonzero: i might have jumped on the last sentence. your first sentence
 
> for signature input should be as small as possible, that satisfies equivocation: PublishEpoch/TargetEpoch, MsgHash. 

is perfectly clear. sorry.

go.sum
```diff
@@ -631,10 +631,11 @@ github.com/subosito/gotenv v1.4.2/go.mod h1:ayKnFf/c6rvx/2iiLrJUk1e6plDbT3edrFNG
 github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7 h1:epCh84lMvA70Z7CTTCmYQn2CKbY8j86K7/FAIr141uY=
 github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7/go.mod h1:q4W45IWZaF22tdD+VEXcAWRA037jwmWEB5VWYORlTpc=
 github.com/tarm/serial v0.0.0-20180830185346-98f6abe2eb07/go.mod h1:kDXzergiv9cbyO7IOYJZWg1U88JhDg3PB6klq9Hg2pA=
```
countvonzero, line 633: i don't understand why you routinely packed deps update in PRs.
this really is not a clean way to update deps. it defeats the purpose of depedantbot

countvonzero (approved):

go.sum
fasmat: I have (most) spacemesh repositories checked out locally in the same workspace. If one of the dependencies of go-spacemesh updates its `go.mod` file `go work sync` will also update the dependencies here. I'll try to avoid this but it makes changes across repositories much easier 🙂

fasmat:
bors merge

PR #4231, opened 2023-03-31
[Merged by Bors] - bootstrapper: stand alone library to generate bootstrap and fallback data

countvonzero:
## Motivation
part of #4089

## Changes
cmd/bootstrapper is a standalone binary that generates bootstrap/fallback data for each epoch.
the binary use a separate bootsrapper.Dockerfile for separate binary release from go-spacemesh.
this binary will be used in the following scenario:
- beacon protocol
  - no beacon generated in epoch N
  - N different beacon generated in epoch N
  
  we will know these failures before epoch N ends. in this case, admin runs the following:

```console
go-bootstrapper N+1 --beacon --bitcoin-endpoint=bitcoin-restful-endpoint --out=gs://my-bucket/my-file --creds=/path/to/gcloud-creds.json
```

- hare protocol
  - no blocks are generated for X layers in epoch N+1:

  we will know this after X layers into epoch N+1. in this case, admin runs the following:

```console
go-bootstrapper N+1 --actives --spacemesh-endpoint=node-grpc-endpoint --out=gs://my-bucket/my-file --creds=/path/to/gcloud-creds.json
```

this means we potentially will write data for the same epoch (N+1) twice. once in epoch N for beacon fallback, and once in epoch N+1 for hare activeset fallback.


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

bootstrap.Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# And compile the project
+RUN --mount=type=cache,id=build,target=/root/.cache/go-build make bootstrapper
+
+#In this last stage, we start from a fresh Alpine image, to reduce the image size and not ship the Go compiler in our production artifacts.
+FROM linux AS spacemesh
+
+# Finally we copy the statically compiled Go binary.
+COPY --from=builder /src/build/go-bootstrapper /bin/
+
+ENTRYPOINT ["/bin/go-bootstrapper"]
```
dshulyak, line 22: i think you can make significantly simpler image, there is no need to bring all deps that we use for spacemesh node

you can check systests image as an example

cmd/bootstrapper/bootstrapper.go
```diff
@@ -0,0 +1,135 @@
...
+	cmd.PersistentFlags().StringVar(&nodeEndpoint, "node-endpoint",
+		"localhost:9092", "specify grpc endpoint for a spacemesh node")
+
+	// network parameters
+	cmd.PersistentFlags().StringVar(&genesisTime, "genesis-time",
+		"", "Time of the genesis layer in 2019-13-02T17:02:00+00:00 format")
+	cmd.PersistentFlags().DurationVar(&layerDuration, "layer-duration",
+		5*time.Minute, "duration per layer")
+	cmd.PersistentFlags().Uint32Var(&layersPerEpoch, "layers-per-epoch",
+		4032, "number of layers per epoch")
```
dshulyak: it was definitely better to get all this data from the node

cmd/bootstrapper/bootstrapper.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			return err
+		}
+		logger := log.NewWithLevel("", lvl)
+		genesis, err := time.Parse(time.RFC3339, genesisTime)
+		if err != nil {
+			return fmt.Errorf("parse genesis time %v: %w ", genesisTime, err)
+		}
+
+		clock, err := timesync.NewClock(
```
dshulyak: i think it is better to make an api call, that will return how long you need to wait until certain point (epoch-offset) and get rid from this clock thing completely

cmd/bootstrapper/bootstrapper.go
```diff
@@ -0,0 +1,135 @@
...
+
+		ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+		defer cancel()
+		appErr := g.Run(ctx)
+
+		shutdownCxt, shutdownCancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer shutdownCancel()
+		if srv != nil {
+			if err = srv.Stop(shutdownCxt); err != nil {
+				logger.Error("failed to gracefully shutdown http server: %w", err)
```
dshulyak: please encode all necessary information into the error and print it only once at the top of the callstack

dshulyak:
what i don't understand, is how do you make a decision when to let protocol run normally, and when to query bootstrap for data? what i thought originally that there will be a command line app:

`bootstrapgen <epoch> --out <file|stdout> --spacemesh-endpoint <spacemesh> --bitcoin-endpoint <bitcoin>`

when we know that bootstrap is required someone will launch it and provide it will write a json file that will be served by nginx. ideally we do it once for genesis, otherwise we let protocol run as normal. the role of the human can also be automated by adding relevant metrics/logs and filtering for them.. but this is optional. you have an app that writes every epoch, how do you propose to filter this data?

for testing it is the same, i don't want to overwrite every epoch as it limits the testing.

dshulyak:
> for testing it is the same, i don't want to overwrite every epoch as it limits the testing.

i guess i answered some of my questions by looking at other change. i think auto-fallback doesn't cover a case when there is an activeset but it is not the same everywhere, but that would mean consensus failure, so maybe it is not that important

bootstrap.Dockerfile
countvonzero: yeah. i actually started with systests/Dockerfile, that didn't work and i struggled a little so ended up just copying go-spacemesh's Dockerfile. 

i'll address your other feedback and try again with a simpler image.

countvonzero:
> > for testing it is the same, i don't want to overwrite every epoch as it limits the testing.
> 
> i guess i answered some of my questions by looking at other change. i think auto-fallback doesn't cover a case when there is an activeset but it is not the same everywhere, but that would mean consensus failure, so maybe it is not that important

this binary has two utilities: bootstrap and fallback

if left running as a server, it will consistently generate fallback data for each epoch. the nodes will receive its update and make local decisions on whether to use the fallback data.

my thinking was, if the logic in the node is correct, then it shouldn't matter whether the fallback data is there or not.
i see that now it's flawed. as it's possible that some nodes use the fallback, and some nodes not. the fallback data actually becomes another source of disagreement. 

it seems better that the fallback data is only generated when the network fails. and when a node receive the fallback data, it HAS to use it so there is consensus.

let me break down the failures and admin action:
- beacon protocol
  - no beacon generated in epoch N
  - N different beacon generated in epoch N

we will know these failures before epoch N ends. in this case, admin runs the following:

```
bootstrapgen N+1 --beacon --out <file|stdout> --bitcoin-endpoint <bitcoin>
```

- hare protocol
  - no blocks are generated for X layers in epoch N+1:

we will know this after X layers into epoch N+1. in this case, admin runs the following:

```
bootstrapgen N+1 --activeset --out <file|stdout> --spacemesh-endpoint <spacemesh>
```

this means we potentially will write data for the same epoch (N+1) twice. once in epoch N for beacon fallback, and once in epoch N+1 for hare activeset fallback.

does this sound OK to you?

dshulyak:
it does sound ok to run it twice, i think it is also ok to overwrite activeset when there are problems with beacon

dshulyak (approved):
how would node work if endpoint is not available? will it throw errors/warnings in the log? i think it is better to expose some metrics that will track such thing, and log only successful update

countvonzero:
bors try

bootstrap.Dockerfile
countvonzero: significantly simplified. the only diff btwn my previous tries and this one was using alpine vs ubuntu. with alpine docker wasn't able to execute the binary.

cmd/bootstrapper/bootstrapper.go
countvonzero: done

cmd/bootstrapper/bootstrapper.go
countvonzero: done

cmd/bootstrapper/bootstrapper.go
countvonzero: done

countvonzero:
> how would node work if endpoint is not available? will it throw errors/warnings in the log? i think it is better to expose some metrics that will track such thing, and log only successful update

i will add such metrics in the next PR that actually tries to use the bootstrap/fallback data.

countvonzero:
bors merge

countvonzero:
bors merge

Issue #3774, opened 2022-11-21
ATX: remove restrictions on positioning ATX 


countvonzero:
## Description

currently positioning ATX must be from the previous epoch. this will prevent network from healing on its own when there are 1+ epochs without ATXs.

- positioning ATX can be a previous ATX of any smeshers from any previous epochs
- ATX need to declare explicitly its publish epoch (instead of inferring from its positioning ATX)

fasmat:
What was the inital reason for introducing the positioning ATX? If we allow any ATX to be chosen and explicitly set a publish epoch what prevents a smesher from always chosing the same ATX? At that point why have it at all?

cc @noamnelke @selfdual-brain

PR #4233, opened 2023-03-31
[Merged by Bors] - atxs: drop posiotining restriction requirement

dshulyak:
related https://github.com/spacemeshos/go-spacemesh/issues/3774

with this change any positioning atx is considered to be valid. additionally PubLayerId was changed to PublishEpoch. the rest are changes in tests.

i will submit part that picks publish epoch based on the current clock in a separate pr.

### DIFF
--- a/activation/validation.go
+++ b/activation/validation.go
@@ -210,14 +210,16 @@ func (*Validator) NIPostChallenge(challenge *types.NIPostChallenge, atxs atxProv
 func (*Validator) PositioningAtx(id *types.ATXID, atxs atxProvider, goldenATXID types.ATXID, pubepoch types.EpochID, layersPerEpoch uint32) error {
 	if *id == types.EmptyATXID {
 		return fmt.Errorf("empty positioning atx")
-	} else if *id != goldenATXID {
-		posAtx, err := atxs.GetAtxHeader(*id)
-		if err != nil {
-			return &ErrAtxNotFound{Id: *id, source: err}
-		}
-		if posAtx.PublishEpoch >= pubepoch {
-			return fmt.Errorf("positioning atx epoch (%v) must be before %v", posAtx.PublishEpoch, pubepoch)
-		}
+	}
+	if *id == goldenATXID {
+		return nil
+	}
+	posAtx, err := atxs.GetAtxHeader(*id)
+	if err != nil {
+		return &ErrAtxNotFound{Id: *id, source: err}
+	}
+	if posAtx.PublishEpoch >= pubepoch {
+		return fmt.Errorf("positioning atx epoch (%v) must be before %v", posAtx.PublishEpoch, pubepoch)
 	}
 	return nil
 }
--- a/activation/validation.go
+++ b/activation/validation.go
@@ -165,8 +165,8 @@ func (*Validator) InitialNIPostChallenge(challenge *types.NIPostChallenge, atxs 
 		if err != nil {
 			return &ErrAtxNotFound{Id: *challenge.CommitmentATX, source: err}
 		}
-		if !challenge.PubLayerID.After(commitmentAtx.PubLayerID) {
-			return fmt.Errorf("challenge publayer (%v) must be after commitment atx publayer (%v)", challenge.PubLayerID, commitmentAtx.PubLayerID)
+		if challenge.PublishEpoch <= commitmentAtx.PublishEpoch {
+			return fmt.Errorf("challenge pubepoch (%v) must be after commitment atx pubepoch (%v)", challenge.PublishEpoch, commitmentAtx.PublishEpoch)
 		}
 	}
 	return nil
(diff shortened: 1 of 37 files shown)


activation/validation.go
```diff
@@ -207,27 +207,17 @@ func (*Validator) NIPostChallenge(challenge *types.NIPostChallenge, atxs atxProv
...
 	if *id == types.EmptyATXID {
 		return fmt.Errorf("empty positioning atx")
-	}
-
-	if *id == goldenATXID && !publayer.GetEpoch().IsGenesis() {
-		return fmt.Errorf("golden atx used for positioning atx in epoch %d, but is only valid in epoch 1", publayer.GetEpoch())
-	}
-
-	if *id != goldenATXID {
+	} else if *id != goldenATXID {
```
fasmat: NIT: I think without the else and inverting the if makes it easier to read:

```golang
	if *id == types.EmptyATXID {
		return fmt.Errorf("empty positioning atx")
	}
	if *id == goldenATXID {
		return nil 
	}

	posAtx, err := atxs.GetAtxHeader(*id)
	if err != nil {
		return &ErrAtxNotFound{Id: *id, source: err}
	}
	if posAtx.PublishEpoch >= pubepoch {
		return fmt.Errorf("positioning atx epoch (%v) must be before %v", posAtx.PublishEpoch, pubepoch)
	}
	return nil
```

fasmat:
Just for my understanding: The positioning ATX is still needed for the TickHeight of the ATX and smeshers are incentivised to pick a recent one to get the highest possible TickHeight?

common/types/activation_test.go
```diff
@@ -26,12 +29,23 @@ func TestRoundEndSerialization(t *testing.T) {
...
+	var object types.ActivationTx
+	f := fuzz.NewWithSeed(1001)
+	f.Fuzz(&object)
+
+	buf := bytes.NewBuffer(nil)
+	enc := scale.NewEncoder(buf)
+	_, err := object.EncodeScale(enc)
+	require.NoError(t, err)
+
+	epoch := types.EpochID(rand.Uint32())
```
fasmat: The assigned value here is overwritten in the next line

```suggestion
	var epoch types.EpochID
```

common/types/activation.go
```diff
@@ -228,8 +223,7 @@ func (atx *ActivationTx) MarshalLogObject(encoder log.ObjectEncoder) error {
 		encoder.AddUint64("vrf_nonce", uint64(*atx.VRFNonce))
 	}
 	encoder.AddString("coinbase", atx.Coinbase.String())
-	encoder.AddUint32("pub_layer_id", atx.PubLayerID.Uint32())
-	encoder.AddUint32("epoch", uint32(atx.PublishEpoch()))
+	encoder.AddUint32("epoch", uint32(atx.PublishEpoch))
```
fasmat: NIT: Use the method you added 🙂 

```suggestion
	encoder.AddUint32("epoch", atx.PublishEpoch.Uint32())
```

common/types/epoch.go
```diff
@@ -11,49 +11,53 @@ import (
...
+}
+
 // EncodeScale implements scale codec interface.
-func (l EpochID) EncodeScale(e *scale.Encoder) (int, error) {
-	n, err := scale.EncodeCompact32(e, uint32(l))
+func (e EpochID) EncodeScale(enc *scale.Encoder) (int, error) {
+	n, err := scale.EncodeCompact32(enc, uint32(e))
 	if err != nil {
 		return 0, err
 	}
```
fasmat: I know this is unrelated to your changes, but I think this is not correct. `n` might be 0 but it also might not be (if some bytes were written to `enc`).

common/types/epoch.go
```diff
@@ -11,49 +11,53 @@ import (
...
 }
 
 // DecodeScale implements scale codec interface.
-func (l *EpochID) DecodeScale(d *scale.Decoder) (int, error) {
-	value, n, err := scale.DecodeCompact32(d)
+func (e *EpochID) DecodeScale(dec *scale.Decoder) (int, error) {
+	value, n, err := scale.DecodeCompact32(dec)
 	if err != nil {
 		return 0, err
 	}
```
fasmat: Again unrelated, but this is definitely wrong. To produce the error some bytes of the underlying `io.Reader`/`scale.Decoder` must have been consumed.

common/types/verified_activation.go
```diff
@@ -48,8 +48,8 @@ func (vatx *VerifiedActivationTx) MarshalLogObject(encoder log.ObjectEncoder) er
 		encoder.AddUint64("vrf_nonce", uint64(*vatx.VRFNonce))
 	}
 	encoder.AddString("coinbase", vatx.Coinbase.String())
-	encoder.AddUint32("pub_layer_id", vatx.PubLayerID.Uint32())
-	encoder.AddUint32("epoch", uint32(vatx.PublishEpoch()))
+	encoder.AddUint32("pub_layer_id", vatx.PublishEpoch.Uint32())
```
fasmat: I think this line can be removed

dshulyak:
> Just for my understanding: The positioning ATX is still needed for the TickHeight of the ATX and smeshers are incentivised to pick a recent one to get the highest possible TickHeight?

yes, afaik that part doesnt change. but now i am not sure if we should pick just the highest one, or the highest one from the latest epoch.

fasmat (approved):

dshulyak:
@fasmat thank you, fixed

dshulyak:
bors merge

dshulyak:
bors merge

Issue #4237, opened 2023-04-01
ballot: include active set hash in signature instead of the set itself


countvonzero:
## Description
as active set data size is large, the node should de-duplicate active set from peers and and transmit/store just one copy of it for all honest ATXs of the epoch.

- when generating the first ballot, we get all the atxs in the previous epoch
  - sort the active set (AS) and generate hash(AS)
  - sign the ballot that includes hash(AS)
  - gossip ballot, and attach AS to the gossip message
  - locally save hash(AS)->AS mapping
- when receiving a ballot via gossip:
  - if the gossip msg doesn't have AS attached, this ballot is invalid
  - validate the ballot as usual
  - locally save hash(AS)->AS mapping
- when a node requesting a ballot during sync, the peer send the ballot, also attaching the corresponding AS

for genesis, we can delay implementing the mapping. simply define the first ballot to always be serialized/deserialized as (x=ballot_containing_hash(AS), AS, signature_of_x). in other words, always send/receive the full data and assume that any other format of transmitting this data is malicious

after genesis, optimization can be done by not transmitting the actual active set and only fetch from peers by its hash.

dshulyak:
i think this was implemented

Issue #4087, opened 2023-02-26
restructure ATX for future pruning


countvonzero:
## Description

changing mesh data structure after mainnet launch is a protocol change. since we know in the future we are going to prune malicious ATXs from the mesh, we want to restructure the data structure such that

- facilitate pruning part of the data
- yet remain verifiable by new nodes

discussion on the impact of pruned ATXs
- an ATX references a malicious ATX as its `PrevATXID`: since they must come from the same identity, both ATXs will be pruned
- an ATX references a malicious ATX as its `CommitmentATX`, then such ATX doesn't get prioritized as described in #3593
- an ATX references a malicious ATX as its `PositioningATX`. since the positioning ATX can be arbitrary and is an input to PoET, making it difficult to prune
- honest identities reference atxs in the first ballot of the epoch, which then is used to compute number of eligible ballots and the ballot weight. if any of the ATXs is pruned, ballot weights cannot be calculated.

ideas:
- for the current epoch, implement https://community.spacemesh.io/t/grading-atxs-for-the-active-set/335
- for old epochs, simply trust the number of slots recorded in the reference ballots. this require adding a field in ballot eligibility.

@pigmej @dshulyak @noamnelke

dshulyak:
it makes sense to add this integer. i think following change works:
- validate that declared integer matches computed integer in the ballot handler if ballot is in the current epoch. it can be done with tortoise state efficiently. as it keeps reference ballots and atxs in memory.
- when doing tortoise, use integer declared in the ballot without additional validation
- when building block we may skip validation as well, as it will be implicit because of first point

dshulyak:
is it ready/not ready? it just hangs open

countvonzero:
> is it ready/not ready? it just hangs open

not ready. @noamnelke is preparing a spec for this.

PR #4238, opened 2023-04-01
[Merged by Bors] - exclude active set from ballot signature

countvonzero:
## Motivation
first part of #4237
part of #4087

## Changes
- exclude active set from ref ballot signature. include the active set instead.
- add a EligibilityCount in ballot (for total number of ballots in the epoch) to prepare for future atx pruning


countvonzero:
bors try

common/types/ballot.go
```diff
@@ -101,6 +105,9 @@ type InnerBallot struct {
 	// can be verified before decoding votes.
 	OpinionHash Hash32
 
+	// total number of ballots the smesher is eligible in this epoch.
+	EligibilityCount uint32
```
dshulyak, line 109: we need to validate that this number matches computed number, or do the alternative. easiest would be to validate it at (*Validator).CheckEligibility

if you don't validate it, and later this number is used in the tortoise - ballot weight will be incorrect. for example computed eligibility count is 10 and this was used when ballots were accepted in Validator.CheckEligibility, node accepted 10 ballots this way. but if EligibilityCount is 1 and it is used in tortoise, then every ballot will have weight of the full atx weight

alternative is to validate only ballots from the current epoch using computed count, and for previous epoch use specified count in the oracle

dshulyak (approved):
please fix validation stuff (in case i didn't miss it)

common/types/ballot.go
countvonzero: thanks. i missed that. added validation as part of (*Validator).CheckEligibility

countvonzero:
bors try

countvonzero:
bors merge

PR #4241, opened 2023-04-02
[Merged by Bors] - hare active set: bootstrap and use first consensus block

countvonzero:
## Motivation
part of #4089 

## Changes
- hare
  - bootstrap/fallback: whenever an update is received for epoch N, use it for epoch N's active set 
  - steady state: use the 1st hare output in epoch N as active set for epoch N (`ActiveSet_N`)
  - confidence param:
    - the bootstrap active set (targeting for epoch 2), is used btwn [`FirstLayer_2`, `FirstLayer_3`+`confidence_param`)
    - `ActiveSet_N` will be used btwn [`FirstLayer_N`+`confidence_param`, `FirstLayer_N+1`+`confidence_param`)

- beacon: 
  - whenever an update is received for an epoch, always use it as the beacon for the epoch

- add debug RPC to query the hare active set used by a node for a given epoch
- add a new systest TestFallback where bootstrapper will update continuously, and check the nodes use the updated value as beacon/activeset instead of the one calculated locally.
- add metrics for update listener for query and update outcome.
- add metrics for block certifier to count number of certificate generated/synced for each epoch.


hare/eligibility/oracle.go
```diff
@@ -388,167 +373,93 @@ func (o *Oracle) Proof(ctx context.Context, nonce types.VRFPostIndex, layer type
...
-		log.Uint32("confidence_param", o.cfg.ConfidenceParam),
-		log.Uint32("epoch_offset", o.cfg.EpochOffset),
-		log.Uint64("layers_per_epoch", uint64(o.layersPerEpoch)),
-		log.FieldNamed("effective_genesis", types.GetEffectiveGenesis()))
-
-	if value, exists := o.activesCache.Get(safeLayerStart.GetEpoch()); exists {
+
+	// note: for simplicity, bootstrap active set will be used btwn
+	// [types.EpochID(2).FirstLayer(), types.EpochID(3).Add(o.cfg.ConfidenceParam)-1]
+	// because the first bootstrap data targets epoch 2 (atxs were published in epoch 1)
```
countvonzero: @dshulyak 
i know this differ from our previous discussion. the reason being that each update published by spacemesh is designated with a target epoch (for beacon and activeset)
so for epoch 2, oracle will receive an update for epoch 2 as the bootstrap value and use it for the first `confidence_param` layers and cache this value as epoch 2 cache.

i didn't want to change the cached value after `confidence_param` has passed, as it's literally the same active set.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

hare/eligibility/oracle.go
```diff
@@ -236,7 +211,7 @@ func (o *Oracle) prepareEligibilityCheck(ctx context.Context, layer types.LayerI
 
 	// validate message
 	if !o.vrfVerifier.Verify(id, msg, vrfSig) {
-		logger.With().Info("eligibility: a node did not pass vrf signature verification",
+		logger.With().Warning("eligibility: a node did not pass vrf signature verification",
```
dshulyak: it should be debug

hare/eligibility/oracle.go
dshulyak: ok let me try to understand what makes sense, without our bootstrap it should work as follows:

confidence lets say 2 layers, with 5 layers epochs, and first epoch after genesis is 2

[1, 1, 2, 2, 2] - 2nd epoch
[2, 2, 3, 3, 3] - 3rd epoch
[3, 3, 4, 4, 4] - 4th epoch

so in normal workflow first two layers are using active set from previous epoch, and only in the first epoch they use values from bootstrap. it is consistent with that you wrote above, or no?

or you made it to work like this:
[1, 1, 1, 1, 1] - 2nd epoch
[1, 1]             - 3rd epoch

---
when we need to use fallback outside first epoch? i think for two reasons:
- no progress in previous epoch, such as

[1, 1, 0, 0, 0]
[0, 0, ?]

for ? - it would make sense to use fallback

- we see that nodes disagree about active set and we want to help them

so there are A

[1, 1, 2a, 2a, 2a]
[2a, 2a, 3a]

and B

[1, 1, 2b, 2b, 2b]
[2b, 2b, 3b]

but in this case we want 3a and 3b to be just 3, and for this we overwrite fallback. this is also i think is not supported

---

and the last case how we handle empty layers

[1, 1, 2, 2, 2]
[0, 0, ?]

i think it is ok to use fallback in this case as well, but confidence params needs to be larger

dshulyak: honestly speaking it is a crutch anyway, so if you prefer

[1, 1, 1, 1, 1] - 2nd epoch
[1, 1, 3, 3, 3] - 3rd epoch

instead of

[1, 1, 2, 2, 2] - 2nd epoch
[2, 2, 3, 3, 3] - 3rd epoch

it makes not significant difference. my version seems cleaner to me, i also think that this is what Tal meant, as he assumed that we need just one block.

--- 


[1, 1, 2a, 2a, 2a]
[2a, 2a, 3a]

[1, 1, 2b, 2b, 2b]
[2b, 2b, 3b]

this case i think is more important, as it means that you can't use download fallback automatically

countvonzero: at the beginning of the network, what you described would be (`bs`: bootstrap, `bk10`: block generated in layer 10)

| layer     | 10 | 11 | 12 | 13 | 14 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | bk10 | bk11 | bk12 | bk13 | bk14 |
| AS     | bs | bs | bk10 | bk10 | bk10 |

| layer     | 15 | 16 | 17 | 18 | 19 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | bk15 | bk16 | bk17 | bk18 | bk19 |
| AS     | bk10 | bk10 | bk15 | bk15 | bk15 |

whereas my approach would be

| layer     | 10 | 11 | 12 | 13 | 14 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | bk10 | bk11 | bk12 | bk13 | bk14 |
| AS     | bs | bs | bs | bs | bs |

| layer     | 15 | 16 | 17 | 18 | 19 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | bk15 | bk16 | bk17 | bk18 | bk19 |
| AS     | bs | bs | bk15 | bk15 | bk15 |

and because the `bk10` is a subset of `bs`, since the cache is keyed by epoch, once saved, i didn't bother updating the cache after confidence param passed. but not a big deal. i can change that.

---

- how we handle empty layers

let's say this is epoch 3. in my current implementation, it would be (`fb3`: fallback for epoch 3)
| layer     | 15 | 16 | 17 | 18 | 19 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | X | X | X | bk18 | bk19 |
| AS     | bk10 | bk10 | fb3 | fb3 | fb3 |

the question here is, should we use `bk18` for active set at layer 19? i.e.
| layer     | 15 | 16 | 17 | 18 | 19 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | X | X | X | bk18 | bk19 |
| AS     | bk10 | bk10 | fb3 | fb3 | bk18 |

i didn't bother switching (same reason as above). my thought is that there is already consensus. why switch to a different/weaker one.

- no progress in previous epoch, such as
> [1, 1, 0, 0, 0]
[0, 0, ?]

in my impl, i don't really care about failures in previous epoch. the only criteria is that by confidence layer, if there is no block generated "for the current epoch", the fallback value is used and saved in cache for subsequent layers.

| layer     | 15 | 16 | 17 | 18 | 19 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | X | X | X | X | X |
| AS     | bk10 | bk10 | fb3 | fb3 | fb3 |

| layer     | 20 | 21 | 22 | 23 | 24 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | bk20 | bk21 | bk22 | bk23 | bk24 |
| AS     | fb3 | fb3 | bk20 | bk20 | bk20 |

- we see that nodes disagree about active set and we want to help them

i have not thought of covering this case. here is my understanding of your description,
node A has
| layer     | 20 | 21 | 22 | 23 | 24 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | bk20A | bk21 | bk22 | bk23 | bk24 |
| AS     | bk15 | bk15 | bk20A | bk20A | bk20A |

and node B has
| layer     | 20 | 21 | 22 | 23 | 24 |
| ---      | :---: | :---: | :---: | :---: | :---: |
| block | bk20B | bk21 | bk22 | bk23 | bk24 |
| AS     | bk15 | bk15 | bk20B | bk20B | bk20B |

correct? 
so if this is network split, then we would have 2 set of hare protocol running in the network. Tal aslo stated something in the past here https://spacemesh.slack.com/archives/CG91F13C7/p1667953163180779

if not a network split, the contents of `bk20A` and `bk20B` should be the same (?) especially after the ATX grading is applied.

a similar problem arise here. say node B changed its mind and reverted to `bk20A` during layer 23. do you think node should switch to `bk20A` for active set in layer 24? i think yes.

my opinion is, once there is consensus, there is no need to change it to a different consensus.
so the last case (`bk20B`->`bk20A`) is the only case that i think worthy of changing.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

hare/eligibility/oracle.go
countvonzero: i think this discussion no longer applies.

dshulyak:
i thought that we want block in consensus (e.g tortoise consensus), ofcourse hare is a prerequisite for that. but there is a difference. if tortoise changes its opinion, what do we want to have as an active set?

if we want tortoise output, then implementation should be a bit different. it is not ok to pick first and cache it forever, instead it makes sense re-pick it if tortoise changed opinion. i also thought that we want to monitor actual empty layers in consensus output. so if it was non-empty and then became empty.. what should be done in such case?

blocks/generator.go
```diff
@@ -60,6 +63,29 @@ func defaultConfig() Config {
...
+	blockCount map[types.EpochID]int
+}
+
+func (s *stats) Add(block *types.Block) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	epoch := block.LayerIndex.GetEpoch()
+	s.blockCount[epoch]++
+	delete(s.blockCount, epoch-2)
+}
```
dshulyak: i think this metric might be useful for this component, but in general it has some limitations. node can download and apply block bypassing this generator and it should be good enough for fallback monitoring purposes

countvonzero:
> i thought that we want block in consensus (e.g tortoise consensus), ofcourse hare is a prerequisite for that. but there is a difference. if tortoise changes its opinion, what do we want to have as an active set?
> 
>
> if we want tortoise output, then implementation should be a bit different. it is not ok to pick first and cache it forever, instead it makes sense re-pick it if tortoise changed opinion. i also thought that we want to monitor actual empty layers in consensus output. so if it was non-empty and then became empty.. what should be done in such case?

@dshulyak the original discussion is to piggyback on hare consensus. the argument is that if there is a block certificate, that means the network agreed. it's not so much about which block is correct for that layer, but that everyone agreed that hare output that block at that time, since all we need is an active set nodes can agree on.

the code path for finding a tortoise verified block is a little more complicated and require consideration for revert as you stated.

api/grpcserver/debug_service.go
```diff
@@ -69,6 +72,20 @@ func (d DebugService) NetworkInfo(ctx context.Context, _ *empty.Empty) (*pb.Netw
 	return &pb.NetworkInfoResponse{Id: d.identity.ID().String()}, nil
 }
 
+// ActiveSet query provides hare active set for the specified epoch.
+func (d DebugService) ActiveSet(ctx context.Context, req *pb.ActiveSetRequest) (*pb.ActiveSetResponse, error) {
+	actives, err := d.oracle.ActiveSet(ctx, types.EpochID(req.Epoch))
+	if err != nil {
+		log.With().Error("failed to get active set", log.Uint32("epoch", req.Epoch), log.Err(err))
```
dshulyak: i prefer return details in status.Errorf and drop this

beacon/beacon.go
```diff
@@ -243,8 +243,13 @@ func (pd *ProtocolDriver) Start(ctx context.Context) {
 }
 
 func (pd *ProtocolDriver) UpdateBeacon(epoch types.EpochID, beacon types.Beacon) {
-	// TODO: implement
-	pd.logger.With().Info("received beacon update", epoch, beacon)
+	pd.mu.Lock()
+	defer pd.mu.Unlock()
+	if err := beacons.AddOverwrite(pd.cdb, epoch, beacon); err != nil {
```
dshulyak: maybe Set is a better name

beacon/beacon.go
```diff
@@ -243,8 +243,13 @@ func (pd *ProtocolDriver) Start(ctx context.Context) {
...
+		pd.logger.With().Error("failed to persist fallback beacon", epoch, beacon, log.Err(err))
+	}
```
dshulyak, line 250: this error handling is not sound. i suggest to return an error and fail the node if it is not possible to persist something (or retry, until it is possible to persist it)

it is definitely not an option just to continue

blocks/generator.go
```diff
@@ -60,6 +63,29 @@ func defaultConfig() Config {
 	}
 }
 
+type stats struct {
+	lock       sync.Mutex
```
dshulyak: mu is more common

blocks/generator.go
```diff
@@ -60,6 +63,29 @@ func defaultConfig() Config {
...
+	defer s.lock.Unlock()
+	epoch := block.LayerIndex.GetEpoch()
+	s.blockCount[epoch]++
+	delete(s.blockCount, epoch-2)
+}
+
+func (s *stats) Get() map[types.EpochID]int {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	result := make(map[types.EpochID]int)
```
dshulyak: it is better to implement such methods without allocating memory on heap

bootstrap/metrics.go
```diff
@@ -0,0 +1,33 @@
...
+	failure = "fail"
+)
+
+var (
+	queryCount = metrics.NewCounter(
+		"query",
+		namespace,
+		"number of queries made to the update site",
+		[]string{"outcome"},
+	)
```
dshulyak, line 24: maybe track latency

sql/certificates/certs.go
```diff
@@ -52,6 +52,28 @@ func GetHareOutput(db sql.Executor, lid types.LayerID) (types.BlockID, error) {
...
 }
 
+func FirstInEpoch(db sql.Executor, epoch types.EpochID) (types.BlockID, error) {
+	var (
+		result types.BlockID
+		err    error
+		rows   int
+	)
+	if rows, err = db.Exec(`
+		select block from certificates where layer between ?1 and ?2 and valid = 1
```
dshulyak: do we keep track of invalid certificates? thats a waste

dshulyak (approved):
i guess it makes sense if it is not going to be a long term solution

blocks/generator.go
countvonzero: i've updated the metrics to track block certificate instead. after all that's what we are looking for.

blocks/generator.go
countvonzero: thanks. noted. moved the code to certifier.go and reuse the mutex there.

sql/certificates/certs.go
countvonzero: we don't keep track of those that were invalid in the first place.
we only keep certificate that was valid before but later became invalid.
1. hare output but has no certificate to back it up (this db entry will have certificate as null)
2. a certificate that was valid when received but no longer valid when the 2nd valid certificate is received.

beacon/beacon.go
countvonzero: changed to fail the node.

countvonzero:
bors try

countvonzero:
bors merge

PR #4244, opened 2023-04-03
[Merged by Bors] - vm: drop additional blake3 hash before signing

### LINKED ISSUE #4234: vm: remove blake3 hash before calling ed25519.Sign for transactions
this was in original design for vm. need to clarify if there was a specific reason for that, but at that time it was said that this is for hardware wallets.

we want to push decodable transaction into hw wallet, so that it can be printed nicely for user to sign. in this case hashing before signing is not an option. we know for sure that precompiles that we have will fit into whatever hw limits are imposed on tx size, if it turns out later that some of the templates won't fit - Verify method will have to be adjusted specifically for that template.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4234

from ledger docs (references in the gh issue)

> The length of the data part is 255 bytes and the whole message sent in a single APDU is 260 bytes with the blocks cla, ins, p1, p2 and one extra byte that encodes the length of the data that follows.

> Receive: the transaction is sent through one or more APDUs. If this transaction is too large to be sent at once, it is copied chunk by chunk into a global buffer.

### DIFF
--- a/genvm/core/hash.go
+++ b/genvm/core/hash.go
@@ -7,8 +7,12 @@ import (
 	"github.com/spacemeshos/go-spacemesh/hash"
 )
 
-// Hash bytes into blake3 hash.
-var Hash = hash.Sum
+func SigningBody(genesis, tx []byte) []byte {
+	full := make([]byte, 0, len(genesis)+len(tx))
+	full = append(full, genesis...)
+	full = append(full, tx...)
+	return full
+}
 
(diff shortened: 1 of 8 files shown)


dshulyak:
@brusherru could you please make a similar change in smapp? it simply removes hashing before pushing message for signing. then we coordinate merge before some release :)

countvonzero (approved):

dshulyak:
bors merge

Issue #4240, opened 2023-04-02
PoS initialization is not working properly while started from GRPC API


brusherru:
## Description
We have two ways to start/continue PoS initialization:
1. Just run go-spacemesh with the node config containing valid `smeshing` opts. (Let's call it "Way A")
2. Call GRPC API `SmesherService.StartSmeshing` ("Way B")

Besides the initialization, GRPC API also makes it possible to set PoS initialization on pause (calling `StopSmeshing` with `deleteFiles: false`) and then continue creating PoS data (call `StartSmeshing` with the same opts).

However, the current go-spacemesh handles `StartSmeshing` incorrectly and fails with an error.
It produces the same error for both cases: when the User initializes PoS the first time or asks to continue initialization.

### Steps to reproduce
1. Run go-spacemesh
2. Call `SmesherService.StartSmeshing` with valid opts. E.g.
    ```
    {"coinbase":{"address":"stest1qqqqqqyfnc665k9xamxtyd2mmnv8tpwu7tl77gcjyncck"},"opts":{"dataDir":"/Users/my-user/pos-data","numUnits":2,"maxFileSize":2147483648,"computeProviderId":1,"throttle":false}}
    ```
### Expected behavior
It starts creating PoS data.
### Actual behavior
It ends up with error. Here are logs:
```
2023-04-02T11:50:59.521-0600	INFO	grpc	finished unary call with code OK	{"module": "grpc", "grpc.start_time": "2023-04-02T11:50:59-06:00", "system": "grpc", "span.kind": "server", "grpc.service": "spacemesh.v1.SmesherService", "grpc.method": "StartSmeshing", "peer.address": "[::1]:51995", "grpc.code": "OK", "grpc.time_ms": 0.072}
2023-04-02T11:50:59.523-0600	INFO	6895c.post	post setup session starting	{"node_id": "6895c819dbc728e26a6cce2185d9386249da5352d321afd782072ac1e71d0c2b", "module": "post", "node_id": "6895c819dbc728e26a6cce2185d9386249da5352d321afd782072ac1e71d0c2b", "commitment_atx": "b8f8d33012", "data_dir": "/Users/my-user/pos-data", "num_units": "2", "labels_per_unit": "1310720", "provider": "1", "name": "post"}
[...]

PR #4245, opened 2023-04-03
[Merged by Bors] - Fixes for #4239 and #4240

### LINKED ISSUE #4239: Make consistent changes due to removing BitsPerLabel
## Description
Previously we get rid of `BitsPerLabel` from the config file and started using `128` as a hard-coded value. See [related changes](https://github.com/spacemeshos/go-spacemesh/commit/d94c2467d880dfc825562b71bb97d554f21597e5#diff-c4c73cc85f7eb81a6ec4579c663fe8ff002ed6009621a6f61e131a312ca90738L28)

1. Clients that count on this value (for example, Smapp uses it to calculate PoS size) — broken.
2. [Proto files](https://github.com/spacemeshos/api/blob/master/spacemesh/v1/smesher_types.proto#L89-L96) were not updated (and as a consequence client, which uses proto files, get falsy `byte_per_label: 0`)

We can inject such hard-coded values into clients and change proto files.
But actually, I'd prefer to keep it in API and not insist on "infecting" clients with some magic constants. ;)
Anyway, to move on with Smapp, I've already added falling back to the magic `128` ;)

## Environment
 - Node Version: v0.2.26-beta.0

poszu:
Closes #4239, #4240

### DIFF
--- a/api/grpcserver/smesher_service.go
+++ b/api/grpcserver/smesher_service.go
@@ -233,6 +233,7 @@ func (s SmesherService) PostConfig(context.Context, *empty.Empty) (*pb.PostConfi
 	cfg := s.postSetupProvider.Config()
 
 	return &pb.PostConfigResponse{
+		BitsPerLabel:  config.BitsPerLabel,
 		LabelsPerUnit: cfg.LabelsPerUnit,
 		MinNumUnits:   cfg.MinNumUnits,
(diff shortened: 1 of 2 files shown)


dshulyak (approved):

poszu:
bors merge

PR #4249, opened 2023-04-03
[Merged by Bors] - Make LayerID an uint32

fasmat:
## Motivation
Similar to EpochID, LayerID can just be a `uint32` for simpler handling in the code and more efficient scale encoding.

## Changes
Change LayerID from a `struct` to a `uint32`

## Test Plan
- existing tests pass

## TODO

## DevOps Notes


common/types/layer.go
```diff
@@ -41,52 +43,64 @@ func GetLayersPerEpoch() uint32 {
 
 // GetEffectiveGenesis returns when actual proposals would be created.
 func GetEffectiveGenesis() LayerID {
-	return NewLayerID(atomic.LoadUint32(&effectiveGenesis))
+	return LayerID(atomic.LoadUint32(&effectiveGenesis))
 }
 
 // NewLayerID creates LayerID from uint32.
 func NewLayerID(value uint32) LayerID {
```
countvonzero: is this still needed?

common/types/layer.go
```diff
@@ -41,52 +43,64 @@ func GetLayersPerEpoch() uint32 {
...
 }
 
-//go:generate scalegen -types LayerID
-
 // LayerID is representing a layer number. Zero value is safe to use, and means 0.
 // Internally it is a simple wrapper over uint32 and should be considered immutable
 // the same way as any integer.
-type LayerID struct {
-	// NOTE(dshulyak) it is made public for compatibility with encoding library.
-	// Don't modify it directly, as it will likely to be made private in the future.
```
countvonzero, line 59: not sure the change and this note is compatible. @dshulyak can you comment?

fasmat: `LayerID` was a `struct` instead of a `uint32` only because scalegen cannot generate a `EncodeScale` and `DecodeScale` for a type that isn't a `struct`.

The two methods are now hand written (just like for `ATXID`, `EpochID` and some oder types where generation isn't possible). The "don't modify/access directly" comment vas violated all over the place btw. 😅

common/types/layer.go
fasmat: Technically no, I can remove it and replace it with `types.LayerID()` instead.

common/types/layer.go
dshulyak: i think that change was misguided, i wouldn't do it now. originally people were having problems with overflow on layer id in tests.. but anyway, i think this is much better

dshulyak (approved):

common/types/layer.go
fasmat: Removed

fasmat:
bors merge

PR #4256, opened 2023-04-05
[Merged by Bors] - set eligibility integer on reference ballot

dshulyak:
i didn't notice that it wasn't part of the epoch data. integers can't overwrite eligibility count in the same epoch, so it should be specified once

### DIFF
--- a/proposals/eligibility_validator.go
+++ b/proposals/eligibility_validator.go
@@ -115,6 +115,7 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot) 
 	if len(activeSets) == 0 {
 		return false, fmt.Errorf("%w: ref ballot %v", errEmptyActiveSet, refBallot.ID())
 	}
+	advertisedCount := refBallot.EpochData.EligibilityCount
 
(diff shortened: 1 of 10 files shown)


proposals/eligibility_validator.go
```diff
@@ -145,8 +146,8 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot)
 	if err != nil {
 		return false, err
 	}
-	if ballot.EligibilityCount != numEligibleSlots {
-		return false, fmt.Errorf("%w: expected %v, got: %v", errIncorrectEligCount, numEligibleSlots, ballot.EligibilityCount)
+	if advertisedCount != numEligibleSlots {
+		return false, fmt.Errorf("%w: expected %v, got: %v", errIncorrectEligCount, numEligibleSlots, advertisedCount)
```
countvonzero: in that case it seems we just need to verify it once with ref ballot and not the following ballots.

countvonzero (approved):
thanks for fixing.

dshulyak:
bors merge

PR #4257, opened 2023-04-05
[Merged by Bors] - Remove Metadata from ActivationTx and Ballot

fasmat:
## Motivation
Activation Transactions contain redundant data that doesn't need to be broadcasted but is required for signature creation & malfeasance proofs.

## Changes
- Remove `ATXMetadata` from `ActvationTx`
  - It is still used to create the msg that is signed for ATX signatures and is part of a Malfeasance proof
- Renamed to `ATXSigMsg` with the fields `ID` and `PublishEpoch`

## Test Plan
- existing tests pass

## TODO

## DevOps Notes


activation/activation.go
```diff
@@ -669,5 +669,5 @@ func (b *Builder) GetPositioningAtxInfo() (types.ATXID, types.LayerID, error) {
 func SignAndFinalizeAtx(signer *signing.EdSigner, atx *types.ActivationTx) error {
 	atx.Signature = signer.Sign(signing.ATX, atx.SignedBytes())
 	atx.SmesherID = signer.NodeID()
-	return atx.Initialize()
+	return nil
```
countvonzero: why? are we sure there is no downstream consumer of this atx?

common/types/activation.go
```diff
@@ -188,16 +174,30 @@ func NewActivationTx(
 	return atx
 }
 
-// SetMetadata sets ATXMetadata.
-func (atx *ActivationTx) SetMetadata() {
-	atx.Target = atx.TargetEpoch()
-	atx.MsgHash = BytesToHash(atx.HashInnerBytes())
+// ATXSigMsg is the data of ActivationTx that is signed.
+type ATXSigMsg struct {
```
countvonzero: as discussed, lets leave ATXMetadata as is.
- use the original name (decided as common term for malfeasance in research)
- use Hash32 instead of ATXID to decouple ID calculation and signed data.

common/types/activation.go
```diff
@@ -188,16 +174,30 @@ func NewActivationTx(
...
+	ID           ATXID
+	PublishEpoch EpochID
```
countvonzero: the epoch ID needs to be the first field for protocol update purpose

malfeasance/handler.go
```diff
@@ -235,17 +235,17 @@ func (h *Handler) validateMultipleATXs(logger log.Log, proof *types.MalfeasanceP
...
 	}
 	logger.With().Warning("received invalid atx malfeasance proof",
-		log.Stringer("smesher", ap.Messages[0].SmesherID),
-		log.Stringer("smesher", ap.Messages[1].SmesherID),
-		log.Object("first", &ap.Messages[0].InnerMsg),
-		log.Object("second", &ap.Messages[1].InnerMsg),
+		log.Stringer("first_smesher", ap.Messages[0].SmesherID),
+		log.Object("first_proof", &ap.Messages[0].InnerMsg),
+		log.Stringer("second_smesher", ap.Messages[1].SmesherID),
+		log.Object("second_proof", &ap.Messages[1].InnerMsg),
```
countvonzero, line 248: why does the order need to change? and why only change atx and not other malfeasance proof?
there is no practical reason to change this. please revert.

countvonzero (commented):
as discussed offline, also include similar change for Ballot?
separate PR would be fine as well.

malfeasance/handler.go
fasmat: I had a different in-between version here. The original code logs 2 fields called "smesher" which I don't think is valid json. Either the second value will just not be logged or the output will look like:

```json
{
    "smesher": "3fd5a",
    "smesher": "3fd5a",
    "first": { ... },
    "second": { ... },
}
```

which a lot of parsers will struggle with.

common/types/activation.go
dshulyak: we need to write down somewhere how we are going to support update for every type. i thought that version for malf proofs is decoded from the outermost layer id

we made malf proof to be `layer_id, malf_proof_enum`. if we need upgrades, we decode layer id, determine version of the enum and then decode the rest

fasmat: @dshulyak is only for the metadata or also for the ATX itself?

I will update the `struct` to look like this:

```go
type ATXMetadata struct {
	Layer LayerID
	Hash  Hash32
}
```

The epoch can be derived from the `LayerID` when checking for malfeasance, and this way it is generic enough to be used for all 3 types of proof we have at the moment.

fasmat: @countvonzero for protocol update purposes can we use to the `LayerID` in the `MalfeasanceProof` type? https://github.com/spacemeshos/go-spacemesh/blob/remove-metadata-from-atx/common/types/malfeasance.go#L22

This way we can do upgrades without having to limit ourselves to a specific layout of the `XXXMetadata` structs.

dshulyak: i think that atx metadata should continue using epoch and order doesn't matter. as a rule we should use only the first field of the object (in this case MalfeasanceProof) for the version, everything else after that field is decoded based on the version 

```
layer := decodeLayer(buf)
switch layerToVersion(layer) {
     case 0:
     case 1:       
}
```

common/types/activation.go
dshulyak: > decided as common term for malfeasance in research

can we please stop doing this "decided in research" thing. if it was decided please specify it somewhere, and lets figure out if it really the best option

dshulyak (approved):

activation/activation.go
fasmat: reverted

fasmat:
The `types.HareMetadata` could theoretically also be removed from a `hare.Message`, since the `MsgHash` contained is also data that doesn't need to be broadcasted.

That seems to be a bigger change and I would like to tackle this in a separate PR.

malfeasance/handler.go
countvonzero: great. can you change it also for other types, ballot and hare please

countvonzero (approved):

fasmat:
bors merge

fasmat:
bors merge

PR #4259, opened 2023-04-05
[Merged by Bors] - Move checking if poet round is in-progress to NiPoST builder

### LINKED ISSUE #4246: Node won't continue waiting for PoET proofs after being restarted
## Description
Node won't continue an already started nipost building procedure when it was restarted when the poet round is already executing.

## Steps to reproduce
1. Run node and wait for it to submit a challenge to poet.
2. Wait for the poet round to start executing.
3. Restart the node

## Actual Behavior
`ATX Builder` detects that poet round was missed (the round is executing) and doesn't continue the in-progress nipost building 

## Expected Behavior
`ATX Builder` should continue building nipost (skip submitting poet challenges and wait for proofs).

## Environment
env-independent

poszu:
Closes #4246

## Changes
Moved check if the poet round able to produce a proof "on time" to NiPoST builder. It is now checked **only if** challenge has not been submitted yet.


poszu:
bors try

dshulyak:
we do write to disk somewhere that challenge was submitted?

activation/nipost.go
```diff
@@ -158,22 +158,27 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
 		if len(poetRequests) == 0 {
-			return nil, 0, &PoetSvcUnstableError{msg: "failed to submit challenge to any PoET"}
+			return nil, 0, &PoetSvcUnstableError{msg: "failed to submit challenge to any PoET", source: ctx.Err()}
 		}
+
 		nipost.Challenge = &challengeHash
 		nb.state.Challenge = challengeHash
 		nb.state.PoetRequests = poetRequests
 		nb.persist()
+		if err := ctx.Err(); err != nil {
```
countvonzero, line 179: 1. i am confused about moving this check from above
2. why checking `ctx` and not `submitCtx`
3. why doesn't `submitPoetChallenges` return error if submitCtx time out.

activation/nipost.go
```diff
@@ -158,22 +158,27 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
-
 		if len(poetRequests) == 0 {
-			return nil, 0, &PoetSvcUnstableError{msg: "failed to submit challenge to any PoET"}
+			return nil, 0, &PoetSvcUnstableError{msg: "failed to submit challenge to any PoET", source: ctx.Err()}
 		}
+
 		nipost.Challenge = &challengeHash
 		nb.state.Challenge = challengeHash
 		nb.state.PoetRequests = poetRequests
 		nb.persist()
```
countvonzero, line 178: does this answer @dshulyak's question [here](https://github.com/spacemeshos/go-spacemesh/pull/4259#issuecomment-1497766960) ?

countvonzero (approved):

activation/nipost.go
dshulyak: thanks, i wasn't sure that they are persisted with a challenge

activation/nipost.go
poszu: Ad 1. The check is moved under `nb.persist()` to save builder state even if it was interrupted to record all the challenges submitted.
Ad 2. Because submitCtx being cancelled (via a deadline) isn't a reason to quit. It only means that submitting to _some_ poets timed out. We only want to quit here if the main ctx is canceled.
Ad 3. It could, but a timeout in submit is not a reason to quit if at least one challenge was submitted successfully. In this case we want to continue.

In general, the process is a kind of best effort:
- we tolerate non-fatal errors if it makes sense to continue,
- quit early if interrupted via `ctx` cancellation.

poszu:
bors merge

poszu:
bors merge

PR #4260, opened 2023-04-05
Simplify poet proof constr in tests

poszu:
## Motivation
`Proof.Members` can be initialized in-place from the challenge array.

## Changes
Avoid calling `copy()` to fill members, but construct it in place.


dshulyak (approved):

countvonzero:
bors merge

Issue #4248, opened 2023-04-03
Consider extracting message processing from hare.consensusProcess


piersy:
## Description
Currently it's not possible to write reliable tests of hare or consensus processes because we don't have a way to hook into message processing to understand when it is safe to progress a round.

We currently hook into the pubsub system to understand message delivery, but message processing in hare is done in a different go routine so we're unable to know when a message delivered over pubsub is actually processed. Our current approach is to set some timeout such that we assume that messages delivered over pubsub will have been processed by hare after the timeout expires.

This is unreliable and this approach will fail randomly in response to changes in the execution environment.

This issue appears in commit hash: TODO

Related files (optionally with line numbers): `hare/flows_test.go`

PR #4261, opened 2023-04-05
Fix multiple cps tests

### LINKED ISSUE #4001: Flaky Test_multipleCPsAndIterations UT

piersy:
## Motivation
Closes #4001

~~Additionally with these flaky tests fixed we can remove the restriction to run our tests on a single core `-p 1` which will speed up CI.~~ It turns out there are other tests that seem to fail with that restriction removed #4262, so this will have to wait.

(Note: I'm OOO till Tues 11th Apr, so likely will not respond till then)

## Changes
Fixed 2 race conditions in `hare/broker.go` where messages could be lost. (for detail see https://github.com/spacemeshos/go-spacemesh/commit/1b6ad3108f8cc16973818062f8fdd70fd1f52a5d).

Re-instated the shared round clock for use in the `Test_multipleCPs...` tests. The shared round clock allows us to enforce synchronisation between participating nodes resulting in more reliable tests, also since we do not rely on timeouts the tests can be made much faster.

The old shared round clock would progress rounds when all nodes had sent a message for a round, this was problematic because we needed to guess at how long we should wait (`processingDelay`) for all the messages to be propagated before actually advancing the round.  I changed this to progress when all node's have received all messages for a round, this mitigates the problem but doesn't actually solve it because there is still a delay between a node receiving a message and `hare.Hare` actually processing it. I raised #4248 with regard to really fixing this.

It looks like `Test_multipleCPsAndIterations` had been broken for some time, the stalled layer was set to be 1 but the tests started at layer 8, so the stalled layer was never encountered. Fixing this revealed the need to have an additional mechanism to drive rounds forward because at the stalled layer messages are dropped and thus the normal mechanism of progressing rounds when nodes have received all messages no longer worked. Extended the shared round clock to allow for this.

Added some go-routine cleanup to avoid panicking due to logging occurring in go-routines after test completion. Also modified `hare.Hare.Close()` to close the broker and the `blockGenCh`.

Increased `Config.ExpectedLeaders` to be `N/2 + 1` where previously it was set to `N/2` which seemed to allow nodes to be able to finish the preround with too few proposals to be able to validate any proposal in the statusround.

Limited the new layer initiation such that a new layer can only be initiated if all nodes have started processing the previous layer, this is needed because the broker only stores early messages for the subsequent layer, so a node that is 2 layers ahead of other nodes will have its messages ignored by their brokers.

~~Removed the restriction to run tests on a single core `-p 1`.~~ (Not quite there yet)

## Test Plan
Ran the `Test_multipleCPs...` tests on CI hundreds of times without failures.

## DevOps Notes


piersy:
Closing for now as this is very out of date and needs some significant effort to resolve conflicts with develop.

PR #4263, opened 2023-04-05
[Merged by Bors] - Use canonical ED for POET

### LINKED ISSUE #4139: crypto: use canonical ed25519 signature in all domains except HARE

fasmat:
## Motivation
Closes #4139, merge after https://github.com/spacemeshos/poet/pull/282

## Changes
switch signatures for PoET to canonical ed25519

## Test Plan
existing tests pass

## TODO

## DevOps Notes


go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	github.com/spacemeshos/fixed v0.0.0-20210523192743-8d17e03c169a
 	github.com/spacemeshos/go-scale v1.1.7
 	github.com/spacemeshos/merkle-tree v0.2.1
-	github.com/spacemeshos/poet v0.7.0
+	github.com/spacemeshos/poet v0.7.1-0.20230405184528-298199ded0e6
```
fasmat: TODO(mafa): release new poet version and update

@poszu where do I update the used poet version in systests?

countvonzero: systest/Makefile

countvonzero: can you refer to the commits in this poet version update for me to understand? thanks

fasmat: The open PR is here: https://github.com/spacemeshos/poet/pull/282/ I just swapped the used ed25519 libraries, since @poszu already considered the upcoming library change when he implemented it.

As soon as it is approved I merge it and release a new poet version that I'll reference in the `systest/Makefile`.

dshulyak (approved):

fasmat:
bors merge

PR #4264, opened 2023-04-06
[Merged by Bors] - malfeasence proofs: use consistent layer for gossiped object

dshulyak:
there were two problems with using clock values for layer id in gossip messages:
- gossip identifies message by hash, if multiple nodes produce the same message with different layer id, multiple copies will be brodcasted over the network. so the valid message structure should be consistent across different nodes
- using clock-based value is not a reliable version, as it will change based on when original objects were received


fasmat (approved):

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #4267, opened 2023-04-06
[Merged by Bors] - Remove Metadata from hare messages

fasmat:
## Motivation
Follow-up to https://github.com/spacemeshos/go-spacemesh/pull/4257 this now removes the metadata from hare messages

## Changes
- The metadata struct was removed from hare messages, which saves some data for every message.
- The check if the extracted NodeID is in the active is done now earlier to fail quicker and cheaper if its not.

## Test Plan
Existing tests pass.

## TODO

## DevOps Notes


fasmat:
bors try

hare/messagevalidation_test.go
```diff
@@ -65,17 +65,6 @@ func TestMessageValidator_ValidateCertificate(t *testing.T) {
...
-	msgs = make([]Message, 0, sv.threshold)
-	for i := 0; i < sv.threshold; i++ {
-		signer, err := signing.NewEdSigner()
-		require.NoError(t, err)
-		// wrong set value
-		m := BuildCommitMsg(signer, NewDefaultEmptySet()).Message
-		msgs = append(msgs, m)
-	}
-	cert.AggMsgs.Messages = msgs
-	require.False(t, sv.validateCertificate(context.Background(), cert))
```
fasmat, line 77: In the unchanged code this part of the test passes because `func (v *syntaxContextValidator) validateAggregatedMessage` returns `false` due to the hash in the Message not matching the actual hash of the message, not because the set value is wrong.

Since this check isn't done any more the `sv.validateCertificate` returns `true` here.

I think the previous intention of the test was to verify that invalid sets fail validation, but according to another test an empty set isn't invalid?

fasmat:
bors try

hare/messagevalidation_test.go
fasmat: As discussed with @countvonzero the actual check here was to see if the message has been altered after signing.

With the change the NodeID extracted from the signature will differ from the ID of the smesher that signed the message.

This is detected in the eligibility oracle where the check for the NodeID being in the active set is now done first so that the check fails faster if this happens.

hare/committracker.go
```diff
@@ -59,35 +59,45 @@ func (ct *commitTracker) OnCommit(ctx context.Context, msg *Msg) {
...
+			return
+		}
+
+		ct.logger.WithContext(ctx).With().Warning("equivocation detected in commit round",
+			log.Stringer("smesher", msg.NodeID),
+			log.Object("prev", &prev.InnerMsg),
+			log.Object("curr", &metadata),
+		)
+		ct.eTracker.Track(msg.NodeID, msg.Round, msg.Eligibility.Count, false)
+		this := &types.HareProofMsg{
```
countvonzero, line 78: we should add NodeID to HareProofMsg as well for future compatibility.
if we move to hare 3.0 where we dont need to do pubkey extraction (msg size no longer critical)
then we don't need to keep the custom lib around forever to verify old malfeasance proof.

hare/committracker.go
```diff
@@ -59,35 +59,45 @@ func (ct *commitTracker) OnCommit(ctx context.Context, msg *Msg) {
 		return
 	}
 
+	msgHash := types.BytesToHash(msg.Message.HashBytes())
```
countvonzero: nit: inline in metadata and use metadata.msgHash below

hare/committracker.go
```diff
@@ -59,35 +59,45 @@ func (ct *commitTracker) OnCommit(ctx context.Context, msg *Msg) {
...
+	metadata := types.HareMetadata{
+		Layer:   msg.Layer,
+		Round:   msg.Round,
+		MsgHash: msgHash,
+	}
 	if prev, ok := ct.seenSenders[msg.NodeID]; ok {
-		if prev.InnerMsg.Layer == msg.Layer &&
-			prev.InnerMsg.Round == msg.Round &&
-			prev.InnerMsg.MsgHash != msg.MsgHash {
-			ct.logger.WithContext(ctx).With().Warning("equivocation detected in commit round",
```
countvonzero, line 66: i honestly feel that this condition capture the essence of malfeasance behavior in a compact form:
the same signer signs two messages for the same layer/round

hare/committracker.go
```diff
@@ -59,35 +59,45 @@ func (ct *commitTracker) OnCommit(ctx context.Context, msg *Msg) {
...
+		if prev.InnerMsg.Layer != msg.Layer || prev.InnerMsg.Round != msg.Round {
```
countvonzero: this check makes me think that mismatching layer / round is expected. but it's actually not.
mismatching layer here would be a programming error.
mismatching round here should not even be processed.
the old code already have this trouble.

hare/eligibility/oracle.go
```diff
@@ -282,14 +283,14 @@ func (o *Oracle) prepareEligibilityCheck(ctx context.Context, layer types.LayerI
 		totalWeight *= uint64(committeeSize)
 		n *= committeeSize
 	}
-	p = fixed.DivUint64(uint64(committeeSize), totalWeight)
+	p := fixed.DivUint64(uint64(committeeSize), totalWeight)
 
-	if minerWeight > maxSupportedN {
+	if n > maxSupportedN {
```
countvonzero, line 289: @noamnelke can you help review this part of code please? i want to make sure the math is correct.

hare/committracker.go
fasmat: Adding the `NodeID` to the `HareProofMsg` doesn't solve that issue. Signatures created with spacemesh ed25519 are not compatible with signatures created with canonical `ed25519`. We will have to keep the custom lib either way...

hare/eligibility/oracle.go
fasmat: @noamnelke specifically we think that some of the casts and overflow checks (see comments) might be problematic.

hare/committracker.go
fasmat: I don't like the verbosity of this check, it is also used in multiple places in the same way. I created a method on `HareMetadata` `Equivocation` that contains this check so instead of:

```golang
if prev.InnerMsg.Layer == msg.Layer &&
	prev.InnerMsg.Round == msg.Round &&
	prev.InnerMsg.MsgHash != msg.MsgHash {
```

one can write

```golang
if prev.InnerMsg.Equivocation(&metadata) {
```

instead. This is also allows for different equivocation checks (if we change them at some point) to be neatly contained within this method.

hare/committracker.go
fasmat: I hope you also find the `Equivocation` method a cleaner approach to detecting two messages of different content in the same layer 🙂

dshulyak (approved):

fasmat:
bors merge

hare/eligibility/oracle.go
fasmat: Created a new PR to specifically address these issues: https://github.com/spacemeshos/go-spacemesh/pull/4275

PR #4269, opened 2023-04-07
[Merged by Bors] - cmd/node: write smesher file in a hex

dshulyak:
it was written in binary, which is not very convenient to work with

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -3,6 +3,7 @@ package node
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


cmd/node/node.go
```diff
@@ -952,16 +953,25 @@ func (app *App) LoadOrCreateEdSigner() (*signing.EdSigner, error) {
...
 			return nil, fmt.Errorf("failed to write identity file: %w", err)
 		}
 
 		log.With().Info("created new identity", edSgn.PublicKey())
 		return edSgn, nil
 	}
+	dst := make([]byte, signing.PrivateKeySize)
+	n, err := hex.Decode(dst, data)
+	if err != nil {
+		return nil, fmt.Errorf("bad hex: %w", err)
```
poszu: The error could be more descriptive, like
```suggestion
		return nil, fmt.Errorf("decoding private key: %w", err)
```

That would end up as: `"decoding private key: encoding/hex: invalid byte"`, which is IMO more helpful in figuring out what is wrong.

poszu (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #4272, opened 2023-04-11
[Merged by Bors] - vm: add methods for base/fixed gas estimation that has access to immutable state

### LINKED ISSUE #4271: vm: use polynomial-based gas estimation in mutlsig templates
full problem is described in https://community.spacemesh.io/t/gas-cost-for-precompiles/348/7#fixed-gas-inconsistency-5

> we are doing 3 different multisig (and vesting) templates (1/k, 2/k, 3/k). the assumption there was that parameter k will not change gas cost. this assumption was clearly wrong, creating (and loading) template with 3/10 (where k=10) and 3/5 should have different cost. as a stop gap solution we can always charge by the upper bound (k=10), but it doesn’t look nice.

proposed solution is to add a method to template that will estimated base gas/fixed gas based on immutable number of variables. for multisig those are n (number of signatures) and k (number of public keys in multisig).

additionally for spawn, fixed gas should include the cost of spawning target account, not the cost of spawning principal account (https://community.spacemesh.io/t/gas-cost-for-precompiles/348/7#selfspawn-vs-spawn-gas-4). practically it means using different method for principal and target templates.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4271

### DIFF
--- a/genvm/cmd/vesting/main.go
+++ b/genvm/cmd/vesting/main.go
@@ -69,25 +69,13 @@ func decodeKeys(dir string) []core.PublicKey {
 	return keys
 }
 
-func getTemplate(k int) core.Address {
-	switch k {
-	case 1:
-		return vesting.TemplateAddress1
-	case 2:
-		return vesting.TemplateAddress2
-	case 3:
-		return vesting.TemplateAddress3
-	}
-	must(fmt.Errorf("no support for %d signature-vesting account", k))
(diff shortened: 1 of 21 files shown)


countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #4275, opened 2023-04-11
[Merged by Bors] - Eligibility oracle: Fix incorrect checks for overflows

fasmat:
## Motivation
Some casts in the hare eligibility oracle don't properly check for overflows.

## Changes
- `minerWeight` larger than `maxSupportedN`: instead of casting to check for an overflow, check first if the value is already too big
- `committeeSize > int(totalWeight)` this might overflow, if `totalWeight` is larger than `math.MaxInt32`. Instead cast `committeeSize` to `uint64`.
- Check for overflows before doing the next calculation, to abort earlier if necessary.

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -351,12 +351,13 @@ func (o *Oracle) CalcEligibility(
 
 	for x := 0; x < n; x++ {
 		if fixed.BinCDF(n, p, x).GreaterThan(vrfFrac) {
-			// even with large N and large P x will be << 2^16, so this cast is safe
+			// even with large N and large P, x will be << 2^16, so this cast is safe
 			return uint16(x), nil
 		}
 	}
 
-	// since BinCDF(n, p, n) is 1 for any p this code is only reached if n << 2^16
+	// BinCDF(n, p, n) is 1 for any p, this code can only be reached if n much smaller
+	// than 2^16 (so that BinCDF(n, p, n-1) is still lower than vrfFrac)
 	return uint16(n), nil
 }
 
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -43,8 +43,8 @@ const (
 )
 
 const (
-	activesCacheSize = 5                 // we don't expect to handle more than two layers concurrently
-	maxSupportedN    = math.MaxInt32 / 2 // higher values result in an overflow
+	activesCacheSize = 5                       // we don't expect to handle more than two layers concurrently
+	maxSupportedN    = (math.MaxInt32 / 2) + 1 // higher values result in an overflow when calculating CDF
 )
 
 var (
--- a/hare/eligibility/oracle.go
+++ b/hare/eligibility/oracle.go
@@ -43,9 +43,8 @@ const (
 )
 
 const (
-	activesCacheSize = 5          // we don't expect to handle more than two layers concurrently
-	maxSupportedN    = 1073741824 // higher values result in an overflow
-	// TODO(mafa): why (MaxUint32 + 1)/4?
+	activesCacheSize = 5                       // we don't expect to handle more than two layers concurrently
+	maxSupportedN    = (math.MaxInt32 + 1) / 2 // higher values result in an overflow
 )
 
 var (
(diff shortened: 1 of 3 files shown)


hare/eligibility/oracle.go
```diff
@@ -40,9 +40,8 @@ const (
 )
 
 const (
-	activesCacheSize = 5          // we don't expect to handle more than two layers concurrently
-	maxSupportedN    = 1073741824 // higher values result in an overflow
-	// TODO(mafa): why (MaxUint32 + 1)/4?
+	activesCacheSize = 5                       // we don't expect to handle more than two layers concurrently
+	maxSupportedN    = (math.MaxInt32 + 1) / 2 // higher values result in an overflow
```
fasmat: Why this value and not `math.MaxInt32`?

hare/eligibility/oracle.go
```diff
@@ -270,26 +269,26 @@ func (o *Oracle) prepareEligibilityCheck(ctx context.Context, layer types.LayerI
 	)
 
 	// ensure miner weight fits in int
-	n := int(minerWeight)
-	if uint64(n) != minerWeight { // TODO(mafa): why not [if minerWeight > maxSupportedN]? leave n uint64 and cast to int (and check for overflow) before returning?
+	n := minerWeight
+	if n > maxSupportedN {
 		logger.Fatal(fmt.Sprintf("minerWeight overflows int (%d)", minerWeight))
```
fasmat: why are we exiting here but not further below when we are checking for an overflow of `n` again?

hare/eligibility/oracle.go
noamnelke: I don't remember. Comment says that higher values result in an overflow, but I don't remember where the overflow would happen. You can try to follow the code, or possibly just try to stick a larger number there and run the code (if it doesn't result in an overflow in your test it doesn't mean it's safe, but if it does than you immediately know what happened).

If you tried to trace the code and couldn't find an issue, then let me know and I'll look deeper into it.

hare/eligibility/oracle.go
noamnelke: You might be right, but just trying to nitpick here: what if you get a higher `n` and this results in an overflow (wrap around) such that the next check for a value of `n` that's over the threshold doesn't trigger?

It sounds like we shouldn't care too much about this case, because it requires crazy weight and the only thing it allows is to get fewer eligibilities, but, you know, my dad always told me "Ordnung muß sein"...

noamnelke: OTOH, here we only checked that `minerWeight` fits in an `int`, so it's actually safer the way you did it. I don't think these nuances are terribly important, so I support removing this check here and keeping your other changes.

noamnelke (approved):

hare/eligibility/oracle.go
fasmat: I know that the second check is necessary - I was just nitpicking the different reactions of the node if the check doesn't pass. Shutting down seems extreme here 😅. Its probably very unlikely but if some node manages to get enough weight to pass the threshold this check will shutdown the whole network (or at least all nodes using `go-spacemesh`).

hare/eligibility/oracle.go
fasmat: I traced n a bit through the code, besides logging it is only used in 2 places:

https://github.com/spacemeshos/go-spacemesh/blob/23ba30c8cecbd13059bccd1a83dfb5750921a5a1/hare/eligibility/oracle.go#L297-L300

Here if `n > 2^16` the `if` will evaluate to `false` - `eligibilityCount` is an `uint16`.

https://github.com/spacemeshos/go-spacemesh/blob/23ba30c8cecbd13059bccd1a83dfb5750921a5a1/hare/eligibility/oracle.go#L356-L360

This is the only other place. Here `x` and `n` can be up to `2^30 - 1` and are cast to an `uint16` before being returned (without a check for an overflow). The returned value is only ever checked against zero, so an overflow someone who should be eligible will be randomly counted as not because they have an exact multiple of `2 ^ 16` for their `n`.

fasmat: The problem I see is that any value above 65536 for `n` will cause overflows during eligibility calculation. `n` is the miner weight (sometimes multiplied by committee size), but miner weights are defined as `uint64`s.

With a large enough weight it will overflow and transmit a `hare.Message` with `Eligibility.Count` lower than it should be.

fasmat:
bors try

fasmat:
bors try

hare/eligibility/oracle.go
```diff
@@ -356,9 +351,12 @@ func (o *Oracle) CalcEligibility(
 
 	for x := 0; x < n; x++ {
 		if fixed.BinCDF(n, p, x).GreaterThan(vrfFrac) {
+			// even with large N and large P x will be << 2^16, so this cast is safe
 			return uint16(x), nil
 		}
 	}
+
+	// since BinCDF(n, p, n) is 1 for any p this code is only reached if n << 2^16
```
countvonzero: i don't understand this comment. what does "if n << 2^16" mean?

countvonzero (approved):

hare/eligibility/oracle.go
fasmat: I don't know how I came up with this but I meant if n is significantly smaller than 2^16. I'll update the comment.

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

countvonzero:
bors merge

fasmat:
bors try

fasmat:
bors merge

PR #4276, opened 2023-04-11
addressgen util

dshulyak:
input is a public key encoded in hexary (without 0x prefix). the only option is -hrp, which is set to smtest by default.

```
go run ./genvm/cmd/addressgen/ 16a4c4050362db7ed31203364900c86bdc181e7ee25c1ed4aa64e3e32a1e0699
```

wallet: smtest1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqgu79e6d
public key: 0x16a4c4050362db7ed31203364900c86bdc181e7ee25c1ed4aa64e3e32a1e0699
coinbase: smtest1qqqqqqrg6p3vfj9fq7dv8lhej37fgpqxz676hagtlzwz2

### DIFF
--- a/genvm/cmd/addressgen/main.go
+++ b/genvm/cmd/addressgen/main.go
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	hrp = flag.String("hrp", "smtest", "network human readable prefix")
+	hrp = flag.String("hrp", "stest", "network human readable prefix")
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


lrettig:
@dshulyak are you sure this belongs in go-sm, rather than in `smcli`? smcli already has a `wallet` feature that generates keypairs in hex format. It seems like a logical extension to add this there.

dshulyak:
yes it is better to have it elsewhere. i think somebody needed it for something

PR #4277, opened 2023-04-11
[Merged by Bors] - Fix Transactions being cached before successful publish

### LINKED ISSUE #4043: Fix tx validation in `TransactionService.SubmitTransaction` api endpoint
## Description
Currently API endpoint behaves strangely.

If you will try to send invalid transaction, for example this one:
```
{
  "transaction": "AAAAAACAuTo3d7O7ZNiSePx57217lBYfzQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAQEBVJ+wdOdEg4jRtwMbQWhqDEMz0eU87uFQM/zTFwOXyNaJMITgwtDY00Q0eBb139ZEO+I5SRRtbU+24rCFZ7NDMJZi5ovNzhAfKDQr6A105iXhtL1Nf9Z7fb5/ONwYwI="
}
```

it will be rejected first time (with "failed to publish transaction" error, more details can be found only in the node logs).
```
{
  "error": "13 INTERNAL: Failed to publish transaction"
}
```

However if you will try to submit the same transaction second time, the Node will response with "okay-ish" message:
```
{
    "status": {
        "details": [],
        "code": 0,
        "message": ""
    },
    "txstate": {
        "id": {
            "id": {
                "type": "Buffer",
                "data": [
                    178, 134, 123, 221, 221, 95, 31, 11, 60, 129, 148, 172, 97,
                    245, 50, 61, 4, 160, 109, 222, 123, 221, 254, 18, 100, 70,
                    9, 85, 77, 221, 61, 199
                ]
            }
        },
        "state": "TRANSACTION_STATE_MEMPOOL"
    }
}
```

## Steps to reproduce
1. Submit invalid transaction to the node first time
2. Submit invalid transaction to the node second time

## Actual Behavior
Node responses with "okay-ish" response second time.

## Expected Behavior
It should reject invalid transaction all the time.

## Environment
 - Node Version: v0.2.21-beta.0

fasmat:
## Motivation
Closes #4043

## Changes
This change validates a nodes own transactions before publishing them with libp2p. This avoids them being cached and therefore multiple submissions of the same (invalid) transaction will continue to fail instead of returning "in mempool"

## Test Plan
- Updated existing tests
- Added tests that ensure a nodes own transactions are validated first before publishing them via libp2p (and relying on the validation there)

## TODO

## DevOps Notes


fasmat:
bors try

api/grpcserver/transaction_service.go
```diff
@@ -63,16 +66,20 @@ func (s TransactionService) SubmitTransaction(ctx context.Context, in *pb.Submit
...
 
 	if !s.syncer.IsSynced(ctx) {
-		return nil, status.Error(codes.FailedPrecondition,
-			"Cannot submit transaction, node is not in sync yet, try again later")
+		return nil, status.Error(codes.FailedPrecondition, "Cannot submit transaction, node is not in sync yet, try again later")
+	}
+
+	if err := s.txHandler.VerifyAndCacheTx(ctx, in.Transaction); err != nil {
+		log.Error("error verifying incoming tx: %v", err)
+		return nil, status.Error(codes.InvalidArgument, "Failed to verify transaction")
```
dshulyak, line 73: i think it is better to return the error and drop the log, as it will be printed grpc middleware anyway

api/grpcserver/interface.go
```diff
@@ -0,0 +1,9 @@
+package grpcserver
+
+import "context"
+
+//go:generate mockgen -package=grpcserver -destination=./mocks.go -source=./interface.go
+
+type txValidator interface {
```
dshulyak: other interfaces are defined right in the api/ . maybe that wasn't necessary, but maybe for consistency worth moving this one as well

txs/conservative_state_test.go
```diff
@@ -65,10 +69,13 @@ type testConState struct {
 	logger log.Log
 	db     *sql.Database
 	mvm    *MockvmState
+
+	key crypto.PrivKey
```
dshulyak: key seems to be unused

api/grpcserver/interface.go
fasmat: I can move it there for now.

Imo in a future PR we can merge the packages `api`, `api/config` and `api/grpcserver` into a single package `grpcserver`: The interfaces defined in the `api` package are only used in the `api/grpcserver` package, while their name - e.g. `api.Syncer` - would suggest they are intended to be used in multiple places. The `api/config` package contains configuration for the `api/grpcserver` package. `config.DefaultConfig()` is less expressive than `grpcserver.DefaultConfig()` and also has the disadvantage that we have to use named imports whenever we use more than one "config" package.

What do you think?

dshulyak: sounds good

dshulyak (approved):

fasmat:
bors merge

api/grpcserver/interface.go
fasmat: created PR: https://github.com/spacemeshos/go-spacemesh/pull/4297

PR #4279, opened 2023-04-12
[Merged by Bors] - check ballot and its atx have same smesher

### LINKED ISSUE #4252: validate that ballot is from the same identity as a referenced activation
we are missing this check in ballots handler, it is essential for "security" of the protocol, and apparently it was just missing forever

countvonzero:
## Motivation
Closes #4252

### DIFF
--- a/proposals/handler.go
+++ b/proposals/handler.go
@@ -45,6 +45,7 @@ var (
 	errKnownBallot           = errors.New("known ballot")
 	errInvalidVote           = errors.New("invalid layer/height in the vote")
 	errMaliciousBallot       = errors.New("malicious ballot")
+	errWrongSmesherID        = errors.New("ballot atx from a different smesher")
 )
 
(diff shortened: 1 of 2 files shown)


dshulyak (approved):

proposals/handler.go
```diff
@@ -45,6 +45,7 @@ var (
 	errKnownBallot           = errors.New("known ballot")
 	errInvalidVote           = errors.New("invalid layer/height in the vote")
 	errMaliciousBallot       = errors.New("malicious ballot")
+	errWrongSmesherID        = errors.New("ballot atx from a different smesher")
```
dshulyak, line 48: i think it is not really necessary to add such errors if this is only for information and testing. in both cases string-based error seems to be sufficient

dshulyak (approved):

dshulyak:
bors merge

proposals/handler.go
countvonzero: noted.

PR #4290, opened 2023-04-13
[Merged by Bors] - api, vm: expose api to parse transaction 

dshulyak:
it allows to check gas and other fields befor submitting transaction.

requires https://github.com/spacemeshos/api/pull/228 . will update go.mod once merged


api/grpcserver/transaction_service.go
```diff
@@ -54,10 +55,29 @@ func NewTransactionService(
...
+	raw := types.NewRawTx(in.Transaction)
+	req := s.conState.Validation(raw)
+	header, err := req.Parse()
+	if errors.Is(err, core.ErrNotSpawned) {
+		return nil, status.Error(codes.NotFound, "account is not spawned")
+	} else if errors.Is(err, core.ErrMalformed) {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	} else if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
```
fasmat, line 73:
```suggestion
	switch {
	case errors.Is(err, core.ErrNotSpawned):
		return nil, status.Error(codes.NotFound, "account is not spawned")
	case errors.Is(err, core.ErrMalformed):
		return nil, status.Error(codes.InvalidArgument, err.Error())
	case err != nil:
		return nil, status.Error(codes.Internal, err.Error())
	}
```

api/grpcserver/transaction_service.go
```diff
@@ -54,10 +55,29 @@ func NewTransactionService(
...
+	if in.Verify && !req.Verify() {
+		return nil, status.Error(codes.InvalidArgument, "signature is invalid")
+	}
+	tx := types.Transaction{RawTx: raw, TxHeader: header}
+	return &pb.ParseTransactionResponse{Tx: castTransaction(&tx)}, nil
+}
+
 // SubmitTransaction allows a new tx to be submitted.
 func (s TransactionService) SubmitTransaction(ctx context.Context, in *pb.SubmitTransactionRequest) (*pb.SubmitTransactionResponse, error) {
-	log.Info("GRPC TransactionService.SubmitTransaction")
```
fasmat, line 61: I found these logs useful in error reports.

api/grpcserver/transaction_service_test.go
```diff
@@ -186,3 +195,101 @@ func BenchmarkStreamResults(b *testing.B) {
...
+		},
+	} {
+		tc := tc
+		t.Run(tc.desc, func(t *testing.T) {
+			resp, err := client.ParseTransaction(context.Background(), &pb.ParseTransactionRequest{
+				Transaction: tc.tx,
+				Verify:      tc.verify,
+			})
+			tc.expect(t, resp, err)
+		})
```
fasmat, line 293: I think typing out each test case as its own `t.Run` makes them much easier to read with about the same amount of code.

fasmat (approved):

api/grpcserver/transaction_service.go
dshulyak: they are logged in grpc middleware

api/grpcserver/transaction_service_test.go
dshulyak: i want two things:
- understand test by looking at the inputs and outputs
- minimize amount of code that i have to change if some interface changes. so if i change ParseTransaction - i want to change it in one place, rather than 5

if i type out every test in Run, then i loose both. another alternative that makes sense to me is to make a function called parseTransactionTest and drop table, but it allows more "flexibility" in writing tests, which i would rather avoid when possible

dshulyak:
bors merge

dshulyak:
bors merge

PR #4295, opened 2023-04-14
[Merged by Bors] - activation: publish atx based on local clock

### LINKED ISSUE #3774: ATX: remove restrictions on positioning ATX

### LINKED ISSUE #4235: atxs: change expectations in the code that waits for the first atx
we have code that waits for first atx https://github.com/spacemeshos/go-spacemesh/blob/develop/activation/activation.go#L284-L306

the goal here is too select good enough atx from the recent epoch, so that new smesher is onboarded right for the next epoch. however we assume a bit small proving time, which will not allow us to select recent atx and will cause new smesher to wait one more additional epoch, before it will start getting rewards.

i think the good enough solution is to tie it to the cycle gap, lets say wait till 4/5 of cycle gap, and then select best available atx.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4235
closes: https://github.com/spacemeshos/go-spacemesh/issues/3774

in this change builder waits till the last point (poet round start - grace period) to get the most fresh and best positioning atx. it will try to submit challenge as long as there is some window before poet round starts, otherwise it will fail and will have to wait for another epoch. this code works the same way for initial atx and for long running smesher.

and, it changes how long we wait for poet to return proofs. grace period now used for two purposes:
- it gives poet time to prepare proof and serve Proof grpc request
- and it should be sufficient to prepare poet pow and submit grpc request with challenge

effective post time should fit into cycle gap - 2 * grace periods.

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -245,8 +245,6 @@ func (b *Builder) run(ctx context.Context) {
 		return
 	case <-b.layerClock.AwaitLayer(types.LayerID(0)):
 	}
-
-	b.waitForFirstATX(ctx)
 	b.loop(ctx)
 }
 
@@ -266,69 +264,6 @@ func (b *Builder) generateProof(ctx context.Context) error {
 	return nil
 }
 
-// waitForFirstATX waits until the first ATX can be published. The return value indicates
-// if the function waited or not (for testing).
-func (b *Builder) waitForFirstATX(ctx context.Context) bool {
-	currentLayer := b.layerClock.CurrentLayer()
-	currEpoch := currentLayer.GetEpoch()
-	if currEpoch == 0 { // genesis miner
-		return false
-	}
-	if prev, err := b.cdb.GetLastAtx(b.nodeID); err == nil {
-		if prev.PublishEpoch == currEpoch {
-			// miner has published in the current epoch
-			return false
-		}
-	}
-
-	// miner didn't publish ATX that targets current epoch.
-	// This estimate work if the majority of the nodes use poet servers that are configured the same way.
-	// TODO: do better when nodes use poet services with different phase shifts.
-	// We must wait till an ATX from other nodes arrives.
-	// To calculate the needed time we find the Poet round end
-	// and add twice network grace plus average PoST duration time.
-	// The first grace time is for poet proof propagation, The second one - for the ATX.
-	// Max wait time depicts an estimated safe moment to still be able to
-	// submit a poet challenge with the obtained ATX.
-	//
-	//                 Grace ──────┐
-	//                 Period      │  ┌─ATX arrives
-	//                     │  PoST │  │   ┌ wait deadline
-	//                     │ ┌───► │  │   │   ┌ Next round start
-	//   ┌────────────────┬┴─►   └─┴─►│   │   ▼───────
-	//   │  POET ROUND N  │           │   │   │ ROUND N+1
-	// ──┴────────────┬───┴───────────▼───▼───┴───────►time
-	//     EPOCH N    │      EPOCH N+1
-	// ───────────────┴───────────────────────────────
-	averagePostTime := time.Second // TODO should probably come up with a reasonable average value.
-	poetRoundEndOffset := b.poetCfg.PhaseShift - b.poetCfg.CycleGap
-	atxArrivalOffset := poetRoundEndOffset + 2*b.poetCfg.GracePeriod + averagePostTime
-	waitDeadline := b.layerClock.LayerToTime(currEpoch.FirstLayer()).Add(b.poetCfg.PhaseShift).Add(-b.poetCfg.GracePeriod)
-
-	var expectedAtxArrivalTime time.Time
-	if time.Now().After(waitDeadline) {
-		b.log.WithContext(ctx).With().Info("missed the window to submit a poet challenge. Will wait for next epoch.")
-		expectedAtxArrivalTime = b.layerClock.LayerToTime((currEpoch + 1).FirstLayer()).Add(atxArrivalOffset)
-	} else {
-		expectedAtxArrivalTime = b.layerClock.LayerToTime(currEpoch.FirstLayer()).Add(atxArrivalOffset)
-	}
-
-	waitTime := time.Until(expectedAtxArrivalTime)
-	timer := time.NewTimer(waitTime)
-	defer timer.Stop()
-
-	b.log.WithContext(ctx).With().Info("waiting for the first ATX", log.Duration("wait", waitTime))
-	select {
-	case <-ctx.Done():
-		return false
-	case <-timer.C:
-	}
-	b.log.WithContext(ctx).With().Info("ready to build first atx",
-		log.Stringer("current_layer", b.layerClock.CurrentLayer()),
-		log.Stringer("current_epoch", b.layerClock.CurrentLayer().GetEpoch()))
-	return true
-}
-
 func (b *Builder) receivePendingPoetClients() *[]PoetProvingServiceClient {
 	return b.pendingPoetClients.Swap(nil)
 }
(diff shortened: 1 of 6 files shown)


activation/activation_test.go
```diff
@@ -173,11 +173,18 @@ func publishAtx(
...
 
 	publishEpoch := posEpoch + 1
-	tab.mclock.EXPECT().CurrentLayer().Return(*currLayer).AnyTimes()
+	// this approach makes no sense for mocking. it rewrites current layer for every
+	// subsequent request, and makes all future calls noop
+	if currentLayerExpect < 0 {
+		tab.mclock.EXPECT().CurrentLayer().Return(*currLayer).AnyTimes()
+	} else {
+		tab.mclock.EXPECT().CurrentLayer().Return(*currLayer).Times(currentLayerExpect)
+	}
```
fasmat: I feel like this could be called by the caller of `publishAtx`?

fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
@fasmat could you please review it again?

activation/activation.go
```diff
@@ -345,22 +280,12 @@ func (b *Builder) loop(ctx context.Context) {
...
 		if err == nil {
-			b.log.WithContext(ctx).With().Info("waiting for atx to propagate before building the next challenge",
-				log.Duration("wait", b.poetRetryInterval),
-			)
-			select {
-			case <-ctx.Done():
-				return
-			case <-time.After(b.poetRetryInterval):
-			}
 			continue
```
fasmat, line 283: I would not immediately retry again. We already had issues with nodes essentially DOS-ing on our poets and ending up in an endless loop without ever recovering.

activation/activation.go
```diff
@@ -403,26 +328,53 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) (*types.NIPostChalle
...
 	}
+	current := b.currentEpoch()
+	prev, err := b.cdb.GetLastAtx(b.nodeID)
+	if err != nil {
+		if !errors.Is(err, sql.ErrNotFound) {
+			return nil, err
+		}
+	} else if prev.PublishEpoch == current+1 {
+		current += 1
+	}
```
fasmat, line 339:
```suggestion
	if err != nil && !errors.Is(err, sql.ErrNotFound) {
		return nil, err
	}
	if prev.PublishEpoch == current+1 {
		current += 1
	}
```

Why is the check for `prev.PublishEpoch` necessary? Will the node eagerly create ATXs for more than one epoch in advance?

activation/activation.go
```diff
@@ -403,26 +328,53 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) (*types.NIPostChalle
...
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-time.After(wait):
+		}
+	} else if until <= 0 {
+		metrics.PublishLateWindowLatency.Observe(-until.Seconds())
+		return nil, fmt.Errorf("%w: builder doesn't have time to submit in epoch %d. poet round already started %v ago",
+			ErrATXChallengeExpired, current, -until)
+	}
```
fasmat: I prefer checking for the error case first:

```suggestion
	until := time.Until(b.poetRoundStart(current))
	if until <= 0 {
		metrics.PublishLateWindowLatency.Observe(-until.Seconds())
		return nil, fmt.Errorf("%w: builder doesn't have time to submit in epoch %d. poet round already started %v ago",	ErrATXChallengeExpired, current, -until)
	}
	metrics.PublishOntimeWindowLatency.Observe(until.Seconds())
	if until > b.poetCfg.GracePeriod {
		wait := until - b.poetCfg.GracePeriod
		b.log.WithContext(ctx).With().Debug("waiting for fresh atxs",
			log.Duration("till poet round", until),
			log.Uint32("current epoch", current.Uint32()),
			log.Duration("wait", wait),
		)
		select {
		case <-ctx.Done():
			return nil, ctx.Err()
		case <-time.After(wait):
		}
	}
```

activation/activation.go
```diff
@@ -506,11 +458,13 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 
 	challenge, err := b.loadChallenge()
 	if err != nil {
-		logger.With().Info("challenge not loaded", log.Err(err))
-		logger.With().Info("building new atx challenge", log.Stringer("current_epoch", b.currentEpoch()))
+		logger.With().Info("building new atx challenge",
+			log.Stringer("current_epoch", b.currentEpoch()),
+			log.Err(err),
```
fasmat, line 462: does this error need to be logged? Won't it just be `sql.NotFound` most of the time?

activation/activation_test.go
```diff
@@ -1113,3 +1078,51 @@ func TestBuilder_UpdatePoetsUnstable(t *testing.T) {
...
+		tc := tc
+		t.Run(tc.desc, func(t *testing.T) {
+			tab := newTestBuilder(t, WithPoetConfig(PoetConfig{
+				PhaseShift:  tc.shift,
+				GracePeriod: tc.grace,
+			}))
+			tab.mclock.EXPECT().CurrentLayer().Return(types.LayerID(0)).AnyTimes()
+			tab.mclock.EXPECT().LayerToTime(gomock.Any()).DoAndReturn(func(lid types.LayerID) time.Time {
+				return genesis.Add(time.Duration(lid) * layerDuration)
+			}).AnyTimes()
```
fasmat, line 1103: NIT:

```suggestion
			tab.mclock.EXPECT().LayerToTime(gomock.Any()).Return(genesis.Add(time.Duration(lid) * layerDuration)).AnyTimes()
```

activation/activation.go
dshulyak: this case is for err == nil. if we exited successfully - node submitted atx in the epoch and will wait for the next round to submit for the next one

activation/activation.go
dshulyak: lets say node successfully submitted atx, and calls this function again (might be after restart). we check that it already submitted atx, and therefore needs to wait for +1 round

dshulyak: btw the suggestion doesn't work, as we can't call prev is error is not nill

activation/activation.go
dshulyak: i wanted to keep it, just in case

activation/activation_test.go
dshulyak: this should return different results based on lid argument

dshulyak:
bors try

activation/activation.go
fasmat: Ah yes, you're right regarding the suggested code. 🙂 

But I'm still failing to understand the `prev.PublishEpoch == current+1` check?

If the prev. ATX is form the current epoch then all is fine and we create one for the next, but if prev. ATX is for the next epoch we create another one for the epoch after that?

dshulyak:
bors try

activation/activation.go
fasmat: 🤦 never mind we wait until `current`...

fasmat (approved):

dshulyak:
bors merge

PR #4297, opened 2023-04-14
[Merged by Bors] - Restructure grpcserver package

fasmat:
## Motivation
Follow up to https://github.com/spacemeshos/go-spacemesh/pull/4277: move config and interfaces / mocks into the `grpcserver` package

## Changes
- some of the handwritten mocks have been replaced by generated ones
- `cmd.Version` and `cmd.Commit` are passed into the service instead of accessing it in the `cmd` package (to avoid cyclic dependencies)
  - they shouldn't change during runtime so this shouldn't break anything

## Test Plan
Existing tests pass

## TODO

## DevOps Notes


activation/validation_test.go
```diff
@@ -65,7 +65,7 @@ func Test_Validation_VRFNonce(t *testing.T) {
 	t.Run("wrong commitmentAtxId", func(t *testing.T) {
 		t.Parallel()
 
-		commitmentAtxId := types.RandomATXID()
```
fasmat, line 68: This test failed in a CI run. The commitmentAtx being random gives a small chance of coincidentally passing the VRFNonce check.

Set it to a fixed value to avoid that in the future.

piersy:
Hey @fasmat why did you merge the config package with the grpcserver package?

That seems like it's going to cause problems with tight coupling and circular dependencies (and already did in this PR).

Now anyone depending on the config package has to pull in all the dependencies of the grpcserver package, which contains a whole ton of functional dependencies which are totally unnecessary for understanding the config which if kept in a separate package only has a dependecy on the std `time` package.

piersy:
@fasmat I'd also like to know why you merged the API package with the grpcserver package, I don't see any problems with that, but I'd like to know your reasoning?

fasmat:
> @fasmat I'd also like to know why you merged the API package with the grpcserver package, I don't see any problems with that, but I'd like to know your reasoning?

The `api` package contained 3 things: `grpcserver`, `config` (the config for the `grpcserver` package) and interfaces & mocks (that ware only used in the `grpcserver` package). I feel like these belong functionally together thats why I merged them. 🙂 

> Hey @fasmat why did you merge the config package with the grpcserver package?
> 
> That seems like it's going to cause problems with tight coupling and circular dependencies (and already did in this PR).
> 
> Now anyone depending on the config package has to pull in all the dependencies of the grpcserver package, which contains a whole ton of functional dependencies which are totally unnecessary for understanding the config which if kept in a separate package only has a dependecy on the std time package.

The `api/config` package does not stand for itself. There were only 2 packages using it: `cmd/node` and `config`. The former already has a dependency on `grpcserver`. The later serves as some kind of meta-package that imo we don't need - see below.

Additionally all of these separate `config` packages introduce types that are called `config.Config` and `config.DefaultConfig`, wouldn't names like `hare.Config`, `tortoise.Config`, etc. make more sense? Imo it makes no sense to have separate sub-`config` packages for everything. They are not used without the package they belong to (except for our global `config` package).

The reason the global `config` exist is because we need to have some way to parse command line options and configuration files. Now the problem is that our configuration files are tightly coupled to our implementation. I cannot change any config of any service without also implicitly changing the layout of our config files. Imo this is wrong; our configuration should not be defined by our implementation but be its separate unit (in its separate package used by `cmd/node` but not dependent on anything else).

countvonzero (approved):

fasmat:
bors merge

PR #4299, opened 2023-04-14
Split unit tests into two batches in CI

piersy:
## Motivation
CI is slow, this should drop time for CI to complete by a significant amount. From the limited results I have it looks like it is dropping overall CI time from about 25min to 15 min. So saving roughly 10 mins.

## Changes
Split unit tests into 2 roughly equal (in running time) chunks that are executed in parallel by the CI.

## Test Plan
See that it works in CI

## DevOps Notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,9 +121,21 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
+        # We use a matrix to split packages into 2 parts to halve the testing time.
+        packages:
+          - part1
+          - part2
         include:
           - os: ubuntu-latest
             allow-failure: "false"
+          - packages: part1
+            package-selection: "go list ./hare/... ./activation/... ./beacon/..."
+          - packages: part2
+            package-selection: "go list ./...  \
+              | grep -v github.com/spacemeshos/go-spacemesh/systest/tests \
+              | grep -v github.com/spacemeshos/go-spacemesh/hare \
+              | grep -v github.com/spacemeshos/go-spacemesh/activation \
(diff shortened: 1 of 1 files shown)


dshulyak:
systests take about 25 minutes to complete in success run. i see that unit tests on macos (slowest) take about the same time. why does it matter that unit tests will become faster?

piersy:
@dshulyak It reduces the feedback loop time to find out that something failed in the CI run.

dshulyak:
i understand that it kindof does that. but the next person will need to decide where to put additional tests, or reorg them to benefit from parallelization.. change doesn't seem right to me. unit tests failure wasn't an issue for me, given if they are not terribly flaky. beside that linux unit tests are significantly faster than macos. i guess i will wait for someone else to confirm that people prefer this approach :)

piersy:
@dshulyak Yes it probably will require occasional re-organizing, but that doesn't seem to me to be a huge burden. Also nothing breaks if you don't re-organize, you just might end up with the tests not split at all.

piersy:
@dshulyak you said the change doesn't seem right, what would be the right change?

dshulyak:
> you said the change doesn't seem right, what would be the right change?

parallelization for testing modules with -p. and slow tests within module with t.Parallel

dshulyak:
i experimented here https://github.com/spacemeshos/go-spacemesh/actions/runs/4729988327/jobs/8393143621?pr=4311 with t.Parallel for slow tests, and removing -p restrictions so that it can run modules in parallel. it became twice faster on linux env. it probably can be made faster by spending more time on  it

piersy:
Ok I'm closing in favour of using t.Parallel and removing core restrictions.

PR #4300, opened 2023-04-14
[Merged by Bors] - Check if all bytes have been consumed during scale decode

### LINKED ISSUE #4133: codec: validate that messages are fully decoded
without such validation it is possible to make node to allocate and keep bloated messages in memory and in state. for example the max valid transaction is around 500 bytes, someone can submit larger message (up to 4MB using grpc api, and 1MB using gossip), where up to first 500 will be a valid message and the rest is bloat.

```go
func DecodeFull(buf []byte, obj scale.Decodable) error {
     // scale.Decoder returns number of bytes read
     // this number must be equal to the size of the original buf
}
```

With transaction it is somewhat more complex, but such validation is also necessary.

fasmat:
## Motivation
Closes #4133

## Changes
- `codec.Decode` was changed to check if all bytes that were provided for decoding were consumed in the process.
  - if not the function now returns a named error: `ErrShortRead`
- In tests checking that layerID / epochID is first value in encoded bytes decoding was updated to not fail on `ErrShortRead`

## Test Plan
- All existing tests pass

## TODO

## DevOps Notes

### DIFF
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -71,7 +71,7 @@ func Decode(buf []byte, value Decodable) error {
 	return nil
 }
 
-// Decode value from a byte buffer.
+// DecodeSome value from a byte buffer without checking for full consumption.
 func DecodeSome(buf []byte, value Decodable) error {
 	if _, err := DecodeFrom(bytes.NewBuffer(buf), value); err != nil {
 		return fmt.Errorf("decode from buffer: %w", err)
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -61,6 +61,18 @@ func Encode(value Encodable) ([]byte, error) {
 
 // Decode value from a byte buffer.
 func Decode(buf []byte, value Decodable) error {
+	n, err := DecodeFrom(bytes.NewBuffer(buf), value)
+	if err != nil {
+		return fmt.Errorf("decode from buffer: %w", err)
+	}
+	if n != len(buf) {
+		return fmt.Errorf("decode from buffer: not all bytes were consumed")
+	}
+	return nil
+}
+
+// Decode value from a byte buffer.
+func DecodeSome(buf []byte, value Decodable) error {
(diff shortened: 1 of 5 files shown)


codec/codec.go
```diff
@@ -61,6 +61,18 @@ func Encode(value Encodable) ([]byte, error) {
...
+		return fmt.Errorf("decode from buffer: %w", err)
+	}
+	if n != len(buf) {
+		return fmt.Errorf("decode from buffer: not all bytes were consumed")
+	}
+	return nil
+}
+
+// Decode value from a byte buffer.
+func DecodeSome(buf []byte, value Decodable) error {
```
fasmat: Not a fan of the name, suggestions welcome! 🙂

Alternatively: we can stick with one `Decode` function and return an error like `ErrShortRead` that we can check for when we only decode some of the bytes from the buffer.

go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/spacemeshos/api/release/go v1.7.0
 	github.com/spacemeshos/economics v0.0.0-20220930194415-799d50b0431d
 	github.com/spacemeshos/fixed v0.0.0-20210523192743-8d17e03c169a
-	github.com/spacemeshos/go-scale v1.1.7
+	github.com/spacemeshos/go-scale v1.1.8-0.20230414162212-bc708740c191
```
fasmat: TODO: merge after https://github.com/spacemeshos/go-scale/pull/55

codec/codec.go
dshulyak: DecodeFull/DecodeAll ?

dshulyak: i agree with what you suggested. also if someone wants to perform decoding without this check they can init scale.Decoder

dshulyak (approved):

fasmat:
bors merge

Issue #4296, opened 2023-04-14
atxs: possible memory leak on atxs/post code path


dshulyak:
![image](https://user-images.githubusercontent.com/1964576/232000552-c3316326-0512-45c1-a660-7b9f56781a14.png)

[pprof.go-spacemesh.alloc_objects.alloc_space.inuse_objects.inuse_space.087.pb.gz](https://github.com/spacemeshos/go-spacemesh/files/11230997/pprof.go-spacemesh.alloc_objects.alloc_space.inuse_objects.inuse_space.087.pb.gz)

memory usage grew almost to 2GB on 2000 nodes network. I don't understand yet what contributed to such growth. On 400 nodes network it was significantly less after almost a week. 

![image](https://user-images.githubusercontent.com/1964576/232001243-dfbcec31-ddfa-4da0-ba0b-1a803e0b9437.png)

dshulyak:
Will try to debug it with higher sampling rate.

dshulyak:
Votes in beacon is most likely explanation, i didn't run network with large number of beacon rounds and high participation before. So this seems what is causing this growth.

dshulyak:
rss grew to more than 1GB (1st screenshot), then some part of it got released but we still keep very large heap ~900MB. which doesn't correspond with data provided by golang, so it does look like that we don't release memory in rust

![image](https://user-images.githubusercontent.com/1964576/232966461-f4cf3014-4153-4087-9bd0-007361079b48.png)

about the same time node started to accept atxs, note that literally nothing else is going on at this time

![image](https://user-images.githubusercontent.com/1964576/232966619-fb16c36e-d7a9-44e2-bbad-3266ecc79988.png)

inuse space (asn is about 150MB) 

![image](https://user-images.githubusercontent.com/1964576/232966716-eb8c2c86-f07b-42f1-8f26-b919e3b9a283.png)

poszu:
I ran valgrind on this test: https://github.com/spacemeshos/post-rs/blob/f7d93cbedd9521f15edc358a6a875de71f2094f7/tests/generate_and_verify.rs#LL11C6-L11C6

It reports "possibly lost memory":
```
==10363== HEAP SUMMARY:
==10363==     in use at exit: 158,824 bytes in 311 blocks
==10363==   total heap usage: 2,073 allocs, 1,762 frees, 1,922,236 bytes allocated
==10363==
[...]

PR #4302, opened 2023-04-16
[Merged by Bors] - beacon: track already voted participants with bitset

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4296

### DIFF
--- a/beacon/handlers.go
+++ b/beacon/handlers.go
@@ -527,10 +527,10 @@ type votesTracker struct {
 }
 
 func (v *votesTracker) register(round types.RoundID) bool {
-	if uint32(round) > wordSize*uint32(len(v.words)) {
+	word := uint64(round) / wordSize
+	if word >= uint64(len(v.words)) {
 		return false
 	}
-	word := uint64(round) / wordSize
 	position := uint64(1) << (uint64(round) % wordSize)
 	if v.words[word]&position > 0 {
 		return false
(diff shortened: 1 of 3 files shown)


beacon/handlers.go
```diff
@@ -510,11 +502,45 @@ func (pd *ProtocolDriver) registerProposed(logger log.Log, epoch types.EpochID,
...
+}
+
+const wordSize = 64
+
+func newVotesTracker(limit uint32) *votesTracker {
+	b := bits.Len32(limit)
+	words := b / wordSize
+	if b%wordSize != 0 {
+		words += 1
+	}
```
countvonzero, line 521: calculate `words` just once instead of on every NodeID since `limit` is a configured constant?

beacon/handlers.go
```diff
@@ -510,11 +502,45 @@ func (pd *ProtocolDriver) registerProposed(logger log.Log, epoch types.EpochID,
...
+	word := uint64(round) / wordSize
+	position := uint64(1) << (uint64(round) % wordSize)
+	if v.words[word]&position > 0 {
+		return false
+	}
+	v.words[word] |= position
+	return true
+}
+
+func (v *votesTracker) voted(round types.RoundID) bool {
```
countvonzero, line 542: it's strange that you check for `round > wordSize*uint32(len(v.words))` in `register` but not in `voted`.

countvonzero (approved):

beacon/handlers.go
dshulyak: limit is from confguration file

dshulyak:
bors merge

PR #4303, opened 2023-04-16
libp2p: disable asn db lookup

dshulyak:
libp2p compiled with asn database that uses about ~150MB. as far as i know we are using it, and can disable until we do.

related: https://github.com/spacemeshos/go-spacemesh/issues/4296


dshulyak:
bors try

dshulyak:
bors try

pigmej (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

fasmat:
Is this still necessary? `v0.3.0` has been released that loads the asn db lazily. If we are not using it, it won't be loaded.

See: https://github.com/libp2p/go-libp2p-asn-util/releases/tag/v0.3.0

EDIT: I'm also not sure the replacement is effective, since we are using `v0.3.0` already and the replacement targets `v0.2.0`

dshulyak:
i probably didn't update this replacement after 0.3.0 was introduced, i was using this for older version, that is compatible with testnet5
but as you noticed it is not necessary anymore

PR #4319, opened 2023-04-19
Ensure that start smeshing cannot fail silently

### LINKED ISSUE #4292: activation: StartSmeshing should not fail silently
related codebase: https://github.com/spacemeshos/go-spacemesh/blob/develop/activation/activation.go#L193-L195

it can fail due to validation, but operator won't even know that smeshing didn't start, which is not that fun..
we also may consider using values specified in post_metadata.json , instead of defaults in the config, if such file already exists

piersy:
## Motivation

Closes https://github.com/spacemeshos/go-spacemesh/issues/4292

## Changes
This change removes the `prepareInitializer` method of the `PostSetupManager` and instead adds `InitializerConfig` which  when called generates, validates and returns the config to be used for POS initialization. `StartSession` has been modified to accept the `InitializationConfig` generated by `InitializerConfig`. Having this separate step allows for up front validation of post params before kicking off the long running task of initialisation, which allows the start smeshing api call to directly return an error to the caller if there is a problem with the smeshing options. Note `InitializerConfig` doesn't modify any state in the `PostSetupManager`, which makes it safe to call concurrently while initialization is occurring.

Also this change updates `Builder.StartSmeshing` to panic if the smeshing operation fails, this ensures that the node will crash if there is any problem with the smeshing background task.

## Test Plan
Updated existing tests and added a test to check that `InitializerConfig` returns an error if given invalid config.

## DevOps Notes

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -181,7 +181,7 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 	ctx, stop := context.WithCancel(b.parentCtx)
 	b.stop = stop
 
-	cfg, err := b.postSetupProvider.PrepareInitializer(b.parentCtx, opts)
+	cfg, err := b.postSetupProvider.InitializerConfig(b.parentCtx, opts)
 	if err != nil {
 		return fmt.Errorf("failed to prepare post initializer: %w", err)
 	}
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -163,17 +163,23 @@ func (b *Builder) Smeshing() bool {
 	return b.started.Load()
 }
 
-// StartSmeshing is the main entry point of the atx builder.
-// It runs the main loop of the builder and shouldn't be called more than once.
-// If the post data is incomplete or missing, data creation
-// session will be preceded. Changing of the post potions (e.g., number of labels),
-// after initial setup, is supported.
+// StartSmeshing is the main entry point of the atx builder. It runs the main
+// loop of the builder in a new go-routine and shouldn't be called more than
+// once. If the post data is incomplete or missing, data creation session will
+// be preceded. Changing of the post potions (e.g., number of labels), after
+// initial setup, is supported. If data creation fails for any reason then the
+// go-routine will panic.
 func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) error {
 	b.smeshingMutex.Lock()
 	defer b.smeshingMutex.Unlock()
 
 	if !b.started.CompareAndSwap(false, true) {
 		return errors.New("already started")
+	}
+
+	err := b.postSetupProvider.PrepareInitializer(b.parentCtx, opts)
+	if err != nil {
+		return fmt.Errorf("failed to prepare post initializer: %w", err)
 	}
 
 	b.coinbaseAccount = coinbase
@@ -190,8 +196,10 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 			// ensure we are ATX synced before starting the PoST Session
 		}
 
-		if err := b.postSetupProvider.StartSession(ctx, opts); err != nil {
-			return err
+		// If start session returns any error other than context.Canceled
+		// (which is how we signal it to stop) then we panic.
+		if err := b.postSetupProvider.StartSession(ctx, opts); err != nil && !errors.Is(err, context.Canceled) {
+			panic(fmt.Sprintf("initialization failed: %v", err))
 		}
 
 		b.run(ctx)
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -177,14 +177,14 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 		return errors.New("already started")
 	}
 
+	b.coinbaseAccount = coinbase
+	ctx, stop := context.WithCancel(b.parentCtx)
+	b.stop = stop
+
 	err := b.postSetupProvider.PrepareInitializer(b.parentCtx, opts)
 	if err != nil {
 		return fmt.Errorf("failed to prepare post initializer: %w", err)
 	}
-
-	b.coinbaseAccount = coinbase
-	ctx, stop := context.WithCancel(b.parentCtx)
-	b.stop = stop
 
 	b.eg.Go(func() error {
 		defer b.started.Store(false)
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -181,7 +181,7 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 	ctx, stop := context.WithCancel(b.parentCtx)
 	b.stop = stop
 
-	err := b.postSetupProvider.PrepareInitializer(b.parentCtx, opts)
+	cfg, err := b.postSetupProvider.PrepareInitializer(b.parentCtx, opts)
 	if err != nil {
 		return fmt.Errorf("failed to prepare post initializer: %w", err)
 	}
@@ -198,7 +198,7 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 
 		// If start session returns any error other than context.Canceled
 		// (which is how we signal it to stop) then we panic.
-		if err := b.postSetupProvider.StartSession(ctx, opts); err != nil && !errors.Is(err, context.Canceled) {
+		if err := b.postSetupProvider.StartSession(ctx, cfg); err != nil && !errors.Is(err, context.Canceled) {
 			panic(fmt.Sprintf("initialization failed: %v", err))
 		}
 
--- a/activation/interface.go
+++ b/activation/interface.go
@@ -58,6 +58,12 @@ type postSetupProvider interface {
 	Status() *PostSetupStatus
 	ComputeProviders() []PostSetupComputeProvider
 	Benchmark(p PostSetupComputeProvider) (int, error)
+	// PrepareInitializer can be called before StartSession to verify if the
+	// given opts are valid. It is not necessary since StartSession also calls
+	// PrepareInitializer, but it provides a means to understand if the post
+	// configuration is valid before kicking off a very long running task
+	// (StartSession can take hours to complete)
+	PrepareInitializer(ctx context.Context, opts PostSetupOpts) error
 	StartSession(context context.Context, opts PostSetupOpts) error
 	Reset() error
 	GenerateProof(ctx context.Context, challenge []byte) (*types.Post, *types.PostMetadata, error)
--- a/activation/interface.go
+++ b/activation/interface.go
@@ -58,12 +58,11 @@ type postSetupProvider interface {
 	Status() *PostSetupStatus
 	ComputeProviders() []PostSetupComputeProvider
 	Benchmark(p PostSetupComputeProvider) (int, error)
-	// PrepareInitializer can be called before StartSession to verify if the
-	// given opts are valid. It is not necessary since StartSession also calls
-	// PrepareInitializer, but it provides a means to understand if the post
-	// configuration is valid before kicking off a very long running task
-	// (StartSession can take hours to complete)
-	PrepareInitializer(ctx context.Context, opts PostSetupOpts) (*SessionConfig, error)
+	// InitializerConfig builds the config to be used by initialization in calls to
+	// StartSession. Having this function be separate from StartSession provides a
+	// means to understand if the post configuration is valid before kicking off a
+	// very long running task (StartSession can take hours to complete).
+	InitializerConfig(ctx context.Context, opts PostSetupOpts) (*SessionConfig, error)
 	StartSession(context context.Context, cfg *SessionConfig) error
 	Reset() error
 	GenerateProof(ctx context.Context, challenge []byte) (*types.Post, *types.PostMetadata, error)
--- a/activation/interface.go
+++ b/activation/interface.go
@@ -62,8 +62,8 @@ type postSetupProvider interface {
 	// StartSession. Having this function be separate from StartSession provides a
 	// means to understand if the post configuration is valid before kicking off a
 	// very long running task (StartSession can take hours to complete).
-	InitializerConfig(ctx context.Context, opts PostSetupOpts) (*SessionConfig, error)
-	StartSession(context context.Context, cfg *SessionConfig) error
+	InitializerConfig(ctx context.Context, opts PostSetupOpts) (*InitializationConfig, error)
+	StartSession(context context.Context, cfg *InitializationConfig) error
 	Reset() error
 	GenerateProof(ctx context.Context, challenge []byte) (*types.Post, *types.PostMetadata, error)
 	CommitmentAtx() (types.ATXID, error)
--- a/activation/interface.go
+++ b/activation/interface.go
@@ -63,8 +63,8 @@ type postSetupProvider interface {
 	// PrepareInitializer, but it provides a means to understand if the post
 	// configuration is valid before kicking off a very long running task
 	// (StartSession can take hours to complete)
-	PrepareInitializer(ctx context.Context, opts PostSetupOpts) error
-	StartSession(context context.Context, opts PostSetupOpts) error
+	PrepareInitializer(ctx context.Context, opts PostSetupOpts) (*SessionConfig, error)
+	StartSession(context context.Context, cfg *SessionConfig) error
 	Reset() error
 	GenerateProof(ctx context.Context, challenge []byte) (*types.Post, *types.PostMetadata, error)
 	CommitmentAtx() (types.ATXID, error)
--- a/activation/post.go
+++ b/activation/post.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/spacemeshos/post/config"
 	"github.com/spacemeshos/post/gpu"
@@ -88,10 +89,11 @@ type PostSetupManager struct {
 	db          *datastore.CachedDB
 	goldenATXID types.ATXID
 
-	mu       sync.Mutex                  // mu protects setting the values below.
-	lastOpts *PostSetupOpts              // the last options used to initiate a Post setup session.
-	state    PostSetupState              // state is the current state of the Post setup.
-	init     *initialization.Initializer // init is the current initializer instance.
+	mu           sync.Mutex                  // mu protects setting the values below.
+	lastOpts     *PostSetupOpts              // the last options used to initiate a Post setup session.
+	state        PostSetupState              // state is the current state of the Post setup.
+	init         *initialization.Initializer // init is the current initializer instance.
+	initializing atomic.Bool                 // keeps track of when initialisation is ocurring.
 }
 
 // NewPostSetupManager creates a new instance of PostSetupManager.
@@ -173,16 +175,13 @@ func (mgr *PostSetupManager) Benchmark(p PostSetupComputeProvider) (int, error) 
 // StartSession starts (or continues) a PoST session. It supports resuming a previously started
 // session, and will return an error if a session is already in progress.
 //
-// Ensure that before calling this method, the node is ATX synced.
+// Insure that before calling this method, the node is ATX synced.
 func (mgr *PostSetupManager) StartSession(ctx context.Context, opts PostSetupOpts) error {
-	if err := mgr.PrepareInitializer(ctx, opts); err != nil {
-		return err
+	// Ensure only one goroutine can execute initialization at a time.
+	if !mgr.initializing.CompareAndSwap(false, true) {
+		return fmt.Errorf("post setup session in progress")
 	}
-
-	if err := mgr.checkAndSetInProgress(); err != nil {
-		return err
-	}
-
+	defer mgr.initializing.Store(false)
 	mgr.logger.With().Info("post setup session starting",
 		log.String("node_id", mgr.id.String()),
 		log.String("commitment_atx", mgr.commitmentAtxId.String()),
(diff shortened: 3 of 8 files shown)


activation/activation.go
```diff
@@ -176,6 +177,11 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 		return errors.New("already started")
 	}
 
+	err := b.postSetupProvider.PrepareInitializer(b.parentCtx, opts)
```
dshulyak: sorry, i am not very familiar with codepath in post setup provider. i will need to read through to understand if moving this function is ok, about the same as you basically

i thought that it is possible to extract side-effect free validation and run it before background work, and then temporarily panic if background works fails for other reasons

piersy: Hi @dshulyak we could try to extract side effect-free validation but whatever we extract it seem's like it won't catch as many problems as just calling `PrepareInitializer` because `PrepareInitializer` is doing the actual work.

Also doing the extraction will take longer and @pigmej mentioned that he'd like this fix "fast"

So yes it's possible to do the extraction, but I'm not sure it would provide any benefit.

Also what do you mean by "_temporarily_ panic"?

pigmej: I think that if I want smeshing and I cannot, we should quit asap because otherwise, the user may have false assumptions. 

Obviously, if it can be done in a pretty way, then we can :) Just having an ugly fix now is better than not at all IMHO.

piersy:
bors try

activation/activation.go
dshulyak: sounds good, if this approach doesn't cause other issues. for example calling PrepareInitializer several times and similar simple things

piersy: My current understanding is that `PrepareInitializer` sets some state in the `PostSetupManager` and calling it again simply overwrites that state. So I don't think multiple calls to it will pose a problem.

piersy: But I realised there is now a race condition with `StartSession`, i need to think about this a bit more

piersy:
bors try

activation/post.go
```diff
@@ -223,55 +246,49 @@ func (mgr *PostSetupManager) StartSession(ctx context.Context, opts PostSetupOpt
...
 	mgr.state = PostSetupStateComplete
 	return nil
 }
 
-func (mgr *PostSetupManager) prepareInitializer(ctx context.Context, opts PostSetupOpts) error {
+// InitializerConfig builds the config to be used by initialization in calls to
+// StartSession. Having this function be separate from StartSession provides a
+// means to understand if the post configuration is valid before kicking off a
+// very long running task (StartSession can take hours to complete).
+func (mgr *PostSetupManager) InitializerConfig(ctx context.Context, opts PostSetupOpts) (*InitializationConfig, error) {
```
poszu, line 262: I find the name of this method a little misleading. It not only builds the config for the initializer. It creates the initializer.

poszu (approved):
Why not just export `prepareInitializer()` method and require it to be called explicitly? Is it to enforce statically that `InitializerConfig` is actually called (by taking its return value as input arg)?

A sidenote about:
> InitializerConfig doesn't modify any state in the PostSetupManager, which makes it safe to call concurrently while initialization is occurring.

The sole fact it is a read-only method doesn't make it thread-safe. If another thread modifies a part of `PostSetupManager` that this method reads concurrently, it's a data race anyway.

pigmej (approved):
Functionality looks ok.

activation/activation.go
```diff
@@ -190,8 +196,10 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 			// ensure we are ATX synced before starting the PoST Session
 		}
 
-		if err := b.postSetupProvider.StartSession(ctx, opts); err != nil {
-			return err
+		// If start session returns any error other than context.Canceled
+		// (which is how we signal it to stop) then we panic.
+		if err := b.postSetupProvider.StartSession(ctx, cfg); err != nil && !errors.Is(err, context.Canceled) {
+			panic(fmt.Sprintf("initialization failed: %v", err))
```
fasmat, line 202: Why panic instead of returning the error? If this is caused by a GRPC request the error that caused it won't be propagated to smapp and other running go-routines will forcefully be shut down as well?

activation/interface.go
```diff
@@ -58,7 +58,12 @@ type postSetupProvider interface {
 	Status() *PostSetupStatus
 	ComputeProviders() []PostSetupComputeProvider
 	Benchmark(p PostSetupComputeProvider) (int, error)
-	StartSession(context context.Context, opts PostSetupOpts) error
+	// InitializerConfig builds the config to be used by initialization in calls to
+	// StartSession. Having this function be separate from StartSession provides a
+	// means to understand if the post configuration is valid before kicking off a
+	// very long running task (StartSession can take hours to complete).
```
fasmat, line 64: These interfaces are private (and primarily exist to be mocked in tests). No need to document the method again if the actual implementation already has a godoc 🙂

activation/post.go
```diff
@@ -46,6 +46,15 @@ type PostSetupOpts struct {
 	ComputeBatchSize  uint64              `mapstructure:"smeshing-opts-compute-batch-size"`
 }
 
+// InitializationConfig serves as a holder for config generated in
+// InitializerConfig to be passed to StartSession. It is not intended to be
+// user configurable.
+type InitializationConfig struct {
+	init            *initialization.Initializer
+	opts            PostSetupOpts
+	commitmentAtxId types.ATXID
```
fasmat, line 55: I don't think the `commitmentAtxId` should be treated differently form the other `PostSetupOpts`

activation/post.go
```diff
@@ -46,6 +46,15 @@ type PostSetupOpts struct {
 	ComputeBatchSize  uint64              `mapstructure:"smeshing-opts-compute-batch-size"`
 }
 
+// InitializationConfig serves as a holder for config generated in
+// InitializerConfig to be passed to StartSession. It is not intended to be
+// user configurable.
+type InitializationConfig struct {
+	init            *initialization.Initializer
```
fasmat, line 53: What is the reasoning behind the config holding an instance of the `Initializer`?

piersy:
Hey @poszu

>Why not just export prepareInitializer() method and require it to be called explicitly? Is it to enforce statically that InitializerConfig is actually called (by taking its return value as input arg)?

I tried this initially, but the problem was that splitting `prepareInitializer()` and `StartSession` meant that I now would need to track if the session was running to prevent multiple goroutines running it and also enforce that `prepareInitializer()` could not be called while a session was running and as you point out also enforce that you can't start a session without first calling `prepareInitializer()`. It just seemed like a safer and simpler approach to make `prepareInitializer()` read only and pass the config to StartSession.

>>InitializerConfig doesn't modify any state in the PostSetupManager, which makes it safe to call concurrently while initialization is occurring.

>The sole fact it is a read-only method doesn't make it thread-safe. If another thread modifies a part of PostSetupManager that this method reads concurrently, it's a data race anyway.

I wasn't trying to say that it's generally thread safe, just that it is safe to call while StartSession is executing, maybe I should change initialization to StartSession, or let me know if you have an idea about how to word the comment?

activation/post.go
```diff
@@ -223,55 +246,49 @@ func (mgr *PostSetupManager) StartSession(ctx context.Context, opts PostSetupOpt
...
 	)
 	mgr.state = PostSetupStateComplete
 	return nil
 }
 
-func (mgr *PostSetupManager) prepareInitializer(ctx context.Context, opts PostSetupOpts) error {
+// InitializerConfig builds the config to be used by initialization in calls to
+// StartSession. Having this function be separate from StartSession provides a
+// means to understand if the post configuration is valid before kicking off a
+// very long running task (StartSession can take hours to complete).
```
fasmat, line 261: I don't understand this reasoning. If before the post configuration was invalid `prepareInitializer` would return an error and `StartSession` would stop immediately. 

Can you explain the reasoning as to why this has now to be called explicitly before calling `StartSession`?

activation/post_test.go
```diff
@@ -49,31 +49,47 @@ func TestPostSetupManager(t *testing.T) {
...
+	_, err := mgr.InitializerConfig(ctx, mgr.opts)
+	req.NoError(err)
+
+	cfg := config.DefaultConfig()
+
+	// Check that invalid options return errors
+	opts := mgr.opts
+	opts.NumUnits = cfg.MaxNumUnits + 1
+	_, err = mgr.InitializerConfig(ctx, opts)
+	req.Error(err)
```
fasmat, line 92: Could `InitializerConfig` be renamed to `ValidatePostOpts` or similar and just return an error if the provided Options are invalid?

I feel like the reason it returns a new object that holds additional data and a new instance of the `Initializer` exposes too much of the internals of the `PostSetupManager`.

I think the logic could be simplified if `commitmentAtx` would just be a part of `PostSetupOpts` that 1) if set gets checked to be a valid ATX and 2) if not set defaults to the logic how it is selected at the moment

activation/post.go
fasmat: I see it similarly. I think this method could just be a "validate" method that only returns an error if the provided `PostSetupOpts` are invalid.

Instantiating a new initializer imo is part of `StartSession` and the CommitmentATX can just be added to the `PostSetupOpts` (and is selected like now if not set explicitly - imo there is nothing wrong with giving the user the option to choose a commitmentATX of their liking).

activation/post.go
piersy: Previously with `prepareInitializer` being called in `StartSession` and since`StartSession` can take days to complete we would call it in a go-routine, so even if it stopped immediately any errors would not be available to the calling code that initiated the routine. The calling code could wait for the result from the go-routine but we wouldn't want to make the API call take day's to return in the case that initialization was successful.

So then the only option I see is to wait for "some" time and if we don't get a failure from the routine in that time, assume that initialization has started successfully. But the problem with that is:
* It's going to be slow, because we'll need to choose a wait time long enough that all machines have completed executing `prepareInitializer` before the time expires.
* It's not reliable since we can't guarantee that in all cases machines will complete running `prepareInitializer` before the time expires.

Alternatively, and what I think is the best option is that we change the api to return a stream that callers listen to, but that I guess is a much bigger change, since we'd need to update clients.

fasmat: > so even if it stopped immediately any errors would not be available to the calling code that initiated the routine.

I can follow that reasoning, but I think a synchronous "ValidatePostOps" method that returns a non-nil error if the provided options are incorrect would solve that issue. In my opinion this method should not return a new instance of initializer and `commitmentAtx` can be part of the `PostSetupOpts`. It would make perfect sense to do that now, with the changes to how it is selected and the fact that any existing ATX is now valid to be chosen for the commitment I don't see the need to treat it differently from all the other options.

> Alternatively, and what I think is the best option is that we change the api to return a stream that callers listen to, but that I guess is a much bigger change, since we'd need to update clients.

This actually pretty much how our API works at the moment 🙂 -> see `SmesherService::PostSetupStatusStream` in `go-spacemesh/api/grpcserver/smesher_service.go`

activation/post.go
piersy: Its tricky. `InitializationConfig` does contain an initializer but it also has the atx id and opts.So if I called it an initializer you could argue that it's not because it has options as well.

It would be nice if I didn't have to update state in the builder and it could just be the initializer, but there seem to be a number of methods that access the bits of state in the builder that are updated.

What would you name it?

activation/activation.go
pigmej: It's not via API. If you have post data in the directory and you configure go-spacemesh /wrongly/ then it will silently exit currently.

We don't want that. ;) The correct fix is for later ;)

activation/interface.go
piersy: Right, thanks :+1:

activation/post.go
piersy: The `Initializer` constructor has many paths that return errors, so if we wan't to get the most possible validation up front, then we should call the constructor as part of validation. Once we have an instance of the `Initializer` it makes sense to use it, instead of duplicating the code to build it elsewhere.

activation/post.go
piersy: I don't have a full understanding of what you are suggesting with regards to the `commitmentAtx` but that change seems to be out of scope with regards to not failing silently.

That aside the reason for returning an intializer is explained here - https://github.com/spacemeshos/go-spacemesh/pull/4319/files/b14c55ae32e6de102c014cb0f164ededec95828f#r1175539280

>>Alternatively, and what I think is the best option is that we change the api to return a stream that callers listen to, but that I guess is a much bigger change, since we'd need to update clients.

>This actually pretty much how our API works at the moment slightly_smiling_face -> see SmesherService::PostSetupStatusStream in go-spacemesh/api/grpcserver/smesher_service.go

Hehe, lol. I was not aware of the `PostSetupStatusStream` but i did initially mention the approach of passing all the feedback through a stream to the caller, however @dshulyak suggested going for the validate up front approach, so that is the approach I took.

activation/post.go
piersy: Also see - https://github.com/spacemeshos/go-spacemesh/pull/4319/files/b14c55ae32e6de102c014cb0f164ededec95828f#r1175539280 for an explanation of why we have an instance of an initializer.

activation/post.go
fasmat: This is true, but calling the initiailzer twice with the same parameters shouldn't return different results in terms of validation.

The reason is that I aim to eventually retire the Initializer service for just an Initialize function. The Initialization service has no real state; the process just needs to be able to be cancelled. The function would be no-op (or return a known error like `ErrAlreadyInitialized`) when called and the initialization is already done.

Alternatively you can call this `PostSession` and remove the `PostSetupOpts`, they are afaik only needed to instantiate the Initialization Service.

activation/post.go
fasmat: Since this might be a bigger change we can create a ticket for that and work on it in a separate PR.

activation/activation.go
fasmat: I still feel like `panic` is the wrong reaction here. How about `log.Fatal`?

activation/post_test.go
piersy: The reason for returning the initializer instance is explained here - https://github.com/spacemeshos/go-spacemesh/pull/4319#discussion_r1175539280

>I feel like the reason it returns a new object that holds additional data and a new instance of the Initializer exposes too much of the internals of the PostSetupManager

I'm not sure what you mean by exposing the internals, the struct is exported but the fields of the struct are not?

> I think the logic could be simplified if commitmentAtx would just be a part of PostSetupOpts that 1) if set gets checked to be a valid ATX and 2) if not set defaults to the logic how it is selected at the moment.

This change sounds like it's not related to not failing loudly, so my inclination would be to not include it in this PR.

From my current, and probably limited understanding of atxs and selecting commitmenbtAtxs, I don't see a reason to make the change, as it would mean that there would be now multiple places in the code where a commitmentAtx is chosen, and optional config fields tend to complicate systems because you now have multiple different ways of calling/configuring components.

activation/post.go
fasmat: > however @dshulyak suggested going for the validate up front approach, so that is the approach I took.

I'm not opposed to this approach. I'm just nitpicking the implementation 🙂.

I would expect a method call to validate the smeshing config in `cmd/node/node.go` that `log.Fatals` if the provided config is invalid (somewhere around here: https://github.com/spacemeshos/go-spacemesh/blob/develop/cmd/node/node.go#L816-L823).

activation/activation.go
piersy: I'm not a fan of `log.Fatal` since you lose the context of where in the stack your failure occurred, which in these critical failures can be pretty useful. However `log.Panic` seems like a sensible change, as we'll now log any extra contextual information that was added to the log.

activation/post_test.go
fasmat: > This change sounds like it's not related to not failing loudly, so my inclination would be to not include it in this PR.

I agree that this is out of scope for this PR.

> ... there would be now multiple places in the code where a commitmentAtx is chosen, and optional config fields tend to complicate systems because you now have multiple different ways of calling/configuring components.

The `PostSetupOpts` contain client side configuration of the PoST the node uses. This includes the nodes public key, the number of space units to use, the directory the data is stored in aso. Since a node is free to choose any ATX as their commitment it should imo be part of these options too.

The selected commitment ATX is even stored along the other options in the `post_metadata.json` file and if the operator chooses to initialize using our Post CLI has to be set explicitly already.

Just because we have code that selects a commitment ATX if the operator doesn't explicitly choose one, should not cause it to be handled differently from other configuration. This behaviour is the same for other parameters too: eg. the post directory can be set explicitly by the operator, but if not the node will select the user directory (based on the OS the node is running on).

activation/activation.go
piersy: However actually when it comes to testing just using panic can be quite useful. Using `log.Fatal ` can be quite confusing if logging is not enabled for the test, as you have previously discovered - https://spacemesh.slack.com/archives/C89QJJY3G/p1679421425885569

activation/post_test.go
fasmat: > I'm not sure what you mean by exposing the internals, the struct is exported but the fields of the struct are not?

I think it would make sense to call the `InitializerConfig` method `NewSession`, the returned objects type `PostSession` and (if possible) drop the `commitmentAtx` and `PostSetupOpts` from it (they are already part of the `Initializer` and as far as I can tell are only needed to log them when `StartSession` is called.

activation/activation.go
fasmat: The instance of `log` that is used to `Fatal` or `Panic` should already contain a lot of context that is necessary to debug an issue; the request ID, the service it was called in, relevant state of the service, etc.

I don't mind using `log.Panic` over `log.Fatal` but `panic` just prints unstructured text that might be more difficult to parse for Grafana/Loki.

piersy: `log.Panic` it is then, although we don't seem to have any consistency around this point currently. I see a lot of plain calls to panic and also lots to `log.Panic` so maybe it would be worth trying to decide on an approach to use consistently.

activation/post_test.go
piersy: The naming sounds sensible to me. But the `PostSetupOpts` are needed because `PostSetupManger.lastOpts` are set to them and the `commitmentAtxId` is needed because `PostSetupManager.commitmentAtxId` is set to it. If we update those fields in `IntializerConfig/NewSession` then we need to go for a more complicated locking model to ensure that they can't be modified while `StartSession` is running.

activation/post.go
piersy: >The reason is that I aim to eventually retire the Initializer service for just an Initialize function. The Initialization service has no real state; the process just needs to be able to be cancelled. The function would be no-op (or return a known error like ErrAlreadyInitialized) when called and the initialization is already done.

How were you planning to return the errors from the function and feed them back to the user?

Regarding `PostSetupOpts` they are needed see - https://github.com/spacemeshos/go-spacemesh/pull/4319#discussion_r1175753970

`NewInitializer` opens files and saves files so it's not guaranteed to succeed if called again with the same params.

activation/post_test.go
piersy: The way I think about the `InitializationConfig/PostSession` is that its a black box, its just a means of neatly ensuring that validation is peformed on the opts and that you have sequential calls to first configure the initialization process and then to run the intialization process.

It's not really important what is in it because its all just an implementation detail of the `PostSetupManager` and in fact we can make `InitializationConfig/PostSession` unexported. I only didn't because the the methods of PostSetupManager are all exported, but now I realise that is only so that they can be mocked.

fasmat: > the PostSetupOpts are needed because `PostSetupManger.lastOpts` are set to them and the `commitmentAtxId` is needed because `PostSetupManager.commitmentAtxId` is set to it

That's true and I guess we have to leave it like this for now, because changing it is out of scope of this PR. My thoughts on it just for future reference:

- `commitmentAtxId` besides logging is only used in `GenerateProof` where it is only passed in to check if the post metadata contains the same `commitmentAtxId`. It is not needed to generate a proof - only for initialization and the same check already happens when we call `StartSession` (indirectly via `NewInitializer`).
- `lastOps` is also only needed in 2 places - to get the size of the PoST during proof generation in `activation/activation.go` and as part of the return value of `Status()`
  - the first one has to be changed - the `GenerateProof` method should return the number of SU it used to generate the proof. That number might be less than the value used during initialization, because we lost data since then (due to disk failures), or it might be more (if we want to allow PoST size increases while the node is running).
  - The second one needs to be thought through again - the data in `lastOpts` is mostly only needed during initialization and is not really valid afterwards. Is there even a need to return network parameters like `ScryptParams` or initialization parameters (like the device used to initialize the data, the size of the batches used during initialization and if we throttle during initialization) as part of the PostStatus? If yes should those still be returned after initialization finished? At the moment we don't even return the values used during initialization, but the ones `StartSession` were called with, which might be different if the node was restarted at some point.

fasmat: Added ticket to fix in future here: https://github.com/spacemeshos/go-spacemesh/issues/4336

activation/post.go
fasmat: Regarding naming: I would prefer `NewSession` that returns a `PostSession` that is then passed to `StartSession` 🙂

activation/post.go
fasmat: Added ticket to fix in future here: https://github.com/spacemeshos/go-spacemesh/issues/4336

activation/post.go
fasmat: The current solution seems to be functionally equivalent to what I proposed in my previous comment so you can ignore it 🙂

activation/post_test.go
piersy: Thanks

piersy:
Closed in favour of #4338

PR #4327, opened 2023-04-20
[Merged by Bors] - activation: reasonable retries for communication with poet

dshulyak:
cycle gap wasn't meant to be spent on requests to poet, it has somewhat different considerations when configured.
automatic retries are still useful, but we can set them to smaller values to tolerate common http problems. if they are not sufficient, we will retry using outer atx builder loop


poszu (approved):

dshulyak:
bors merge

PR #4333, opened 2023-04-24
[Merged by Bors] - tortoise: add voting targets from encoded votes

### LINKED ISSUE #3608: tortoise: count votes for multiple block targets
depends on: https://github.com/spacemeshos/go-spacemesh/issues/3607

Each distinct combination of block id, layer id, tick height is a different voting target and weight is counted separately for every such target. The reason for doing so is because we don't know what is the "real" layer id and height is before we will fetch the block, but we want to fetch only blocks that are upvoted by majority.

- [ ] implementation plan. should include efficient way to represent and count votes for multiple targets

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3608

- there is a new `data` boolean on blockInfo in tortoise state. it is set to true on the direct call to OnBlock to tortoise. tortoise will always vote against the block, if the data is not available locally.
- tortoise.Results provides api to query consensus state from a given layer, and returns results in-between (from, verified]. verified is maintained internally. if Data is set to false, with Valid - true, caller is expected to try to fetch data from peers, and then call tortoise.OnBlock
- voting targets are also added directly from votes in the ballot. in normal "online" case voting target will be always added together with data with a call to tortoise.OnBlock

### DIFF
--- a/tortoise/algorithm.go
+++ b/tortoise/algorithm.go
@@ -223,6 +223,10 @@ func (t *Tortoise) OnBallot(ballot *types.Ballot) {
 type DecodedBallot struct {
 	*types.Ballot
 	info *ballotInfo
+	// after validation is finished we need to add new vote targets
+	// for tortoise from the decoded votes. minHint identifies the boundary
+	// until which we have to scan.
+	minHint types.LayerID
 }
 
 // DecodeBallot decodes ballot if it wasn't processed earlier.
@@ -231,7 +235,7 @@ func (t *Tortoise) DecodeBallot(ballot *types.Ballot) (*DecodedBallot, error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	waitBallotDuration.Observe(float64(time.Since(start).Nanoseconds()))
-	info, err := t.trtl.decodeBallot(ballot)
+	info, min, err := t.trtl.decodeBallot(ballot)
 	if err != nil {
 		errorsCounter.Inc()
 		return nil, err
@@ -247,7 +251,7 @@ func (t *Tortoise) DecodeBallot(ballot *types.Ballot) (*DecodedBallot, error) {
 			info.opinion().ShortString(), ballot.OpinionHash.ShortString(), ballot.ID(),
 		)
 	}
-	return &DecodedBallot{Ballot: ballot, info: info}, nil
+	return &DecodedBallot{Ballot: ballot, info: info, minHint: min}, nil
 }
 
 // StoreBallot stores previously decoded ballot.
@@ -259,10 +263,7 @@ func (t *Tortoise) StoreBallot(decoded *DecodedBallot) error {
 	if decoded.IsMalicious() {
 		decoded.info.malicious = true
 	}
-	if err := t.trtl.storeBallot(decoded.info); err != nil {
-		errorsCounter.Inc()
-		return err
-	}
+	t.trtl.storeBallot(decoded.info, decoded.minHint)
 	return nil
 }
 
--- a/tortoise/algorithm.go
+++ b/tortoise/algorithm.go
@@ -296,3 +296,43 @@ func (t *Tortoise) GetMissingActiveSet(epoch types.EpochID, atxs []types.ATXID) 
 	}
 	return missing
 }
+
+// Results returns layers that crossed threshold in range (from, verified].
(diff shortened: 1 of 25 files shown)


tortoise/algorithm.go
```diff
@@ -298,3 +296,43 @@ func (t *Tortoise) GetMissingActiveSet(epoch types.EpochID, atxs []types.ATXID)
 	}
 	return missing
 }
+
+// Results returns layers that crossed threshold in range (from, verified].
```
countvonzero, line 300: not intuitive that `from` is not inclusive. is there a particular reason why it cannot be [from, verified]?
do you expect caller to remember the last verified and so they do (lastVerified, verified]?

tortoise/state.go
```diff
@@ -365,3 +373,104 @@ func sortBlocks(blocks []*blockInfo) {
...
+	for _, header := range exceptions.Against {
+		from = minLayer(from, header.LayerID)
+		layerdiff, exist := diff[header.LayerID]
+		if !exist {
+			layerdiff = map[types.BlockID]headerWithSign{}
+			diff[header.LayerID] = layerdiff
+		}
+		existing, exist := layerdiff[header.ID]
+		if exist {
+			return votes{}, 0, fmt.Errorf("conflicting votes on the same id with different heights (%d, %d)", existing.header.Height, header.Height)
```
countvonzero: does it make sense to include header.ID and header.LayerID in the error?

countvonzero: but how do you determine the two votes are different height?
i can only tell that this block ID is included twice in the against list.

tortoise/state.go
```diff
@@ -365,3 +373,104 @@ func sortBlocks(blocks []*blockInfo) {
...
+		layerdiff[header.ID] = headerWithSign{header, against}
+	}
+	for _, header := range exceptions.Support {
+		from = minLayer(from, header.LayerID)
+		layerdiff, exist := diff[header.LayerID]
+		if !exist {
+			layerdiff = map[types.BlockID]headerWithSign{}
+			diff[header.LayerID] = layerdiff
+		}
+		layerdiff[header.ID] = headerWithSign{header, support}
```
countvonzero, line 442: sorry i know this is not new code.
can you explain why does the code care about duplicate in the Against list but not in the Support/Abstain list

tortoise/state.go
```diff
@@ -365,3 +373,104 @@ func sortBlocks(blocks []*blockInfo) {
...
+	}
+	for _, lid := range exceptions.Abstain {
+		from = minLayer(from, lid)
+		_, exist := diff[lid]
+		if !exist {
+			diff[lid] = map[types.BlockID]headerWithSign{}
+		}
+	}
+	if from <= evicted {
```
countvonzero, line 453: why not check this condition at the beginning of the method?

tortoise/tortoise.go
```diff
@@ -556,33 +545,27 @@ func (t *turtle) loadBallots(lid types.LayerID) error {
...
-		return nil
-	}
-	t.logger.With().Debug("on block", log.Inline(block))
-	binfo := &blockInfo{
-		id:     block.ID(),
-		layer:  block.LayerIndex,
-		hare:   neutral,
-		height: block.TickHeight,
+func (t *turtle) onBlock(header types.Vote) {
+	if !header.LayerID.After(t.evicted) {
```
countvonzero: use `<` ?

i just realized i can do that recently. so happy.

countvonzero (approved):

tortoise/algorithm.go
dshulyak: i changed that in the followup pr to select an inclusive range

tortoise/state.go
dshulyak: so it is either multiple of the same, or same id / different height. both are not good

tortoise/state.go
dshulyak: it can be updated in any of the loops inside the func. basically it is the minimum from base ballot or any vote

tortoise/state.go
dshulyak: support overwrites any previous supported block with same id in the same layer, so i thought that it is ok to keep same semantics for single encoded ballot. but actually there are only upsides if we drop ballot with such votes. will make a change, thanks

dshulyak:
bors merge

PR #4334, opened 2023-04-24
[Merged by Bors] - Bump post to v0.5.5

poszu:
## Motivation
Bumps post to use the new 8/56 proving algo

### DIFF
--- a/activation/post.go
+++ b/activation/post.go
@@ -57,7 +57,7 @@ type PostProvingOpts struct {
 func DefaultPostProvingOpts() PostProvingOpts {
 	return PostProvingOpts{
 		Threads: 1,
-		Nonces:  20,
+		Nonces:  16,
 	}
 }
 
--- a/Makefile-libs.Inc
+++ b/Makefile-libs.Inc
@@ -94,7 +94,7 @@ else
   endif
 endif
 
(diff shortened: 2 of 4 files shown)


poszu:
Bors try

fasmat (approved):

activation/post.go
```diff
@@ -57,7 +57,7 @@ type PostProvingOpts struct {
 func DefaultPostProvingOpts() PostProvingOpts {
 	return PostProvingOpts{
 		Threads: 1,
-		Nonces:  20,
+		Nonces:  16,
```
fasmat, line 60: I'm failing to understand why this parameter (that can be arbitrarily chosen by the operator of a node) influences the values we use for K1, K2, K3. Maybe you can enlighten me? 🙂

pigmej: the default is "expected nonces" / "sensible amount of nonces" Kx affects those expected nonces (or the other way around if you look from the other side).

poszu:
Bors merge

poszu:
Bors merge

pigmej:
@brusherru that means that the tool to check the performance have to go with nonces in 16 increments, 16,32,48 and soo on.

cc @poszu

PR #4338, opened 2023-04-25
[Merged by Bors] - Ensure start smeshing cannot fail silently (take 2)

### LINKED ISSUE #4292: activation: StartSmeshing should not fail silently

piersy:
## Motivation

Closes https://github.com/spacemeshos/go-spacemesh/issues/4292

## Changes

This change moves the  `prepareInitializer` method of `PostSetupManager.StartSession` and exports it, putting in place the requirement to call `PrepareInitializer` before calling `StartSession`. This provides a means for POST opts to be validated synchronously before starting the long running task of `StartSession` which allows the start smeshing api call to directly return an error to the caller if there is a problem with the smeshing options.

Also this change updates `Builder.StartSmeshing` to panic if the smeshing operation fails, this ensures that the node will crash if there is any problem with the smeshing background task. While not optimal in terms of user experience this at least means that users will now be aware of problems with the smeshing process.

Note: This replaces #4319 which suffered from the problem that a call to `StartSmeshing`, while initialization was active would crash the node. Now `PrepareInitializer` acts as a gate and will instead return an error if the node is initializing, which prevents calls to `StartSession` while the node is initializing.

## Test Plan
Updated existing tests and added a test to check that `PrepreInitializer` returns an error if given invalid config and a test to check that `PrepareInitialzer` and `StartSession` return errors if not called in sequence.

I have manually verified that if `StartSession` returns an error the `activation.Builder` will panic, but I can't write an automated test for it since the go routine that calls `StartSession` is initiated by the `activation.Builder` and there's not a way to neatly insert a recover statement there just for tests.

## DevOps Notes

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -163,17 +163,23 @@ func (b *Builder) Smeshing() bool {
 	return b.started.Load()
 }
 
-// StartSmeshing is the main entry point of the atx builder.
-// It runs the main loop of the builder and shouldn't be called more than once.
-// If the post data is incomplete or missing, data creation
-// session will be preceded. Changing of the post potions (e.g., number of labels),
-// after initial setup, is supported.
+// StartSmeshing is the main entry point of the atx builder. It runs the main
+// loop of the builder in a new go-routine and shouldn't be called more than
+// once without calling StopSmeshing in between. If the post data is incomplete
+// or missing, data creation session will be preceded. Changing of the post
+// options (e.g., number of labels), after initial setup, is supported. If data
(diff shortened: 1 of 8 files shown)


fasmat (approved):

fasmat:
bors merge

Issue #3609, opened 2022-10-03
sync: fetch blocks that cross positive threshold in tortoise


dshulyak:
more info: https://community.spacemesh.io/t/deferred-block-fetching-a-k-a-dangling-pointers/309
depends on: https://github.com/spacemeshos/go-spacemesh/issues/3607 https://github.com/spacemeshos/go-spacemesh/issues/3608

- [ ] implementation plan

dshulyak:
see https://github.com/spacemeshos/go-spacemesh/issues/3608#issuecomment-1515859095 for tortoise interface. beside that expected changes in sync should be minimal. syncer should not eagerly ask peers for block. there should be two conditions to ask for a block:
- peer shared a valid certificate within hdist
- tortoise outputs that block is in consensus, but missing data locally

PR #4339, opened 2023-04-26
fetcher: dont ask for blocks

dshulyak:
related to: https://github.com/spacemeshos/go-spacemesh/issues/3609

there is no reason to ask for known blocks, as we are not supposed too query simply based on advertisement.
there is code that queries for a block after downloading certificate, and additionally code will have to query when mesh wants to apply block, but misses data.


dshulyak:
merged into https://github.com/spacemeshos/go-spacemesh/pull/4344

PR #4340, opened 2023-04-26
mesh: refactor comms with tortoise

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/3609


dshulyak:
bors try

PR #4343, opened 2023-04-27
[Merged by Bors] - miner: cache proposal eligibility

countvonzero:


miner/oracle.go
```diff
@@ -89,7 +89,12 @@ func (o *Oracle) GetProposalEligibility(lid types.LayerID, beacon types.Beacon,
 		return nil, fmt.Errorf("failed to get valid atx for node for target epoch %d: %w", epoch, err)
 	}
 
-	return o.calcEligibilityProofs(atx, epoch, beacon, nonce)
+	ee, err := o.calcEligibilityProofs(atx, epoch, beacon, nonce)
+	if err != nil {
+		return nil, err
+	}
+	o.cache = ee
```
dshulyak, line 96: so if it is cached something and fallback needs to be used?

countvonzero: this is not part of the fallback. only beacon and hare active set.

dshulyak: if eligibility computation is using beacon - shouldn't it change as well? otherwise it simply will be delayed by everyone else in the tortoise, and probably ignored in hare all together

dshulyak: so there is actually no good option, if smesher made a reference ballot - it has to stick to the previous beacon and eligibilities. we literally should never overwrite beacon if some value exists in the network, as it will create partition

dshulyak (approved):

miner/oracle.go
countvonzero: it depends on how many nodes in the network has beacon value. if the majority of the network does not have beacon (partition already exists), its still better to create fallback beacon. if the network have N different valid beacon values, then there seems no point meddling.

countvonzero:
bors merge

PR #4344, opened 2023-04-27
[Merged by Bors] - syncer: fetch missing blocks reported by tortoise

### LINKED ISSUE #3609: sync: fetch blocks that cross positive threshold in tortoise

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3609

- peer doesn't ask for blocks as a part of layer info
- instead before applying layer it asks tortoise for blocks that could be applied, but missing data (such as block validated by hare or tortoise). if we fail to fetch it from peers we will continue doing that periodically, until either sync downloads them or tortoise threshold changes
- another important behavior was already implemented. if we received a certificate for a block we will query that certificate

there is an opportunity to simplify mesh code, i will try to followup on that in https://github.com/spacemeshos/go-spacemesh/pull/4340


dshulyak:
bors try

dshulyak:
bors try

mesh/mesh.go
```diff
@@ -316,7 +317,7 @@ func (msh *Mesh) processValidityUpdates(ctx context.Context, logger log.Log, upd
 					if err := blocks.SetValid(dbtx, bid); err != nil {
 						return err
 					}
-				} else if err := blocks.SetInvalid(dbtx, bid); err != nil {
+				} else if err := blocks.SetInvalid(dbtx, bid); err != nil && !errors.Is(err, sql.ErrNotFound) {
```
dshulyak: add a test for this. another edge case with multiple same id targets where only one is valid

countvonzero: if we ignore sql.ErrNotFound here, when is the next chance that the block will be set valid in code?
supposedly mesh only gets validity updates from tortoise once when layer is verified

dshulyak: i refactored this part

dshulyak:
@countvonzero please take a look, i refactored part that updates validity

mesh/mesh.go
```diff
@@ -308,15 +306,12 @@ func (msh *Mesh) processValidityUpdates(ctx context.Context, logger log.Log, upd
...
-					bid, lid, log.Bool("validity", valid))
-				if valid {
-					if err := blocks.SetValid(dbtx, bid); err != nil {
-						return err
-					}
-				} else if err := blocks.SetInvalid(dbtx, bid); err != nil {
+		for i := range results {
+			for _, block := range results[i].Blocks {
+				if !block.Data {
+					continue
```
countvonzero, line 312: i think we should return error here instead of continuing. 

from the code logic, since it errors out when there are missing blocks, this should not happen. 
this is the only place in code that block validity is updated. if indeed there are missing blocks at this point, the block validity will not be set ever if the code continue.

countvonzero (approved):

mesh/mesh.go
dshulyak: we don't ask peers for any block that crossed negative threshold. it is probably one of the main reasons for doing this change

dshulyak:
bors merge

dshulyak:
https://grafana.spacemesh.dev/explore?orgId=1&left=%7B%22datasource%22:%22loki-gke-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22test-xims%5C%22,%20pod%3D%5C%22smesher-14-f696c7f4f-fr6td%5C%22%7D%20%7C%3D%20%5C%22proposalBuilder%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-gke-dev%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%22now-3h%22,%22to%22:%22now%22%7D%7D

image

dshulyak:
image

dshulyak:
bors merge

Issue #4090, opened 2023-02-26
recover from a checkpoint state


countvonzero:
## description

when multiple errors cascade in the network, it's very likely that network stalls to a grind. in situations where mesh data is in a bad state such that tortoise cannot make progress, we want to revert to a genesis-like state and restart tortoise.

## considerations:
- preserve global state (accounts/stubs balances)
- accounts do not need to be re-spawned (via preserving account state)
- prevent replay attack on transactions (via reserving account nonce)
- preserve ATXs so that smeshers do not have to re-initialize the storage (keep the same genesis ID)
- it's ok to reset vesting schedule (the only vesting account is for investors/devs)
- rewards for staying in the network (?)

## proposal
the spacemesh administrator will provide a json file that contains necessary data for the network to recover.
the revived network will use the same genesis ID so that miners can use existing identities and PoST data.

### recovery data
- reboot layer: the first layer the revived network will start running hare protocol
  - tortoise will tally votes starting from this layer
  - syncer and mesh will manage mesh and states from this layer

- active set: the bootsrapping active set at reboot layer

- beacon: the bootstrapping beacon at reboot epoch

- snapshot of all ATXs: this is the most recent ATX of each miner. it serves two main purposes:
  - miners do not need to re-initialize their PoST data
  - used as previous/positioning ATX for building next ATXs in the revived network

  each ATX should contain
    - coinbase (for the first epoch rewards after recovery)
    - space unit (numUnit) (?) 
    - tick height (base tick height + tick count), for new ATX's base tick height
    - vrf nonce (for building new ATX)
    - commitment atx (for building new ATX)
    - publish layer (?)
    - sequence
    - initial post indices (?)

- snapshot of accounts at a known good layer

### a known good layer to recover state from
the spacemesh administrator manually determine a known good layer. (manner TBD)

### dump recovery data
given a known good layer, issue an RPC to a node that does the following
- start a sql transaction, do
- query latest accounts data before reboot layer 
- query latest atxs data before reboot layer
- dump result into json file

the spacemesh administrator should dump recovery data from multiple nodes to verify that
the dumped data are identical.

### load recovery data
node operator download the recovery data from spacemesh website and restart go-spacemesh 
binary with --recovery_data pointing to this file. the node will do the following on startup
- backup the old database
- create new database
- fill the account table with the account snapshot
- fill the atxs table with most recent ATXs from miners
- set effective genesis layer to reboot layer - 1

### implementation
[...]

PR #4345, opened 2023-04-28
[Merged by Bors] - nipost: use file instead of database to persist nipost challenge and builder state

countvonzero:
## Motivation
part of #4090

## Changes
- save nipost challenge and builder state to a file
- remove kvstore table


activation/nipost_state.go
```diff
@@ -0,0 +1,80 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/spacemeshos/go-spacemesh/codec"
+	"github.com/spacemeshos/go-spacemesh/common/types"
+)
+
+const (
+	challengeFilename = "NIPostChallenge"
+	builderFilename   = "NIPostBuilderState"
```
dshulyak: please use better filenames (for example nipost_challenge.bin, nipost_builder_state.bin)

activation/nipost_state.go
```diff
@@ -0,0 +1,80 @@
...
+
+func saveNipostChallenge(dir string, ch *types.NIPostChallenge) error {
+	if _, err := os.Stat(dir); err != nil {
+		return fmt.Errorf("post dir not created: %w", err)
+	}
+	data, err := codec.Encode(ch)
+	if err != nil {
+		return fmt.Errorf("encode nipost challenge: %w", err)
+	}
+	err = os.WriteFile(filepath.Join(dir, challengeFilename), data, ownerReadWrite)
```
dshulyak: as this data is somewhat critical i think we should report success only after fsync and also write checksum that is checked on the load side

write without fsync is not guaranteed to land on disk. or may land partially, in this case there will be undefined behavior on load

countvonzero: thanks. done

countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

PR #4356, opened 2023-05-04
[Merged by Bors] - Verify no panic on StartSmeshing if prepare fails

piersy:
## Motivation

An additional test for #4338 

Currently if there is an error duing a smeshing session, started by StartSession then we panic, but we don't want to panic if a call to PrepareInitializer returns an error, because:

A) That error can be directly returned to callers of the API.
B) Panicking when PrepareInitializer returns an error could cause an existing smeshing session that had not encountered any problems to be abandoned.

This test verifies this behaviour.

## Changes
One extra test.

## Test Plan
The test is passing

## DevOps Notes


fasmat (approved):

piersy:
bors merge

activation/activation_test.go
```diff
@@ -314,6 +314,15 @@ func TestBuilder_RestartSmeshing(t *testing.T) {
 	})
 }
 
+// This test is intended to check that if PrepareInitializer returns an error
+// then that error is returned instead of crashing, which is what should happen if
+// StartSession returns an error.
```
dshulyak: i think such comments are not useful, the code below expresses very clearly what is the test about

piersy: The reason I added this test is to check that we don't crash when StartSmeshing is called and PrepareInitializer returns an error. There's nothing in the code of this test that would allow you to understand that.

Ideally I'd test for a panic when StartSession is called and then verify that if PrepareInitializer returns an error there is no panic, this would then make it clear from the code that PrepareInitializer returning an error ensures that no panic occurs. But we can't test for a panic because StartSession starts its own go-routine to run StartSession.

fasmat: I'm leaning towards @dshulyak opinion.

Here we had a bug/regression that caused a go routine started by `StartSession` to exit early without indicating this to the caller. Ideally we have a test that would have failed without the fix and now passes.

Before it didn't panic it was failing silently. So imo a proper test should return a non-`nil` error on `StartSmeshing` with gomock and assert that `log.Panic` is called (while before it would silently return).

This test is additional to check the new behavior that if `PrepareInit` returns an error, it is also returned by `StartSession` without calling `StartSmeshing`.

I prefer good names for those 2 tests (e.g. `Test_PanicOnErrStartSmeshing` and `Test_PropagateErrOnPrepareInitializer` or similar) and if additional context is necessary just add a comment with the link to the issue that the test belongs to (`// regression test for https://github.com/spacemeshos/go-spacemesh/issues/4292`).

piersy: Ok, so the thing I'm missing is how to assert that  `log.Panic` is called?

dshulyak: by recovering from panic, or there are some nuances? i think there is some helper like require.Panic

fasmat: We have a `log.Logger` interface. `NewBuilder` can take this a parameter instead of the concrete type `log.Log`. Then you an mock the logger for tests. 🙂 

`log.Logger` needs to be extended to include a `Panic` method for this to work, but the change is small enough that imo it can be part of this PR.

fasmat: > by recovering from panic, or there are some nuances? i think there is some helper like require.Panic

I would prefer to mock the logger and asserting this way. Also I'm not sure if `logtest` actually panics when `Panic` is called (imo it shouldn't).

piersy: @dshulyak the nuance is that recover only works to recover from panics emitted by the same goroutine that recover is called in and in this case the builder starts a new goroutine to run `StartSession` and so we cannot recover panics from that.

piersy: @fasmat, I will try mocking the log

activation/activation.go
```diff
@@ -188,6 +194,7 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 
 	b.eg.Go(func() error {
 		defer b.started.Store(false)
+		time.Sleep(time.Second)
```
dshulyak: i think this needs to be removed

activation/activation.go
```diff
@@ -199,7 +206,7 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 		// If start session returns any error other than context.Canceled
 		// (which is how we signal it to stop) then we panic.
 		if err := b.postSetupProvider.StartSession(ctx); err != nil && !errors.Is(err, context.Canceled) {
-			b.log.Panic(fmt.Sprintf("initialization failed: %v", err))
+			b.log.Panic("initialization failed: %v", err)
```
countvonzero, line 207: there is a lot of effort putting in testing this panic behavior...

have you considered adding a streaming API of the StartSmeshing equivalent and only return success after StartSession returns?

activation/activation.go
piersy: Thanks, left over from testing.

activation/activation.go
piersy: Yes, that was one of the approaches I initially considered and still looks to be ultimately the best approach to me, but it was @dshulyak that suggested that I take this route of validating up front. Going down the route of using a streaming API will mean all clients will need to update, I'm not presently aware of how much extra work that would entail, but seems like it coule have been a factor in @dshulyak's decision making.

activation/activation.go
```diff
@@ -118,6 +118,12 @@ func WithPoetConfig(c PoetConfig) BuilderOption {
 	}
 }
 
+func WithLogger(l log.Logger) BuilderOption {
```
fasmat: The builder now receives the `log` instance twice: as a parameter and as a functional option. Can we just use one of the two? I'd prefer `WithLogger` but I'm also OK with not having the functional option and instead just passing it in as a parameter.

piersy: Yeah, that's not great. I think its better to go with passing the param directly, since as my understanding of the functional option pattern is that it is reserved for optional parameters and the log is not optional.

dshulyak: in other places i use log.Nop as a default, thats seems to make it optional

activation/activation_test.go
```diff
@@ -314,6 +315,46 @@ func TestBuilder_RestartSmeshing(t *testing.T) {
...
+	// Stub these methods in case they get called
+	tab.mpost.EXPECT().GenerateProof(gomock.Any(), gomock.Any()).AnyTimes()
+	tab.mclock.EXPECT().AwaitLayer(gomock.Any()).AnyTimes()
+
+	// Set expectations
+	tab.mpost.EXPECT().PrepareInitializer(gomock.Any(),
+		gomock.Any()).Return(errors.New("avoid panic"))
+	// We don't expect this call to be made, but we need StartSession to return
+	// an error if it is called so that l.Panic is then called. If we didn't set this then
+	// the test could pass even if StartSession was called.
```
fasmat: This is not true. A call to a mock that is not `EXPECT()`-ed will cause the test to fail.

activation/activation_test.go
```diff
@@ -314,6 +315,46 @@ func TestBuilder_RestartSmeshing(t *testing.T) {
...
+	tab.mpost.EXPECT().StartSession(gomock.Any()).Return(errors.New("err")).MaxTimes(0)
+	tab.StartSmeshing(tab.coinbase, PostSetupOpts{})
+
+	// We wait a second since the call to l.Panic is happening in a different goroutine
+	time.Sleep(time.Second)
```
fasmat: This is a good use of `gomock.WithContext`:

```golang
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	ctrl, ctx := gomock.WithContext(ctx, t)
	l := log.NewMockLogger(ctrl)
	panicCalled := make(chan struct{})
	l.EXPECT().Panic(gomock.Any(), gomock.Any()).Do(func(format string, args ...any) {
		close(panicCalled)
	})

	// ... additional expectations

	select {
	case <-panicCalled:
		// Success
	case <-ctx.Done():
		require.Fail(t, "test timed out or failed")
	}
```

This will allow the test to pass as soon as `log.Panic` is called (0.02s on my machine) or fail if it isn't after 5 seconds. No need to sleep 🙂. `gomock.WithContext` allows to fail the test quicker, when a different error than a timeout occurs (i.e. unexpected call of a method of a mocked interface)

activation/activation_test.go
piersy: Thanks, I've removed the expectation and updated the comment.

activation/activation_test.go
piersy: Thanks! Although I only see it working for the first part of the tests that verifies there is a panic. In order to verify that there isn't a panic, we just have to wait some time and assume that if there was going to be a panic, it would have already happened. You could call this the _optimistic_ approach :D

fasmat: Now it becomes philosophical 😄, either:

- we write the test knowing how the service works internally; then we don't need to sleep because we know that the `log.Panic` won't be called when `PrepareInitializer` returns an error
- we write the test and treat the service as a black box; then we should make sure that no go routine is started that could call `StartSmeshing` when `PrepareInitializer` returns an error; we can do this by setting the `errGroup` to `nil` for this part of the test or mocking it and asserting that no go-routine was started

While the second is safer, I think the first is OK here - also if someone changes the method to not return when `PrepareInitializer` returns an error and a go routine is started by `StartSmeshing`, go detects this even without the sleep:

```
panic: Fail in goroutine after TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError has completed
```

piersy: I don't think the first approach is a valid approach, the whole point of this test is to verify that if `PrepareInitializer` returns an error then `log.Panic` is not called, if we assume that's already the case then we don't need the test.

The second approach as you describe it, is not blackbox, because you'd need to  manipulate the internal `errGroup` with an understanding that use of the `errGroup` implies that eventually `log.Panic` would be called.

So IMO the blackbox approach would be to either sleep at the end of the test or rely on the go test framework detecting a failure after the test has completed.

I prefer the sleep (_even though I really don't like having to use sleep in tests_) because:
*  it provides a much clearer error ( relying on the go test framework simply provides a stack trace which lets you know there was some problem with gomock)
* 1 second is not a hugely significant amount of time to wait when our tests take more than 5 minutes to run.
* The test process can finish before a fail in a go-routine occurs and in that case the failure goes undetected(since it doesn't actually happen).

It's not perfect but the sleep is like a best effort with the go test framework providing some backup in case the sleep is not long enough.

dshulyak: in my opinion sleeps in tests should not be used at all, relying on sleep is just bad code, regardless of reasons

activation/activation.go
piersy: @fasmat & @dshulyak I removed the `WithLogger` option but that was before @dshulyak suggested using log.Nop, so If you think It's worth it I can switch back to using `WithLogger` with  log.Nop ?

fasmat (approved):

activation/activation_test.go
piersy: Upon further discussion @fasmat still preferred to remove the sleep and have a faster test, so that is what I have done.

piersy:
bors merge

activation/activation_test.go
piersy: @dshulyak In this case what would you do then?

piersy:
bors merge

dshulyak:
seems to be stuck in code coverage

piersy:
bors merge

piersy:
@dshulyak 

> seems to be stuck in code coverage

I've seen that before, I don't think its due to any changes in this PR though, it seems to happen randomly.

PR #4362, opened 2023-05-05
[Merged by Bors] - tortoise: use rlock for read-only methods and meter store

dshulyak:
should enable parallel ballot decoding

### DIFF
--- a/tortoise/algorithm.go
+++ b/tortoise/algorithm.go
@@ -45,7 +45,7 @@ type Tortoise struct {
 	ctx    context.Context
 	cfg    Config
 
-	mu   sync.Mutex
+	mu   sync.RWMutex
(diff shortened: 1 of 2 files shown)


countvonzero (approved):
curious what profiling data you had that trigger this change.

dshulyak:
> curious what profiling data you had that trigger this change.

i didn't profile. this is just a right thing to do. this is on critical code path, and there is some amount of computation (hashing, decoding votes) so holding mutex is a mistake

dshulyak:
bors merge

dshulyak:
https://github.com/spacemeshos/go-spacemesh/issues/4367

dshulyak:
bors merge

PR #4365, opened 2023-05-05
[Merged by Bors] - bootstrapper: allow to bootstrap a specific epoch

countvonzero:
## Motivation
part of #4090

## Changes
for systest
- allow the server binary to bootstrap any epoch (not just epoch 2)
  in the checkpoint recovery case, we will checkpoint/recover in the middle of epoch 5, we then need to bootstrap 
  - the remainder of epoch 5: so hare can start working immediately at the restore layer
  - epoch 6: the result of beacon protocol in epoch 5 is not persisted across restart, if generated.
 
- change bootstrap active set to be the complete active set (instead of partial 3/4)
- add retry to query network parameters

### DIFF
--- a/bootstrap/schema.json
+++ b/bootstrap/schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://spacemesh.io/checkpoint.schema.json.1.0",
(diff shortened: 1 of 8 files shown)


dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #4366, opened 2023-05-05
[Merged by Bors] - syncer: sync atxs at the end of epoch

### LINKED ISSUE #4284: sync: download activations from peers at the end of publish epoch
if node missed gossip window it will download atx's only at the start of next epoch. downloading them sooner doesn't change anything for newly joined node, as it will also miss beacon protocol and will have to wait for some number of layers.

but if node computed beacon, and got disconnected for short period of time  - it will very likely miss atleast some activations. it will result in skewed tortoise active set, and soon will also affect voting on proposals (activations grading). it would be better to download activations based on the publish epoch, rather than the start of target epoch.

we should set a layer at which we should ask for atxs, such that they still will be considered good based on grading scheme, but after close enough to the end of the epoch.

countvonzero:
## Motivation
Closes #4284

## Changes
- sync atxs at the end of each epoch, where the end of epoch is configured by param

- in fetcher peers cache, exclude own peer ID from the cache.
Screenshot from 2023-05-03 23-10-25

- remove unused param SyncRequestTimeout

- add metrics for syncer/fetch

- use single instance of logger


countvonzero:
bors try

cmd/node/node.go
```diff
@@ -814,12 +814,45 @@ func (app *App) listenToUpdates(ctx context.Context, appErr chan error) {
...
 
-func (app *App) startServices(ctx context.Context, appErr chan error) error {
-	if err := app.fetcher.Start(); err != nil {
-		return fmt.Errorf("failed to start fetcher: %w", err)
+func (app *App) waitConnectedPeers(ctx context.Context) error {
+	want := uint64(app.Config.P2P.MinPeers)
+	ready := func() bool {
+		return app.host.PeerCount() >= want
+	}
+	timeout := time.Now().Add(10 * time.Minute)
```
dshulyak: context.WithTimeout

cmd/node/node.go
```diff
@@ -814,12 +814,45 @@ func (app *App) listenToUpdates(ctx context.Context, appErr chan error) {
...
+	wait := 10 * time.Second
+	for {
+		if ready() {
```
dshulyak: why not to inline `app.host.PeerCount() >= want` ? closure is not more readable than a single line condition

cmd/node/node.go
```diff
@@ -814,12 +814,45 @@ func (app *App) listenToUpdates(ctx context.Context, appErr chan error) {
...
+			app.log.With().Info("peers ready",
+				log.Uint64("count", app.host.PeerCount()),
+				log.Uint64("want", want),
+			)
+			return nil
+		}
+		app.log.With().Warning("peers not ready",
```
dshulyak: this should be debug if necessary. there are appropriate metrics that allow to see what is the actual number of peers

cmd/node/node.go
```diff
@@ -814,12 +814,45 @@ func (app *App) listenToUpdates(ctx context.Context, appErr chan error) {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		}
 	}
+}
+
+func (app *App) startServices(ctx context.Context, appErr chan error) error {
 	app.eg.Go(func() error {
-		app.startSyncer(ctx)
+		if err := app.waitConnectedPeers(ctx); err != nil {
```
dshulyak: i don't think that this is good solution for the problem. we already had problems in the past that sync was waiting for 10 peers, while there were less in the network and therefore gossip was stuck...

dshulyak (commented):
i think it is a big mistake to wait for a specific number of peers before starting sync, as gossip also depends on the synced state. we will need to find another solution for a problem that you are facing

dshulyak:
> in fetcher peers cache, exclude own peer ID from the cache.

fetcher should not ask over the network based on the object that it sent itself. if it happens thats a programming mistake

dshulyak:
the problem that is being solved is about fetching atxs that didn't made it into the checkpoint or something else?

i think something along next line is more robust, and doesn't carry a risk that node will be stuck because it has high min peers number. syncer should maintain a boolean per peer to track if atxs were requested. starting at 3/4 (or maybe 4/5) of the epoch we can ask a subset of peers on every tick to share atx headers that target next epoch. for example if sync ticks every 10s, on every tick we can ask 10% of peers to share atxs. once peer shared non empty list -  boolean is set to true and it is not asked again in that epoch. obviously this data structure needs to be cleaned up periodically.

countvonzero:
> > in fetcher peers cache, exclude own peer ID from the cache.
> 
> fetcher should not ask over the network based on the object that it sent itself. if it happens thats a programming mistake

true in a steady state. with 2 caveats:
1. there is already a bug in current code where we call `RegisterPeerHashes` in the gossip handlers. obviously we will call that on the nodes' own peer ID as well, and this fixed that. 

2. when the node's own atx A9 isn't included in the checkpoint data, and A9 references B9 as positioning ATX in the same epoch that is also not included in the checkpoint data. then it will needs to fetch B9 from the network. tho i think i will change the approach by preserving B9 from its previous database instead.

but this fix is still needed for case 1.

countvonzero:
> the problem that is being solved is about fetching atxs that didn't made it into the checkpoint or something else?

yes.
 
> i think something along next line is more robust, and doesn't carry a risk that node will be stuck because it has high min peers number. syncer should maintain a boolean per peer to track if atxs were requested. starting at 3/4 (or maybe 4/5) of the epoch we can ask a subset of peers on every tick to share atx headers that target next epoch. for example if sync ticks every 10s, on every tick we can ask 10% of peers to share atxs. once peer shared non empty list - boolean is set to true and it is not asked again in that epoch. obviously this data structure needs to be cleaned up periodically.

thanks for the suggestion. in this specific case, i actually need the node starts syncing ATX (a form of re-gossip) right away after recovering from checkpoint. so maybe modify the timing to
1. layers btwn checkpoint layer and restore layer
2. starting 4/5 of the epoch


then i can also resolve #4284 as well.

countvonzero:
bors try

countvonzero:
@dshulyak please take another look. updated the PR description as well. thank you

countvonzero:
bors try

fetch/fetch.go
```diff
@@ -612,6 +620,9 @@ func (f *Fetch) getHash(ctx context.Context, hash types.Hash32, h datastore.Hint
 
 // RegisterPeerHashes registers provided peer for a list of hashes.
 func (f *Fetch) RegisterPeerHashes(peer p2p.Peer, hashes []types.Hash32) {
+	if peer == f.host.ID() {
+		return
```
dshulyak, line 624: this should be a programmer mistake. can you please log an .Error ?

syncer/data_fetch.go
```diff
@@ -374,20 +377,55 @@ func (d *DataFetch) receiveOpinions(ctx context.Context, req *opinionRequest, pe
...
 	if len(peers) == 0 {
 		return errNoPeers
 	}
-	peer := peers[rand.Intn(len(peers))]
+	peer := p2p.NoPeer
+	if _, ok := d.atxSynced[epoch]; !ok {
+		d.atxSynced[epoch] = map[p2p.Peer]struct{}{}
+		delete(d.atxSynced, epoch-1)
+	}
+	for _, p := range peers {
```
dshulyak, line 391: asking all peers at once is not optimal, they have to select collection from disk, which is with unknown size. beside that total number of peers may be very high. i suggest to randomly pick subset of peers, for example 5-10% every 10 seconds (interval that you configured)

dshulyak (approved):

fetch/fetch.go
countvonzero: this is actually not a programmer mistake. when we publish an object (atx/proposal), the object go through the gossip handlers, which calls `RegisterPeerHashes()` on the peer.

alternatively i can add a check in those gossip handlers to not call `RegisterPeerHashes` if it's self-publish data. but i figure it's simpler to simply add the check here.

syncer/data_fetch.go
countvonzero: i think there is a misunderstanding. 
the code ask ATX IDs from 1 peer at a time.

this loop breaks as soon as it finds a peer that the node hasn't synced ATX from.
so the last 1/5 of epoch, ~800 layers of the epoch, the node will ask one peer every 10 sec on the atx IDs they know about. that translates to 40K new peers to query ATX IDs published in current epoch.

for mainnet i was thinking about setting the interval to 30 secs.

dshulyak: yes sorry. i was mislead by a loop

fetch/fetch.go
dshulyak: my bad, i thought code is registering that stuff only for data that is not available locally

countvonzero:
bors merge

PR #4368, opened 2023-05-08
[Merged by Bors] - checkpoint: generate checkpoint data

countvonzero:
## Motivation
part of #4090 

## Changes
- add db fields for data that needs to survive across checkpoint so that at the point of restore, these data will be queried on its own without parsing the atx blob

- checkpoint runner that saves the following data in a json file 
  - the account state at snapshot layer
  - the last two atx for each miner

- add checkpoint rpc that return the checkpoint data


api/grpcserver/smesher_service.go
```diff
@@ -19,10 +19,13 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
+type CheckpointRunnerFunc func() CheckpointRunner
```
dshulyak: why it needs to create new instances?

checkpoint/schema.json
```diff
@@ -0,0 +1,103 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://spacemesh.io/checkpoint.schema.json.1.0",
+  "title": "checkpoint data",
+  "description": "checkpoint data for recovery",
+  "type": "object",
+  "required": [
+    "version",
+    "data"
+  ],
```
dshulyak, line 10: just realized that it will be nicer to use protobufs and pbjson

checkpoint/runner.go
```diff
@@ -0,0 +1,154 @@
...
+			Restore:      restoreLayer.Uint32(),
+		},
+	}
+
+	var (
+		atxSnapshot  []atxs.CheckpointAtx
+		acctSnapshot []*types.Account
+		err          error
+	)
+	if err = r.db.WithTx(ctx, func(tx *sql.Tx) error {
```
dshulyak:
```
// WithTx will pass initialized deferred transaction to exec callback.
// Will commit only if error is nil.
func (db *Database) WithTx(ctx context.Context, exec func(*Tx) error) error {
	return db.withTx(ctx, beginImmediate, exec)
}
```

this will acquire write lock. i think better to use read only transaction (e.g begin;)

checkpoint/util.go
```diff
@@ -0,0 +1,53 @@
...
+	if err = json.Unmarshal(data, &v); err != nil {
+		return fmt.Errorf("unmarshal checkpoint data: %w", err)
+	}
+	if err = sch.Validate(v); err != nil {
+		return fmt.Errorf("validate checkpoint data: %w", err)
+	}
+	return nil
+}
+
+func savefile(logger log.Log, fs afero.Fs, dst string, data []byte) error {
```
dshulyak: do you have a ballpark estimate how large snaphot can be? it would probably make more sense to stream data into a file, instead of holding full blob in memory

api/grpcserver/smesher_service.go
```diff
@@ -34,8 +37,20 @@ func (s SmesherService) RegisterService(server *Server) {
 }
 
 // NewSmesherService creates a new grpc service using config data.
-func NewSmesherService(post postSetupProvider, smeshing activation.SmeshingProvider, streamInterval time.Duration, postOpts activation.PostSetupOpts) *SmesherService {
-	return &SmesherService{post, smeshing, streamInterval, postOpts}
+func NewSmesherService(
```
dshulyak: is it a lot of work to introduce this outside of smesher service? doesn't seem to be related

api/grpcserver/smesher_service.go
```diff
@@ -276,3 +291,17 @@ func (s SmesherService) UpdatePoetServers(ctx context.Context, req *pb.UpdatePoe
 	}
 	return nil, status.Errorf(codes.Internal, "failed to update poet server")
 }
+
+func (s SmesherService) Checkpoint(ctx context.Context, req *pb.CheckpointRequest) (*pb.CheckpointResponse, error) {
+	data, err := s.checkpoint().Generate(ctx, types.LayerID(req.SnapshotLayer), types.LayerID(req.RestoreLayer))
```
dshulyak: to be honest when we talked about that i thought that Checkpoint method will stream checkpoint data to a requestor, instead of writing to some local directory

checkpoint/runner.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return r
+}
+
+func (r *Runner) Generate(ctx context.Context, snapshotLayer types.LayerID, restoreLayer types.LayerID) ([]byte, error) {
+	checkpoint := Checkpoint{
+		Version: SchemaVersion,
+		Data: InnerData{
+			CheckpointId: fmt.Sprintf("snapshot-%d-restore-%d", snapshotLayer, restoreLayer),
+			Restore:      restoreLayer.Uint32(),
```
dshulyak: why does it need to be specified when checkpoint is made? i thought this is ortogonal, e.g you can make checkpoint at time A and then decide later when you want to restore it

api/grpcserver/smesher_service.go
countvonzero: yes. this is mainly for systests, as i can only rely on calling fatal/panic to preserve post directory. there needs to be a way for the node to restart (with the same flags) and find the checkpoint file and recover.

the plan for the node operator is to do this manually:
- spacemesh generate a checkpoint file and upload to https://recovery.spacemesh.network
- node operator download that file to some local dir /local/path/to/snapshot-15-restore-18
- node operator restart the node with --checkpoint=file:///local/path/to/snapshot-15-restore-18
- go-spacemesh code copy that file to $dataDir/recovery/ and recover from that file. once recovery is done, backup $dataDir/recovery/ to $dataDir/recovery.$timestamp/
 
 for systest, i need a Recover RPC that does the equivalent of --checkpoint cmd line without redeploying. here is my systest sequence: see code in
https://github.com/countvonzero/go-spacemesh/blob/0dd140106bcb3f75d26bfcc669dd92058d24db26/systest/tests/checkpoint_test.go

- issue checkpoint RPC to every node
- check all the returned checkpoint data are the same
- issue recovery RPC to every node (node will copy the local checkpoint file to $dataDir/recovery/) and restart
- run testSmeshing on them
- add two new nodes
- run testSmeshing on these two nodes

checkpoint/runner.go
countvonzero: it is orthogonal yes, just convenient for systest.

when we deliver this checkpoint data, we can change the restore layer to whatever we want. but since i intend to deliver the checkpoint as a file w. cmd line option (not rpc).

api/grpcserver/smesher_service.go
countvonzero: it was just convenient to put it in the same private API. will create an admin service instead.

checkpoint/util.go
countvonzero: assuming 10K atx and 20K accounts (each holding 5KB state) in the network, we are looking at 
400B/atx and 10KB/account. that will be
8MB (2 atx per miner) +200MB = 208MB

yes. i will change it to streaming.

checkpoint/schema.json
countvonzero: do you know if the encoding is more compact?

api/grpcserver/smesher_service.go
countvonzero: the node doesn't need the checkpoint runner instance in the steady state. 
the RPC interface for checkpoint is mostly for systest purposes.

checkpoint/util.go
dshulyak: account definitely can't be 5kb even in json. maybe around 1kb is max

checkpoint/schema.json
dshulyak: if encoded in binary protobuf, yes it will be more compact. i think in current version all bytes are encoded in hex so thats a double. if base58/base64 thats less than double but still 1.7 maybe

but it probably not worth to change now

api/grpcserver/smesher_service.go
dshulyak: so this is an ephemeral object just to tie together checkpointing logic. maybe it could be just a function

checkpoint/runner.go
```diff
@@ -0,0 +1,154 @@
...
+			BaseTickHeight: catx.BaseTickHeight,
+			TickCount:      catx.TickCount,
+			PublicKey:      hex.EncodeToString(catx.SmesherID.Bytes()),
+			Sequence:       catx.Sequence,
+			Coinbase:       hex.EncodeToString(catx.Coinbase.Bytes()),
+		})
+	}
+	for _, acct := range acctSnapshot {
+		a := Account{
+			Address: hex.EncodeToString(acct.Address.Bytes()),
```
dshulyak: it is also an option to leave encoding up to json, instead of doing hex.EncodeToString manually

checkpoint/runner_test.go
```diff
@@ -0,0 +1,189 @@
...
+		NumUnits:       v.NumUnits,
+		BaseTickHeight: v.BaseTickHeight(),
+		TickCount:      v.TickCount(),
+		PublicKey:      hex.EncodeToString(v.SmesherID.Bytes()),
+		Sequence:       v.Sequence,
+		Coinbase:       hex.EncodeToString(v.Coinbase.Bytes()),
+	}
+}
+
+func TestRunner_Generate(t *testing.T) {
```
dshulyak: i would try to define some table based framefork so that it can be easily extended. in this case inputs are atxs and accounts and a target layer, and output is encoded json, right? so maybe it makes sense to have a table like:


```
{
   "empty"
    nil,
    nil,
    "accounts: [], atxs: []",
}
{
    "only atxs",
    []*types.VerifiedAtx{},
    nil,
    "json with data that we get from atx",
}
``` 

frankly speaking when i see such tests as below i am not even sure what they testing

dshulyak (approved):

checkpoint/runner.go
dshulyak: it will have to be explained to everyone who will generate this thing, i would rather split this two operations, but don't have clear suggestion how exactly

api/grpcserver/smesher_service.go
dshulyak: sounds good. i also thought that Recover can be used to download object from stream and then store it in a directory

api/grpcserver/smesher_service.go
countvonzero: yes. but having an interface at the admin API makes it easier to test the stream RPC.
the call will generate a file on disk, and the RPC will stream the file content to the client.

will try harder in later PR to make these methods pure functional.

checkpoint/schema.json
countvonzero: filed https://github.com/spacemeshos/go-spacemesh/issues/4377

checkpoint/util.go
countvonzero: changed to streaming directly write to a file.
tho not so much streaming. i wasn't able to figure out how to write json's structured data in chunks.
ended up creating a struct that holds all the data, and encode it directly to a file.

please let me know if there is a better approach.

checkpoint/runner.go
countvonzero: the node operator should get this file from the spacemesh url.
for ppl who wants to generate the checkpoint file to verify, they can specify this in the RPC request to create the exact same file.

but will try harder to decouple this in the future. i also think it's best to decouple these two settings.

checkpoint/runner.go
countvonzero: made this change and leave it to json to encode binary data

checkpoint/runner_test.go
countvonzero: tried my best to re-write the test setup. now it listed the atxs and accounts explicitly.
linter now doesn't allow unnamed struct fields.

sql/atxs/atxs.go
```diff
@@ -288,3 +307,47 @@ func GetAtxIDWithMaxHeight(db sql.Executor) (types.ATXID, error) {
...
+	NumUnits       uint32
+	BaseTickHeight uint64
+	TickCount      uint64
+	SmesherID      types.NodeID
+	Sequence       uint64
+	Coinbase       types.Address
+}
+
+// LatestTwo returns the latest two ATXs per smesher.
+func LatestTwo(db sql.Executor) ([]CheckpointAtx, error) {
```
dshulyak: feels a bit too much code-wise, is it tested only in runner? i would test it with variable number of atxs per smesher, e.g some has 2, some 3, some 1

also simpler version is to get latest per identity, and then prev by reference

checkpoint/runner.go
```diff
@@ -0,0 +1,163 @@
...
+			if acct.TemplateAddress != nil {
+				a.Template = acct.TemplateAddress.Bytes()
+			}
+			if acct.State != nil {
+				a.State = acct.State
+			}
+			checkpoint.Data.Accounts = append(checkpoint.Data.Accounts, a)
+		}
+	}
+	if err = tx.Commit(); err != nil {
```
dshulyak: why do you need commit? it shouldn't commit anything

checkpoint/runner.go
```diff
@@ -0,0 +1,163 @@
...
+			Restore:      restore.Uint32(),
+		},
+	}
+
+	tx, err := db.Tx(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("create db tx: %s", err)
+	}
+	defer tx.Release()
+	// using scoped block to signal GC
```
dshulyak: ~if this is for atxSnapshot - it will be on heap, so this optimization will not have any effect.~
actually this would be interesting to doublecheck, i might be wrong, maybe it still allows to release stuff earlier

checkpoint/util.go
dshulyak: yes it is not exactly what i meant, you still hold full checkpoint in memory. json probably will not allocate memory for whole checkpoint. but otherwise it is still not optimal

in order to stream checkpoint you essentially need to write incrementally, not sure if standard codec supports that, so the program should do smth like

```
fmt.Println("{")
fmt.Println('"layer": 10')
fmt.Printn('atxs: [')
for item in atxs {
     fmt.Println(item.JSONString())
}
```

there should be lib that can do that, but i don't know if needed

dshulyak (approved):

checkpoint/util.go
countvonzero: filed https://github.com/spacemeshos/go-spacemesh/issues/4379 to follow up

sql/atxs/atxs.go
countvonzero: it has unit test. i now expanded it to take an argument N and tested latest 1, 2, and 3.

i decided to stick to this query, instead of doing what you suggested (taking latest, and then its prev) because there are 
two possible approaches:
- parse the whole blob for all miners, then query its previous atx. 
the simplest version is put all the atx blobs into memory. (latest+prevs)
to optimize i do one query with blob (the latest), then one query without blob (prevs).

- add a field in table prev_atx to avoid parsing the whole atx blob into memory
this feels unnecessary as prev_atx isn't part of the checkpointed data

checkpoint/runner.go
countvonzero: yes. removed

checkpoint/runner.go
countvonzero: right. the dynamic allocation is on heap :( thanks.

countvonzero:
bors try

countvonzero:
bors merge

PR #4373, opened 2023-05-10
[Merged by Bors] - cmd/node: use shared filelock by default

dshulyak:
originally i added file lock for a specific data directory, as it seemed nicer that we can run multiple instances of go-spacemesh. however there are ongoing reports that people somehow run more than one, and submit different atxs, etc.

it would be very stupid if we will have malfeasence proofs because of that. so i would rather request people to configure separate file locks if they wish to run many instances on same computer, to do so there is an option `--filelock`, that defaults to `/tmp/spacemesh.lock`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -69,6 +69,7 @@ func (cfg *Config) DataDir() string {
 // BaseConfig defines the default configuration options for spacemesh app.
 type BaseConfig struct {
 	DataDirParent string `mapstructure:"data-folder"`
+	FileLock      string `mapstructure:"filelock"`
 
 	ConfigFile string `mapstructure:"config"`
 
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -154,6 +155,7 @@ func DefaultTestConfig() Config {
 func defaultBaseConfig() BaseConfig {
 	return BaseConfig{
 		DataDirParent:       defaultDataDir,
+		FileLock:            "/tmp/spacemesh.lock",
```
pigmej: I think we need to make it platform independent?

pigmej (changes requested):

config/config.go
dshulyak: yeah i realized that it is probablyy no ok on windows. will look up what works there

dshulyak: actually there was os.TempDir

pigmej (approved):

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

pigmej:
bors try

Issue #4088, opened 2023-02-26
p2p: drop misbehaving peer


countvonzero:
## Description

when the node receives a invalid message via gossip or sync, the node should drop the peer

specific to sync, we also need to identity when a peer failed to deliver content for hash that it claimed to have.

related to #3551

dshulyak:
#### invalid messages (or any violation of the protocol rules)
any syntactically invalid message can't be sent by mistake. they are also may not be cheap
to download or verify. therefore we want to drop peer immediately. and schedule peer for blacklisting by ip address. blacklisting it immediately is unnecessary risk for network health, as we will do it by ip address it may blacklist other nodes as well.

we can make a threshold such as 5  syntactically invalid messages result in blacklisting peer for 30 minutes.

#### contextually invalid messages
if peer sends contextually invalid message (such as atx equivocation) it can't be disqualified immediately, as it may not know that the identity equivocated. we don't have strict rules in the implementation to know when peer actually misbehaves. when we will add them handler should report rejection.

#### wasting bandwidth (duplicates, overflowing limits)
libp2p gossipsub provides several configuration parameters to set expectations on message limits. peer will be throttled by dropping published messages, it may potentially waste bandwidth. we need to evaluate such ddos separately, as right now we probably have easier means to make node unusable. 

so for mid-term we should setup provided options to penalize peers that overuse limits or send already known messages.

#### integration
both gossipsub and sync should use decorator, that drops a peer on reject and calls book.Penalize(id, ip).
on book.Penalize track penalty per ip, once penalty grows to blacklist threshold - ip is blacklisted for some delay. penalty for the duration of several layers should be sufficient to prevent ddos.

book should not add or share addresses of the peers that are currently blacklisted. 

libp2p needs to reject connections from blacklisted peers, this is to be implemented by implemeting ConnGater interface. that is configured for libp2p.Host.

book must maintain time based cache for blacklisted peers that expires periodically. for simplicity we can run goroutine in background that expires ips periodically. number of blacklisted ips should be limited to some reasonable number, taking into account memory requirement. it is also not necessary to persist them.

countvonzero:
> blacklisting it immediately is unnecessary risk for network health, as we will do it by ip address it may blacklist other nodes as well.

would like to learn in what scenario would blocking one IP address also blacklist other nodes.

piersy:
> would like to learn in what scenario would blocking one IP address also blacklist other nodes.

[...]

PR #4374, opened 2023-05-10
[Merged by Bors] - Drop p2p connections on ValidationReject

piersy:
## Motivation
We don't want to stay connected to peers that we know are exhibiting invalid behaviour.
This is a step towards completing #4088

## Changes
Change the signature of `pubsub.GossipHandler` to return an error instead of a `pubsub.ValidationResult` this makes the interface used by both sync handlers and pubsub handlers consistent, which allows for the use of a single decorator that will drop a peer if the handler returns a specific error. The decorator (`pubsub.DropPeerOnError`) is applied to all pubsub handlers and sync handlers.

Added a test to verify disconnection on pubsub handlers, to support this I exposed the `p2p.Host` from `node.App`

## Test Plan
The test shows that this is working. A test or tests to verify this behaviour for sync handlers has yet to be added

## DevOps Notes


p2p/pubsub/wrapper.go
```diff
@@ -34,6 +36,19 @@ func (ps *PubSub) Register(topic string, handler GossipHandler) {
 		rst := handler(log.WithNewRequestID(ctx), pid, msg.Data)
 		metrics.ProcessedMessagesDuration.WithLabelValues(topic, castResult(rst)).
 			Observe(float64(time.Since(start)))
+		if rst == ValidationReject {
+			// We want to disconnect the peer and also penalize it which could result in it being blacklisted.
+
+			// Ok we close the peer here but the deadpeerhandler may try a redial if the state of the conn is connected
+			err := ps.host.Network().ClosePeer(pid)
+			if err != nil {
+				ps.logger.With().Error("failed to close peer",
```
dshulyak: this should be Debug. for information maybe it makes sense to add a counter how many peers are disconnected

but also this approach is not sufficient, as mentioned we need to drop peer also when it gives bad message over sync. thats why i suggested to decorate handlers

dshulyak (changes requested):

p2p/pubsub/wrapper.go
piersy: After discussion regarding how to handle bad sync messages, it was decided that handlers should return errors as opposed to validation results and then both sync and pubsusb handlers can share a single wrapper that handles peer dropping.

piersy: @dshulyak I've made the changes we discussed, please take a look.

p2p/pubsub/pubsub.go
```diff
@@ -139,15 +141,40 @@ const (
 	ValidationReject = pubsub.ValidationReject
 )
```
dshulyak: i think we should remove all aliases above

p2p/pubsub/pubsub.go
```diff
@@ -139,15 +141,40 @@ const (
...
+	}
+}
+
+// DropPeerOnError wraps a gossip handler to provide a handler that drops a
+// peer if the wrapped handler returns an error that is the targetErr.
+func DropPeerOnError(handler GossipHandler, h host.Host, logger log.Log, targetErr error) GossipHandler {
+	return func(ctx context.Context, peer peer.ID, data []byte) error {
+		err := handler(ctx, peer, data)
+		if errors.Is(err, targetErr) {
+			// We want to disconnect the peer and also penalize it which could result in it being blacklisted.
```
dshulyak: imo this comment is not that useful

p2p/pubsub/pubsub.go
```diff
@@ -139,15 +141,40 @@ const (
...
+			err := h.Network().ClosePeer(peer)
+			if err != nil {
+				logger.With().Debug("failed to close peer",
+					log.String("peer", peer.ShortString()),
+					log.Err(err),
+				)
```
dshulyak, line 160: we should also track how many peers we are dropping. maybe just a prometheus counter here

p2p/pubsub/pubsub.go
```diff
@@ -139,15 +141,40 @@ const (
 	ValidationReject = pubsub.ValidationReject
 )
 
+// ValidationRejectErr is returned by a GossipHandler to indicate that the
+// pubsub validation result is ValidationReject. ValidationAccept is indicated
+// by a nil error and ValidationIgnore is indicated by any error that is not a
+// ValidationRejectErr.
+var ValidationRejectErr = errors.New("validation reject")
```
dshulyak: as linter said this should be ErrValidationReject, maybe ErrReject

p2p/bad_peer_test.go
```diff
@@ -0,0 +1,154 @@
...
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		app1.Cleanup(ctx)
+		app2.Cleanup(ctx)
+	})
+	g := errgroup.Group{}
+	g.Go(func() error {
+		return app1.Start(ctx)
+	})
+	<-app1.Started()
```
dshulyak: would be better if this couldn't get stuck for whole test timeout if app can't start

p2p/bad_peer_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/protocol"
+	varint "github.com/multiformats/go-varint"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/spacemeshos/go-spacemesh/cmd/node"
```
dshulyak: i don't think that we should import node here. is it possible to write a test without a node, or otherwise add this test to a node module

cmd/node/node.go
```diff
@@ -749,19 +749,28 @@ func (app *App) initServices(
 		app.hare,
 		app.edVerifier,
 	)
-	fetcher.SetValidators(atxHandler, poetDb, proposalListener, blockHandler, proposalListener, app.txHandler, malfeasanceHandler)
+	fetcher.SetValidators(
+		fetch.ValidatorFunc(pubsub.DropPeerOnError(atxHandler.HandleAtxData, app.host, lg, pubsub.ValidationRejectErr)),
```
dshulyak: can we maybe do DropPeerOnReject (as a wrapper around DropPeerOnError, if you want to keep that)?

blocks/handler.go
```diff
@@ -9,13 +9,14 @@ import (
...
 	"github.com/spacemeshos/go-spacemesh/p2p"
+	"github.com/spacemeshos/go-spacemesh/p2p/pubsub"
 	"github.com/spacemeshos/go-spacemesh/sql"
 	"github.com/spacemeshos/go-spacemesh/sql/blocks"
 	"github.com/spacemeshos/go-spacemesh/system"
 )
 
 var (
-	errMalformedData  = errors.New("malformed data")
+	errMalformedData  = fmt.Errorf("malformed data: %w", pubsub.ValidationRejectErr)
```
dshulyak: does it make more sense to have message like: "reject: some reason", so the template in this case "%w: malformed data"?

p2p/bad_peer_test.go
dshulyak: it is also better to use mocknet from libp2p. tcp stack was very flaky on ci runners in the past

cmd/node/node.go
piersy: I thought it could be useful to make it more general, but If you don't think so I'll just make it `DropPeerOnValidationReject` and hardcode the error.

blocks/handler.go
piersy: Yes it does, thanks!

p2p/pubsub/pubsub.go
piersy: Do you know the reason for adding them in the first place, they seemed kind of unnecessary to me?

cmd/node/node.go
dshulyak: i meant that you can keep both of them, but define "shortcut" for Reject, because it is used everywhere

p2p/pubsub/pubsub.go
dshulyak: i defined aliases so that libp2p is not imported directly across project codebase, it is easier to prevent misuse this way

p2p/bad_peer_test.go
piersy: We can add it to the node module, but what is the problem you see with importing node here?

p2p/bad_peer_test.go
piersy: Some of the configuration is done in node, so to test that we need to use node. I think it's nice also to know that the peer dropping is actually working in an approximately real system as opposed to a much more contained test that would involve the mocknet.

I'm interested to know more about the flaky tcp stack, I've never heard of a tcp stack being flaky?

dshulyak: it is environment issue, i am pretty sure that you will see it again with this test

p2p/bad_peer_test.go
dshulyak: ~you will see a cycle if you want to import this module in the tests for node~ test in *_test.go, but i still prefer to avoid node imports here.

and even conceptually, there should be a hierarchy of imports in the project. higher level things (as node) should not be imported by lower level things (such as encoding, p2p, database)

i think it will also lead to failures in p2p module, because of some changes in node. in my opinion thats a mess

piersy: With regards to the hierarchy totally agree with respect to production code. But for tests I think this can be a good approach.

Consider a situation where you want to test an API and you wan't to test it with the full stack active, I.E not mocking out any stuff so having a full node running. You could put these tests in node, but does it really make sense to put the tests in node when they are testing an API that is defined in a different package? If you do that for all the APIs the node package will end up with a lot of tests that don't really have anything to do with node. Structurally there is no problem with importing node in test packages because you can't depend on a test package.

> i think it will also lead to failures in p2p module, because of some changes in node. in my opinion thats a mess

I don't see this as a problem, if node is not supporting correct functioning of the API then that's a real problem. For example if the API is not registered in node then it's not going to work, that is something we want to know about.

p2p/bad_peer_test.go
piersy: >it is environment issue, i am pretty sure that you will see it again with this test

Maybe this has been resolved? I've managed to run this test 23000 times on CI without a single failure.

cmd/node/node.go
piersy: I din't feel like keeping the original implementation and the shortcut, it's easy enough to split out if we need to.  Having both added an extra layer of indirection that I felt was not necessary.

I would have been happy just with the more general version though, IMO the shortcut doesn't provide enough value to justify it's use over the more general version.

EG.

```
pubsub.DropPeerOnError(atxHandler.HandleAtxData, app.host, lg, pubsub.ValidationRejectErr)
```
VS
```
pubsub.DropPeerOnValidationRejectError(atxHandler.HandleAtxData, app.host, lg)
```

But I don't feel that strongly about this.

p2p/bad_peer_test.go
piersy: I've moved the test to the node package.

dshulyak (approved):

dshulyak:
there is this failure

piersy:
bors merge

dshulyak (commented):
```
diff --git a/cmd/node/node.go b/cmd/node/node.go
index 318d1542d..733b5ef58 100644
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -763,7 +763,7 @@ func (app *App) initServices(
        fetcher.SetValidators(
                fetch.ValidatorFunc(pubsub.DropPeerOnValidationReject(atxHandler.HandleAtxData, app.host, lg)),
                fetch.ValidatorFunc(pubsub.DropPeerOnValidationReject(poetDb.ValidateAndStoreMsg, app.host, lg)),
-               fetch.ValidatorFunc(pubsub.DropPeerOnValidationReject(proposalListener.HandleSyncedProposal, app.host, lg)),
+               fetch.ValidatorFunc(pubsub.DropPeerOnValidationReject(proposalListener.HandleSyncedBallot, app.host, lg)),
```

there was a typo

piersy:
bors merge

PR #4376, opened 2023-05-11
[Merged by Bors] - Add membership proof to nipost and remove members list from poet proof

### LINKED ISSUE #3929: Avoid storing the list of members of a Poet round in the proof
### Motivation
Currently, the Poet proof is stored in the following structure, along with a potentially huge list of all members of the respective poet round: https://github.com/spacemeshos/go-spacemesh/blob/5c7f4867ccbe1750b31862e9a4d5576621ebb3d3/common/types/activation.go#L343-L349

It is excessive (the proof is big). It should be enough to only store the merkle proof proving the sequential work. It should be the responsibility of the NiPoST to prove the NiPostChallenge it references is actually a member of the PoET.

### Proposal
The proposed changes are to:
- avoid storing a list of members in a poet proof,
- store a merkle proof proving membership in NiPoST,
- modify NiPoST validation as follows.

#### Updated structures:
```go
type NIPost struct {
	// A proof that the challenge for the PoET, which is
	// constructed from fields in the activation transaction,
	// is a member of the Poet's proof.
	// Proof.ProvenLeaves contains the challenge.
	// Proof.Root must match the Poet's POSW statement.
	Proof poetShared.MerkleProof

	// Post is the proof that the prover data is still stored (or was recomputed) at
	// the time he learned the challenge constructed from the PoET.
	Post *Post

	// PostMetadata is the Post metadata, associated with the proof.
	// The proof should be verified upon the metadata during the syntactic validation,
	// while the metadata should be verified during the contextual validation.
	PostMetadata *PostMetadata
}
```
```go
// PoetProof is the full PoET service proof of elapsed time. It includes the leaf count and the actual PoSW.
type PoetProof struct {
	poetShared.MerkleProof
	LeafCount      uint64
}
```

A node should store `PoetProof` along with the `statement` in a key-value store. The `Statement` will be later used to verify if the nipost challenge of an ATX is actually a member of a PoSW it references. For example:
```go
type PoetProofWithStatement {
	PoetProof
	// The input to Poet's POSW.
	// It's the root of a merkle tree built from all of the members
	// that are included in the proof.
	Statement []byte
}
```

#### Validation of a PoET in nipost builder
1. Construct a merkle tree from the list of poet round's members.
2. The root of the merkle tree is the PoET proof's `statement`.
3. Validate the PoET's merkle proof (POSW) with the `statement`
4. Save `PoetProof` in DB.

When a `PoetProof` is already saved in DB we consider it validated.

#### Validation of NiPoST in ATX handler
1. verify `NIPost.Proof`
2. obtain (from DB/fetch) `PoetProof` by its ref (`NIPost.PostMetadata.Challenge`)
  a. if poet proof needs to be fetched, validate it against `NIPost.Proof.Root`, which is (should be) the POSW `statement`
  b. otherwise verify if `NIPost.Proof.Root` == `proof.Statement` - this check proves the "elapsed time" part
3. verify NIPost.Post the same way as it is done now - this check proves that the smesher still held data after "elapsed time"

poszu:
Closes #3929, see the issue for motivation and explanation.


poszu:
bors try

poszu:
bors try

dshulyak:
just in case if it will be useful - [previous failure](https://grafana.spacemesh.dev/explore?orgId=1&left=%7B%22datasource%22:%22loki-gke-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-gke-dev%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bnamespace%3D%5C%22test-enkj%5C%22,%20pod%3D%5C%22smesher-17-7ddc68888-st2k2%5C%22%7D%20%7C~%20%5C%22atxHandler%5C%22%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%22now-12h%22,%22to%22:%22now%22%7D%7D)

image

by the logs from that env, all except 1 node (lucky smesher-18) failed with this error

poszu:
> the logs from that env, all except 1 node (lucky smesher-18) failed with this error

Yeah, I realized that `Nipost.Membership` merkle proof validation doesn't work in some cases that were not covered by UT. Fix is on the way.

poszu:
bors try

common/types/activation.go
```diff
@@ -298,15 +298,26 @@ func (atx *ActivationTx) Verify(baseTickHeight, tickCount uint64) (*VerifiedActi
...
 type NIPost struct {
-	// Challenge is the challenge for the PoET which is
-	// constructed from fields in the activation transaction.
-	Challenge *Hash32
+	// Membership proves that the challenge for the PoET, which is
+	// constructed from fields in the activation transaction,
+	// is a member of the poet's proof.
+	// Proof.Leaf is the challenge submitted to poet service.
+	// Proof.Root must match the Poet's POSW statement.
+	Membership *MerkleProof
```
dshulyak: does it have to be a pointer?

activation/nipost.go
```diff
@@ -350,38 +356,73 @@ func (nb *NIPostBuilder) getBestProof(ctx context.Context, challenge *types.Hash
...
-		nb.log.With().Info("selected the best proof", log.Uint64("leafCount", bestProof.LeafCount), log.Binary("ref", ref[:]))
-		return ref, nil
+		nb.log.With().Info("selected the best proof", log.Uint64("leafCount", bestProof.poet.LeafCount), log.Binary("ref", ref[:]))
+		return ref, bestProof.membership, nil
+	}
+
+	return types.PoetProofRef{}, nil, ErrPoetProofNotReceived
+}
+
+func constructMerkleProof(challenge *types.Hash32, members []types.Member) (*types.MerkleProof, error) {
```
dshulyak: challenge can be passed by value

activation/validation.go
```diff
@@ -62,31 +59,56 @@ func (v *Validator) NIPost(nodeId types.NodeID, commitmentAtxId types.ATXID, nip
...
+func validateMerkleProof(proof *types.MerkleProof) error {
+	nodes := make([][]byte, 0, len(proof.Nodes))
+	for _, n := range proof.Nodes {
+		nodes = append(nodes, n.Bytes())
+	}
+	ok, err := merkle.ValidatePartialTree(
+		[]uint64{proof.LeafIndex},
+		[][]byte{proof.Leaf.Bytes()},
+		nodes, proof.Root.Bytes(),
+		merkle.GetSha256Parent,
```
dshulyak: i think there is a mistake in that function. usually merkle trees will prefix leafs with 0x0 and inner nodes with 0x1

mainly following https://www.rfc-editor.org/rfc/rfc6962#section-2.1 . need to check why exactly, but i would rather add that prefix

common/types/activation.go
```diff
@@ -298,15 +298,26 @@ func (atx *ActivationTx) Verify(baseTickHeight, tickCount uint64) (*VerifiedActi
 	return vAtx, nil
 }
 
+// Merkle proof proving that a given leaf is included in the root of merkle tree.
+type MerkleProof struct {
+	Root      Hash32
+	Nodes     []Hash32 `scale:"max=32"`
```
fasmat, line 306: What does `Nodes` contain? Only the nodes on the path to the leaf?

api/grpcserver/grpcserver_test.go
```diff
@@ -228,8 +229,26 @@ func TestMain(m *testing.M) {
 }
 
 func newNIPostWithChallenge(challenge *types.Hash32, poetRef []byte) *types.NIPost {
+	tree, err := merkle.NewTreeBuilder().
+		WithLeavesToProve(map[uint64]bool{0: true}).
+		Build()
+	if err != nil {
+		panic("failed to add leaf to tree")
+	}
```
fasmat, line 239:
```suggestion
	require.NoError(t, err, "failed to add leaf to tree")
```

activation/validation.go
dshulyak: image

https://en.wikipedia.org/wiki/Merkle_tree

activation/validation_test.go
```diff
@@ -621,3 +616,27 @@ func validateNIPost(minerID types.NodeID, commitmentAtx types.ATXID, nipost *typ
...
+	t.Run("valid proof", func(t *testing.T) {
+		t.Parallel()
+
+		err := validateMerkleProof(proof)
+		require.NoError(t, err)
+	})
+	t.Run("invalid proof", func(t *testing.T) {
+		t.Parallel()
+
+		invalidProof := *proof
```
fasmat: I'm not sure but isn't this possibly racy?

activation/validation.go
```diff
@@ -62,31 +59,56 @@ func (v *Validator) NIPost(nodeId types.NodeID, commitmentAtxId types.ATXID, nip
...
 	}
 
-	var ref types.PoetProofRef
-	copy(ref[:], nipost.PostMetadata.Challenge)
-	proof, err := v.poetDb.GetProof(ref)
-	if err != nil {
-		return 0, fmt.Errorf("poet proof is not available %x: %w", nipost.PostMetadata.Challenge, err)
+	poetChallenge := nipost.Membership.Leaf
+	if !bytes.Equal(poetChallenge[:], expectedChallenge[:]) {
+		return 0, fmt.Errorf("invalid `Challenge` in membership merkle proof leaf; expected: %x, given: %x", expectedChallenge, poetChallenge)
```
dshulyak: i think this check and additional 32 bytes are redundant, you can use nipost hash for merkle computation

common/types/activation.go
poszu: No, I only followed the convention

activation/validation.go
poszu: I don't understand - what is wrong with that function? Whether the leafs are prefixed for hashing or not is responsibility of `merkle-tree` library, not this code. Here I only pass the proof for validation. The `LeafIndex` is needed for the validation code to know whether leaf is left or right sibling

activation/validation.go
poszu: True. I will remove `Leaf` from `types.MerkleProof`

activation/validation.go
```diff
@@ -62,31 +59,56 @@ func (v *Validator) NIPost(nodeId types.NodeID, commitmentAtxId types.ATXID, nip
...
+		return 0, err
 	}
 
 	if err := v.Post(nodeId, commitmentAtxId, nipost.Post, nipost.PostMetadata, numUnits, opts...); err != nil {
 		return 0, fmt.Errorf("invalid Post: %v", err)
 	}
 
+	var ref types.PoetProofRef
+	copy(ref[:], nipost.PostMetadata.Challenge)
+	proof, err := v.poetDb.GetProof(ref)
```
dshulyak: we also need a check that membership.Root matches PoetProofMessage.Statement . otherwise i can use any membership with any poet proof

common/types/activation.go
poszu: Yes. The blue nodes in https://medium.com/crypto-0-nite/merkle-proofs-explained-6dd429623dc5

activation/validation.go
dshulyak: i don't understand you. it doesn't matter which piece of code is responsible for that. that function doesn't do the right thing

api/grpcserver/grpcserver_test.go
poszu: Not possible as this is in a global context to initialize a global variable (which is then needed in `TestMain()` :( . Refactoring globals in this test is out of scope.

activation/validation_test.go
poszu: No, as long as `proof` is readonly.

api/grpcserver/grpcserver_test.go
fasmat: You mean the `nipost` global variable? Can we make that variable local instead? It is only used in 2 places (both in the same test) I think that's OK to include in this PR if `newNIPostWithChallenge` is updated as well. 🙂

fasmat: NVM, just saw its not a test but `TestMain`, then forget my suggestion.

common/types/activation.go
poszu: Changed to value

activation/validation.go
poszu: Can you explain why it doesn't do the right thing?

poszu:
bors try

activation/validation.go
dshulyak: there is a screen above that explains second preimage attack. i personally would just fix it, but if we don't want to do so for some reason i suggest to talk with Iddo about how smart is to have such code

activation/validation.go
```diff
@@ -62,31 +60,54 @@ func (v *Validator) NIPost(nodeId types.NodeID, commitmentAtxId types.ATXID, nip
...
-			return true
+func validateMerkleProof(leaf []byte, proof *types.MerkleProof) error {
+	nodes := make([][]byte, 0, len(proof.Nodes))
+	for _, n := range proof.Nodes {
+		nodes = append(nodes, n.Bytes())
+	}
+	ok, err := merkle.ValidatePartialTree(
+		[]uint64{proof.LeafIndex},
+		[][]byte{leaf},
+		nodes, proof.Root.Bytes(),
```
dshulyak: i think it is better to pass nipost.Statement directly here and drop proof.Root field, and that validation

dshulyak (approved):

activation/validation.go
poszu: I understand the description in Wikipedia, but I don't see how this function is wrong. AFAIR, the https://github.com/spacemeshos/merkle-tree, which poet uses to construct merkle tree of all members, doesn't support that mitigation form screen (prepending 0x00 / 0x01).

Do you mean that we should first fix merkle-tree?

@noamnelke could you help understand if this is a problem?

dshulyak: yes thats what i mean. i asked Iddo in research channel.
my point is that any "good" merkle tree implementation should do that. but in practice this attack may not be relevant (by happenstance)

noamnelke: I don't think this affects us, but there's also no reason not to add the prefix (in the merkle lib). It's easy, so let's just do it.

activation/poet.go
noamnelke, line 1: So we're still getting the full list of members from the PoET server. This is a good way to implement this security fix quickly, but it's probably suboptimal in the sense that nodes will have to download a long list of members while they only need a small Merkel proof. I suppose we'll fix this as part of future PoET work that's planned anyway.

poszu: Yes. There is an open issue (started but on hold) to split poet proof API into:
- fetching members,
- fetching proof.

Members could be fetched right after the poet round start.

activation/validation.go
poszu: Implemented prepending data for hashing with 0x00 for leafs and 0x01 for nodes. Also changed hashing function for tree nodes to blake3.

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

activation/validation.go
```diff
@@ -62,31 +61,52 @@ func (v *Validator) NIPost(nodeId types.NodeID, commitmentAtxId types.ATXID, nip
...
+	if !ok {
+		hexNodes := make([]string, 0, len(proof.Nodes))
+		for _, n := range proof.Nodes {
+			hexNodes = append(hexNodes, n.Hex())
 		}
+		return fmt.Errorf(
+			"invalid merkle proof, calculated root does not match the proof root, leaf: %v, nodes: %v, expected root: %v",
+			util.Encode(leaf),
+			hexNodes,
+			util.Encode(expectedRoot),
```
dshulyak, line 106: what do you think about pattern in types/missing.go ? it allows to print error nicely by implementing logging interface

dshulyak (approved):

activation/validation.go
poszu: I'm all for it, I also implemented `MarshalLogObject` for some types in the past. I'm not sure how you would use it here though.

poszu:
bors merge

activation/validation.go
poszu: I misunderstood what you meant. Do you mean to create an error type to represent a particular failure and implement logging i/f for it? It seems a bit of an overkill if the error is used only in one place and if it is not desired to handle it in any way (like in this case).

PR #4378, opened 2023-05-12
[Merged by Bors] - tortoise: assign hare and opinion fields

dshulyak:
this doesn't fix any bugs, i just messed up with cherry picks a bit and left those fields unassigned. they are logged in mesh with "consensus results" message and may confuse people, when they see them empty


countvonzero (approved):

dshulyak:
bors merge

PR #4380, opened 2023-05-12
[Merged by Bors] - Bundle profiler tool

### LINKED ISSUE #4349: Pack post-rs/profiler tool next to the go-spacemesh binaries
## Description
Smapp requires the `profiler` tool from https://github.com/spacemeshos/post-rs
The easiest proper way is to pack it all together as a part of CI.
We need to tweak GHA workflows.

brusherru:
## Motivation
It closes #4349
Part of https://github.com/spacemeshos/smapp/pull/1242
Actually, we don't need it bundled in the same archive with go-spacemesh, but it seems to me as the easiest way to keep the profiler up to date with post-rs library used in the go-spacemesh.

**❗ We need to add some valid `POSTRS_SETUP_REV` to this branch before merging. ❗** 

## Changes
- Added `make get-profiler`
- It will download the profiler and unzip it into the `build` directory
- Then it will be packed in the same archive

## Test Plan
To test these changes I've pushed also `test-4380-profiler-tool` branch and pushed the tag `v0.3.2-beta.1`
It contains valid `POSTRS_SETUP_REV` (to the one with the profiler tool).

## DevOps Notes
- [ ] This PR does not rely on a new version of external services.
        It relies on `post-rs`.


pigmej (approved):

Makefile-libs.Inc
```diff
@@ -52,9 +52,14 @@ endif
 POSTRS_SETUP_REV = 0.1.12
```
countvonzero: 0.1.13?

Makefile-libs.Inc
```diff
@@ -71,7 +76,12 @@ $(BINDIR_POSTRS_SETUP_LIBS): $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
 $(PROJ_DIR)$(POSTRS_SETUP_ZIP):
 	curl -L $(POSTRS_SETUP_URL_ZIP) -o $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
 
+$(BIN_DIR)$(POSTRS_PROFILER_BIN):
+	curl -L $(POSTRS_PROFILER_URL) -o $(PROJ_DIR)$(POSTRS_PROFILER_ZIP)
+	unzip -o -j $(PROJ_DIR)$(POSTRS_PROFILER_ZIP) -d $(BIN_DIR)
```
poszu, line 81: I'd suggest renaming it to something like `post_proving_profiler.exe` because just `profiler.exe` is ambiguous (what does it profile?)

poszu (commented):
IMHO, it doesn't belong here because go-spacemesh has no use for it, but I understand we are short on time and need to cut corners.

I think it should be placed in smapp CI. go-spacemesh could somehow expose a version of post-rs it depends on in an easy-to-scrap way. Perhaps a file `post-rs-version.txt` could be put into release artifacts? Then smapp CI could pick this up and fetch the right profiler version. What do you think?

andres-spacemesh:
bors merge

fasmat:
I agree with @poszu. I don't think the profiler belongs into `go-spacemesh`. The node doesn't use this tool.

I think like this should be part of release process / packaging of `smapp`.

pigmej:
Was discussed with Smapp, and it will be decoupled later.

PR #4382, opened 2023-05-13
[Merged by Bors] - Revert "tortoise: use rlock for read-only methods and meter store"

dshulyak:
This reverts commit 9552e34ad00e80afdd8ed3cc2e6ea54a725c8221.

mistakes were made. i forgot that even on read path i added some functions like that, to prevent accidental crashes:

```
func (s *state) epoch(eid types.EpochID) *epochInfo {
	epoch, exist := s.epochs[eid]
	if !exist {
		epochsNumber.Inc()
		epoch = &epochInfo{atxs: map[types.ATXID]atxInfo{}}
		s.epochs[eid] = epoch
	}
	return epoch
}
```

### DIFF
--- a/tortoise/algorithm.go
+++ b/tortoise/algorithm.go
@@ -42,7 +42,7 @@ type Tortoise struct {
 	ctx    context.Context
 	cfg    Config
 
-	mu   sync.RWMutex
+	mu   sync.Mutex
 	trtl *turtle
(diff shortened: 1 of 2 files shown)


countvonzero (approved):

countvonzero:
bors merge

PR #4383, opened 2023-05-13
[Merged by Bors] - mesh: rework processing of the consensus results

dshulyak:
most of this change is about making consensus results processing to be more straightforward. and it also removes fetching of the blocks referenced in ballots.

tortoise outputs latest consensus results with computed opinion as in example:
```json
[
  {
    "layer": 24,
    "opinion": "55bfc8b672",
    "verified": true,
    "blocks": [
      {
        "id": "59e7b07d06",
        "layer": 24,
        "height": 23563,
        "valid": true,
        "invalid": false,
        "hare": true,
        "data": true
      }
    ]
  },
  {
    "layer": 25,
    "opinion": "1a5fe28f37",
    "verified": false,
    "blocks": [
      {
        "id": "e2cda88192",
        "layer": 25,
        "height": 23563,
        "valid": false,
        "invalid": false,
        "hare": true,
        "data": true
      }
    ]
  }
]
```

mesh is supposed to take this result and ensure that the vm state (and the rest of db) is consistent with results computed by tortoise. 

tortoise tracks when opinion has changed, and therefore it can provide results that start before currently applied layer. in such case mesh goes over layers under the question and finds which one it needs to revert.

after that mesh applies all layers consistently with tortoise results. with the nuance that we want to apply hare within hdist, unless block upvoted by hare crossed negative threshold. see layer 25 from the json above.


dshulyak:
bors try

countvonzero (approved):

countvonzero:
bors merge

dshulyak:
bors try

dshulyak:
clearly a bug somewhere. this looks like a correct consensus results

image
image

but on the node that was online

tortoise computed correct state
image

and then there is some mess because of the revert 
image

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

mesh/mesh.go
```diff
@@ -281,87 +234,88 @@ func (msh *Mesh) processValidityUpdates(ctx context.Context, logger log.Log, upd
...
+		log.Context(ctx),
+		log.Uint32("layer_id", lid.Uint32()),
+	)
 
-	// set processed layer even if later code will fail, as that failure is not related
-	// to the layer that is being processed
-	if err := msh.setProcessedLayer(logger, layerID); err != nil {
-		return err
-	}
+	msh.trtl.TallyVotes(ctx, lid)
```
countvonzero, line 276: why does tallying vote need to be inside the locked block?

mesh/mesh.go
```diff
@@ -281,87 +234,88 @@ func (msh *Mesh) processValidityUpdates(ctx context.Context, logger log.Log, upd
...
-
-	if !to.Before(from) {
-		if err := msh.pushLayersToState(ctx, logger, from, to); err != nil {
+	if next := msh.LatestLayerInState() + 1; next < msh.pendingUpdates.min {
+		msh.pendingUpdates.min = next
+		pending = true
+	}
+	if pending {
+		var err error
+		results, err = msh.trtl.Results(msh.pendingUpdates.min, msh.pendingUpdates.max)
```
countvonzero, line 293: i am confused by Updates() and Results() api.
the code is counting on `msh.pendingUpdates.min` to remember the min layer from the last tortoise update.
and if the mesh somehow was't able to apply the block from call, code asks for trtl for more prior updates that mesh didn't persist. these two apis are very similar.

would it make sense to create have a channel btwn from trtl->mesh that pushes the min updated layer each time?
that way mesh can query results btwn min(instate+1, trtl min updated layer) and the latest verified layer. then we'll just need Results()

countvonzero (approved):
the partition tests used to the only tests that give me confidence for this area of code.
is it possible to add back those tests?

dshulyak:
> is it possible to add back those tests?

i will add them back, but they are not quite stable. will be looking into that

mesh/mesh.go
dshulyak: > latest verified layer

max is not verified layer, as we want to get also latest optimistic results with computed opinion

maybe i can make updates to return *Range instead of []result.Layer

mesh/mesh.go
dshulyak: Mesh.mu is used only for ProcessLayer, i don't know any reason for TallyVotes to be outside ProcessLayer pipeline

mesh/mesh.go
dshulyak: i will look into that in https://github.com/spacemeshos/go-spacemesh/issues/4425

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #4384, opened 2023-05-13
[Merged by Bors] - p2p/server: limit request size alloc to 10kb

### LINKED ISSUE #3970: p2p server: set limits for request and response

dshulyak:
lack of limit before alloc can be trivially abused for oom

closes: https://github.com/spacemeshos/go-spacemesh/issues/3970


fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #4385, opened 2023-05-13
[Merged by Bors] - codec: dont copy bytes from previous encoding

dshulyak:
previous code could copy bytes from previously encoded value


fasmat (approved):

dshulyak:
bors merge

PR #4387, opened 2023-05-15
recover from a checkpoint file 

countvonzero:
## Motivation
part of #4090 

## Changes
- add a command line arg --checkpoint-file to specify URI for checkpoint file. the file can be from
  - an URL (https://)
  - local file (file://)
  the node will download/copy the file to local disk and execute the recovery logic

- add a command line arg --restore-layer to specify which layer to restart the mesh

- recovery logic include the following steps:
  - read recovery file
  - check if the node's most recent ATX is in the checkpoint file. if not, copy its latest ATX and dependencies (prev/positioning ATX,  and the poet proof) from the old db
  - backup the old database files
  - create a new database file, save the checkpoint ATXs, accounts and the data for the node's most recent ATX if any
  - set the effective genesis layer to (restore layer - 1)
  - resume the app init logic

- sqlite recovery table
only allow 1 entry, and records the the restore layer.

- ATXs re-gossip 
when ATXs are not included in the checkpoint file, the network will re-gossip these ATXs by syncing them from peers as soon as a node recovers the checkpoint data. this is achieved by syncer requesting current epoch ATXs immediately after recovery

- gossip handlers
reject data (proposal/ballot/blocks) that are before the restore layer

- debug API service
add querying account states at specific layer

- admin recover RPC
the only way (i can figure out) to preserve PoST data is to make the node commit suicide (log fatal) after receiving the recover RPC.

- checkpoint systest does the following:
  - submit transactions to spawn account and transfer coins
  - get accounts nonce/balance
  - issue checkpoint RPC to every node
  - check all the returned checkpoint data are the same
  - issue recovery RPC to every node (node will copy the local checkpoint file to $dataDir/recovery/) and restart
  - check that all accounts have the same nonce/balance as before the checkpoint
  - run testSmeshing() the nodes that tests miners are generating proposals
  - make boostrapper serve the checkpoint data
  - add two new nodes with --checkpoint=http://bootstrapper-0:80/checkpoint
  - run testSmeshing on these two nodes


tortoise/algorithm.go
```diff
@@ -133,7 +133,8 @@ func (t *Tortoise) OnBeacon(eid types.EpochID, beacon types.Beacon) {
 		log.Uint32("evicted", evicted.Uint32()),
 		log.Stringer("beacon", beacon),
 	)
-	if eid <= evicted {
+	// when the node recover from checkpoint, the new effective genesis can be the same as the beacon genesis
```
dshulyak: what is beacon genesis?

countvonzero: sorry. i mean to say 
"the new effective genesis can be the target epoch for the beacon"

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
closed for the changes were merged in 3 separate PRs

PR #4390, opened 2023-05-15
[Merged by Bors] - PoST: update dependency and documentation

fasmat:
## Motivation
This moves installation of dependencies back into `make install` rather than requiring to install them explicitly.

## Changes
Update `Makefile` and `ci.yml`

## TODO

## DevOps Notes


Makefile-libs.Inc
```diff
@@ -100,7 +100,17 @@ $(BINDIR_POSTRS_SETUP_LIBS): $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
 $(PROJ_DIR)$(POSTRS_SETUP_ZIP):
 	curl -L $(POSTRS_SETUP_URL_ZIP) -o $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
 
-get-postrs-lib: $(PROJ_DIR)$(POSTRS_SETUP_ZIP) $(BINDIR_POSTRS_SETUP_LIBS)
+install-deps:
+ifeq ($(platform), $(filter $(platform), linux linux-arm64))
+	sudo apt-get update -q && sudo apt-get install -qy ocl-icd-opencl-dev libpocl2 clinfo
```
pigmej: That has one terrible side effect IMHO. Now we will strongly support ONLY Ubuntu-like systems. And you'll not be able to use makefile on ANY other Linux distro. Not to mention that this requires also sudo (so it will break in any non-elevated scenarios)

I think it's bad idea to do it like that.

pigmej (changes requested):

Makefile-libs.Inc
piersy: @pigmej how would you do it?

Why is the sudo thing such a problem? When I executed this in a non elevated situation I was prompted for the password and then it worked.

Makefile-libs.Inc
```diff
@@ -100,7 +100,17 @@ $(BINDIR_POSTRS_SETUP_LIBS): $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
...
+else
+ifeq ($(platform), windows)
+	choco install opencl-intel-cpu-runtime
+endif
```
piersy: What about mac systems?

Makefile-libs.Inc
fasmat: @pigmej it requires a linux system with `apt`. I added a check to not execute this line when `clinfo` is already installed.

We are only supporting Ubuntu anyway? I don't see the issue.

Makefile-libs.Inc
fasmat: Don't know, at least on CI it isn't required to install anything for the tests to run on macOS...

Makefile-libs.Inc
pigmej: Since when we support "only ubuntu" ? :)

pigmej: @fasmat I really dislike that "already installed' because you're checking only one of 3 packages, easily can be false positive, apt is smart enough to do noop if all is installed.

pigmej: if you want to use clinfo to check the setup then we have to `clinfo -l` and check if there is at least one platform / device combination.

fasmat: > @fasmat I really dislike that "already installed' because you're checking only one of 3 packages, easily can be false positive, apt is smart enough to do noop if all is installed.

without the check `apt` will update every time I call `make test` which is quite annoying. So I would have to remove `apt update` but then `apt install` might not work.

How would you implement `install-deps`?

piersy: Hey @pigmej 
>That has one terrible side effect IMHO. Now we will strongly support ONLY Ubuntu-like systems. And you'll not be able to use makefile on ANY other Linux distro.

What approach would you take?

piersy: > Since when we support "only ubuntu" ? :)

What do we support, is it written down anywhere?

Makefile-libs.Inc
fasmat: I added documentation on how to set up OpenCL to the `README.md`

Makefile-libs.Inc
fasmat: I'm not sure if we explicitly have it written down somewhere but imo:

- Ubuntu 22.04 (amd64, arm64 only experimentally)
- macOS + macOS m1 (no idea which version we have as minimum propably macOS 12)
- Windows 10 (amd64)

At least we have our pipelines set up for those.

fasmat:
bors try

pigmej (approved):

fasmat:
bors merge

fasmat:
bors merge

PR #4404, opened 2023-05-18
[Merged by Bors] - recover 1of3: process checkpoint file

countvonzero:
## Motivation
part of #4090
PR 1 of 3 to break down large PR https://github.com/spacemeshos/go-spacemesh/pull/4387

## Changes
- add a command line arg --checkpoint-file to specify URI for checkpoint file. the file can be from
  - an URL (https://)
  - local file (file://)
  the node will download/copy the file to local disk and execute the recovery logic

- add a command line arg --restore-layer to specify which layer to restart the mesh

- recovery logic include the following steps:
  - read recovery file
  - check if the node's most recent ATX is in the checkpoint file. if not, copy its latest ATX and dependencies (prev/positioning ATX, and the poet proof) from the old db
  - backup the old database files
  - create a new database file, save the checkpoint ATXs, accounts and the data for the node's most recent ATX if any
  - set the effective genesis layer to (restore layer - 1)
  - resume the app init logic

- sqlite recovery table
  - only allow 1 entry, and records the the restore layer.

- admin recover RPC
  the only way (i can figure out) to preserve PoST data is to make the node commit suicide (log fatal) after receiving the recover RPC.

- debug API service
  - add querying account states at specific layer


dshulyak:
> the only way (i can figure out) to preserve PoST data is to make the node commit suicide (log fatal) after receiving the recover RPC.

i think there are nicer options, like cancaling context that blocks app.Start, but we need to be sure that graceful termination is robust.

checkpoint/runner.go
```diff
@@ -23,59 +22,21 @@ const (
...
-			Restore:      restore.Uint32(),
+			CheckpointId: fmt.Sprintf("snapshot-%d", snapshot),
 		},
 	}
 
 	tx, err := db.Tx(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("create db tx: %s", err)
 	}
-	defer tx.Release()
```
dshulyak: why?

checkpoint/recovery.go
```diff
@@ -0,0 +1,383 @@
...
+
+func RecoveryFilename(dataDir, base string, restore types.LayerID) string {
+	matches := regexp.MustCompile(fileRegex).FindStringSubmatch(base)
+	if len(matches) > 0 {
+		return filepath.Join(RecoveryDir(dataDir), base)
+	}
+	return filepath.Join(RecoveryDir(dataDir), fmt.Sprintf("%s-restore-%d", base, restore.Uint32()))
+}
+
+func ParseRestoreLayer(fname string) (types.LayerID, error) {
```
dshulyak, line 70: is there any other other option beside parsing it from the filename? usually it is avoided to store metadata in filenames

there is an option with two distinct files, merging them into tarball (e.g one will be meta.json, and another recovery.json) and then distributing tarball is more robust

checkpoint/recovery.go
```diff
@@ -0,0 +1,383 @@
...
+	restore types.LayerID,
+) (*sql.Database, error) {
+	oldRestore, err := recovery.CheckpointInfo(db)
+	if err != nil {
+		return nil, fmt.Errorf("get last checkpoint: %w", err)
+	}
+	if oldRestore >= restore {
+		return nil, nil
+	}
+	logger.With().Info("Recover from uri", log.String("uri", uri))
```
dshulyak: is there any reason to upper case logs?

checkpoint/recovery.go
```diff
@@ -0,0 +1,383 @@
...
+	own.atx, err = atxs.Get(db, atxid)
+	if err != nil {
+		return own, fmt.Errorf("get own atx: %w", err)
+	}
+	own.preserve = append(own.preserve, own.atx)
+	deps := map[types.ATXID]struct{}{own.atx.PrevATXID: {}}
+	deps[own.atx.PositioningATX] = struct{}{}
+	if own.atx.CommitmentATX != nil {
+		deps[*own.atx.CommitmentATX] = struct{}{}
+	}
```
dshulyak, line 255: it such atx is preserved and re-gossiped and commitment atx is not in the state, what do we do?

dshulyak: do i understand it correctly, that you preserve whole chain and then nodes will have to download it? i think this is not an option

dshulyak:
i also noticed that coverage is <70%. i think we should aim for 100% for new modules

checkpoint/recovery.go
dshulyak: i am somewhat confused about atxs that are kept for recovery. originally there was an idea that we keep a set of "golden" atxs, that are hardcoded when checkpoint is distributed. it meant that the only part that has to be kept around is post data.

later there was a re-gossip idea, but (re)-gossiping requires full body of the atx that was generated using old code. so e.g post proof, poet proof, other atx validation must be kept atleast for some time. is it definitely better?

countvonzero: there are two kinds of ATXs:

A. checkpointed: 
  - these are the golden atxs you were referring to. 
  - the body is not preserved and these cannot be gossiped.
  - new miners have to get these atxs from the checkpoint file.
  - currently we take the latest 2 atxs from each miner to form the set of checkpointed atxs
 
B. re-gossiped own atxs:
  - these are full atxs that preserved at the time of recovering from checkpoint file when a miner finds that its own latest atx is not included in the checkpoint file.
  - these are re-gossiped via syncer syncing aggressively at the time of recovery
  - the whole chain is not preserved, only the directly referenced atxs, and those referenced ones should refer back to one in set A. 
 
the whole point of B is so that miners do not have to wait another epoch if their atxs missed the checkpoint. as you pointed out, atxs in group B will carry the old format, which may defeat the purpose of the checkpoint recovery. 

but the worst case scenario is similar, if they are no longer valid under the new code after checkpoint, then we drop them (without marking them malicious). if they are still valid, then we gain on UX.

in any case, we can drop B from this PR and initiate a separate discussion with research.

checkpoint/recovery.go
countvonzero: this code is only exercised in systest (with the flag --recover-from-default-dir). 
i agree with you that we should not infer anything from filename. but in this specific case for systest testing, i truly run out of ideas (restarting the nodes without the ability to feed it new flag)

checkpoint/recovery.go
countvonzero: or maybe we can use a config to control whether we want to accept atxs in group B?

dshulyak: i guess allowing them by default and disabling by config option seems reasonable

dshulyak (approved):

checkpoint/runner.go
countvonzero: it's replaced by the line below. (my IDE complained about un-captured error

dshulyak: i noticed. i probably commented on wrong line. just seemed strange. maybe setting for ide have to be fixed, linter in go-spacemesh doesn't complain on this

checkpoint/recovery.go
```diff
@@ -0,0 +1,383 @@
...
+	return restore, nil
+}
+
+func ReadCheckpointAndDie(ctx context.Context, logger log.Log, dataDir, uri string, restore types.LayerID) error {
+	fs := afero.NewOsFs()
+	file, err := copyToLocalFile(ctx, logger, fs, dataDir, uri, restore)
+	if err != nil {
+		return fmt.Errorf("copy checkpoint file before restart: %w", err)
+	}
+	logger.With().Panic("restart to recover from checkpoint", log.String("file", file))
```
poszu, line 100: Instead of panicking here, you could give a `stop func()` to `AdminService` that would cancel `App`'s context and remove killing from this function entirely. Then, `AdminService::Recover()` would call `checkpoint.ReadCheckpoint(...)` and call the `stop()` function. It will allow testing it nicely by simply mocking the `stop` function. Example:

```go
// api/grpcserver/admin_service.go
type AdminService struct {
	stop func()
	// other stuff
}

func (a AdminService) Recover(ctx context.Context, req *pb.RecoverRequest) (*empty.Empty, error) {
	file, err := checkpoint.ReadCheckpoint(ctx, a.logger, a.dataDir, req.Uri, types.LayerID(req.RestoreLayer))
	if err != nil {
		a.logger.WithContext(ctx).With().Error("failed to read checkpoint file", log.Err(err))
		return nil, status.Errorf(codes.Internal,
			fmt.Sprintf("read checkpoint %s and restore %d: %s", req.Uri, req.RestoreLayer, err.Error()))
	}
	a.logger.WithContext(ctx).With().Info("read checkpoint, shutting down app", log.String("file", file)) 
	a.stop() // <- it only cancels the App's context and lets GRPC response be sent :)
	return &empty.Empty{}, nil
}
```

```go
// checkpoint/recovery.go
func ReadCheckpoint(ctx context.Context, logger log.Log, dataDir, uri string, restore types.LayerID) (string error) {
	fs := afero.NewOsFs()
	file, err := copyToLocalFile(ctx, logger, fs, dataDir, uri, restore)
	if err != nil {
		return fmt.Errorf("copy checkpoint file before restart: %w", err)
	}
	return file, nil
}
```

```go
// somewhere in node.go

// This ctx should be passed around to make sure everything quits as expected when its cancelled
ctx, cancel := context.WithCancel(ctx)
...
grpcserver.NewAdminService(cancel, /* other args */)
```

checkpoint/util.go
```diff
@@ -62,3 +80,71 @@ func ValidateSchema(data []byte) error {
...
+	}()
+	return rf.copy(fs, srcf)
+}
+
+func httpToLocalFile(ctx context.Context, resource *url.URL, fs afero.Fs, dst string) error {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, resource.String(), nil)
+	if err != nil {
+		return fmt.Errorf("create http request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
```
poszu: Why set content type on a GET request? I think it's not needed and won't be used by the server.

common/types/activation.go
```diff
@@ -184,6 +186,17 @@ func NewActivationTx(
...
+// a golden ATX is not verifiable, and is only allowed to be prev atx or positioning atx.
+func (atx *ActivationTx) Golden() bool {
+	return atx.golden
+}
+
+// SetGolden set atx to golden.
+func (atx *ActivationTx) SetGolden() {
+	atx.golden = true
+}
+
```
poszu, line 199: Why not simply make the field public instead? Is it to avoid it being serialized by scale?

sql/atxs/atxs_test.go
```diff
@@ -50,6 +50,32 @@ func TestGet(t *testing.T) {
...
+		require.NoError(t, atxs.Add(db, atx))
+		expected[atx.ID()] = struct{}{}
+	}
+
+	all, err := atxs.All(db)
+	require.NoError(t, err)
+	for _, got := range all {
+		delete(expected, got)
+	}
+	require.Empty(t, expected)
```
poszu: I think this could be simplified with `require.ElementsMatch()`. For example:
```suggestion
	var expected []types.ATXID
	for _, atx := range atxList {
		require.NoError(t, atxs.Add(db, atx))
		expected = append(expected, atx.ID())
	}

	all, err := atxs.All(db)
	require.NoError(t, err)
	require.ElementsMatch(t, expected, all)
```

checkpoint/recovery.go
countvonzero: Dmitry also mentioned this idea here https://github.com/spacemeshos/go-spacemesh/pull/4404#issuecomment-1556804794
this approach is of course better, but i don't want to change in this PR for following reasons:

- this codepath is only for systest (with a special flag on). in production, for node operator to recover from a checkpoint, we require them to use a different config that points to a recovery file from url and download from there.

- context canceling and the node exit gracefully is not well tested, and it doesn't trigger restart. i will have to change the systest node restart policy

common/types/activation.go
countvonzero: right. this boolean is inferred.
also, golden atx should never be sync'ed or gossiped

countvonzero:
bors try

countvonzero:
> i also noticed that coverage is <70%. i think we should aim for 100% for new modules

i did my best. most of the uncovered lines are error handling for os and sqlite.

countvonzero:
bors merge

PR #4405, opened 2023-05-18
[Merged by Bors] - recover 2of3: adjust to new effective genesis after recovery

countvonzero:
## Motivation
part of https://github.com/spacemeshos/go-spacemesh/issues/4090
PR 2 of 3 to break down large PR https://github.com/spacemeshos/go-spacemesh/pull/4387
loosely dependent on PR 1 of 3 https://github.com/spacemeshos/go-spacemesh/pull/4404

## Changes
- ATXs re-gossip
when ATXs are not included in the checkpoint file, the network will re-gossip these ATXs by syncing them from peers as soon as a node recovers the checkpoint data. this is achieved by syncer requesting current epoch ATXs immediately after recovery

- gossip handlers
reject data (proposal/ballot/blocks) that are before the restore layer

### DIFF
--- a/tortoise/algorithm.go
+++ b/tortoise/algorithm.go
@@ -119,14 +119,13 @@ func (t *Tortoise) OnWeakCoin(lid types.LayerID, coin bool) {
 func (t *Tortoise) OnBeacon(eid types.EpochID, beacon types.Beacon) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
-	evicted := t.trtl.evicted.GetEpoch()
+	firstInWindow := t.trtl.evicted.Add(1).GetEpoch()
 	t.logger.With().Debug("on beacon",
-		log.Uint32("epoch_id", eid.Uint32()),
-		log.Uint32("evicted", evicted.Uint32()),
+		log.Uint32("epoch id", eid.Uint32()),
+		log.Uint32("first epoch", firstInWindow.Uint32()),
 		log.Stringer("beacon", beacon),
 	)
-	// when the node recover from checkpoint, the new effective genesis can be the same as the beacon's target epoch.
-	if eid <= evicted && eid != types.GetEffectiveGenesis().GetEpoch() {
+	if eid < firstInWindow {
 		return
 	}
 	epoch := t.trtl.epoch(eid)
--- a/tortoise/algorithm.go
+++ b/tortoise/algorithm.go
@@ -125,7 +125,8 @@ func (t *Tortoise) OnBeacon(eid types.EpochID, beacon types.Beacon) {
 		log.Uint32("evicted", evicted.Uint32()),
 		log.Stringer("beacon", beacon),
 	)
-	if eid <= evicted {
+	// when the node recover from checkpoint, the new effective genesis can be the same as the beacon's target epoch.
+	if eid <= evicted && eid != types.GetEffectiveGenesis().GetEpoch() {
 		return
 	}
 	epoch := t.trtl.epoch(eid)
--- a/tortoise/recover.go
+++ b/tortoise/recover.go
@@ -26,6 +26,14 @@ func Recover(db *datastore.CachedDB, beacon system.BeaconGetter, opts ...Opt) (*
 	if err != nil {
 		return nil, fmt.Errorf("failed to load latest known layer: %w", err)
 	}
+
+	if types.GetEffectiveGenesis() != types.FirstEffectiveGenesis() {
+		// need to load the golden atxs after a checkpoint recovery
+		if err := recoverEpoch(types.GetEffectiveGenesis().Add(1).GetEpoch(), trtl, db, beacon); err != nil {
+			return nil, err
+		}
+	}
+
 	epoch, err := atxs.LatestEpoch(db)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load latest epoch: %w", err)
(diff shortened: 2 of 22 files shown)


tortoise/algorithm.go
```diff
@@ -133,7 +133,8 @@ func (t *Tortoise) OnBeacon(eid types.EpochID, beacon types.Beacon) {
 		log.Uint32("evicted", evicted.Uint32()),
 		log.Stringer("beacon", beacon),
 	)
-	if eid <= evicted {
+	// when the node recover from checkpoint, the new effective genesis can be the same as the beacon's target epoch.
+	if eid <= evicted && eid != types.GetEffectiveGenesis().GetEpoch() {
```
dshulyak: this can't be done like this, as it will be never evicted from memory. after recovery evicted epoch should be before types.GetEffectiveGenesis()

tortoise/recover.go
```diff
@@ -25,6 +25,18 @@ func Recover(db *datastore.CachedDB, beacon system.BeaconGetter, opts ...Opt) (*
 	if err != nil {
 		return nil, fmt.Errorf("failed to load latest known layer: %v", err)
 	}
+
+	if latest == 0 && types.GetEffectiveGenesis() != types.FirstEffectiveGenesis() {
```
dshulyak: what if latest is not zero and we want this data in tortoise state? i think this data should be loaded regardless of latest

tortoise/algorithm.go
countvonzero: yeah, sorry. 
the evicted layer is set to the new genesis layer. however, the `evicted` variable here lost the layer resolution.
changed to use prev epoch for evicted epoch when genesis and restore layer are in the same epoch,

tortoise/recover.go
countvonzero: done

tortoise/algorithm.go
```diff
@@ -120,6 +120,11 @@ func (t *Tortoise) OnBeacon(eid types.EpochID, beacon types.Beacon) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	evicted := t.trtl.evicted.GetEpoch()
+	// when the node recover from checkpoint, the genesis epoch and recover epoch can be the same
+	if types.GetEffectiveGenesis() != types.FirstEffectiveGenesis() &&
+		evicted == (types.GetEffectiveGenesis()+1).GetEpoch() {
+		evicted--
```
dshulyak, line 126: this doesn't look good

is it possible to initialize tortoise state correctly? right now t.evicted is set to genesis - 1. what it needs to be set so that it will work both for recovery and normal case?

tortoise/recover.go
```diff
@@ -25,6 +25,18 @@ func Recover(db *datastore.CachedDB, beacon system.BeaconGetter, opts ...Opt) (*
...
+		first := types.GetEffectiveGenesis().GetEpoch()
+		for _, epoch := range []types.EpochID{first, first + 1} {
+			if err := db.IterateEpochATXHeaders(epoch, func(header *types.ActivationTxHeader) bool {
+				trtl.OnAtx(header)
+				return true
+			}); err != nil {
+				return nil, err
+			}
+		}
+	}
```
dshulyak: there is a function recoverEpoch now

i think the only thing that is missing is:


```
if ... {
      recoverEpoch(first)
}
```

first+1 will be recovered either if latest is 0, or normally in the loop

tortoise/algorithm.go
dshulyak: in order for eviction to work correctly - evicted needs to be either 0 or genesis - epoch_size - 1. otherwise it breaks state consistency, as there will be an epoch that is below evicted.

i guess the problem is that evicted is also used in loops where weight is counted
there can be `first` and it will be equal to genesis.Sub(1) and `evicted` will be 0. and both will move forward on eviction

tortoise/tortoise_test.go
```diff
@@ -985,6 +985,33 @@ func tortoiseVotingWithCurrent(tortoise voter) sim.VotesGenerator {
 	}
 }
 
+func TestOnBeacon(t *testing.T) {
+	s := sim.New()
```
dshulyak: sim is not used, why not just New(...) ?

tortoise/algorithm.go
countvonzero: note that `evicted` here is a local variable for epoch, i am not changing `t.trtl.evicted`.

took a closer look at the eviction logic. it seems that the current code at develop isn't consistent.
say epoch size is 10. window size is 100. ideally data for epoch 2 should be removed when t.trtl.evicted == 29.
but the current code will disregard `OnBeacon()` for epoch 2 when t.trtl.evicted == 20.

i think the code should simply change to 
```go
evicted := t.trtl.evicted.Add(1).Epoch()  // we still care about t.trtl.evicted.Add(1)
if eid < evicted {  // note "<" instead of "<<"
  return
} 
```
this will also work for the recovery case.
the calculation in `lookbackWindowStart()` needs to be adjusted for recovery tho.

UPD: `lookbackWindowStart()` seems fine as is.

dshulyak: > evicted := t.trtl.evicted.Add(1).Epoch()  // we still care about t.trtl.evicted.Add(1)
if eid < evicted {  // note "<" instead of "<<"
  return
} 

yes, this looks good to me as well

dshulyak (approved):

countvonzero:
bors try

countvonzero:
bors merge

PR #4408, opened 2023-05-19
[Merged by Bors] - Offload PoST verification to pool of verifying workers

### LINKED ISSUE #4388: atx: rate limit number of in-flight verifications (default to 3/4 of cores rounding up)
results from running 1k network. at the start of cycle gap - node started to spin up multiple threads (assuming for verification). many nodes crushed with oom, as i set 2GB as a hard limit for every smesher.

image

image

## proposed solution

we need to rate limit activations that are verified at the same time, to facilitate thread reuse. we could reuse libp2p/gossipsub code for that, but we can't do so at the moment due to async nature of verification (e.g it can ask peers over the network for missing data).

the proposal is to implement simple rate limiter, and allow configurable number of threads. defaulting to 3/4 of the cpu count, rounding up.

poszu:
## Motivation
Closes #4388 

## Changes
Created a layer between nipostValidator and verifying code in https://github.com/spacemeshos/post that offloads verifying PoST onto a pool of workers. A free worker picks up proof, verifies it, and sends back the result via a provided channel.

## Test Plan
- [x] Existing tests should pass
- [ ] Survive the big 1k nodes network

### DIFF
--- a/activation/post.go
+++ b/activation/post.go
@@ -66,8 +66,12 @@ type PostProofVerifyingOpts struct {
 }
 
 func DefaultPostVerifyingOpts() PostProofVerifyingOpts {
+	workers := runtime.NumCPU() * 3 / 4
+	if workers < 1 {
+		workers = 1
+	}
 	return PostProofVerifyingOpts{
-		Workers: runtime.NumCPU() * 3 / 4,
+		Workers: workers,
 	}
 }
 
--- a/activation/post.go
+++ b/activation/post.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime"
 	"sync"
 
 	"github.com/spacemeshos/post/config"
@@ -55,6 +56,18 @@ func DefaultPostProvingOpts() PostProvingOpts {
 	return PostProvingOpts{
 		Threads: 1,
 		Nonces:  16,
+	}
+}
+
+// PostProvingOpts are the options controlling POST proving process.
+type PostProofVerifyingOpts struct {
+	// Number of workers spawned to verify proofs.
+	Workers int `mapstructure:"smeshing-opts-verifying-workers"`
+}
+
+func DefaultPostVerifyingOpts() PostProofVerifyingOpts {
+	return PostProofVerifyingOpts{
+		Workers: runtime.NumCPU() * 3 / 4,
 	}
 }
 
--- a/activation/post_verifier.go
+++ b/activation/post_verifier.go
@@ -1,0 +1,106 @@
+package activation
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spacemeshos/post/config"
+	"github.com/spacemeshos/post/shared"
+	"github.com/spacemeshos/post/verifying"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+type verifyPostJob struct {
+	proof    *shared.Proof
+	metadata *shared.ProofMetadata
+	opts     []verifying.OptionFunc
+	result   chan error
+}
+
+type OffloadingPostVerifier struct {
+	eg      errgroup.Group
+	log     log.Log
+	workers []*postVerifierWorker
+	channel chan<- *verifyPostJob
+}
+
+type postVerifierWorker struct {
+	verifier PostVerifier
+	log      log.Log
+	channel  <-chan *verifyPostJob
+}
+
+type verifierFunc func(*shared.Proof, *shared.ProofMetadata, ...verifying.OptionFunc) error
+
+func (f verifierFunc) Verify(p *shared.Proof, m *shared.ProofMetadata, opts ...verifying.OptionFunc) error {
+	return f(p, m, opts...)
+}
+
+// NewPostVerifier creates a new post verifier.
+func NewPostVerifier(cfg PostConfig, logger log.Log) PostVerifier {
+	c := config.Config(cfg)
+	verify := func(p *shared.Proof, m *shared.ProofMetadata, opts ...verifying.OptionFunc) error {
+		logger.With().Debug("verifying post", log.FieldNamed("proof_node_id", types.BytesToNodeID(m.NodeId)))
+		return verifying.Verify(p, m, c, logger.Zap(), opts...)
+	}
+	return verifierFunc(verify)
+}
+
+// NewOffloadingPostVerifier creates a new post proof verifier with the given number of workers.
+// The verifier will distribute incoming proofs between the workers.
+// It will block if all workers are busy.
+func NewOffloadingPostVerifier(verifiers []PostVerifier, logger log.Log) *OffloadingPostVerifier {
+	numWorkers := len(verifiers)
+	channel := make(chan *verifyPostJob, numWorkers)
+	workers := make([]*postVerifierWorker, 0, numWorkers)
+
+	for i, verifier := range verifiers {
+		workers = append(workers, &postVerifierWorker{
+			verifier: verifier,
+			log:      logger.Named(fmt.Sprintf("worker-%d", i)),
+			channel:  channel,
+		})
+	}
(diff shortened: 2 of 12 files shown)


activation/post.go
```diff
@@ -58,6 +59,18 @@ func DefaultPostProvingOpts() PostProvingOpts {
...
 
+// PostProvingOpts are the options controlling POST proving process.
+type PostProofVerifyingOpts struct {
+	// Number of workers spawned to verify proofs.
+	Workers int `mapstructure:"smeshing-opts-verifying-workers"`
+}
+
+func DefaultPostVerifyingOpts() PostProofVerifyingOpts {
+	return PostProofVerifyingOpts{
+		Workers: runtime.NumCPU() * 3 / 4,
```
dshulyak: what about 1 cpu? without rounding up - it will be 0

dshulyak: maybe it is sufficient to set min to 1?

activation/post_verifier.go
```diff
@@ -0,0 +1,106 @@
...
+func (v *OffloadingPostVerifier) Verify(p *shared.Proof, m *shared.ProofMetadata, opts ...verifying.OptionFunc) error {
+	job := &verifyPostJob{
+		proof:    p,
+		metadata: m,
+		opts:     opts,
+		result:   make(chan error),
+	}
+	v.channel <- job
+	return <-job.result
+}
```
fasmat: This method will block if there are more than `numWorkers` calls to it.

This ensures that only a certain amount of ATXs are verified at the same time but it doesn't prevent a large number of go-routines waiting for this method to return when a lot of ATX are being processed at the same time.

poszu: Yes. The problem was not the number of goroutines but the fact that Go spawned a thread for the long-running verification process. Now there are max `numWorkers` concurrent verifications going on at the same time.

poszu:
bors try

activation/post_verifier.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (v *OffloadingPostVerifier) Verify(p *shared.Proof, m *shared.ProofMetadata, opts ...verifying.OptionFunc) error {
+	job := &verifyPostJob{
+		proof:    p,
+		metadata: m,
+		opts:     opts,
+		result:   make(chan error),
+	}
+	v.channel <- job
```
dshulyak: i think this will get stuck if workers were stopped concurrently with the call to verify

there should be something like
```
select {
   case <-terminated:
        return errTerminated (or context error)
   case v.channel <- job:
}

select {
    <-job.result:
    <-terminated:
}
```

so it would be better not to pass context to Start and instead control lifecycle with Start()/Stop() and initialize something for termination

activation/post_verifier.go
```diff
@@ -0,0 +1,106 @@
...
+		workers: workers,
+		channel: channel,
+	}
+}
+
+func (v *OffloadingPostVerifier) Start(ctx context.Context) {
+	v.log.Info("starting post verifier")
+	for _, worker := range v.workers {
+		worker := worker
+		v.eg.Go(func() error { worker.start(); return nil })
```
dshulyak: it can really be just a function instead of worker struct

activation/post_verifier.go
dshulyak: actually not stuck it will crush because of close(v.channel) and then send on closed channel

activation/post_verifier.go
```diff
@@ -0,0 +1,106 @@
...
+		result:   make(chan error),
+	}
+	v.channel <- job
+	return <-job.result
+}
+
+func (w *postVerifierWorker) start() {
+	w.log.Info("starting post proof verifier worker")
+	for job := range w.channel {
+		job.result <- w.verifier.Verify(job.proof, job.metadata, job.opts...)
```
dshulyak: and then if you fix problem outlined above, this may get stuck if termination was concurrent with w.verifier.Verify call. so job result either has to be with capacity = 1, or this should also select on two channels:


```
select {
    job.result <- w.verifier.Verify(job.proof, job.metadata, job.opts...)
    <-job.terminated:
}
```

activation/post_verifier.go
fasmat: `start` is a blocking call and the `errGroup` ensures that if the service is stopped all workers are stopped as well.

That being said, I think passing the context into the `worker.start()` method and checking there:

```golang
for {
	select {
	case job := <-w.channel: {
		job.result <- w.verifier.Verify(job.proof, job.metadata, job.opts...)
	case <-ctx.Done() {
		return
	}
}
```

makes the code a bit easier. The `v.channel` then doesn't need to be closed. At the moment if the `ctx` from `Start(ctx)` is closed and `Verify` is called the code will panic. With my suggestion the `Verify` method instead would block.

This either could be OK (the app is exiting anyway) or `Verify` needs to be updated as well. Probably by also receiving a `ctx` - which might be a good idea anyway if `Verify` can take a longer time.

activation/post_verifier.go
fasmat: I was thinking the same, but with a different solution 🙂 : https://github.com/spacemeshos/go-spacemesh/pull/4408/files/55ba220f4d4806c8237af96d08a8e4835aea7f0f#r1199031986

activation/post_verifier.go
dshulyak: that seems to work too. what i meant is that worker.start can be just a function, instead of a method on worker struct

dshulyak: > Probably by also receiving a ctx - which might be a good idea anyway if Verify can take a longer time.

the question here is what context will be used by Verify, as that method is called by libp2p handler. not sure if it cancels context when we stop the node and in what order

fasmat: I think `libp2p` is context aware. The Handler that receives the Gossip data receives a ctx with it: `func (h *Handler) HandleGossipAtx(ctx context.Context, peer p2p.Peer, msg []byte)` and eventually calls `Verify`.

The context that's passed into `HandleGossipAtx` I think comes from `libp2p`. If it works similar to how the `net/http` handlers use contexts it will be canceled when either the connection drops or a timeout is reached.

fasmat: But I haven't looked too deep into it and we should confirm it first.

activation/post_verifier.go
poszu: Thanks for pointing that out! I will restrain from closing the channel to workers and use a context to signal workers to stop instead.

activation/post_verifier.go
poszu: Good catch! IMO, actually both "precautions" should be implemented. The `result` channel should be buffered to avoid a worker being stuck infinitely if the sender never reads from it (i.e. its context is canceled and it never reads the result).

It should also select on `ctx.Done()` to allow termination.

activation/post_verifier.go
poszu: I'd rather stick with a separate struct for workers

activation/post_verifier_test.go
```diff
@@ -50,5 +50,39 @@ func TestOffloadingPostVerifier(t *testing.T) {
 
 func TestPostVerfierDetectsInvalidProof(t *testing.T) {
 	verifier := activation.NewPostVerifier(activation.PostConfig{}, log.NewDefault(t.Name()))
-	require.Error(t, verifier.Verify(&shared.Proof{}, &shared.ProofMetadata{}))
+	require.Error(t, verifier.Verify(context.Background(), &shared.Proof{}, &shared.ProofMetadata{}))
+}
+
+func TestPostVerifierVerifyAfterStop(t *testing.T) {
```
fasmat, line 56: Do you think an additional test where `Verify` doesn't return and the context passed to `Start` is canceled to check if everything shuts down correctly makes sense?

fasmat (approved):

poszu:
Bors try

dshulyak (approved):

dshulyak:
i want to include my change (https://github.com/spacemeshos/go-spacemesh/pull/4383) into next 10k network attempt, so it may take some time. we should probably merge this one before the test

poszu:
bors merge

PR #4410, opened 2023-05-19
add genesis id and first layer as p2p protocol prefix 

countvonzero:
## Motivation
part of #4090 

## Changes
- p2p handshake: add mesh first layer to handshake message
- gossip: add genesis id and mesh first layer into the gossip topic
- fetch: add mesh first layer into protocol

### DIFF
--- a/p2p/handshake/handshake.go
+++ b/p2p/handshake/handshake.go
@@ -35,7 +35,7 @@ func WithLog(logger log.Log) Opt {
 }
 
 const (
-	hsprotocol    = "/handshake/1"
+	hsprotocol    = "handshake/1"
 	streamTimeout = 10 * time.Second
 )
 
@@ -52,18 +52,19 @@ type HandshakeAck struct {
 }
 
 // New instantiates handshake protocol for the host.
-func New(h host.Host, genesisID types.Hash20, opts ...Opt) *Handshake {
+func New(h host.Host, genesisID types.Hash20, prefix string, opts ...Opt) *Handshake {
 	ctx, cancel := context.WithCancel(context.Background())
 	hs := &Handshake{
 		logger:    log.NewNop(),
 		genesisID: genesisID,
+		protocol:  protocol.ID(fmt.Sprintf("%s/%s", prefix, hsprotocol)),
 		h:         h,
 		cancel:    cancel,
 	}
 	for _, opt := range opts {
 		opt(hs)
 	}
-	h.SetStreamHandler(protocol.ID(hsprotocol), hs.handler)
+	h.SetStreamHandler(hs.protocol, hs.handler)
 	emitter, err := h.EventBus().Emitter(new(EventHandshakeComplete))
 	if err != nil {
 		hs.logger.With().Panic("failed to initialize emitter for handshake", log.Err(err))
@@ -83,6 +84,7 @@ type Handshake struct {
 
 	emitter   event.Emitter
 	genesisID types.Hash20
+	protocol  protocol.ID
 	h         host.Host
 
 	cancel context.CancelFunc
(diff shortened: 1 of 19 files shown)


dshulyak (commented):
lets try to keep this change isolated. it tries to modify protocol, but still didn't do it in a couple of places for no good reason. beside the inconsistency, it also spills protocol-prefixing across the codebase making the end result look messy

i propose:
- remove all changes across spacemesh-specific codebase
- configure protocolPrefix for p2p.Host during initialization (maybe no need to use full genesis id, as it will be very verbose)
- prefix everything in the p2p.Host
 
technically it can be done by implementing SetStreamHandler and NewStream methods of the p2p.Host interface. pubsub topics actually can be kept as is, as they will be using prefixed streams

p2p/handshake/handshake.go
```diff
@@ -43,7 +43,8 @@ const (
 
 // HandshakeMessage is a handshake message.
 type HandshakeMessage struct {
-	GenesisID types.Hash20
+	GenesisID  types.Hash20
+	FirstLayer types.LayerID
```
dshulyak: we can generalize it and define libp2p stream protocol that looks as /genesis/layer/hs/1. then naturally if we update genesis id or layer id - peers won't be able to setup handshake

dshulyak:
or maybe better alternative to keep it isolated is to add some method to p2p.Host, like PrefixedProtocol(protocol) string

countvonzero:
> pubsub topics actually can be kept as is, as they will be using prefixed streams

for SM-specific code that calls SetStreamHandler and NewStream, yes. i added these impl in p2p.Host and prefix the protocol.
but i don't see how i can influence pubsub component to use the prefixed streams. pubsub will use the libp2p host.Host interface directly. can you help me understand @dshulyak 

the PR is here https://github.com/spacemeshos/go-spacemesh/pull/4412 but i cannot verify that gossip will use prefixed streams.

countvonzero:
> the PR is here https://github.com/spacemeshos/go-spacemesh/pull/4412 but i cannot verify that gossip will use prefixed streams.

ok. i get it now. i passed p2p.Host down to pubsub and discovery (instead of host.Host from libp2p)

countvonzero:
closed in favor of https://github.com/spacemeshos/go-spacemesh/pull/4412

countvonzero:
bors try

p2p/pubsub/wrapper.go
```diff
@@ -63,7 +70,8 @@ func (ps *PubSub) Register(topic string, handler GossipHandler) {
 }
 
 // Publish message to the topic.
-func (ps *PubSub) Publish(ctx context.Context, topic string, msg []byte) error {
+func (ps *PubSub) Publish(ctx context.Context, protocol string, msg []byte) error {
+	topic := makeTopic(ps.topicPrefix, protocol)
```
dshulyak, line 74: we capture this topic in metrics, how it will look like in prometheus?

p2p/handshake/handshake.go
```diff
@@ -52,18 +52,19 @@ type HandshakeAck struct {
...
 		genesisID: genesisID,
+		protocol:  protocol.ID(fmt.Sprintf("%s/%s", prefix, hsprotocol)),
 		h:         h,
 		cancel:    cancel,
 	}
 	for _, opt := range opts {
 		opt(hs)
 	}
-	h.SetStreamHandler(protocol.ID(hsprotocol), hs.handler)
+	h.SetStreamHandler(hs.protocol, hs.handler)
```
dshulyak, line 67: i reliazed that i don't know what happens if node connects, and it doesn't have any protocols in common (because of prefix), then what? will it remain using a connection or not?

countvonzero: ClosePeer() will be called on this peer. the doc stated.

// ClosePeer closes the connection to a given peer

btw, i wonder if this handshake protocol is still required after you added the noise prologue? is it possible we can remove this handshake all together?

p2p/pubsub/wrapper.go
countvonzero: it will look like "/xxxxx/1234/bo" for protocol "bo". are you concerned about the data?
to me it's nice if we will be able to see cross-protocol gossip metrics

btw, i will make this prefix value the same as in the noise prologue.

countvonzero (commented):
here is the metrics for cpu/mem usage. from left to right

- #4471 develop #4410

Screenshot from 2023-06-07 11-15-02

Screenshot from 2023-06-07 11-14-56

countvonzero:
closed as superseded by https://github.com/spacemeshos/go-spacemesh/pull/4485

PR #4412, opened 2023-05-23
[Merged by Bors] - p2p: automatically prefix p2p protocol id with genesis id and layer

countvonzero:
## Motivation
part of #4090 
alternative impl of #4410

## Changes
prefix p2p protocol with "/genesisid/layer"


cmd/node/node.go
```diff
@@ -1200,7 +1200,7 @@ func (app *App) Start(ctx context.Context) error {
 	p2plog := app.addLogger(P2PLogger, lg)
 	// if addLogger won't add a level we will use a default 0 (info).
 	cfg.LogLevel = app.getLevel(P2PLogger)
-	app.host, err = p2p.New(ctx, p2plog, cfg, app.Config.Genesis.GenesisID(),
+	app.host, err = p2p.New(ctx, p2plog, cfg, app.Config.Genesis.GenesisID(), types.GetEffectiveGenesis()+1,
```
dshulyak: i think it will be cleaner to pass only string instead of `app.Config.Genesis.GenesisID(), types.GetEffectiveGenesis()+1` to p2p.New

dshulyak (approved):

cmd/node/node.go
piersy: Passing a string seems more general, if we ever wanted to run more than one host, the string could be useful.

countvonzero: genesis ID is still used in handshake msg. i didn't want to change the logic there.

> if we ever wanted to run more than one host, the string could be useful

can you elaborate?

piersy: You may want to run a host that doesn't have the genesis id or effective genesis in it's protocol prefix, so having a string would allow for that

countvonzero: > You may want to run a host that doesn't have the genesis id or effective genesis in it's protocol prefix, so having a string would allow for that

i am curious about scenarios you are describing. can you give an example of why we want to run different hosts with different configuration?

countvonzero:
bors try

cmd/node/node.go
piersy: No I can't, I'm just thinking it makes for a more general interface, and I'm pretty bad at predicting the future, so I find general interfaces useful when things change in ways that I wasn't expecting.

Intuitively though having the p2p package (a networking package) know about genesis ids and layers, seems to be tying it into knowledge that is out of it's domain. But as you pointed out it was already tied in due to the use in the handshake.

countvonzero:
bors try

countvonzero:
bors merge

PR #4414, opened 2023-05-23
systest: reenable partition tests

### LINKED ISSUE #4402: systest: enable partition tests
they were disabled in https://github.com/spacemeshos/go-spacemesh/pull/4393 to make ci non-flaky.

by the looks nodes take very long to reach tortoise confidence threshold, after partition. not clear where is the problem

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4402


dshulyak:
bors try

PR #4428, opened 2023-05-25
[Merged by Bors] - tortoise: dont skip genesis data

dshulyak:
there was a codepath that didn't load any data if node was restarted before the first ballot was sent into the network


fasmat (approved):
What's the reasoning for the including the earlier layers? Should it always have started from 0 and we just overlooked it or did requirements change?

dshulyak:
@fasmat i overlooked that it can be restarted during the genesis, so the broken part was

> if latest <= types.GetEffectiveGenesis() {
		return trtl, nil

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
https://github.com/spacemeshos/go-spacemesh/issues/4433

dshulyak:
bors merge

PR #4431, opened 2023-05-25
[Merged by Bors] - fetch: fix out of memory issue in handleMeshHashReq

### LINKED ISSUE #4427: fetch: FuzzMeshHashRequest crashes with out of memory error.
## Description
As part of the work on https://github.com/spacemeshos/go-spacemesh/issues/4439 the following issue was identified:

When fuzzing `FuzzMeshHashRequest` eventually the test will crash with a generic error message:

```shell
--- FAIL: FuzzMeshHashRequest (24.30s)
    fuzzing process terminated by unexpected signal; no crash will be recorded: signal: killed
```

## Steps to reproduce
1. Run `go test ./fetch -v -run=FuzzMeshHashRequest -fuzz=^FuzzMeshHashRequest$` for at least 30 seconds
2. Observe memory consumption of the process (keeps increasing)
3. Eventually `go test` will be terminated by the OS because it consumes more memory than available.

## Actual Behavior
Process termination due to high memory consumption

## Expected Behavior
Fuzzing should continue without a crash

## Environment
 - OS: Linux
 - Node Version: develop `2a36d0455f024bb7fcb2d126bc096526211ee292`

## Additional Info 

The culprit is seems to be `layers.GetAggHashes`. It will cause a crash by building a very large SQL query:

```golang
func TestGetAggHashesOom(t *testing.T) {
	db := sql.InMemory()

	lids := make([]types.LayerID, 0, math.MaxUint32)
	for i := 0; i < math.MaxUint32; i++ {
		lids = append(lids, types.LayerID(i))
	}

	GetAggHashes(db, lids)
}
```

fasmat:
## Motivation
Closes #4427

## Changes
- Prevent requests for more than 1000 layers at a time
- Earlier / quicker check for `uint32` overflow

The overflow error could be mitigated by changing the request to only contain `From`, `To` and `Delta`/`By`, since `Steps` can be derived from the other 3. Then only the following checks would suffice to prevent overflows:

- `From` < `To`
- `Delta` <= `To` - `From`

I didn't include this change because because I'm not sure if non-backwards compatible changes are still OK?

## Test Plan
- `FuzzMeshHashRequest` would crash or panic within 15 - 20 seconds before (on my machine) now it doesn't find any issues even after 60 seconds
- Memory consumption during testing doesn't increase above 6 GB any more (before it would rise to > 16 GB before it crashes)

## TODO

## DevOps Notes

### DIFF
--- a/fetch/mesh_data.go
+++ b/fetch/mesh_data.go
@@ -229,6 +229,16 @@ func (f *Fetch) PeerEpochInfo(ctx context.Context, peer p2p.Peer, epoch types.Ep
 }
 
 func iterateLayers(req *MeshHashRequest) ([]types.LayerID, error) {
+	// TODO(mafa): there is MaxHashesInReq config in syncer that could be passed here instead of hardcoding this value
+	if req.Steps == 0 || req.Steps > 100 {
+		return nil, fmt.Errorf("%w: %v", errBadRequest, req)
+	}
+
+	// check for overflow
+	if req.Delta == 0 || (req.Delta*req.Steps)/req.Delta != req.Steps {
+		return nil, fmt.Errorf("%w: %v", errBadRequest, req)
+	}
+
 	var diff uint32
 	if req.To.After(req.From) {
 		diff = req.To.Difference(req.From)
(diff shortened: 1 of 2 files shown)


dshulyak (approved):

fasmat:
bors merge

dshulyak:
i opened a bug about latest failure https://github.com/spacemeshos/go-spacemesh/issues/4433 . it failed this day on 1 or 2 changes, no idea what is causing this

fasmat:
bots merge

dshulyak:
same failure

fasmat:
> same failure

I'll investigate to check if the change is the problem or if it's unrelated.

fasmat:
bors merge

dshulyak:
bors merge

PR #4438, opened 2023-05-26
[Merged by Bors] - tortoise: more expressive tools for writing tests

dshulyak:
unlike previous testing framework for tortoise this one allows to manually define all objects (and objects properties) and the order in which tortoise processes them. the goal is to enable writing exact execution patterns without significant effort. additionally it helps to find ordering issues (such as missing adding ballot weight to block weight, if ballot was received before block) by computing random topological order and executing same test multiple times with random order. it can also be extended to compute all topological order for better and determinstic coverage.


tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+		for _, vote := range layer {
+			hasher.WriteSupport(vote.ID, vote.Height)
+		}
+		hasher.Sum(rst[:0])
+		hasher.Reset()
+	}
+	return rst
+}
+
+func (b *ballotAction) decodeVotes(votes [][]types.Vote) {
```
countvonzero: interesting. i'd call this encodeVotes, as you are populating `votes` from `b.Opinion`.

tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+	if b.base != nil {
+		b.base.decodeVotes(votes)
+	}
+	genesis := types.GetEffectiveGenesis()
+	for _, vote := range b.Opinion.Support {
+		pos := vote.LayerID.Difference(genesis)
+		matches := false
+		for _, existing := range votes[pos] {
+			if existing == vote {
+				matches = true
```
countvonzero, line 276: break?

tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+		}
+		if !matches {
+			votes[pos] = append(votes[pos], vote)
+		}
+	}
+	for _, vote := range b.Opinion.Against {
+		pos := vote.LayerID.Difference(genesis)
+		for i, existing := range votes[pos] {
+			if existing == vote {
+				votes[pos] = append(votes[pos][:i], votes[pos][i:]...)
```
countvonzero: do you mean `votes[pos] = append(votes[pos][:i], votes[pos][i+1:]...)`

tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+	effective := types.GetEffectiveGenesis()
+	epochSize := types.GetLayersPerEpoch()
+	tb.Cleanup(func() {
+		types.SetEffectiveGenesis(effective.Uint32())
+		types.SetLayersPerEpoch(epochSize)
+	})
+	s := &session{tb: tb}
+	return s.
+		withLayerSize(50).
+		withEpochSize(10).
```
countvonzero, line 325: `epochSize`?

tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+	// or optionally a subset of them if all takes too long
+	rst := []action{}
+	empty := []action{}
+	forward := map[action]map[action]struct{}{}
+	reverse := map[action]map[action]struct{}{}
+	for _, act := range s.actions {
+		for _, dep := range act.deps() {
+			if reverse[act] == nil {
+				reverse[act] = map[action]struct{}{}
+			}
```
countvonzero, line 570: move to the outer loop?

tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+	s.tally(1)
+	s.hareblock(1, "aa", 0)
+	for i := 0; i < n; i++ {
+		s.smesher(i).atx(1).ballot(2, new(bopt).
+			eligibilities(s.layerSize/n).
+			votes(new(evotes).support(1, "aa", 0)),
+		)
+	}
+	s.hareblock(2, "bb", 0)
+	s.tallyWait(2)
```
countvonzero, line 678: can you explain when you would use `tally()` vs `tallyWait()` in test?

tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+	b := types.BallotTortoiseData{}
+	b.AtxID = a.header.ID
+	b.Layer = types.LayerID(lid)
+	hs := hash.Sum(a.header.ID[:], []byte(strconv.Itoa(int(lid))))
+	copy(b.ID[:], hs[:])
+
+	val := &ballotAction{
+		BallotTortoiseData: b,
+	}
+	for _, opt := range opts {
```
countvonzero, line 187: options relating to `EpochData` is incompatible if a.reference != nil and should be set conditionally to be more realistic?

tortoise/fixture_test.go
```diff
@@ -0,0 +1,791 @@
...
+			eligibilities(s.layerSize/n).
+			votes(new(evotes).support(1, "aa", 0)),
+		)
+	}
+	s.hareblock(2, "bb", 0)
+	s.tallyWait(2)
+	s.updates(t, new(results).
+		verified(0).
+		verified(1).block("aa", 0, valid|hare|data).
+		next(2).block("bb", 0, hare|data),
```
countvonzero, line 682: is there a scenario where `hare` and `data` are different value?

countvonzero (approved):

tortoise/fixture_test.go
dshulyak: i didn't get it. can you clarify?

tortoise/fixture_test.go
dshulyak: tallyWait waits for all previous events to finish, so it cannot be reorded if executed in random topological order. tally can be reorder

tortoise/fixture_test.go
dshulyak: you mean in whole application? data can be required even if block is not confirmed by hare, hare might be true and data is false if there is a bug elsewhere and node didn't save the block but saved hare output

tortoise/fixture_test.go
dshulyak: Opinion is an encoded form of votes, atleast i referred to it so far and tortoise has EncodeVotes method

tortoise/fixture_test.go
dshulyak: yes, thanks

tortoise/fixture_test.go
dshulyak: i will set Ref only if EpochData is nil

tortoise/fixture_test.go
dshulyak: realized that reverse should have no key is act.deps() is empty. moving to outer loop breaks it

dshulyak:
bors merge

tortoise/fixture_test.go
countvonzero: i meant you already defined epochSize on line 316

PR #4442, opened 2023-05-27
[Merged by Bors] - logging: revisit verbosity

dshulyak:
there are multiple logs that print duplicate information on progress, this significantly decreases readability.
for example optimistic execution printed 3-4 logs with similar info, which is clearly useful only for debug. if thats necessary developer should raise that module to debug level.

in some other cases metric is a cheaper and often better alternative, e.g a count of some object or computed value.

### DIFF
--- a/miner/oracle.go
+++ b/miner/oracle.go
@@ -154,7 +154,7 @@ func (o *Oracle) calcEligibilityProofs(atx *types.ActivationTxHeader, epoch type
 		)
 	}
 
-	logger.With().Info("tortoise eligibility for an epoch",
+	logger.With().Info("proposal eligibility for an epoch",
 		log.Uint32("total_num_slots", numEligibleSlots),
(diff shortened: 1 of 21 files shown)


miner/oracle.go
```diff
@@ -154,7 +154,7 @@ func (o *Oracle) calcEligibilityProofs(atx *types.ActivationTxHeader, epoch type
 		)
 	}
 
-	logger.With().Info("proposal eligibility calculated",
+	logger.With().Info("tortoise eligibility for an epoch",
```
countvonzero: i am not sure tortoise eligibility is better than proposal eligibility

countvonzero (approved):

miner/oracle.go
dshulyak: reverted

dshulyak:
bors merge

PR #4444, opened 2023-05-28
[Merged by Bors] - tortoise: don't evict until updates are consumed

dshulyak:


tortoise/tortoise_test.go
```diff
@@ -1052,6 +1053,7 @@ func TestBaseBallotEvictedBlock(t *testing.T) {
 		tortoise.TallyVotes(ctx, last)
 		verified = tortoise.LatestComplete()
 		require.Equal(t, last.Sub(1), verified)
+
```
countvonzero: i am surprised empty line at the end doesn't fail lint.

countvonzero (approved):

tortoise/tortoise_test.go
dshulyak: likely because it is inside the for loop. i removed

dshulyak:
bors merge

PR #4447, opened 2023-05-29
[Merged by Bors] - tortoise: tracer to debug execution interactively

dshulyak:
tortoise provides EnableTracer option, if enabled it will dump to stderr all inputs and outputs from tortoise consensus using zap library.

this logs will be available in cloud and can be downloaded either with grafana or logcli. if downloaded from grafana, next post processing script is useful:
> jq -c '.[].line | fromjson | {"t": .t, "o": .o}' ./EXAMPLE.json

note that traces are verbose and should be decreased in size, as an example i added tortoise/data/failed_nodes.json from one of the nodes.

once trace is collected it can be used as a regression test, and also for interactive debugging. for interactive debugging:
1. install delve https://github.com/go-delve/delve/tree/master/Documentation/installation
2. run delve with selected trace `dlv exec ./trace -- -breakpoint -level=debug ./tortoise/data/failed_nodes.json`
3. continue will stop after every step, debug log will also print what happened during that step
```
> (dlv) continue
2023-05-30T11:42:03.864+0200	DEBUG	trace	using trace	{"path": "./tortoise/data/failed_nodes.json", "name": ""}
> [hardcoded-breakpoint] github.com/spacemeshos/go-spacemesh/tortoise.RunTrace() ./tortoise/tracer.go:101 (hits total:0) (PC: 0xdf038f)
Warning: debugging optimized function
    96:				return err
    97:			}
    98:			if err := ev.Run(runner); err != nil {
    99:				return err
   100:			}
=> 101:			if breakpoint != nil {
   102:				breakpoint()
   103:			}
   104:		}
   105:	}
   106:	
```


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors cancel

dshulyak:
bors try

cmd/node/node.go
```diff
@@ -600,10 +600,16 @@ func (app *App) initServices(ctx context.Context, poetClients []activation.PoetP
 	if trtlCfg.BadBeaconVoteDelayLayers == 0 {
 		trtlCfg.BadBeaconVoteDelayLayers = app.Config.LayersPerEpoch
 	}
-	trtl, err := tortoise.Recover(app.cachedDB, beaconProtocol,
-		tortoise.WithContext(ctx),
```
fasmat, line 604: I think this functional option might have been accidentally dropped.

tortoise/algorithm.go
```diff
@@ -231,12 +288,24 @@ type DecodedBallot struct {
 	minHint types.LayerID
 }
 
-// DecodeBallot decodes ballot if it wasn't processed earlier.
+// decodeBallot decodes ballot if it wasn't processed earlier.
```
fasmat: Why small caps when the function is public?

tortoise/algorithm.go
```diff
@@ -281,22 +353,25 @@ func (t *Tortoise) OnHareOutput(lid types.LayerID, bid types.BlockID) {
...
+		missing = atxs
+	} else {
+		for _, atx := range atxs {
+			_, exists := edata.atxs[atx]
+			if !exists {
+				missing = append(missing, atx)
+			}
 		}
 	}
 	return missing
```
fasmat: I personally think the previous code was easier to understand than the changed code (functional they seem to be identical?)

cmd/node/node.go
dshulyak: not accidental, i need to remove WithContext completely, it was used long time ago

tortoise/algorithm.go
piersy: I agree, I don't see a reason for the else which creates additional nesting and makes it harder for me to read.

dshulyak: the reason for this was that i added a trace for this method as well, but then decided to remove it temporarily.  with nesting i had to add trace in one place

dshulyak:
bors try

piersy:
Hi @dshulyak I'm not familiar with interactive debugging, could you explain how you would go about interactively debugging a trace, such as the failed nodes trace?

dshulyak:
> I'm not familiar with interactive debugging, could you explain how you would go about interactively debugging a trace, such as the failed nodes trace?

i tried to explain in the pr description

tortoise/tracer.go
```diff
@@ -0,0 +1,487 @@
...
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				return nil
+			}
+			return err
+		}
+		if err := ev.Run(runner); err != nil {
+			return err
+		}
+		if breakpoint != nil {
```
piersy, line 101: Having the breakpoint here means that you can't manipulate the event contents. It seems like that would be something that could be useful in interactive debugging.

dshulyak: good idea. no reason to prevent that. will more it above

tortoise/tortoise_test.go
```diff
@@ -3053,3 +3054,20 @@ func TestUpdates(t *testing.T) {
 		require.Equal(t, id, updates[0].Blocks[0].Header.ID)
 	})
 }
+
+func TestData(t *testing.T) {
```
piersy, line 3058: I think we need some better description of what these traces are, possibly a description field at the top of the trace. The existing trace `failed_nodes.json` makes it sound like there is some failure that occurs in the trace, yet the trace is being run without error here, so what behaviour is the trace exhibiting?

dshulyak: that a trace from the system test with a similar name

tortoise/algorithm.go
dshulyak: reverted

tortoise/algorithm.go
dshulyak: it is a leftover after refactoring. reverted

tortoise/tracer.go
dshulyak: added breakpoint after ev is assigned

dshulyak: sorry, had to revert it. i creates some issues with access to runner. don't want to debug it now
> (dlv) print runner.trt.trtl
(unreadable could not find loclist entry at 0x37045 for address 0xdea7e

tortoise/tortoise_test.go
```diff
@@ -3053,3 +3054,20 @@ func TestUpdates(t *testing.T) {
...
+	t.Parallel()
+	data, err := filepath.Abs("./data")
```
piersy, line 3061:
```suggestion
	// The files in data are named after the systest tests so for example `failed_nodes.json` is a trace taken
	// from a run of the `TestFailedNodes` systest.
	data, err := filepath.Abs("./data")
```

dshulyak: sorry, this is not useful. i am not going to stick to what you wrote.
i will think about your comment and add something to make it more obvious to others over time

piersy: I don't understand your point about this not being useful? In the absence of anything else this is the only thing in the code that would let someone understand where the traces have come from.

dshulyak: sorry i don't know how to be more clear. i will add more details later

tortoise/algorithm.go
```diff
@@ -237,6 +287,18 @@ func (t *Tortoise) DecodeBallot(ballot *types.Ballot) (*DecodedBallot, error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	waitBallotDuration.Observe(float64(time.Since(start).Nanoseconds()))
+	decoded, err := t.decodeBallot(ballot)
+	if t.tracer != nil {
+		ev := &DecodeBallotTrace{Ballot: ballot}
```
countvonzero: not interested in ballot.IsMalicious?

countvonzero (approved):

tortoise/algorithm.go
dshulyak: thanks, it needs to be tracked for decode as well

dshulyak:
bors merge

PR #4450, opened 2023-05-30
[Merged by Bors] - systest: partition tests

### LINKED ISSUE #4402: systest: enable partition tests

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4402

the result of the partition could be that some block crosses local threshold, but doesn't cross the global threshold. lets say in total weight the fractions would be 55/45. 55 voting in support, 45 voting against the block. once local threshold is crossed  55-side will be always voting in support, 45-side will be seeing that block crosses local threshold, but not global, but won't have data to vote for the block. as such it won't be possible to reach global threshold indefinitely, as it will always remain at ~10 of total weight, as 45-side will never ask for the block.

the change is to force 45-side to ask for the block even if it crossed local threshold.

other changes:
- make traces shorter by removing all unnecessary data (new types to communicate with tortoise) and encoding byte arrays (ids, node) with base64
- enable lfs for tortoise/data/
- collect traces from multiple ci tests for regression testing


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
FailedNodes flake. not related

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

tortoise/tortoise.go
```diff
@@ -402,12 +403,19 @@ func (t *turtle) verifyLayers() {
 			}
 			success = t.full.verify(logger, target)
 		}
+
+		layer := t.layer(target)
 		if !success {
+			// notify mesh in two additional cases:
+			// - if layer was verified, and became undecided
+			// - if layer is undecided outside hdist distance
```
countvonzero, line 411: do you think this will cause unnecessary churn? usually if the best cause block is executed in-place, and only reverted when it was verified invalid.

countvonzero (approved):

tortoise/tortoise.go
dshulyak: in best case this condition has no impact, i doublechecked in [TestSmeshing](https://grafana.spacemesh.dev/explore?orgId=1&left=%7B%22datasource%22:%22loki-gke-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-gke-dev%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bnamespace%3D%5C%22test-cgdz%5C%22,%20pod%3D%5C%22smesher-11-55996f6664-8snm4%5C%22%7D%20%7C%3D%20%5C%22consensus%20results%5C%22%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221685699059384%22,%22to%22:%221685700055021%22%7D%7D)

it will have to do more work than optimal if there is a partition that causes verified layer to go backward (e.g we knew about 100 weight, and then got 100 more weight). but it is still ok, as we already applied necessary block before doing expensive operations

dshulyak:
bors merge

PR #4455, opened 2023-05-31
[Merged by Bors] - syncer: stay in synced state if data sync is successful

countvonzero:
## Motivation
a node should not become out of sync if the network simply has no data, and should be ready to receive gossip when the network was given fallback data.

change syncer to compare last synced layer, which continue to advance despite empty layers, instead of the latest layer with data in the mesh.

tested against devnet-401-short which is empty since genesis.


dshulyak (approved):

countvonzero:
bors try

countvonzero:
bors merge

PR #4457, opened 2023-05-31
[Merged by Bors] - mesh: prevent incorrect latest layer updates on error and unexpected inputs

dshulyak:
revert didn't work correctly if applied layer was set to a lower layer.

this is still not enough to re-enable partition tests, there seems to be another flake

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -393,7 +393,15 @@ func (msh *Mesh) applyResults(ctx context.Context, results []result.Layer) error
 			log.Context(ctx),
 			log.Stringer("applied", target),
 		)
-		msh.setLatestLayerInState(layer.Layer)
+		if latest := msh.LatestLayerInState(); latest > layer.Layer {
+			msh.logger.With().Warning("reverted layer without reverting state",
+				log.Context(ctx),
+				log.Uint32("latest", latest.Uint32()),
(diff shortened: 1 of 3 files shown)


dshulyak:
bors try

dshulyak:
bors try

countvonzero (approved):

mesh/mesh.go
```diff
@@ -219,15 +219,12 @@ func (msh *Mesh) setProcessedLayer(layerID types.LayerID) error {
 // the block in consensus, and reverts state before that layer
 // if such layer is found.
 func (msh *Mesh) ensureStateConsistent(ctx context.Context, results []result.Layer) error {
-	var (
-		changed types.LayerID
-		inState = msh.LatestLayerInState()
```
countvonzero, line 224: the question is... how did inState become incorrect?

dshulyak: i found two ways:
- if loop that updates inState layer is terminated with error, we would update it to some outdated layer
- or if loop terminated earlier, for example we applied 10, 11, 12 (12 because of hare), later hare falls out of hdist window and not yet verified by tortoise because of partition (opinion on that layer is abstain)

dshulyak:
bors merge

PR #4460, opened 2023-06-01
[Merged by Bors] - beacon: add domain to VRF signatures to distinguish between beacon messages

### LINKED ISSUE #4274: beacon: messages should not be decodable as another message type within same domain
we are using same domain for all beacon messages (proposal, first voting, followup, weak coin). from the definitions below it is quite possible that certain messages will be decodable as other messages of the different type (for example first voting, can be decoded as a followup), signature will pass and we may even consider it to be an equivocation in future.

```go
	ValidProposals            []Proposal `scale:"max=1000"` // number of proposals is expected to be under 100, 1000 is a safe upper bound
	PotentiallyValidProposals []Proposal `scale:"max=1000"` // number of proposals is expected to be under 100, 1000 is a safe upper bound
}

	EpochID        types.EpochID
	RoundID        types.RoundID
	VotesBitVector []byte `scale:"max=128"` // 128 bytes = 1024 bits and we limit the number of proposals to 1000
}

type Message struct {
	Epoch        types.EpochID
	Round        types.RoundID
	Unit         uint32
	NodeID       types.NodeID
	VrfSignature types.VrfSignature
}

type ProposalMessage struct {
	EpochID      types.EpochID
	NodeID       types.NodeID
	VRFSignature types.VrfSignature
}
```

We should either add another non-overlapping prefix for every different message within domain, or encode message as a scale enum.

fasmat:
## Motivation
Closes https://github.com/spacemeshos/go-spacemesh/issues/4274

## Changes
- Signatures of different messages can now not be verified any more if the message is decoded for the wrong domain
- This should prevent messages like `FirstVotingMessage` accidentally be parsed and validated as `FollowupVotingMessage`

## Test Plan
- Added tests for new domains
- Updated existing tests to check if the correct domain is used

## TODO

## DevOps Notes

### DIFF
--- a/proposals/util_test.go
+++ b/proposals/util_test.go
@@ -59,7 +59,7 @@ func TestComputeWeightPerEligibility(t *testing.T) {
 func TestComputeWeightPerEligibility_EmptyRefBallotID(t *testing.T) {
 	types.SetLayersPerEpoch(layersPerEpoch)
 	signer, err := signing.NewEdSigner(
-		signing.WithKeyFromRand(rand.New(rand.NewSource(1001))),
+		signing.WithKeyFromRand(rand.New(rand.NewSource(1000))),
 	)
 	require.NoError(t, err)
 	beacon := types.Beacon{1, 1, 1}
@@ -76,7 +76,7 @@ func TestComputeWeightPerEligibility_EmptyRefBallotID(t *testing.T) {
 func TestComputeWeightPerEligibility_FailToGetRefBallot(t *testing.T) {
 	types.SetLayersPerEpoch(layersPerEpoch)
 	signer, err := signing.NewEdSigner(
-		signing.WithKeyFromRand(rand.New(rand.NewSource(1001))),
+		signing.WithKeyFromRand(rand.New(rand.NewSource(1000))),
 	)
 	require.NoError(t, err)
 	beacon := types.Beacon{1, 1, 1}
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -956,7 +956,7 @@ func (pd *ProtocolDriver) sendFirstRoundVote(ctx context.Context, epoch types.Ep
 	if err != nil {
 		pd.logger.With().Fatal("failed to serialize message for signing", log.Err(err))
 	}
-	sig := pd.edSigner.Sign(signing.BEACON, encoded)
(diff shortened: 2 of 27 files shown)


fasmat:
bors try

proposals/util_test.go
```diff
@@ -59,7 +59,7 @@ func TestComputeWeightPerEligibility(t *testing.T) {
 func TestComputeWeightPerEligibility_EmptyRefBallotID(t *testing.T) {
 	types.SetLayersPerEpoch(layersPerEpoch)
 	signer, err := signing.NewEdSigner(
-		signing.WithKeyFromRand(rand.New(rand.NewSource(1001))),
+		signing.WithKeyFromRand(rand.New(rand.NewSource(1000))),
```
piersy, line 62: Why have you changed the source?

piersy (approved):
There is something up with the coverage job, but apart from that it looks good to me!

proposals/util_test.go
fasmat: The test uses signatures to determine eligibility. Since I changed which data goes into a signature the previous identities would not be eligible any more. I changed which identities are being used in the test to some that are eligible for the given active set.

fasmat:
bors merge

dshulyak:
bors cancel

fasmat:
> There is something up with the coverage job, but apart from that it looks good to me!

@dshulyak the reason seems to be this line here: https://github.com/spacemeshos/go-spacemesh/blob/266048c6814c66a636b0aadb84fb5906f8e89da5/tortoise/trace/main.go#L15

If I remove this line `make cover` works without issues. Alternatively we can also update https://github.com/spacemeshos/go-spacemesh/blob/266048c6814c66a636b0aadb84fb5906f8e89da5/Makefile#L5 to

```bash
UNIT_TESTS ?= $(shell go list ./...  | grep -v systest/tests | grep -v tortoise/trace)
```

which is probably the correct approach anyway. I'll create a PR

fasmat:
PR to fix code coverage job: https://github.com/spacemeshos/go-spacemesh/pull/4463

dshulyak (approved):

dshulyak:
bors merge

fasmat:
bors merge

PR #4462, opened 2023-06-02
[Merged by Bors] - set effective genesis correctly

countvonzero:
## Motivation
when a node restart after a checkpoint recovery, it will have the wrong effective genesis


dshulyak (approved):

countvonzero:
bors merge

PR #4464, opened 2023-06-02
[Merged by Bors] - Randomx k2 pow

poszu:
## Motivation
This pull request updates the post protocol to use a new proof of work algorithm based on RandomX.

## Changes
- renamed K2Pow to Pow in Post Proof structure,
- bumped `post` repo
- added `[32]byte` difficulty for the new PoW algo, the existing `uint64` difficulty is left for backward compatibility on the testnet-5 (it's only used to verify the existing ATXs - new ATXs are created using RandomX PoW).
## Test Plan
- existing tests should pass
- test it on devnet-401 by updating all nodes

## DevOps Notes
  - added configuration option  `post-pow-difficulty [32]byte`. It's possible to configure it with a HEX string


poszu:
bors try

systest/parameters/bignet/smesher.json
```diff
@@ -37,6 +37,6 @@
         "post-k1": 273,
         "post-k2": 300,
         "post-k3": 100,
-        "post-k2pow-difficulty": 469112881707866048
+        "post-pow-difficulty": 469112881707866048
```
poszu: These need to be updated too.

dshulyak:
the comment generated by copilot looks like bs

dshulyak:
we need to consider how this will be delievered to testnet05.

there is an option to have it in backward compatible way, e.g from epoch X atxs must provide proofs with randomx. or do a checkpoint restart. i think doing it with checkpoint restart is lame, but 1st option will be more work

systest/parameters/bignet/smesher.json
pigmej: why do we need to change that param?

poszu: Because difficulty changed from an uint64 to an uint256 (represented as 64 bytes)

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

poszu:
bors try

activation/post_verifier.go
```diff
@@ -33,20 +33,29 @@ type postVerifierWorker struct {
 	channel  <-chan *verifyPostJob
 }
 
-type verifierFunc func(context.Context, *shared.Proof, *shared.ProofMetadata, ...verifying.OptionFunc) error
+type postVerifier struct {
+	inner  *verifying.ProofVerifier
```
fasmat: You can use struct embedding here to not require re-defining the `Close` method (`Verify` can still be extended).

activation/post_verifier_test.go
```diff
@@ -86,3 +88,25 @@ func TestPostVerifierVerifyAfterStop(t *testing.T) {
...
+		}, log.NewDefault(t.Name()))
+
+	var eg errgroup.Group
+	eg.Go(func() error {
+		v.Start(ctx)
+		return nil
+	})
+
+	cancel()
+	err := v.Verify(ctx, &shared.Proof{}, &shared.ProofMetadata{})
```
fasmat, line 108: Do you think it makes sense to also start `v.Verify` in it's own go-routine and then cancel, to ensure it is actually blocked before released again?

Otherwise it already receives a canceled context when is called.

activation/post.go
```diff
@@ -23,13 +25,34 @@ type PostSetupProvider initialization.Provider
...
-	K3              uint32 `mapstructure:"post-k3"`
-	K2PowDifficulty uint64 `mapstructure:"post-k2pow-difficulty"`
+	MinNumUnits     uint32   `mapstructure:"post-min-numunits"`
+	MaxNumUnits     uint32   `mapstructure:"post-max-numunits"`
+	LabelsPerUnit   uint64   `mapstructure:"post-labels-per-unit"`
+	K1              uint32   `mapstructure:"post-k1"`
+	K2              uint32   `mapstructure:"post-k2"`
+	K3              uint32   `mapstructure:"post-k3"`
+	K2PowDifficulty uint64   `mapstructure:"post-k2pow-difficulty"`
+	PowDifficulty   [32]byte `mapstructure:"post-pow-difficulty"`
```
fasmat: This is missing in `cmd/root.go` to be added as command line parameter like the other options here.

poszu:
bors try

activation/post.go
```diff
@@ -2,8 +2,10 @@ package activation
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	reflect "reflect"
```
fasmat: import doesn't need to be named

poszu:
bors try

activation/post_verifier_test.go
poszu: Putting `v.Verify()` in its own goroutine wouldn't ensure that it is called before the context is canceled.

activation/post.go
poszu: Fixed

activation/post_verifier_test.go
fasmat: You could use `context.WithTimeout` with a short timeout (100 milliseconds?).

That doesn't 100% guarantee that `v.Verify` is called with a context that is not (yet) canceled, but makes it very unlikely.

activation/post.go
```diff
@@ -23,13 +25,59 @@ type PostSetupProvider initialization.Provider
...
+	decodedLen := hex.DecodedLen(len(text))
+	if decodedLen != 32 {
+		return fmt.Errorf("expected 32 bytes, got %d", decodedLen)
+	}
+	_, err := hex.Decode(d[:], text)
+	return err
+}
+
+// Decodes a hex string representation of PoW difficulty into a [32]byte.
+func DecodePowDifficulty(f reflect.Type, t reflect.Type, data any) (any, error) {
```
fasmat: I don't think you need this hook if `PowDifficulty` already implements `TextUnmarshaler`. Instead you should be able to use `mapstructure.TextUnmarshallerHookFunc()`.

cmd/root.go
```diff
@@ -18,6 +19,20 @@ func ResetConfig() {
...
+
+// Set implements pflag.Value.Set.
+func (f powDifficultyFlag) Set(value string) error {
+	return f.UnmarshalText([]byte(value))
+}
+
+// Type implements pflag.Value.Type.
+func (powDifficultyFlag) Type() string {
+	return "PowDifficulty"
+}
```
fasmat: Imo these can be methods of `PowDifficulty` then the wrapper type isn't needed.

fasmat (approved):

activation/post_verifier_test.go
poszu: Relying on time is the source of flakiness, especially for slower runners. It's been a plague in poet UTs.

I'd rather refactor the code the be more testable, or change the test differently.

fasmat: Generally yes, here in this specific case not.

`context` is meant to be used this way. This is different from using sleeps and also doesn't cause races.

The test finishes equally if the context is canceled before `v.Verify` is called or after, but this way we make sure that it actually blocks before it's released again.

poszu:
bors merge

PR #4467, opened 2023-06-02
[Merged by Bors] - hare: report tortoise weakcoin asap

### LINKED ISSUE #4465: tortoise weakcoin should be recorded as soon as preround ended
## Description
currently coinflip is only recorded when hare terminates successfully.
but tortoise weakcoin doesn't need to depend on hare termination.

countvonzero:
## Motivation
Closes #4465

## Changes
use a separate report channel for weakcoin


hare/hare.go
```diff
@@ -47,10 +47,14 @@ type Consensus interface {
...
 	ID() types.LayerID
 	Set() *Set
-	Coinflip() bool
 	Completed() bool
 }
 
+type WeakCoinOutput interface {
+	ID() types.LayerID
+	Coinflip() bool
+}
```
dshulyak: can't it be just a struct? whats the point of interface?

dshulyak (approved):

hare/hare.go
countvonzero: yeah. was thinking of minimizing the change while keeping parallel to termination output (and reusing the same impl for test).
now changed both outputs to struct.

countvonzero:
bors merge

PR #4470, opened 2023-06-03
[Merged by Bors] - syncer sync genesis epoch during genesis 

### LINKED ISSUE #4469: syncer: fetch activations during genesis
if node is started in epoch 1 it will not ask peers for activations that target epoch 2.
as a consequence it will have low base ballot weight, if it publishes atx that targets epoch 3

countvonzero:
## Motivation

Closes #4469 

## Changes
- make syncer sync genesis epoch during genesis
- make hare use bootstrap active set when recover layer is the first of epoch. it would previously use last epoch's active set


dshulyak (approved):

countvonzero:
bors merge

PR #4473, opened 2023-06-04
[Merged by Bors] - bootstrapper: use unique filename for each update

countvonzero:
## Motivation
the update uses the same filename in gs bucket and therefore nodes will get different data around the update push time
as gs bucket is multi-region, and GCP cache is not controllable (at least not at our current gcp plan)

## Changes
- upon recover, remove old boostrap files 

- bootstrapper upload each kind of update with an unique suffix. for epoch 30
  - for a boostrap update, filename would be epoch-30-update-bs
  - for a beacon-only update => epoch-30-update-bc
  - for an activeset-only update => epoch-30-update-as

- once a node gets a bootstrap update, it will not query for beacon or activeset update as a bootstrap update supersede the others.

- node will always query updates for 3 epochs. if current epoch is 30, node will query all 3 url suffix for epochs (29, 30, 31). active set from the previous epoch to help with validating block certificate within hdist

- remove etag checking as update is now immutable, node will only download the same update once.


countvonzero:
bors try

countvonzero:
bors try

dshulyak (approved):
> node will generally query 2-3 epochs, depending on the confidence param set in hare active set. if current epoch is 30, node will query all 3 url suffix for

i think it makes sense to always lookup 3 epochs (past, present, future). present and future are clear, but we may also get active set for past so that certificates can be succesfully validated within hdist

countvonzero:
bors try

countvonzero:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
last try, if it fails please skip FailedNodes test

countvonzero:
bors merge

countvonzero:
bors merge

Issue #4456, opened 2023-05-31
Implement hare 3.0


piersy:
- [x] https://github.com/spacemeshos/go-spacemesh/issues/4494
- [x] https://github.com/spacemeshos/go-spacemesh/issues/4563
- [x] https://github.com/spacemeshos/go-spacemesh/issues/4648
- [x] https://github.com/spacemeshos/go-spacemesh/issues/4649
- [x] https://github.com/spacemeshos/go-spacemesh/issues/4650

dshulyak:
i think we should use hare 2 instead of hare 3 for the sake of the implementation versioning. hare 2 didn't really came out of the private research discussions

piersy:
Ok, so we'll call it hare 2.0 in the code, but we'll use what we are currently referring to as the hare 3.0 spec.

piersy:
So it was decided that it would be easier if we just refer to this as hare 3.0 and have a gap in the versioning, so from now on both the conceptual design and the implementation will be referred to as hare3 or hare 3.0

PR #4478, opened 2023-06-05
Hare 2 sketch

piersy:
## Motivation
Relates to #4456 

Note that fort he purposes of having more understandable versioning we are will assign the resulting implementation a version number of 2, even though the implementation is derived from the 3rd conceptual version of hare.

This is a high level sketch of how the hare3 protocol could fit together. There are many parts that are not implemented and there are likely many missing elements, but this is intended to give an idea of what the structure of the protocol as a whole will look like when completed.

Since the hare protocol is a fairly critical part of the system the structure has been designed in such a way as to minimise the complexity of the protocol by pushing all parts that are not part of the conceptual state machine defined by the hare 3.0 paper out of the protocol. E.G message encoding/decoding, signature verification and generally any code that could result in an error. Below is a diagram showing a very high level view of the design.

It probably makes more sense to look at `hare2/runner/runner.go` first since that shows how the protocol would be used at a high level and then look at `hare2/algorithm.go` which contains the detail.

hare_2_high_level_design

### DIFF
--- a/hare2/algorithm.go
+++ b/hare2/algorithm.go
@@ -23,19 +23,20 @@ package hare2
 // 	Gossip([]byte) (action Action, output []byte)
 // }
 
+//--------------------------------------------------
 // Take 2 below
 
-type NetworkGossiper interface {
-	NetworkGossip([]byte)
-}
+// type NetworkGossiper interface {
+// 	NetworkGossip([]byte)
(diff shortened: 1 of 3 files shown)


piersy:
Closing to change naming to hare3 see #4481

PR #4479, opened 2023-06-05
[Merged by Bors] - Install git LFS with make install

fasmat:
## Motivation
Add info about Git LFS to readme

## Changes
Just README.md

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,5 +199,5 @@ fuzz:
 .PHONY: fuzz
 
 lfs:
-	git lfs install
+	@git lfs install
 .PHONY: lfs
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ all: install build
 .PHONY: all
 
 install:
+	git lfs install
 	go mod download
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.52.0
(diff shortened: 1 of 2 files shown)


pigmej (commented):
We wanted to keep LFS as an option not as mandatory, can we make that also visible in the readme?

Lfs is needed only if one wants to run a full test suite, doesn't it?

poszu (approved):

fasmat:
@pigmej at the moment git LFS is a prerequisite for all tests, since `make test` requires it to be installed.

I can change that if that shouldn't be the case and exclude the tests that need LFS from `make test` and instead move them to their own `make` target?

piersy:
@pigmej @fasmat I'd prefer to keep make test running all the tests, it seems arbitrary to not run some tests because they use large files.

pigmej:
I second @piersy here. My point was only about the readme part. We don't need lfs to BUILD or run go-sm.

fasmat:
So how to proceed? Close this PR without merging?

fasmat:
I would like to move `make lfs` at least to `make install` so it is not called every time we run `make test` (it is only required once).

dshulyak (approved):

fasmat:
bors merge

PR #4480, opened 2023-06-05
[Merged by Bors] - Refactor Mesh Request Handler

### LINKED ISSUE #4389: fetcher: rewrite mesh hash request
the code in https://github.com/spacemeshos/go-spacemesh/pull/4381#issuecomment-1546538014 should be rewritten in a more robust way.

i found only of the problems, and it happens that there is another straightforward vulnerability.

ideally it should not use complex arithmetic for variable size allocation, maybe just let request a range from between two layers and limit the size of the request

fasmat:
## Motivation
Closes #4389 

`MeshHashReq` contains redundant information that can potentially be abused by an attacker to crash a node by forcing the node to allocate more memory than what is available on the system.

## Changes
- Removed `Steps` from request and renamed `Delta` to `By`
- Calculate `Steps` from the other 3 parameters
- Limit the number of hashes that can be requested in a single request to `MaxHashesPerReq` (defaults to 100)
- Do not echo back the layers that were requested
- Instead of using a string builder to query layers from DB use SQL statements

## Test Plan
- Added tests for changed functionality

## TODO

## DevOps Notes


fetch/wire_types.go
```diff
@@ -35,15 +35,14 @@ type ResponseBatch struct {
 }
 
 type MeshHashRequest struct {
-	From, To     types.LayerID
-	Delta, Steps uint32
+	From, To types.LayerID
+	By       uint32
```
poszu: For me, `Step` or `StepBy` would be easier to understand. What do you think?

fetch/mesh_data.go
```diff
@@ -272,7 +269,7 @@ func (f *Fetch) PeerMeshHashes(ctx context.Context, peer p2p.Peer, req *MeshHash
 	if err != nil {
 		f.logger.Fatal("failed to encode mesh hash request", log.Err(err))
 	}
-	lids, err := iterateLayers(req)
+	lids, err := iterateLayers(req, f.cfg.MaxHashesInReq)
```
poszu: Perhaps the `PeerMeshHashes` API could be changed so that `iterateLayers` wouldn't be needed in the first place? Currently, the response contains a slice of layer ids which seem redundant because the caller knows which layers he requested `Range(from, by, to)`. The caller can build this list on its own.

I mean:
```go
type MeshHashes struct {
// remove `Layers`:
//	Layers []types.LayerID `scale:"max=1000"` // depends on syncer Config `MaxHashesInReq`, defaults to 100, 1000 is a safe upper bound
	Hashes []types.Hash32  `scale:"max=1000"` // depends on syncer Config `MaxHashesInReq`, defaults to 100, 1000 is a safe upper bound
}
```

fasmat: That's a good point, at the moment an adversary could respond with a different set of Layers and their Hashes and the requesting node would not check if it received the Hashes for the Layers it requested.

I'll rewrite this part.

fetch/wire_types.go
piersy: I prefer Step

fetch/handler.go
```diff
@@ -20,15 +20,17 @@ import (
...
 	logger log.Log
 	cdb    *datastore.CachedDB
+	cfg    Config
 	bs     *datastore.BlobStore
 	msh    meshProvider
 	beacon system.BeaconGetter
 }
 
-func newHandler(cdb *datastore.CachedDB, bs *datastore.BlobStore, m meshProvider, b system.BeaconGetter, lg log.Log) *handler {
+func newHandler(cdb *datastore.CachedDB, cfg Config, bs *datastore.BlobStore, m meshProvider, b system.BeaconGetter, lg log.Log) *handler {
```
piersy, line 29: What about just passing MaxHashesInReq here, since it is the only config value used?

fetch/handler.go
```diff
@@ -209,28 +212,45 @@ func (h *handler) handleMeshHashReq(ctx context.Context, reqData []byte) ([]byte
...
 		h.logger.WithContext(ctx).With().Debug("failed to iterate layers", log.Err(err))
 		return nil, err
 	}
-	if len(lids) > 0 {
-		hashes, err = layers.GetAggHashes(h.cdb, lids)
-		if err != nil {
-			h.logger.WithContext(ctx).With().Warning("failed to get mesh hashes", log.Err(err))
-			return nil, err
-		}
+	hashes, err = layers.GetAggHashes(h.cdb, req.From, req.To, req.By)
```
piersy: What is the meaning of an aggregated hash?

fasmat: Its the hash of a layer. I just rewrote this function not to query multiple layer hashes by passing in a slice of `[]types.LayerID` and using a string builder. This was not only very inefficient, an adversary could crash a node by causing it to build an SQL query thats large enough such that the node runs out of memory.

See also https://github.com/spacemeshos/go-spacemesh/issues/4427 and https://github.com/spacemeshos/go-spacemesh/pull/4431.

The previous fix was just symptomatic by limiting how many layers can be requested, but the inefficient SQL hasn't been updated in the previous PR.

fetch/mesh_data.go
fasmat: Removed echoing the layers that were requested. This also removed the need for `iterateLayers` and allowed me to update `GetAggHashes` to use more efficient SQL.

fetch/handler.go
fasmat: `newHandler` is part of the fetcher, so I see no issue passing in the fetcher config.

Imo it shouldn't only consider `MaxHashesInReq` but also other configuration like `RequestTimeout` and `MaxRetriesForRequest`, but this is out of the scope for this PR.

fetch/wire_types.go
```diff
@@ -35,21 +35,19 @@ type ResponseBatch struct {
 }
 
 type MeshHashRequest struct {
```
piersy, line 49: It took me a long time to understand what By really meant, I think a comment here would be useful to any future reader.

```suggestion
// The MeshHashRequest is used by the ForkFinder to request sets of hashes from
// a peer that are used to determine the point at which a divergence occured in
// this node and the peer's history. From and To define the beginning and end
// hash of the range of hashes to request and in the case that the length of
// that range exceeds Config.MaxHashesInReq, instead of requesting every hash
// we sample Config.MaxHashesInReq hashes across the range and By is set to the
// Step used in sampling. Eg if By is 2 we will request every other hash.
type MeshHashRequest struct {
```

fetch/handler.go
piersy: I'd say no point passing the whole object if only one field is used, but in this case if it should be using more fields from the config I think it's fine.

fetch/wire_types.go
piersy: However I don't think By should be part of the request. By/Step should be returned by the handler, since the requestor cannot know what the node it is requesting from will have set as it's MaxHashesInReq. The caller should just request a range of hashes and the responder should return a sampling of that range ensuring that the number of hashes don't exceed MaxHashesInReq, so the responder will define By/Step at the point they handle the request.

fasmat: I disagree 🙁 

If the requestor requests a range of layers and the responder can decide how many hashes they would like to return form that range they can just decide to always just return 1, completely defeating the purpose of a batch request.

`MaxHashesInReq` can be different for requester & responder, that is correct. In that case the requester has to adapt their range / step size to fit accordingly. This case is not yet handled and is out of scope for this PR.

piersy: I disagree with your disagreement :D

In this scenario where the responder responds with 1 hash, that means their MaxHashesInReq is 1.

With the current implementation the requesting node will just never get any hashes, their request will always be rejected. If you add the ability to adapt the requester will keep on requesting probably halving their request size with each request until their request is serviced, but there's no need for all that wasted rejected network traffic, if the responder simply returns a response that has been fitted to their MaxHashesInReq.

fetch/handler.go
piersy: But why do we call them *aggregated*hash ?

fasmat: Don't know but changing it is not part of this PR 🙂

fetch/wire_types.go
fasmat: That would mean reverting most of the changes in this PR. The goal was to remove the redundant `Steps` parameter from the request both to avoid buffer overflows and node crashes due to oom issues.

The goal imo is not to rewrite the fork finding logic.

piersy: Fair enough, I was not aware that this would require such an in depth re-write. I'll add a ticket for this.

piersy: https://github.com/spacemeshos/go-spacemesh/issues/4486

fasmat:
bors try

fasmat:
bors try

fetch/fetch.go
```diff
@@ -81,6 +81,7 @@ type Config struct {
 	MaxRetriesForPeer    int
 	BatchSize, QueueSize int
 	RequestTimeout       time.Duration // in seconds
+	MaxHashesInReq       int
```
countvonzero: this config is a bit confusing. 
syncer is the initiator here. fetch is just a general lib to do the leg work.
maybe be just a sync config, and pass as a parameter to fetch. or simply make it a hard-coded constant in fetch. i prefer the latter as its simpler.

countvonzero (approved):
do i understand correctly that
- the dos vector is addressed by limiting number of hashes requested
- the rest of the code change is simplification
?

countvonzero (approved):
do i understand correctly that
- the dos vector is addressed by limiting number of hashes requested
- the rest of the code change is simplification
?

fetch/handler.go
```diff
@@ -209,28 +212,45 @@ func (h *handler) handleMeshHashReq(ctx context.Context, reqData []byte) ([]byte
 		h.logger.WithContext(ctx).With().Warning("failed to parse mesh hash request", log.Err(err))
 		return nil, errBadRequest
 	}
-	lids, err := iterateLayers(&req)
-	if err != nil {
+	if err := h.validateReq(req); err != nil {
 		h.logger.WithContext(ctx).With().Debug("failed to iterate layers", log.Err(err))
```
poszu: The error message needs updating :)

fetch/handler.go
```diff
@@ -209,28 +212,45 @@ func (h *handler) handleMeshHashReq(ctx context.Context, reqData []byte) ([]byte
...
+		log.Stringer("layer_to", req.To),
+		log.Uint32("by", req.By),
+		log.Int("count_hashes", len(hashes)),
+	)
 	return data, nil
 }
+
+func (h *handler) validateReq(req MeshHashRequest) error {
+	if req.By == 0 {
+		return fmt.Errorf("%w: %v", errBadRequest, req)
```
poszu: How about simplifying the life of our future selves and tell what was actually wrong in the request:
```suggestion
		return fmt.Errorf("%w: By must not be 0 (%v)", errBadRequest, req)
```

fetch/handler.go
```diff
@@ -209,28 +212,45 @@ func (h *handler) handleMeshHashReq(ctx context.Context, reqData []byte) ([]byte
...
+	if req.To.Before(req.From) {
+		return fmt.Errorf("%w: %v", errBadRequest, req)
+	}
+
+	diff := req.To.Difference(req.From)
+	count := diff/req.By + 1
+	if diff%req.By != 0 {
+		// last layer is not a multiple of By, so we need to add it
+		count++
+	}
```
poszu: The same code appears in `GetAggHashes()` and in find_fork.go. How about making it a method of `MeshHashRequest` and unit-testing it?

fetch/wire_types.go
poszu: @fasmat you didn't respond to my suggestion to change `By` to `Step` or `StepBy`. What do you think?

fetch/fetch.go
piersy: I think a constant is preferable, allowing nodes to set different values will break the network as described [here](https://github.com/spacemeshos/go-spacemesh/issues/4486)

piersy (changes requested):
I think the max hashes should be hardcoded, then we don't need to deal with any issues arising from nodes setting different values.

fetch/wire_types.go
fasmat: @poszu  `Step` did have a different meaning before (number of layers expected instead of increment size) and I want to avoid confusion. I think `By` is quite clear in its meaning:

`From` Layer X `To` Layer Y `By` Z Layers.

fetch/fetch.go
fasmat: One is a parameter on the client/requesting side and one is a parameter on the serving side.

`syncer.Config.MaxHashesInReq` defines how many layers are asked for in one to another peer, while `fetch.Config.MaxHashesInReq` defines how many layers are at most served in a single request. Both imo should be configurable. Slower nodes/nodes with less bandwidth might want to limit the number of layers they request or serve in a single query.

For HTTP requests that is usually done using pagination, where the client can decide how many items they want to receive in a single request and the server can limit the maximum number of items they want to serve in a single response (usually returns error 422 if limit is larger than it wants to serve).

fasmat:
> do i understand correctly that
> 
> * the dos vector is addressed by limiting number of hashes requested
> * the rest of the code change is simplification
>   ?

Yes, `MaxHashesPerReq` prevents a node from crashing due to a malicious request to serve too many layers at once. Additionally the removal of `Steps` removes possible ambiguity in a request since that parameter can be derived from the others.

fetch/handler.go
fasmat: I added a `Count` and `Validate` Method to `MeshHashReq` to reduce code duplication.

I don't think the code in `GetAggHashes()` should be de-duplicated. We cannot pass `MeshHashReq` as parameter to it and imo shouldn't - `MeshHashReq` is API layer not DB layer. Moving the "logic" to the `sql` package or `types` package imo also isn't a solution because it doesn't belong there.

Both places are already unit tested -> `GetAggHashes()` has tests that validates that the correct number of layers is returned and `MeshHashRequest.Count()` is tested in `TestForkFinder_FindFork_Edges` and `TestForkFinder_FindFork_Permutation`

fetch/wire_types.go
fasmat: Updated description

fasmat:
@piersy @countvonzero 

Going through the review makes me realise that maybe the current implementation can be improved further. The fork finder sets up a boundary based on the latest layer two nodes agree on (`From`) and the oldest two nodes disagree on (`To`). This boundary is then used to request all hashes between those two layers (limited by `MaxHashesInReq` - currently hard coded to 100 on both the client and server side).

This means that if a fork is detected a request is made that can request up to 100 hashes, but if `From` and `To` differ by more than 100 at least another request (for at least 3 hashes) has to be made to find the layer where the divergence happened.

What if we change that logic to instead do a binary search and update the boundaries based on if the hash in the middle between `From` and `To` matches or not? This would simplify the code further by using a recursive search: If the hash matches update `From`, if it differs update `To` -> call recursively until `From` and `To` differ by no more than 1 layer -> `To` is the layer where the fork occurred.

So instead of 1 request with 100 hashes, we would do 5-6 requests for one hash each (assuming `From` and `To` differ by 100).

This would also remove the need for `GetAggHashes`, the need for `MaxHashesInReq`, some of the validation currently needed to check for overflows, etc.

fasmat:
bors try

fetch/wire_types.go
```diff
@@ -34,11 +38,51 @@ type ResponseBatch struct {
...
+		return fmt.Errorf("%w: To before From", errBadRequest)
+	}
+
+	diff := r.To.Difference(r.From)
+	count := diff/r.By + 1
+	if diff%r.By != 0 {
+		// last layer is not a multiple of By, so we need to add it
+		count++
+	}
+	if count > MaxHashesInReq {
```
poszu: Why not:
```suggestion
	if r.Count() > MaxHashesInReq {
```

fetch/wire_types.go
poszu: For me `Step` is clearer. I think it's the same for @piersy too. But I'm not blocking the PR on it.

poszu (approved):

fetch/wire_types.go
```diff
@@ -34,22 +38,60 @@ type ResponseBatch struct {
 	Responses []ResponseMessage `scale:"max=1000"` // depends on fetch config `BatchSize` which defaults to 20, more than 1000 seems unlikely
 }
 
+// MeshHashRequest is used ForkFinder to request the hashes of layers from
```
piersy:
```suggestion
// MeshHashRequest is used by ForkFinder to request the hashes of layers from
```

piersy:
> @piersy @countvonzero
> 
> Going through the review makes me realise that maybe the current implementation can be improved further. The fork finder sets up a boundary based on the latest layer two nodes agree on (`From`) and the oldest two nodes disagree on (`To`). This boundary is then used to request all hashes between those two layers (limited by `MaxHashesInReq` - currently hard coded to 100 on both the client and server side).
> 
> This means that if a fork is detected a request is made that can request up to 100 hashes, but if `From` and `To` differ by more than 100 at least another request (for at least 3 hashes) has to be made to find the layer where the divergence happened.
> 
> What if we change that logic to instead do a binary search and update the boundaries based on if the hash in the middle between `From` and `To` matches or not? This would simplify the code further by using a recursive search: If the hash matches update `From`, if it differs update `To` -> call recursively until `From` and `To` differ by no more than 1 layer -> `To` is the layer where the fork occurred.
> 
> So instead of 1 request with 100 hashes, we would do 5-6 requests for one hash each (assuming `From` and `To` differ by 100).
> 
> This would also remove the need for `GetAggHashes`, the need for `MaxHashesInReq`, some of the validation currently needed to check for overflows, etc.

Sounds good to me, I'm just wondering why that approach wasn't taken initially, is it because the overhead of each request would be high?

A side note, I'm still seeing `syncer.Config.MaxHashesInReq` existing in this PR

fasmat:
@piersy: I updated the ForkFinder also to use a const for `MaxHashesInReq` instead of having it a configurable parameter

fetch/wire_types.go
piersy: As a native English speaker By doesn't really make sense to me, but I'm not going to block the PR on it, and it could all be moot if we go for the binary search approach

fasmat:
New Issue for refactoring (after genesis): https://github.com/spacemeshos/go-spacemesh/issues/4517

fasmat:
bors merge

fetch/wire_types.go
```diff
@@ -34,22 +38,54 @@ type ResponseBatch struct {
...
 
+// MeshHashRequest is used by ForkFinder to request the hashes of layers from
+// a peer to find the layer at which a divergence occurred in the local mesh of
+// the node.
+//
+// From and To define the beginning and end layer to request. By defines the
+// increment between layers to limit the number of hashes in the response, e.g.
+// 2 means only every other hash is requested. On the requester side the number
+// of hashes in the response is limited by `MaxHashesInReq` in `syncer.Config`,
+// on the responder side it is limited by `fetch.MaxHashesInReq`.
```
piersy:
```suggestion
// 2 means only every other hash is requested. The number
// of hashes in the response is limited by `fetch.MaxHashesInReq`.
```

syncer/find_fork.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/spacemeshos/go-spacemesh/sql/layers"
 )
 
+const MaxHashesInReq = 100
```
piersy: Lets remove this and use `fetch.MaxHashesInReq` directly

piersy (approved):
Looking good!

fasmat:
bors merge

PR #4481, opened 2023-06-06
hare 3.0 sketch

piersy:
## Motivation
Relates to #4456 

This PR is intended to allow for an early look at the design of hare3 so that any criticisms/suggestions/issues can be raised early and incorporated into the final implementation without the need for significant re-work.

This is a high level sketch of how the hare3 protocol could fit together. There are many parts that are not implemented and there are likely many missing elements, but this is intended to give an idea of what the structure of the protocol as a whole will look like when completed.

Since the hare protocol is a fairly critical part of the system the structure has been designed in such a way as to minimise the complexity of the protocol by pushing all parts that are not part of the conceptual state machine defined by the hare 3.0 paper out of the protocol. E.G message encoding/decoding, signature verification and generally any code that could result in an error. Below is a diagram showing a very high level view of the design.

It probably makes more sense to look at `hare3/runner/runner.go` first since that shows how the protocol would be used at a high level and then look at `hare3/algorithm.go` which contains the detail.
hare3-diagram

### DIFF
--- a/hare2/algorithm.go
+++ b/hare2/algorithm.go
@@ -23,19 +23,20 @@ package hare2
 // 	Gossip([]byte) (action Action, output []byte)
 // }
 
+//--------------------------------------------------
 // Take 2 below
 
-type NetworkGossiper interface {
-	NetworkGossip([]byte)
-}
+// type NetworkGossiper interface {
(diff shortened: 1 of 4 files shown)


piersy:
Superceded by #4564

PR #4483, opened 2023-06-06
[Merged by Bors] - Revert VRF signature domain

### LINKED ISSUE #4274: beacon: messages should not be decodable as another message type within same domain

fasmat:
## Motivation
Closes #4274 

## Changes
- Reverts the addition of a domain byte to VRFSignatures.
- Ensure that VRFSignatures in `beacon` contain a unique enum value so they can uniquely be decoded and verified

## Test Plan
- Revert tests that were changed before.

## TODO

## DevOps Notes

### DIFF
--- a/signing/signer.go
+++ b/signing/signer.go
@@ -19,7 +19,6 @@ const (
 	ATX Domain = iota
 	BEACON_FIRST_MSG
 	BEACON_FOLLOWUP_MSG
-	BEACON_PROPOSAL
 	BALLOT
 	HARE
 	POET
@@ -34,8 +33,6 @@ func (d Domain) String() string {
 		return "BEACON_FIRST_MSG"
 	case BEACON_FOLLOWUP_MSG:
 		return "BEACON_FOLLOWUP_MSG"
-	case BEACON_PROPOSAL:
-		return "BEACON_PROPOSAL"
 	case BALLOT:
 		return "BALLOT"
 	case HARE:
--- a/signing/signer.go
+++ b/signing/signer.go
@@ -16,12 +16,14 @@ import (
 type Domain byte
 
 const (
-	ATX Domain = iota
-	BEACON_FIRST_MSG
-	BEACON_FOLLOWUP_MSG
-	BALLOT
-	HARE
-	POET
+	ATX Domain = 0
+
+	BALLOT = 2
+	HARE   = 3
+	POET   = 4
+
+	BEACON_FIRST_MSG    = 10
+	BEACON_FOLLOWUP_MSG = 11
 )
 
 // String returns the string representation of a domain.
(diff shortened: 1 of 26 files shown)


piersy:
Hey @fasmat why are you reverting this change?

signing/signer.go
```diff
@@ -19,7 +19,6 @@ const (
 	ATX Domain = iota
 	BEACON_FIRST_MSG
 	BEACON_FOLLOWUP_MSG
-	BEACON_PROPOSAL
 	BALLOT
 	HARE
 	POET
```
dshulyak: is it possible to make everything except beacon compatible with testnet05? otherwise we can't deploy this change

dshulyak: maybe iota should be dropped and we should just manually assign indexes, iota is too easy to break

fasmat: Before https://github.com/spacemeshos/go-spacemesh/pull/4460 the concrete values were:

```golang
const (
	ATX domain = 0
	BEACON = 1
	BALLOT = 2
	HARE = 3
	POET = 4
)
```

I can change it to

```golang
const (
	ATX domain = 0

	BALLOT = 2
	HARE = 3
	POET = 4

	BEACON_FIRST_MSG = 10
	BEACON_FOLLOWUP_MSG = 11
)
```

would this work or does it need to be compatible with the changes introduced in the previous PR?

fasmat:
> Hey @fasmat why are you reverting this change?

The previous change was bigger than necessary, the VRF Signer should not have been extended with domain indicators. See the discussion starting here for more details: https://spacemesh.slack.com/archives/C89QJJY3G/p1685988586158159

signing/signer.go
dshulyak: > would this work or does it need to be compatible with the changes introduced in the previous PR?

i think it should be enough. we didn/t deploy previous change

dshulyak (approved):

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

dshulyak:
bors merge

PR #4451, opened 2023-05-30
[Merged by Bors] - Use go version from go.mod

fasmat:
## Motivation
Instead of explicitly specifying the go-version for building go-spacemesh use the version specified in `go.mod`

## Changes
`actions/setup-go` can parse the version specified in `go.mod` to install the same go version when building / testing the source code. `check-latest` ensures that the newest patch version is installed instead of a cached version, so we always have the latest fixes of the go runtime & standard library in our builds.

## Test Plan
n/a

## TODO

- [x] Explain motivation or link existing issue(s)
- [x] Test changes and document test plan
- [x] Update documentation as needed

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #4484, opened 2023-06-06
[Merged by Bors] - Fix release workflow: move git checkout before reading go.mod

brusherru:
## Motivation
No issue. The bug was introduced in #4451 

## Changes
Just moving `actions/checkout` step before installing go ;)

## Proofs that it works
- before the fix: https://github.com/spacemeshos/go-spacemesh/actions/runs/5184781196
- after the fix: https://github.com/spacemeshos/go-spacemesh/actions/runs/5187132323


fasmat (approved):
lgtm 👍

fasmat:
bors merge

PR #4485, opened 2023-06-06
[Merged by Bors] - p2p: set prologue to a unique genesis id and latest set genesis

dshulyak:
in noise protocol there is an option to set [prologue](https://noiseprotocol.org/noise.html#prologue).

> Noise protocols have a prologue input which allows arbitrary data to be hashed into the h variable. If both parties do not provide identical prologue data, the handshake will fail due to a decryption error. This is useful when the parties engaged in negotiation prior to the handshake and want to ensure they share identical views of that negotiation.

this is not exactly our use case, but we don't want to establish secure connection if peers don't need to speak to each other.

### DIFF
--- a/fetch/fetch_test.go
+++ b/fetch/fetch_test.go
@@ -327,13 +327,13 @@ func TestFetch_PeerDroppedWhenMessageResultsInValidationReject(t *testing.T) {
 
 	// Good host
 	genesisID := types.Hash20{}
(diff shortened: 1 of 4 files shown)


dshulyak:
bors try

countvonzero (approved):
coooool. thanks!

dshulyak:
bors merge

PR #4492, opened 2023-06-07
[Merged by Bors] - p2p: drop handshake protocol

dshulyak:
additional handshake protocol is redundant, since we enforce same genesis during noise handshake

we can re-enable it, if in future we will have additional use cases


countvonzero (approved):

dshulyak:
bors merge

PR #4493, opened 2023-06-07
[Merged by Bors] - Persist POST

### LINKED ISSUE #4491: Persist the last POST
## Description
Currently neither the initial nor the subsequent proofs are persisted. Thus it might happen that the node is shut down between the proof is created and it is used. That proof would be lost and would have to be recreated, which is hard.

## Affected code
- especially the initial POST,
- every post built by `nipostBuilder`,

## Proposed solution
The last proof should be persisted (in a DB or in a file, like the nipost builder state).

poszu:
## Motivation
Closes #4491 

## Changes
- persist the initial POST in a file;
- don't erase the nipost builder's state after it completes to avoid running it from scratch if the app was closed between creating NiPOST and publishing an ATX.

## Test Plan
Existing tests should pass

## DevOps Notes


activation/activation.go
```diff
@@ -261,19 +261,28 @@ func (b *Builder) run(ctx context.Context) {
...
+		return nil
+	}
+	// ...and if we don't have an initial POST persisted already.
+	if _, err := loadPost(b.nipostBuilder.DataDir()); err == nil {
+		return nil
 	}
 
+	// Create the initial post and save it.
+	startTime := time.Now()
+	var err error
```
dshulyak, line 276: it makes sense to group variables in var ( ... ) block

activation/activation.go
```diff
@@ -261,19 +261,28 @@ func (b *Builder) run(ctx context.Context) {
...
+	b.initialPost, b.initialPostMeta, err = b.postSetupProvider.GenerateProof(ctx, shared.ZeroChallenge)
+	if err != nil {
+		return fmt.Errorf("post execution: %w", err)
+	}
+	metrics.PostDuration.Set(float64(time.Since(startTime).Nanoseconds()))
+
+	if err := savePost(b.nipostBuilder.DataDir(), b.initialPost); err != nil {
+		b.log.With().Warning("failed to save initial post: %w", log.Err(err))
```
dshulyak, line 284: i would fail and terminate app here for simplicity

dshulyak (approved):

activation/activation.go
poszu: What's the benefit? I prefer to define variables next to the place they are used.

activation/activation.go
poszu: I was thinking about whether I should return an error, panic, or log and continue. I picked the latter because it's not really a fatal error and the application _can_ continue. Terminating would mean that the post needs to be regenerated and there's really no guarantee that `savePost()` would succeed on the second try.

activation/activation.go
dshulyak: i meant that you defined both startTime and err close to each other, but one with var and another without

activation/activation.go
dshulyak: but that failure means something is not ok with os or disk, e.g can't fit the data for example. it is not like it can crash due to some random flake and nobody cares

dshulyak (changes requested):
i think we should not write such copilot comments, they are just wasting reviewer time. beside that all this useless links and robot faces will be added to github history, making it futher unreadable

activation/activation.go
poszu: `startTime` is initialized while `err` is not. Using `var (...)` adds 2 extra lines w/o any real benefit.
```
startTime := time.Now()
var err error
```
vs.
```go
var (
	startTime = time.Now()
	err       error
)
```

dshulyak (approved):

activation/activation.go
fasmat: does `err` need to be declared explicitly? The next line could just be:

```go
b.initialPost, b.initialPostMeta, err := b.postSetupProvider.GenerateProof(ctx, shared.ZeroChallenge)
```

poszu:
bors merge

activation/activation.go
poszu: @fasmat it cannot. The compiler complains with:
> non-name b.initialPost on left side of :=

activation/nipost_state.go
```diff
@@ -57,58 +59,81 @@ func read(path string) ([]byte, error) {
 	return data[:end], nil
 }
 
-func saveNipostChallenge(dir string, ch *types.NIPostChallenge) error {
-	if _, err := os.Stat(dir); err != nil {
-		return fmt.Errorf("post dir not created: %w", err)
-	}
-	data, err := codec.Encode(ch)
+func load(filename string, dst scale.Decodable) error {
```
fasmat, line 62: `load` seems to be inefficient:

1. Read all data into memory
2. Remove last 8 bytes
3. Calculate CRC of data in memory (i.e. has to be copied again)
4. Parse last 8 bytes as Uint64
5. Return all data except last 8 bytes to be decoded
6. Decode data as scalable (i.e. copy data again)

This can be done more efficiently:

1. Move CRC to beginning of file and read first as uint64
2. Use an `io.TeeReader` to calculate CRC while decoding a the scale encoded data without the need of loading data into memory and copying twice.

fasmat: The same can be done in reverse: `write` already does it more efficiently; it uses `io.MultiWriter` to write data to a buffer and CRC at the same time.

I would only change it to not use a buffer, but instead write 8 (zero) bytes to a file then use the `io.MultiWriter` to calculate and write to the file at the same time and then seek to the beginning of the file and update the CRC.

The new file can then atomically replace the old one with [atomic.ReplaceFile](https://github.com/natefinch/atomic/blob/v1.0.1/file_unix.go#L12)

poszu: Good idea :+1:, though it's used infrequently and the data is small so it shouldn't matter.

I just refactored the existing code to reuse it and fixing inefficiencies is out of scope. Could you open an issue for it?

activation/activation.go
fasmat: :+1: `:=` doesn't work with fields.

activation/nipost_state.go
fasmat: Created issue here: https://github.com/spacemeshos/go-spacemesh/issues/4506

poszu:
bors try

dshulyak:
that failure is caused by one of the issues described in https://github.com/spacemeshos/go-spacemesh/issues/4497

poszu:
bors try-
bors merge

PR #4500, opened 2023-06-09
[Merged by Bors] - Fix hare threshold calculations

### LINKED ISSUE #4499: Incorrect threshold calculation for Hare BFT consensus protocol
## Description
As part of our consensus logic there is a rule for calculating whether values should be considered. The original rule was that we need to see` >= f+1` support for a value, the intent of this rule is to ensure that at least one honest party supported a value. It was modified (see [here](https://community.spacemesh.io/t/hare-equivocations-defense/329/11) for discussion) to be ` >= f+1-k` where k is a count of known non-honest nodes. This modification ensures that we still only need support from a single honest node, even if we detect up to f dishonest nodes.

This rule was not being enforced correctly since we introduced a [CountInfo](https://github.com/spacemeshos/go-spacemesh/blob/2a7f0e0bbe517b7c1401e39892452592b12ce7a3/hare/algorithm.go#L142-L145) object which provided 2 different mechanisms to mark dishonest nodes, but when performing the threshold calculation it only considered dishonest nodes marked through one of the mechanisms, `dhCount` is not included in the `Meet` calculation.

```go
type CountInfo struct {
	// number of nodes that are honest, dishonest and known equivocators.
	// known equivocators are those the node only has eligibility proof but no hare message.
	numHonest, numDishonest, numKE int
	hCount, dhCount, keCount       int
}

func (ci *CountInfo) Meet(threshold int) bool {
	// at least one honest party with good message
	return ci.hCount > 0 && ci.hCount+ci.keCount >= threshold
}
```

This logic is used in commit, pre-round, status and notify trackers, so they are likely all not functioning correctly.

piersy:
## Motivation
Closes #4499 

## Changes
Modifies the Meet method calculation for CountInfo

## Test Plan
Added tests for all affected trackers, and for the CountInfo object

## DevOps Notes


dshulyak (dismissed):

hare/algorithm_test.go
```diff
@@ -174,6 +175,44 @@ func (mev *mockEligibilityValidator) ValidateEligibilityGossip(context.Context,
 	return atomic.LoadInt32(&mev.valid) != 0
 }
 
+// We assert the the theshold X is met a long as there is one honest vote and
+// the sum of honest, dishonest and known equivocator is >= X.
```
countvonzero: this description is not exactly right. the condition also requires at least 1 honest vote.

hare/algorithm_test.go
```diff
@@ -174,6 +175,44 @@ func (mev *mockEligibilityValidator) ValidateEligibilityGossip(context.Context,
...
+func TestCountInfoMeetsTrheshold(t *testing.T) {
```
countvonzero, line 183: i think the test below is sufficient to deliver your point.
```suggestion
func TestCountInfoMeetsTrheshold(t *testing.T) {
        c := &CountInfo{}
	c.IncKnownEquivocator(1)
	require.False(t, c.Meet(3))
	c.IncDishonest(1)
	require.False(t, c.Meet(3))
	c.IncHonest(1)
	require.True(t, c.Meet(3))
}
```

countvonzero (approved):

hare/algorithm_test.go
piersy: That's in there right?
>We assert the the theshold X is met **a long as there is one honest vote** and

Although I have noticed a typo **a -> as**

piersy: Oh do you mean it should be *at least one* ?

hare/algorithm_test.go
piersy: This test doesn't assert that at least one honest vote is required to meet the threshold.

I.E the following implementation of Meet would pass
```
ci.hCount+ci.dhCount+ci.keCount >= threshold
```

piersy:
bors cancel

piersy:
bors merge

dshulyak:
bors try

piersy:
bors merge

piersy:
bors merge

PR #4514, opened 2023-06-14
[Merged by Bors] - tortoise: switch back into verifying only if same layers can be verified

### LINKED ISSUE #4503: tortoise: switch into verifying mode only when layers up to the last one can be verified
in the implementation we will switch from full into verifying if any layer crossed pessimistic threshold. it doesn't cause any major issues, but creates a spam in logs and forces consumers to do more work, as tortoise will report same consensus results multiple times.

to avoid this, tortoise should be more pessimistic about when it switches back into verifying mode. check that all layers confirmed in full mode were confirmed in verifying too and only then switch

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4503
depends on: https://github.com/spacemeshos/go-spacemesh/pull/4438

in the implementation we will switch from full into verifying if any layer crossed pessimistic threshold. it doesn't cause any major issues, but creates a spam in logs and forces tortoise users to do more work, as tortoise will report same consensus results multiple times.

to avoid this, tortoise should be more pessimistic about when it switches back into verifying mode. check that all layers confirmed in full mode were confirmed in verifying too and only then switch


dshulyak:
bors try

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #4516, opened 2023-06-14
[Merged by Bors] - tortoise, rewards: enforce minimal active set weight

### LINKED ISSUE #4466: tortoise, rewards: enforce minimal active set weight in eligibilities computation
there is a quite bad dos opportunity because every ballot declares their own tortoise active set.
in the worst case smesher can declare their own atx as the only one in active set, and get eligibility to create ~50~ 1 ballot per layer.
there is nothing to gain, but it will add a lot of data to store, and may even cause oom in tortoise.

the proposed solution is to define configurable minimal active set weight. this weight will be enforced in ballot weight computation.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4466
depends on: https://github.com/spacemeshos/go-spacemesh/pull/4438

without minimal active set weight any smesher can be eligible to produce 1 ballot per layer.
this will result in spam and non-optimal resources consumption.

this change introduce minimal active set weight for eligible number of slots computation. if weight of the
atx specified in the first ballot active set is lower than minimal active set weight then protocol will use minimal.
this value must be the same across whole network


proposals/util/util.go
```diff
@@ -29,12 +29,19 @@ func CalcEligibleLayer(epochNumber types.EpochID, layersPerEpoch uint32, vrfSig
...
+}
+
 // GetNumEligibleSlots calculates the number of eligible slots for a smesher in an epoch.
-func GetNumEligibleSlots(weight, totalWeight uint64, committeeSize uint32, layersPerEpoch uint32) (uint32, error) {
+func GetNumEligibleSlots(weight, minWeight, totalWeight uint64, committeeSize uint32, layersPerEpoch uint32) (uint32, error) {
 	if totalWeight == 0 {
 		return 0, ErrZeroTotalWeight
 	}
-	numberOfEligibleBlocks := weight * uint64(committeeSize) * uint64(layersPerEpoch) / totalWeight // TODO: ensure no overflow
+	numberOfEligibleBlocks := weight * uint64(committeeSize) * uint64(layersPerEpoch) / maxWeight(minWeight, totalWeight) // TODO: ensure no overflow
```
countvonzero: prolly can rename this variable to just `numEligible`?

countvonzero (approved):

dshulyak:
bors merge

PR #4522, opened 2023-06-15
[Merged by Bors] - hare: re-gossip eligibility proof for previously known malicious identity

### LINKED ISSUE #4521: hare does not gossip eligibility upon receiving a msg from a previous malicious identity
## Description

@piersy reported 
```
So we have a bug in the broker whereby we don't gossip malfeasance proofs for previously known malicious identities
Here is where the broker checks in the mesh for a proof it will send that proof to Hare on the malfeasance channel:
https://github.com/spacemeshos/go-spacemesh/blob/develop/hare/broker.go#L189-L200
And this is where hare reads it from the malfeasance channel, and checks to see if this is a known malicious identity, and if so does not relay the proof to the network:
https://github.com/spacemeshos/go-spacemesh/blob/1a0f48a164c803ebce89d90bf008295cdc2c3ab2/hare/hare.go#L605C1-L616
[broker.go](https://github.com/spacemeshos/go-spacemesh/blob/develop/hare/broker.go)
<https://github.com/[spacemeshos/go-spacemesh](https://github.com/spacemeshos/go-spacemesh)|spacemeshos/go-spacemesh>spacemeshos/go-spacemesh | Added by [GitHub](https://spacemesh.slack.com/services/B01UK3XL8K0)
```
```
Another bug that would manifest if the above bug was fixed is that in the broker no record is kept of if a malfeasance proof has been sent for an identity, so a malicious actor can send as many equivocating messages as they like and honest nodes would gossip a malfeasance for each equivocating message.
```

countvonzero:
## Motivation
Closes #4521

## Changes
- immediately re-gossip the eligibility when receiving a hare msg from a previously known malicious identity
- share eligibility tracker btwn broker and consensus processes so we can check whether we have previously gossiped eligibility proof for the same identity/layer/round
- make sure broker's old data is cleaned up on every layer tick

### DIFF
--- a/hare/broker.go
+++ b/hare/broker.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/p2p"
@@ -33,7 +34,7 @@ type Broker struct {
 	roleValidator validator                // provides eligibility validation
 	stateQuerier  stateQuerier             // provides activeness check
 	nodeSyncState system.SyncStateProvider // provider function to check if the node is currently synced
-	mchOut        chan<- *types.MalfeasanceGossip
+	publisher     pubsub.Publisher
 	outbox        map[types.LayerID]chan any
 	pending       map[types.LayerID][]any // the buffer of pending early messages for the next layer
 	latestLayer   types.LayerID           // the latest layer to attempt register (successfully or unsuccessfully)
@@ -51,7 +52,7 @@ func newBroker(
 	roleValidator validator,
 	stateQuerier stateQuerier,
 	syncState system.SyncStateProvider,
-	mch chan<- *types.MalfeasanceGossip,
+	publisher pubsub.Publisher,
 	limit int,
 	log log.Log,
 ) *Broker {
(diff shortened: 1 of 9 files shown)


countvonzero:
bors try

hare/broker.go
```diff
@@ -348,64 +353,71 @@ func (b *Broker) handleEarlyMessage(logger log.Log, layer types.LayerID, nodeID
 	return nil
 }
 
-func (b *Broker) cleanOldLayers() {
+func (b *Broker) CleanOldLayers(current types.LayerID) {
```
piersy, line 356: Why did you decide to call this for each layer tick?

This is going to cause problems because the tracker can be deleted for a layer before the layer is unregistered, which can result in sending more than one malfeasance proof for a given identity.

Previously layers would get unregistered when the consensus process completed or at registration time if the limit was exceeded. It seems simplest to me to remove eligibility trackers at those points, which will keep eligibility trackers and outboxes in sync.

countvonzero: > This is going to cause problems because the tracker can be deleted for a layer before the layer is unregistered

thanks. can you help me identity the codepath where this would happen?

`CleanOldLayers` figures out the highest layer without an entry in `outbox`, i.e. it's already unregistered, or never registered. when a node gets spammed, we will create a tracker for those early messages (same problem you spotted earlier with `pending`) and they will never get cleaned. it is safer to always clean up after each tick. because after `onTick` returns, the node either has an entry in outbox for the current layer or it doesnt.

i deliberately do not want to keep trackers and outbox in sync for 2 reasons:
- i don't want to get into the `isEarly` logic like `outbox`
- it gets clean up regardless of whether a layer register successfully or not.

countvonzero: i also added `pending` to be cleaned up with `trackers`.

piersy: > thanks. can you help me identity the codepath where this would happen?

Actually I think it's fine, I hadn't realised that CleanOldLayers is not going to affect currently running layers.

piersy (approved):
Looks good!

countvonzero:
bors merge

PR #4524, opened 2023-06-15
[Merged by Bors] - tortoise: use zap directly and make logging cheaper

dshulyak:
noticed that our wrapper, string formatting for fixed float, and logger initialization make tortoise unnecessarily slow during restarts. flamegraph below explains the problem

### DIFF
--- a/tortoise/full_test.go
+++ b/tortoise/full_test.go
@@ -61,7 +61,7 @@ func TestFullBallotFilter(t *testing.T) {
 			config := Config{}
 			config.BadBeaconVoteDelayLayers = tc.distance
 			require.Equal(t, tc.expect, newFullTortoise(config, state).shouldBeDelayed(
-				logtest.New(t), &tc.ballot))
+				logtest.Zap(t), &tc.ballot))
 		})
 	}
 }
@@ -321,7 +321,7 @@ func TestFullCountVotes(t *testing.T) {
 	} {
 		tc := tc
 		t.Run(tc.desc, func(t *testing.T) {
-			logger := logtest.New(t)
+			logger := logtest.Zap(t)
 			tortoise := defaultAlgorithm(t)
 			var activeset []types.ATXID
 			for i := range tc.activeset {
@@ -544,7 +544,7 @@ func TestFullVerify(t *testing.T) {
 				}
 				layer.blocks = append(layer.blocks, block)
 			}
-			require.Equal(t, tc.validity != nil, full.verify(logtest.New(t), target))
+			require.Equal(t, tc.validity != nil, full.verify(logtest.Zap(t), target))
 			if tc.validity != nil {
 				for i, expect := range tc.validity {
 					require.Equal(t, expect, layer.blocks[i].validity)
--- a/tortoise/tortoise.go
+++ b/tortoise/tortoise.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/spacemeshos/fixed"
+	"go.uber.org/zap"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/log"
@@ -21,7 +22,7 @@ var errBeaconUnavailable = errors.New("beacon unavailable")
 
 type turtle struct {
 	Config
-	logger log.Log
+	logger *zap.Logger
 
 	*state
 
@@ -41,7 +42,7 @@ type turtle struct {
 }
 
 // newTurtle creates a new verifying tortoise algorithm instance.
-func newTurtle(logger log.Log, config Config) *turtle {
+func newTurtle(logger *zap.Logger, config Config) *turtle {
 	t := &turtle{
 		Config: config,
 		state:  newState(),
@@ -84,10 +85,10 @@ func (t *turtle) evict(ctx context.Context) {
 	if !ok {
 		return
 	}
-	t.logger.With().Debug("evict in memory state",
-		log.Stringer("pending", t.pending),
-		log.Stringer("from_layer", t.evicted.Add(1)),
-		log.Stringer("upto_layer", windowStart),
+	t.logger.Debug("evict in memory state",
+		zap.Stringer("pending", t.pending),
+		zap.Stringer("from_layer", t.evicted.Add(1)),
+		zap.Stringer("upto_layer", windowStart),
 	)
 	if !windowStart.After(t.evicted) {
 		return
@@ -128,7 +129,7 @@ func (t *turtle) EncodeVotes(ctx context.Context, conf *encodeConf) (*types.Opin
 		return nil, err
 	}
 	var (
-		logger = t.logger.WithContext(ctx)
+		logger = t.logger.With(log.ZContext(ctx))
 		err    error
 
 		current = t.last.Add(1)
@@ -152,18 +153,18 @@ func (t *turtle) EncodeVotes(ctx context.Context, conf *encodeConf) (*types.Opin
 			opinion, err = t.encodeVotes(ctx, base, t.evicted.Add(1), current)
 			if err == nil {
 				metrics.LayerDistanceToBaseBallot.WithLabelValues().Observe(float64(t.last - base.layer))
-				logger.With().Info("encoded votes",
-					log.Stringer("base ballot", base.id),
-					log.Stringer("base layer", base.layer),
-					log.Stringer("voting layer", current),
-					log.Inline(opinion),
+				logger.Info("encoded votes",
+					zap.Stringer("base ballot", base.id),
+					zap.Stringer("base layer", base.layer),
+					zap.Stringer("voting layer", current),
+					zap.Inline(opinion),
 				)
 				return opinion, nil
 			}
-			logger.With().Debug("failed to encode votes using base ballot id",
-				base.id,
-				log.Err(err),
-				log.Stringer("current layer", current),
+			logger.Debug("failed to encode votes using base ballot id",
+				zap.Stringer("ballot", base.id),
+				zap.Error(err),
+				zap.Stringer("current layer", current),
 			)
 		}
 	}
@@ -180,9 +181,10 @@ func (t *turtle) encodeVotes(
 	start types.LayerID,
 	current types.LayerID,
 ) (*types.Opinion, error) {
-	logger := t.logger.WithContext(ctx).WithFields(
-		log.Stringer("base layer", base.layer),
-		log.Stringer("current layer", current),
+	logger := t.logger.With(
+		log.ZContext(ctx),
+		zap.Stringer("base layer", base.layer),
+		zap.Stringer("current layer", current),
 	)
 	votes := types.Votes{
 		Base: base.id,
@@ -197,7 +199,7 @@ func (t *turtle) encodeVotes(
 			return nil, fmt.Errorf("ballot %s can't be used as a base ballot", base.id)
 		}
 		if lvote.vote != abstain && !layer.hareTerminated {
-			logger.With().Debug("voting abstain on the layer", lvote.lid)
+			logger.Debug("voting abstain on the layer", zap.Uint32("lid", lvote.lid.Uint32()))
 			votes.Abstain = append(votes.Abstain, lvote.lid)
 			continue
 		}
@@ -213,14 +215,14 @@ func (t *turtle) encodeVotes(
 			}
 			switch vote {
 			case support:
-				logger.With().Debug("support before base ballot", log.Inline(block))
+				logger.Debug("support before base ballot", zap.Inline(block))
 				votes.Support = append(votes.Support, block.header())
 			case against:
-				logger.With().Debug("explicit against overwrites base ballot opinion", log.Inline(block))
+				logger.Debug("explicit against overwrites base ballot opinion", zap.Inline(block))
 				votes.Against = append(votes.Against, block.header())
 			case abstain:
-				logger.With().Error("layers that are not terminated should have been encoded earlier",
-					log.Inline(block), log.Stringer("reason", reason),
+				logger.Error("layers that are not terminated should have been encoded earlier",
+					zap.Inline(block), zap.Stringer("reason", reason),
 				)
 			}
 		}
@@ -229,7 +231,7 @@ func (t *turtle) encodeVotes(
 	for lid := base.layer; lid.Before(current); lid = lid.Add(1) {
 		layer := t.layer(lid)
 		if !layer.hareTerminated {
-			logger.With().Debug("voting abstain on the layer", lid)
+			logger.Debug("voting abstain on the layer", zap.Uint32("lid", lid.Uint32()))
 			votes.Abstain = append(votes.Abstain, lid)
 			continue
 		}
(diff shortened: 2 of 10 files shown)


dshulyak:
image

tortoise/full_test.go
```diff
@@ -321,7 +321,7 @@ func TestFullCountVotes(t *testing.T) {
 	} {
 		tc := tc
 		t.Run(tc.desc, func(t *testing.T) {
-			logger := logtest.New(t)
+			logger := logtest.Zap(t)
```
fasmat, line 324: I think we can do completely away with our wrapper:

```suggestion
			logger := zaptest.New(t)
```

tortoise/util.go
```diff
@@ -95,23 +96,21 @@ func verifyLayer(logger log.Log, blocks []*blockInfo, getDecision func(*blockInf
...
-					return nil
-				})),
-		)
-	}
-	return true
+	return true, changes
+}
+
+func zapBlocks(blocks []*blockInfo) zap.Field {
+	return zap.Array("blocks", log.ArrayMarshalerFunc(func(encoder log.ArrayEncoder) error {
```
fasmat: Same here. Imo we can go all the way and use `zapcore` directly instead of the aliases in the `go-spacemesh/log` package.

tortoise/verifying.go
```diff
@@ -32,23 +33,23 @@ func (v *verifying) resetWeights(voted types.LayerID) {
...
-		log.ShortStringer("local opinion", prev.opinion),
-		log.Bool("bad beacon", ballot.conditions.badBeacon),
-		log.Stringer("weight", ballot.weight),
-		log.Uint64("reference height", prev.verifying.referenceHeight),
-		log.Uint64("ballot height", ballot.reference.height),
-		log.Bool("counted", counted),
+	logger.Debug("count ballot in verifying mode",
+		zap.Uint32("lid", ballot.layer.Uint32()),
+		zap.Stringer("ballot", ballot.id),
+		log.ZShortStringer("ballot opinion", ballot.opinion()),
```
fasmat, line 47: Do we need the indirection with `log.ZShortStringer`? This does the same without the need to allocate additional objects just to print a short string:

```suggestion
		zap.String("ballot opinion", ballot.opinion().ShortString()),
```

tortoise/verifying_test.go
```diff
@@ -116,7 +116,7 @@ func TestVerifyingProcessLayer(t *testing.T) {
 	} {
 		tc := tc
 		t.Run(tc.desc, func(t *testing.T) {
-			logger := logtest.New(t)
+			logger := logtest.Zap(t)
```
fasmat, line 119: See comment above

tortoise/tortoise.go
```diff
@@ -180,10 +182,6 @@ func (t *turtle) encodeVotes(
 	start types.LayerID,
 	current types.LayerID,
 ) (*types.Opinion, error) {
-	logger := t.logger.WithContext(ctx).WithFields(
-		log.Stringer("base layer", base.layer),
-		log.Stringer("current layer", current),
-	)
```
fasmat, line 186: I think a dedicated instance of logger with pre-set fields should consume less memory than re-declaring all fields on every log, since it only has to be allocated once:

```golang
logger := t.logger.With(
	log.ZContext(ctx),
	zap.Stringer("base layer", base.layer),
	zap.Stringer("current layer", current),
)
```

And then re-using `logger` should theoretically consume less memory than repeating the fields on every log.

tortoise/tortoise.go
```diff
@@ -348,59 +367,56 @@ func (t *turtle) onLayer(ctx context.Context, last types.LayerID) {
...
 	}
 	return nil
 }
 
 func (t *turtle) verifyLayers() {
 	// TODO(dshulyak) simplify processing of layers and notifications
-	var (
-		logger = t.logger.WithFields(
-			log.Stringer("last layer", t.last),
-		)
```
fasmat, line 384: Same here as above

tortoise/tortoise.go
```diff
@@ -725,17 +760,17 @@ func (t *turtle) onBallot(ballot *types.BallotTortoiseData) error {
...
-		logger.With().Debug("ballot has different beacon",
-			log.Uint32("layer_id", lid.Uint32()),
-			log.Stringer("block", bid),
-			log.ShortStringer("ballot_beacon", beacon),
-			log.ShortStringer("epoch_beacon", epoch.beacon),
+		t.logger.Debug("ballot has different beacon",
+			zap.Uint32("layer_id", lid.Uint32()),
+			zap.Stringer("block", bid),
+			log.ZShortStringer("ballot_beacon", beacon),
+			log.ZShortStringer("epoch_beacon", epoch.beacon),
```
fasmat, line 773: Imo we don't need the `ZShortStringer` wrapper:

```suggestion
			zap.String("ballot_beacon", beacon.ShortString()),
			zap.String("epoch_beacon", epoch.beacon.ShortString()),
```

tortoise/verifying.go
dshulyak: ballot.opinion().ShortString() allocates the string. so it is completly the opposite of what is needed

tortoise/tortoise.go
dshulyak: no this allocates logger and a lot of fields regardless if this fields will be even used,  this should not be used like that at all

tortoise/tortoise.go
dshulyak: we do need it

tortoise/verifying.go
fasmat: I don't think it makes any difference, please correct me if my Benchmark is wrong:

```golang
import (
	"testing"

	"go.uber.org/zap"
	"go.uber.org/zap/zaptest"
)

type shortString struct {
	val string
}

func (s *shortString) ShortString() string {
	return s.val
}

func Benchmark_ShortString(b *testing.B) {
	s := &shortString{val: "test"}

	log := zaptest.NewLogger(b)

	for i := 0; i < b.N; i++ {
		log.Info("test", zap.String("val", s.ShortString()))
	}
}

func Benchmark_ZShortString(b *testing.B) {
	s := &shortString{val: "test"}

	log := zaptest.NewLogger(b)

	for i := 0; i < b.N; i++ {
		log.Info("test", ZShortStringer("val", s))
	}
}
```

Results:

```bash
Benchmark_ShortString-6           467226              5961 ns/op            1446 B/op         15 allocs/op
--- BENCH: Benchmark_ShortString-6
    logger.go:130: 2023-06-15T09:00:07.697Z     INFO    test    {"val": "test"}
        ... [output truncated]
Benchmark_ZShortString-6          503100              8767 ns/op            1432 B/op         16 allocs/op
--- BENCH: Benchmark_ZShortString-6
    logger.go:130: 2023-06-15T09:00:10.511Z     INFO    test    {"val": "test"}
        ... [output truncated]
PASS
ok      github.com/spacemeshos/go-spacemesh/log 7.271s
```

fasmat: it would even indicate that the indirection causes an additional allocation and slowdown

tortoise/tortoise.go
poszu: > we do need it

Would you mind to explain why?

tortoise/verifying.go
dshulyak: that variable is not a string, but bytes - to make a short string it needs to be encoded with hex. therefore your example is not useful

tortoise/tortoise.go
fasmat: Again I'm not sure about this:

```golang
func Benchmark_DedicatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b)
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	logger := log.With(
		zap.Stringer("base layer", baseLayer),
		zap.Stringer("current layer", currentLayer),
	)

	for i := 0; i < b.N; i++ {
		logger.Info("test")
	}
}

func Benchmark_RepeatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b)
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	for i := 0; i < b.N; i++ {
		log.Info("test",
			zap.Stringer("base layer", baseLayer),
			zap.Stringer("current layer", currentLayer),
		)
	}
}
```

Results:

```shell
Benchmark_DedicatedLogger-6       502114              8032 ns/op            1498 B/op         14 allocs/op
--- BENCH: Benchmark_DedicatedLogger-6
    logger.go:130: 2023-06-15T09:11:15.790Z     INFO    test    {"base layer": "10", "current layer": "20"}
        ... [output truncated]
Benchmark_RepeatedLogger-6        506257              8183 ns/op            1755 B/op         17 allocs/op
--- BENCH: Benchmark_RepeatedLogger-6
    logger.go:130: 2023-06-15T09:11:19.851Z     INFO    test    {"base layer": "10", "current layer": "20"}
        ... [output truncated]
PASS
PASS
```

tortoise/tortoise.go
dshulyak: beacon and some other things are stored as bytes , not string. in order to transform them into string, you need to encode them in hex. therefore such allocation should be performed lazily, only when needed

tortoise/tortoise.go
dshulyak: but you againt doing the wrong thing..

tortoise/verifying.go
poszu: @fasmat
```go
log.Info("test", zap.String("val", s.ShortString()))
```
will call ` s.ShortString()` eagerly even if INFO is disabled.

tortoise/tortoise.go
poszu: Makes sense

tortoise/tortoise.go
fasmat: wdym?

fasmat: In both cases the fields are evaluated lazily, so even if there is transformation / encoding caused by calling `String()` this only happens when the log is actually printed?

poszu: @fasmat, consider:
```go
func Benchmark_DedicatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b, zaptest.Level(zap.ErrorLevel))
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	for i := 0; i < b.N; i++ {
		logger := log.With(
			zap.Stringer("base layer", baseLayer),
			zap.Stringer("current layer", currentLayer),
		)
		logger.Info("test")
	}
}

func Benchmark_RepeatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b, zaptest.Level(zap.ErrorLevel))
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	for i := 0; i < b.N; i++ {
		log.Info("test",
			zap.Stringer("base layer", baseLayer),
			zap.Stringer("current layer", currentLayer),
		)
	}
}
```

>Benchmark_DedicatedLogger-20    	 2257372	       539.4 ns/op	    1414 B/op	       8 allocs/op
Benchmark_RepeatedLogger-20     	32398957	        35.94 ns/op	     128 B/op	       1 allocs/op

fasmat: @poszu your test is incorrect, you are allocating a new logger every time which is not the case in our code:

```golang
func Benchmark_DedicatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b, zaptest.Level(zap.ErrorLevel))
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	logger := log.With(
		zap.Stringer("base layer", baseLayer),
		zap.Stringer("current layer", currentLayer),
	)

	for i := 0; i < b.N; i++ {
		logger.Info("test")
	}
}

func Benchmark_RepeatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b, zaptest.Level(zap.ErrorLevel))
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	for i := 0; i < b.N; i++ {
		log.Info("test",
			zap.Stringer("base layer", baseLayer),
			zap.Stringer("current layer", currentLayer),
		)
	}
}
```

> Benchmark_DedicatedLogger-6   	262413741	         4.553 ns/op	       0 B/op	       0 allocs/op
> Benchmark_RepeatedLogger-6    	30456820	        40.82 ns/op	     128 B/op	       1 allocs/op

dshulyak: no, this is wrong. With() will allocate immediately

poszu: It is exactly the case, that's why I moved it into the loop. The point is that (before this PR changes) the `logger` was created on each function call, regardless of the log level. In effect, we pay the cost even if the log is not rendered in the end.

tortoise/full_test.go
dshulyak: logtest also check env variable

tortoise/verifying_test.go
dshulyak: logtest checks env varaiable

pigmej (approved):
Makes perf improvement. Good for me. Even if it would require some adjustments later then I'd still go with it now and land on the networks.

tortoise/tortoise.go
fasmat: > no, this is wrong. With() will allocate immediately

It creates a copy of the logger, if the original logger had lazy evaluated fields those will still not be evaluated.

> The point is that (before this PR changes) the logger was created on each function call, regardless of the log level.

But this is still comparing apples to oranges.

The creation of a child logger costs time and memory yes, but we save that on every subsequent call to the logger, because the fields don't have to be evaluated again. So it depends on how many logs statements we do with the new instance:

```golang
func Benchmark_DedicatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b, zaptest.Level(zap.ErrorLevel))
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	for i := 0; i < b.N; i++ {
		logger := log.With(
			zap.Stringer("base layer", baseLayer),
			zap.Stringer("current layer", currentLayer),
		)

		for j := 0; j < 15; j++ {
			logger.Info("test")
		}
	}
}

func Benchmark_RepeatedLogger(b *testing.B) {
	log := zaptest.NewLogger(b, zaptest.Level(zap.ErrorLevel))
	baseLayer := bytes.NewBufferString("10")
	currentLayer := bytes.NewBufferString("20")

	for i := 0; i < b.N; i++ {
		for j := 0; j < 15; j++ {
			log.Info("test",
				zap.Stringer("base layer", baseLayer),
				zap.Stringer("current layer", currentLayer),
			)
		}
	}
}
```

> Benchmark_DedicatedLogger-6   	 1984636	       538.2 ns/op	    1412 B/op	       8 allocs/op
> Benchmark_RepeatedLogger-6    	 2044924	       603.3 ns/op	    1920 B/op	      15 allocs/op

The breakeven is somewhere between 10 and 15 logs, more than that and the child logger becomes more efficient than repeating fields, independent of if the logs are printed or not.

dshulyak: > It creates a copy of the logger, if the original logger had lazy evaluated fields those will still not be evaluated.

it will call String() method even if log is not printed

fasmat: > it will call String() method even if log is not printed

That's not true. `With()` calls this method to create a new instance of `Logger`: https://github.com/uber-go/zap/blob/master/logger.go#L290 no fields are evaluated in the process.

Try it yourself! On my machine `DedicatedLogger` outperforms `RepeatedLogger` if logs are disabled after 13 `log.Info` lines, if the logs are enabled its already after 3!

poszu: >Try it yourself!

image

dshulyak:
```

type stringer struct {
}

func (s *stringer) String() string {
	fmt.Println("tried it myself")
	return ""
}

func TestString(t *testing.T) {
	log := logtest.New(t).Zap().With(zap.Stringer("s", &stringer{}))
	log.Debug("dassdsa")
}
```

> TEST_LOG_LEVEL=info go test ./tortoise -run=TestString -v
 === RUN   TestString
tried it myself
--- PASS: TestString (0.00s)
PASS
ok  	github.com/spacemeshos/go-spacemesh/tortoise	(cached

fasmat: OK, I concede 🙇

dshulyak:
bors merge

PR #4525, opened 2023-06-15
[Merged by Bors] - actvation: add atxs to tortoise if they are stored

dshulyak:
this codepath fails to call OnAtx because node can't broadcast malfeasence proofs for some reason


countvonzero (approved):

dshulyak:
bors merge

PR #4530, opened 2023-06-15
[Merged by Bors] - Discard objects from future

### LINKED ISSUE #4411: atxs: discard objects from the future
this is another ddos opportunity.

objects from the future will not be used by protocol but they will be kept in state and in memory (in particular in tortoise).

it should be enough to drop atxs after the next epoch. so handlers should accepts activations from (genesis, current+1] epoch. 

additionally we should check that ballot layer is within target epoch of the atx, if this check is not present

fasmat:
## Motivation
Closes #4411 

## Changes
- Add check to `SyntacticallyValidateAtx` to ensure that an ATX cannot claim to be published in an epoch further in the future than current epoch + 1
- Add check to handlers for Proposals and Ballots to ensure their layer is within the target epoch of their referenced ATXs

## Test Plan
- Add test for new verification
- Updated tests in `activation/handler.go` to contain less code repetition. Additionally sub-tests now all run in parallel, the main tests do not, because some of them use different `LayersPerEpoch` which would be racy if they were executed in parallel.
- Added tests for Proposal and Ballot handlers for new checks.

## TODO

## DevOps Notes

### DIFF
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -518,7 +518,7 @@ func (h *Handler) handleGossipAtx(ctx context.Context, peer p2p.Peer, msg []byte
 		return fmt.Errorf("nil nipst in gossip for atx %s", atx.ShortString())
 	}
 
-	h.registerHashes(&atx, peer) // TODO(mafa): should this be done after the ATX was validated?
+	h.registerHashes(&atx, peer)
 	if err := h.fetcher.GetPoetProof(ctx, atx.GetPoetProofRef()); err != nil {
 		return fmt.Errorf("received atx (%v) with syntactically invalid or missing PoET proof (%x): %w",
 			atx.ShortString(), atx.GetPoetProofRef().ShortString(), err,
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -143,7 +143,6 @@ func (h *Handler) ProcessAtx(ctx context.Context, atx *types.VerifiedActivationT
 		atx.ID(),
 		atx.PublishEpoch,
 		log.Stringer("smesher", atx.SmesherID),
-		atx.PublishEpoch,
 	)
 	if err := h.ContextuallyValidateAtx(atx); err != nil {
 		h.log.WithContext(ctx).With().Warning("atx failed contextual validation",
@@ -353,7 +352,7 @@ func (h *Handler) storeAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 		return fmt.Errorf("checking if node is malicious: %w", err)
 	}
 	var proof *types.MalfeasanceProof
-	if err = h.cdb.WithTx(ctx, func(dbtx *sql.Tx) error {
+	if err := h.cdb.WithTx(ctx, func(dbtx *sql.Tx) error {
 		if !malicious {
 			prev, err := atxs.GetByEpochAndNodeID(dbtx, atx.PublishEpoch, atx.SmesherID)
 			if err != nil && !errors.Is(err, sql.ErrNotFound) {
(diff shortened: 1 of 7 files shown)


activation/handler.go
```diff
@@ -513,24 +518,26 @@ func (h *Handler) handleGossipAtx(ctx context.Context, peer p2p.Peer, msg []byte
 		return fmt.Errorf("nil nipst in gossip for atx %s", atx.ShortString())
 	}
 
-	h.registerHashes(&atx, peer)
+	h.registerHashes(&atx, peer) // TODO(mafa): should this be done after the ATX was validated?
```
fasmat: Would be glad about inputs 🙂

dshulyak:
i discovered that we already perform check that ballot is created in the correct epoch https://github.com/spacemeshos/go-spacemesh/blob/develop/proposals/eligibility_validator.go#L134-L137

fasmat:
> i discovered that we already perform check that ballot is created in the correct epoch https://github.com/spacemeshos/go-spacemesh/blob/develop/proposals/eligibility_validator.go#L134-L137

Wouldn't it make more sense to already do this when the ballot is received? We already check if the referenced ATX comes from the same user as the Ballot/Proposal so the check is relatively cheap. I added it in `proposals/handler.go`

dshulyak:
> Wouldn't it make more sense to already do this when the ballot is received? We already check if the referenced ATX comes from the same user as the Ballot/Proposal so the check is relatively cheap. I added it in proposals/handler.go

checking eligibility is always executed on the handler codepath

activation/handler.go
dshulyak: it needs to be registered before handler calls fetcher. this is the way fetcher keeps track of which peers has missing objects

fasmat:
> > Wouldn't it make more sense to already do this when the ballot is received? We already check if the referenced ATX comes from the same user as the Ballot/Proposal so the check is relatively cheap. I added it in proposals/handler.go
> 
> checking eligibility is always executed on the handler codepath

Shall I remove the checks in the handlers for Ballots and Proposals then? Since we are already checking if the owner of the ATX is the same as the node that signed the Ballot/Proposal, the additional check for the same epoch is cheap and we can discard invalid Ballots/Proposals even before checking their eligibility.

dshulyak:
i think it makes sense to keep only original check. it is already run before verifying eligibility vrf signature

fasmat:
The check for matching owner of ATX and ballot also happens twice (once in the handler when receiving and again in `CheckEligibility`). Can this also be removed?

EDIT: I decided to remove the additional checks I implemented as well as the redundant checks for the same SmesherID

fasmat:
bors try

activation/handler.go
```diff
@@ -353,7 +358,7 @@ func (h *Handler) storeAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 		return fmt.Errorf("checking if node is malicious: %w", err)
 	}
 	var proof *types.MalfeasanceProof
-	if err = h.cdb.WithTx(ctx, func(dbtx *sql.Tx) error {
+	if err := h.cdb.WithTx(ctx, func(dbtx *sql.Tx) error {
```
countvonzero, line 361: curious why you want to overwrite these?

countvonzero (approved):

dshulyak (approved):

activation/handler.go
fasmat: No particular reason, I just like to limit the scope of a variable to smaller blocks if possible.

fasmat:
bors merge

PR #4535, opened 2023-06-16
[Merged by Bors] - validate malfeasance proof before saving to db

### LINKED ISSUE #4527: generated malfeasance proof should be validated before saved
## Description
in testnet-05 network, due to #4526, hare malfeasance proofs were saved but failed validation later.
currently malfeasance proofs were saved eagerly (in a db transaction along with the offending data)
so even if it failed validation later the bad data remained in db.

Screenshot from 2023-06-15 04-27-42

Screenshot from 2023-06-15 04-27-58

countvonzero:
## Motivation
Closes #4527


activation/handler.go
```diff
@@ -420,7 +423,7 @@ func (h *Handler) storeAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 		}
 		encodedProof, err := codec.Encode(&gossip)
 		if err != nil {
-			h.log.Fatal("failed to encode MalfeasanceGossip", log.Err(err))
+			h.log.With().Fatal("failed to encode MalfeasanceGossip", log.Err(err))
```
dshulyak: imo programmer mistakes should be with panic, so that trace will be dumped to stderr

mesh/mesh.go
```diff
@@ -53,11 +54,12 @@ type Mesh struct {
 }
 
 // NewMesh creates a new instant of a mesh.
-func NewMesh(cdb *datastore.CachedDB, c layerClock, trtl system.Tortoise, exec *Executor, state conservativeState, logger log.Log) (*Mesh, error) {
+func NewMesh(cdb *datastore.CachedDB, c layerClock, f system.Fetcher, trtl system.Tortoise, exec *Executor, state conservativeState, logger log.Log) (*Mesh, error) {
```
dshulyak: is there an alternative solution that doesn't add fetcher to the mesh?  i don't think that it is ok to add dependency that can do network requests here

hare/hare.go
```diff
@@ -137,6 +138,7 @@ func New(
 	ch chan LayerOutput,
 	syncState system.SyncStateProvider,
 	beacons system.BeaconGetter,
+	fetcher system.Fetcher,
```
dshulyak: this doesn't make sense as well

mesh/mesh.go
dshulyak: is it possible to store the proof inside the handler and don't save it in other modules? handlers can't error on sync validation though, but that should be adjusted anyway

countvonzero: changed malfeasance proof handler to have static functions to validate and save the malfeasance proof with an optional database transaction. so validate and save now is done in the same module.

dshulyak (approved):

countvonzero:
bors merge

dshulyak:
@countvonzero i missed the huge problem in this change. write lock should never be acquired while doing expensive computation (such as verifying signatures, and other check in ValidateAndSave), while db lock is held nobody else can write. it will lead to software failing in spectacular ways. i understand the intentation, but it can't be done in a such naive way

countvonzero:
> @countvonzero i missed the huge problem in this change. write lock should never be acquired while doing expensive computation (such as verifying signatures, and other check in ValidateAndSave), while db lock is held nobody else can write. it will lead to software failing in spectacular ways. i understand the intentation, but it can't be done in a such naive way

ok. i think i will move the call to ValidateAndSave outside the critical block.

countvonzero:
@dshulyak https://github.com/spacemeshos/go-spacemesh/issues/4556

Issue #4510, opened 2023-06-13
proposals: ensure that ballot is used only once


dshulyak:
the proposal referencing same ballot is accepted and upvoted by hare. 

- node should reject proposal if there is another one with same ballot
  can be implemented by making tortoise return an error if ballot already in the state. error has to result in Ignore for p2p, and failure if ballot is added with proposal 
- hare should not vote on proposals from same atx, even is they are stored by the node 

codepath starts https://github.com/spacemeshos/go-spacemesh/blob/develop/proposals/handler.go#L229-L245

this is a first part of the problem, but there was also a bigger issue that node accepted proposal that can't pass signature verification afterwards. it does seem a vulnerability that we can block all together, but would be nice to understand the root cause

below is the layer and proposals that reference same ballot. the only different field is signature

3557 12eb5dc3e0 766227542b
3558 5d1dd48cc3 8e77baa165

dshulyak:
this is a first part of the problem, but there was also a bigger issue that node accepted proposal that can't pass signature verification afterwards, below is the layer and proposals that reference same ballot. it does seem a vulnerability that we can block alltogether, but would be nice to understand the root cause

3557 12eb5dc3e0 766227542b
3558 5d1dd48cc3 8e77baa165

dshulyak:
this behavior was caused by creating a second proposal with a slightly different ballot, ballot had exactly the same id, but different votes, because votes were encoded with different base

```
types.Votes{
  	Base: types.BallotID{
- 		0x49, 0xa3, 0x78, 0xa9, 0xbe, 0x5e, 0xef, 0xe7, 0xe5, 0xbb, 0x63, 0x06, 0x7f, 0x48, 0x80, 0xfb,
- 		0x2e, 0xac, 0x55, 0x1a,
+ 		0x99, 0x29, 0x70, 0xfc, 0x86, 0xe8, 0xde, 0x94, 0x03, 0xbe, 0x15, 0x82, 0xac, 0x17, 0x46, 0x0d,
+ 		0xe0, 0x2c, 0xf1, 0x3e,
  	},
  	Support: nil,
  	Against: nil,
  	Abstain: {s"3556"},
```

so i think that somebody ended up running two nodes with same data

dshulyak:
@countvonzero i opened two changes to make hare and tortoise resilient to this behavior. i will leave issue open if you want to add malfeasence proof with a separate domain for proposals

countvonzero:
[...]

PR #4536, opened 2023-06-16
[Merged by Bors] - tortoise: error if same ballots is added to the state

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4510

proposal that references same ballot was accepted when the second proposal was processed concurrently with the first.
in that case ballot wasn't stored in the database yet, and could pass existence check


countvonzero (approved):

dshulyak:
bors merge

PR #4537, opened 2023-06-16
[Merged by Bors] - hare: dont vote on malicious proposals

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4510

we don't count ballot weight, but we still store equivocating ballots. this is more of a safety measure as we shouldn't even store equivocating proposal.

### DIFF
--- a/hare/hare.go
+++ b/hare/hare.go
@@ -470,10 +470,21 @@ func goodProposals(ctx context.Context, logger log.Log, msh mesh, nodeID types.N
 	if ownHdr != nil {
 		ownTickHeight = ownHdr.TickHeight()
 	}
+	atxs := map[types.ATXID]int{}
+	for _, p := range props {
+		atxs[p.AtxID]++
+	}
 	for _, p := range props {
 		if p.IsMalicious() {
 			logger.With().Warning("not voting on proposal from malicious identity",
 				log.Stringer("id", p.ID()),
+			)
+			continue
+		}
+		if atxs[p.AtxID] > 1 {
+			logger.With().Warning("proposal with same atx added twice in the recorded set",
+				log.Stringer("id", p.ID()),
+				log.Stringer("atxid", p.AtxID),
 			)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


hare/hare.go
```diff
@@ -470,7 +470,24 @@ func goodProposals(ctx context.Context, logger log.Log, msh mesh, nodeID types.N
...
+		atxs[p.AtxID]++
+	}
+	for _, p := range props {
+		if p.IsMalicious() {
+			logger.With().Warning("not voting on proposal from malicious identity",
+				log.Stringer("id", p.ID()),
+			)
+			continue
+		}
+		if atxs[p.AtxID] > 1 {
```
piersy: Why do we need this check?

Won't the proposal be set to malicious when being retrieved from the mesh if there were multiple ballots from the same smesher?

dshulyak: it doesn't do that if smesher registered several proposals with same ballot. that should be fixed.
but i would prefer to have this check regardless of that, as this such bug is too expensive

piersy: Ahh ok so this is the case of multiple proposals with one good ballot? So I guess we'll need a new malfeasance proof type for this?

piersy:
```suggestion
		// Check for multiple proposals sharing a single ballot.
		if atxs[p.AtxID] > 1 {
```

piersy (approved):
One suggested change, but apart from that looks good.

hare/hare.go
dshulyak: or simply reject proposal, but in my mind it is more important to be safe here

hare/hare.go
```diff
@@ -470,7 +470,24 @@ func goodProposals(ctx context.Context, logger log.Log, msh mesh, nodeID types.N
...
+			logger.With().Warning("proposal with same atx added twice in the recorded set",
```
countvonzero: not necessarily twice. might be informational to log the count `atxs[p.AtxID]`

countvonzero (approved):

hare/hare.go
dshulyak: sorryy i don't want to have this message. this is a bug that i observed, it may not be the only one, and i want to have this check regardless of that bahavior

dshulyak:
bors merge

PR #4545, opened 2023-06-19
[Merged by Bors] - p2p: parametrize accept queue length and extend libp2p observability

dshulyak:
bootnodes were slow to accept new connections, when we had to restart whole network. accept queue may have a direct impact on that, so we can increase it to 32/64. it will increase upper bound for resources usage on bootnodes, but should be fine.

new option is `---p2p-accept-queue=64`, and config p2p: {"p2p-accept-queue": 64}.

metrics for tcp transport and libp2p resource manager will be collected if enabled with `--p2p-metrics`. or p2p: {"p2p-metrics": true}


pigmej (approved):

dshulyak:
bors merge

Issue #4542, opened 2023-06-18
sync: investigate "protocol not supported"


countvonzero:
## Description
on testnet-05, one can see the following logs.

syncer had completed a hash resolution with peer `12D3KooWE8cdznCXh9Lu7y5Sgstpi4KC7FLwn1PvFnBoc1fRKWUz` but then when it starts to sync data from that peer then complained `protocols not supported`.

```
2023-06-18 10:17:01.787 | {"L":"WARN","T":"2023-06-18T13:17:01.787-0400","N":"45b2f.sync","M":"failed to sync atxs","node_id":"45b2f9e93836dba4cd50dc4294bdad30ce2d6f722954dac292dbdf2628c897f1","module":"sync","sessionId":"09d4ffee-f793-4c32-84ba-3c7e9bc2f1e0","current":"5151","errmsg":"get epoch info (peer 12D3KooWE8cdznCXh9Lu7y5Sgstpi4KC7FLwn1PvFnBoc1fRKWUz): protocols not supported: [ax/1]","name":"sync"}  
2023-06-18 10:16:41.785 | {"L":"WARN","T":"2023-06-18T13:16:41.785-0400","N":"45b2f.sync","M":"failed to sync atxs","node_id":"45b2f9e93836dba4cd50dc4294bdad30ce2d6f722954dac292dbdf2628c897f1","module":"sync","sessionId":"5e257c11-3373-406a-83a6-92e5d96050c1","current":"5151","errmsg":"get epoch info (peer 12D3KooWE8cdznCXh9Lu7y5Sgstpi4KC7FLwn1PvFnBoc1fRKWUz): protocols not supported: [ax/1]","name":"sync"} 
2023-06-18 10:16:21.785 | {"L":"WARN","T":"2023-06-18T13:16:21.785-0400","N":"45b2f.sync","M":"failed to sync atxs","node_id":"45b2f9e93836dba4cd50dc4294bdad30ce2d6f722954dac292dbdf2628c897f1","module":"sync","sessionId":"d4c4da07-c677-438a-9d48-a70fe9f2677b","current":"5151","errmsg":"get epoch info (peer 12D3KooWE8cdznCXh9Lu7y5Sgstpi4KC7FLwn1PvFnBoc1fRKWUz): protocols not supported: [ax/1]","name":"sync"}
[...]

PR #4546, opened 2023-06-19
[Merged by Bors] - p2p: enable discovery after all local protocols were registered

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4542

libp2p communicates the set of enabled protocols after completing secure handshake.
if some of the protocols were not registered before completing initial identify protocol round, it will result in errors
as in attached github issue.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -905,7 +905,9 @@ func (app *App) initServices(ctx context.Context, poetClients []activation.PoetP
 			peersync.WithConfig(app.Config.TIME.Peersync),
 		)
 	}
-
+	if err := app.host.Start(); err != nil {
+		return err
+	}
 	return nil
 }
 
--- a/p2p/peerexchange/discovery.go
+++ b/p2p/peerexchange/discovery.go
@@ -44,6 +44,7 @@ type Discovery struct {
 	host   host.Host
 	cfg    Config
 
+	ctx    context.Context
 	cancel context.CancelFunc
 	eg     errgroup.Group
 
@@ -60,6 +61,7 @@ func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
 		cfg:    config,
 		logger: logger,
 		host:   h,
+		ctx:    ctx,
 		cancel: cancel,
 		book:   book.New(),
 	}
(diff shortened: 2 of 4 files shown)


pigmej (approved):

p2p/peerexchange/discovery.go
```diff
@@ -103,10 +105,15 @@ func New(logger log.Log, h host.Host, config Config) (*Discovery, error) {
...
 		}
 	}
-	d.scanPeers(ctx)
 	return d, nil
 }
 
+// StartScan starts background goroutine to connect with known peers
+// and scan for new ones.
+func (d *Discovery) StartScan() {
+	d.scanPeers(d.ctx)
```
fasmat, line 114: Instead of passing the context to the constructor, imo it should be passed to `StartScan` instead.

node/node.go
```diff
@@ -905,7 +905,9 @@ func (app *App) initServices(ctx context.Context, poetClients []activation.PoetP
 			peersync.WithConfig(app.Config.TIME.Peersync),
 		)
 	}
-
+	if err := app.host.Start(); err != nil {
+		return err
+	}
```
fasmat, line 910: can be simplified to

```suggestion
	return app.host.Start()
```

p2p/peerexchange/discovery.go
dshulyak: lifecycle is controlled by Start/Stop . passing context to contstructor is not a good pattern if you have Start/Stop

dshulyak:
bors merge

p2p/peerexchange/discovery.go
fasmat: yes, that's what I said 🙂

fasmat: The context at the moment instantiated when `Discovery` is created, instead I think it would be better to pass it to `StartScan`.

dshulyak: context is used in two other places when discovery is created

dshulyak: sorryy
> lifecycle is controlled by Start/Stop . passing context to contstructor is not a good pattern if you have Start/Stop

i wanted to say that passing ctx to Start is not a good pattern if you have Start/Stop. because then services may be stopped even without calling Stop

node/node.go
dshulyak: i intentionally made it so that next person can write their own thing without changing previous code

p2p/peerexchange/discovery.go
fasmat: `context.Context` should not be used to control the lifecycle of an object. Instead it should only be used to be able to cancel the function/method it is passed into. `StartScan` breaks this pattern which is used for all other methods of `Discovery`.

fasmat: what I'm trying to say is: canceling the context should not be equivalent to calling `Stop` instead it should just cancel the starting

PR #4547, opened 2023-06-19
Wait for Builder goroutines to finish in UT

### LINKED ISSUE #4544: TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError
> panic: Fail in goroutine after TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError has completed

goroutine 2163 [running]:
testing.(*common).Fail(0xc0000d6b60)
	/opt/hostedtoolcache/go/1.19.10/x64/src/testing/testing.go:824 +0x1a8
testing.(*common).FailNow(0xc0000d6b60)
	/opt/hostedtoolcache/go/1.19.10/x64/src/testing/testing.go:847 +0x51
testing.(*common).Fatalf(0xc0000d6b60, {0x1c3c0b3, 0x2e}, {0xc00011fbd0, 0x5, 0x5})
	/opt/hostedtoolcache/go/1.19.10/x64/src/testing/testing.go:955 +0xaf
github.com/golang/mock/gomock.(*Controller).Call.func1(0xc000185e60, {0x1b48860?, 0xc0002862a0}, {0x1c0064a, 0x4}, {0x0, 0x0, 0x0})
	/home/runner/go/pkg/mod/github.com/golang/mock@v1.6.0/gomock/controller.go:231 +0x46d
github.com/golang/mock/gomock.(*Controller).Call(0xc000185e60, {0x1b48860, 0xc0002862a0}, {0x1c0064a, 0x4}, {0x0, 0x0, 0x0})
	/home/runner/go/pkg/mod/github.com/golang/mock@v1.6.0/gomock/controller.go:247 +0xce
github.com/spacemeshos/go-spacemesh/log.(*MockLogger).With(0xc0002862a0)
	/home/runner/work/go-spacemesh/go-spacemesh/log/log_mock.go:125 +0x9b
github.com/spacemeshos/go-spacemesh/activation.(*Builder).generateInitialPost(0xc000317880, {0x1f1e578, 0xc00032a000})
	/home/runner/work/go-spacemesh/go-spacemesh/activation/activation.go:284 +0x3fb
github.com/spacemeshos/go-spacemesh/activation.(*Builder).run(0xc000317880, {0x1f1e578, 0xc00032a000})
	/home/runner/work/go-spacemesh/go-spacemesh/activation/activation.go:250 +0x68
github.com/spacemeshos/go-spacemesh/activation.(*Builder).StartSmeshing.func1()
	/home/runner/work/go-spacemesh/go-spacemesh/activation/activation.go:209 +0x31f
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75 +0x87
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:72 +0x12f
FAIL	github.com/spacemeshos/go-spacemesh/activation	0.483s

poszu:
## Motivation
The UT is flaky because it mocks a logger and an extra warning log is sometimes emitted. The log is emitted only sporadically because there is a race between the background Builder's goroutine (spawned in b.run() and the test finishing and clearing the test temp directory).
Closes #4544 

## Changes
Wait for Builder's goroutines to finish before returning from the test.

## Test Plan
The test should not be flaky anymore

## DevOps Notes

### DIFF
--- a/activation/activation_test.go
+++ b/activation/activation_test.go
@@ -345,6 +345,9 @@ func TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError(t 
 		require.Fail(t, "test timed out or failed")
 	}
 
+	tab.stop()
+	tab.eg.Wait()
+
 	// Now verify that a panic does not occur if PrepareInitializer returns an error
 	l = log.NewMockLogger(gomock.NewController(t))
 	tab = newTestBuilder(t)
@@ -366,6 +369,7 @@ func TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError(t 
 	goroutineCount := runtime.NumGoroutine()
 	tab.StartSmeshing(tab.coinbase, PostSetupOpts{})
 	require.Equal(t, goroutineCount, runtime.NumGoroutine())
+	tab.eg.Wait()
 }
 
 func TestBuilder_StopSmeshing_failsWhenNotStarted(t *testing.T) {


activation/activation_test.go
```diff
@@ -366,6 +369,7 @@ func TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError(t
 	goroutineCount := runtime.NumGoroutine()
 	tab.StartSmeshing(tab.coinbase, PostSetupOpts{})
 	require.Equal(t, goroutineCount, runtime.NumGoroutine())
```
fasmat, line 371: imo counting the go routines isn't necessary if we call `eg.Wait()` in the test.

activation/activation_test.go
```diff
@@ -345,6 +345,9 @@ func TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError(t
 		require.Fail(t, "test timed out or failed")
 	}
 
+	tab.stop()
+	tab.eg.Wait()
```
fasmat, line 349: I think we can just end the test here and start a new test.

first test checks if an error on `StartSmeshing` calls `log.Panic` and second test checks if an error on `PrepareInitializer` does not call `StartSession`.

activation/activation_test.go
piersy: If smeshing had been started `eg.wait()` would take a very long time to return. So i think the go-routine count is useful or we could timeout while waiting for `eg.wait()` to return.

fasmat: But smeshing never starts in this test, because `PrepareInitializer` returns an error and we are asserting that this error is propergated and returned by `tab.StartSmeshing`.

Imo this tests is sufficient to check that `StartSmeshing` doesn't start a smeshing session if `PrepareInitializer` fails:

```golang
func TestBuilder_StartSmeshing_SessionNotStartedOnFailPrepare(t *testing.T) {
	l := log.NewMockLogger(gomock.NewController(t))
	tab := newTestBuilder(t)
	tab.log = l

	// Stub these methods in case they get called
	tab.mpost.EXPECT().GenerateProof(gomock.Any(), gomock.Any()).AnyTimes().Return(&types.Post{}, &types.PostMetadata{}, nil)
	tab.mclock.EXPECT().AwaitLayer(gomock.Any()).AnyTimes()

	// Set expectations, we don't expect StartSession to be called, since
	// StartSmeshing should exit early if PrepareInitializer returns an error.
	testError := errors.New("avoid panic")
	tab.mpost.EXPECT().PrepareInitializer(gomock.Any(), gomock.Any()).Return(testError)

	require.ErrorIs(t, tab.StartSmeshing(tab.coinbase, PostSetupOpts{}), testError)
	tab.eg.Wait()
}
```

`eg.Wait()` should return almost immediately here and if for some reason it doesn't then this test should fail (in a sense that it will time out after whatever we have currently defined as test timeout).

fasmat: > If smeshing had been started `eg.wait()` would take a very long time to return. So i think the go-routine count is useful or we could timeout while waiting for `eg.wait()` to return.

Counting the go routines doesn't guarantee that eg.wait() doesn't block, afaik the runtime can start go routines any time (for instance to garbage collect) so the count might be off anyway.

Additionally if smeshing had been started the test would fail with a proper error message anyway (along the lines of `StartSmeshing` has unexpectedly been called). If it does immediately or after the test times out is imo not as important as making sure the test properly cleans up after itself when it passes 🤷 

EDIT: `eg.Wait()` will never return if smeshing started and it wasn't stopped explicitly with `StopSmeshing()`, so the test is guaranteed to time out, and even if that wouldn't be the case the "`StartSmeshing` has unexpectedly been called" message from go-mock would mark the test as failed too.

poszu: @fasmat, I just want to unblock an annoyingly flaky test, not to re-implement it. TBH, I'd remove it completely and probably refactor the code to not panic in the first place.

piersy: @poszu Panic was the [chosen approach](https://github.com/spacemeshos/go-spacemesh/issues/4292#issuecomment-1514361910) here , but it does make testing tricky. The best solution IMO is to go for a streaming API, but of course that implies a larger change.

@fasmat The use of `runtime.NumGoroutine()` as you point out is not a good idea, due to runtime created go routines. So I would remove that. But then as you noted `eg.Wait()` can get stuck so to prevent that blocking the entire test run we can just timeout in a few seconds, as we know that this should return pretty much instantly.

activation/activation_test.go
```diff
@@ -366,6 +369,7 @@ func TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError(t
 	goroutineCount := runtime.NumGoroutine()
 	tab.StartSmeshing(tab.coinbase, PostSetupOpts{})
 	require.Equal(t, goroutineCount, runtime.NumGoroutine())
+	tab.eg.Wait()
```
piersy, line 372: So I think this would solve the issue with getting stuck, it is quite tricky though. Given the trickyness of this test, I'd also be happy with just removing the test altogether.
```suggestion
	closeChan := make(chan struct{})
	go func() {
		tab.eg.Wait()
		closeChan <- struct{}{}
	}()
	select {
	case closeChan <- struct{}{}:
	case <-time.NewTimer(time.Second * 5).C:
		tab.stop()
		<-closeChan
		t.Fatal("timed out waiting for close")
	}
```

fasmat: This kinda defeats the whole purpose of the `errgroup`: If the timer fires then the test is aborted with a go routine still running in the background when the next test starts.

Imo we should just let the test get stuck, we have timeouts in place that stop test execution if a test times out: `go test -timeout 5m` so if a single test takes longer than 5 mins to complete test execution will be aborted, which imo is a safer approach.

activation/activation_test.go
fasmat: `eg.Wait()` getting stuck is not a bad thing imo. It signals that the test failed, and we have a test timeout. So if a test takes longer than 5 minutes to complete, test execution will be aborted.

fasmat: @poszu "re-implementing" is a bit much for splitting a test into 2 without changing the implementation. 😅
The tests also only changes a single assert.

first half:
Screenshot 2023-06-20 at 11 50 20

second half:
Screenshot 2023-06-20 at 11 50 47

dshulyak (approved):

activation/activation_test.go
piersy: It _is_ waiting for the go-routines to finish, after the timeout it waits on the close channel for the errgroup to return.

But it is tricky! Which is probably why you misread it.

fasmat:
Fixed as part of https://github.com/spacemeshos/go-spacemesh/pull/4575

PR #4555, opened 2023-06-20
[Merged by Bors] - tortoise: add on malfeasence handler

### LINKED ISSUE #4538: tortoise: ignore equivocating atxs for the purposes of threshold calculation
implementation stores equivocating atxs and tortoise uses its weight both in local and global threshold computation.
it makes it very easy to halt protocol by creating many atxs from the same smesher. note that we won't count weight from atxs ballots, so this is purely about halting tortoise

OnAtx method should be updated with information about whether atx equivocated (malicious bool). it should also update previously added atx and set its weight to zero.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4538

main change is to handle equivocating atxs in tortoise, so that they can't skew global threshold by making it unreachable.
for this purpose there is OnMalfeasance method that is expected to be called when equivocating ballot or atx is found, hare equivocation doesn't matter for tortoise, so it can be added later at its own timing.

tortoise will store mapping with all malicious identities without pruning them, even considering some unrealistic number of such identities (1_000_000) that will only be about ~30MB, which is clearly not a ddos vector.

field for ballot.Malicious can be dropped in future pr.


dshulyak:
bors try

malfeasance/handler.go
```diff
@@ -72,7 +75,7 @@ func (h *Handler) HandleMalfeasanceProof(ctx context.Context, peer p2p.Peer, dat
 		updateMetrics(p.Proof)
 		return nil
 	}
-	return ValidateAndSave(ctx, h.logger, h.cdb, h.cdb, h.edVerifier, h.cp, &p)
```
countvonzero: if you do it report malfeasance to tortoise in `HandleSyncedMalfeasanceProof` and `HandleMalfeasanceProof` (before the self check), then malfeasance proofs reported by hare would also be captured.

countvonzero (commented):
i am removing calls to `ValidateAndSave` in atx handler, hare, and mesh in https://github.com/spacemeshos/go-spacemesh/pull/4558
it's enough to report malfeasance to tortoise on the gossip codepath, as every proof, internally or externally, will go through it.

dshulyak:
> it's enough to report malfeasance to tortoise on the gossip codepath, as every proof, internally or externally, will go through it.

if node is not synced then it won't be called. beside that OnMalfesance has to be called before the call to OnBallot/OnAtx with object that equivocated

countvonzero:
> if node is not synced then it won't be called. 

`HandleSyncedMalfeasanceProof` will take care of the sync path when a node is not synced.

> beside that OnMalfesance has to be called before the call to OnBallot/OnAtx with object that equivocated

but malfeasance proof isn't valid if an identity doesn't exist locally. in syncer, it syncs ATXs first before syncing malfeasance proofs.

dshulyak:
> HandleSyncedMalfeasanceProof will take care of the sync path when a node is not synced.

this absolutely should not depend on what data is given by peer

> but malfeasance proof isn't valid if an identity doesn't exist locally. in syncer, it syncs ATXs first before syncing malfeasance proofs.

it should not pass 2 atxs in the same epoch from identity that equivocated, which is exactly what i am doing in the change.
so the order that i need is:
on atx 1
on malfeasance
on atx 2 

and the same for ballots

tortoise/tortoise.go
```diff
@@ -525,7 +525,9 @@ func (t *turtle) computeEpochHeight(epoch types.EpochID) {
 	einfo := t.epoch(epoch)
 	heights := make([]uint64, 0, len(einfo.atxs))
 	for _, info := range einfo.atxs {
-		heights = append(heights, info.height)
+		if !info.malfeasant {
+			heights = append(heights, info.height)
+		}
 	}
 	einfo.height = getMedian(heights)
```
countvonzero, line 531: say node A published its first atx for epoch X+1 in epoch X, then published its second atx for epoch X in the 2nd layer of epoch X+1. the epoch weight will not be deducted as a result of makrMalfeasant?

tortoise/state.go
```diff
@@ -69,6 +70,10 @@ type (
 
 		// to efficiently find base and reference ballots
 		ballotRefs map[types.BallotID]*ballotInfo
+
+		// malnodes is a collection with all nodes that equivocated in history.
+		// each node id is 32 bytes. 100 000 of such nodes is only about ~3MB
+		malnodes map[types.NodeID]struct{}
```
countvonzero, line 76: this seems parallel with the lru cache in CachedDB, and we are spending 2X memory for the malicious IDs. is there any way to combine?

tortoise/tortoise.go
dshulyak: i am confused by this 'timestamps', they have no impact whatsoever. 
if smesher submitted two atxs for the same epoch, first is counted towards the threshold, second is not.

i discovered when i actually need to revert ballots weights, and probably will revert all malfeseant atxs within the window as well, but this is not in this change

tortoise/state.go
dshulyak: i understand that, but i don't see how to dedup.
i don't want to add dependency to that cache, and using tortoise to query for malfeasant ids will conflict with executing important things. would rather think that about memory if number of that ids will grow significantly

dshulyak: i can actually add another rwlock for identities, and then every module can query that, but better not in this change

countvonzero: FYI. the DBCache also caches honest identities (for hare).

tortoise/tortoise.go
countvonzero: the timing was based on how you calculated epoch weight. i recall the code calculated epoch weight at the first layer of every epoch.

maybe leave a TODO for reverting ballot/epoch weights?

countvonzero (approved):

tortoise/tortoise.go
dshulyak: >the timing was based on how you calculated epoch weight. i recall the code calculated epoch weight at the first layer of every epoch.

i doublechecked current code, it adds to epoch.weight only in once place, when onAtx is called

i will keep TODO on equivocation test, as it won't be reliable without reverting weight

dshulyak:
bors merge

activation/handler.go
```diff
@@ -408,14 +412,14 @@ func (h *Handler) storeAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 	}
 	if proof != nil {
 		h.cdb.CacheMalfeasanceProof(atx.SmesherID, proof)
+		h.tortoise.OnMalfeasance(atx.SmesherID)
```
countvonzero, line 415: i just realized that tortoise.OnMalfeasance() will be called twice. once here, and once when Publish() is called on the malfeasance proof https://github.com/spacemeshos/go-spacemesh/pull/4555/files#diff-32a06abd05b71236130e135e1db758587402f061b1fa6dd766248333fafcc293R442

dshulyak: it can be called any number of times, only the first one matters

PR #4558, opened 2023-06-21
[Merged by Bors] - move malfeasance proof validation outside db transaction

### LINKED ISSUE #4556: malfeasance proof: do not lock database while validating proofs
## Description
see discussion in https://github.com/spacemeshos/go-spacemesh/pull/4535

countvonzero:
## Motivation
Closes #4556

## Changes
- skip validating proof when its generated by the node
- add tests to validate malfeasance proofs generated by the node


activation/handler.go
```diff
@@ -359,50 +359,40 @@ func (h *Handler) storeAtx(ctx context.Context, atx *types.VerifiedActivationTx)
...
-				)
+			}
+			proof = &types.MalfeasanceProof{
+				Layer: atx.PublishEpoch.FirstLayer(),
+				Proof: types.Proof{
+					Type: types.MultipleATXs,
+					Data: &atxProof,
+				},
+			}
+			if err := malfeasance.ValidateAndSave(ctx, h.log, h.cdb, h.edVerifier, nil, &types.MalfeasanceGossip{
```
dshulyak: i think this was a mistake to introduce this function. signatures must be verified already when atxs are accepted. if it is not valid thats a programmer mistake. this method should simply store proof

dshulyak: i think we should try to make cross-module interfaces, i had to touch this now and it just a complete mess to work with

countvonzero: but the problem before was that the SmesherID was not saved and went unnoticed. it's about verifying the generated proof (and not just the signature...)

dshulyak: maybe there should be a test that generated proof is actually ok?

countvonzero:
bors try

activation/handler.go
countvonzero: ok. i have updated tests that validate the saved proof and restored code back to just saving the proof (without validating)

countvonzero:
bors try

activation/handler.go
```diff
@@ -385,10 +384,8 @@ func (h *Handler) storeAtx(ctx context.Context, atx *types.VerifiedActivationTx)
 						Data: &atxProof,
 					},
 				}
-				if err := malfeasance.ValidateAndSave(ctx, h.log, h.cdb, dbtx, h.edVerifier, nil, &types.MalfeasanceGossip{
-					MalfeasanceProof: *proof,
-				}); err != nil && !errors.Is(err, malfeasance.ErrKnownProof) {
-					return fmt.Errorf("adding malfeasance proof: %w", err)
+				if err := h.cdb.AddMalfeasanceProof(atx.SmesherID, proof, dbtx); err != nil {
```
dshulyak: this api call is still a risk. one thing is that it is confusing, as cache has access to its own db pointer. 
but also there is mutex on this codepath, if this mutex will be abused then every write will be delayed. this kind of bugs are extremely hard to find and debug later

i personally would avoid using that directly, even at inconvenience that proof won't be cached immediately

dshulyak (approved):

activation/handler.go
countvonzero: thanks for the suggestion. i agree.

the tricky thing is honest nodes' maliciousness is also cached (so we don't need to hit db every time a (hare) msg is received). so it's important to keep the db and cache in sync. 

i replaced `AddMalfeasanceProof` with `CacheMalfeasanceProof`.
now the code is changed to saving the proof directly to db, and then call `cdb.CacheMalfeasanceProof()` when the db transaction is successfully terminated.

countvonzero:
bors merge

PR #4564, opened 2023-06-21
Hare 3 initial implementation

piersy:
This PR is intended to give a look into the current state of the hare3 implementation (it is not functional).

There is a document [description.md](https://github.com/spacemeshos/go-spacemesh/blob/72e20642fea9939d4b0f7765f1430719843b692c/hare3/description.md) that provides a high level description of the protocol and its parts.

There is also the current state of the protocol implementation under `./hare3`.


hare3/gradecast.go
```diff
@@ -0,0 +1,126 @@
...
+	grade  uint8
+}
+
+// Gradecaster acts as a specialized value store, it ingests messages that can
+// be retrieved after 3 rounds. Retrieved messages will be graded either 1 or 2
+// based on the input grade and when they were ingested relative to the
+// msgRound. Messages received too late or with insufficient grades will not be
+// retrievable. Subsequent malicious messages can also render previously good
+// messages un-retrievable.
+type Gradecaster interface {
```
dshulyak, line 26: what is the point of it being an interface? does it simplify testing? i would rather not mock this stuff

hare3/graded_gossip.go
```diff
@@ -0,0 +1,97 @@
+package hare3
+
+import "github.com/spacemeshos/go-spacemesh/common/types"
+
+// GradedGossiper works as a filter, for the given message inputs it returns
+// one of 3 results indicating what action to take with the message it was
+// invoked for.
+type GradedGossiper interface {
```
dshulyak, line 8: what the point of it being an interface? does it simplify testing?

hare3/algorithm.go
```diff
@@ -0,0 +1,478 @@
...
+	// Each index "i" holds values considered valid up to round "i+1"
+	Vi [][]types.Hash20
+	// Each index "i" holds a map of sets of values from valid proposals
+	// indexed by their hash received in iteration "i"
+	Ti  []map[types.Hash20][]types.Hash20
+	Si  []types.Hash20
+	tgg ThresholdGradedGossiper
+	gc  Gradecaster
+	lc  LeaderChecker
+	mu  *sync.Mutex
```
dshulyak, line 215: please put it on top of stuff that is synchronized in final version

dshulyak (commented):
one thing that will be very important to redefine for initial integration are messages, they should be encoded as a scale enum where every specific message doesn't carry misleading junk (similar to protobuf oneof), and there should be no recursive types in them

there is an example with malfeasance proof where message basically encoded as `byte || scale type ` , and decoded then checks `byte` value before choosing how to decode.

hare3/algorithm.go
```diff
@@ -0,0 +1,478 @@
...
+	round  AbsRound
+	values []types.Hash20
+}
+
+// with int8 we have a max iterations of 17 before we overflow.
+// AbsRound (short for absolute round) represents a round of the hare protocol. The hare protocol progresses in iterations and there are 7 rounds per iteration, except for the first iteration which has a one off pre-round  the abs round starts at -1 and increments for each successive round.
+//
+// from the abs round both round and message type can be
+// the iteration and the round within that iteration can be derived.
+type AbsRound int8
```
dshulyak, line 489: this needs to be larger (max number of iterations is 40) and unsigned if it is going to be encoded with scale. note that we use compact encoding with scale, where integer is encoded into 1, 2, 4, 9 bytes depending on its size

hare3/graded_gossip_test.go
```diff
@@ -0,0 +1,41 @@
+package hare3
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGradedGossip(t *testing.T) {
+	gg := NewDefaultGradedGossiper()
```
dshulyak, line 10: please use table tests, or named subtests with extracted function like you mentioned previously

hare3/algorithm_test.go
```diff
@@ -0,0 +1,36 @@
...
+	values1 = []types.Hash20{val1}
+	values2 = []types.Hash20{val1, val2}
+	values3 = []types.Hash20{val1, val2, val3}
+
+	rPre = NewAbsRound(0, -1)
+	r0   = NewAbsRound(0, 0)
+	r1   = NewAbsRound(0, 1)
+	r2   = NewAbsRound(0, 2)
+	r3   = NewAbsRound(0, 3)
+	r4   = NewAbsRound(0, 4)
```
dshulyak, line 33: let's avoid global state for tests. if you want to reuse same setup for tests that can be accomplished with simple object

hare3/gradecast.go
piersy: Initially I started with an interface because I didn't have the implementation, I just added an implementation yesterday, and was considering whether the interface was still useful. I was planning to leave the interfaces till I'd finished all the testing in case a situation came up where we needed to mock something. I can't see that we would need to mock anything but it's no more work to remove the interfaces later rather than now.

hare3/graded_gossip.go
piersy: Same answer as above

hare3/algorithm.go
piersy: Do you mean the following ?

```
mu  *sync.Mutex
x   someType // synchronised by mu

y   someOtherType // not synchronised by mu
```
Is this a convention that we have, or in general a convention? I've never heard of this before.

hare3/algorithm.go
piersy: Thank you, I was wondering if this would be sufficient

hare3/graded_gossip_test.go
piersy: The reason I didn't do this here is I think its quite clear with the comment what is happening at each point, and when the test fails you will get a line number taking you to the exact point of the failure. I think the comments are too long to serve as good subtest names, so by adding subtests you'll end up with lots of tests with largely meaningless names which are adding boilerplate to the test. I don't really see what you'll gain from doing that?

hare3/algorithm_test.go
piersy: I not sure I would consider this global state, since these are not intended to be written, just read. Unfortunately due to the types involved they can't be made consts. Could you elaborate on your suggested sue of a simple object, I'm not sure exactly what you are suggesting?

hare3/graded_gossip_test.go
dshulyak: it allows to debug a single failed test instead of all. for example if can add breakpoint or print statement inside a function that is tested and run single test, and it will trigger only in that function

tbh i don't see downsides of doing subtests, even if you leave name field empty

hare3/algorithm.go
dshulyak: yes like that, it is a convention atleast in spacemesh codebase, and i definitely stumbled on it before, though i will need to find sources if you need them

another approach that i like is to nest sychronized fields after mutex, it is general practice in cockroachdb codebase

```
mu struct {
    sync.Mutex
    x someType
}
```

piersy: The nesting is nice, more explicit and less likely to get mangled by future updates

piersy: I'd be in favour of nesting, unless you'd rather stick with the line locality approach?

hare3/algorithm_test.go
dshulyak: really poor implementation of this pattern is in spacemesh api modules. in your case it is smaller but it is still not good.
for example if i need more than 3 ids, will i add more here? same for values1/2/3

i think better pattern is to write stateful tester class that will generate and manage collections of objects. so for example in this simple case it would be

```go
type hareTester struct {
    ids []int
    values []types.Hash20
}

func (t hareTester) getValues(positions ...int) []types.Hash32 {
    // and here it just selects values using positions, maybe even extends values is position is not found
}
```

hare3/algorithm.go
```diff
@@ -0,0 +1,502 @@
...
+	hardLocked []bool
+	// Each index "i" holds the locked value for iteration "i"
+	Li []*types.Hash20
+	// Each index "i" holds values considered valid up to round "i+1"
+	// TODO convert this to an object that accepts grades, so that it can be more closely mapped to the paper
+	Vi [][]types.Hash20
+	// Each index "i" holds a map of sets of values from valid proposals
+	// indexed by their hash received in iteration "i"
+	Ti  []map[types.Hash20][]types.Hash20
+	Si  []types.Hash20
```
dshulyak, line 212: could you help me understand why protocol needs to store history for all previous iterations? and not two at most

hare3/algorithm.go
```diff
@@ -0,0 +1,502 @@
...
+type Protocol struct {
+	// round encodes both iteration and round in a single value.
+	round AbsRound
+	// Each index "i" holds an indication of whether iteration "i" is hard locked
+	hardLocked []bool
+	// Each index "i" holds the locked value for iteration "i"
+	Li []*types.Hash20
+	// Each index "i" holds values considered valid up to round "i+1"
+	// TODO convert this to an object that accepts grades, so that it can be more closely mapped to the paper
+	Vi [][]types.Hash20
```
dshulyak, line 226: i is not an iteration here?

hare3/algorithm.go
```diff
@@ -0,0 +1,502 @@
...
+		if j > 0 {
+			// Check if values received from thresh gossip, we can only do this
+			// safely if we know j > 0, so this can't be done before that
+			// check.
+			values := p.tgg.RetrieveThresholdMessages(NewAbsRound(j-1, 5), 2)
+			_, set = matchInPreviousIterations(values, p.Ti, j)
+		}
+		// If we didn't get a set from threshold gossip then use our local candidate
+		if set == nil {
+			set = p.Vi[1]
```
dshulyak, line 355: could you name this as some constant?

dshulyak (commented):
types in hare3 is actually simpler than i thought, e.g. there is basically one type. it is still needs to be defined outside original hare package though

have you considered what is the hare3 bandwidth usage without gossip overhead? max 800 messages, 6 rounds, each message ~1KB (50 * 20), so - 5MB or 2.5MB depending if stop gossiping once threshold reached, am i right?

hare3/algorithm.go
```diff
@@ -0,0 +1,502 @@
...
+
+// with int8 we have a max iterations of 17 before we overflow.
+// AbsRound (short for absolute round) represents a round of the hare protocol. The hare protocol progresses in iterations and there are 7 rounds per iteration, except for the first iteration which has a one off pre-round  the abs round starts at -1 and increments for each successive round.
+//
+// from the abs round both round and message type can be
+// the iteration and the round within that iteration can be derived.
+type AbsRound int8
+
+// NewAbsRound constructs a new AbsRound from the given iteration (j) and round (r).
+func NewAbsRound(j, r int8) AbsRound {
```
dshulyak, line 492: i think this type doesn't have an advantage over type where Round and Iteration are stored separately

hare3/algorithm_test.go
piersy: Hi @dshulyak sorry for the delay in getting back to you here. Is your concern that we have global state or is it that we'll need to keep adding values?

dshulyak: the pattern with global state setup is a pain for next person that will work on this module

hare3/algorithm.go
piersy: That's what is specified in the whitepaper, at several points the protocol inspects the whole history of valid proposals to check whether the message it's considering is voting for a valid proposal. As to why the protocol inspects the whole history and not just the last 2 iterations I'd suggest asking Tal.

hare3/algorithm.go
piersy: It's more compact, than storing both separately, its also easier to compare since a given abs round only has one representation, whereas if we stored round and iteration separately there can be multiple representations for a single AbsRound (E.G. {iteration:1, round:0} {iteration:0, round:7} ).

What would be the benefit of storing round and iteration separately?

hare3/algorithm.go
piersy: Not sure what you mean here, also are you referring to p.Vi or set ?

The current names are not particularly intuitive, but they do follow the naming in the whitepaper which is helpful for cross checking the code against the whitepaper.

dshulyak: i am just confused about `1`

hare3/algorithm.go
dshulyak: but we will need to fit atleast 40 iterations and max 6 rounds, so 2 bytes either way, right?

we know when to increment iteration and when to increment round, so i thought that it is simpler to have struct with two fields and avoid additional arithmetic

> {iteration:0, round:7}

this is invalid, right?

piersy:
> types in hare3 is actually simpler than i thought, e.g. there is basically one type. it is still needs to be defined outside original hare package though

What types are you referring to here?

> have you considered what is the hare3 bandwidth usage without gossip overhead? max 800 messages, 6 rounds, each message ~1KB (50 * 20), so - 5MB or 2.5MB depending if stop gossiping once threshold reached, am i right?

I haven't given it consideration yet.

To clarify though there are 7 rounds (8 if you include the preround)
But messages are only sent for the preround, proposal, commit and notify rounds.
The protocol needs at least 2 iterations to complete, and so that makes 7 rounds in total  (4 in iteration 0 and 3 in iteration 1) where nodes send a message.

So we'd be looking at 800 * 7 messages being received by each node for an instance of hare 3.

hare3/algorithm.go
piersy: Oh, right. Yes I plan to wrap this in a type that makes it more clear, 1 here means the proposals that were received in round 1 or lower, which implies that the proposals have grade d-1 or higher (where d is the degree of the grading, and is 5). So 1 means proposals with grade 4 or 5.

hare3/algorithm_test.go
piersy: I don't see why that is.

IIUC what you are proposing is that we build some object that can then be used to generate test values for each test.

I see that as more of a pain because every test then needs to define the test values it needs and likely the naming will not be consistent between tests so it becomes harder to read all the tests.

So you'll end up with more code required for each test, it will take longer to read all the tests, because there won't be consistency between the variables used in each test and also you need to actually write the code for the value generator object.

Am I misunderstanding your suggestion, is there something I'm not considering?

hare3/algorithm.go
piersy: >but we will need to fit atleast 40 iterations and max 6 rounds, so 2 bytes either way, right?

Yes 2 bytes either way, by compact I meant we could fit more iterations in if we used a single value, several thousand as opposed to 256, it doesn't seem particularly important to have that though.

>we know when to increment iteration and when to increment round, so i thought that it is simpler to have struct with two fields and avoid additional arithmetic

If you look at the implementation of `NextRound` it just increments the round, it doesn't need to know about iterations or rounds, so we don't really know when to increment iteration or round explicitly in the current implementation.

I feel like if we represented round and iteration separately we'd just end up moving the existing logic to handle updating and calculating the absolute round, but I don't think it would get simpler.

>>{iteration:0, round:7}
>this is invalid, right?

You could consider it invalid or you could consider it as iteration 1 round 0, if you consider it as invalid then that's going to  require some code to validate updates, and that implies errors and that introduces extra complexity.

hare3/algorithm.go
dshulyak: i think it makes more sense if hardLocked and locked are single variables which are reset when softLock / hardLock rounds start

do you see a reason why Ti is not just a map[types.Hash20][]types.Hash20 ?when you run softLock / hardLock there should be only values from prev iterations

dshulyak:
> What types are you referring to here?

i meant messages that go over the wire. i misunderstood it yesterday, it looks like Prepare/Propose contains the slice, and Commit/Notify only a single hash, so my estimate was completely wrong

hare3/algorithm_test.go
dshulyak: so you are saying that you will be able to test everything with 3 ids and 3 different set of values? can you check API module, does testing there look ok to you?

hare3/runner/runner.go
```diff
@@ -0,0 +1,394 @@
...
+		// hare3/weakcoin for an explanation of weak)
+		if round == hare3.Preround.Round()+1 {
+			r.weakcoinChan <- WeakCoinResult{
+				Layer: r.layer,
+				Coin:  r.coinChooser.Choose(),
+			}
+		}
+		select {
+		// We await the beginning of the round, which is achieved by calling AwaitEndOfRound with (round - 1).
+		case <-r.clock.AwaitEndOfRound(uint32(round - 1)):
```
dshulyak: this is another part where i am confused, based on my understanding protocol needs to wait for 5 network delays (in iteration 0, and then 4). according to this clock it looks like it will wait for soft and hard llock too

- preround -> WAIT -> propose (only in first)
- propose -> WAIT -> WAIT -> commit
- commit -> WAIT -> notify
- notify -> WAIT -> TERMINATED ?

piersy: If you're referring to the note mentioning that round 1 (soft-lock) is not required in iteration 0,  I understood that as meaning it's not required to execute the code of round 1 as opposed to meaning that you go straight from round 0 to round 2 with only one WAIT.

The reason for my interpretation is that otherwise the specification of how valid values is calculated does not make sense, since they are intended to be calculated at the beginning of rounds 0, 1, 2 and 3.

dshulyak: no, this is not what i was referring to. this code will run waits in the wrong places

dshulyak: preround -> WAIT -> softlock -> WAIT -> hardlock -> WAIT propose

but it should be

preround -> WAIT -> softlock -> hardlock -> propose -> WAIT -> WAIT

dshulyak:
following changes are **required** to proceed with review, sorry for the list but it is easier for me this way

protocol changes:
- hare 3 should use new types, that encode preround and propose with a collection. commit, notify with a single hash
- hare 3 should output result for building block after receiving notify messages
- hard lock round should be skipped completely in the very first iteration. i was wrong about soft lock, so the skip is only for hard lock

other required changes:
- hare3 should use field based logger as any other module in the system. no strings formatting. also zap can be used directly, as this is where we are moving
- hare3 should not initialize logger based on the input from network
- duplicates from p2p layer are expected, and should not spam warnings
- no crash on database error, if you want exit protocol with error. what you did might be right approach so please create an issue to consider it later
- hardLock, softLock not a collection, map with values also not a collection, you can add iteration to each value in the map. i know that it is different from paper
- info logs should be limited to data when hare starts and terminates, right now it is spammy
- change should be limited only to hare3, if other changes are needed in other modules they should be in separate prs
- mutex should be a first field everywhere
- this code is not sufficiently tested, i suggest implementing generator for messages and testing that output matches expected after a sequence of messages was processed. i would like to have framework similar to tortoise (latest) but significantly simpler
- hare instance can be instantiated when layer starts, and 1st message is expected only after wakeup delta. with this assumption handling of early message is not required - if someone sends message before your layer started thats invalid. 

other asks:
- verbosity in the code can be lower, in some places code is simpler then the comments that describe it 
- code can be also more compact, i suggest limiting hare3 files to protocol.go that contains everything needed for clean state machine like you have. and hare.go that handles messages and routes them to the protocol
- interface for output hare result / coin using channels wasn't implemented
- metrics are needed

dshulyak:
we decided to go with alternative implementation plan

PR #4566, opened 2023-06-21
[Merged by Bors] - checkpoint: query the vrf nonce for the right epoch

countvonzero:
## Motivation
current code queries the vrf nonce using the wrong epoch.

the test failed after layers per epoch changed to 2, but passed after the code was corrected


fasmat (approved):

countvonzero:
bors merge

PR #4570, opened 2023-06-22
[Merged by Bors] - tortoise: dont notify updates caller if opinion is the same after restart

### LINKED ISSUE #4553: tortoise: when restoring state from disk don't re-notify mesh about valid blocks
when tortoise reads state from disk it will re-notify mesh about already emitted blocks, it is inneficient as it makes mesh to
check if all of those blocks are applied correctly. tortoise should instead check if emitted opinion matches stored, and don't send an update in such case

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4553


countvonzero (approved):

dshulyak:
bors merge

PR #4571, opened 2023-06-22
[Merged by Bors] - systets: setup nodes with random equivocation behavior

dshulyak:
second half of the cluster is expected to equivocate, as they are using non-unique private keys. 
test watches consensus hash and expects it to be the same across the cluster


dshulyak:
bors try

systest/tests/equivocation_test.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	for i := honest; i < len(keys); i += 2 {
+		_, priv, err := ed25519.GenerateKey(nil)
+		require.NoError(t, err)
+		keys[i] = priv
+		keys[i+1] = priv
+	}
+	cctx.Log.Infow("fraction of nodes will have keys set up for equivocations",
+		zap.Int("honest", honest),
+		zap.Int("equivocators", len(keys)-honest),
```
countvonzero, line 42: number of equivocators should be `(len(keys)-honest)/2`, as every 2 have the same key.

countvonzero, line 42: number of equivocators should be `(len(keys)-honest)/2`, as every 2 have the same key.

systest/tests/equivocation_test.go
```diff
@@ -0,0 +1,110 @@
...
+	for i := 0; i < honest; i++ {
+		_, priv, err := ed25519.GenerateKey(nil)
+		require.NoError(t, err)
+		keys[i] = priv
+	}
+	for i := honest; i < len(keys); i += 2 {
+		_, priv, err := ed25519.GenerateKey(nil)
+		require.NoError(t, err)
+		keys[i] = priv
+		keys[i+1] = priv
```
countvonzero, line 38: len(keys) needs to be a multiple of 4? otherwise this may be out-of-bound on the last one

systest/tests/equivocation_test.go
```diff
@@ -0,0 +1,110 @@
...
+			mu.Lock()
+			results[client.Name][num] = consensus
+			mu.Unlock()
+			return true, nil
+		})
+	}
+	eg.Wait()
+	reference := results[cl.Client(0).Name]
+	for i := 1; i < cl.Total(); i++ {
+		assert.Equal(t, reference, results[cl.Client(i).Name],
```
countvonzero, line 85: is this supposed to succeed for any number of nodes?
for example, if there are 12 nodes, 6 are honest and 3 are malicious. would the test still pass?

countvonzero (commented):
can you help me understand the change in tortoise? it looks more like simplification to me. or is it fixing something?

systest/tests/equivocation_test.go
dshulyak: i prefer this log to be all - honest. this is the number of total nodes that are expected to equivocate, not  number of keys

systest/tests/equivocation_test.go
dshulyak: what is the difference?

systest/tests/equivocation_test.go
dshulyak: why 4? it should be multiple of two, which is done above

dshulyak:
> can you help me understand the change in tortoise? it looks more like simplification to me. or is it fixing something?

it simplifies and ensures that consensus hash is always recomputed, i found inconistencies with this tests

systest/tests/equivocation_test.go
dshulyak: actually sorry the length of total nodes themself can be odd, i adjustedd honest in a  such way so that equivocators are always even

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
i need to do more work for equivocation to work correctly, please see other changes

systest/tests/equivocation_test.go
countvonzero: actually my question is stupid. nvm

dshulyak:
@countvonzero i simplified this test by removing partition and opened https://github.com/spacemeshos/go-spacemesh/issues/4620

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #4575, opened 2023-06-22
[Merged by Bors] - Fix unit tests on windows

### LINKED ISSUE #3624: Few tests from `txs` pkg fails on Windows

### LINKED ISSUE #3626: Test `TestSpacemeshApp_NodeService` from `cmd/node` pkg fails on Windows

### LINKED ISSUE #3628: Test `TestClock_AwaitRound` from `hare` pkg fails on Windows in CI
## Description
Test `TestClock_AwaitRound` from `hare` pkg fails on Windows in CI

## Steps to reproduce
This test fails only in CI. It doesn't fail when I run the test locally. To reproduce, execute unit tests on CI on Windows platform.

## Actual Behavior
```sh
=== FAIL: hare TestClock_AwaitRound (0.03s)
    clock_test.go:55: 
        	Error Trace:	D:\a\go-spacemesh\go-spacemesh\hare\clock_test.go:55
        	Error:      	too slow
        	Test:       	TestClock_AwaitRound
```

## Environment
 - OS: CI, Windows

### LINKED ISSUE #3630: Test `Test_HandleProposal_Success` from `beacon` pkg fails in Windows CI
## Description
Test `Test_HandleProposal_Success` from `beacon` pkg fails in Windows CI.

## Steps to reproduce
This test times out only in CI. It doesn't fail when I run the test locally. To reproduce, execute unit tests on CI on Windows platform.

## Actual Behavior
```sh
=== FAIL: beacon Test_HandleProposal_Success (0.00s)
    handlers_test.go:114: 
        	Error Trace:	D:\a\go-spacemesh\go-spacemesh\beacon\handlers_test.go:114
        	            				D:\a\go-spacemesh\go-spacemesh\beacon\handlers_test.go:236
        	Error:      	Not equal: 
        	            	expected: struct { valid beacon.proposalSet; potentiallyValid beacon.proposalSet }{valid:beacon.proposalSet{"?\xe2*F":struct {}{}}, potentiallyValid:beacon.proposalSet{"\xda\x04\xb7\x16":struct {}{}}}
        	            	actual  : struct { valid beacon.proposalSet; potentiallyValid beacon.proposalSet }{valid:beacon.proposalSet{"?\xe2*F":struct {}{}, "\xda\x04\xb7\x16":struct {}{}}, potentiallyValid:beacon.proposalSet(nil)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,10 +1,9 @@
        	            	 (struct { valid beacon.proposalSet; potentiallyValid beacon.proposalSet }) {
        	            	- valid: (beacon.proposalSet) (len=1) {
        	            	+ valid: (beacon.proposalSet) (len=2) {
        	            	   (string) (len=4) "?\xe2*F": (struct {}) {
        	            	+  },
        	            	+  (string) (len=4) "\xda\x04\xb7\x16": (struct {}) {
        	            	   }
        	            	  },
        	            	- potentiallyValid: (beacon.proposalSet) (len=1) {
        	            	-  (string) (len=4) "\xda\x04\xb7\x16": (struct {}) {
        	            	-  }
        	            	- }
        	            	+ potentiallyValid: (beacon.proposalSet) <nil>
        	            	 }
        	Test:       	Test_HandleProposal_Success
```

## Environment
 - OS: Windows, CI

### LINKED ISSUE #4544: TestBuilder_StartSmeshingAvoidsPanickingIfPrepareInitializerReturnsError

### LINKED ISSUE #4578: checkpoint: fails on windows
## Description
Checkpoint fails with various issues on Windows:

- sql files cannot be accessed
- files cannot be found
- etc.

fasmat:
## Motivation
There are still a few unit tests that fail on windows consistently. This PR address those.

Closes #3630 
Closes #3628
Closes #3626 
Closes #3624
Closes #4544 
Closes #4578 

## Changes
- Some timeouts have been increased to avoid tests failing due to the slower windows runner
- Ensure that temporary files are closed / removed before a test ends to avoid a test failing on windows because temporary files are still open.
- Re-enable all tests that were disabled for windows
- Speed up windows CI by disabling Windows defender
- Fix possible deadlock in `boostrap.Updater` when closing the service
- Free file lock when startup fails, also lock earlier and unlock later

## Test Plan
n/a

## TODO

## DevOps Notes


.github/workflows/ci.yml
```diff
@@ -183,21 +178,22 @@ jobs:
...
           sudo sysctl -w net.link.generic.system.hwcksum_rx=0
-      - name: disable TCP/UDP offload
+      - name: disable TCP/UDP offload - Windows
         if: ${{ matrix.os == 'windows-latest' }}
         run: |
           Disable-NetAdapterChecksumOffload -Name * -TcpIPv4 -UdpIPv4 -TcpIPv6 -UdpIPv6
+      - name: disable Windows Defender - Windows
+        if: ${{ matrix.os == 'windows-latest' }}
+        run: |
+          Set-MpPreference -DisableRealtimeMonitoring $true
```
poszu, line 196: What happened without this?

bootstrap/updater.go
```diff
@@ -181,13 +184,15 @@ func (u *Updater) Start(ctx context.Context) error {
...
-	defer u.mu.Unlock()
 	for _, ch := range u.subscribers {
 		close(ch)
 	}
-	_ = u.eg.Wait()
+	u.subscribers = nil
+	u.mu.Unlock()
+	u.stop()
+	return u.eg.Wait()
 }
```
poszu: How is it ensured that none of the channels in `u.subscribers` is written to after they are closed? Perhaps the order of closing the channels and `u.eg.Wait()` should be inverted?

bootstrap/updater_test.go
```diff
@@ -242,7 +240,7 @@ func TestStartClose(t *testing.T) {
 		}
 	}, time.Second, 100*time.Millisecond)
 	checkUpdate1(t, got)
-	cancel()
+	updater.Close()
```
poszu:
```suggestion
	require.NoError(t, updater.Close())
```

checkpoint/recovery_test.go
```diff
@@ -108,53 +110,51 @@ func TestReadCheckpointAndDie(t *testing.T) {
...
-				require.ErrorIs(t, err, os.ErrNotExist)
-			} else {
-				require.NoError(t, err)
-				require.True(t, bytes.Equal([]byte(checkpointdata), got))
-			}
+		uri := fmt.Sprintf("%s/snapshot-15", ts.URL)
+		dataDir := t.TempDir()
+		require.Panics(t, func() {
+			checkpoint.ReadCheckpointAndDie(ctx, logtest.New(t), dataDir, uri, types.LayerID(recoverLayer))
 		})
```
poszu: An idea for improving testability of "dying". Instead of calling `panic` internally, `ReadCheckpointAndDie()` could take an interface:

```go
type Killer interface {
   Kill()
}
```
In the application code, the implementation of `Killer` can call panic or cancel the app's main context. In UTs - it would be nicely mocked.

bootstrap/updater.go
```diff
@@ -154,6 +155,8 @@ func (u *Updater) Start(ctx context.Context) error {
 		return fmt.Errorf("data dir not set %s", u.cfg.DataDir)
 	}
 	u.once.Do(func() {
+		ctx, cancel := context.WithCancel(ctx)
+		u.stop = cancel
```
piersy: I don't think this is safe, Start and Close are called in different goroutines and so there's no guarantee that Close will see any change to u.stop, and also u.stop may not have been set by by the time that Close calls it.

node/node.go
```diff
@@ -451,6 +455,7 @@ func (app *App) Initialize() (err error) {
 			log.Int("hare_limit_iterations", app.Config.HARE.LimitIterations),
 			log.Duration("hare_round_duration", app.Config.HARE.RoundDuration))
 
+		app.fileLock.Unlock()
```
piersy: All these different places to unlock seems like it could trip up updaters of this code.

I think I'd be inclined to do something like this:
```go
	err = func() error {
		gpath := filepath.Join(app.Config.DataDir(), genesisFileName)
		var existing config.GenesisConfig
		if err := existing.LoadFromFile(gpath); err != nil {
			if !errors.Is(err, os.ErrNotExist) {
				return fmt.Errorf("failed to load genesis config at %s: %w", gpath, err)
			}
			if err := app.Config.Genesis.Validate(); err != nil {
				return err
			}
			if err := app.Config.Genesis.WriteToFile(gpath); err != nil {
				app.fileLock.Unlock()
				return fmt.Errorf("failed to write genesis config to %s: %w", gpath, err)
			}
		} else {
			diff := existing.Diff(app.Config.Genesis)
			if len(diff) > 0 {
				return fmt.Errorf("genesis config was updated after initializing a node, if you know that update is required delete config at %s.\ndiff:\n%s", gpath, diff)
			}
		}

		// tortoise wait zdist layers for hare to timeout for a layer. once hare timeout, tortoise will
		// vote against all blocks in that layer. so it's important to make sure zdist takes longer than
		// hare's max time duration to run consensus for a layer
		maxHareRoundsPerLayer := 1 + app.Config.HARE.LimitIterations*hare.RoundsPerIteration // pre-round + 4 rounds per iteration
		maxHareLayerDuration := app.Config.HARE.WakeupDelta + time.Duration(maxHareRoundsPerLayer)*app.Config.HARE.RoundDuration
		if app.Config.LayerDuration*time.Duration(app.Config.Tortoise.Zdist) <= maxHareLayerDuration {
			log.With().Error("incompatible params",
				log.Uint32("tortoise_zdist", app.Config.Tortoise.Zdist),
				log.Duration("layer_duration", app.Config.LayerDuration),
				log.Duration("hare_wakeup_delta", app.Config.HARE.WakeupDelta),
				log.Int("hare_limit_iterations", app.Config.HARE.LimitIterations),
				log.Duration("hare_round_duration", app.Config.HARE.RoundDuration))

			return errors.New("incompatible tortoise hare params")
		}
		return nil
	}()
	if err != nil {
		app.fileLock.Unlock()
		return err
	}

```

node/node.go
```diff
@@ -423,17 +423,21 @@ func (app *App) Initialize() (err error) {
 	var existing config.GenesisConfig
 	if err := existing.LoadFromFile(gpath); err != nil {
 		if !errors.Is(err, os.ErrNotExist) {
+			app.fileLock.Unlock()
```
poszu: Should we also unlock it in case the app terminates, for example on panic?

checkpoint/recovery.go
```diff
@@ -107,13 +107,14 @@ func copyToLocalFile(ctx context.Context, logger log.Log, fs afero.Fs, dataDir,
 	if err != nil {
 		return "", fmt.Errorf("%w: parse recovery URI %v", err, uri)
 	}
-	dst := RecoveryFilename(dataDir, filepath.Base(parsed.Path), restore)
+	dst := RecoveryFilename(dataDir, filepath.Base(parsed.String()), restore)
 	if parsed.Scheme == "file" {
-		_, err = fs.Stat(parsed.Path)
+		src := filepath.Join(parsed.Host, parsed.Path)
```
piersy: Why have you added the host here?

node/node_test.go
```diff
@@ -841,30 +830,39 @@ func TestGenesisConfig(t *testing.T) {
...
 	t.Run("no error if no diff", func(t *testing.T) {
 		app := New()
 		app.Config = getTestDefaultConfig(t)
 		app.Config.DataDirParent = t.TempDir()
 
 		require.NoError(t, app.Initialize())
+		t.Cleanup(func() { app.Cleanup(context.Background()) })
+
 		app.Cleanup(context.Background())
 		require.NoError(t, app.Initialize())
```
poszu: I had to think long why you call `t.Cleanup(func() { app.Cleanup(context.Background()) })` and immediately cleanup the app one line below. I suggest instead:
```suggestion
		require.NoError(t, app.Initialize())
		app.Cleanup(context.Background())

		require.NoError(t, app.Initialize())
		app.Cleanup(context.Background())
```

checkpoint/recovery.go
fasmat: On windows `parsed.Host` is the volume: `file://C:\Users\piersy` -> `parsed.Host` = `C:` and `parsed.Path` = `/Users/piersy`

fasmat: This is btw. already merged to the `v0.3.10` branch, it was part of the fix for checkpoint.

sql/database.go
```diff
@@ -226,9 +230,15 @@ func (db *Database) Exec(query string, encoder Encoder, decoder Decoder) (int, e
...
+	defer db.closeMux.Unlock()
+	if db.closed {
+		return nil
+	}
 	if err := db.pool.Close(); err != nil {
 		return fmt.Errorf("close pool %w", err)
 	}
+	db.closed = true
 	return nil
 }
```
poszu: How about using `sync.Once`?

config/genesis.go
```diff
@@ -59,20 +59,22 @@ func (g *GenesisConfig) Diff(other *GenesisConfig) string {
...
 func (g *GenesisConfig) WriteToFile(filename string) error {
-	buf, err := json.Marshal(g)
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_WRONLY, 0o644)
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(filename, buf, 0o644)
+	defer f.Close()
+	return json.NewEncoder(f).Encode(g)
 }
```
piersy, line 78: Why have you changed these?

bootstrap/updater.go
fasmat: That's true, but the underlying issue is a different one:

The context passed into start is valid for longer than the lifetime of `Start`, essentially giving 2 options to stop the Updater: calling `cancel` on the passed in context or calling `Close`, both of which DO NOT stop the service at the moment.

This is a hacky fix to make `Close` properly shut down the `Updater`. The correct fix is to not pass a context into `Start` at all.

config/genesis.go
fasmat: This speeds up tests loading and saving configuration. It is quite inefficient loading a whole file into memory just to then pass all the data to an Unmarshaller. Using Encoder and Decoder makes this more memory efficient and faster.

.github/workflows/ci.yml
fasmat: This disables the windows defender. Without it `go build` and `go test` are noticeably faster.

You can test it yourself: open the task manager and call `make build` on your windows machine. 90% of the CPU will be used by Windows defender, this disables it.

bootstrap/updater.go
fasmat: This would cause a deadlock.

node/node.go
fasmat: I don't think we can 😞

bootstrap/updater.go
piersy: The correct fix doesn't seem like such a big change to me.

* Pass the context at construction time instead of in Start.
* Make Close just call the cancel function of the context.
* Cleanup channels when returning from Start.

Seems like it should work

fasmat: > Pass the context at construction time

No. This is not how contexts are intended be used. If a function/method returns it should not use the passed in context at all any more.

fasmat: See also: https://go.dev/blog/context-and-structs

If the function/method that receives the context returns it should not hold on to a reference of the context any more, also not onto child contexts. I just tried to limit the scope of my change but `Start` and `Close` shouldn't use any context at all.

sql/database.go
fasmat: I thought about that to, but that doesn't work if `db.pool.Close` returns an error. In that case close cannot call `db.pool.Close` again / only one caller of `Close` receives the error.

```golang
// Close closes all pooled connections.
func (db *Database) Close() error {
	var err error
	db.closeOncer.Do(func() {
		err = db.pool.Close()
	})
	return err
}
```

node/node.go
poszu: Why not? It would be called in case of panic if it was deferred.

fasmat: If I `defer app.fileLock.Unlock()` then the lock will be released when `Initialize` returns which is not what we want.

bootstrap/updater.go
piersy: Thanks for the link, it seems that we are misusing contexts throughout the codebase!

Having checked there doesn't seem to be any value in the context passed to Start, since the only time its cancelled is after [run](https://github.com/spacemeshos/go-spacemesh/blob/fix-unittests-on-windows/node/node.go#L135) has returned and run will have already initiated the the node cleanup.

So a better approach would be:

* Remove the context
* Make Close clean up go-routines ... etc

bootstrap/updater.go
fasmat: `Updater::Subscribe` is the creator of the channel. The only one who can write to the channel is the `Updater` the caller of the method can only read from it.

`u.mu` prevents concurrent read and write access to the `u.subscribers` slice.

fasmat: I switched the order of waiting and clearing the slice

bootstrap/updater.go
fasmat: This is indeed the case but this PR is already quite large and I prefer multiple small PRs over a few large 🙂

fasmat: That being said I addressed this immediate concern by removing the passed in context from `Start` and use a channel to indicate closure over a `context.CancelFunc`.

checkpoint/recovery_test.go
piersy: What would Killer offer over require.Panics in this case?

sql/database.go
piersy: You could persist the close error on the struct, then that approach would be a little neater than the current approach.

fasmat: Yes in the case of returning the error that would work, but if `db.pool.Close()` was not successful another call to `Database.Close()` should try again in my opinion instead of returning the error from the last try.

node/node.go
poszu: Of course not, it would need some refactoring.

bootstrap/updater.go
piersy: Yes .. but! The purpose of this PR is to fix the unit tests and this change introduces a new problem that could cause tests to fail randomly. So I'd really be inclined to fix that as part of the PR. I not going to block it on this but lets see what @poszu thinks?

node/node.go
fasmat: good point, the current logic makes no sense (it locks too late and unlocks too early).

I created 2 new methods: `app.Lock()` and `app.Unlock`, the first of which is called earlier than `Initialize` and if successful is immediately followed by a `defer app.Unlock` so that unlocking is the last thing the node does (instead of the first thing when shutting down).

node/node.go
fasmat: The current logic makes no sense, I updated the code as described here: https://github.com/spacemeshos/go-spacemesh/pull/4575#discussion_r1239938660

bootstrap/updater.go
fasmat: As I said, I addressed this immediate concern, but I won't update the whole codebase to use context correctly 😉

checkpoint/recovery_test.go
fasmat: Normally I would agree @poszu, but the documentation to `ReadCheckpointAndDie` says that it is only used for testing and intentionally panics with `zap.Logger.Panic`. So I'm hesitant to invest too much time in making this cleaner...

Edit: this test is essentially testing that other tests behave correctly 😅

bootstrap/updater.go
piersy: Oh sorry! Great, I misread your comment :D

fasmat:
bors try

bootstrap/updater.go
```diff
@@ -143,51 +145,57 @@ func (u *Updater) Load(ctx context.Context) error {
...
 	if len(u.cfg.DataDir) == 0 {
 		return fmt.Errorf("data dir not set %s", u.cfg.DataDir)
 	}
 	u.once.Do(func() {
 		u.eg.Go(func() error {
+			ctx := log.WithNewSessionID(context.Background())
 			if err := u.Load(ctx); err != nil {
 				return err
 			}
-			wait := time.Duration(0)
```
countvonzero, line 161: this is intended to make it run immediately the first time on startup. this data is important for other component and should be read asap

checkpoint/recovery_test.go
countvonzero: this test ito make sure systest will run correctly.

hare/clock_test.go
```diff
@@ -28,14 +26,10 @@ func TestClock_AwaitWakeup(t *testing.T) {
...
-	if util.IsWindows() && util.IsCi() {
-		t.Skip("Skipping test in Windows on CI (https://github.com/spacemeshos/go-spacemesh/issues/3628)")
-	}
-
 	r := require.New(t)
 
-	tolerance := 5 * time.Millisecond
-	waitTime := 20 * time.Millisecond
+	tolerance := 20 * time.Millisecond
+	waitTime := 80 * time.Millisecond
```
countvonzero, line 14: do you know why this is needed?

fasmat: My assumption is because the windows runner is so much slower than the ubuntu and macos runner that 25ms is just not enough for the timer that is used to fire the signal.

`AwaitEndOfRound` uses `time.AfterFunc` which internally uses `time.Timer`. According to the documentation it "will send the current time on its channel after at least duration d". Essentially that means the timer will never be early, but there are no guarantees about delays.

So `<-time.After(1 * time.Second)` will fire earliest in 1 second, but could also fire later (on a busy / slow system).

The assertions in this test compare `AwaitEndOfRound` to `time.After` which has the same issue (as in it can never be early but it can be late). So the test becomes less stable the slower / busier the system is it is running on and to combat this I increased the thresholds a few milliseconds.

bootstrap/updater.go
fasmat: Updated

bootstrap/updater.go
```diff
@@ -143,51 +145,57 @@ func (u *Updater) Load(ctx context.Context) error {
...
 			}
 		})
 	})
 	return nil
 }
 
-func (u *Updater) Close() {
+func (u *Updater) Close() error {
+	select {
+	case <-u.stop: // prevent calling Close twice from a panic
```
countvonzero: i like this protection, but not sure about the comment. from a panic? how?

bootstrap/updater.go
```diff
@@ -126,7 +128,7 @@ func New(clock layerClock, opts ...Opt) *Updater {
 	return u
 }
 
-func (u *Updater) Subscribe() chan *VerifiedUpdate {
+func (u *Updater) Subscribe() <-chan *VerifiedUpdate {
 	u.mu.Lock()
 	defer u.mu.Unlock()
 	ch := make(chan *VerifiedUpdate, 10)
```
countvonzero, line 140: check if stop channel is close or not?

node/node.go
```diff
@@ -418,7 +422,23 @@ func (app *App) Initialize() (err error) {
...
+	}
+	if err := app.fileLock.Unlock(); err != nil {
+		log.With().Error("failed to unlock file",
+			log.String("path", app.fileLock.Path()),
+			log.Err(err),
+		)
+	}
+}
 
+// Initialize sets up an exit signal, logging and checks the clock, returns error if clock is not in sync.
```
countvonzero: the comment doesn't reflect the code.

node/node_test.go
```diff
@@ -547,7 +534,8 @@ func TestSpacemeshApp_NodeService(t *testing.T) {
 	}
 
 	// This stops the app
```
countvonzero: this comment is confusing after your change

sql/database.go
```diff
@@ -7,7 +7,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/go-llsqlite/llsqlite"
+	sqlite "github.com/go-llsqlite/llsqlite"
```
countvonzero: why?

syncer/syncer.go
```diff
@@ -219,6 +221,8 @@ func (s *Syncer) IsBeaconSynced(epoch types.EpochID) bool {
 func (s *Syncer) Start(ctx context.Context) {
```
countvonzero: you don't have to pass in ctx either?

countvonzero:
i find it a little confusing when reviewing the code. i am not sure which part fixes windows tests and which part are general code improvements.

bootstrap/updater.go
fasmat: Yeah, don't know what I was thinking when I wrote that comment... Updated it.

bootstrap/updater.go
fasmat: Updated to return an error if Updater has already been closed

node/node.go
fasmat: Updated comment.

sql/database.go
fasmat: `goimports` automatically adds this when I safe the file. I assume because the package in `llsqlite` is named `sqlite` and not `llsqlite`.

node/node_test.go
fasmat: Updated

syncer/syncer.go
fasmat: Removed ctx from `Start`, since it is a relatively small change and related to fixing some tests for windows.

I don't want to do this in too many places though, since this PR should only as much as is needed to get the windows build / unit tests running

fasmat:
> i find it a little confusing when reviewing the code. i am not sure which part fixes windows tests and which part are general code improvements.

I kept my changes to a minimum and only rewrote parts of the code that were necessary to change to get failing tests to pass on windows. There are more possible improvements in the code which I will create issues for and fix separately.

countvonzero (approved):

fasmat:
bors merge

PR #4586, opened 2023-06-24
[Merged by Bors] - hare: remove vrf nonce from hare eligibility

countvonzero:
## Motivation
while adding a equivocation systest, @dshulyak found inconsistent behavior in the network regarding known malicious identities.

```
if node submitted two different atxs with two different vrf nonce's, we will get the same one from database
so depending on which atx was received first - nodes will see all further eligibilities differently
```

Tal agreed to remove vrf nonce from hare eligibility for now and rely on hare 3 to deliver a more comprehensive defense.


countvonzero:
bors try

fasmat (approved):

countvonzero:
bors merge

PR #4597, opened 2023-06-26
[Merged by Bors] - tx: get rid of sleeps in tests

fasmat:
## Motivation
In #4575 sleeps were added to tx tests to ensure that they pass on windows. This fixes the code to not require the sleeps any more.

## Changes
TX are now sorted in the priority queue according to the following:

- largest fee
- if equal earliest receive time
- (new) if also equal transaction ID

on *nix systems it is near impossible for two transactions to be received on at the same time (within 2 ns). On windows however for two timestamps to be considered equal they can be up to 15 ms apart. Tis might cause the order of transactions to randomly change in the pool, that's why an additional check was added to determine the priority.

## TODO

## DevOps Notes


countvonzero (approved):

fasmat:
bors merge

PR #4600, opened 2023-06-27
[Merged by Bors] - beacon: speed up slow test

fasmat:
## Motivation
`TestBeacon_proposalPassesEligibilityThreshold` is a test that takes ~ 10 seconds to complete on my machine. The fastest GH runner (ubuntu) needs ~ 70 seconds for it. If we want to also execute it on slower runners (`linux/arm64` and `darwin/arm64`), we should speed it up so it doesn't time out.

## Changes
Reduce amount of proposals that are processed in test from 10k to 1k thereby speeding up the test by a factor of ~ 10.

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/beacon/beacon_test.go
+++ b/beacon/beacon_test.go
@@ -874,19 +874,19 @@ func TestBeacon_proposalPassesEligibilityThreshold(t *testing.T) {
 		wEarly, wOntime int
 	}{
 		{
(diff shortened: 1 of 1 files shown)


poszu (approved):

fasmat:
bors merge

PR #4602, opened 2023-06-27
[Merged by Bors] - txs: fix flaky test on windows

fasmat:
## Motivation
This fixes the flaky test `TestSelectProposalTXs` on windows.

## Changes
`SelectProposalTXs` is supposed to return a random collection of transactions from the mempool. To generate this random set it uses `time.Now().UnixNano()` as a seed. On windows this will return the same integer if called twice within less than ~ 15 ms.

The test is fast enough that between the first and second call to `SelectProposalTXs` no new timestamp will be available. 

The code was updated to not seed the PRNG on every call, but rather only once.

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/txs/conservative_state_test.go
+++ b/txs/conservative_state_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"math"
 	"math/rand"
+	"reflect"
 	"sort"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


poszu (approved):

fasmat:
There are 3 ways to approach a fix for this problem:

1. Fixing the test: retry until the set changes, on *nix this is basically immediately, on windows this is after ~ 15 ms
2. Fixing the implementation: seed the rng only once, this guarantees that any call generates a new random set, but it makes it predictable (because not secure random number generator)
3. Use secure random numbers: solves the issue but makes shuffling a lot slower.

Right now I went for 1. I'm not sure this is the best fix for the issue, but I'm open to opinions.

fasmat:
bors merge

Issue #4605, opened 2023-06-27
nightly build to run slow verification


countvonzero:
## Description
nightly build to verify code that cannot do on every merge

- fuzz test
- release candidate

fasmat:
We should create a release in every nightly build just to check if the release job is still working, since we had issues in the past were the release workflow broke and we only noticed when we wanted to create a new release.

Also worth considering:
- creating docker images and pushing them to docker hub for all supported platforms
- some basic checks on a created release / docker image -> start node, query status, shut down node
- tagging nightly releases (might be helpful in bisecting regressions in the future)

pigmej:
We should add that we should deploy the "normal" nightly network here and run some longer tests too that can be implied by the first points that Kimmy stated but thought that it makes sense to make it explicitly obvious :)

dshulyak:
foud out that infra for fuzzinng needs to be stateful. when golang fuzzer finds execution that increase code coverage it will add it to the initial corpus. so the larger the state the better utilization of the fuzzer, and betteer guarantee of finding regression

PR #4607, opened 2023-06-28
[Merged by Bors] - vm: fuzz parse

### LINKED ISSUE #4324: vm: fuzz vm validation/parsing by crafting smart inputs
submitting any random inputs will not get many results from fuzzing, as tx will very likely will be disqualified at the beginning (with invalid version or something else which is not very interesting). we want to enable deeper fuzzing.

the proposal is to implement fuzzer that uses valid fields progressively for every wallet/txtype combination, e.g in a following way:

fuzz [valid, invalid]
fuzz [valid, valid, invalid]

it can be implemented for every transaction/wallet pair.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4324

extend seed corpus for tx fuzzer. seed corpus helps fuzzer to start with inputs that enable to allow it to extend coverage.
for example adding existing principal with known template enables it to fuzz code paths after existence check, same for the method and some other fields.

in seed corpus there 3000 test cases, in around 2 hours of fuzzing on my computer fuzzer found 400 inputs that extend coverage. in order to make use of it we need to run it periodically and keep state between runs (https://github.com/spacemeshos/go-spacemesh/issues/4605)

### DIFF
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -49,6 +49,14 @@ func putEncoderBuffer(b *bytes.Buffer) {
 	encoderPool.Put(b)
 }
 
+func MustEncode(value Encodable) []byte {
+	buf, err := Encode(value)
+	if err != nil {
+		panic(err)
+	}
+	return buf
+}
+
 // Encode value to a byte buffer.
 func Encode(value Encodable) ([]byte, error) {
 	b := getEncoderBuffer()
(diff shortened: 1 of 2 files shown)


fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #4612, opened 2023-06-29
[Merged by Bors] - checkpoint: preserve own atx chain

### LINKED ISSUE #4624: checkpoint: correctly preserve miners inflight atxs and its dependencies
## Description
checkpoint recovery on testnet-05 exposed the following problem:
- if a miner is still initializing the storage before the checkpoint, their atx will be rejected later if the commitment atx it used was not part of the snapshot.
- when a miner's new atx is not part of the snapshot (either its waiting for poet proof or just published), the preserved atx will not be accepted by peers if
  - the number of atxs per miners were not enough to capture the dependency chain of a miner's own atx
  - if the miner's atx depends on a golden atx of a previous checkpoint snapshot

countvonzero:
## Motivation
Closes #4624 

## Changes
code
- if miner doesn't have any atx, check post data dir for metadata and preserve the commitment atx (and its deps) used for initialization
- if miner has an atx that's not part of the snapshot, preserve its chain of deps as far as it can
- if miner has an pending atx in nipost metadata, also preserve its positioning atx
- abort preserving atx if any of its dependency is part of the snapshot of previous checkpoint
- atx dependencies are recursively discovered. the order is important for the validation path
- once node start services after recovering from checkpoint, validate and save the preserved atxs and poet proofs.

systest TestCheckpoint
- publish snapshot that only contains half of the miners atxs. the other miners need to preserve the chain of its own atxs
- using checkpoint url for testing (instead of local snapshot files)

cleanup
- remove recovery from local file
- move checkpoint related data to common/types, as post init dependency is introduced to checkpoint package.

note:
i tried shutting down nodes gracefully in admin::Recover() RPC instead of panicking during TestCheckpoint. but kubernetes doesn't restart the nodes automatically, which means the nodes will not restart to do the recovery.

### DIFF
--- a/checkpoint/recovery.go
+++ b/checkpoint/recovery.go
@@ -13,6 +13,7 @@ import (
 	"github.com/spacemeshos/post/initialization"
 	"github.com/spf13/afero"
 
+	"github.com/spacemeshos/go-spacemesh/activation"
 	"github.com/spacemeshos/go-spacemesh/bootstrap"
 	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/spacemeshos/go-spacemesh/common/types"
@@ -381,18 +382,12 @@ func collectOwnAtxDeps(
 	var ref types.ATXID
 	var own bool
 	if atxid == types.EmptyATXID {
-		// check if any atx is being built
-		if cfg.PostDataDir == "" {
-			return nil, nil, nil
+		if m, _ := initialization.LoadMetadata(cfg.PostDataDir); m != nil {
+			ref = types.ATXID(types.BytesToHash(m.CommitmentAtxId))
+			logger.With().Debug("found commitment atx from metadata",
+				log.Stringer("commitment atx", ref),
+			)
 		}
-		m, err := initialization.LoadMetadata(cfg.PostDataDir)
-		if err != nil {
-			return nil, nil, fmt.Errorf("read post metdata %s: %w", cfg.PostDataDir, err)
-		}
-		ref = types.ATXID(types.BytesToHash(m.CommitmentAtxId))
-		logger.With().Debug("found commitment atx from metadata",
-			log.Stringer("commitment atx", ref),
-		)
 	} else {
 		ref = atxid
 		logger.With().Debug("found own atx",
@@ -401,15 +396,44 @@ func collectOwnAtxDeps(
 		own = true
 	}
 
+	// check for if miner is building any atx
+	nipostCh, _ := activation.LoadNipostChallenge(cfg.PostDataDir)
+	if ref == types.EmptyATXID && nipostCh == nil {
+		return nil, nil, nil
+	}
+
 	all := map[types.ATXID]struct{}{cfg.GoldenAtx: {}, types.EmptyATXID: {}}
 	for _, catx := range data.atxs {
 		all[catx.ID] = struct{}{}
 	}
-	logger.With().Info("trying to preserve atx and deps",
-		ref,
-		log.Bool("own", own),
+	var (
+		deps   []*types.VerifiedActivationTx
+		proofs []*types.PoetProofMessage
 	)
-	return collectDeps(db, cfg.GoldenAtx, ref, all)
+	if ref != types.EmptyATXID {
+		logger.With().Info("collecting atx and deps",
+			ref,
+			log.Bool("own", own),
+		)
+		deps, proofs, err = collectDeps(db, cfg.GoldenAtx, ref, all)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	if nipostCh != nil {
+		logger.With().Info("collecting pending atx and deps",
+			log.Object("nipost", nipostCh),
+		)
+		// any previous atx in nipost should already be captured earlier
+		// we only care about positioning atx here
+		deps2, proofs2, err := collectDeps(db, cfg.GoldenAtx, nipostCh.PositioningATX, all)
+		if err != nil {
+			return nil, nil, fmt.Errorf("deps from nipost positioning atx (%v): %w", nipostCh.PositioningATX, err)
+		}
+		deps = append(deps, deps2...)
+		proofs = append(proofs, proofs2...)
+	}
+	return deps, proofs, nil
 }
 
 func collectDeps(
@@ -446,7 +470,7 @@ func collect(
 	if atx.Golden() {
 		return fmt.Errorf("atx %v belong to previous snapshot. cannot be preserved", ref)
 	}
-	if atx.CommitmentATX != nil && *atx.CommitmentATX != goldenAtxID {
+	if atx.CommitmentATX != nil {
 		if err = collect(db, goldenAtxID, *atx.CommitmentATX, all, deps); err != nil {
 			return err
 		}
@@ -455,21 +479,15 @@ func collect(
 		if err != nil {
 			return fmt.Errorf("get commitment for ref atx %v: %w", ref, err)
 		}
-		if commitment != goldenAtxID {
-			if err = collect(db, goldenAtxID, commitment, all, deps); err != nil {
-				return err
-			}
-		}
-	}
-	if atx.PrevATXID != types.EmptyATXID {
-		if err = collect(db, goldenAtxID, atx.PrevATXID, all, deps); err != nil {
+		if err := collect(db, goldenAtxID, commitment, all, deps); err != nil {
 			return err
 		}
 	}
-	if atx.PositioningATX != goldenAtxID {
-		if err = collect(db, goldenAtxID, atx.PositioningATX, all, deps); err != nil {
-			return err
-		}
+	if err = collect(db, goldenAtxID, atx.PrevATXID, all, deps); err != nil {
+		return err
+	}
+	if err = collect(db, goldenAtxID, atx.PositioningATX, all, deps); err != nil {
+		return err
 	}
 	*deps = append(*deps, atx)
 	all[ref] = struct{}{}
--- a/checkpoint/recovery.go
+++ b/checkpoint/recovery.go
@@ -102,13 +102,13 @@ func Recover(
 	}
 	defer db.Close()
 	preserve, err := RecoverWithDb(ctx, logger, db, fs, cfg)
-	if err != nil {
-		if errors.Is(err, ErrCheckpointNotFound) {
-			logger.With().Info("no checkpoint file available. not recovering",
-				log.String("uri", cfg.Uri),
-			)
-			return nil, nil
-		}
+	switch {
+	case errors.Is(err, ErrCheckpointNotFound):
+		logger.With().Info("no checkpoint file available. not recovering",
+			log.String("uri", cfg.Uri),
+		)
+		return nil, nil
+	case err != nil:
 		return nil, err
 	}
 	return preserve, nil
--- a/checkpoint/recovery.go
+++ b/checkpoint/recovery.go
@@ -1,9 +1,7 @@
 package checkpoint
 
 import (
-	"container/list"
 	"context"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -162,27 +160,30 @@ func copyToLocalFile(ctx context.Context, logger log.Log, fs afero.Fs, dataDir, 
 	return dst, nil
 }
 
+type PreservedData struct {
+	Deps   []*types.VerifiedActivationTx
+	Proofs []*types.PoetProofMessage
+}
+
 func Recover(
 	ctx context.Context,
 	logger log.Log,
 	fs afero.Fs,
 	cfg *RecoverConfig,
-) error {
+) (*PreservedData, error) {
 	db, err := sql.Open("file:" + filepath.Join(cfg.DataDir, cfg.DbFile))
 	if err != nil {
-		return fmt.Errorf("open old database: %w", err)
+		return nil, fmt.Errorf("open old database: %w", err)
 	}
 	defer db.Close()
-	newdb, err := RecoverWithDb(ctx, logger, db, fs, cfg)
+	preserve, newdb, err := RecoverWithDb(ctx, logger, db, fs, cfg)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	if newdb != nil {
-		if err = newdb.Close(); err != nil {
-			return fmt.Errorf("close new db: %w", err)
-		}
+	if err = newdb.Close(); err != nil {
+		return nil, fmt.Errorf("close new db: %w", err)
 	}
-	return nil
+	return preserve, nil
 }
 
 func RecoverWithDb(
@@ -191,22 +192,22 @@ func RecoverWithDb(
 	db *sql.Database,
 	fs afero.Fs,
 	cfg *RecoverConfig,
-) (*sql.Database, error) {
+) (*PreservedData, *sql.Database, error) {
 	oldRestore, err := recovery.CheckpointInfo(db)
 	if err != nil {
-		return nil, fmt.Errorf("get last checkpoint: %w", err)
+		return nil, nil, fmt.Errorf("get last checkpoint: %w", err)
 	}
 	if oldRestore >= cfg.Restore {
 		types.SetEffectiveGenesis(oldRestore.Uint32() - 1)
-		return nil, nil
+		return nil, nil, nil
 	}
 	if err = fs.RemoveAll(filepath.Join(cfg.DataDir, bootstrap.DirName)); err != nil {
-		return nil, fmt.Errorf("remove old bootstrap data: %w", err)
+		return nil, nil, fmt.Errorf("remove old bootstrap data: %w", err)
 	}
 	logger.With().Info("recover from uri", log.String("uri", cfg.Uri))
 	cpfile, err := copyToLocalFile(ctx, logger, fs, cfg.DataDir, cfg.Uri, cfg.Restore)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	return recoverFromLocalFile(ctx, logger, db, fs, cfg, cpfile)
 }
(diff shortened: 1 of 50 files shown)


dshulyak:
what is the expected behavior when format has changed and user preserved own atx? will it crush its own node when it tries to encode atx? will it be disconnected from whole network because it will serve malformed data repeatedly?

checkpoint/recovery.go
```diff
@@ -216,65 +219,149 @@ type recoverydata struct {
...
-		return own, nil
+		// check if any atx is being built
+		if cfg.PostDataDir == "" {
+			return nil, nil, nil
+		}
+		m, err := initialization.LoadMetadata(cfg.PostDataDir)
+		if err != nil {
+			return nil, nil, fmt.Errorf("read post metdata %s: %w", cfg.PostDataDir, err)
+		}
+		seed = types.ATXID(types.BytesToHash(m.CommitmentAtxId))
```
dshulyak: maybe top, prev or ref. i think seed is not in the right context

checkpoint/recovery.go
```diff
@@ -216,65 +219,149 @@ type recoverydata struct {
...
+		logger.With().Debug("found commitment atx from metadata", seed)
+	} else {
+		seed = atxid
+		logger.With().Debug("found own atx", seed)
```
dshulyak: named fields (e.g when thing is passed to the logger without name) have to be deprecated eventually

checkpoint/recovery.go
```diff
@@ -216,65 +219,149 @@ type recoverydata struct {
...
 	}
+	var seed types.ATXID
+	var own bool
 	if atxid == types.EmptyATXID {
-		return own, nil
+		// check if any atx is being built
+		if cfg.PostDataDir == "" {
+			return nil, nil, nil
+		}
+		m, err := initialization.LoadMetadata(cfg.PostDataDir)
```
dshulyak: do you want to check for positioning atx in nipost challenge? i think it becomes even more error prone , but technically same problem exists there

checkpoint/recovery.go
```diff
@@ -347,30 +441,35 @@ func recoverFromLocalFile(
...
-			for _, atx := range own.preserve {
+		if len(deps) > 0 {
+			logger.With().Info("preserving deps for own atx",
+				log.Context(ctx),
+				log.Int("num atx", len(deps)),
+			)
+			for i, atx := range deps {
 				if err = atxs.Add(tx, atx); err != nil {
 					return fmt.Errorf("preserve atx %s: %w", atx.ID().String(), err)
 				}
```
dshulyak: i think this object (and poet proof) should pass validation before saving them to database, as it is locally configurable, nodes may save this and then advertise it to the whole cluster which will make them eventually disconnected

checkpoint/recovery.go
fasmat: It is part of the seed for post, but I would also prefer to just call it `commitment`.

checkpoint/recovery.go
countvonzero: yes. i have a local change that does exactly that... but testing in checkpoint test is trickier. still working on that

checkpoint/recovery.go
countvonzero: @fasmat seed is in the context of using this atx as a starting point for expanding dependencies. will change it to ref as dmitry suggested

checkpoint/recovery.go
dshulyak: i realized that under current constraints (e.g snapshotted atxs) this solution works only for the very first checkpoint, or more specifically for those that started initialization before first checkpoint. in all other cases snapshotted atx will be on the path

countvonzero: not necessarily. it will still work for
- own atx deps are traced to the current snapshot
- miner starts initializing in btwn snapshots and uses a commitment atx A not in old snapshot. this is likely as the algo tries to pick the max height. the only caveat is A cannot reference anything in the old snapshot. 

maybe when picking commitment/positioning atxes we should prioritize Golden() == false? but that should already happen by choosing max height.

dshulyak: i think anything that algo picks will have snapshotted atx on the path because of the reference to positioning

countvonzero: but prevAtx is a problem

dshulyak: yes that too. but positioning will be actually either the highest from the snapshot or one on top of that, depending on the timing

countvonzero: @dshulyak 
what do you think that we change the algo for a miner's first atx after snapshot to
- use golden atx for positioning/commitment atxes
- use empty atx for prev 

or alternatively we keep ALL historical snapshot atxes

should bring this up for discussion in research.

countvonzero:
> what is the expected behavior when format has changed and user preserved own atx? will it crush its own node when it tries to encode atx? will it be disconnected from whole network because it will serve malformed data repeatedly?

yes. i have a follow up change to validate own atxs before preserving it. if it fails, then it will give up.
ideally we will also just set the flag to false when we know a priori that format is incompatible

checkpoint/recovery.go
dshulyak: > use golden atx for positioning/commitment atxes

blocks in the next epoch will have lower height than the ballots from the previous one. it will definitely create more mess, not sure if it is worth it to analyze. for instance it will be one more consideration for concentration attack

> or alternatively we keep ALL historical snapshot atxes

what is the problem with configuring 4-10 epochs, as discussed yesterday? continuous post init should be supported and optimal from maximizing rewards. it is also fine to just keep all atx's metadata in the snapshot, to avoid such problems completelyy

countvonzero: > what is the problem with configuring 4-10 epochs, as discussed yesterday?

i don't see a problem. i assume @fasmat is working on that separately.

dshulyak: i don't understand the advantage of doing `preserve-own-atx` given it limited utility. especially if it requires making some adjustments in consensus

countvonzero: > i don't understand the advantage of doing preserve-own-atx given it limited utility. especially if it requires making some adjustments in consensus

are you suggesting to forgo support for inflight atxs (those already built but not in snapshot, and those still initializing)?
i am happy to delete the whole codepath actually, but the UX is going to be bad.

dshulyak: > those already built but not in snapshot

i forgot that we may be forced to do the snapshot close to cycle gap or shortly after, network should be literally dying to justify that. thanks, i remember previous chat now

so if 4-10 previous atxs are already in snapshot, then every smesher should succeed with `preserve-own-atx`?

countvonzero: > so if 4-10 previous atxs are already in snapshot, then every smesher should succeed with `preserve-own-atx`

yes, unless this miner start initializing 4 epochs ago. (assuming miners publish every epoch)
if it takes > 4 epochs to initialize their storage, we need to warn users about this possibility of being excluded from checkpoint snapshot.

checkpoint/recovery.go
countvonzero: added. thanks

countvonzero:
bors try

common/types/checkpoint.go
```diff
@@ -1,17 +1,17 @@
-package checkpoint
+package types
```
fasmat, line 1: We tend to put all type definitions in this package, but I'm not a fan of separating types from the implementation. I know they are also needed by systests, but the issue here is that the implementation defines the API, when it should be the other way round.

EDIT: I don't think this needs to be addressed in this PR but imo we should reconsider the way we organize our codebase.

checkpoint/util.go
```diff
@@ -104,7 +110,13 @@ func httpToLocalFile(ctx context.Context, resource *url.URL, fs afero.Fs, dst st
...
-		return fmt.Errorf("http get bootstrap file: %w", err)
+		if _, ok := err.(*url.Error); ok {
+			return ErrCheckpointNotFound
+		}
+		return fmt.Errorf("http get recovery file: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return ErrCheckpointNotFound
 	}
 	defer resp.Body.Close()
```
fasmat: `errors.As` is a more thorough check than casting the error. Also the body should be closed if `Do` didn't return an error, even if the status code is not `http.StatusOK`.

```golang
	resp, err := (&http.Client{}).Do(req)
	urlErr := &url.Error{}
	switch {
	case errors.As(err, &urlErr):
		return ErrCheckpointNotFound
	case err != nil:
		return fmt.Errorf("http get recovery file: %w", err)
	}
	defer resp.Body.Close()
	if resp.StatusCode != http.StatusNotFound {
		return ErrCheckpointNotFound
	}
```

node/bad_peer_test.go
```diff
@@ -56,11 +56,11 @@ func TestPeerDisconnectForMessageResultValidationReject(t *testing.T) {
 	})
 	g := errgroup.Group{}
 	g.Go(func() error {
-		return app1.Start(ctx)
+		return app1.Start(ctx, nil)
```
fasmat: Suggestion: use a functional option like `WithCheckpoint` to avoid having to add `nil` to every call of `Start`.

EDIT: or remove the parameter completely (see below)

node/node.go
```diff
@@ -150,15 +151,17 @@ func GetCommand() *cobra.Command {
 					return fmt.Errorf("could not retrieve identity: %w", err)
 				}
 
-				if err := app.LoadCheckpoint(ctx); err != nil {
+				preserve, err := app.LoadCheckpoint(ctx)
+				if err != nil {
 					return err
 				}
```
fasmat: Alternatively to the functional option, `app.LoadCheckpoint` could just set a field that `app.Start` can access later.

node/node.go
```diff
@@ -1394,10 +1367,57 @@ func (app *App) Start(ctx context.Context) error {
...
+					app.log.With().Error("failed to preserve atx after checkpoint",
+						log.Inline(vatx),
+						log.Err(err),
+					)
+				} else {
+					app.log.With().Info("preserved atx after checkpoint",
+						log.Inline(vatx),
+					)
+				}
+			}
```
fasmat: Deep nesting makes the code hard to follow imo, maybe `continue` would help?

```suggestion
		for _, vatx := range preserve.Deps {
			encoded, err := codec.Encode(vatx)
			if err != nil {
				app.log.With().Error("failed to encode atx after checkpoint",
					log.Inline(vatx),
					log.Err(err),
				)
				continue
			}
			if err := app.atxHandler.HandleAtxData(ctx, p2p.NoPeer, encoded); err != nil {
				app.log.With().Error("failed to preserve atx after checkpoint",
					log.Inline(vatx),
					log.Err(err),
				)
				continue
			}
			app.log.With().Info("preserved atx after checkpoint",
				log.Inline(vatx),
			)
		}
```

node/node.go
```diff
@@ -1394,10 +1367,57 @@ func (app *App) Start(ctx context.Context) error {
 			app.host.ID().String(), app.Config.Genesis.GenesisID().ShortString())
 	}
 
-	if err := app.startServices(ctx, appErr); err != nil {
+	if err := app.startServices(ctx); err != nil {
 		return err
 	}
 
+	// need post verifying service to start first
+	if preserve != nil {
```
fasmat: I feel like the whole block inside the if could be its own method `app.restore(preserve)`.

Also I think `chkptData` or similar could be a better name for `preserve`.

node/node.go
```diff
@@ -1394,10 +1367,57 @@ func (app *App) Start(ctx context.Context) error {
...
+						log.Err(err),
+					)
+				} else {
+					app.log.With().Info("preserved poet proof after checkpoint",
+						log.Stringer("atx id", preserve.Deps[i].ID()),
+						log.String("poet proof ref", preserve.Deps[i].GetPoetProofRef().ShortString()),
+					)
+				}
+			}
+		}
```
fasmat: Same as the other, `continue` helps to prevent deep nesting.

```suggestion
		for i, poetProof := range preserve.Proofs {
			encoded, err := codec.Encode(poetProof)
			if err != nil {
				app.log.With().Error("failed to encode poet proof after checkpoint",
					log.Stringer("atx id", preserve.Deps[i].ID()),
					log.Object("poet proof", poetProof),
					log.Err(err),
				)
				continue
			}
			if err := app.poetDb.ValidateAndStoreMsg(ctx, p2p.NoPeer, encoded); err != nil {
				app.log.With().Error("failed to preserve poet proof after checkpoint",
					log.Stringer("atx id", preserve.Deps[i].ID()),
					log.String("poet proof ref", preserve.Deps[i].GetPoetProofRef().ShortString()),
					log.Err(err),
				)
				continue
			}
			app.log.With().Info("preserved poet proof after checkpoint",
				log.Stringer("atx id", preserve.Deps[i].ID()),
				log.String("poet proof ref", preserve.Deps[i].GetPoetProofRef().ShortString()),
			)
		}
```

systest/tests/checkpoint_test.go
```diff
@@ -102,6 +103,43 @@ func TestCheckpoint(t *testing.T) {
...
+	endpoint := fmt.Sprintf("http://%s:%d", ip, 80)
+	updateUrl := fmt.Sprintf("%s/updateCheckpoint", endpoint)
+	tctx.Log.Infow("submit checkpoint data", "update url", updateUrl)
+	require.NoError(t, updateCheckpointServer(tctx, updateUrl, newSnapshot.Bytes()))
+
+	queryUrl := fmt.Sprintf("%s/checkpoint", endpoint)
+	tctx.Log.Debugw("query checkpoint data", "query url", queryUrl)
+	data, err := query(tctx, queryUrl)
+	require.NoError(t, err)
+	require.True(t, bytes.Equal(newSnapshot.Bytes(), data))
```
fasmat: This helps to see what the difference is should the assertion fail:

```suggestion
	require.Equal(t, newSnapshot.String(), string(data))
```

Also imo instead of serializing the expected data to json, it would be better to deserialize the received bytes to an object and compare those two. Arguably these two are the same: `{"field1":1,"field2":2}` and `{"field2":2,"field1":1}` but comparing them as strings will cause the assertion to fail. Same with nested objects and order of elements in arrays.

checkpoint/recovery_test.go
```diff
@@ -195,24 +145,30 @@ func TestRecoverFromHttp(t *testing.T) {
...
 			db := sql.InMemory()
-			newdb, err := checkpoint.RecoverWithDb(ctx, logtest.New(t), db, fs, cfg, types.NodeID{2, 3, 4}, tc.uri, types.LayerID(recoverLayer))
-			if len(tc.expErr) > 0 {
-				require.ErrorContains(t, err, tc.expErr)
+			preserve, err := checkpoint.RecoverWithDb(ctx, logtest.New(t), db, fs, cfg)
+			if tc.expErr != nil {
+				require.ErrorIs(t, err, tc.expErr)
 				return
 			}
 			require.NoError(t, err)
```
fasmat, line 165: No need to be careful. `errors.Is` can handle `nil`:

```suggestion
			require.ErrorIs(t, err, tc.expErr)
```

https://go.dev/play/p/yZfoyOh7n2g

checkpoint/recovery_test.go
```diff
@@ -362,172 +393,383 @@ func TestRecover_OwnAtxInCheckpoint(t *testing.T) {
...
+	validateAndPreserveData(t, olddb, vatxs[1:], proofs[1:])
+	// the proofs are not valid, but save them anyway for the purpose of testing
+	for i, vatx := range vatxs {
+		if i == 0 {
+			continue
+		}
+		encoded, err := codec.Encode(proofs[i])
+		require.NoError(t, err)
+		require.NoError(t, poets.Add(olddb, types.PoetProofRef(vatx.GetPoetProofRef()), encoded, proofs[i].PoetServiceID, proofs[i].RoundID))
+	}
```
fasmat: Ranging over proofs instead of vatxs simplifies the code a bit

```suggestion
	for i, proof := range proofs[1:] {
		encoded, err := codec.Encode(proof)
		require.NoError(t, err)
		require.NoError(t, poets.Add(olddb, types.PoetProofRef(vatx[i+1].GetPoetProofRef()), encoded, proof.PoetServiceID, proof.RoundID))
	}
```

checkpoint/recovery.go
```diff
@@ -156,30 +85,33 @@ func copyToLocalFile(ctx context.Context, logger log.Log, fs afero.Fs, dataDir,
...
+		if errors.Is(err, ErrCheckpointNotFound) {
+			logger.With().Info("no checkpoint file available. not recovering",
+				log.String("uri", cfg.Uri),
+			)
+			return nil, nil
 		}
+		return nil, err
 	}
-	return nil
+	return preserve, nil
```
fasmat: switching makes the error handling a bit easier to understand I think:

```golang
	preserve, err := RecoverWithDb(ctx, logger, db, fs, cfg)
	switch {
	case errors.Is(err, ErrCheckpointNotFound):
		logger.With().Info("no checkpoint file available. not recovering",
			log.String("uri", cfg.Uri),
		)
		return nil, nil
	case err != nil:
		return nil, err
	default:
		return preserve, nil
	}
```

node/node.go
countvonzero: but these are not checkpointData. at this point the checkpoint data is already saved in db.
here `preserve` means the data we want to preserve from the previous db, and is trying to save them to the new db via the gossip validation path.

systest/tests/checkpoint_test.go
countvonzero: in the systest i am trying hard to be black box (not knowing the data transpired btwn nodes.
here all i care is the data is the same as the one i just updated.

checkpoint/recovery_test.go
countvonzero: i did this intentionally because i am returning immediately when there is error,

checkpoint/recovery_test.go
countvonzero: but i think the suggested approach is more error-prone because it manipulating the index. here skipping 0 is very explicit. i will range over proofs instead but keeping continue.

checkpoint/recovery.go
countvonzero: for me if/else block is easier to read.....but ok.

node/node.go
countvonzero: done. thanks for the suggestion.

node/node.go
countvonzero: moved to a separate function and used continue.

countvonzero:
bors try

systest/tests/checkpoint_test.go
fasmat: The problem is that if that assertion fails it doesn't display way:

```
FAIL TestName
     Reason: Should be true
```

Which is especially hard to fix later should the test become flaky and a failing test doesn't say why it failed, for this I think using strings and `require.Equal` would be helpful over `require.True`.

The second (less important) issue is that both values can be different but arguably the test should still pass in my example above.

fasmat (approved):

systest/tests/checkpoint_test.go
countvonzero: the only reason i serialize data into json is so that i can manipulate the snapshot such that half of the miners need to preserve their own atxs after checkpoint. i'll deserialize the byte to match the content.

countvonzero:
bors merge

Issue #4599, opened 2023-06-27
Flaky tests on windows 


dshulyak:
=== Failed
=== FAIL: activation TestBuilder_RetryPublishActivationTx (10.02s)
    activation_test.go:1068: 
        	Error Trace:	D:/a/go-spacemesh/go-spacemesh/activation/activation_test.go:1068
        	Error:      	failed waiting for required number of tries to occur
        	Test:       	TestBuilder_RetryPublishActivationTx
    controller.go:137: missing call(s) to *activation.MocknipostBuilder.BuildNIPost(is anything, is anything) D:/a/go-spacemesh/go-spacemesh/activation/activation_test.go:1043
    controller.go:137: aborting test due to missing call(s)

=== FAIL: hare TestRecvDelayedDishonest (8.80s)
    consensus_test.go:111: Validity 1 failed: output does not contain the intersection of honest parties

=== FAIL: hare TestEquivocation (10.02s)
    consensus_test.go:111: Validity 1 failed: output does not contain the intersection of honest parties

https://github.com/spacemeshos/go-spacemesh/actions/runs/5387202693

dshulyak:
they are probably generally flaky. just failed for me on winddows

fasmat:
So far I have only seen `TestRecvDelayedDishonest` fail on windows:

https://github.com/spacemeshos/go-spacemesh/actions/runs/5424306542/jobs/9863491112

after this test is fixed we can probably add Windows to the `unittest` job and have it failing block merges.

countvonzero:
i found this one as well :(

=== Failed
=== FAIL: hare TestHare_onTick (10.03s)
    hare.go:337: Unexpected call to *mocks.MockRolacle.IsIdentityActiveOnConsensusView([context.Background.WithCancel.WithValue(type log.correlationIDType, val af86acdf-1a48-4386-a54e-0caec80fe276).WithValue(type log.correlationIDType, val <not Stringer>).WithValue(type log.correlationIDType, val 456e7430-ef47-44a4-8bb6-29e26384a443) 71a7692c195432e63fa4504fcc5c8c8b2a4d83936b42f84aa703d534fc17be05 9]) at D:/a/go-spacemesh/go-spacemesh/hare/hare.go:337 because: 
        expected call at D:/a/go-spacemesh/go-spacemesh/hare/hare_test.go:335 doesn't match the argument at index 2.
        Got: 9 (types.LayerID)
[...]

PR #4616, opened 2023-06-29
[Merged by Bors] - Fix slow and failing arm64 unit tests

fasmat:
## Motivation
On arm64 systems (mac M1 and linux) we had a regression where some tests timed out after 5 minutes.
This PR fixes the issues and promotes all runners except windows to be mandatory to pass for a successful build.

## Changes
- Arm64 runners are now configured to not use caching (seems to fail for some reason)
- All runners are now part of the mandatory `unittest` job, only windows is still in `unittests-slower` that is allowed to fail.
- Speed up some tests in the `activation` and `config/presets` packages by initializing fewer space units and using the correct proofing parameters
- Disable parallel testing in `FuzzParse` to avoid overwhelming the scheduler on windows

## Test Plan
- all test on all runners pass
- windows is allowed to fail for now

## TODO

## DevOps Notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,7 @@ jobs:
         run: sudo apt-get update -q && sudo apt-get install -qy ocl-icd-opencl-dev libpocl2
       - name: Override SDKROOT for macOS
         if: ${{ contains(matrix.os, 'macos') && runner.arch == 'arm64' }}
-        run: echo "SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk" >> "$GITHUB_ENV"
+        run: echo "SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk" >> $GITHUB_ENV
       - name: setup env
         run: make install
       - name: build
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -187,13 +208,15 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - [self-hosted, linux, arm64]
-          - macos-latest
-          - [self-hosted, macos, arm64]
           - windows-latest
```
poszu, line 211: You could remove the matrix completely and just use `runs-on: windows-latest`. Also, at this point all the special steps for ubuntu and mac can be removed.

fasmat: I left them there so if we want to disable runners again we just have to move the elements from one list to the other.

Eventually I want to move windows to the `unittests` job and completely remove `unittest-slower`

poszu: I understand the motivation, but I think this is a bad idea. If the stuff there is not actually used, there's a high risk it will get outdated pretty soon. IMO, we shouldn't just keep unused stuff around "just in case it is needed in the future".

poszu (approved):
I'd rather remove the unused stuff from unittest-slower, but that's not a blocker.

fasmat:
bors merge

.github/workflows/ci.yml
fasmat: Added it as info to https://github.com/spacemeshos/go-spacemesh/issues/4599 and assigned myself.

PR #4617, opened 2023-06-29
[Merged by Bors] - checkpoint: make number of atx stored per identity configurable

### LINKED ISSUE #4608: checkpoint: snapshot must include all atxs from a configurable number of recent epochs
## Description
The current checkpoint includes only the latest 2 ATXs of every smesher and their commitment ATXs. This is not sufficient to guarantee that a smesher that started initialising before the checkpoint and publishes their initial ATX after it has their commitment ATX available in the snapshot.

## Actual Behavior
Snapshot includes 2 most recent ATXs of every known identity and their commitment ATX

## Expected Behavior
Snapshot includes `n` most recent ATXs of every known identity and their commitment ATX.
`n` needs to be a configurable parameter that is set to 4 for mainnet and 10 for testnet.

fasmat:
## Motivation
Closes #4608

Do not merge before https://github.com/spacemeshos/api/pull/246

## Changes
- creating a checkpoint requires an additional parameter `numAtxs` that specifies how many atxs per smesher shall be persisted
- snapshot now contains the command that was used to create it
- Updated tests to handle new parameter correctly

## Test Plan
- new tests were added to check new behaviour of checkpoint

## TODO

## DevOps Notes

### DIFF
--- a/api/grpcserver/admin_service.go
+++ b/api/grpcserver/admin_service.go
@@ -24,17 +24,19 @@ const chunksize = 1024
 
 // AdminService exposes endpoints for node administration.
 type AdminService struct {
-	logger  log.Log
-	db      *sql.Database
-	dataDir string
+	logger           log.Log
+	db               *sql.Database
+	dataDir          string
+	checkpointEpochs int
 }
 
 // NewAdminService creates a new admin grpc service.
-func NewAdminService(db *sql.Database, dataDir string, lg log.Log) *AdminService {
+func NewAdminService(db *sql.Database, dataDir string, lg log.Log, numEpochs int) *AdminService {
 	return &AdminService{
-		logger:  lg,
-		db:      db,
-		dataDir: dataDir,
+		logger:           lg,
+		db:               db,
+		dataDir:          dataDir,
+		checkpointEpochs: numEpochs,
 	}
 }
 
@@ -48,7 +50,7 @@ func (a AdminService) CheckpointStream(req *pb.CheckpointStreamRequest, stream p
 	// - on the client side (default limit on the receiving end)
 	// - locally as the node already loads db query result in memory
 	snapshot := types.LayerID(req.SnapshotLayer)
-	err := checkpoint.Generate(stream.Context(), afero.NewOsFs(), a.db, a.dataDir, snapshot)
+	err := checkpoint.Generate(stream.Context(), afero.NewOsFs(), a.db, a.dataDir, snapshot, a.checkpointEpochs)
 	if err != nil {
 		return status.Errorf(codes.Internal, fmt.Sprintf("failed to create checkpoint: %s", err.Error()))
 	}
--- a/api/grpcserver/admin_service.go
+++ b/api/grpcserver/admin_service.go
@@ -50,7 +50,7 @@ func (a AdminService) CheckpointStream(req *pb.CheckpointStreamRequest, stream p
 	snapshot := types.LayerID(req.SnapshotLayer)
 	numAtxs := int(req.NumAtxs)
 	if numAtxs < 2 {
-		numAtxs = 2
+		numAtxs = 2 // default
 	}
 	err := checkpoint.Generate(stream.Context(), afero.NewOsFs(), a.db, a.dataDir, snapshot, numAtxs)
 	if err != nil {
--- a/api/grpcserver/admin_service.go
+++ b/api/grpcserver/admin_service.go
@@ -21,6 +21,7 @@ import (
 )
 
 const chunksize = 1024
+const defaultNumAtxs = 4
 
 // AdminService exposes endpoints for node administration.
 type AdminService struct {
@@ -49,8 +50,8 @@ func (a AdminService) CheckpointStream(req *pb.CheckpointStreamRequest, stream p
 	// - locally as the node already loads db query result in memory
 	snapshot := types.LayerID(req.SnapshotLayer)
 	numAtxs := int(req.NumAtxs)
-	if numAtxs < 2 {
-		numAtxs = 2 // default
+	if numAtxs < defaultNumAtxs {
+		numAtxs = defaultNumAtxs
 	}
 	err := checkpoint.Generate(stream.Context(), afero.NewOsFs(), a.db, a.dataDir, snapshot, numAtxs)
 	if err != nil {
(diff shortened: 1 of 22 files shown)


dshulyak:
but what was the reason to make it a node parameter, instead of specifying it in request?
it makes no sense to me, if we decide that we want 5 instead of 4 or 2 - then we need to update config and restart the node before we can generate snapshot

fasmat:
I assumed it should be a network parameter so every node keeps the same number of epochs in their snapshot, but I can also change it to be part of the request.

dshulyak:
> I assumed it should be a network parameter so every node keeps the same number of epochs in their snapshot, but I can also change it to be part of the request.

it would be better if it can be a request parameter, with a sane default.
right now it is not a network parameter, it is part of the centralized upgrades, we are generating a snapshot and distributing it

perhaps in future there can be an argument to make it decentralized, but hopefully we will get rid of snapshots completely within a year

fasmat:
> it would be better if it can be a request parameter, with a sane default.

Changed the number of ATXs preserved per identity to be a request parameter rather than a network parameter and made 2 the default (if requests specifies less than that or none).

api/grpcserver/admin_service.go
```diff
@@ -48,7 +48,11 @@ func (a AdminService) CheckpointStream(req *pb.CheckpointStreamRequest, stream p
 	// - on the client side (default limit on the receiving end)
 	// - locally as the node already loads db query result in memory
 	snapshot := types.LayerID(req.SnapshotLayer)
-	err := checkpoint.Generate(stream.Context(), afero.NewOsFs(), a.db, a.dataDir, snapshot)
+	numAtxs := int(req.NumAtxs)
+	if numAtxs < 2 {
+		numAtxs = 2 // default
+	}
+	err := checkpoint.Generate(stream.Context(), afero.NewOsFs(), a.db, a.dataDir, snapshot, numAtxs)
```
countvonzero, line 58: i think NumAtxs should be a node config, not an rpc argument. the goal is that the whole network should generate the same snapshot.

checkpoint/runner_test.go
```diff
@@ -23,14 +26,33 @@ func TestMain(m *testing.M) {
...
-var allAtxs = []*types.ActivationTx{
-	newatx(types.ATXID{12}, &types.ATXID{1}, 1, 0, 123, []byte("smesher1")),
-	newatx(types.ATXID{13}, nil, 2, 1, 0, []byte("smesher1")),
-	newatx(types.ATXID{14}, nil, 3, 2, 0, []byte("smesher1")),
-	newatx(types.ATXID{15}, nil, 4, 3, 0, []byte("smesher1")),
-	newatx(types.ATXID{23}, &types.ATXID{2}, 3, 0, 152, []byte("smesher2")),
-	newatx(types.ATXID{32}, &types.ATXID{3}, 2, 0, 211, []byte("smesher3")),
-	newatx(types.ATXID{34}, nil, 4, 1, 0, []byte("smesher3")),
+var allAtxs = map[types.NodeID][]*types.ActivationTx{
+	// smesher 1 has had 7 ATXs, one in each epoch from 1 to 7
```
countvonzero: :s/has had/has/g

checkpoint/runner_test.go
```diff
@@ -44,28 +66,59 @@ var allAccounts = []*types.Account{
...
-			},
 		},
 	}
+
+	if numAtxs < 2 {
+		require.Fail(t, "numEpochs must be at least 2")
+	}
+
+	atxs := make([]checkpoint.ShortAtx, 0, numAtxs*len(allAtxs))
+	for _, identity := range allAtxs {
```
countvonzero: the map is identity to atxs.... the variable name should be atxs instead of identity?

checkpoint/runner_test.go
```diff
@@ -44,28 +66,59 @@ var allAccounts = []*types.Account{
...
+	}
+
+	result.Data.Atxs = atxs
+
+	accounts := make(map[types.Address]*types.Account)
+	for _, account := range allAccounts {
+		if account.Layer <= snapshot {
+			if a, ok := accounts[account.Address]; !ok {
+				accounts[account.Address] = account
+			} else {
```
countvonzero: else if account.Layer > a.Layer

countvonzero:
also the PR title needs adjusting. thanks

api/grpcserver/admin_service.go
countvonzero: just read https://github.com/spacemeshos/go-spacemesh/pull/4617#issuecomment-1614551619

my argument is that node operators should be able to generate the snapshot themselves to verify against spacemesh published version. so nodes have to agree on the default number. 2 above is too small. for mainnet i thought we agreed on 4. 

the default constant should be defined as a constant as well.

countvonzero: @dshulyak 
considering the following:
- different network can have different sensible defaults (maybe 4 for mainnet, 10 for testnet)
- we want the node operators to be able to validate our snapshot easily

i feel it's simpler to put it as a node configuration.

however, for spacemesh operator, it would be simpler if it's an RPC argument. maybe we should have an RPC to update the default number of ATXs for snapshot instead, and leave the Checkpoint RPC unchanged?

countvonzero: i have a hard time rationalizing making it a rpc req argument.
two purposes for the Checkpoint RPC:
A. spacemesh operator to generate the snapshot
B. node operator to validate the snapshot generated by spacemesh

for A, there is currently no way for spacemesh operators to know that numAtxs==4 is not enough, and that we need to generate another snapshot where numAtxs is 5. also by the time we realize the number of atxs in the snapshot is not enough, the damage is already done. so allowing as argument doesn't make sense to me.

the key for me here is B and A need to generate the same data. so they should use the same default value.
i still think a config value makes more sense.

fasmat: The way I implemented it now:

- `numAtxs` is a second parameter for the request that specifies how many ATXs per identity should be included in the checkpoint
- the checkpoint now contains a "command" field that shows the command that was used to generate it.

countvonzero: thanks. that works for me.

countvonzero (approved):

fasmat:
bors merge

PR #4626, opened 2023-07-03
[Merged by Bors] - sync: fetch Commitment ATX as dependency of incoming ATX

### LINKED ISSUE #4615: sync: commitment ATX not fetched as a dependency of incoming ATX
## Description
An incoming ATX triggers fetching its dependencies (previous and positioning ATX). This should also trigger fetching the commitment referenced in the ATX if present.

## Additional Resources 
The relevant method to update is https://github.com/spacemeshos/go-spacemesh/blob/27ebba93d5d68b9a8314cba9b05296fe17b476b4/activation/handler.go#L561-L584

fasmat:
## Motivation
Closes #4615

## Changes
Fetch commitment ATX as dependency when handling incoming ATXs.

## Test Plan
- Added tests to check for correct handling of commitment ATX
- Updated existing tests to be executed in parallel

## TODO

## DevOps Notes

### DIFF
--- a/activation/handler.go
+++ b/activation/handler.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/spacemeshos/post/shared"
 	"golang.org/x/exp/maps"
-	"golang.org/x/exp/slices"
 
 	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/spacemeshos/go-spacemesh/common/types"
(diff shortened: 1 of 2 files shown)


activation/handler.go
```diff
@@ -557,21 +558,28 @@ func (h *Handler) handleGossipAtx(ctx context.Context, peer p2p.Peer, msg []byte
...
 		atxIDs = append(atxIDs, atx.PositioningATX)
 	}
 
 	if atx.PrevATXID != types.EmptyATXID {
-		logger.With().Debug("going to fetch prev atx", atx.PrevATXID, atx.ID())
-		if len(atxIDs) < 1 || atx.PrevATXID != atxIDs[0] {
+		logger.With().Debug("fetching prev atx", atx.PrevATXID, atx.ID())
+		if !slices.Contains(atxIDs, atx.PrevATXID) {
 			atxIDs = append(atxIDs, atx.PrevATXID)
 		}
```
poszu: NIT: a set could be used instead to automatically avoid duplicates. But then it would require conversion back to slice...

poszu (approved):

activation/handler_test.go
```diff
@@ -1078,50 +1078,105 @@ func BenchmarkGetAtxHeaderWithConcurrentProcessAtx(b *testing.B) {
...
+		coinbase := types.Address{2, 4, 5}
+		challenge := newChallenge(1, prevATX, goldenATXID, types.LayerID(22).GetEpoch(), nil)
+		nipost := newNIPostWithChallenge(t, types.HexToHash32("55555"), []byte("66666"))
+		atx := newAtx(t, sig, challenge, nipost, 2, coinbase)
+
+		atxHdlr.mockFetch.EXPECT().GetAtxs(gomock.Any(), []types.ATXID{atx.PrevATXID}).Return(nil)
+		require.NoError(t, atxHdlr.FetchAtxReferences(context.Background(), atx))
+	})
+
+	t.Run("valid prev ATX and empty pos ATX", func(t *testing.T) {
```
countvonzero, line 1115: empty pos ax is not valid

activation/handler_test.go
```diff
@@ -1078,50 +1078,105 @@ func BenchmarkGetAtxHeaderWithConcurrentProcessAtx(b *testing.B) {
...
+		challenge := newChallenge(1, types.EmptyATXID, posATX, types.LayerID(22).GetEpoch(), nil)
+		nipost := newNIPostWithChallenge(t, types.HexToHash32("55555"), []byte("66666"))
+		atx := newAtx(t, sig, challenge, nipost, 2, coinbase)
+		atx.CommitmentATX = &goldenATXID
+
+		atxHdlr.mockFetch.EXPECT().GetAtxs(gomock.Any(), []types.ATXID{atx.PositioningATX}).Return(nil)
+		require.NoError(t, atxHdlr.FetchAtxReferences(context.Background(), atx))
+	})
+
+	t.Run("empty prev ATX, empty pos ATX", func(t *testing.T) {
```
countvonzero, line 1156: ditto

activation/handler.go
fasmat: I changed it to a set since the order of the IDs doesn't matter and it is marginally faster.

fasmat:
bors merge

activation/handler_test.go
fasmat: That is correct, but when `FetchAtxReferences` is called we haven't checked yet for syntactically valid ATXs. So a node can send an ATX without a pos ATX, because it is only checked later if this is even valid or not.

activation/handler_test.go
fasmat: See above; I can just remove this two test cases if you prefer but the code also checks if the two fields reference `EmptyATXID` or not so I wouldn't cover all branches of the code then.

PR #4635, opened 2023-07-04
admin events

dshulyak:
api: https://github.com/spacemeshos/api/pull/247

launch standalone node
> ./build/go-spacemesh --preset=standalone --genesis-time=2023-07-04T9:05:00.000Z

watch events
> grpcurl -plaintext 0.0.0.0:10093 spacemesh.v1.AdminService.EventsStream

```
{
  "timestamp": "2023-07-04T08:55:00.355785009Z",
  "help": "Published proposal. Rewards will be received, once proposal is included into the block.",
  "proposal": {
    "layer": 150,
    "proposal": "i5iBp35b53G7OdTKsd9Y1tRc5HM="
  }
}
{
  "timestamp": "2023-07-04T08:55:01.025198892Z",
  "help": "Node started post execution for the challenge from poet.",
  "postStart": {
    "challenge": "mvy+3M6Ja2e8GWRsCP5Rqzuw+n8xIKtgnDGz296zKuU="
  }
}
```

admin events can be watched by subscribing to spacemesh.v1.AdminService.EventsStream. each event has 4 fields:
- timestamp
- failure. denotes that certain expected step has failed. we considered an alternative to provide a special failure events, but it would require maintaining a mapping on client side, and it is unclear if we need that. for now if any step has failure=true, user will be referred to logs or relevant documentation.
- help. this is initial approach to communicate what each events means to facilitate integration. later smapp can write better explanation and we will deprecate help.
- details. uses protobuf oneof, it enables smapp to reliably use those fields for custom representation and use this events in tests for better assertions. i considered encoding details in json, but it makes everything else worse but makes it a bit easier to add new events, which is probably not the right tradeoff

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -19,6 +19,7 @@ import (
 	"github.com/spacemeshos/go-spacemesh/codec"
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/datastore"
+	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/p2p/pubsub"
 	"github.com/spacemeshos/go-spacemesh/signing"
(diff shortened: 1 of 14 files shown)


dshulyak:
this wasn't registering new commits for some reason

PR #4638, opened 2023-07-04
[Merged by Bors] - Update CI tooling and fix complaints

fasmat:
## Motivation
The tooling used in go-spacemesh CI is outdated. This PR fixes the issue.

## Changes
- update tools: `gotestsum`, `golangci-lint` and `staticcheck`
- update configuration for new version of `golangci-lint`
- fix new complaints

## TODO

## DevOps Notes

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -590,7 +590,7 @@ linters-settings:
   staticcheck:
     # SAxxxx checks in https://staticcheck.io/docs/configuration/options/#checks
     # Default: ["*"]
-    checks: ["all", "-SA5001"]
+    checks: ["all"]
 
 issues:
   # List of regexps of issue texts to exclude.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,11 +9,11 @@ run:
 
   # Exit code when at least one issue was found.
   # Default: 1
-  issues-exit-code: 1
+  # issues-exit-code: 1
 
   # Include test files or not.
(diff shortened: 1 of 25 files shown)


.golangci.yml
```diff
@@ -270,114 +570,27 @@ linters-settings:
...
-  # https://golangci-lint.run/usage/linters/#disabled-by-default
-  # disable:
-    
-  # Run only fast linters from enabled linters set (first run won't be fast)
-  # Default: false
-  fast: false
+  staticcheck:
+    # SAxxxx checks in https://staticcheck.io/docs/configuration/options/#checks
+    # Default: ["*"]
+    checks: ["all", "-SA5001"]
```
piersy: -SA5001 seems new, why are we adding it?

piersy (approved):
One question, otherwise looks good!

.golangci.yml
fasmat: This check disallows calling `Close` without checking the returned error: https://staticcheck.dev/docs/checks/#SA5001

Seems like I misunderstood the check. I assumed that this is supposedly how we should handle a deferred close:

```golang
f, err := os.Open("")
defer func() {
    _ = f.Close()
}()
```

But actually it's because `Close` is deferred before `err` is checked. Will remove the exception and instead fix the code that causes the problems.

fasmat:
bors merge

Issue #4614, opened 2023-06-29
Provide a way to get commitmentATX for post via api call / cmd


pigmej:
For proper usage of `postcli` we would need to have a way to get commitmentATX that can be used for post, otherwise users of `postcli` are forced to start using smapp to get commitmentATX or use goldenATX always.

dima-stefantsov:
According to https://discord.com/channels/623195163510046732/820605182970036245/1125251881615228990 it is already possible with `go-spacemesh with start smeshing arg`, which generates `postdata_metadata.json` which has nodeid and atxid needed, all from cli.

dshulyak:
grpcurl -plaintext 127.0.0.1:9092 spacemesh.v1.ActivationService.Highest | jq -r '.atx.id.id' | base64 -d | xxd -p -c 64

PR #4639, opened 2023-07-04
[Merged by Bors] - activation api servce: add Highest RPC for max height atx

countvonzero:
## Motivation
part of #4614

depend on https://github.com/spacemeshos/api/pull/248

## Changes
add ActivationService::Highest() that returns the max height atx id in the node.

### DIFF
--- a/api/grpcserver/activation_service.go
+++ b/api/grpcserver/activation_service.go
@@ -13,11 +13,13 @@ import (
 )
 
 type activationService struct {
+	goldenAtx   types.ATXID
 	atxProvider atxProvider
 }
 
-func NewActivationService(atxProvider atxProvider) *activationService {
+func NewActivationService(atxProvider atxProvider, goldenAtx types.ATXID) *activationService {
 	return &activationService{
+		goldenAtx:   goldenAtx,
 		atxProvider: atxProvider,
 	}
 }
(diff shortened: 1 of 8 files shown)


api/grpcserver/activation_service.go
```diff
@@ -30,11 +32,24 @@ func (s *activationService) RegisterService(server *Server) {
 
 // Get implements v1.ActivationServiceServer.
 func (s *activationService) Get(ctx context.Context, request *pb.GetRequest) (*pb.GetResponse, error) {
-	if l := len(request.Id); l != types.ATXIDSize {
+	var atxId types.ATXID
```
dshulyak, line 35: i don't see why we should have such hacks

api/grpcserver/activation_service.go
```diff
@@ -30,11 +32,24 @@ func (s *activationService) RegisterService(server *Server) {
...
+	l := len(request.Id)
+	if l == 0 {
```
dshulyak: i don't see why we should have such hacks. it is very easy to extend protobufs with one more method

fasmat: `pb.GetRequest` is a generated type. I'm not sure its easy to add methods to it.

EDIT: nvm, I think I misunderstood what you meant - I assume you want an additional method besides `Get` that doesn't receive an ID? If yes, I agree 🙂

dshulyak: yes method for the service :) Highest() or Commitment()

countvonzero: thanks. added PR in api repository https://github.com/spacemeshos/api/pull/248
and updated the code here as well.

api/grpcserver/activation_service.go
```diff
@@ -53,3 +56,26 @@ func (s *activationService) Get(ctx context.Context, request *pb.GetRequest) (*p
...
+				Id: &pb.ActivationId{Id: s.goldenAtx.Bytes()},
+			},
+		}, nil
+	}
+	atx, err := s.atxProvider.GetFullAtx(highest)
+	if err != nil || atx == nil {
+		return nil, status.Error(codes.NotFound, fmt.Sprintf("atx id %v not found: %v", highest, err.Error()))
+	}
+
+	if highest != atx.ID() {
```
fasmat: I think this is probably not needed. If that check fails that would mean that `GetFullAtx(atxID)` returned an ATX that doesn't have `atxID` as its ID.

countvonzero: i don't fully understand. it's unlikely but better not return something that's wrong in API.
why do you think it's not needed?

dshulyak: i think as well that this check looks strange, it is a huge programmer mistake if it happens.
there should be a way not to write things like this

dshulyak (approved):

api/grpcserver/activation_service.go
fasmat: @countvonzero: why should a query to db for an ATX with a specific ID return an ATX with a different ID and no error? I hope we have tests in place that this cannot happen.

It's comparable to:

```golang
f, err := os.Open("test.txt")
if err != nil {
    return err
}
if f.Name() != "test.txt" {
    return fmt.Errorf("wrong file name")
}
```

I wouldn't think the second `if` is necessary.

countvonzero: ok. in that case do you object that i remove the same check in Get()?

fasmat: I wasn't aware that there was such a check, but yes, imo it can go there as well 🙂

countvonzero:
bors merge

PR #4641, opened 2023-07-05
[Merged by Bors] - node: admin events

dshulyak:
api: https://github.com/spacemeshos/api/pull/247

launch standalone node
> ./build/go-spacemesh --preset=standalone --genesis-time=2023-07-04T9:05:00.000Z

watch events
> grpcurl -plaintext 0.0.0.0:10093 spacemesh.v1.AdminService.EventsStream

```
{
  "timestamp": "2023-07-04T08:55:00.355785009Z",
  "help": "Published proposal. Rewards will be received, once proposal is included into the block.",
  "proposal": {
    "layer": 150,
    "proposal": "i5iBp35b53G7OdTKsd9Y1tRc5HM="
  }
}
{
  "timestamp": "2023-07-04T08:55:01.025198892Z",
  "help": "Node started post execution for the challenge from poet.",
  "postStart": {
    "challenge": "mvy+3M6Ja2e8GWRsCP5Rqzuw+n8xIKtgnDGz296zKuU="
  }
}
```

admin events can be watched by subscribing to spacemesh.v1.AdminService.EventsStream. each event has 4 fields:
- timestamp
- failure. denotes that certain expected step has failed. we considered an alternative to provide a special failure events, but it would require maintaining a mapping on client side, and it is unclear if we need that. for now if any step has failure=true, user will be referred to logs or relevant documentation.
- help. this is initial approach to communicate what each events means to facilitate integration. later smapp can write better explanation and we will deprecate help.
- details. uses protobuf oneof, it enables smapp to reliably use those fields for custom representation and use this events in tests for better assertions. i considered encoding details in json, but it makes everything else worse but makes it a bit easier to add new events, which is probably not the right tradeoff


dshulyak:
> --- PASS: TestAdminEvents (66.84s)

windows failing because of something else

node/node_test.go
```diff
@@ -906,6 +906,69 @@ func TestFlock(t *testing.T) {
...
+	cfg.DataDirParent = t.TempDir()
+	cfg.FileLock = filepath.Join(cfg.DataDirParent, "LOCK")
+	cfg.SMESHING.Opts.DataDir = t.TempDir()
+	cfg.Genesis.GenesisTime = time.Now().Add(5 * time.Second).Format(time.RFC3339)
+	types.SetLayersPerEpoch(cfg.LayersPerEpoch)
+
+	app := New(WithConfig(&cfg), WithLog(logtest.New(t)))
+	signer, err := app.LoadOrCreateEdSigner()
+	require.NoError(t, err)
+	app.edSgn = signer // it should not be like this
```
fasmat: is there an issue to fix it?

node/node_test.go
```diff
@@ -906,6 +906,69 @@ func TestFlock(t *testing.T) {
...
+	require.NoError(t, app.Initialize())
+	ctx, cancel := context.WithCancel(context.Background())
+	var eg errgroup.Group
+	eg.Go(func() error {
+		if err := app.Start(ctx); err != nil {
+			return err
+		}
+		app.Cleanup(context.Background())
+		app.eg.Wait() // it should not be like this
```
fasmat: same as above (should probably be the same issue)

events/events.go
```diff
@@ -0,0 +1,205 @@
...
+				Proposal: proposal[:],
+			},
+		},
+	)
+}
+
+func emitUserEvent(help string, failure bool, details pb.IsEventDetails) {
+	mu.RLock()
+	defer mu.RUnlock()
+	if reporter != nil {
```
fasmat, line 195: I think inversing the if and returning early makes the code clearer.

Any reason that `emitUserEvent` is a function instead of a method on `reporter`? It's a bit odd that the reporter has to be initialized with `InitializeReporter` and every function in the package needs to check if it is `nil` or not.

If it was a service and we don't want events to be reported we can just pass a `NoOpEventReporter` that implements the same interface as the actual reporter.

Probably not for this PR but technical debt that we should clean up in the future.

activation/nipost.go
```diff
@@ -206,11 +209,13 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 	if nb.state.NIPost.Post == nil {
 		nb.log.With().Info("starting post execution", log.Binary("challenge", nb.state.PoetProofRef[:]))
 		startTime := time.Now()
+		events.EmitPostStart(nb.state.PoetProofRef[:])
```
fasmat, line 212: Any reason why this event (and the one below) take `[]byte` as parameter instead of the actual type?

fasmat (approved):

api/grpcserver/admin_service.go
```diff
@@ -101,3 +102,28 @@ func (a AdminService) Recover(ctx context.Context, req *pb.RecoverRequest) (*emp
...
+	// send empty header after subscribing to the channel.
+	// this is optional but allows subscriber to wait until stream is fully initialized.
+	if err := stream.SendHeader(metadata.MD{}); err != nil {
+		return status.Errorf(codes.Unavailable, "can't send header")
+	}
+	for {
+		select {
+		case <-stream.Context().Done():
+			return nil
+		case <-sub.Full():
```
piersy, line 121: It would be nice to send all the buffered events at this point. Otherwise I think they'll just get lost

events/events.go
piersy: I'm also in favour of passing the reporter to whatever services need to report. I made a ticket for this [here](https://github.com/spacemeshos/go-spacemesh/issues/4651)

events/events.go
```diff
@@ -0,0 +1,205 @@
...
+func EmitPoetWait(current, publish types.EpochID, wait time.Duration) {
+	const help = "Node needs to wait for poet registration window in current epoch to open. " +
+		"Once opened it will submit challenge and wait till poet round ends in publish epoch."
+	emitUserEvent(
+		help,
+		false,
+		&pb.Event_PoetWait{PoetWait: &pb.EventPoetWait{
+			Current: current.Uint32(),
+			Publish: publish.Uint32(),
+			Wait:    durationpb.New(wait),
```
piersy, line 66: I think putting the specific time here (and for other events with the Wait fields) would be better for users. It will, I think be clearer and easier to remember. Also if the UI were to try to convert this duration into the actual time to wait till, it will be off, due to the delay in delivering the event. But the other way around you can always accurately generate the time left from an exact timestamp.

events/events.go
dshulyak: yes it should not be global, thats legacy.. which i am extending ofcourse, but can't fix it right now

events/events.go
dshulyak: there is a timestamp for every event, i think what you suggest can be accomplished using that timestamp on ui if needed

piersy: I still think an exact time here would be preferable since we do calculate exact times for events (I.E. a layer has an exact time that should be identical for all users) using the timestamp you won't actually be able to calculate an exact time, the time will be off by some amount. Probably a minuscule amount but even so I think the logs would be clearer and, debugging issues with users could be significantly simpler if we could rely on them logging exact times for events, because you could filter or grep by exact time.

activation/nipost.go
dshulyak: tried to change that, but zero challenge is defined as a slice, so i would rather leave it as it is

events/events.go
dshulyak: not all of this events are determined by layer number. both poet events are time based, e.g we have a deadline based on some other logic. for atx this is a wait time till the start of the target epoch, after that user will receive different event

as you said time won't be accurate, but the inaccuracy will be so small that it doesn't matter for human to make any decisions. tbh i am not even sure that smapp should print this wait time, i just added more stuff for @brusherru to consider what to use

so for me the diff between timestamp and duration is so irrelevant that i can't justify going back and change things

dshulyak:
bors merge

events/events.go
piersy: My point was more regarding making debugging simpler, by having consistent logs for all users. I don't think it will make a difference when it comes to taking decisions. I do however think that having the time to wait is very important for users, since these times could be huge ( on the order of weeks ) we will need to manage users expectations.

dshulyak:
bors cancel

dshulyak:
bors merge

PR #4647, opened 2023-07-05
[Merged by Bors] - Fix possible nil ptr dereference in HTTP retry check

### LINKED ISSUE #4645: Possible crash in poet client CheckRetry callback
## Description
The `checkRetry()` callback in poet client might dereference a nil pointer 

## Steps to reproduce
Unknown

## Actual Behavior
crash

## Expected Behavior
no crash, don't retry

## Environment
?

## Additional Resources 
https://github.com/spacemeshos/go-spacemesh/blob/b907b50279fc046b8fa53c7da578bb30024ef930/activation/poet.go#L49-L58

image

poszu:
## Motivation
Closes #4645 

## Changes
Checking for an error before dereferencing `resp`

## Test Plan
Added unit tests covering the whole `checkRetry()`.

### DIFF
--- a/activation/poet.go
+++ b/activation/poet.go
@@ -50,7 +50,9 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
 	if retry, err := retryablehttp.DefaultRetryPolicy(ctx, resp, err); retry || err != nil {
 		return retry, err
 	}
-
+	if err != nil {
+		return false, nil
+	}
 	if resp.StatusCode == http.StatusNotFound {
 		return true, nil
 	}
--- a/activation/poet.go
+++ b/activation/poet.go
@@ -47,16 +47,10 @@ func defaultPoetClientFunc(address string, cfg PoetConfig) (PoetProvingServiceCl
 }
 
 func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, error) {
-	if retry, err := retryablehttp.DefaultRetryPolicy(ctx, resp, err); retry || err != nil {
-		return retry, err
-	}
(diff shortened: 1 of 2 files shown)


activation/poet.go
```diff
@@ -50,7 +50,9 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
 	if retry, err := retryablehttp.DefaultRetryPolicy(ctx, resp, err); retry || err != nil {
```
dshulyak: this method will not fail with nil resp? does it make sense to exit immediately with false is resp is nil, just for additionnal defense?

dshulyak (approved):

activation/poet.go
fasmat: afaik `resp` is only `nil` if `err` is not, at least that's how `net/http` deals with failed requests.

That being said, we only want to overwrite the behaviour in the case of 404? Could we then write `checkRetry` like this:

```golang
func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, error) {
	if resp != nil && resp.StatusCode == http.StatusNotFound {
		return true, err
	}
	return retryablehttp.DefaultRetryPolicy(ctx, resp, err)
}
```

poszu: @fasmat, much nicer, thx. Updated the code.

poszu:
bors merge

poszu:
`TestCheckpoint` timed out. There's nothing obvious in the logs - retrying

bors merge

poszu:
Weird timeout in macos unittests...

bors merge

PR #4657, opened 2023-07-06
[Merged by Bors] - Prevent panicking when fetching without any peers

piersy:
## Motivation
There was a debug panic that was not removed from #4374, and also `randomPeer()` would fatally exit if there were no peers.
We don't actually want to panic, when trying to service a fetch request and there are no peers, so now we check to see if we have no peers and simply return, the fetch operation is performed on a ticker and so will be re-executed after some period.

## DevOps Notes

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -457,6 +457,7 @@ func (f *Fetch) organizeRequests(requests []RequestMessage) map[p2p.Peer][][]Req
 	peer2requests := make(map[p2p.Peer][]RequestMessage)
 	peers := f.host.GetPeers()
 	if len(peers) == 0 {
+		log.Info("cannot send fetch: no peers found")
 		// in loop() we will try again after the batchTimeout
 		return nil
 	}
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -459,9 +459,6 @@ func (f *Fetch) organizeRequests(requests []RequestMessage) map[p2p.Peer][][]Req
 	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
 	peer2requests := make(map[p2p.Peer][]RequestMessage)
 	peers := f.host.GetPeers()
-	if len(peers) == 0 {
-		panic("no peers")
-	}
 
 	for _, req := range requests {
 		p, exists := f.hashToPeers.GetRandom(req.Hash, req.Hint, rng)
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -98,9 +98,6 @@ func DefaultConfig() Config {
 
 // randomPeer returns a random peer from current peer list.
 func randomPeer(peers []p2p.Peer) p2p.Peer {
-	if len(peers) == 0 {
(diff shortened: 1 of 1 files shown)


fetch/fetch.go
```diff
@@ -459,9 +459,6 @@ func (f *Fetch) organizeRequests(requests []RequestMessage) map[p2p.Peer][][]Req
 	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
 	peer2requests := make(map[p2p.Peer][]RequestMessage)
 	peers := f.host.GetPeers()
-	if len(peers) == 0 {
-		panic("no peers")
```
countvonzero, line 463: my concern is that there must have been a codepath that caused you to add this panic when you worked on the PR.
if we simply removed here, then it is simply kicking the can down the road. can you look deeper to see why you added this in the first place?

piersy: Let me dig in

piersy: I added that when I added a test of connection dropping, I just wanted to be sure that the connections I was establishing in the tests were persisting, before they were dropped forcibly.

piersy: So its just debug and can go.

piersy: However there is this, which is called just below. Looks like we should change that as well

 https://github.com/spacemeshos/go-spacemesh/blob/e12e38f328296553664d7592f67b921be5be2e49/fetch/fetch.go#L100-L105

countvonzero: so i recall there was a earlier check on len(peers) before `organizeRequests`. 
but never mind that now. i think the easiest fix is to return no peer error from `requestHashBatchFromPeers()` instead of keep going. like i said. removing panic just kicking the can down the road.

piersy: @countvonzero I've removed the call to `log.Fatal` as well.

fetch/fetch.go
```diff
@@ -98,9 +98,6 @@ func DefaultConfig() Config {
 
 // randomPeer returns a random peer from current peer list.
 func randomPeer(peers []p2p.Peer) p2p.Peer {
-	if len(peers) == 0 {
-		log.Fatal("cannot send fetch: no peers found")
-	}
```
countvonzero, line 103: i think here you do need to keep it. otherwise peers[] will panic with index error as well.

countvonzero (approved):
in any case, not super important. just think that it's better to panic with sane error than with an array index OOB

fetch/fetch.go
piersy: This is protected against by the check in organizeRequests.

piersy:
@countvonzero  I don't see a reason to check this slice for zero length any more than any other slice passed as a parameter if the outcome either way is to panic. Doing this across the codebase would add a lot of boilerplate, and not provide much value if any.

piersy:
bors merge

piersy:
bors merge

dshulyak:
bors merge

PR #4659, opened 2023-07-07
[Merged by Bors] - atx: check last 2 epochs for highest tick atx

countvonzero:
## Motivation
due to different poet speed, and once poet phase shift is implemented, the highest ticked atx might be still from previous epoch (from those fast poets), and that those that built on top of them are not yet published.

to strike a balance btwn honest miners and malicious one that inject high tick atx to an old epoch, we used last two epochs to select the highest ticked atx for positioning and commitment atx.


dshulyak (approved):

fasmat (approved):

countvonzero:
bors merge

PR #4661, opened 2023-07-07
multisig: add test for 1of1

countvonzero:


dshulyak (approved):
i don't see the benefit of having this test, there are parametrized tests in genvm (maybe no 1-1, but can easily added) and parametrized tests in system tests (i definitely had 1-1 for vesting, which is what you are after probably), but feel free to merge it you want

countvonzero:
> i don't see the benefit of having this test, there are parametrized tests in genvm (maybe no 1-1, but can easily added) 

that was my first attempt. but it failed at "duplicates" and i didn't invest time to fix it.

> and parametrized tests in system tests (i definitely had 1-1 for vesting, which is what you are after probably), but feel free to merge it you want

thanks. the PR is more like a form of asking question. will close.

PR #4662, opened 2023-07-07
[Merged by Bors] - node: make sure that hrp is configurable

dshulyak:


dshulyak:
@fasmat could you check this change, it was included in https://github.com/spacemeshos/go-spacemesh/pull/4663

fasmat (approved):

fasmat:
I would not use different binaries for different networks. The HRP has to be provided as part of the config or via CLI parameters. Maybe we have to refactor the Adress generator as a service then as well that gets injected the config to receive it.

dshulyak:
> I would not use different binaries for different networks. The HRP has to be provided as part of the config or via CLI parameters. Maybe we have to refactor the Adress generator as a service then as well that gets injected the config to receive it.

ok, i will remove that from description, so we can reconsider whats better later

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #4663, opened 2023-07-07
[Merged by Bors] - api: node info

dshulyak:
requires: https://github.com/spacemeshos/go-spacemesh/pull/4662
api: https://github.com/spacemeshos/api/pull/250

add handler with:
- network hrp, as it is hardcoded in smapp
- first/effective genesis, for clients to understand that checkpoint happened
- epoch size, just useful info

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,7 +15,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/beacon"
 	"github.com/spacemeshos/go-spacemesh/bootstrap"
 	"github.com/spacemeshos/go-spacemesh/checkpoint"
-	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/fetch"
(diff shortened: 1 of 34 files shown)


dshulyak:
@brusherru @kacpersaw

config/config.go
```diff
@@ -111,6 +109,8 @@ type BaseConfig struct {
 
 	DatabaseConnections     int  `mapstructure:"db-connections"`
 	DatabaseLatencyMetering bool `mapstructure:"db-latency-metering"`
+
+	NetworkHRP string `mapstructure:"network-hrp"`
```
fasmat: Do we need the global if the HRP is part of the config? If it is too much effort, I would at least create an issue for this to refactor in the future.

dshulyak: the main problem is that hrp is used when address is printed. there are a couple of logs with addresses across the codebase. maybe we can reconsider

fasmat (approved):

dshulyak:
bors merge

PR #4664, opened 2023-07-07
[Merged by Bors] - syncer: dont block initialization before genesis

dshulyak:
if ticker returns 0 we should consider node to be "synced" and "atx synced" and allow to run initialization


fasmat (approved):

dshulyak:
bors merge

PR #4667, opened 2023-07-07
[Merged by Bors] - cmd: generate as many private/public key as you want with golden atxid

dshulyak:
```
Usage of genesisgen:
  -extra string
    	genesis extra data. valid size is between [1, 255]
  -n int
    	number of keys (default 10)
  -time string
    	genesis time. format 2006-01-02T15:04:05Z07:00
```


pigmej (approved):

dshulyak:
bors merge

PR #4669, opened 2023-07-08
[Merged by Bors] - api, events: cache last 100 admin events in memory

dshulyak:
grpc subscriber might be slow to join or disconnect due to flakes and other reasons, we don't want it to miss
important events because of that.

added simple ring buffer, and used it to cache last 100 user events. they are copied (pointers) and are sent to the newly created stream.

### DIFF
--- a/api/grpcserver/admin_service.go
+++ b/api/grpcserver/admin_service.go
@@ -109,7 +109,6 @@ func (a AdminService) EventsStream(req *pb.EventStreamRequest, stream pb.AdminSe
 		return status.Errorf(codes.Unavailable, "can't send header")
 	}
 	buf.Iterate(func(ev events.UserEvent) bool {
-		fmt.Println(ev.Event.Help)
 		err = stream.Send(ev.Event)
 		return err == nil
 	})
--- a/api/grpcserver/admin_service.go
+++ b/api/grpcserver/admin_service.go
@@ -98,7 +98,7 @@ func (a AdminService) Recover(_ context.Context, _ *pb.RecoverRequest) (*empty.E
 }
 
 func (a AdminService) EventsStream(req *pb.EventStreamRequest, stream pb.AdminService_EventsStreamServer) error {
-	sub, err := events.Subscribe[events.UserEvent]()
+	sub, buf, err := events.SubscribeUserEvents(events.WithBuffer(1000))
 	if err != nil {
 		return status.Errorf(codes.FailedPrecondition, err.Error())
 	}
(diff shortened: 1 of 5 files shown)


events/reporter.go
```diff
@@ -511,3 +544,71 @@ func CloseEventReporter() {
 		reporter = nil
 	}
 }
+
+func newRing[T any](size int) *Ring[T] {
```
fasmat, line 548: Did you consider `container/ring`? https://pkg.go.dev/container/ring

Would save us maintaining our own implementation. It's not generic, but since it is only used by the `UserEvent` emitter, we can safely assume that any cast to `UserEvent` will not panic.

api/grpcserver/admin_service.go
```diff
@@ -108,6 +108,13 @@ func (a AdminService) EventsStream(req *pb.EventStreamRequest, stream pb.AdminSe
 	if err := stream.SendHeader(metadata.MD{}); err != nil {
 		return status.Errorf(codes.Unavailable, "can't send header")
 	}
+	buf.Iterate(func(ev events.UserEvent) bool {
+		err = stream.Send(ev.Event)
+		return err == nil
+	})
```
fasmat, line 114: Instead of returning `buf`, I think it would be safer to pass the iterate function to `SubscribeUserEvents`. This would eliminate the need to export the `Ring` itself.

events/reporter.go
```diff
@@ -411,8 +420,30 @@ type EventReporter struct {
...
-	stopChan           chan struct{}
+	events             struct {
+		sync.Mutex
+		buf     *Ring[UserEvent]
+		emitter event.Emitter
+	}
+	stopChan chan struct{}
+}
+
+func (r *EventReporter) EmitUserEvent(ev UserEvent) error {
```
fasmat: Can be made private I think

events/reporter.go
```diff
@@ -411,8 +420,30 @@ type EventReporter struct {
...
+	r.events.Lock()
+	defer r.events.Unlock()
+	r.events.buf.insert(ev)
+	return r.events.emitter.Emit(ev)
+}
+
+func (r *EventReporter) SubUserEvents(opts ...SubOpt) (*BufferedSubscription[UserEvent], *Ring[UserEvent], error) {
```
fasmat: Can be made private I think

events/reporter.go
dshulyak: container/ring (for this purpose list is the same) maintains structure using pointers. i want something that i can copy fast (e.g in this case i need to copy 100 8-byte pointers with single copy) and then use it without synchronizing with original ring. for example i write events to stream, if i will hold lock during this time it may screw up consensus

api/grpcserver/admin_service.go
dshulyak: i copy and then iterate. i don't want to always copy when i iterate

events/reporter.go
dshulyak: done

fasmat (approved):

dshulyak:
bors merge

Issue #4074, opened 2023-02-21
flaky test: TestCache_Account_Add_NonceTooSmall


fasmat:
#### Description

Failed during unrelated change: https://github.com/spacemeshos/go-spacemesh/actions/runs/4231221021/jobs/7349474661

```
=== FAIL: txs TestCache_Account_Add_NonceTooSmall (0.00s)
    cache_test.go:655: 
        	Error Trace:	/home/runner/work/go-spacemesh/go-spacemesh/txs/cache_test.go:655
        	Error:      	Target error should be in err chain:
        	            	expected: "bad nonce"
        	            	in chain: 
        	Test:       	TestCache_Account_Add_NonceTooSmall
```

poszu:
The test is still flaky: https://github.com/spacemeshos/go-spacemesh/actions/runs/5517581731/jobs/10060394729

PR #4678, opened 2023-07-10
[Merged by Bors] - Support optional minerID in K2 PoW input

### LINKED ISSUE #4656: add miner id to k2pow and its verification
## Description
to prepare for https://community.spacemesh.io/t/proof-of-work-and-post-proof-generation/361, and to mitigate against an adversary creating a large number of small identities, but stores only half of the claimed storage, for genesis we should inject miner id into k2pow work and verification of k2pow to ensure the miner does its own k2pow work

poszu:
## Motivation
Closes #4656 

## Changes
Adds an optional minerID that created the k2 pow (_k2 pow creator id_) to post proving and verifying.

## Reasoning for making it optional
In order to be able to test in on a real network (devnet-401) pre-genesis. For genesis we will set `post-minerid-in-k2-pow-since-epoch = 0` and later remove the config field along with branches in code.

## Test Plan
- existing UTs
- systests
- devnet-401 (set `post-minerid-in-k2-pow-since-epoch` to the epoch near in the future)

### DIFF
--- a/Makefile-libs.Inc
+++ b/Makefile-libs.Inc
@@ -50,7 +50,7 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.3.0
+POSTRS_SETUP_REV = 0.4.0-rc2
 POSTRS_SETUP_ZIP = libpost-$(platform)-v$(POSTRS_SETUP_REV).zip
 POSTRS_SETUP_URL_ZIP ?= https://github.com/spacemeshos/post-rs/releases/download/v$(POSTRS_SETUP_REV)/$(POSTRS_SETUP_ZIP)
 POSTRS_PROFILER_ZIP = profiler-$(platform)-v$(POSTRS_SETUP_REV).zip
--- a/Makefile-libs.Inc
+++ b/Makefile-libs.Inc
@@ -50,7 +50,7 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.4.0-rc2
+POSTRS_SETUP_REV = 0.4.0
 POSTRS_SETUP_ZIP = libpost-$(platform)-v$(POSTRS_SETUP_REV).zip
 POSTRS_SETUP_URL_ZIP ?= https://github.com/spacemeshos/post-rs/releases/download/v$(POSTRS_SETUP_REV)/$(POSTRS_SETUP_ZIP)
 POSTRS_PROFILER_ZIP = profiler-$(platform)-v$(POSTRS_SETUP_REV).zip
--- a/activation/nipost_test.go
+++ b/activation/nipost_test.go
@@ -123,6 +123,7 @@ func TestNIPostBuilderWithClients(t *testing.T) {
 		t.Parallel()
 		postCfg := DefaultPostConfig()
 		postCfg.PowDifficulty[0] = 64
+		postCfg.MinerIDInK2PowSinceEpoch = 2
 		postProvider := newTestPostManager(t, withPostConfig(postCfg))
 		logger := logtest.New(t).WithName("validator")
 		verifier, err := NewPostVerifier(postProvider.Config(), logger)
@@ -138,18 +139,19 @@ func TestNIPostBuilderWithClients(t *testing.T) {
 		)
 		_, err = v.NIPost(
 			context.Background(),
+			challenge.PublishEpoch,
 			postProvider.id,
 			postProvider.commitmentAtxId,
 			nipost,
 			challengeHash,
 			postProvider.opts.NumUnits,
 			verifying.WithLabelScryptParams(postProvider.opts.Scrypt),
-			verifying.WithPowCreator(postProvider.id.Bytes()),
 		)
 		r.NoError(err)
 
 		_, err = v.NIPost(
 			context.Background(),
+			0,
 			postProvider.id,
 			postProvider.commitmentAtxId,
 			nipost,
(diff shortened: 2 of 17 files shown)


Makefile-libs.Inc
```diff
@@ -50,7 +50,7 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.3.0
+POSTRS_SETUP_REV = 0.4.0-rc2
```
poszu: Bump when post-rs v0.4.0 is released

go.mod
```diff
@@ -39,7 +39,7 @@ require (
 	github.com/spacemeshos/go-scale v1.1.10
 	github.com/spacemeshos/merkle-tree v0.2.2
 	github.com/spacemeshos/poet v0.8.6
-	github.com/spacemeshos/post v0.8.2
+	github.com/spacemeshos/post v0.8.3-rc1
```
poszu: Bump when post v0.8.3 is released

poszu:
bors try

poszu:
bors try

node/node.go
```diff
@@ -649,6 +649,7 @@ func (app *App) initServices(ctx context.Context, poetClients []activation.PoetP
 		trtl,
 		app.addLogger(ATXHandlerLogger, lg),
 		poetCfg,
+		app.Config.POST,
```
fasmat: This breaks layers of abstraction. Imo the postCfg doesn't need to be passed to the `atxHandler`, it is already passed to the validator (line 552).

I think it would be cleaner if instead the publish epoch is passed to `nipostValidator.NIPost` and `nipostValidator.Post` which then apply the logic of checking if it needs to be added to the pow or not.

activation/post.go
```diff
@@ -444,12 +446,15 @@ func (mgr *PostSetupManager) GenerateProof(ctx context.Context, challenge []byte
...
 
-	proof, proofMetadata, err := proving.Generate(ctx, challenge, mgr.cfg.ToConfig(), mgr.logger.Zap(),
+	opts := []proving.OptionFunc{
 		proving.WithDataSource(mgr.cfg.ToConfig(), mgr.id.Bytes(), mgr.commitmentAtxId.Bytes(), mgr.lastOpts.DataDir),
 		proving.WithNonces(mgr.provingOpts.Nonces),
 		proving.WithThreads(mgr.provingOpts.Threads),
 		proving.WithPowFlags(mgr.provingOpts.Flags),
-	)
+	}
+	opts = append(opts, options...)
```
fasmat, line 455: This has the drawback that the caller can now overwrite all of these options, which I think is not intended.

We should add at least a `// FIXME:` comment here to address this in the future.

poszu: I know, but I wouldn't consider it a drawback. Eventually, I see knobs like `threads` and `nonces` and `pow flags` as _options_ .

node/node.go
poszu: :+1:
Done

activation/post.go
fasmat: Yes, but we don't want to pass these options n every function call, as they are not expected to change while the node is running 🙂. Thats also why they are set once via the `provingOpts` when initializing this service.

activation/nipost_test.go
```diff
@@ -114,33 +116,97 @@ func TestNIPostBuilderWithClients(t *testing.T) {
...
+			nipost,
+			challengeHash,
+			postProvider.opts.NumUnits,
+			verifying.WithLabelScryptParams(postProvider.opts.Scrypt),
+			// no pow creator id -> proof invalid
+		)
+		r.Error(err)
+	})
+	t.Run("POST without pow creator ID", func(t *testing.T) {
+		t.Parallel()
```
fasmat, line 166: There is no upside in executing these tests in parallel. Both generating and verifying proofs is done with a mutex. `t.Parallel()` only slows down the tests in this case.

fasmat (approved):

activation/nipost_test.go
```diff
@@ -114,33 +116,97 @@ func TestNIPostBuilderWithClients(t *testing.T) {
...
+			context.Background(),
+			challenge.PublishEpoch,
+			postProvider.id,
+			postProvider.commitmentAtxId,
+			nipost,
+			challengeHash,
+			postProvider.opts.NumUnits,
+			verifying.WithLabelScryptParams(postProvider.opts.Scrypt),
+		)
+		r.NoError(err)
```
fasmat: this `r.NoError` will fail the parent test, not the subtest. I'd recommend just using `require.NoError(t, err)` instead 🙂

fasmat (approved):

activation/nipost_test.go
poszu: Actually, neither generation nor verification is done behind a mutex lock. Initialization is (per provider). On my PC the test speedup is around 2x. It would be even better if the test was refactored to reused inited data.

fasmat: Yes, you are right. Only initialization is done behind a mutex.

fasmat: But this will still cause all the tests to initially lock on initialization before continuing with generating and verifying proofs.

poszu:
bors merge

activation/post.go
poszu: Actually changing them while the node is running could be a good UX improvement :thinking: The user could pick how much CPU resources should be spent, maybe use more when the PC is idle etc.

poszu:
bors merge

poszu:
bors merge

poszu:
flaky `TestCache_Account_Add_NonceTooSmall` (https://github.com/spacemeshos/go-spacemesh/issues/4074)

bors merge

PR #4679, opened 2023-07-10
[Merged by Bors] - prepare config for mainnet

dshulyak:
config is extracted into a separate file so that we can track any accidental changes in protocol parameters easier

### DIFF
--- a/config/mainnet.go
+++ b/config/mainnet.go
@@ -41,6 +41,7 @@ func MainnetConfig() Config {
 		},
 		Genesis: &GenesisConfig{
 			GenesisTime: "2023-07-14T08:00:00Z",
+			Accounts:    map[string]uint64{},
 		},
 		Tortoise: tortoise.Config{
 			Hdist:                    200,
--- a/config/mainnet.go
+++ b/config/mainnet.go
@@ -33,11 +33,12 @@ func MainnetConfig() Config {
 			LayerDuration:  5 * time.Minute,
 			LayersPerEpoch: 4032,
 
-			TxsPerProposal: 700,       // TODO
+			TxsPerProposal: 700,       // https://github.com/spacemeshos/go-spacemesh/issues/4559
 			BlockGasLimit:  100107000, // 3000 of spends
 
-			OptFilterThreshold: 90, // TODO
-			TickSize:           9331200,
+			OptFilterThreshold: 90,
+
+			TickSize: 9331200,
 		},
 		Genesis: &GenesisConfig{
 			GenesisTime: "2023-07-14T08:00:00Z",
@@ -49,7 +50,8 @@ func MainnetConfig() Config {
 			WindowSize:               10000,
 			MaxExceptions:            1000,
 			BadBeaconVoteDelayLayers: 4032,
-			MinimalActiveSetWeight:   100 * 9331200, // TODO known units per tick size
+			// TODO update it with safe but reasonble minimum weight in network before first ballot
+			MinimalActiveSetWeight: 1000 * 9331200,
 		},
 		HARE: hareConfig.Config{
 			N:               200,
--- a/config/mainnet.go
+++ b/config/mainnet.go
@@ -36,8 +36,8 @@ func MainnetConfig() Config {
 			TxsPerProposal: 700,       // TODO
 			BlockGasLimit:  100107000, // 3000 of spends
 
-			OptFilterThreshold: 90,        // TODO
-			TickSize:           1_000_000, // TODO
+			OptFilterThreshold: 90, // TODO
+			TickSize:           9331200,
 		},
 		Genesis: &GenesisConfig{
 			GenesisTime: "2023-07-14T08:00:00Z",
@@ -48,7 +48,7 @@ func MainnetConfig() Config {
 			WindowSize:               10000,
 			MaxExceptions:            1000,
 			BadBeaconVoteDelayLayers: 4032,
-			MinimalActiveSetWeight:   100 * 1_000_000, // TODO known units per tick size
+			MinimalActiveSetWeight:   100 * 9331200, // TODO known units per tick size
 		},
 		HARE: hareConfig.Config{
 			N:               200,
@@ -64,12 +64,15 @@ func MainnetConfig() Config {
 		Beacon: beacon.Config{
 			Kappa:                    40,
 			Q:                        big.NewRat(1, 3),
+			Theta:                    big.NewRat(1, 4),
 			GracePeriodDuration:      10 * time.Minute,
 			ProposalDuration:         4 * time.Minute,
 			FirstVotingRoundDuration: 30 * time.Minute,
 			RoundsNumber:             300,
 			VotingRoundDuration:      4 * time.Minute,
 			WeakCoinRoundDuration:    4 * time.Minute,
+			VotesLimit:               100,
+			BeaconSyncWeightUnits:    800,
 		},
 		POET: activation.PoetConfig{
 			PhaseShift:        240 * time.Hour,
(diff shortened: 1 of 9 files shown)


config/mainnet.go
```diff
@@ -0,0 +1,105 @@
...
+			PhaseShift:        240 * time.Hour,
+			CycleGap:          12 * time.Hour,
+			GracePeriod:       1 * time.Hour,
+			RequestRetryDelay: 10 * time.Second,
+			MaxRequestRetries: 10,
+		},
+		POST: activation.PostConfig{
+			MinNumUnits:   4,
+			MaxNumUnits:   1000,
+			LabelsPerUnit: 4294967296, // TODO 64GB / 16 ?
```
pigmej: 64GiB SU
4 SU minimum
No MaxNumUnits

config/mainnet.go
```diff
@@ -0,0 +1,105 @@
...
+			K1:            26,
+			K2:            37,
+			K3:            37,
```
pigmej, line 97: I think it's not needed anymore @poszu did we deleted it for good?

p2p/host.go
```diff
@@ -31,7 +31,7 @@ func DefaultConfig() Config {
 	return Config{
 		Listen:             "/ip4/0.0.0.0/tcp/7513",
 		Flood:              false,
-		MinPeers:           6,
+		MinPeers:           20,
```
pigmej, line 34: Why?

config/mainnet.go
```diff
@@ -0,0 +1,105 @@
...
+			NetworkHRP:          "sm",
+
+			LayerDuration:  5 * time.Minute,
+			LayersPerEpoch: 4032,
+
+			TxsPerProposal: 700,       // TODO
+			BlockGasLimit:  100107000, // 3000 of spends
+
+			OptFilterThreshold: 90,        // TODO
+			TickSize:           1_000_000, // TODO
```
pigmej: 80000 per second
that is 93312000000 ticks per epoch `(epoch_length * 60 * 60 - cycle_gap_seconds) * poet_ticks_per_second`

`epoch_length` 14 * 24 hours.
`cycle_gap_seconds` 43200

To normalize to 10k we need to make `tick_height = 9331200`


The poets are making 77-78k consistently. That leaves some buffer for future.

config/mainnet.go
dshulyak: i believe it is still used, i used it while testing to make post less cpu heavy.
it is actually the number of labels that will be verified. it can be removed ofcourse and always use k2, but i found it useful during testing

config/mainnet.go
dshulyak: so 64 / 16 = 4294967296.
i think we always have default max num units, i need to check if don't overflow some values otherwise

p2p/host.go
dshulyak: i want node to connect with more peers than we will have bootnodes (usually we have 10, therefore double), this way they will always eagerly look for other nodes and won't wait until they disconnected from bootnodes

config/mainnet.go
```diff
@@ -0,0 +1,108 @@
...
+		Genesis: &GenesisConfig{
+			GenesisTime: "2023-07-14T08:00:00Z",
+		},
+		Tortoise: tortoise.Config{
+			Hdist:                    200,
+			Zdist:                    2,
+			WindowSize:               10000,
+			MaxExceptions:            1000,
+			BadBeaconVoteDelayLayers: 4032,
+			MinimalActiveSetWeight:   100 * 9331200, // TODO known units per tick size
```
dshulyak: we can update it right before 2nd epoch. this is used as dos protection against identities spawning too many ballots

config/mainnet.go
pigmej: re maxnum units

to fit uint64 I assumed:
`587202560` total number of units in an epoch
and then we can keep it for 100years and yet we will not consume more than 53bits on tortoise.

that yields: `15267266560000000` over a year.  

why `587202560` ? It's 35EiB in 64GiB units.

dshulyak: but have you factored tick count into this formula? because we work with spacetimeunits in consensus

config/mainnet.go
```diff
@@ -0,0 +1,108 @@
...
+			Q:                        big.NewRat(1, 3),
+			Theta:                    big.NewRat(1, 4),
+			GracePeriodDuration:      10 * time.Minute,
+			ProposalDuration:         4 * time.Minute,
+			FirstVotingRoundDuration: 30 * time.Minute,
+			RoundsNumber:             300,
+			VotingRoundDuration:      4 * time.Minute,
+			WeakCoinRoundDuration:    4 * time.Minute,
+			VotesLimit:               100,
+			BeaconSyncWeightUnits:    800,
```
dshulyak, line 83: beacon should complete in around ~41 hour

config/mainnet.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+		POET: activation.PoetConfig{
+			PhaseShift:        240 * time.Hour,
+			CycleGap:          12 * time.Hour,
+			GracePeriod:       1 * time.Hour,
```
dshulyak, line 81: setting grace period to 1h means that post effectively should take 10h

config/mainnet.go
```diff
@@ -0,0 +1,108 @@
...
+			RequestRetryDelay: 10 * time.Second,
+			MaxRequestRetries: 10,
+		},
+		POST: activation.PostConfig{
+			MinNumUnits:   4,
+			LabelsPerUnit: 4294967296,
```
dshulyak, line 94: 64 << 30 / 16

fasmat: or just 2 << 32 😄

fasmat: So one unit will be 64 GiB?

dshulyak: yes, with min post data size 256GB

config/mainnet.go
```diff
@@ -0,0 +1,110 @@
...
+			K1:            26,
+			K2:            37,
+			K3:            37,
+			PowDifficulty: activation.DefaultDifficulty,
```
pigmej: @poszu can we calculate difficulty, please?

p2p/host.go
pigmej: Ok understood. thx.

config/mainnet.go
dshulyak: i thought that we always use this and then scale it based on number of units based on what @poszu explained

config/mainnet.go
poszu: As @dshulyak said, this is the number of verified labels (the proof contains K2 indices, but the verifier checks K3 of them). 

Currently, K3 should equal K2.

config/mainnet.go
poszu: According to https://replit.com/@poszu/PowDifficulty, the pow difficulty should be `00037ec8ec25e6d2c00000000000000000000000000000000000000000000000` if we target 10 minut PoW for 256GiB storage for 64 nonces for 500H/s machine.

dshulyak: @poszu can you push that into this branch please?

brusherru:
For the protocol, the current Smapp implementation requires these fields in the node config:
- `post`
  - `post-labels-per-unit`
  - `post-max-numunits`
- `poet`
  - `cycle-gap`
- `genesis`
  - `genesis-time`
  - `genesis-extra-data`
- `main`
  - ~~`network-hrp`~~ (`sm` by default, required only for testnet)
  - `layer-duration`
  - `layers-per-epoch`

These params are used by Smapp to calculate time, pos size, genesis id, and so on.

countvonzero:
bors merge

countvonzero (approved):

countvonzero:
bors merge

PR #4680, opened 2023-07-10
[Merged by Bors] - feat: persist nipost state efficiently

### LINKED ISSUE #4506: Persisting nipost state is done inefficiently
## Description
NiPoST state persistence can be optimised to avoid coping data when writing or reading state from disk. The general process should be:

writing:
- create a new temporary file
- write CRC (as zeroes) to file
- use a `io.MultiWriter` to write to the file and calculate `CRC` simultaneously
- seek back to the beginning of the file and update CRC
- use [atomic.ReplaceFile](https://github.com/natefinch/atomic/blob/v1.0.1/file_unix.go#L12) to replace existing file with new file

reading:
- read CRC
- use a `io.TeeReader` to calculate CRC while reading the file
- decode data by reading from `io.TeeReader`
- compare calculated CRC with CRC read from file
- check if all data has been read from file while decoding (i.e. did read end in `io.EOF`?)

## Affected code
- Functions in `activation/nipost_state.go`
- Possibly other places where data is persisted to disk
- Discussion can be found here: https://github.com/spacemeshos/go-spacemesh/pull/4493/files#r1226351812

0xjac:
## Motivation

Closes #4506

## Changes

- write: compute CRC as the file is written. (requires two writes: content & CRC)
- read: compute CRC as the file is read.
- use CRC Size constant from CRC pkg.
- *extra: Properly close the files when reading and writing*

The changes match what was requested in #4506 with regards to storing NiPoST state.

## Test Plan

Unit tests `TestWriteCRC` & `TestReadCRC` were added to specifically test the writing and reading of valid and invalid CRCs.  
Note: The write tests do not rely on `read()` and the read tests do not rely on `write()`.

I am unsure if a more elaborate plan is needed.

## Documentation
As far as I can see, there is no documentation related to the NiPoST state storage format.

## DevOps Notes

  This PR changes error messages when writing and reading NiPoST states from file. Specifically:
    - write: new message `create temporary file <path>: <err>`
    - write: replace `save file <path>: err` with `save file from <tmpfile>, <path>: <err>`
    - read: replace `read file <path>: <err>` with `open file <path>: <err>`
    - read: new message `failed to get file info <path>: <err>`
    - read: change in `wrong checksum <read_cksum>, computed <computed_cksum>`: checksums are displayed in (uppercase) hex rather than decimal.

### DIFF
--- a/activation/nipost_state.go
+++ b/activation/nipost_state.go
@@ -1,7 +1,6 @@
 package activation
 
 import (
-	"bytes"
 	"encoding/binary"
 	"fmt"
 	"hash/crc64"
@@ -20,43 +19,71 @@ const (
 	challengeFilename = "nipost_challenge.bin"
 	builderFilename   = "nipost_builder_state.bin"
 	postFilename      = "post.bin"
-	crc64Size         = 8
 )
 
 func write(path string, data []byte) error {
-	buf := bytes.NewBuffer(nil)
-	checksum := crc64.New(crc64.MakeTable(crc64.ISO))
-	w := io.MultiWriter(buf, checksum)
-	if _, err := w.Write(data); err != nil {
-		return fmt.Errorf("write data %v: %w", path, err)
+	tmp, err := os.CreateTemp("", filepath.Base(path))
+	if err != nil {
+		return fmt.Errorf("create temporary file %s: %w", tmp.Name(), err)
 	}
 
-	crc := make([]byte, crc64Size)
+	checksum := crc64.New(crc64.MakeTable(crc64.ISO))
+	w := io.MultiWriter(tmp, checksum)
+	if _, err = w.Write(data); err != nil {
+		_ = tmp.Close()
+		return fmt.Errorf("write data %v: %w", tmp.Name(), err)
+	}
+
+	crc := make([]byte, crc64.Size)
 	binary.BigEndian.PutUint64(crc, checksum.Sum64())
-	if _, err := buf.Write(crc); err != nil {
-		return fmt.Errorf("write checksum %v: %w", path, err)
+	if _, err = tmp.Write(crc); err != nil {
+		_ = tmp.Close()
+		return fmt.Errorf("write checksum %s: %w", tmp.Name(), err)
 	}
-	if err := atomic.WriteFile(path, buf); err != nil {
-		return fmt.Errorf("save file %v: %w", path, err)
+
+	if err = tmp.Close(); err != nil {
+		return fmt.Errorf("failed to close tmp file %s: %w", tmp.Name(), err)
 	}
+
+	if err = atomic.ReplaceFile(tmp.Name(), path); err != nil {
+		return fmt.Errorf("save file from %s, %s: %w", tmp.Name(), path, err)
+	}
+
 	return nil
 }
 
 func read(path string) ([]byte, error) {
-	data, err := os.ReadFile(path)
+	file, err := os.Open(path)
 	if err != nil {
-		return nil, fmt.Errorf("read file %v: %w", path, err)
+		return nil, fmt.Errorf("open file %s: %w", path, err)
 	}
+
+	defer file.Close()
+
+	fInfo, err := file.Stat()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file info %s: %w", path, err)
+	}
+
+	data := make([]byte, fInfo.Size()-crc64.Size)
 	checksum := crc64.New(crc64.MakeTable(crc64.ISO))
-	end := len(data) - crc64Size
-	if _, err = checksum.Write(data[:end]); err != nil {
-		return nil, fmt.Errorf("read checksum %v: %w", path, err)
+	if _, err = io.TeeReader(file, checksum).Read(data); err != nil {
+		return nil, fmt.Errorf("read file %s: %w", path, err)
 	}
-	saved := binary.BigEndian.Uint64(data[end:])
-	if saved != checksum.Sum64() {
-		return nil, fmt.Errorf("wrong checksum %d, computed %d", saved, checksum.Sum64())
+
+	saved := make([]byte, crc64.Size)
+	if _, err = file.Read(saved); err != nil {
+		return nil, fmt.Errorf("read checksum %s: %w", path, err)
 	}
-	return data[:end], nil
+
+	savedChecksum := binary.BigEndian.Uint64(saved)
+
+	if savedChecksum != checksum.Sum64() {
+		return nil, fmt.Errorf(
+			"wrong checksum 0x%X, computed 0x%X", savedChecksum, checksum.Sum64())
+	}
+
+	return data, nil
 }
 
 func load(filename string, dst scale.Decodable) error {
--- a/activation/nipost_state.go
+++ b/activation/nipost_state.go
@@ -22,7 +22,7 @@ const (
 )
 
 func write(path string, data []byte) error {
-	tmp, err := os.CreateTemp("", filepath.Base(path))
+	tmp, err := os.Create(fmt.Sprintf("%s.tmp", path))
 	if err != nil {
 		return fmt.Errorf("create temporary file %s: %w", tmp.Name(), err)
 	}
--- a/activation/nipost_state_test.go
+++ b/activation/nipost_state_test.go
@@ -10,6 +10,75 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestWriteCRC(t *testing.T) {
+	tests := []struct {
+		name     string
+		data     []byte
+		crc      []byte
+		validCRC bool
+	}{
+		{
+			name:     "validCRC",
+			data:     []byte("spacemesh"),
+			crc:      []byte{0xC6, 0x10, 0x72, 0xDF, 0x52, 0xD7, 0x74, 0x0E},
+			validCRC: true,
+		}, {
+			name:     "badCRC",
+			data:     []byte("spacemesh"),
+			crc:      []byte{0xC7, 0x11, 0x73, 0xE0, 0x53, 0xD8, 0x7, 0x0F},
+			validCRC: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			path := filepath.Join(t.TempDir(), tt.name)
+			require.NoError(t, write(path, tt.data))
+
+			raw, err := os.ReadFile(path)
+			require.NoError(t, err)
+			require.Equal(t, tt.validCRC, bytes.Equal(raw, append(tt.data, tt.crc...)))
+		})
+	}
+}
+
+func TestReadCRC(t *testing.T) {
+	tests := []struct {
+		name   string
+		data   []byte
+		crc    []byte
+		errMsg string
+	}{
+		{
+			name:   "validCRC",
+			data:   []byte("spacemesh"),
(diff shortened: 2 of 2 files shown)


0xjac:
Regarding the change of format for the files. If keeping the content of the existing files is requested, I think the best solution is to have some kind of migration logic where the files are read the old way and persisted the new way.

Also note: the issue #4506 talks about NiPoST states but also mentions other places where data is persisted to disk. I had a brief look and found that:
- `p2p/book/book.go` has some [persistence logic using CRC-64](https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/book/book.go#L389-L439)
- `p2p/identity.go` has a [TODO to add a CRC-64 check](https://github.com/spacemeshos/go-spacemesh/blob/develop/p2p/identity.go#L52)

However those seem unrelated to the NiPoST state. Let me know if you want me to implement CRC properly for those too in a separaate PR.

activation/nipost_state.go
```diff
@@ -20,43 +19,64 @@ const (
...
+	w := io.MultiWriter(tmp, checksum)
+	if _, err = w.Write(data); err != nil {
+		return fmt.Errorf("write data %v: %w", tmp.Name(), err)
 	}
 
-	crc := make([]byte, crc64Size)
 	binary.BigEndian.PutUint64(crc, checksum.Sum64())
-	if _, err := buf.Write(crc); err != nil {
-		return fmt.Errorf("write checksum %v: %w", path, err)
+	if _, err = tmp.WriteAt(crc, 0); err != nil {
```
dshulyak: is it a common practice to add crc at the start? what are the disadvantages of adding it as a suffix? i think crc as a suffix allows to preserve sequential writes, it obviously doesn't matter here

dshulyak (changes requested):
> Regarding the change of format for the files. If keeping the content of the existing files is requested, I think the best solution is to have some kind of migration logic where the files are read the old way and persisted the new way.

i think it would be needed, it is a pain to update othrwise. but also in order to simplify deployment i would run migration on go-spacemesh startup. e.g it can look up files with version 1 and migrate them to version 2. otherwise good change

we are merging last minute cleanups and preparations for release, once it is out, this change will be merged as well.

activation/nipost_state.go
0xjac: The [first point of the specification on Wikipedia](https://en.wikipedia.org/wiki/Cyclic_redundancy_check) seems to indicate it is. To be taken with a grain of salt.

That being said I would argue having some fixed length data first and the variable length data afterwards makes for a better design. especially if the variable length data is really large. It is also the way it was specified in the issue.

Also looking further into it, adding a zeros prefix does not change the CRC so I can rewrite the `write` function to remove the initial write of the prefix.

Or I can rewrite it with the CRC as a suffix. The logic might be a bit more complicated for the read because we need the logic to stop before the end of the file but that can work too.

0xjac:
> > Regarding the change of format for the files. If keeping the content of the existing files is requested, I think the best solution is to have some kind of migration logic where the files are read the old way and persisted the new way.
> 
> i think it would be needed, it is a pain to update othrwise. but also in order to simplify deployment i would run migration on go-spacemesh startup. e.g it can look up files with version 1 and migrate them to version 2. otherwise good change
> 
> we are merging last minute cleanups and preparations for release, once it is out, this change will be merged as well.

The hard part is those files have no concept of versioning so I can't just run something at startup every time. The migration has to happen during the update of the software (which might be platform dependent).

I'm not sure exactly how you proceed in those cases. I can think of the following solutions:
- You have some mechanism to run code when upgrading `go-spacemesh`
- You have some mechanism to run an update/migration only once on startup and not every time.
- Create a separate command and document that the user run it after updating the software.

Let me know which way you want to go about this?

Edit: this is only relevant if CRC is a prefix. As pointed in the original issue, it is a pain to change the format after the launch. But if time is short and you don't want to break anything you could keep it as a suffix and once you have some sort of migration mechanism in place between versions of spacemesh itself, then you can move the CRC as a prefix.

dshulyak:
> The hard part is those files have no concept of versioning so I can't just run something at startup every time. The migration has to happen during the update of the software (which might be platform dependent).

i think the most future proof approach is to add a version. to run the first migration we can have some heuristic, like if after crc we have `{` then it is a version 0. sometimes migrations won't be possible at all, as content of the file is used as a challenge for downstream protocols, and we will have to keep several versions in the codebase/state

regarding upgrades i think we want to keep it simple for now and try to do all of it on spacemesh startup.

@fasmat what do you think?

0xjac:
Regarding version information, I would include that as part of the "content" so the version information is also taken into account for the CRC. So that would move it out of scope for this issue. The location of the CRC (rpefix/suffix) would be a lower level issue than file version.

Another quick fix, tho I'm not sure you would like it is to change the filenames (and you could even include version info in there). That would allow to read the files properly based on the name. Something like:
- `nipost_challenge.bin` → `nipost_challenge.2.bin`
- `nipost_builder_state.bin` → `nipost_builder_state.2.bin`
- `post.bin` → `post.2.bin`

And if the names bother you, in the future once everyone is using the `.2.bin` you can rename it back to `.bin`.

fasmat:
Thanks for your contribution @0xjac!

@dshulyak  The main challenge that I see here is that `write` and `read` are used for multiple data types and files. In the long run we need a better way to generalize our persistence of data. I think `write` and `read` should take `io.Writer`/`io.Reader` and `scale.Encodable`/`scale.Decodable` instead of a file name and `[]byte`.

This also has the advantage that the caller of those functions can read/write a few bytes before passing the arguments to set or get the version or other metadata of the file or data type. This would also make changes to the file format in the future easier.

If we want to include this change we should already add a version to the new file, to avoid migrating a file twice and to improve from there. Yes, your suggestion to do it at start up keeps the scope of this change limited if we decide to include it before genesis.

@0xjac I think adding a few bytes at the beginning of the file (for now maybe just encoding 0x0001 as first version) is better than renaming the files, otherwise we might need to keep track of 2 filenames per file in the future.

0xjac:
Happy to add a version at the beginning if this is what you want but:
- `0x0001` do you want a whole 4 bytes for the version (i.e. over 4 billion possible versions)?
- How would you approach the issues of current files which do not have a version prefix? Especially on the unlikely event that the start of the content matches a known version prefix?

Overall, I would consider versioning for the content and the encoding method of the data. But CRC happens below that, once your versioned content is encoded into bytes and needs to be saved to a file. The way it should work, I think, is:
1. Split the CRC and content.
2. Compute the content's CRC
3. Compare the computed CRC with the one of the file
4. Pass the content up and only at this point look for versioning and the decoded content, etc.

Steps 1 to 3 should not really change unless the way integrity is handled changes (very unlikely). Version changes should happen if you switch from something like json to cbor or change the structure of the data.

I don't know how much of a pain rebuilding those files are. But my take would be screw it. Do CRC properly before launch as a prefix (and maybe sneak in the version there too) and if some users have to recompute some data even if it is a pain it's only for testnet and on mainnet at least it will be fine. And we can provide a small script (happy to do it) for testnet users to manually run this time to update their file so they don't have to recompute anything.

@fasmat regarding your comment on a better way to generalize our persistence of data in the long run. I agree. Happy to give it a shot after the merge if you want. I would also suggest to add a mechanism to perform some changes on updates similar to db migration but more generic as on the long run I am sure some things will change and this will be needed. (At this point, handling breaking changes of the persistence logic should be easier.)

0xjac:
@dshulyak @fasmat 

> Also looking further into it, adding a zeros prefix does not change the CRC so I can rewrite the write function to remove the initial write of the prefix.

I was wrong here. So the only advantage of having the checksum first is that the logic to read it is a bit easier. Hence, screw it. Let's keep it at the end and keep compatibility with previous format.

I've updated the code to have the checksum at the end. The only extra logic and cost is the `fstat` syscall to get the file size. This is needed to stop reading as we reach the checksum.

Since the format is the same, there is no need for versioning or anything at this point. This can be tackled later, esp. as genesis is near.

This PR only improves the efficiency of the read/write with respect to the checksum computation. Let me know what you think.

*edit 1: Updated PR description to reflect changes.*
*edit 2: The existing code was not explicitly closing the file after reading/writing. This is now fixed too.*

fasmat:
Thanks for all your effort @0xjac 🙂 

The code looks good to me and keeping compatibility with the existing implementation, should make integration easier. However, since we are so close to Genesis and this code is used in multiple places, we won't be able to test the change in time for it to be included in the release.

However I would like to merge it after Genesis and include it in the next release 🙂

fasmat (approved):

fasmat:
bors try

fasmat:
@0xjac besides the linter complaining the CI looks good. Could you run `make lint-fix` on your branch and commit the changes? 🙂

0xjac:
@fasmat Done. Sorry about that my IDE took my global linter not the local one and it missed a couple things.

Note; I've also rebased the branch on `upstream/develop`

fasmat:
bors merge

activation/nipost_state.go
```diff
@@ -20,43 +19,71 @@ const (
...
 	checksum := crc64.New(crc64.MakeTable(crc64.ISO))
-	end := len(data) - crc64Size
-	if _, err = checksum.Write(data[:end]); err != nil {
-		return nil, fmt.Errorf("read checksum %v: %w", path, err)
+	if _, err = io.TeeReader(file, checksum).Read(data); err != nil {
+		return nil, fmt.Errorf("read file %s: %w", path, err)
+	}
+
+	saved := make([]byte, crc64.Size)
+	if _, err = file.Read(saved); err != nil {
```
dshulyak, line 75: won't it fail with io unexpected eof here for existing files without crc? can we please add such test?

fasmat: The code stats the file, and first reads all data except the last 8 (`crc64.size`) bytes from the file into data. Then it reads the remaining 8 bytes into `saved` and compares if saved matches the calculated CRC.

There can only be an `io.EOF` here if the file changes size while the node is reading it 😕

dshulyak (approved):
thanks, ignore my last question i confused things

activation/nipost_state.go
0xjac: Yes, if the CRC is missing the last 8 bits will be read as CRC and then the checksum verification will fail. I'll add a quick test for that case.

If you want a test where the file is changed between the reads I can try to make one but that might be a bit trickier. Because we pass the path as a string. But if you want I can mock the filesystem for `read` and test it.

fasmat:
bors merge

fasmat:
bors merge

fasmat:
The change seems to fail because the `/tmp` directory on the nodes we use for system tests is on a different device than the path the file ultimately is written to. `atomic.ReplaceFile` fails with the following error: "rename /tmp/nipost_challenge.bin283715628 /data/post/nipost_challenge.bin: invalid cross-device link"

I assume we have to create the file we write to in the same directory as path for this to work as we expect it. Probably replacing `os.CreateTemp("", filepath.Base(path))` with `os.Create(fmt.Sprintf("%s.tmp", path))` should fix the issue.

dshulyak:
btw do we store this nipost state next to post data? if so it is reasonable to assume that tmp will be on different disk, so it is not only systest specific

0xjac:
I'm away from my laptop at the moment but I can fix it maybe tonight or Sunday

fasmat:
> btw do we store this nipost state next to post data? if so it is reasonable to assume that tmp will be on different disk, so it is not only systest specific

yes and agree 🙂

fasmat:
bors try

fasmat:
bors try

fasmat:
bors merge

PR #4700, opened 2023-07-14
[Merged by Bors] - genesis: configuration updates for v1.0.2

dshulyak:
- ledger from https://github.com/spacemeshos/wiki/wiki/Genesis-ledger
- updated max num units to max uint32, as node expects non-zero number there


pigmej (approved):
Accounts look ok.

dshulyak:
bors merge

PR #4701, opened 2023-07-14
fix: missing Highest

beck-8:
## Motivation

```bash
$ grpcurl -plaintext 127.0.0.1:9092 list spacemesh.v1.ActivationService.Highest
Error invoking method "spacemesh.v1.ActivationService.Highest": service "spacemesh.v1.ActivationService" does not include a method named "Highest"
```
`spacemesh.v1.ActivationService.Highest`This method is not exported.

## Changes

before
```bash
$ grpcurl -plaintext 127.0.0.1:9092 list spacemesh.v1.ActivationService
spacemesh.v1.ActivationService.Get
```
after
```bash
$ grpcurl -plaintext 127.0.0.1:9092 list spacemesh.v1.ActivationService
spacemesh.v1.ActivationService.Get
spacemesh.v1.ActivationService.Highest
```
## TODO

- [ ] Explain motivation or link existing issue(s)
- [ ] Test changes and document test plan
- [ ] Update documentation as needed

## DevOps Notes

- [x] This PR does not require configuration changes (e.g., environment variables, GitHub secrets, VM resources)
- [x] This PR does not affect public APIs
- [x] This PR does not rely on a new version of external services (PoET, elasticsearch, etc.)
- [x] This PR does not make changes to log messages (which monitoring infrastructure may rely on)

PR #4702, opened 2023-07-14
[Merged by Bors] - config: enable activation service by default on public endpoinnt

dshulyak:
while looking at https://github.com/spacemeshos/go-spacemesh/pull/4701 i found that "activation" wasn't
enabled by default on public endpoint


fasmat (approved):

dshulyak:
bors merge

PR #4704, opened 2023-07-14
[Merged by Bors] - cmd: beacon alarm tool to monitor boot nodes agreeing on beacon values

fasmat:
## Motivation

This tool is intended to be used to monitor if the bootnodes agree on the beacon value for the next epoch. It queries the central Mimir server for the metrics of all nodes of a given namespace and pushes its own metric to indicate if the bootnodes agree or not.

**NOTE:** The tool itself has been added to the internal tools repository. This PR now only includes some minor code changes (mostly exporting non-exported functions) that are needed for the tool to work properly.

## Description

- Add a new tool `beaconalarm` to the tools repository
- Tool starts a prometheus server that exposes a single metric: `spacemesh_beacon_beacons_alarm` that per epoch is either 0 (everything is fine) or 1 (nodes didn't agree on beacon value or failed to fetch metric)
- It connects to Mimir - our central metric server and fetches the beacon metrics of bootnodes in a given namespace
  - The server URL to connect to can be specified with `--prometheus` authentication details need to be provided via `MIMIR_ORG`, `MIMIR_USR` and `MIMIR_PWD` environment variables.
  - The namespace to monitor can be specified via command line with `--k8sNamespace`
  - The network config (required to know when nodes are supposed to agree on beacon for which epoch) can be specified with `--config`. The config file is expected to have the same layout as the config file for the node; only the "beacon", "main" and "genesis" sections are needed, the rest is ignored
- The tool starts at genesis and pulls all beacon values (one per epoch) and exposes a metric to listen to if there is or isn't agreement on it
  - After it finishes parsing all passed epochs it will wait until a new beacon value becomes available: `beacon-proposal-duration`+ `beacon-first-voting-round-duration` + (`beacon-rounds-number`-1) * (`beacon-voting-round-duration+beacon-weak-coin-round-duration` after the start of an epoch (+ 30 seconds of buffer so the metric can be scraped by prometheus / mimic before the tool tries to fetch it).

## Test Plan

- CI in tools repository

## TODO


dshulyak:
are there any advantages of keeping this tool in the node repository? i also realized that i will just move this binary out of go-spacemesh https://github.com/spacemeshos/go-spacemesh/pull/4715

fasmat:
Yeah, I think a dedicated "go-sm tools" repository makes sense for tools like this. The only issue I see is that `beaconalarm` uses code from the `go-spacemesh` repository.

I'd prefer if the tool directory does not depend on `go-spacemesh` but rather both depend on a "core" repository that contains code that is intended to be shared between multiple modules.

dshulyak:
can we please move it out of go-spacemesh? it doesn't benefit from expensive ci and has nothing to do with the node. it can import same modules externally

poszu (approved):

fasmat:
bors merge

PR #4716, opened 2023-07-17
[Merged by Bors] - post: update to latest version with inflight verification

### LINKED ISSUE #181: Verify POS initialization inflight
We could catch if initialized labels are invalid very early on when they are created and before writing to a file. This would slow down initialization (how much depends on the % of labels checked), but give a higher confidence that the generated POS is valid and allow catching problems early.

It's essential that the problem of invalid labels is communicated to users in a way that is understandable so that they can take action by themself. We should not just `panic` with the message "labels are invalid" because users will be confused.

fasmat:
## Motivation
Closes https://github.com/spacemeshos/post/issues/181

## Changes
Updates PoST with inflight verification of labels

## Test Plan
n/a

## TODO

## DevOps Notes


poszu (approved):

fasmat:
bors merge

activation/post.go
```diff
@@ -317,20 +317,27 @@ func (mgr *PostSetupManager) StartSession(ctx context.Context) error {
...
 	switch {
 	case errors.Is(err, context.Canceled):
 		mgr.logger.Info("post setup session was stopped")
 		mgr.state = PostSetupStateStopped
 		return err
+	case errors.As(err, &errLabelMismatch):
+		mgr.logger.With().Error("post setup session failed due to an issue with the initialization provider", log.Err(errLabelMismatch))
+		mgr.state = PostSetupStateError
+		events.EmitInitFailure(mgr.id, mgr.commitmentAtxId, errLabelMismatch)
+		return nil
```
poszu, line 333: Why return nil ?

fasmat: We return nil, because in the case of this error we don't want to shut down the node, just signal the error to smapp and log it.

PR #4717, opened 2023-07-17
[Merged by Bors] - use atx grading for block proposal 

countvonzero:
## Motivation

part of #4089

## Changes
- grade atxs as described in https://community.spacemesh.io/t/grading-atxs-for-the-active-set/335, let s be the start of the epoch, and δ the network propagation time.

  - grade 0/evil: ATX was received at time t >= s-3δ, or an equivocation proof was received by time s-δ.
  - grade 1/acceptable: ATX was received at time t < s-3δ before the start of the epoch, and no equivocation proof was received by time s-δ.
  - grade 2/good: ATX was received at time t < s-4δ, and no equivocation proof was received for that id until time s.

- miner only include grade 2/good ATXs in its active set for its first ballot of the epoch

- newly sync'ed node uses the first block of epoch for active set in ref ballot, since it doesn't have accurate received timestamp for atxs or malfeasance proof


miner/interface.go
```diff
@@ -11,7 +11,7 @@ import (
 //go:generate mockgen -package=miner -destination=./mocks.go -source=./interface.go
 
 type proposalOracle interface {
-	GetProposalEligibility(types.LayerID, types.Beacon, types.VRFPostIndex) (*EpochEligibility, error)
+	ProposalEligibility(types.LayerID, types.Beacon, types.VRFPostIndex) (*EpochEligibility, error)
```
fasmat, line 14: ❤️  naming

miner/util.go
```diff
@@ -0,0 +1,38 @@
...
+	"golang.org/x/exp/maps"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/sql"
+	"github.com/spacemeshos/go-spacemesh/sql/ballots"
+	"github.com/spacemeshos/go-spacemesh/sql/blocks"
+	"github.com/spacemeshos/go-spacemesh/sql/certificates"
+)
+
+func ActiveSetFromBlock(db sql.Executor, epoch types.EpochID) ([]types.ATXID, error) {
```
fasmat, line 15: I think the naming doesn't fit. It generates the ActiveSet from the first block in the provided epoch. Maybe `EpochActiveSet` or similar is a better name?

fasmat, line 15: Would it make sense to have a dedicated test for this function? It is indirectly covered by other tests, but maybe crucial enough to have a few test cases directly?

datastore/store.go
```diff
@@ -239,7 +239,7 @@ func (db *CachedDB) GetLastAtx(nodeID types.NodeID) (*types.ActivationTxHeader,
 	}
 }
 
-// GetEpochAtx gets the atx header of specified node ID in the specified epoch.
+// GetEpochAtx gets the atx header of specified node ID published in the specified epoch.
 func (db *CachedDB) GetEpochAtx(epoch types.EpochID, nodeID types.NodeID) (*types.ActivationTxHeader, error) {
```
fasmat, line 243: I think this method does caching incorrectly.

It always goes to the database and then stores the found ATX in the cache. I think it should check the cache first and then go the DB if the ATX wasn't found?

miner/oracle_test.go
```diff
@@ -17,51 +17,72 @@ import (
...
 	*Oracle
-	nodeID    types.NodeID
 	edSigner  *signing.EdSigner
 	vrfSigner *signing.VRFSigner
+	mClock    *MocklayerClock
+	mSync     *mocks.MockSyncStateProvider
 }
 
-func generateNodeIDAndSigner(tb testing.TB) (types.NodeID, *signing.EdSigner, *signing.VRFSigner) {
+func genSigners(tb testing.TB) (*signing.EdSigner, *signing.VRFSigner) {
```
fasmat, line 41: Like the change! 👍

fasmat (approved):
LGTM.

dshulyak (approved):

datastore/store.go
countvonzero: the cache is by atxid. there is no cache for the first atx id for a node. that's why it always have to do a lookup.

miner/util.go
countvonzero: EpochActiveSet means many things. it can mean the active set is derived from
- hare active set (determined by hare rules)
- tortoise active set (all epoch atx ids)
- block active set (derived from block)
- fallback (overwritten manually)

i think the naming fits here.

countvonzero: https://github.com/spacemeshos/go-spacemesh/issues/4757

countvonzero:
bors merge

countvonzero:
bors merge

miner/util.go
fasmat: It is confusing to me because instead of passing in a Block ID, I have to pass in an Epoch ID.

countvonzero: i guess i can rename it to ActiveSetFromFirstBlockInEpoch?

countvonzero:
bors merge

miner/util.go
fasmat: Yeah, I cannot think of a better name and you already stated that `EpochActiveSet` would be confusing for other reasons.

PR #4718, opened 2023-07-17
[Merged by Bors] - use atx grading for hare

### LINKED ISSUE #4089: implement hare active set

### LINKED ISSUE #4757: add test for ActiveSetFromBlock
## Description
following up on https://github.com/spacemeshos/go-spacemesh/pull/4717

countvonzero:
## Motivation

Closes #4089
Closes #4757

## Changes
- hare only vote for proposals with grade 1/acceptable and 2/good atxs in the activeset


hare/hare.go
```diff
@@ -441,9 +442,21 @@ func (h *Hare) removeCP(ctx context.Context, lid types.LayerID) {
...
 
 // goodProposals finds the "good proposals" for the specified layer. a proposal is good if
-// it has the same beacon value as the node's beacon value.
+// - it has the same beacon value as the node's beacon value.
+// - its miner is not malicious
+// - its active set contains only grade 1 or grade 2 atxs
+// see (https://community.spacemesh.io/t/grading-atxs-for-the-active-set/335#proposal-voting-4)
 // any error encountered will be ignored and an empty set is returned.
-func goodProposals(ctx context.Context, logger log.Log, msh mesh, nodeID types.NodeID, lid types.LayerID, epochBeacon types.Beacon) []types.ProposalID {
+func goodProposals(
```
fasmat, line 450: It seems that besides the changes in this file only a few tests were cleaned up compared to https://github.com/spacemeshos/go-spacemesh/pull/4717.

If this is the case the PR LGTM

fasmat (approved):

miner/atx_grader.go
```diff
@@ -0,0 +1,38 @@
...
+// grade 2: ATX was received at time t < s-4δ, and no equivocation proof was received for that id until time s.
+type AtxGrade int
+
+const (
+	Evil AtxGrade = iota
+	Acceptable
+	Good
+)
+
+func GradeAtx(msh mesh, nodeID types.NodeID, atxReceived, epochStart time.Time, delta time.Duration) (AtxGrade, error) {
```
dshulyak: this function can be more generic to make it reusable, i think there is no need for it to load malfeasance proof.
there could be one more on top that loads it and can error

`func GradeAtx(atxReceived, proofReceived, epochStart time.Time, delta time.Duration) Grade`

miner/atx_grader.go
```diff
@@ -0,0 +1,38 @@
...
+// let s be the start of the epoch, and δ the network propagation time.
+// grade 0: ATX was received at time t >= s-3δ, or an equivocation proof was received by time s-δ.
+// grade 1: ATX was received at time t < s-3δ before the start of the epoch, and no equivocation proof was received by time s-δ.
+// grade 2: ATX was received at time t < s-4δ, and no equivocation proof was received for that id until time s.
+type AtxGrade int
+
+const (
+	Evil AtxGrade = iota
+	Acceptable
+	Good
```
dshulyak: can we add two more grades? it also makes sense to call them Grade1-5 explicitly, as this Evil/Acceptable/Good won't work anymore

i don't remember deltas in those grades, but there should be a picture somewhere

dshulyak (approved):
i realized that we don't persist grades, so it doesn't matter that much if this change is 3-graded or 5-graded

miner/util.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/spacemeshos/go-spacemesh/sql/certificates"
 )
 
-func ActiveSetFromBlock(db sql.Executor, epoch types.EpochID) ([]types.ATXID, error) {
+func ActiveSetFromEpochFirstBlock(db sql.Executor, epoch types.EpochID) ([]types.ATXID, error) {
```
fasmat, line 15: ❤️

miner/util_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/datastore"
+	"github.com/spacemeshos/go-spacemesh/log/logtest"
+	"github.com/spacemeshos/go-spacemesh/signing"
+	"github.com/spacemeshos/go-spacemesh/sql"
+	"github.com/spacemeshos/go-spacemesh/sql/blocks"
+	"github.com/spacemeshos/go-spacemesh/sql/certificates"
+)
+
+func TestActiveSetFromEpochFirstBlock(t *testing.T) {
```
fasmat, line 18: Thanks for adding the test ❤️

fasmat (approved):

countvonzero:
bors merge

PR #4719, opened 2023-07-18
[Merged by Bors] - validate genesis before using

zhiqiangxu:
Currently if run `fastnet` without specifying a `genesis-time`, will panic:
```
2023-07-18T10:02:33.246+0200	INFO	00000.defaultLogger	Looking for identity file at `/home/dd/post/data/key.bin`
panic: code should have run Validate before this method
```

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -1194,6 +1194,10 @@ func (app *App) LoadOrCreateEdSigner() (*signing.EdSigner, error) {
 	filename := filepath.Join(app.Config.SMESHING.Opts.DataDir, edKeyFileName)
 	log.Info("Looking for identity file at `%v`", filename)
 
+	if err := app.Config.Genesis.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid genesis: %w", err)
+	}
+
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -1194,6 +1194,10 @@ func (app *App) LoadOrCreateEdSigner() (*signing.EdSigner, error) {
 	filename := filepath.Join(app.Config.SMESHING.Opts.DataDir, edKeyFileName)
 	log.Info("Looking for identity file at `%v`", filename)
 
+	if err := app.Config.Genesis.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid genesis: %w", err)
+	}
+
```
dshulyak: i think it is better to call Initialize before LoadOrCreateSigned in the GetCommand

zhiqiangxu: Updated accordingly.

dshulyak:
bors try

dshulyak:
bors merge

dshulyak (approved):
bors merge

PR #4721, opened 2023-07-18
[Merged by Bors] - Verify self-generated POST proofs to catch errors early

poszu:
## Motivation
Currently, if a node generated invalid POST proof (possibly because it has invalid POST data), it will learn it after Poet round ended (2 weeks) - upon publishing the ATX. It's suboptimal, we can check the proof right after it was created to give the operator more time to react.

## Changes
Verify the generated POST proofs immediately after they are created.

## Test Plan
- UT
- system test
- [x] manual test joining the mainnet

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/spacemeshos/post/proving"
 	"github.com/spacemeshos/post/shared"
-	"github.com/spacemeshos/post/verifying"
 	"go.uber.org/atomic"
 	"golang.org/x/sync/errgroup"
 
@@ -75,6 +74,7 @@ type Builder struct {
 	nipostBuilder     nipostBuilder
 	postSetupProvider postSetupProvider
 	initialPost       *types.Post
+	validator         nipostValidator
 
 	// smeshingMutex protects `StartSmeshing` and `StopSmeshing` from concurrent access
 	smeshingMutex sync.Mutex
@@ -123,6 +123,12 @@ func WithContext(ctx context.Context) BuilderOption {
 func WithPoetConfig(c PoetConfig) BuilderOption {
 	return func(b *Builder) {
 		b.poetCfg = c
+	}
+}
+
+func WithValidator(v nipostValidator) BuilderOption {
+	return func(b *Builder) {
+		b.validator = v
 	}
 }
 
@@ -264,7 +270,11 @@ func (b *Builder) run(ctx context.Context) {
 
 	if post != nil {
 		b.log.With().Info("initial post created", log.Object("post", post), log.Object("metadata", metadata))
-		if err := b.postSetupProvider.VerifyProof(ctx, post, metadata, verifying.WithPowCreator(b.nodeID.Bytes())); err != nil {
+		commitmentAtxId, err := b.postSetupProvider.CommitmentAtx()
+		if err != nil {
+			b.log.With().Panic("failed to fetch commitment ATX ID.", log.Err(err))
+		}
+		if err := b.validator.Post(ctx, types.EpochID(0), b.nodeID, commitmentAtxId, post, metadata, b.postSetupProvider.LastOpts().NumUnits); err != nil {
 			events.EmitInvalidPostProof()
 			b.log.With().Fatal("initial POST proof is invalid. Probably initialized POST data is corrupted. Please verify the data with postcli and regenerate the corrupted files.", log.Err(err))
 		}
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/spacemeshos/post/proving"
 	"github.com/spacemeshos/post/shared"
+	"github.com/spacemeshos/post/verifying"
 	"go.uber.org/atomic"
 	"golang.org/x/sync/errgroup"
 
@@ -211,6 +212,7 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 			return nil
 		case err != nil:
 			b.log.Panic("initialization failed: %v", err)
+			return err
 		}
 
 		b.run(ctx)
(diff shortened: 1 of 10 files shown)


activation/activation.go
```diff
@@ -211,6 +212,7 @@ func (b *Builder) StartSmeshing(coinbase types.Address, opts PostSetupOpts) erro
 			return nil
 		case err != nil:
 			b.log.Panic("initialization failed: %v", err)
+			return err
```
poszu, line 221: Note: this is only needed to pass the unit test that mocks `log.Panic`... (and doesn't actually panic, allowing the code to proceed)

activation/post.go
```diff
@@ -489,6 +491,34 @@ func (mgr *PostSetupManager) GenerateProof(ctx context.Context, challenge []byte
 	return p, m, nil
 }
 
+func (mgr *PostSetupManager) VerifyProof(ctx context.Context, proof *types.Post, metadata *types.PostMetadata, options ...verifying.OptionFunc) error {
```
countvonzero: what's the rationale not reusing the existing code path in `Validator.Post()`?

poszu: It was unergonomic (would need to query `postSetupProvider` anyway to obtain data) and would require more changes (`postSetupProvider` is already available in both `atxBuilder` and `nipostBuilder`). I figured that as the `PostSetupManager` already provides `func GenerateProof()` it could also provide `func VerifyProof()`.

countvonzero: but if codepath diverge, the benefit is lost. miners still only found out proof is invalid after poet round end.
would it be cleaner to pass the validator to atx builder in node.go?

activation/activation.go
```diff
@@ -254,11 +256,21 @@ func (b *Builder) SmesherID() types.NodeID {
...
 	}
 
+	if post != nil {
+		b.log.With().Info("initial post created", log.Object("post", post), log.Object("metadata", metadata))
+		if err := b.postSetupProvider.VerifyProof(ctx, post, metadata, verifying.WithPowCreator(b.nodeID.Bytes())); err != nil {
+			events.EmitInvalidPostProof()
+			b.log.With().Fatal("initial POST proof is invalid. Probably initialized POST data is corrupted. Please verify the data with postcli and regenerate the corrupted files.", log.Err(err))
+		}
+		b.initialPost = post
+	}
```
fasmat: Imo this should be part of `generateInitialPost` and it shouldn't return `post` and `metadata`

activation/activation_test.go
```diff
@@ -1104,21 +1113,36 @@ func TestBuilder_InitialProofGeneratedOnce(t *testing.T) {
...
-	tab.mpost.EXPECT().GenerateProof(gomock.Any(), shared.ZeroChallenge, gomock.Any()).Return(&types.Post{}, &types.PostMetadata{}, nil)
-	require.NoError(t, tab.generateInitialPost(context.Background()))
+	tab.mpost.EXPECT().Config().AnyTimes().Return(PostConfig{})
+	tab.mpost.EXPECT().GenerateProof(gomock.Any(), shared.ZeroChallenge, gomock.Any()).Return(&types.Post{}, &types.PostMetadata{
+		Challenge: shared.ZeroChallenge,
+	}, nil)
+	p, m, err := tab.generateInitialPost(context.Background())
+	require.NoError(t, err)
+	require.NotNil(t, p)
+	require.NotNil(t, m)
```
fasmat: All these assertions for post and metadata being `nil` or not `nil` aren't needed to verify that `generateInitialPost` behaves as expected in my opinion. The fact that `tab.mpost.GenerateProof()` wasn't called on the second call to `generateInitialPost` is already enough to proof that no additional proof was generated, i.e. `tab.initialPost` didn't change

activation/post.go
fasmat: I agree with @countvonzero there shouldn't be two code paths to verify a Post Proof

activation/activation.go
poszu: IMO not. Separating logic results in cleaner code as there are code paths in `generateInitialPost ` that return early.

fasmat: There are 2 ways `generateInitialPost` returns with `post != nil`. Once on line 310 (where I think this logging belongs) and once on line 289, where this logging is misleading:

In this case no post was created, an existing post was loaded from disk instead. This also arguably doesn't need to be verified again, because it already was before it was saved to disk. Corruption or alteration on disk is prevented by having a checksum in place that would already detect this. Imo there is no need to verify the data again and the logging is misleading.

poszu: >This also arguably doesn't need to be verified again

It needs to be ATM because the initial post had NOT been verified before saving to the disk (there are users with initial post created already). In the future, we can remove re-verifying it.

activation/post.go
poszu: Changed to use the `nipostValidator` interface.

activation/activation.go
```diff
@@ -254,11 +262,25 @@ func (b *Builder) SmesherID() types.NodeID {
...
+		b.log.With().Info("initial post created", log.Object("post", post), log.Object("metadata", metadata))
+		commitmentAtxId, err := b.postSetupProvider.CommitmentAtx()
+		if err != nil {
+			b.log.With().Panic("failed to fetch commitment ATX ID.", log.Err(err))
+		}
+		if err := b.validator.Post(ctx, types.EpochID(0), b.nodeID, commitmentAtxId, post, metadata, b.postSetupProvider.LastOpts().NumUnits); err != nil {
+			events.EmitInvalidPostProof()
+			b.log.With().Fatal("initial POST proof is invalid. Probably initialized POST data is corrupted. Please verify the data with postcli and regenerate the corrupted files.", log.Err(err))
+		}
+		b.initialPost = post
```
fasmat: How about making this a method:

```go
func (b *Builder) verifyPost(post *types.Post, metadata *types.PostMetadata) error {
	commitmentAtxId, err := b.postSetupProvider.CommitmentAtx()
	if err != nil {
		b.log.With().Panic("failed to fetch commitment ATX ID.", log.Err(err))
	}
	if err := b.validator.Post(ctx, types.EpochID(0), b.nodeID, commitmentAtxId, post, metadata, b.postSetupProvider.LastOpts().NumUnits); err != nil {
		events.EmitInvalidPostProof()
		b.log.With().Fatal("initial POST proof is invalid. Probably initialized POST data is corrupted. Please verify the data with postcli and regenerate the corrupted files.", log.Err(err))
		return err
	}
	b.initialPost = post
	return nil
}
```

and call this method on line 298:

```go
	if post, err := loadPost(b.nipostBuilder.DataDir()); err == nil {
		return b.verifyPost(post, &types.PostMetadata{
			Challenge:     shared.ZeroChallenge,
			LabelsPerUnit: b.postSetupProvider.Config().LabelsPerUnit,
		})
	}
```

and on line 320
```go
	return b.verifyPost(post, metadata)
```

then `generateInitialPost` only needs to return an error. Additionally I would log inside `generateProof` if the proof was successfully verified (or loaded from disk).

activation/activation.go
countvonzero: > It needs to be ATM because the initial post had NOT been verified before saving to the disk

nice upgrade path!
can you please file an issue to drop re-verifying the initial post?

countvonzero (approved):

fasmat (approved):
❤️  LGTM

poszu:
bors merge

activation/activation.go
poszu: Created the issue: https://github.com/spacemeshos/go-spacemesh/issues/4722

PR #4723, opened 2023-07-18
[Merged by Bors] - public metrics

dshulyak:
subset of metrics to record:
- connections
- version
- init size started / completed
- post proof time

they will be recorded with 5 first characters from p2p public key, we probably don't want to gather information about smesher key.

example config:
```json
{
    "main": {
        "metrics-push": "https://public-metrics-gateway.spacemesh.dev/",
        "metrics-push-period": "1s",
        "metrics-push-user": "XXX",
        "metrics-push-pass": "XXXXXX",
        "metrics-push-header": {
            "X-Scope-OrgId": "XXXXXXX"
        }
    }
}
```

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -54,6 +54,7 @@ import (
 	"github.com/spacemeshos/go-spacemesh/malfeasance"
 	"github.com/spacemeshos/go-spacemesh/mesh"
 	"github.com/spacemeshos/go-spacemesh/metrics"
+	"github.com/spacemeshos/go-spacemesh/metrics/public"
 	"github.com/spacemeshos/go-spacemesh/miner"
 	"github.com/spacemeshos/go-spacemesh/node/mapstructureutil"
 	"github.com/spacemeshos/go-spacemesh/p2p"
@@ -453,9 +454,9 @@ func (app *App) Initialize() error {
 	timeCfg.TimeConfigValues = app.Config.TIME
 
 	app.setupLogging()
-
 	app.introduction()
 
+	public.Version.WithLabelValues(cmd.Version).Set(1)
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


node/node.go
```diff
@@ -1374,9 +1375,13 @@ func (app *App) Start(ctx context.Context) error {
...
-		metrics.StartPushingMetrics(app.Config.MetricsPush,
-			app.Config.MetricsPushUser, app.Config.MetricsPushPass, app.Config.MetricsPushPeriod,
-			app.host.ID().String(), app.Config.Genesis.GenesisID().ShortString())
+		metrics.StartPushingMetrics(
+			app.Config.MetricsPush,
+			app.Config.MetricsPushUser,
+			app.Config.MetricsPushPass,
+			app.Config.MetricsPushHeader,
+			app.Config.MetricsPushPeriod,
+			app.host.ID().String()[:5], app.Config.Genesis.GenesisID().ShortString())
```
countvonzero, line 1384: what do you think about hashing the first 5 char?

countvonzero (approved):

node/node.go
dshulyak: maybe hash all then and use 5 chars? i can do

dshulyak: actually it doesn't hide anything, if i want i can hash all identities found in the network and match them

countvonzero: yes. but at least its a step towards anomynizing. somebody have to pay the computation to get it....

pigmej: But the same info will be publicly available in the epoch when ATX will be published, right? So why are we trying to hide it?

countvonzero: this is p2p id

dshulyak:
bors merge

PR #4725, opened 2023-07-19
[Merged by Bors] - Fix misleading error

poszu:
## Motivation
Users often get confused by the log message when the nipost_challenge.bin doesn't exist. This is a normal situation, yet the log includes "errormsg".

## Changes
Don't log an error in case the file is not found.

Additionally, includes a fix for a bug discovered during testing. The code panicked when the file was too small (for example empty).

## Test Plan
- UT
- [x] manual test

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -527,9 +527,11 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 
 	challenge, err := b.loadChallenge()
 	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			logger.With().Error("failed to load atx challenge", log.Err(err))
+		}
 		logger.With().Info("building new atx challenge",
 			log.Stringer("current_epoch", b.currentEpoch()),
-			log.Err(err),
 		)
 		challenge, err = b.buildNIPostChallenge(ctx)
 		if err != nil {
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -528,7 +528,7 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 	challenge, err := b.loadChallenge()
 	if err != nil {
 		if !errors.Is(err, os.ErrNotExist) {
-			logger.With().Error("failed to load atx challenge", log.Err(err))
+			logger.With().Warning("failed to load atx challenge", log.Err(err))
 		}
 		logger.With().Info("building new atx challenge",
 			log.Stringer("current_epoch", b.currentEpoch()),
--- a/activation/nipost_state_test.go
+++ b/activation/nipost_state_test.go
@@ -58,6 +58,16 @@ func TestReadCRC(t *testing.T) {
 			data:   []byte("spacemesh"),
 			crc:    []byte{0xC7, 0x11, 0x73, 0xE0, 0x53, 0xD8, 0x75, 0x0F},
 			errMsg: "wrong checksum 0xC71173E053D8750F, computed 0xC61072DF52D7740E",
+		}, {
+			name:   "file too short",
+			data:   []byte("123"),
+			crc:    []byte{},
+			errMsg: "too small",
+		}, {
+			name:   "file empty",
+			data:   []byte(""),
+			crc:    []byte{0xC6, 0x10, 0x72, 0xDF, 0x52, 0xD7, 0x74, 0x0E},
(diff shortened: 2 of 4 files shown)


dshulyak (approved):

activation/nipost_state_test.go
```diff
@@ -58,6 +58,16 @@ func TestReadCRC(t *testing.T) {
...
+			name:   "file too short",
+			data:   []byte("123"),
+			crc:    []byte{},
+			errMsg: "too small",
+		}, {
+			name:   "file empty",
+			data:   []byte(""),
+			crc:    []byte{0xC6, 0x10, 0x72, 0xDF, 0x52, 0xD7, 0x74, 0x0E},
+			errMsg: "wrong checksum 0xC61072DF52D7740E, computed 0x0",
 		},
```
fasmat, line 71: I'm confused by this test. The CRC64 of an empty file should be `0x0000000000000000` and if the file only contains those bytes it will pass reading & CRC validation

activation/activation.go
```diff
@@ -527,9 +527,11 @@ func (b *Builder) PublishActivationTx(ctx context.Context) error {
 
 	challenge, err := b.loadChallenge()
 	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			logger.With().Error("failed to load atx challenge", log.Err(err))
+		}
```
fasmat: I would change it to debug logging instead of only logging it when there is no file.

If it is another error (e.g. CRC validation fails) it will log the error and start building a new challenge just like if no file was found. If it would be an error I would expect a different behaviour (e.g. stopping smeshing)

activation/nipost_state_test.go
poszu: What is confusing? It's exactly as you said. CRC of nothing is 0. The test file contains a wrong CRC and no data hence it should fail with "wrong checksum 0xC61072DF52D7740E, computed 0x0" message.

activation/activation.go
poszu: I don't understand. This branch will be entered when the file exists and it is corrupted. The change is to not log when the file simply doesn't exist.

activation/nipost_state_test.go
fasmat: but isn't this just a repetition of the "badCRC" testcase?

activation/activation.go
fasmat: OK, let me rephrase: This is not an error, the node doesn't behave differently because it couldn't load the ATX challenge (it does the same no matter if there is no file or a corrupted file). Imo the appropriate logging level is Warn or Debug, because any error doesn't change anything in the flow of the program.

poszu: For sure it must not be a `Debug`. A corrupted nipost_challenge.bin is a serious thing and the user must be made aware of it. Whether Error or Warn is better, IDK. What are the guidelines here?

activation/nipost_state_test.go
poszu: No, it's an edge case when the input is empty.

poszu:
bors merge

activation/activation.go
fasmat: > A corrupted nipost_challenge.bin is a serious thing and the user must be made aware of it.

But then just logging it and deleting / overwriting it with a new challenge is probably wrong? Maybe this is serious enough to have the node stop smeshing with an error that is logged and emitted as event.

> What are the guidelines here?

We don't have a logging strategy yet. Here are my 2 cents: anything we log as `Error` must be severe enough to cause the node to deviate from its normal behavior, `Warn` logs on the other hand are issues the user should be informed about but the node can continue to operate as normal. `Fatal` are errors the node can't deal with and has to exit (should ideally never be necessary).

activation/nipost_state_test.go
fasmat: Again back to my first comment: an empty input is not an error, it has a valid checksum. This test does not check if empty input is handled correctly (that would be "" and 0x0), it checks if a wrong checksum is detected (as does the "badCRC" testcase)

poszu:
bors merge

poszu:
flake in `TestNodesUsingDifferentPoets`, unrelated to this change. 

bors merge

poszu:
bors merge

PR #4736, opened 2023-07-21
[Merged by Bors] - p2p: enable dht based disovery

dshulyak:
this makes use of "native" libp2p discovery by enabling kad dht, peers exchange (adjusting score) and relay servers. the primary reason is to benefit from hole punching and multiple transports support (quic, webrtc).

in general peers will be obtained by one of the following methods:
- getting more peers from dht. in order for peer to be added to public dht it needs to have dialable address and advertise it to the neighbors. this is pretty much analogous to our current setup, but it also supports multiple addresses so we can enable quic in next change. as it is udp based it may work on nat/firewall devices where tcp was restricted.
- peer exchange (as a part of gossipsub) will advertise peers to other nodes as part of PRUNE message, once other node sees relayed address it will initiate holepunch.  if peer is not publicly dialable it will request relay reservation from the bootnodes. once received this reservation will be communicated to bootnode as one of the peer addresses.
- additionally we will persist peers in p2p/connected.txt every 30minutes, and try to reconnect with them on startup to decrease load on bootnodes/relays. 

configuration changes:
- `p2p-bootnode` option needs to be set explicitly on bootnodes in order to start dht in server mode, and adjust gossipsub configuration
- `p2p-reachability` can be either `public` or `private`. we have to set `public` on bootnodes, otherwise autonat might be confused by out setup, beside that it is only useful for testing.
- `p2p-holepunching` will enable relay client and holepunching. enabled by default
- `p2p-disable-legacy-discovery` disables our own discovery. it can be run in parallel, but we will have no reason to keep it if all works as expected
- `p2p-private-network` disables dht filtering by public address, it is mostly useful for testing
- `relay-server` section:
     - `enable` enable relay server, has to be enabled on bootnodes
     - `reservations` number of relay reservations, 512 by default
     -  `ttl` reservation ttl, 20 minutes  

example of the p2p config section:
```json
"p2p": {
        "listen": "/ip4/0.0.0.0/tcp/5000",
        "p2p-disable-legacy-discovery": true,
        "relay-server": {
            "enable": "true"
        },
        "p2p-reachability": "public",
        "p2p-bootnode": true,
}
```


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
spacemeshos/go-spacemesh-dev:8ef576f

dshulyak:
bors try

dshulyak:
config for testing

```json
{
    "api": {
        "grpc-public-services": [],
        "grpc-private-services": []
    },
    "p2p": {
        "listen": "/ip4/0.0.0.0/tcp/5000",
        "p2p-disable-legacy-discovery": true,
        "bootnodes": [
            "/dns4/holepunch-101-bootnode-0.spacemesh.network/tcp/5000/p2p/12D3KooWQMAAL9nkgXJgTM2psJLMbWRgZLjAxJozx7dNkKYczs2V",
            "/dns4/holepunch-101-bootnode-1.spacemesh.network/tcp/5000/p2p/12D3KooWF2bhnqsnu2UjxJGZs8KCbzvd91vnud29KaG9rzTNFH78",
            "/dns4/holepunch-101-bootnode-2.spacemesh.network/tcp/5000/p2p/12D3KooWQKosE9LZraMfFPg9QRYyKwxT4ipsar37oT6wRy8WqWos",
            "/dns4/holepunch-101-bootnode-3.spacemesh.network/tcp/5000/p2p/12D3KooWKMQ5j15x2gzfXfX7uisss7uwkPnqJyoHcfyiYfHgSo3F"
        ]
    },
    "genesis": {
        "genesis-time": "2023-07-21T17:00:00.00Z",
        "genesis-extra-data": "holepunch",
        "genesis-accounts": {}
    }
}
```

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

p2p/dhtdiscovery/discovery.go
```diff
@@ -0,0 +1,235 @@
...
+				}
+			},
+		}
+		d.h.Network().Notify(notifiee)
+		ticker := time.NewTicker(d.period)
+		for {
+			select {
+			case <-ticker.C:
+			case <-disconnected:
+			case <-d.ctx.Done():
```
countvonzero, line 147: a little bit clearer to put this case as the first one.

p2p/upgrade.go
```diff
@@ -94,27 +87,55 @@ func Upgrade(h host.Host, opts ...Opt) (*Host, error) {
...
+		discovery.WithBootnodes(bootnodes),
+		discovery.WithLogger(fh.logger.Zap()),
 	}
+	if cfg.Bootnode {
+		dopts = append(dopts, discovery.Server())
+	}
+	if cfg.PrivateNetwork {
+		dopts = append(dopts, discovery.Private())
+	}
+	if !cfg.Bootnode {
```
countvonzero: combine with line 120

countvonzero (approved):
general question. when you enable legacy with dht, how do you differentiate the peers discovered via each mechanism?

dshulyak:
> general question. when you enable legacy with dht, how do you differentiate the peers discovered via each mechanism?

there is no differentiation, if two discoveries run in parallel each will use its own protocol (new - kad dht, old - peerexchange). if some of the peers don't talk one of the protocols then that discovery will skip that peer

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

dshulyak:
bors merge

countvonzero:
bors merge

PR #4746, opened 2023-07-25
Increase retries for retrieving profiler

piersy:
The git machines serving this data seem to be somewhat flaky, increasing the retries seems to help.

## DevOps Notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
     needs: filter-changes
     if: ${{ needs.filter-changes.outputs.nondocchanges == 'true' }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
@@ -137,7 +137,7 @@ jobs:
     needs: filter-changes
     if: ${{ needs.filter-changes.outputs.nondocchanges == 'true' }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
--- a/Makefile-libs.Inc
+++ b/Makefile-libs.Inc
@@ -75,10 +75,10 @@ $(BINDIR_POSTRS_SETUP_LIBS): $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
 	touch $@
 
 $(PROJ_DIR)$(POSTRS_SETUP_ZIP):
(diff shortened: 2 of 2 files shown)


fasmat:
Increasing the number of retries doesn't help against this specific error unfortunately 🙁

.github/workflows/ci.yml
```diff
@@ -99,7 +99,7 @@ jobs:
     needs: filter-changes
     if: ${{ needs.filter-changes.outputs.nondocchanges == 'true' }}
     strategy:
-      fail-fast: true
+      fail-fast: false
```
fasmat, line 102: Why the change? Especially for the `build` job failing faster should be better? 🤔

Makefile-libs.Inc
```diff
@@ -75,10 +75,10 @@ $(BINDIR_POSTRS_SETUP_LIBS): $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
 	touch $@
 
 $(PROJ_DIR)$(POSTRS_SETUP_ZIP):
-	curl -sSL --retry 5 --retry-max-time 60 $(POSTRS_SETUP_URL_ZIP) -o $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
+	wget -nv -c $(POSTRS_SETUP_URL_ZIP) -O $(PROJ_DIR)$(POSTRS_SETUP_ZIP)
```
fasmat, line 78: It's not an issue with `curl`. It rather seems like some github servers actively reset connections from our build agent to them. Without a proxy I doubt we can solve the issue.

It seems the issue persists, then we will have to move our self-hosted build agent from `unittests` to `unittests-slower`. Those are allowed to fail and don't block us from merging pull requests. Just removing `[self-hosted, linux, arm64]` from `unittests:strategy:matrix:os` and adding it to `unittests-slower:strategy:matrix:os` should suffice to work around the issue for now until we find a more permanent solution.

fasmat (changes requested):

.github/workflows/ci.yml
piersy: A temporary change to more easily see the effects on all matrix entries. Otherwise without this, if a change fails for one matrix entry you don't get to find out if it would have worked on any other entries.

Makefile-libs.Inc
piersy: This was an attempt to get retries happening since `--retry-all-errors` didn't work on all platforms for curl, wget retries 20 times by default. But it does not seem to be making a difference.

piersy:
@fasmat I'll close for now, as I'll be on vacation till the 7th August and this doesn't seem to be helping.

PR #4749, opened 2023-07-26
[Merged by Bors] - Remove global logger

fasmat:
## Motivation
 Using a global logger makes testing harder and logging configuration more complicated. This PR removes most uses of the global logger to address this.

## Changes
- Instead of the global logger services now use their own logging instance, this should reduce the number of log lines produced by `0000.defaultLogger`
- There are still a few places where the global logger is used, primarily the `common/types` package where it cannot be replaced without some bigger changes. Issue for this: https://github.com/spacemeshos/go-spacemesh/issues/4751

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/api/grpcserver/grpc.go
+++ b/api/grpcserver/grpc.go
@@ -4,11 +4,11 @@ import (
 	"net"
 	"time"
 
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/reflection"
-
-	"github.com/spacemeshos/go-spacemesh/log"
 )
 
 // ServiceAPI allows individual grpc services to register the grpc server.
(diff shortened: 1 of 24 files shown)


node/node.go
```diff
@@ -1033,6 +1029,7 @@ func (app *App) startServices(ctx context.Context) error {
 }
 
 func (app *App) initService(ctx context.Context, svc grpcserver.Service) (grpcserver.ServiceAPI, error) {
+	// TODO(mafa): add app.log.WithName("service") to all services
```
fasmat, line 1032: TODO

fasmat:
bors try

api/grpcserver/grpc.go
```diff
@@ -19,28 +19,31 @@ type ServiceAPI interface {
 // Server is a very basic grpc server.
 type Server struct {
 	Listener   string
+	logger     *zap.Logger
```
poszu: Why be inconsistent here? `log.Logger` is used everywhere else.

fasmat: Because of line 62 where a sugared logger is used. If that log can be structured as well then the `log.Logger` can be used.

poszu: `Logger::Info()` works identically as zap's sugared logger. In fact, `log.Log` uses it underneath: https://github.com/spacemeshos/go-spacemesh/blob/c4275d506b32a71dbb40ab485b1e65554db0ef6f/log/zap.go#L25-L27

fasmat: Updated

poszu (approved):

fasmat:
bors merge

fasmat:
bors merge

PR #4752, opened 2023-07-28
[Merged by Bors] - Add check for next poet round on mainnet to codebase.

### LINKED ISSUE #4745: node configuration: fetch poet proofs from all poets in epoch 1
## Description

Due to a hardware issue no node was able to submit their proof to PoET 111. This PoET has been restored with the state of PoET 110. By default nodes only query poets for proof where they were successfully able to submit their challenge to.

For the upcoming end of the first PoET round (2023-08-06 8:00 UTC) this behaviour should be changed such that all nodes also query their proofs from PoET 111 instead of only from the ones they were able to submit to.

## Implementation hints

- `getBestProof` in `nipost.go` needs to be updated to have a different behavior for the end of the current PoET Round
  - instead of iterating over all submitted requests it should iterate over all known poets listed in the configuration
- Add tests that verify that changed behavior works as expected for the current PoET round and reverts back to the current behavior after

fasmat:
## Motivation
Closes #4745 

## Changes
- When fetching a poet proof for the publish epoch 1 the state will be updated to duplicate the request for PoET 110 to PoET 111 such that the node will also try to fetch from there even if it didn't submit a proof to it.
- `NIPoSTBuilder` was updated to store clients in a map instead of in a slice for easier access.

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/activation/nipost.go
+++ b/activation/nipost.go
@@ -378,6 +378,7 @@ func membersContainChallenge(members []types.Member, challenge types.Hash32) (ui
 	return 0, fmt.Errorf("challenge is not a member of the proof")
 }
 
+// TODO(mafa): remove after next poet round; https://github.com/spacemeshos/go-spacemesh/issues/4753
 func (nb *NIPostBuilder) addPoet111ForPubEpoch1(ctx context.Context) error {
 	// because poet 111 had a hardware issue when challenges for round 0 were submitted, no node could submit to it
 	// 111 was recovered with the poet 110 DB, so all successful submissions to 110 should be able to be fetched from there as well
@@ -420,6 +421,7 @@ func (nb *NIPostBuilder) addPoet111ForPubEpoch1(ctx context.Context) error {
 }
 
 func (nb *NIPostBuilder) getBestProof(ctx context.Context, challenge types.Hash32, publishEpoch types.EpochID) (types.PoetProofRef, *types.MerkleProof, error) {
+	// TODO(mafa): remove after next poet round; https://github.com/spacemeshos/go-spacemesh/issues/4753
 	if publishEpoch == 1 {
 		err := nb.addPoet111ForPubEpoch1(ctx)
 		if err != nil {
--- a/activation/nipost.go
+++ b/activation/nipost.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spacemeshos/poet/shared"
 	"github.com/spacemeshos/post/proving"
 	"github.com/spacemeshos/post/verifying"
+	"golang.org/x/exp/slices"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/spacemeshos/go-spacemesh/activation/metrics"
@@ -28,7 +29,10 @@ import (
 // proving clients, and thus enormously reduce the cost-per-proof for PoET since each additional proof adds only
 // a small number of hash evaluations to the total cost.
 type PoetProvingServiceClient interface {
+	Address() string
+
 	PowParams(ctx context.Context) (*PoetPowParams, error)
+
 	// Submit registers a challenge in the proving service current open round.
 	Submit(ctx context.Context, prefix, challenge []byte, signature types.EdSignature, nodeID types.NodeID, pow PoetPoW) (*types.PoetRound, error)
 
@@ -64,7 +68,7 @@ type NIPostBuilder struct {
 	nodeID            types.NodeID
 	dataDir           string
 	postSetupProvider postSetupProvider
-	poetProvers       []PoetProvingServiceClient
+	poetProvers       map[string]PoetProvingServiceClient
 	poetDB            poetDbAPI
 	state             *types.NIPostBuilderState
 	log               log.Log
@@ -82,6 +86,16 @@ func WithNipostValidator(v nipostValidator) NIPostBuilderOption {
 	}
 }
 
+// withPoetClients allows to pass in clients directly (for testing purposes).
+func withPoetClients(clients []PoetProvingServiceClient) NIPostBuilderOption {
+	return func(nb *NIPostBuilder) {
+		nb.poetProvers = make(map[string]PoetProvingServiceClient, len(clients))
+		for _, client := range clients {
+			nb.poetProvers[client.Address()] = client
+		}
+	}
+}
+
 type poetDbAPI interface {
 	GetProof(types.PoetProofRef) (*types.PoetProof, *types.Hash32, error)
 	ValidateAndStore(ctx context.Context, proofMessage *types.PoetProofMessage) error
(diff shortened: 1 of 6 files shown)


fasmat:
bors try

activation/nipost.go
```diff
@@ -349,7 +378,57 @@ func membersContainChallenge(members []types.Member, challenge types.Hash32) (ui
...
+	if poet110 == -1 {
+		return fmt.Errorf("poet 110 is not in the state, cannot add poet 111")
+	}
+	poet111 := nb.state.PoetRequests[poet110]
+	poet111.PoetServiceID.ServiceID = ID111.ServiceID
+	nb.state.PoetRequests = append(nb.state.PoetRequests, poet111)
+	nb.persistState()
+	nb.log.Info("pub epoch 1 mitigation: poet 111 added to the state")
+	return nil
+}
```
poszu, line 421: Why not just query all `nb.poetProvers` instead (not the ones on `nb.state.PoetRequests`)? If we are not in the members of a poet proof, we simply discard it. The change would be much smaller and fewer places for mistakes.

node/node.go
```diff
@@ -797,18 +797,21 @@ func (app *App) initServices(ctx context.Context, poetClients []activation.PoetP
...
 		app.Config.SMESHING.Opts.DataDir,
 		app.addLogger(NipostBuilderLogger, lg),
 		app.edSgn,
 		poetCfg,
 		app.clock,
 		activation.WithNipostValidator(app.validator),
 	)
+	if err != nil {
+		app.log.Panic("failed to create nipost builder: %v", err)
+	}
```
poszu, line 814: Too bad error handling is mixed (sometimes a panic, sometimes bubble up).

activation/nipost.go
```diff
@@ -91,19 +105,28 @@ type poetDbAPI interface {
 func NewNIPostBuilder(
 	nodeID types.NodeID,
 	postSetupProvider postSetupProvider,
-	poetProvers []PoetProvingServiceClient,
 	poetDB poetDbAPI,
+	poetServers []string,
```
poszu, line 109: Why changed to `string`? It allowed passing a mocked instance in tests nicely. Now there are 2 ways of passing poet provers and in tests, you had to pass an empty list explicitly. What's the benefit?

activation/nipost.go
```diff
@@ -91,19 +105,28 @@ type poetDbAPI interface {
...
 	poetCfg PoetConfig,
 	layerClock layerClock,
 	opts ...NIPostBuilderOption,
-) *NIPostBuilder {
+) (*NIPostBuilder, error) {
+	poetClients := make(map[string]PoetProvingServiceClient, len(poetServers))
+	for _, address := range poetServers {
+		client, err := NewHTTPPoetClient(address, poetCfg)
+		if err != nil {
+			return nil, fmt.Errorf("cannot create poet client: %w", err)
```
countvonzero, line 121: curious. what's the benefit of parsing here vs parsing in node.go?

activation/nipost.go
```diff
@@ -349,7 +378,57 @@ func membersContainChallenge(members []types.Member, challenge types.Hash32) (ui
...
+	// 111 was recovered with the poet 110 DB, so all successful submissions to 110 should be able to be fetched from there as well
+
+	client111, ok := nb.poetProvers["https://poet-111.spacemesh.network"]
+	if !ok {
+		// poet 111 is not in the list, no action necessary
+		return nil
+	}
+
+	nb.log.Info("pub epoch 1 mitigation: poet 111 is in the list of poets, adding it to the state as well")
+	client110 := nb.poetProvers["https://poet-110.spacemesh.network"]
```
countvonzero, line 393: why add poet 110 as well?

activation/nipost.go
```diff
@@ -349,7 +378,57 @@ func membersContainChallenge(members []types.Member, challenge types.Hash32) (ui
 	return 0, fmt.Errorf("challenge is not a member of the proof")
 }
 
-func (nb *NIPostBuilder) getBestProof(ctx context.Context, challenge types.Hash32) (types.PoetProofRef, *types.MerkleProof, error) {
+// TODO(mafa): remove after next poet round; https://github.com/spacemeshos/go-spacemesh/issues/4753
```
countvonzero, line 381: i thought the whole point is simply querying every poet in the config? 
that way there is no need to hard-code anything as long as 111 is in the config?

node/node.go
fasmat: I went for panic because the other services immediately proceeding also panic on error. It also seems like the correct response to me (without the service being able to be initialized it doesn't make sense to continue and just exist with the error).

activation/nipost.go
fasmat: I tried the way you describe and the change looked bigger and more involved (i.e. changes in multiple places instead of just one). This way it is literally only the 40 lines you highlighted.

Going the route of not looking at `nb.state.PoetRequests` but hardcode to query all would require changes in multiple places and also runs the risk of all nodes (even ones that didn't submit) to query the PoET when time comes, which will put a much bigger load on the server.

This way the change is contained to this one function, that updates the state. It also only causes nodes to fetch from PoET 111 if they were previously able to submit to PoET 110 since otherwise they wouldn't be contained in the returned proof anyway and would just needlessly cause load on the PoET.

activation/nipost.go
fasmat: The service now receives the configuration and decides by itself what to do with it. Imo it is not correct that the node initializes http clients and passes them to the service, when the service should do that itself.

There is still only one way to pass PoET provers: via this parameter. The second one is a functional option that is not exported and thereby only usable in tests (a pattern that we use in other services as well).

activation/nipost.go
fasmat: Imo the node doesn't need to know about the dependencies of the `NIPoSTBuilder`. It passes in the configuration to the service and the service does with it what is needed. This way `newHTTPPoetClient` could even be private and the interface for it as well, since it is not needed outside the `activation` package.

activation/nipost.go
fasmat: PoET 110 isn't added.

The logic is as follows: if the configuration contains both 110 and 111, but the state only contains 110 duplicate the state of 110 for 111.

activation/nipost.go
fasmat: Again as I explained in https://github.com/spacemeshos/go-spacemesh/pull/4752/files/466e168e9486621f54a7675258ca21676f828885#r1278978871

The change to just query every poet in the config is a) a bigger change than the current solution and b) would cause more unnecessary load on the system.

Only clients that submitted to 110 for round 0 should fetch from 111 when the time comes, since only those will even be able to receive a response.

activation/nipost.go
countvonzero: mixing critical change with non-critical change simply slow down the review process.

activation/nipost.go
countvonzero: does it really matter? doesn't adding 111 only suffice?

countvonzero (commented):
will leave it to @poszu to do the final approval. thanks

activation/nipost.go
fasmat: I can't follow. PoET 111 was restored with a copy of the state of PoET 110, someone who didn't submit to 110 won't be able to receive a proof from 111?

activation/nipost.go
fasmat: I had to touch the code in question anyway. The field `poetProvers` was updated from `[]PoetProvers` to `map[string]PoetProvers` to be able to get one of the entries by Address (which wasn't possible before).

So either I add this code to iterate over `[]PoetProvers` call `Address()` on them and add them to the map, or I initialize them here directly. Seems like a related change to me 🤷.

activation/nipost.go
poszu: I'm not sure, it seems to be a step back. The API lost flexibility as now the HTTP interface to Poets is hardcoded in the NipostBuilder (previously it could be any client).

poszu (approved):

fasmat:
bors merge

fasmat:
bors cancel

fasmat:
bors merge

Issue #4604, opened 2023-06-27
malfeasance proof for proposals


countvonzero:
## Description
when multiple proposals enclose the same ballot, the smesher should be considered malicious and therefore a malfeasance proof should be generated as a result. if not, adversaries can trivially bombard nodes with multiple Proposal gossips like this.

please refer to https://github.com/spacemeshos/go-spacemesh/issues/3915 for 
- [x] https://github.com/spacemeshos/go-spacemesh/pull/4759 
- detecting / verifying / saving / regossip malfeasance proof

countvonzero:
this is added for genesis only because it changes Proposal data structure.

more accurately, it changes the way proposal data is signed.

dshulyak:
original issue https://github.com/spacemeshos/go-spacemesh/issues/4510

countvonzero:
see also https://github.com/spacemeshos/go-spacemesh/issues/4632 for handling malicious atxs and ballots

PR #4759, opened 2023-07-31
[Merged by Bors] - proposal: sign proposal data like ballot

countvonzero:
## Motivation
part of #4604

the minimum change required before 1st block


countvonzero:
@dshulyak 

do you think we should sign Proposal in a different domain? currently it's signed with `signing.Ballot` domain.

dshulyak:
it seems possible to construct malfeasance proof using proposal and ballot signature for same layer and it will be seen as valid. below is a valid malfeasance proof right?

proposal : layer || proposal hash || smesher || signature
ballot      : layer || ballot hash      || smesher || signature

because signature will be valid but msg hash is different. if they are in different domain there is no such problem

countvonzero:
> it seems possible to construct malfeasance proof using proposal and ballot signature for same layer and it will be seen as valid. below is a valid malfeasance proof right?
> 
> proposal : layer || proposal hash || smesher || signature ballot : layer || ballot hash || smesher || signature
> 
> because signature will be valid but msg hash is different. if they are in different domain there is no such problem

yes. agreed. thank for confirming.

countvonzero:
bors try

signing/signer.go
```diff
@@ -18,9 +18,10 @@ type Domain byte
 const (
 	ATX Domain = 0
 
-	BALLOT = 2
-	HARE   = 3
-	POET   = 4
+	PROPOSAL = 1
```
dshulyak, line 21: looks strange that we didn't use 1 before

dshulyak (approved):

signing/signer.go
countvonzero: @fasmat do you know why? 

maybe unrelated: 
staticcheck complains
`SA9004: only the first constant in this group has an explicit type (staticcheck)` 
if i don't leave a empty line.

dshulyak: i think we had beacon as 1, and then @fasmat implemented change to split it into different domains

countvonzero:
bors merge

PR #4765, opened 2023-08-02
[Merged by Bors] - hare3 implementation

dshulyak:
implementation is splitted into several files:

#### protocol.go
implements protocol execution logic close to the original paper, baring several differences in syntax and things that we need but are not covered in the paper. interface for interacting with protocol has two methods:

- onMessage
- next(active bool)

only active nodes are publishing messages

#### types.go
types that go over the wire

```go
type Message struct {
	Body
	Sender    types.NodeID
	Signature types.EdSignature
}

type Body struct {
	Layer types.LayerID
	IterRound
	Value       Value
	Eligibility types.HareEligibility
}

type Value struct {
	// Proposals is set in messages for preround and propose rounds.
	Proposals []types.ProposalID `scale:"max=200"`
	// Reference is set in messages for commit and notify rounds.
	Reference *types.Hash32
}
```

signature domain and malfeasance proof remains the same, as we don't intend to run more than one hare in parallel.

#### hare.go
integrates protocol with spacemesh data model, network, synchronization. manages multiple protocol instances (one per layer).

#### legacy_oracle.go
wrapper for oracle that uses active set in consensus, it outputs maximum grades for identities that are in consensus and 0 for everyone else.

#### compat
intergration for weakcoin and hare results to avoid changes across the app

remaining:
- [ ] test commit with different locked value in protocol_test.go (remaining for 100%)


dshulyak:
bors try

dshulyak:
image

tx count in 21 and 22 layers is a bit off, might be mistimed somehow

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
```
     poets_test.go:146: 
        	Error Trace:	/build/systest/tests/poets_test.go:146
        	Error:      	Received unexpected error:
        	            	rpc error: code = Unavailable desc = can't reach poet service (poet service is unstable:  (failed to query poet 'http://poet-1:80' for ID: getting poet ID: doing request: GET http://poet-1:80/v1/info giving up after 1 attempt(s): Get "http://poet-1:80/v1/info": dial tcp 10.20.57.19:80: connect: connection refused)). retry later
```

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
this change can be reviewed, i may add more unit tests but otherwise it is good. the problem that we discussed on a call is not applicable because of the timings when threshold gossip is called. i will followup in the slack

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

hare3/types.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (m *Message) ToMetadata() types.HareMetadata {
+	return types.HareMetadata{
+		Layer:   m.Layer,
+		Round:   m.Absolute(),
+		MsgHash: m.ToHash(),
+	}
+}
+
+func (m *Message) ToMalfeasenceProof() types.HareProofMsg {
```
countvonzero: :s/Malfeasence/Malfeasance/

hare3/types.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (m *Message) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddUint32("lid", m.Layer.Uint32())
+	encoder.AddUint8("iter", m.Iter)
+	encoder.AddString("round", m.Round.String())
+	encoder.AddString("sender", m.Sender.ShortString())
+	if m.Value.Proposals != nil {
+		encoder.AddArray("full", zapcore.ArrayMarshalerFunc(func(encoder log.ArrayEncoder) error {
+			for _, id := range m.Value.Proposals {
+				encoder.AppendString(types.Hash20(id).ShortString())
```
countvonzero, line 140: throughout the codebase we do id.ShortString() (as Hash32).
here the output is shorter, but might be more difficult for debugging if we search loki for the hash32 string.

hare3/types.go
```diff
@@ -0,0 +1,147 @@
...
+		IterRound: m.IterRound,
+	}
+}
+
+func (m *Message) ToBytes() []byte {
+	return codec.MustEncode(m)
+}
+
+func (m *Message) Validate() error {
+	if (m.Round == commit || m.Round == notify) && m.Value.Reference == nil {
```
countvonzero, line 124: how about when reference is set outside commit/notify?

node/node.go
```diff
@@ -763,6 +766,24 @@ func (app *App) initServices(ctx context.Context) error {
 		tortoiseWeakCoin{db: app.cachedDB, tortoise: trtl},
 		app.addLogger(HareLogger, lg),
 	)
+	if app.Config.HARE3.Enable {
+		logger := app.addLogger(HareLogger, lg).Zap()
+		app.hare3 = hare3.New(
+			app.clock, app.host, app.cachedDB, app.edVerifier, app.edSgn, app.hOracle, newSyncer,
+			hare3.WithLogger(logger),
+			hare3.WithConfig(app.Config.HARE3),
+			hare3.WithEnableLayer(types.GetEffectiveGenesis()),
```
countvonzero: fix for before merge

hare3/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+		"validation_errors",
+		namespace,
+		"number of validation errors. not expected to be at zero",
+		[]string{"error"},
+	)
+	notRegisteredError = validationError.WithLabelValues("not_registered")
+	malformedError     = validationError.WithLabelValues("malformed")
+	signatureError     = validationError.WithLabelValues("signature")
+	oracleError        = validationError.WithLabelValues("oracle")
+	maliciuosError     = validationError.WithLabelValues("malicious")
```
countvonzero: :s/maliciuos/malicious/

hare3/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+
+	droppedMessages = metrics.NewCounter(
+		"dropped_msgs",
+		namespace,
+		"number of messages dropped by gossip",
+		[]string{},
+	).WithLabelValues()
+
+	validationLatency = metrics.NewHistogramWithBuckets(
+		"validation_seconds",
```
countvonzero, line 49: why force the resolution of time at this level? 
seems more appropriate to do at prometheus in case the duration is improved to sub-second?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+	resp, err := session.submit(&input{
+		Message:   msg,
+		msgHash:   msg.ToHash(),
+		malicious: malicious,
+		atxgrade:  g,
+	})
+	if err != nil {
+		return err
+	}
+	submitLatency.Observe(time.Since(start).Seconds())
```
countvonzero, line 286: is this meant to include time for validation?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+
+type Config struct {
+	Enable          bool          `mapstructure:"enable"`
+	EnableAfter     types.LayerID `mapstructure:"enable-layer"`
+	Committee       uint16        `mapstructure:"committee"`
+	Leaders         uint16        `mapstructure:"leaders"`
+	IterationsLimit uint8         `mapstructure:"iterations-limit"`
+	PreroundDelay   time.Duration `mapstructure:"preround-delay"`
+	RoundDuration   time.Duration `mapstructure:"round-duration"`
+	ProtocolName    string
```
countvonzero, line 41: is there a reason why this should be configurable?
this makes checking prometheus metrics difficult if diff network has diff value (testnet vs mainnet...)

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+
+// roundStart returns expected time for iter/round relative to
+// layer start.
+func (cfg *Config) roundStart(round IterRound) time.Duration {
+	duration := cfg.PreroundDelay
+	if round.Round == preround {
+		return duration
+	}
+	delay := time.Duration(round.Delay(IterRound{Round: preround})) - 1
+	return duration + cfg.RoundDuration*delay
```
countvonzero: `cfg.PreroundDelay + round.Absolute() * cfg.RoundDuration`?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+}
+
+func DefaultConfig() Config {
+	return Config{
+		Committee:       800,
+		Leaders:         10,
+		IterationsLimit: 40,
```
countvonzero, line 79: should there be a limit on how many layers/sessions can run concurrently (especially when iteration limit is raised to 40)?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+}
+
+func (h *Hare) Coins() <-chan WeakCoinOutput {
+	return h.coins
+}
+
+func (h *Hare) Start() {
+	h.log.Info("started", zap.Inline(&h.config))
+	h.eg.Go(func() error {
+		h.pubsub.Register(h.config.ProtocolName, h.Handler)
```
countvonzero: you want to start getting hare messages even before the node is sync'ed?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+	delay := time.Duration(round.Delay(IterRound{Round: preround})) - 1
+	return duration + cfg.RoundDuration*delay
+}
+
+func DefaultConfig() Config {
+	return Config{
+		Committee:       800,
+		Leaders:         10,
+		IterationsLimit: 40,
+		PreroundDelay:   25 * time.Second,
```
countvonzero, line 80: `hare config's WakeupDelta` is used in different places (miner and activation) as the expected msg delivery time.
it's now set the same (25 sec) so no problems. FYI just in case you want to shorten this value for hare3

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+)
+
+type Config struct {
+	Enable          bool          `mapstructure:"enable"`
+	EnableAfter     types.LayerID `mapstructure:"enable-layer"`
+	Committee       uint16        `mapstructure:"committee"`
+	Leaders         uint16        `mapstructure:"leaders"`
+	IterationsLimit uint8         `mapstructure:"iterations-limit"`
+	PreroundDelay   time.Duration `mapstructure:"preround-delay"`
+	RoundDuration   time.Duration `mapstructure:"round-duration"`
```
countvonzero, line 40: https://github.com/spacemeshos/go-spacemesh/blob/391cb83ba602b0b785c8a8d2302684c9f54bbeef/node/node.go#L444

we need to do a similar sanity check against tortoise's zdist.

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+		malformedError.Inc()
+		return fmt.Errorf("%w: validation %s", pubsub.ErrValidationReject, err.Error())
+	}
+	h.tracer.OnMessageReceived(msg)
+	h.mu.Lock()
+	session, registered := h.sessions[msg.Layer]
+	h.mu.Unlock()
+	if !registered {
+		notRegisteredError.Inc()
+		return fmt.Errorf("layer %d is not registered", msg.Layer)
```
countvonzero, line 256: i think we should handle early messages. seems important to get as many preround messages as possible?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+
+func (h *Hare) Start() {
+	h.log.Info("started", zap.Inline(&h.config))
+	h.eg.Go(func() error {
+		h.pubsub.Register(h.config.ProtocolName, h.Handler)
+		enabled := types.MaxLayer(h.nodeclock.CurrentLayer(), h.config.EnableAfter)
+		h.log.Debug("starting at layer", zap.Uint32("lid", enabled.Uint32()))
+		for next := enabled + 1; ; next++ {
+			select {
+			case <-h.nodeclock.AwaitLayer(next):
```
countvonzero, line 223: isn't it possible that AwaitLayer() notify too late? if preround delay has expired based on current time, there is no point starting hare.

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+
+func (h *Hare) run(layer types.LayerID, beacon types.Beacon, inputs <-chan *sessionInput) error {
+	// oracle may load non-negligible amount of data from disk
+	// we do it before preround starts, so that load can have some slack time
+	// before it needs to be used in validation
+	current := IterRound{Round: preround}
+
+	start := time.Now()
+	vrf := h.oracle.active(h.signer.NodeID(), layer, current)
+	activeLatency.Observe(time.Since(start).Seconds())
```
countvonzero, line 361: it's possible that Handler() cause the active set to calculated first due to early message.
why not record this in legacy_oracle instead?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+		}
+		return []types.ProposalID{}
+	}
+	var (
+		beacon types.Beacon
+		result []types.ProposalID
+	)
+	own, err := h.db.GetEpochAtx(lid.GetEpoch()-1, h.signer.NodeID())
+	if err != nil && !errors.Is(err, sql.ErrNotFound) {
+		h.log.Warn("no atxs in the requisted epoch",
```
countvonzero: :s/requisted/requested/

countvonzero: the handling here doesn't seem correct.
whether it's ErrNotFound or not, as long as there is an error, the code should not proceed.

for ErrNotFound, the miner has to be active in the preround. it's a real error if we cannot find its atx from last epoch.

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+			h.log.Warn("proposal with same atx added several times in the recorded set",
+				zap.Int("n", n),
+				zap.Stringer("id", p.ID()),
+				zap.Stringer("atxid", p.AtxID),
+			)
+			continue
+		}
+		if own != nil {
+			hdr, err := h.db.GetAtxHeader(p.AtxID)
+			if err != nil {
```
countvonzero: log error? otherwise we will silently propose empty set.
i think it's more robust to to make `Proposals()` return error as well. it should give up on the active role on error.

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+		}
+		if n := atxs[p.AtxID]; n > 1 {
+			h.log.Warn("proposal with same atx added several times in the recorded set",
+				zap.Int("n", n),
+				zap.Stringer("id", p.ID()),
+				zap.Stringer("atxid", p.AtxID),
+			)
+			continue
+		}
+		if own != nil {
```
countvonzero: this has to be a programming error. own cannot be nil here

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+		case <-h.ctx.Done():
+			return h.ctx.Err()
+		case h.results <- ConsensusOutput{Layer: layer, Proposals: out.result}:
+		}
+		sessionResult.Inc()
+	}
+	return nil
+}
+
+func (h *Hare) Proposals(lid types.LayerID, epochBeacon types.Beacon) []types.ProposalID {
```
countvonzero: why make this public API?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+		return fmt.Errorf("%w: invalid signature", pubsub.ErrValidationReject)
+	}
+	malicious, err := h.db.IsMalicious(msg.Sender)
+	if err != nil {
+		maliciuosError.Inc()
+		return fmt.Errorf("database error %s", err.Error())
+	}
+	start := time.Now()
+	g := h.oracle.validate(msg)
+	oracleLatency.Observe(time.Since(start).Seconds())
```
countvonzero, line 269: do you expect this to be different from `activeLatency`? i thought they would be doing almost the same thing. so depending on which goroutine caused the active set to be calculated first, the value will be very wildly skewed.

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+	h.tracer.OnActive(vrf)
+
+	walltime := h.nodeclock.LayerToTime(layer).Add(h.config.PreroundDelay)
+	var proposals []types.ProposalID
+	if vrf != nil {
+		h.log.Debug("active in preround", zap.Uint32("lid", layer.Uint32()))
+		// initial set is not needed if node is not active in preround
+		select {
+		case <-h.wallclock.After(h.wallclock.Until(walltime)):
+		case <-h.ctx.Done():
```
countvonzero, line 370: abort?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+		proposalsLatency.Observe(time.Since(start).Seconds())
+	}
+	proto := newProtocol(proposals, h.config.Committee/2+1)
+	if err := h.onOutput(layer, current, proto.next(vrf != nil), vrf); err != nil {
+		return err
+	}
+	h.log.Debug("ready to accept messages", zap.Uint32("lid", layer.Uint32()))
+	walltime = walltime.Add(h.config.RoundDuration)
+	for {
+		select {
```
countvonzero, line 384: i wonder if processing inputs and the wait of executing the next round should be in the same select block.
if both cases are ready to go, select will randomly select the case. so isn't possible that the input processing could delay the execution of the next round?

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+		zap.Bool("active", vrf != nil),
+	)
+	if out.message != nil {
+		h.eg.Go(func() error {
+			out.message.Layer = layer
+			out.message.Eligibility = *vrf
+			out.message.Sender = h.signer.NodeID()
+			out.message.Signature = h.signer.Sign(signing.HARE, out.message.ToMetadata().ToBytes())
+			if err := h.pubsub.Publish(h.ctx, h.config.ProtocolName, out.message.ToBytes()); err != nil {
+				h.log.Error("failed to publish", zap.Inline(out.message))
```
countvonzero: log `err`

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+			zap.Uint32("epoch", layer.GetEpoch().Uint32()),
+			zap.Uint32("lid", layer.Uint32()),
+			zap.Error(err),
+		)
+		return
+	}
+	inputs := make(chan *sessionInput)
+	ctx, cancel := context.WithCancel(h.ctx)
+	h.mu.Lock()
+	h.sessions[layer] = sessionInputs{
```
countvonzero: syncer needs to coordinate with hare such that when hare is working on a layer, syncer doesn't try to advance the state by calling mesh.ProcessLayer(). the logic now is in layer patrol.

hare3/legacy_oracle.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type legacyOracle struct {
+	log    *zap.Logger
+	oracle oracle
+	config Config
+}
+
+func (lg *legacyOracle) validate(msg *Message) grade {
+	if msg.Eligibility.Count == 0 {
```
countvonzero, line 26: is it valid to send msg with count == 0? it's almost malicious to flood gossip like this...

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+	return types.CalcProposalHash32Presorted(proposals, nil)
+}
+
+type messageKey struct {
+	IterRound
+	Sender types.NodeID
+}
+
+type input struct {
+	*Message
```
countvonzero, line 47: when can this be nil?

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+		p.validValues[g] = p.thresholdGossip.output(IterRound{Round: preround}, g)
+	}
+	if p.Round == preround && active {
+		out.message = &Message{Body: Body{
+			IterRound: p.IterRound,
+			Value:     Value{Proposals: p.initial},
+		}}
+	} else if p.Round == hardlock && p.Iter > 0 {
+		if p.result != nil {
+			out.terminated = true
```
countvonzero, line 165: so best case `out.terminated` is set to true at iteration 1 round 1.
and the session will end as soon as next() returns. 

i thought it still needs to participate in gossip for 1 more iteration?

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+					Value:     Value{Reference: ref},
+				}}
+			}
+		}
+	} else if p.Round == notify && active {
+		ref := p.result
+		if ref == nil {
+			ref, _ = p.thresholdProposals(IterRound{Iter: p.Iter, Round: commit}, grade5)
+		}
+		if ref != nil {
```
countvonzero, line 255: i had assume that hare can output result here already and signal ready for block generation?
the paper also indicated that Pi terminates in case 1 of notify round.

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+func (p *protocol) next(active bool) output {
+	out := output{}
+	p.execution(&out, active)
+	if p.Round >= softlock && p.coin != nil && !p.coinout {
+		coin := p.coin.LSB() != 0
+		out.coin = &coin
+		p.coinout = true
+	}
+	if p.Round == preround && p.Iter == 0 {
+		p.Round = softlock // skips hardlock
```
countvonzero: does the implementation deliberately change from skipping softlock to skipping hardlock for the first iteration?
if so, can you please document the diff from paper somewhere, along with the fact that round number starts from 0 instead of -1?

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+}
+
+func (p *protocol) next(active bool) output {
+	out := output{}
+	p.execution(&out, active)
+	if p.Round >= softlock && p.coin != nil && !p.coinout {
+		coin := p.coin.LSB() != 0
+		out.coin = &coin
+		p.coinout = true
```
countvonzero, line 273: intuitively the protocol can output the coin as soon as preround ends, and false if there are no preround msgs.
is it because you want to wait for late preround messages?

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+	malicious     bool
+	values        []types.ProposalID
+	vrf           types.VrfSignature
+}
+
+func (g *gradecast) receive(current IterRound, input *input) {
+	if stored, exist := g.state[input.key()]; !exist {
+		g.state[input.key()] = &gradecasted{
+			grade:     input.atxgrade,
+			received:  current,
```
countvonzero: it seems received should be part of input and set in `hare.Handler()`?

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+		},
+	}
+}
+
+type protocol struct {
+	IterRound
+	coinout         bool
+	coin            *types.VrfSignature // smallest vrf from preround messages. not a part of paper
+	initial         []types.ProposalID  // Si
+	result          *types.Hash32       // set after Round 6. Case 1
```
countvonzero: the comment here says `Round 6`. it's unclear to me which round you are referring to, as the round number in paper is different. also in code it's set in round hardlock, tho i think it can be set in notify round (and output)

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+	return duration + cfg.RoundDuration*delay
+}
+
+func DefaultConfig() Config {
+	return Config{
+		Committee:       800,
+		Leaders:         10,
+		IterationsLimit: 40,
+		PreroundDelay:   25 * time.Second,
+		RoundDuration:   10 * time.Second,
```
countvonzero, line 81: i don't understand why round duration is shorter than communication delay (PreroundDelay). 
is it because you expect the hare messages to be much smaller than block proposals?

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+
+// Protocol 1. graded-gossip. page 10.
+type gradedGossip struct {
+	state map[messageKey]*input
+}
+
+func (g *gradedGossip) receive(input *input) (bool, *types.HareProof) {
+	// Case 1: will be discarded earlier
+	other, exist := g.state[input.key()]
+	if exist {
```
countvonzero, line 304: in protocol 1, the description is Bvk, sid = 1. here `exist` only means Bvk,ir = 1 (ir -> iterround)

hare3/hare.go
```diff
@@ -0,0 +1,571 @@
...
+	if resp.equivocation != nil && !malicious {
+		h.log.Debug("registered equivocation",
+			zap.Uint32("lid", msg.Layer.Uint32()),
+			zap.Stringer("sender", resp.equivocation.Messages[0].SmesherID))
+		proof := resp.equivocation.ToMalfeasenceProof()
+		if err := identities.SetMalicious(
+			h.db, resp.equivocation.Messages[0].SmesherID, codec.MustEncode(proof), time.Now()); err != nil {
+			h.log.Error("failed to save malicious identity", zap.Error(err))
+		}
+		h.db.CacheMalfeasanceProof(resp.equivocation.Messages[0].SmesherID, proof)
```
countvonzero: you don't gossip the malfeasance proof?

hare3/protocol.go
```diff
@@ -0,0 +1,451 @@
...
+	gossip, equivocation := p.gradedGossip.receive(msg)
+	if !gossip {
+		return false, equivocation
+	}
+	// gradecast and thresholdGossip should never be called with non-equivocating duplicates
+	if msg.Round == propose {
+		p.gradecast.receive(p.IterRound, msg)
+	} else {
+		p.thresholdGossip.receive(msg)
+	}
```
countvonzero: in the paper, both gradecast and thresholdGossip operate on the output of gradedGossip. here you broke the data into two distinct state. i don't understand this.

i had imagined that upon invocation of gradecast and thresholdGossip, each sub protocol will just take the state of gradedGossip as input and output the result.

countvonzero (commented):
this is the first part of the review. sorry for the many questions.

i've yet to understand the core of gradecast and thresholdGossip (protocol.go), from the paper and from code. will follow up with more feedback on those.

in general there should be a high-level document that describes how hare 3 works as implemented. the paper is mainly for academic purpose and doesn't map easily (at least for me) to the implementation.

for example, the input/output to the overall protocols and each sub-protocol, the information flow, the responsibility of each component. this document should be specific to the spacemesh protocol (as opposed to the general Byzantine Agreement in the paper). 
 
i understand there is an urgency for deployment and that such document takes time. we can discuss how to go about creating such doc after the PR is merged.

hare3/legacy_oracle.go
dshulyak: this is why it is dropped with grade0, as syntactically invalid message. adding malfeasence proof for syntactically invalid message is not a good idea

hare3/hare.go
dshulyak: this limit is based on iterations limit, why would i add another number?

hare3/hare.go
dshulyak: it shouldn't be used in miner and activation, i dont know why you decided to use it there. propagation for activation is not the same as ballot propagation

hare3/hare.go
dshulyak: yes it depends on the message size and expected amount of work

hare3/hare.go
dshulyak: i think there is a misunderstanding. we are getting hare when node is not synced, they are simply dropped. in this code they will be dropped as well, because session is not registered. and layer won't be registered because node is not synced

hare3/hare.go
dshulyak: it is definitely not expected, if it happens thats likely a bug in the code, or cpu is overloaded to a point that goroutines are not scheduled  

> if preround delay has expired based on current time, there is no point starting hare.

this is not true, you will finish hare successfully even if wake up notify is 5-10s late.

hare3/hare.go
dshulyak: i don't think that this code needs any special handling. preround message sent 25s seconds into the layer, session is registered when layer starts. nodes that are misconfigured by 25s will have no impact on the consensus

any party that sends out preround before layer started, will have empty preround, which is the same as not receiving any. if it sends preround 20s earlier then everyone else, thats 5s to register a layer. if event doesn't fire in 5s, thats 100% broken software or all cores are completely saturated with higher priority work

hare3/hare.go
dshulyak: data is expected to be loaded on first call to active. beside that active and vrf check doesn't run exactly the same code. after i am done with graded oracle there will be no data loaded at all. so this metric would accurately reflect cost of different operations

hare3/hare.go
dshulyak: this is meant to include time that it takes for protocol to process this message. e.g synchronization and actual work during protocol.onMessage

hare3/hare.go
dshulyak: i don't. any online node will save it from 2 messages. offline node will have to get it from sync

hare3/hare.go
dshulyak: i don't understand what it has to do with this code. that logic is not affected. after hare3 outputs result - block generator will do its own thing

hare3/hare.go
dshulyak: the message will have to be **very** early. in general it is possible, but as i explained above there won't be data loading on this or that code path.

metrics are separate mainly because i need to know how expensive calls are in consensus goroutine

hare3/hare.go
dshulyak: all expensive computation for input processing is moved to the handler. practically msg processing is reduced to single map get, and assignment

btw what you are describing is not specific to this approach. scheduling is generally can be unfair. if there is a way to overload with messages (and there is no such way, because of eligibility) then round can be delayed, because we have to validate tons of signatures for example

hare3/hare.go
dshulyak: i actually copied this part from function in hare. i assumed that you wrote it this way so that it works for inactive smesher. but in my case this code won't be even executed for inactive, so i can change

hare3/hare.go
dshulyak: see comment above

hare3/metrics.go
dshulyak: in case of subsecond it will be displayed as normal float 0.001 for example. nanoseconds are quite hard to see, with fraction of seconds is more usable for me

hare3/protocol.go
dshulyak: it is nil because *Message is allocated on heap earlier, before input is created

hare3/protocol.go
dshulyak: > i had imagined that upon invocation of gradecast and thresholdGossip, each sub protocol will just take the state of gradedGossip as input and output the result.

so in practice it can be reduced to single map instead of 3 different? should be possible, will try that

hare3/protocol.go
dshulyak: result is set in iteration 1 / round 1. terminated is set when nodes stops participating in gossip (best case in iteration 2)

hare3/protocol.go
dshulyak: yes, coin is output as early as you said. but if preround messages are delayed it is output later when atleast one is received

hare3/protocol.go
dshulyak: thanks, i think that was mistake. i will change implementation to skip softlock to propose

hare3/protocol.go
dshulyak: all timing is tracked here. saving current round before validating signature doesn't matter for correctness

hare3/protocol.go
dshulyak: no, result is output after waiting for notify. i think i also clarified this in messages above. termination can be done without waiting for notify, but this is optional

hare3/protocol.go
dshulyak: i don't understand, can you clarify?

case 2  is handled by dropping message if identity is malicious (paper), or it is a duplicate from p2p layer

hare3/protocol.go
dshulyak: will update, notify

dshulyak:
> for example, the input/output to the overall protocols and each sub-protocol, the information flow, the responsibility of each component. this document should be specific to the spacemesh protocol (as opposed to the general Byzantine Agreement in the paper).

this is exactly the opposite of what i want. this subprotocols should be as isolated from spacemesh protocol as possible, so that reasoning about them is actually possible, and not muddled by completely irrelevant details.

when you say spacemesh specific, what is that? how proposals are selected? choice of p2p?

hare3/types.go
dshulyak: honestly this is wrong.. id is 20 bytes, why would it be casted to 32 with additional allocation and copy across the codebase. i will change short string to return same data as hash32

hare3/protocol.go
dshulyak: thanks, much more compact now

hare3/hare.go
dshulyak: > i don't understand why round duration is shorter than communication delay (PreroundDelay).

note that calling this as communication delays is misleading. communication delay depends on 2 things: size of the object, work to verify that object. 

activations are naturally the most expensive to verify. ballots/proposals may be large, and less expensive (though right now we have not so optimal code there). hare1 was large, and it wasn't clear if code is doing good job. hare3 is the cheapest

hare3/hare.go
dshulyak: the plan is to update this value to hare 3.1, with graded oracle

hare3/hare.go
dshulyak: synced condition is applied in onLayer method

hare3/hare.go
dshulyak: practically if it votes with empty set, it gives up active role

hare3/hare.go
dshulyak: i changed code  to reflect that

hare3/hare.go
dshulyak: changed

hare3/metrics.go
dshulyak: actually another reason is what you are using as lowest boundary. i thihk that anything that takes less than 1ms is negligible in such protocol (currently lowest boundary is 0.5s)

hare3/hare.go
countvonzero: true. basically didn't have the empirical data to set the propagation time by data type and just used one value for worst case scenario.

hare3/hare.go
```diff
@@ -0,0 +1,592 @@
...
+func (h *Hare) Coins() <-chan WeakCoinOutput {
+	return h.coins
+}
+
+func (h *Hare) Start() {
+	h.pubsub.Register(h.config.ProtocolName, h.Handler)
+	current := h.nodeclock.CurrentLayer() + 1
+	enabled := types.MaxLayer(current, h.config.EnableLayer)
+	enabled = types.MaxLayer(enabled, types.GetEffectiveGenesis()+1)
+	disabled := types.MinLayer(types.LayerID(math.MaxUint32), h.config.DisableLayer)
```
countvonzero, line 214: i don't see a function to set DisableLayer in the option

hare3/hare.go
```diff
@@ -0,0 +1,592 @@
...
+	"github.com/spacemeshos/go-spacemesh/sql/beacons"
+	"github.com/spacemeshos/go-spacemesh/sql/identities"
+	"github.com/spacemeshos/go-spacemesh/sql/proposals"
+	"github.com/spacemeshos/go-spacemesh/system"
+)
+
+type Config struct {
+	Enable          bool          `mapstructure:"enable"`
+	EnableLayer     types.LayerID `mapstructure:"enable-layer"`
+	DisableLayer    types.LayerID `mapstructure:"disable-layer"`
```
countvonzero, line 35: do you envision EnableLayer and DisableLayer will be set at the same time?
i feel safer if we have some sanity check to check btwn h1config.Disable, h3config.EnableLayer and h3.config.DisableLayer to avoid human error.

hare3/hare.go
countvonzero: so my concern is for the worst case scenario (40 iterations) it'll take 25+10*7*40 == 2_825 sec == 47 min.
if we don't control how many session/layer can run hare concurrently, we will have 9 hare3 instances running all the time for sustained consensus failures.

if that's ok memory-wise, then i agree there is no need to add another number.

hare3/hare.go
countvonzero: it happened before on the cloud nodes where there was a long period of inactivity for the pod. i guess it's what you said (cpu overloaded). 

> this is not true, you will finish hare successfully even if wake up notify is 5-10s late.

you mean it will succeed as a passive role? 
but it's also possible to wake up when the hare protocol already terminated.

hare3/hare.go
countvonzero: so if you don't call `LayerPatrol.SetHareInCharge` on a layer, and say the consensus run for more than 5 min, then syncer will sync data from peers and call `mesh.ProcessLayer()` on that layer. i don't know how the interaction would be btwn block generator, mesh and tortoise if that happens. never tested that flow. but at the very least, it would make debugging annoying.

hare3/hare.go
countvonzero: tho it seems the unfairness can be avoided simply by moving input processing to a separate goroutine and manage with a errgroup? future code edits may not guarantee the input processing to be light weight.

hare3/protocol.go
countvonzero: this is confusing for me as using a ptr here implies that it's optional. the current implementation always initialize the struct with a valid msg.

hare3/protocol.go
countvonzero: i misread the code.

hare3/types.go
countvonzero: yeah. it is wrong. tho at the moment i put higher value on debugging than correctness here.
i think it's because the low-level fetcher expect hash32 to download data. i once changed these ids to hash20 and the fetch failed. didn't follow up back then

hare3/hare.go
countvonzero: yes. but there was still code to decode/validate. not simply dropped on as soon as.
there was actually discussion to only register gossip handler after node is ready to handle gossip msgs (fully sync'ed)
otherwise a new node syncing from genesis needs to decode all hare messages only to drop it.

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+type gset struct {
+	values   []types.ProposalID
+	grade    grade
+	smallest types.VrfSignature
+}
+
+// Protocol 2. gradecast. page 13.
+func (g *gossip) gradecast(target IterRound) []gset {
```
countvonzero, line 338: for the purpose of this implementation, it maybe more straightforward to only pass Iteration and always assume proposal round.

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+	// unlike paper we use 5-graded gossip for gradecast as well
+	var rst []gset
+	for key, value := range g.state {
+		if key.IterRound == target && !value.malicious {
```
countvonzero: the last condition `!value.malicious` seems incorrect. the code always set `malicious` to true when the 2nd message is received, along with `otherReceived`. in this code block, `otherReceived` will always be nil.

i think it should be
`if value.otherReceived != nil || !value.malicious`

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+	sort.Slice(rst, func(i, j int) bool {
+		return rst[i].smallest.Cmp(&rst[j].smallest) == -1
+	})
+	return rst
+}
+
+// Protocol 3. thresh-gossip. Page 15.
+// output returns union of sorted proposals received
+// in the given round with minimal specified grade.
+func (g *gossip) thresholdGossip(filter IterRound, grade grade) []types.ProposalID {
```
countvonzero: here the grade is what i understood differently from the paper.

say we are at iteration 0 round softlock (s==1 in protocol 3)
- valid messages include malicious messages as well
- for a given value that is in all sets, find the minimal grade in all valid messages, say g, and that no malicious msg carries higher grade than g. get a count Cv of all the values that satisfy this condition
- get the count of all malicious identities in this round with grade at least g, call it Cm
- if Cv > 0 and Cv+Cm > f, then output (side, r, v, 5) where 5 is d+1-s (5+1-1)

the minimal grade g here doesn't affect the output grade of threshold-gossip
this seems to imply that you also need to remember the atxgrade of the 2nd msg (along with `otherReceived`)

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+		if all[id] >= g.threshold {
+			rst = append(rst, id)
+		}
+	}
+	types.SortProposalIDs(rst)
+	return rst
+}
+
+// thresholdGossipRef returns all references to proposals in the given round with minimal grade.
+func (g *gossip) thresholdGossipRef(filter IterRound, grade grade) []types.Hash32 {
```
countvonzero: see comments from `thresholdGossip`

hare3/protocol.go
countvonzero: tho it's still not intuitive to me why we don't record the round as soon as when the message is received, but not a big deal

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+		// -1 - implementation rounds starts from 0
+		g := grade5 - grade(p.Round-2)
+		p.validValues[g] = p.gossip.thresholdGossip(IterRound{Round: preround}, g)
+	}
+	if p.Round == preround && active {
+		out.message = &Message{Body: Body{
+			IterRound: p.IterRound,
+			Value:     Value{Proposals: p.initial},
+		}}
+	} else if p.Round == hardlock && p.Iter > 0 {
```
countvonzero, line 163: might even want to log fatal/panic if p.round == hardlock && p.Iter == 0

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+		ref, values := p.thresholdProposals(IterRound{Iter: p.Iter - 1, Round: notify}, grade5)
+		if ref != nil && p.result == nil {
+			p.result = ref
+			out.result = values
+			if values == nil {
+				// receiver expects non-nil result
+				out.result = []types.ProposalID{}
+			}
+		}
+		if ref, _ := p.thresholdProposals(IterRound{Iter: p.Iter - 1, Round: commit}, grade4); ref != nil {
```
countvonzero: should this be conditional on `if p.result != nil`

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+					Value:     Value{Reference: ref},
+				}}
+			}
+		}
+	} else if p.Round == notify && active {
+		ref := p.result
+		if ref == nil {
+			ref, _ = p.thresholdProposals(IterRound{Iter: p.Iter, Round: commit}, grade5)
+		}
+		if ref != nil {
```
countvonzero, line 255: commit round case 2 requires also ref is part of validProposals?

hare3/protocol.go
```diff
@@ -0,0 +1,397 @@
...
+				// receiver expects non-nil result
+				out.result = []types.ProposalID{}
+			}
+		}
+		if ref, _ := p.thresholdProposals(IterRound{Iter: p.Iter - 1, Round: commit}, grade4); ref != nil {
+			p.locked = ref
+			p.hardLocked = true
+		} else {
+			p.locked = nil
+			p.hardLocked = false
```
countvonzero, line 181: does p.validProposals need to be reset?

countvonzero:
> this is exactly the opposite of what i want. this subprotocols should be as isolated from spacemesh protocol as possible, so that reasoning about them is actually possible, and not muddled by completely irrelevant details.
> 
> when you say spacemesh specific, what is that? how proposals are selected? choice of p2p?

what i am looking for, is to translate the paper into a technical document about the hare protocol, assuming no prior knowledge of hare protocol from the reader
- without the math proofs
- with spacemesh specific descriptions .eg.
  - how GradeKey() is realized
  - how proposals are selected
  - how malicious identities are discovered / handled
  - the intuitions behind grading for the overall protocol
  ...

the goal of this document should be for other devs to understand the implementation at a high level without having to read the academic paper.

having said that, i understand that this a big-ish project and includes technical debts (things that are not documented in hare 1). so i am thinking we may need help from some kind of professional tech writer.

hare3/protocol.go
dshulyak: thresholdProposals and commit both ensure that reference exists in valid proposals

hare3/protocol.go
dshulyak: i like that IterRound used everywhere

hare3/protocol.go
dshulyak: thanks, i missed that when refactored it to use single state

hare3/protocol.go
dshulyak: i think not. protocol can terminate using valid reference in some previous iteration, even if in current proposal wasn't created at all

hare3/protocol.go
dshulyak: i think this is enough to test that skip, and not overload code with not so useful conditions

hare3/protocol.go
dshulyak: to save on execution time? i think the important part is that notify is always sent for  p.result, regardless of what is seenn as committed in the last round

hare3/hare.go
dshulyak: the idea is that hare31 will start where hare3 stops. i will check that DisableLayer is higher than EnableLayer if protocol is enabled

hare3/hare.go
dshulyak: my point was that you control max concurrent sessions with IterationsLimit, there is no reason to introduce one more variable

hare3/hare.go
dshulyak: i think this synchronization is unnecessary, but will add it back

hare3/hare.go
dshulyak: > it happened before on the cloud nodes where there was a long period of inactivity for the pod. i guess it's what you said (cpu overloaded).

i am not sure if this is the case. i would rather believe that it was due to bugs in this clock that we use for unclear reason. or simply logs were missing. i run nodes that are heavily throttled in my tests, it is unrealistic for a node not to send an event

> you mean it will succeed as a passive role?

it depends how much clock is off, if it is off within small boundary (e.g +-5s) it will contribute to the protocol as active

> but it's also possible to wake up when the hare protocol already terminated.

why does it matter? it has to wake up signficantly later for this to happen. in that case session will just sit idle

hare3/hare.go
dshulyak: i don't need that function

hare3/hare.go
dshulyak: i don't see why they won't be light weight unless someone will make mistakes

hare3/protocol.go
dshulyak: it is to avoid copy of the data

hare3/protocol.go
dshulyak: sorry i don't follow where min grade step came from, min is in the method parameters (gave an example below)

more practically speaking, lets use preround messages as an example (hardlock is skipped):
- at softlock votes are computed from min(atx grade, msg grade) == 5
- at propose votes are computed from min(atx grade, msg grade) >= 4
- at wait1 votes are computed from min(atx grade, msg grade) >= 3
- at wait2 votes are computed from min(atx grade, msg grade) >= 2

note that msg grade is implicit because of the timing when thresholdGossip is called. e.g at hardlock we will receive preround messages within single delay. at propose - within 2 delays and so on.

then we ensure that votes crossed threshold, and there is atleast one honest vote.

dshulyak:
@countvonzero could you please clarify with requested changes what i am expected to fix? is it only about that threshold gossip or more? tbh this discussion is a bit too much for me, i would prefer to move somehow asynchronously

hare3/protocol.go
dshulyak: mapping should always exist locally, otherwise protocol is in invalid state

hare3/protocol.go
dshulyak: the same for commit:

notify: min(atx grade, msg grade) >= 5
hardlock: min(atx grade, msg grade) >= 4
softlock: min(atx grade, msg grade) >= 3
propose: min(atx grade, msg grade) >= 2
commit: min(atx grade, msg grade) >= 1

same caveat about msg grade, it is implicit because when threshold gossip is called.

i also rechecked this part of the protocol:
- valid messages do not include malicious, they are very clearly separate, the protocol checks that there is atleast one non-malicious. see 4th in 3(a)
- filtering in 3(a) is applied almost as is in the code, filter by round (1st), filter by grade (3rd, 4th)
- then we count malicious messages, why would they be a part of valid messages?
- and threshold is passed if there is atleast 1 valid/good message, and total number of messages (together) with equivocation crossed threshold

hare3/hare.go
countvonzero: ok. somehow i was reading an older version of code where config.EnableLayer is set with `WithEnableLayer`

hare3/hare.go
countvonzero: my assumption was that 40 iterations is a security parameter. that's why i didn't think it's good to mix the two params.

hare3/protocol.go
countvonzero: this is weird as in paper the variable T is suffixed with iteration.
shouldn't each iteration operates independently on the proposal round msgs of that iteration?

countvonzero (approved):

hare3/hare.go
dshulyak: but if there will be additional parameter for max concurrent iterations, then it will have to overwrite iterations limit right? so i would rather change 40 iterations. for example i definitely don't want to start with 40 on mainnet

hare3/protocol.go
dshulyak: i read that condition as S included in Ti in any iteration between [0; j-1]

countvonzero: ah yes. you are right

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
cluster was prematurely terminated. can't find any failures

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

PR #4770, opened 2023-08-03
[Merged by Bors] - p2p: disable autoscaling and drop from bootnodes on enough peers

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4771

total hardcap without autoscaling should be high peers + min peers + transient new connections (min peers).
once node reached high-peers watermark, it will disconnect from bootnodes, and reconnect only once crossing min-peers watermark


pigmej:
I have run it on several nodes number of connections was under the control.

pigmej (approved):

dshulyak:
bors merge

PR #4773, opened 2023-08-04
[Merged by Bors] - Remove best provider option for OpenCL provider ID

### LINKED ISSUE #212: Remove automatic selection for best provider
### Description

At the moment the node defaults to provider -1 in their config if not set to a different provider by the user. This value is problematic for multiple reasons:

1. The provider ID is an `uint` which means `-1` will be cast to either `2^32 - 1` or `2^64 - 1` depending on the system.
2. The CPU provider of post-rs has the hard coded ID `2^32 -1` which can cause the node to do a benchmarks when not intended.

### Acceptance criteria

- Remove `-1` as identifier for "select best provider". The user has to select a provider explicitly via Smapp or the config file.
- Remove automatic benchmarking; but not manually triggered (used by Smapp to display speeds of providers to user)
- Default to CPU provider in config, so that unit and system tests still behave as expected

fasmat:
## Motivation
Closes https://github.com/spacemeshos/post/pull/212

Merge after: https://github.com/spacemeshos/post/pull/213 and https://github.com/spacemeshos/api/pull/256

## Changes
- Smeshing Provider ID is still parsed as `uint32` from config and command line (to not unexpectedly break anything) but now can be `nil` or unset.
- `-1` has been removed as "automatically select best provider" option. Instead the user has to manually set the provider of their choice.

## Test Plan
- New tests were added that check that start smeshing fails without provider if initialization isn't complete and passes when initialization is completed
- Tests were added that ensure that provider ID is parsed correctly when passed via CLI or config

## TODO

## DevOps Notes

### DIFF
--- a/activation/post.go
+++ b/activation/post.go
@@ -222,8 +222,8 @@ func DefaultPostSetupOpts() PostSetupOpts {
 	}
 }
 
-func (o PostSetupOpts) ToInitOpts() (config.InitOpts, error) {
-	initOpts := config.InitOpts{
+func (o PostSetupOpts) ToInitOpts() config.InitOpts {
+	return config.InitOpts{
 		DataDir:          o.DataDir,
 		NumUnits:         o.NumUnits,
 		MaxFileSize:      o.MaxFileSize,
@@ -232,7 +232,6 @@ func (o PostSetupOpts) ToInitOpts() (config.InitOpts, error) {
 		Scrypt:           o.Scrypt,
 		ComputeBatchSize: o.ComputeBatchSize,
 	}
-	return initOpts, nil
 }
 
 // PostSetupManager implements the PostProvider interface.
@@ -421,15 +420,11 @@ func (mgr *PostSetupManager) PrepareInitializer(ctx context.Context, opts PostSe
 		return err
 	}
 
-	initOpts, err := opts.ToInitOpts()
-	if err != nil {
-		return err
-	}
 	newInit, err := initialization.NewInitializer(
 		initialization.WithNodeId(mgr.id.Bytes()),
 		initialization.WithCommitmentAtxId(mgr.commitmentAtxId.Bytes()),
 		initialization.WithConfig(mgr.cfg.ToConfig()),
-		initialization.WithInitOpts(initOpts),
+		initialization.WithInitOpts(opts.ToInitOpts()),
 		initialization.WithLogger(mgr.logger.Zap()),
 	)
 	if err != nil {
--- a/activation/post.go
+++ b/activation/post.go
@@ -5,8 +5,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math"
 	"runtime"
+	"strconv"
 	"sync"
 
 	"github.com/spacemeshos/post/config"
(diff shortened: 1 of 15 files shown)


activation/post.go
```diff
@@ -169,23 +170,32 @@ func DefaultPostSetupOpts() PostSetupOpts {
...
 		ComputeBatchSize: opts.ComputeBatchSize,
 	}
 }
 
-func (o PostSetupOpts) ToInitOpts() config.InitOpts {
-	return config.InitOpts{
+func (o PostSetupOpts) ToInitOpts() (config.InitOpts, error) {
+	if o.ProviderID == math.MaxUint64 {
+		return config.InitOpts{}, fmt.Errorf("no provider ID specified: %d", o.ProviderID)
+	}
```
poszu: Why use an uin64 and a magic number `math.MaxUint64` to express optionality instead of `*uint32` as in the post repo?

fasmat: Because command line / config parsing doesn't have a `nil` option for values that are not set. Instead parameters passed to the node have to use some kind of default value.

Usually this is done by making the default `0` and checking for that, but since `0` can be a valid ProviderID I had to be a bit more creative.

I can move this check from here to `cmd/root.go` where the value is actually parsed, but the only way to react to an invalid or missing value is to `panic` there.

go.mod
```diff
@@ -36,13 +36,13 @@ require (
 	github.com/pyroscope-io/pyroscope v0.37.2
 	github.com/santhosh-tekuri/jsonschema/v5 v5.3.1
 	github.com/seehuhn/mt19937 v1.0.0
-	github.com/spacemeshos/api/release/go v1.17.0
+	github.com/spacemeshos/api/release/go v1.17.1-0.20230804164905-1af0fe873e1f
```
fasmat: Update to new version

go.mod
```diff
@@ -36,13 +36,13 @@ require (
...
 	github.com/spacemeshos/economics v0.1.0
 	github.com/spacemeshos/fixed v0.1.0
 	github.com/spacemeshos/go-scale v1.1.10
 	github.com/spacemeshos/merkle-tree v0.2.2
 	github.com/spacemeshos/poet v0.8.7
-	github.com/spacemeshos/post v0.8.11
+	github.com/spacemeshos/post v0.8.12-0.20230804123744-7e32021db1c1
```
fasmat: Update to new version

fasmat:
bors try

activation/post.go
```diff
@@ -82,12 +83,58 @@ type PostSetupOpts struct {
...
+
+	i, err := strconv.ParseUint(string(text), 10, 32)
+	if err != nil {
+		return fmt.Errorf("failed to parse PoST Provider ID (%s): %w", text, err)
+	}
+
+	id.value = new(uint32)
+	*id.value = uint32(i)
+	return nil
+}
```
poszu: You could keep supporting a number value in JSON config file with:
```diff
diff --git a/activation/post.go b/activation/post.go
index 7154659ef..07e07a250 100644
--- a/activation/post.go
+++ b/activation/post.go
@@ -5,10 +5,12 @@ import (
        "encoding/hex"
        "errors"
        "fmt"
+       reflect "reflect"
        "runtime"
        "strconv"
        "sync"

+       "github.com/mitchellh/mapstructure"
        "github.com/spacemeshos/post/config"
        "github.com/spacemeshos/post/initialization"
        "github.com/spacemeshos/post/proving"
@@ -135,6 +137,22 @@ func (id *PostProviderID) UnmarshalText(text []byte) error {
        return nil
 }

+func DecodePostProviderIdHookFunc() mapstructure.DecodeHookFunc {
+       return func(f, t reflect.Type, data any) (any, error) {
+               if f.Kind() != reflect.Float64 {
+                       return data, nil
+               }
+               if t != reflect.TypeOf(PostProviderID{}) {
+                       return data, nil
+               }
+
+               p := PostProviderID{value: new(uint32)}
+               *p.value = uint32(data.(float64))
+
+               return p, nil
+       }
+}
+
 // PostProvingOpts are the options controlling POST proving process.
 type PostProvingOpts struct {
        // Number of threads used in POST proving process.
diff --git a/node/node.go b/node/node.go
index a03f3a487..fcef50d50 100644
--- a/node/node.go
+++ b/node/node.go
@@ -250,6 +250,7 @@ func LoadConfigFromFile() (*config.Config, error) {
                mapstructure.StringToSliceHookFunc(","),
                mapstructureutil.BigRatDecodeFunc(),
                mapstructure.TextUnmarshallerHookFunc(),
+               activation.DecodePostProviderIdHookFunc(),
        )

        // load config if it was loaded to the viper
```

node/node_test.go
```diff
@@ -760,6 +760,78 @@ func TestConfig_Preset(t *testing.T) {
...
+	tt := []struct {
+		name         string
+		cli          string
+		config       string
+		updatePreset func(*testing.T, *config.Config)
+	}{
+		{
+			name:   "smeshing-opts-provider",
+			cli:    "--smeshing-opts-provider=1337",
+			config: `{"smeshing": {"smeshing-opts": {"smeshing-opts-provider": "1337"}}}`,
```
poszu: ... then this will work:
```suggestion
			config: `{"smeshing": {"smeshing-opts": {"smeshing-opts-provider": 1337}}}`,
```

activation/post.go
fasmat: Thanks for the hint, I ended up doing something similar that now works with integers and if no value is set.

activation/post.go
```diff
@@ -82,12 +83,57 @@ type PostSetupOpts struct {
...
+
+	id.value = new(int64)
+	*id.value = int64(i)
+	return nil
+}
+
+// SetInt64 sets the value of the PostProviderID to the given int64.
+func (id *PostProviderID) SetInt64(value int64) {
+	id.value = new(int64)
+	*id.value = value
```
poszu: Wouldn't this work:
```suggestion
	id.value = &value
```

poszu (approved):

activation/post.go
fasmat: In principle yes, but consider the following code:

```go
value := int64(5)
id := new(ProstProviderID)

id.SetInt64(value)

value = int64(2)
```

with the unchanged code `id` doesn't change when `value` is changed. If we use the reference of the passed in variable, `id` changes when `value` changes, which would be unexpected behaviour - at least to me.

EDIT: this is incorrect, see comment below.

fasmat: Nvm forget my previous comment, the parameter is passed by value not reference, so it makes no difference 🙂

fasmat:
bors merge

fasmat:
bors merge

PR #4774, opened 2023-08-04
hare: new node passively participate in the epoch it becomes synced

countvonzero:
## Motivation

following up on https://github.com/spacemeshos/go-spacemesh/pull/4718
new node should not actively participate in hare because it doesn't have correct timestamp for atxs and malfeasance proof

## Changes
for node that become synced in current epoch, only participate in hare passively.
start active participation in the next epoch.


dshulyak:
so if it was restarted the node is not participating in hare? that will certainly halt whole network

countvonzero:
doesn't work :(

PR #4775, opened 2023-08-04
[Merged by Bors] - p2p: add connection gater to prevent dials over max limit

### LINKED ISSUE #4771: Node ignores high-peers value defined in the config map
### Environment

* Smapp version: v1.0.11
* Node version: v1.0.7
* OS: Ubuntu 22.04

### Describe the bug

Spacemesh node constantly has over 100 connections when it's run through Smapp. It seems that `high-peers` value is not used although it's present in the config map. Values defined in the config map are:
```json
  "p2p": {
    "disable-reuseport": false,
    "p2p-disable-legacy-discovery": true,
    "bootnodes": [
        "/dns4/mainnet-bootnode-10.spacemesh.network ...",
        "... omitted ...",
        "/dns4/mainnet-bootnode-19.spacemesh.network ..."
    ],
    "min-peers": 10,
    "low-peers": 20,
    "high-peers": 30
  },
```

The issue started to appear with Smapp version `v1.0.11`. I didn't have any problems with `v1.0.9`.

### Steps to reproduce

1. Download `spacemesh_app_1.0.11_amd64.deb` from the release page
2. Install
3. Run Smapp
4. Observe the number of Connected neighbors in the Network tab

### Expected behavior

The number of Connected neighbors should be between 10 and 30.

### Actual behavior

The number of Connected neighbors is constantly over 100. The node is running for 3 hours already and the value is not decreasing. Restart also doesn't help.

### Logs

N/A

### Additional info
```
xxx@xxx:/opt/Spacemesh/node$ ./go-spacemesh version
v1.0.7+65f516c6fca65eaf495b227aad526390eaa8eb20
```
```
xxx@xxx:~$ grpcurl -plaintext 0.0.0.0:9092 spacemesh.v1.NodeService.Version
{
  "versionString": {
    "value": "v1.0.7"
  }
}
```
```
xxx@xxx:~$ grpcurl -plaintext 0.0.0.0:9092 spacemesh.v1.NodeService.Status
{
  "status": {
    "connectedPeers": "112",
    "isSynced": true,
    ...
  }
}
```

It could be that the issue is with the `go-spacemesh` and not Smapp, but I'm not sure.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4771

this change will reject dials from outbound peers when inbound number is 0.8 of high peers number.
it will not allow dials to other peers when total number of peers is higher then 1.1 of high peers number.
and node is allowed to always dial direct peers

note that dials are concurrent, so if there is a burst it is still possible that they will go over the limit for short period of time

### DIFF
--- a/p2p/gater.go
+++ b/p2p/gater.go
@@ -1,0 +1,34 @@
+package p2p
+
+import (
+	"github.com/libp2p/go-libp2p/core/control"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/multiformats/go-multiaddr"
+)
+
+type gater struct {
+	h   host.Host
+	max int
+}
+
+func (*gater) InterceptPeerDial(_ peer.ID) bool {
+	return true
+}
+
+func (g *gater) InterceptAddrDial(pid peer.ID, m multiaddr.Multiaddr) bool {
+	return len(g.h.Network().Peers()) <= g.max
+}
+
+func (g *gater) InterceptAccept(n network.ConnMultiaddrs) bool {
+	return len(g.h.Network().Peers()) <= g.max
+}
+
+func (*gater) InterceptSecured(_ network.Direction, _ peer.ID, _ network.ConnMultiaddrs) bool {
+	return true
+}
+
+func (*gater) InterceptUpgraded(_ network.Conn) (allow bool, reason control.DisconnectReason) {
+	return true, 0
+}
--- a/p2p/gater.go
+++ b/p2p/gater.go
@@ -1,6 +1,7 @@
 package p2p
 
 import (
+	"github.com/libp2p/go-libp2p/core/connmgr"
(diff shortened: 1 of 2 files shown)


pigmej:
1) Connections just sporadically go out of the `high`. 🟢 
2) Direct peers are working even when `high` is exceeded 🟢 

side note:
*WITHOUT* that patch my limits were ok when I did not have port forwarded (fewer peers connecting??), with port forwarded then limits were often exceeded but then it was going lower.

dshulyak:
> Connections just sporadically go out of the high

can you clarify this part? i expect them to go above high if node is making many concurrent dials, so for example you are at 80 peers, node make 50 dials at the same time, connections will to 130, but that should be rare. is it consistent with your observations or not at all?

pigmej:
Yeah, more or less what you described. Sporadically going out of bounds and then quickly down.

pigmej (approved):

countvonzero (approved):

dshulyak:
bors try

dshulyak:
bors merge

PR #4779, opened 2023-08-05
[Merged by Bors] - beacon: handle malicious identity

### LINKED ISSUE #4772: beacon: vote against proposal from malicious identities
## Description
when we have a malfeasance proof for an identity, protocol should vote against its proposals and give it weight 0 in voting.

countvonzero:
## Motivation
Closes #4772

## Changes
for malicious identity's 
- proposal: categorized as potentially valid (voting against)
- first vote: accept with weight 0
- following vote: accept with weight 0

malicious identity's weight is not included in the total epoch weight

### DIFF
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -304,40 +304,46 @@ func (pd *ProtocolDriver) OnAtx(atx *types.ActivationTxHeader) {
 	if !ok {
 		return
 	}
-	if id, ok := s.minerAtxs[atx.NodeID]; ok && id != atx.ID {
+	if mi, ok := s.minerAtxs[atx.NodeID]; ok && mi.atxid != atx.ID {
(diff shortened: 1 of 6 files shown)


dshulyak (commented):
doesn't it require grading too? otherwise how we can be sure that malicious identities are seen the same way

countvonzero:
> doesn't it require grading too? otherwise how we can be sure that malicious identities are seen the same way

i specifically asked Tal/Iddo in barcelona. they said it only requires 2 grades going into the protocol. (malicious or not) because this is tortoise style consensus. (i don't fully understand why but tortoise did only use 2 grades).

so i only count identities that are malicious before the protocol starts.

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

Issue #4504, opened 2023-06-11
sync, tortoise: improve speed and efficiency of participation in protocol after restart


dshulyak:
the pattern occurs when all nodes are restarted in the network, example:

- layer 100 is verified, and everyone votes consistently on it
- no ballots and blocks in layers 101-102 
- layers starting from 103 will have a block, but ballots in layers 103-105 will vote abstain on previous layers making them uncountable in verifying mode
- after layer 105 all opinions should be consistent in the network, and with some number of layers we should be able to cross pessimitic threshold

the downside is that network had to switch into full mode to verify ballots, with long hdist this period will take quite some time and definitely not optimal
  
it can be tested in system tests by failing all nodes (don't delete them, as it will also drop the state)

dshulyak:
i run the test to explore this behavior. by launching 20 nodes and interupting them all for 30s, every 30m [chaos mesh template](https://github.com/dshulyak/go-spacemesh/blob/debug/restart/systest/parameters/chaos/failall.yaml)

additionally i set both out of sync and gossip sync threshold to 0 [change](https://github.com/dshulyak/go-spacemesh/commit/a3e3803718a6bbce29ea8be321cd98bb641c9ea7). this is not important for tortoise behavior, but it helps to get into normal operation faster.

[...]

PR #4780, opened 2023-08-05
sync: remove gossip sync

countvonzero:
## Motivation
part of #4504

## Changes
- remove gossip sync state
- always sync to the current layer before changing to synced state


countvonzero:
bors try

countvonzero:
bors try

dshulyak:
so what do need to proceed with this? i was thinking about changing approach, from asking syncer to verifying in each component individually if it has enough data. this would be:

#### hare

- beacon exists
- identity is active with provided beacon

#### tortoise (proposals)

- if reference ballot exists - always participate
otherwise
- beacon exists
- first block active set or graded active set exists

#### beacon

- any active set exists

PR #4782, opened 2023-08-07
[Merged by Bors] - Don't rewrite poet config

poszu:
## Motivation
The poet config is unnecessarily and wrongly rewritten in node setup code.

## Changes
Don't rewrite poet config at all.

### DIFF
--- a/activation/nipost.go
+++ b/activation/nipost.go
@@ -197,6 +197,7 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 		log.Time("poet proof deadline", poetProofDeadline),
 		log.Stringer("publish epoch", pubEpoch),
 		log.Stringer("target epoch", challenge.TargetEpoch()),
+		log.Any("config", nb.poetCfg),
 	)
 
 	challengeHash := challenge.Hash()
--- a/activation/nipost.go
+++ b/activation/nipost.go
@@ -197,7 +197,6 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 		log.Time("poet proof deadline", poetProofDeadline),
 		log.Stringer("publish epoch", pubEpoch),
 		log.Stringer("target epoch", challenge.TargetEpoch()),
-		log.Any("config", nb.poetCfg),
 	)
 
 	challengeHash := challenge.Hash()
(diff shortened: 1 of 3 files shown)


countvonzero (approved):

activation/nipost.go
```diff
@@ -197,6 +197,7 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 		log.Time("poet proof deadline", poetProofDeadline),
 		log.Stringer("publish epoch", pubEpoch),
 		log.Stringer("target epoch", challenge.TargetEpoch()),
+		log.Any("config", nb.poetCfg),
```
fasmat: This will print redundant information here (mostly what's already logged at the other fields). I think it would be better to instead log the relevant config parameters in the `poetClient`, like MaxRetries and RetryDelay.

fasmat (approved):

dshulyak:
would it be nice to have a test that correct config is being used? also true for other components obviously

poszu:
> would it be nice to have a test that correct config is being used? also true for other components obviously

Yes, but engineering a systest to check it would be really painful. Would you happen to have an idea of how to test it best?

dshulyak:
maybe just a unit test to check that correct config is being used in poet client.
or systest, that ensures that we always select best poet proof even if it becomes available later, thinking:
- throttle one of the poets with lower cpu limits - slow poet
- make poet round end later (by 10s for example) on the fast poet
- assert that all atxs are using fast poet

but it is indeed time consuming to setup all that

poszu:
>     * throttle one of the poets with lower cpu limits - slow poet
> 
>     * make poet round end later (by 10s for example) on the fast poet

Actually there is no need to throttle if one of the poets runs for less time, because then it will have less ticks.

dshulyak (approved):
> Actually there is no need to throttle if one of the poets runs for less time, because then it will have less ticks.

true, but i assumed that 10s will not necessarily guarantee more ticks

activation/poet.go
```diff
@@ -63,6 +63,9 @@ func withCustomHttpClient(client *http.Client) PoetClientOpts {
 
 // NewHTTPPoetClient returns new instance of HTTPPoetClient connecting to the specified url.
 func NewHTTPPoetClient(baseUrl string, cfg PoetConfig, opts ...PoetClientOpts) (*HTTPPoetClient, error) {
+	// TODO(brozansk): Take a logger and use it instead of global logger
```
fasmat, line 66: The `retryablehttp.Client` also takes a `Logger` field. Maybe we can also init that with the logger passed to `NewHTTPPoetClient`?

poszu: I created a ticket to do so: https://github.com/spacemeshos/go-spacemesh/issues/4783

poszu:
bors merge

PR #4795, opened 2023-08-09
[Merged by Bors] - p2p: add ip4/ip6 blocklists

### LINKED ISSUE #4788: Make possible to disable local network scanning in discovery in the config
Currently, go-spacemesh will eagerly /scan/ local networks for peers. 

In some environments, that's not good and could be seen as suspicious network activity by cloud providers and some local network providers too.

Currently, the workaround is to block it via a firewall:
```
iptables -A OUTPUT -d 192.168.0.0/16 -j DROP
iptables -A OUTPUT -d 172.16.0.0/12 -j DROP
iptables -A OUTPUT -d 10.0.0.0/8 -j DROP
```

But there should be in code equivalent for more polite behavior of go-spacemesh.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4788

default ip4/ip6 blocked ranges https://github.com/spacemeshos/go-spacemesh/blob/a8f86a19fffde6141110b21d398b65bf017d2c57/p2p/host.go#L48-L66C1

to disable them set `"ip4-blocklist": []` or `"ip6-blocklist": []` in config. 
if `private-network` is set to true, blocklists will be ignored. that option meant to be used only in testing networks.
`direct` peers will bypass blocklists validation.


p2p/gater.go
```diff
@@ -43,3 +78,38 @@ func (*gater) InterceptSecured(_ network.Direction, _ peer.ID, _ network.ConnMul
...
+func (g *gater) isBlocked(m multiaddr.Multiaddr) bool {
+	blocked := false
+	multiaddr.ForEach(m, func(c multiaddr.Component) bool {
+		switch c.Protocol().Code {
+		case multiaddr.P_IP4:
+			blocked = inAddrRange(net.IP(c.RawValue()), g.ip4blocklist)
+		case multiaddr.P_IP6:
+			blocked = inAddrRange(net.IP(c.RawValue()), g.ip6blocklist)
+		}
+		return false
```
countvonzero: there is no need to inspect all components?

p2p/gater.go
```diff
@@ -28,8 +60,11 @@ func (g *gater) InterceptPeerDial(pid peer.ID) bool {
 	return len(g.h.Network().Peers()) <= g.outbound
 }
 
-func (*gater) InterceptAddrDial(pid peer.ID, m multiaddr.Multiaddr) bool {
-	return true
+func (g *gater) InterceptAddrDial(pid peer.ID, m multiaddr.Multiaddr) bool {
+	if !g.InterceptPeerDial(pid) {
+		return false
+	}
+	return !g.isBlocked(m)
```
countvonzero: nit: double-negative is always confusing to me.
would be slightly more readable for me if it's return `g.allowed(m)`

countvonzero (approved):

p2p/gater.go
dshulyak: i think there is an assumption that ip4/ip6 is a first component, and i just copied this code. but i changed it to iterate just in case

pigmej (approved):
Works for me.

dshulyak:
bors merge

PR #4802, opened 2023-08-09
[Merged by Bors] - p2p: remain connected to bootnode until it prunes you

dshulyak:
this was a fix i did in rush, we now have a connection gater that will enforce that node doesn't dial over the limit.
it is better if regular node remains connected after it got enough peers, this way if some peers got disconnected sporadically it will get more. eventually node will be pruned by a bootnode due to the settings we are running on them


countvonzero (approved):

fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #4803, opened 2023-08-09
[Merged by Bors] - Fix failing Windows tests

fasmat:
## Motivation
Several failing windows tests have been fixed.

## Changes
- `TestFetch_PeerDroppedWhenMessageResultsInValidationReject`: the instantiated p2p hosts weren't stopped correctly leading to a still open instance of `levelds.Datastore` that prevented the test from properly cleaning up
- `TestPrologue`: stopping discovery did not close the `levelds.Datastore` if dht was used
- `TestAdminEvents` was indirectly affected by the issue above
- `TestConnectedPersist` and `TestConnectedBrokenCRC` the name of the temporary file used to persist peers atomically is not legal on windows file systems, so writing the file always failed.

## Test Plan
n/a

## TODO

## DevOps Notes

### DIFF
--- a/fetch/fetch_test.go
+++ b/fetch/fetch_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/golang/mock/gomock"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/spacemeshos/go-spacemesh/codec"
@@ -312,7 +313,7 @@ func TestFetch_RegisterPeerHashes(t *testing.T) {
 func TestFetch_PeerDroppedWhenMessageResultsInValidationReject(t *testing.T) {
 	lg := logtest.New(t)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
-	t.Cleanup(cancel)
+	defer cancel()
 	cfg := Config{
 		time.Minute * time.Duration(2000), // make sure we never hit the batch timeout
 		3,
@@ -328,13 +329,13 @@ func TestFetch_PeerDroppedWhenMessageResultsInValidationReject(t *testing.T) {
 	// Good host
 	h, err := p2p.New(ctx, lg, p2pconf, []byte{})
 	require.NoError(t, err)
-	t.Cleanup(func() { h.Close() })
+	t.Cleanup(func() { assert.NoError(t, h.Stop()) })
 
 	// Bad host, will send a message that results in validation reject
 	p2pconf.DataDir = t.TempDir()
 	badPeerHost, err := p2p.New(ctx, lg, p2pconf, []byte{})
 	require.NoError(t, err)
-	t.Cleanup(func() { badPeerHost.Close() })
+	t.Cleanup(func() { assert.NoError(t, badPeerHost.Stop()) })
 
 	err = h.Connect(ctx, peer.AddrInfo{
 		ID:    badPeerHost.ID(),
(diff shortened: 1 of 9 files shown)


dshulyak (approved):
bors merge

fasmat:
bors cancel

sorry it seems the issue isn't resolved with this PR 😞

fasmat:
I found a proper fix and fixed the other consistently failing tests for windows 🙂

fetch/fetch_test.go
```diff
@@ -312,7 +313,7 @@ func TestFetch_RegisterPeerHashes(t *testing.T) {
 func TestFetch_PeerDroppedWhenMessageResultsInValidationReject(t *testing.T) {
 	lg := logtest.New(t)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
-	t.Cleanup(cancel)
+	defer cancel()
```
poszu, line 316: What's the difference here? When to use one or the other?

p2p/dhtdiscovery/discovery.go
```diff
@@ -278,5 +281,12 @@ func newDht(ctx context.Context, h host.Host, public, server bool, dir string) (
 	} else {
 		opts = append(opts, dht.Mode(dht.ModeAutoServer))
 	}
-	return dht.New(ctx, h, opts...)
+	dht, err := dht.New(ctx, h, opts...)
+	if err != nil {
+		ds.Close()
```
poszu: The error from `ds.Close()` is not handled. Perhaps it could be at least logged?

poszu (approved):

fetch/fetch_test.go
fasmat: Order of execution - `defer`s are executed before cleanups:

```golang
func Test_Defer(t *testing.T) {
	t.Cleanup(func() { t.Log("cleanup") })
	defer t.Log("deferred")
	t.Cleanup(func() { t.Log("cleanup2") })
	defer t.Log("deferred2")
}
```

prints:

```
=== RUN   Test_Defer
    fetch_test.go:409: deferred2
    fetch_test.go:409: deferred
    fetch_test.go:407: cleanup2
    fetch_test.go:405: cleanup
--- PASS: Test_Defer (0.00s)
PASS
```

p2p/dhtdiscovery/discovery.go
fasmat: I added a log here, since if the returned error is only logged anyway and triggers the node to shut down.

fasmat:
bors merge

p2p/persist.go
```diff
@@ -46,7 +46,7 @@ func writePeers(h host.Host, dir string) error {
 	}
 
 	checksum := crc64.New(crc64.MakeTable(crc64.ISO))
-	tmp, err := os.CreateTemp(dir, "connected-***")
+	tmp, err := os.CreateTemp(dir, "connected.tmp")
```
dshulyak, line 49: how it will work if connected.tmp exists? will it overwrite it?

poszu: The second argument is a pattern to which a random string is appended. See https://pkg.go.dev/os#CreateTemp
>CreateTemp creates a new temporary file in the directory dir, opens the file for reading and writing, and returns the resulting file. The filename is generated by taking pattern and adding a random string to the end. If pattern includes a "\*", the random string replaces the last "\*". If dir is the empty string, CreateTemp uses the default directory for temporary files, as returned by TempDir. Multiple programs or goroutines calling CreateTemp simultaneously will not choose the same file. The caller can use the file's Name method to find the pathname of the file. It is the caller's responsibility to remove the file when it is no longer needed.

fasmat: According to the documentation if no `*` are specified in the name, a random string is appended to its name. The problem was that `os.CreateTemp` only replaces the last `*` with a random string and `connected-**123972` is not a valid filename on windows.

So now the filename is `connected.tmp123798` where the number at the end is a different randomly generated one every time `os.CreateTemp` is called.

It would probably have been better if I had used the pattern `connected-*.tmp` or similar to ensure it works on any platform. Do you want me to change it?

PR #4809, opened 2023-08-10
[Merged by Bors] - Fix grpc logging levels

poszu:
## Motivation
Changing the log level for the `GrpcLogger` is currently impossible.

## Changes
- added log level for the `GrpcLogger` "grpc" to `LoggingConfig struct`
- use the GRPClogger for all GRPC services instead of the app log.

## Test Plan
Manual test if changing log level is reflected in logs.

### DIFF
--- a/config/logging.go
+++ b/config/logging.go
@@ -17,6 +17,7 @@ const (
 type LoggerConfig struct {
 	Encoder                   LogEncoder `mapstructure:"log-encoder"`
 	AppLoggerLevel            string     `mapstructure:"app"`
+	GrpcLoggerLevel           string     `mapstructure:"grpc"`
 	P2PLoggerLevel            string     `mapstructure:"p2p"`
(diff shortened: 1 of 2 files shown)


config/logging.go
```diff
@@ -17,6 +17,7 @@ const (
 type LoggerConfig struct {
 	Encoder                   LogEncoder `mapstructure:"log-encoder"`
 	AppLoggerLevel            string     `mapstructure:"app"`
+	GrpcLoggerLevel           string     `mapstructure:"grpc"`
```
fasmat, line 20: This field needs a default value (a few lines down in this file)

fasmat (approved):

config/logging.go
poszu: Interesting that it works anyway and gets an INFO level by default 🤷

poszu:
bors merge

Issue #4751, opened 2023-07-27
logging: remove global logger


fasmat:
## Description
After https://github.com/spacemeshos/go-spacemesh/pull/4749 there are still some parts in go-spacemesh left that use the global logger instead of a service-specific one. This makes testing & configuration of those logs more difficult. 

## Affected code
Specifically the following packages still use the global logger in some places:

- `events`
- `common/types`
- `hare`
- events in `grpcserver`

Todo:
- [x] #4783 
- [ ] remove uses of global logger in `package events`
- [ ] remove global logger completely

The logger in those packages is used in functions that are not part of a service and thereby do not yet have a logger thats injected into them. Consider removing them by one of the following:

- adding the functions as methods to an existing or new service
- returning an error and have the caller log the event
- drop the log (in case it is already logged upstream)

PR #4816, opened 2023-08-11
[Merged by Bors] - Use ctxzap in GRPC services

poszu:
## Motivation
The start of GRPC calls was manually logged. It can be automated and enriched with contextual information about the GRPC connection using middleware. We already use Zap's [UnaryServerInterceptor](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware@v1.4.0/logging/zap#UnaryServerInterceptor) that appends a logger into the `Context` of a GRPC call. It can be later extracted with [ctxzap](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap).

## Changes
- log start of GRPC call with middleware
- use context-aware logging in GRPC services with ctxzap
- give time to send a response on AdminService/Recover before shutting down
- shutdown with os.Exit instead of panic on /Recover

## Test Plan
- existing tests pass
- added test for AdminService/Recover


api/grpcserver/admin_service.go
```diff
@@ -92,8 +94,9 @@ func (a AdminService) CheckpointStream(req *pb.CheckpointStreamRequest, stream p
 	}
 }
 
-func (a AdminService) Recover(_ context.Context, _ *pb.RecoverRequest) (*empty.Empty, error) {
-	a.logger.Panic("going to recover from checkpoint")
+func (a AdminService) Recover(ctx context.Context, _ *pb.RecoverRequest) (*empty.Empty, error) {
+	ctxzap.Info(ctx, "going to recover from checkpoint")
+	a.recover(ctx)
 	return &empty.Empty{}, nil
 }
```
fasmat: Should we remove this endpoint? Otherwise someone might call it and crash their node with it. We can re-add it when the functionality is available?

Or we could just make it a no-op.

node/node.go
```diff
@@ -1024,32 +1025,47 @@ func (app *App) startServices(ctx context.Context) error {
...
-		return grpcserver.NewNodeService(app.host, app.mesh, app.clock, app.syncer, cmd.Version, cmd.Commit, logger.WithName("Node")), nil
+		return grpcserver.NewNodeService(app.host, app.mesh, app.clock, app.syncer, cmd.Version, cmd.Commit), nil
 	case grpcserver.Admin:
-		return grpcserver.NewAdminService(app.db, app.Config.DataDir(), logger.WithName("Admin")), nil
+		return grpcserver.NewAdminService(app.db, app.Config.DataDir(), func(context.Context) {
+			go func() {
+				// Allow time for the response to be sent.
+				time.Sleep(time.Second)
+				os.Exit(0)
+			}()
```
fasmat: I am unsure about this, seems kinda dangerous having an endpoint that "crashes" the node when called. See also my other comment

api/grpcserver/transaction_service.go
```diff
@@ -236,7 +234,7 @@ func (s TransactionService) TransactionsStateStream(in *pb.TransactionsStateStre
 			// In order to read transactions, we first need to read layer blocks
 			layerObj, err := s.mesh.GetLayer(layer.LayerID)
 			if err != nil {
-				s.logger.With().Error("error reading layer data for updated layer", layer.LayerID, log.Err(err))
+				ctxzap.Error(stream.Context(), "error reading layer data for updated layer", layer.LayerID.Field().Zap(), zap.Error(err))
```
fasmat, line 237: Instead of extending `log.Field` with a `Zap()` method I think it would be better to have `LayerID` implement the `zapcore.ObjectMarshaler` interface.

This way we don't need to extend the (hopefully soon to be removed) `log` package and the field is evaluated lazily instead of eagerly (in case the log isn't even printed).

api/grpcserver/transaction_service.go
```diff
@@ -279,7 +277,7 @@ func (s TransactionService) TransactionsStateStream(in *pb.TransactionsStateStre
 						if in.IncludeTransactions {
 							tx, err := s.conState.GetMeshTransaction(txid)
 							if err != nil {
-								s.logger.Error("could not find transaction %v from layer %v: %v", txid, layer, err)
+								ctxzap.Error(stream.Context(), "could not find transaction from layer", txid.Field().Zap(), layer.Field().Zap(), zap.Error(err))
```
fasmat, line 280: Same as above, `txid` should in my oppinion implement `zapcore.ObjectMarshaler` instead of extending `log.Fied`

api/grpcserver/mesh_service.go
```diff
@@ -352,8 +347,17 @@ func (s MeshService) readLayer(ctx context.Context, layerID types.LayerID, layer
...
-		s.logger.With().Error("could not find activations from layer",
-			log.String("missing", fmt.Sprint(matxs)), layer.Index())
+		ctxzap.Error(
+			ctx,
+			"could not find activations from layer",
+			zap.Array("missing", zapcore.ArrayMarshalerFunc(func(ae zapcore.ArrayEncoder) error {
+				for _, atxid := range matxs {
+					ae.AppendString(atxid.ShortString())
+				}
+				return nil
```
fasmat: Since this seems to be used in multiple places it might make sense to add a general way to do this. In the file `ATXID` is defined we can add:

```golang
type ATXIDs []ATXID

func (ids ATXIDs) MarshalLogArray(enc zapcore.ArrayEncoder) error {
	for _, id := range ids {
		enc.AppendString(id.ShortString())
	}
	return nil
}
```

and then when using it:

```golang
ctxzap.Error(ctx, "could not find activations from layer",
	zap.Array("missing", types.ATXIDs(matxs)),
	layer.Index().Field().Zap(),
)
```

api/grpcserver/mesh_service.go
```diff
@@ -147,7 +139,12 @@ func (s MeshService) getFilteredActivations(ctx context.Context, startLayer type
 	// Look up full data
 	atxs, matxs := s.mesh.GetATXs(ctx, atxids)
 	if len(matxs) != 0 {
-		s.logger.Error("could not find activations %v", matxs)
+		ctxzap.Error(ctx, "could not find activations", zap.Array("matxs", zapcore.ArrayMarshalerFunc(func(enc zapcore.ArrayEncoder) error {
+			for _, atxid := range matxs {
+				enc.AppendString(atxid.ShortString())
+			}
+			return nil
+		})))
```
fasmat: See my other comment regarding a new type for `[]ATXID`

fasmat (changes requested):

api/grpcserver/transaction_service.go
poszu: Yes, 100% agree. I added this `Zap()` method as a shortcut before we migrate to not have to re-implement (and avoid having 2 implementations for log field serialization).

api/grpcserver/admin_service.go
poszu: I thought this is used... Anyway, I only changed the way it kills the app to:
- avoid unnecessary stack trace (it's not an error after all),
- allow controlling what happens (mostly for testing).

> someone might call it and crash their node with it

This is exactly the purpose of this endpoint, no? :D Btw, it's in the _private_ endpoint group.

fasmat: I'm not sure, I'd like to hear @countvonzero's opinion on it:

- Should we leave the endpoint killing the node
- Update it to a no-op (and maybe just log that not yet available)
- Remove it completely

EDIT: 🤦  I think I understood, we want the node to exit when we call that endpoint?

fasmat: Then OK, but I wouldn't pass in the `recover` function from the outside. You can set the field to

```golang
s.recover = func(context.Context) {
	go func() {
		// Allow time for the response to be sent.
		time.Sleep(time.Second)
		os.Exit(0)
	}()
}
```

in `NewAdminService` and overwrite it in the one test that calls it?

fasmat: One more thing: just exiting doesn't clean up `lock` files as far as I know? Does this maybe leave the node in a corrupted state or cause data loss if called at the wrong moment?

poszu: I'm not sure. I think file locks are released when locked file descriptors are closed. But what happens if the application simply exits/crashes? Is the behavior different on a `panic()` vs `os.Exit()`?

fasmat: Both exit immediately as far as I know. I'm not sure if the library we are using for lock files registers some handler that removes the file when the process exits 🤔.

poszu:
@fasmat is there anything else you require changes for? Regarding `Zap()` func, I tried different approaches but everything is worse. I think it's fine to have it in the transition period (as we discussed offline). Wdyt?

api/grpcserver/admin_service.go
fasmat: Since no issue came up so far regarding this, it should be OK.

fasmat (approved):
For now it's probably OK. There is already an issue in the backlog regarding removing the global logger: https://github.com/spacemeshos/go-spacemesh/issues/4751

Maybe we can update that issue to completely remove the `log` package?

poszu:
bors merge

poszu:
bors merge

poszu:
bors try

poszu:
@countvonzero could you please check my changes around recovery? I changed it to quit asynchronously after 1 second to give time to send a GRPC response. Do you see any potential problems with it?

poszu:
bors try

countvonzero:
> @countvonzero could you please check my changes around recovery? I changed it to quit asynchronously after 1 second to give time to send a GRPC response. Do you see any potential problems with it?

thanks for double-checking. there is no problem with Recover() RPC. it's also only used for systest.

poszu:
Bors merge

PR #4818, opened 2023-08-11
hare: disable atx grading

countvonzero:
## Motivation
hare is using a malfeasance proof received later than the proposal's activeset creation time.
disable it for now.


countvonzero:
bors try

PR #4819, opened 2023-08-11
[Merged by Bors] - separate layer average size before some point and after

dshulyak:
- all ballots before layer 8150 (can be changed) will use 0 layer size for eligibility. so they are eligible for one layer
- after layer 8150 they will start using 50
- this size will be respected on the validation in handler and tortoise
- proposer will check if it already made a ballot in this epoch and if it did, it will use that counter referenced in that ballot
- hdist was lowered to 10, in order for full tortoise to kick in faster if we will be experiencing problems with hare during the upgrade


countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #4820, opened 2023-08-11
[Merged by Bors] - hare stop atx grading at layer N

countvonzero:
## Motivation
hare is using a malfeasance proof received later than the proposal's activeset creation time.
disable it at layer N and after


dshulyak (approved):

countvonzero:
bors merge

PR #4833, opened 2023-08-13
[Merged by Bors] - tortoise: cap ballot layer by current according to the clock

dshulyak:
if there is a ballot with a significantly higher layer it will make tortoise stuck, because global threshold will overestimated


dshulyak:
bors try

countvonzero (approved):

dshulyak:
bors merge

PR #4839, opened 2023-08-14
[Merged by Bors] - fetch: check hash for received data

### LINKED ISSUE #3797: fetcher: validate that received object matches requested
there are leftovers of such validation, but it doesn't work now and in fact disabled, since some objects are not hashed as is.

it needs to be enabled somehow, otherwise we don't know if received object matches requested, which is secure since every handler validates received object. but not correct from the application dependencies point of view.

countvonzero:
## Motivation
Closes #3797

## Changes
- add SyncHandler in p2p for dropping peers
- in data handler's path, check the expected hash received on the sync path

### DIFF
--- a/blocks/handler.go
+++ b/blocks/handler.go
@@ -55,7 +55,7 @@ func NewHandler(f system.Fetcher, db *sql.Database, m meshProvider, opts ...Opt)
 }
 
 // HandleSyncedBlock handles Block data from sync.
-func (h *Handler) HandleSyncedBlock(ctx context.Context, peer p2p.Peer, data []byte) error {
+func (h *Handler) HandleSyncedBlock(ctx context.Context, expHash types.Hash32, peer p2p.Peer, data []byte) error {
 	logger := h.logger.WithContext(ctx)
 
 	var b types.Block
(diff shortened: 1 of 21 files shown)


malfeasance/handler.go
```diff
@@ -52,14 +52,18 @@ func NewHandler(
...
+func (h *Handler) HandleSyncedMalfeasanceProof(ctx context.Context, expHash types.Hash32, _ p2p.Peer, data []byte) error {
 	var p types.MalfeasanceProof
 	if err := codec.Decode(data, &p); err != nil {
 		numMalformed.Inc()
 		h.logger.With().Error("malformed message (sync)", log.Context(ctx), log.Err(err))
 		return errMalformedData
 	}
-	return validateAndSave(ctx, h.logger, h.cdb, h.edVerifier, h.cp, h.tortoise, &types.MalfeasanceGossip{MalfeasanceProof: p})
+	nodeID, err := validateAndSave(ctx, h.logger, h.cdb, h.edVerifier, h.cp, h.tortoise, &types.MalfeasanceGossip{MalfeasanceProof: p})
+	if err == nil && types.Hash32(nodeID) != expHash {
```
dshulyak, line 64: is it because we don't request them by hash, but instead all at once?

blocks/handler.go
```diff
@@ -65,6 +65,11 @@ func (h *Handler) HandleSyncedBlock(ctx context.Context, peer p2p.Peer, data []b
 	}
 	// set the block ID when received
 	b.Initialize()
+
+	if b.ID().AsHash32() != expHash {
+		return fmt.Errorf("fetched wrong block hash. want %s, got %s", expHash.ShortString(), b.ID().String())
```
dshulyak: it would be also optimal to drop such peer, there should be a special error for that (pubsub.ErrValidationReject)

dshulyak (approved):

malfeasance/handler.go
countvonzero: we request the list of malicious IDs (types.NodeID) and have a special protocol to request malfeasance proof for that ID. so for malfeasance proof, their hash on the sync path is the SmesherID

countvonzero:
bors try

countvonzero:
bors merge

PR #4840, opened 2023-08-14
[Merged by Bors] - block certificate: only sync latest certs in catch up sync

countvonzero:
## Motivation
currently for a new node joining the network, it needs to sync all block certificate from genesis.
there is no need. we can rely on full tortoise to advance the state during catch up sync and only sync cert right before it catch up to current layer.

### DIFF
--- a/mesh/mesh.go
+++ b/mesh/mesh.go
@@ -29,6 +29,8 @@ import (
 	"github.com/spacemeshos/go-spacemesh/system"
 )
 
+var ErrMissingBlock = errors.New("missing blocks")
+
 // Mesh is the logic layer above our mesh.DB database.
 type Mesh struct {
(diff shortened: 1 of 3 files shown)


dshulyak:
> we can rely on verifying tortoise to advance the state during catch up sync and only sync cert right before it catch up to current layer.

you meant on full tortoise?

dshulyak (approved):

countvonzero:
> > we can rely on verifying tortoise to advance the state during catch up sync and only sync cert right before it catch up to current layer.
> 
> you meant on full tortoise?

yes. corrected in PR description

countvonzero:
bors merge

countvonzero:
bors merge

PR #4845, opened 2023-08-14
[Merged by Bors] - Add peer info api

### LINKED ISSUE #4778: p2p, api: add api to return information about your peers
add an api endpoint that will return list of peers from host.Network().Peers() and additionally provide information if that peers has a special role (such as bootnode or direct peer). in future we can collect interesting stats about peers and add that to the same endpoint, so format should allow to extend it without significant pain in future

one option is below

```json
[{
  "id": "...",
  "addresses": "...",
  "uptime": "99h", // or maybe connection timestamp
  "tag": ["bootnode", "direct"] 
}]
```

piersy:
## Motivation
Closes #4778

Note that this is in a draft state because it needs https://github.com/spacemeshos/api/pull/258 to be merged and I guess a release of the API package, before it can be merged. I plan to remove the replace directive for the api in the mod file when, the api has been released.

## Changes
Adds a new streaming API endpoint to the node service to return info about connected peers.

Note this PR changes the PostProviderID field to be exported to allow for config to be easily serialised/deserialised in order to perform a deep copy.

## Test Plan
Added a test, the test doesn't check whether tags are since that part would make things a bit trickier. However through manual checking I can see that tags are being assigned.

```shell
$ grpc_cli call  localhost:9092 spacemesh.v1.NodeService.PeerInfo ''
connecting to localhost:9092
peers {
  id: "12D3KooWRkZMjGNrQfRyeKQC9U58cUwAfyQMtjNsupixkBFag8AY"
  connections {
    address: "/ip4/34.150.215.241/tcp/5000"
    uptime: "29.348584085s"
    outbound: true
  }
  tags: "bootnode"
}
Rpc succeeded with OK status
```

## DevOps Notes


activation/post.go
```diff
@@ -90,15 +90,15 @@ type PostSetupOpts struct {
 }
 
 type PostProviderID struct {
-	value *int64
+	InnerValue *int64
```
dshulyak: was change in this module necessary?

node/util_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/config"
+	"github.com/spacemeshos/go-spacemesh/log"
+)
+
+// NewNetwork creates a network of fully connected nodes.
+func NewNetwork(conf config.Config, l log.Log, size int) ([]*TestApp, func() error, error) {
```
dshulyak: please rename to NewTestNetwork

p2p/dhtdiscovery/discovery.go
```diff
@@ -292,3 +292,24 @@ func (d *Discovery) newDht(ctx context.Context, h host.Host, public, server bool
...
+
+func (d *Discovery) IsDirect(id peer.ID) bool {
+	return containsPeer(d.direct, id)
+}
+
+func (d *Discovery) IsBootnode(id peer.ID) bool {
+	return containsPeer(d.bootnodes, id)
+}
+
+func (d *Discovery) IsBackup(id peer.ID) bool {
```
dshulyak: i think information about backup is not necessary to expose. this is a list of saved connections

p2p/dhtdiscovery/discovery.go
```diff
@@ -292,3 +292,24 @@ func (d *Discovery) newDht(ctx context.Context, h host.Host, public, server bool
 	d.datastore = ds
 	return nil
 }
+
+func containsPeer(addrInfo []peer.AddrInfo, id peer.ID) bool {
```
dshulyak: please move all of this to p2p/host.go

api/grpcserver/node_service.go
```diff
@@ -106,6 +106,30 @@ func (s NodeService) NodeInfo(context.Context, *empty.Empty) (*pb.NodeInfoRespon
 	}, nil
 }
 
+func (s NodeService) PeerInfo(context.Context, *empty.Empty) (*pb.PeerInfoResponse, error) {
```
dshulyak: better to use AdminService for this. we don't set NodeService to 127.0.0.1 by default, i would prefer to not expose this info accidentally

api/grpcserver/node_service.go
```diff
@@ -106,6 +106,30 @@ func (s NodeService) NodeInfo(context.Context, *empty.Empty) (*pb.NodeInfoRespon
...
+				Outbound: c.Outbound,
+			}
+		}
+		peers[i] = &pb.PeerInfo{
+			Id:          info.ID.String(),
+			Connections: connections,
+			Tags:        info.Tags,
+		}
+	}
+	return &pb.PeerInfoResponse{
```
dshulyak: on some nodes we may have up to 10_000 peers. i would like to avoid memory spikes when reading peers from them, if it would be the case then it is better to switch to streaming

activation/post.go
piersy: This was to allow for serialisation/serialisation of the config in order to deep copy it in the test. Being able to serialize/deserialize config seems like it could  be a useful property to have.

p2p/dhtdiscovery/discovery.go
piersy: @dshulyak Are you referring just to containsPeer or all the IsXXX functions as well?

activation/post.go
fasmat: Our config is (de-)serializable using `viper`. This change might introduce an option to set an illegal ID as provider. This config will cause the node to fail at startup with the error that `-2` is not a legal provider ID:

```json
{
    "smeshing-opts-provider": -2
}
```

With the change this becomes possible (I think  - can't test at the moment):

```json
{
    "smeshing-opts-provider": {
        "InnerValue": -2
    }
}
```

node/util_test.go
```diff
@@ -0,0 +1,153 @@
...
+	err := gob.NewEncoder(&buf).Encode(conf)
+	if err != nil {
+		cleanup()
+		return nil, nil, err
+	}
+	marshaled := buf.Bytes()
+
+	for i := 0; i < size; i++ {
+		var c config.Config
+		err := gob.NewDecoder(bytes.NewBuffer(marshaled)).Decode(&c)
```
fasmat:
```suggestion
		c := conf
```

Encoding / decoding shouldn't be necessary here. All values that are changed between the copies are value types and the node afaik doesn't alter the config that is passed into it.

node/util_test.go
```diff
@@ -0,0 +1,153 @@
...
+		if err != nil {
+			cleanup()
+			return nil, nil, err
+		}
+		dir, err := os.MkdirTemp("", "")
```
fasmat: Use `t.TempDir()` instead to have the directory automatically deleted when the test ends (successfully or not)

node/util_test.go
```diff
@@ -0,0 +1,153 @@
...
+		c.FileLock = filepath.Join(c.DataDirParent, "LOCK")
+
+		app, err := NewApp(&c, l)
+		if err != nil {
+			cleanup()
+			return nil, nil, err
+		}
+		g.Go(func() error {
+			err := app.Start(ctx)
+			println(err)
```
fasmat:
```suggestion
			t.Logf("failed to start instance %d: %v", i, err)
```

You might need to check for `context.Canceled` before logging.

node/util_test.go
fasmat: Giving the function `tb testing.TB` as parameter also further clarifies that is only intended to be used in tests and allows access to `testing.T` in the function 🙂

node/util_test.go
```diff
@@ -0,0 +1,153 @@
...
+			if err != nil {
+				cleanup()
+				return nil, nil, err
+			}
+		}
+	}
+	return apps, cleanup, nil
+}
+
+func NewApp(conf *config.Config, l log.Log) (*App, error) {
```
fasmat: I think `NewApp` is confusing when `New` also exists. Since it seems to be intended to be used in tests only I think at least we should make it private?

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/spacemeshos/go-spacemesh
 
 go 1.19
 
+replace github.com/spacemeshos/api/release/go v1.18.0 => github.com/spacemeshos/api/release/go v1.18.1-0.20230811144450-baf8688a08b2
```
fasmat: Please do not use `replace`. This is not its intended use, you can directly update the dependency instead.

`replace` is intended to be able to forcefully overwrite transient dependencies when e.g. a critical vulnerability needs to be fixed and the dependency using it is slow to update. `replace` statements are not (always) applied to dependencies of `github.com/spacemeshos/go-spacemesh` and ignored by modules importing `github.com/spacemeshos/go-spacemesh`.

So any module importing `github.com/spacemeshos/go-spacemesh` needs to copy all `replace` statements listed here or it might not even compile, see also https://go.dev/ref/mod#go-mod-file-replace

EDIT: For its second use - locally replacing dependencies while working on multiple modules - a more convenient method was introduced with Go 1.18: `go work`, see also https://go.dev/doc/tutorial/workspaces

api/grpcserver/interface.go
```diff
@@ -54,9 +54,10 @@ type postSetupProvider interface {
 	Config() activation.PostConfig
 }
 
-// peerCounter is an api to get amount of connected peers.
-type peerCounter interface {
+// Peers is an api to get peer related info.
+type Peers interface {
```
fasmat: Interfaces are usually private in go-spacemesh if they only exist to be able to mock a dependency (i.e. are not intended to be used outside the package they are defined in).

Making the interface public and importing it in another package might introduce a dependency between packages we could avoid.

node/util_test.go
piersy: Thanks :+1:

node/util_test.go
piersy: I'm happy to change this, it will remove the need to modify PostProviderID, however it seems like it will be hard to enforce going forward since there doesn't seem to be any written standard around how to handle config. Let me know your thoughts?

node/util_test.go
piersy: I was thinking that somewhere down the line it might be useful to share this functionality between packages (I.E you may want a network for a test in a package which is not node) which is why I left the names Exported.

But happy to change this if you feel strongly about it, at that point it probably makes sense to also change all the names to be unexported.

go.mod
piersy: Sorry I meant to leave a note on the PR, the replace is temporary. I don't know what is the exact process we use when updating the API package but I assume we need to release it and update this PR to point to the release before it can be merged, at which point I planned to remove the replace.

node/util_test.go
piersy: So I've just re-worked the NewTestNetwork function to accept and use an instance of testing.T, it worked out nicely thanks!

node/util_test.go
piersy: Now that this is taking a testing.T parameter, do you still feel that a rename is necessary? @fasmat

activation/post.go
piersy: Yes I think it is possible. So I'll revert this modification and not copy the config.

node/util_test.go
piersy: So given the problem exposing the internals of PostProviderID, I've decided to remove the deep copy.

p2p/dhtdiscovery/discovery.go
dshulyak: yes i meant all IsXXX

api/grpcserver/admin_service.go
```diff
@@ -128,3 +130,27 @@ func (a AdminService) EventsStream(req *pb.EventStreamRequest, stream pb.AdminSe
 		}
 	}
 }
+
+func (a AdminService) PeerInfo(context.Context, *empty.Empty) (*pb.PeerInfoResponse, error) {
```
dshulyak: will repost so that comment is not lost

> on some nodes we may have up to 10_000 peers. i would like to avoid memory spikes when reading peers from them, if it would be the case then it is better to switch to streaming

go.mod
fasmat: You can directly go get into the `go-spacemesh` branch. Just make sure to merge & release the API changes before you merge and release the `go-spacemesh` changes 🙂

api/grpcserver/node_service.go
piersy: @dshulyak Switched to streaming now

go.mod
piersy: I felt like the use of replace just made it more clear that this wasn't a version that was intended to be merged, I.E. reducing the chance of accidentally merging a non released version. (And you spotted it immediately :D )

node/node.go
```diff
@@ -1100,10 +1100,10 @@ func (app *App) startAPIServices(ctx context.Context) error {
 		app.jsonAPIService.StartService(ctx, public...)
 	}
 	if app.grpcPublicService != nil {
-		app.grpcPublicService.Start()
+		<-app.grpcPublicService.Start()
```
fasmat: Hm... I don't think `Start()` should be asynchronous to begin with? Using a channel here to await the call can be easily missed.

Can you change this method to be synchronous or if its too much for this PR add a TODO/Issue?

In my opinion only this should be on a different go routine (ideally using `errgroup.Group`) in `grpc.go`:

```go
if err := s.GrpcServer.Serve(lis); err != nil {
	s.logger.Error("error stopping grpc server: %v", err)
}
```

And when stopping `errgroup.Group::Wait()` to make sure it stopped.

p2p/dhtdiscovery/discovery.go
piersy: Ok all moved

node/node.go
piersy: Yeah agreed. Doesn't look like too much to change it here.

piersy: Ok actually there's also the json server that uses the same approach but it's got nothing to do with this PR, I think it makes sense to handle them in one go, so I'll raise a ticket for that.

piersy: * https://github.com/spacemeshos/go-spacemesh/issues/4861

go.mod
fasmat: Another thing is if one forgets the `replace` statement it will be there until someone removes it, if one updates the dependency directly and forgets to update again when the API repo releases a new version, @dependabot will remind us 🙂

piersy: Noted, thanks!

node/util_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// To save an epoch of startup time, we bootstrap (meaning we manually set
+	// it) the beacon for epoch 2 so that in epoch 3 hare can start.
+	bootstrapEpoch := (types.GetEffectiveGenesis() + 1).GetEpoch()
+	bootstrapBeacon := types.Beacon{}
+	genesis := conf.Genesis.GenesisID()
+	copy(bootstrapBeacon[:], genesis[:])
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	g := errgroup.Group{}
```
fasmat: This limits any test to at most 20 seconds before all nodes are forcefully shut down. I think just a `cancelCtx` would fit better:

```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	var g errgroup.Group
```

node/util_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		app := NewApp(t, &c, l)
+		instanceIndex := i
+		g.Go(func() error {
+			err := app.Start(ctx)
+			if err != nil && !errors.Is(err, context.Canceled) {
+				t.Logf("failed to start instance %d: %v", instanceIndex, err)
+			}
+			return err
+		})
```
fasmat: Should the test be stopped as soon as the first node encounters some error? Maybe:

```go
g.Go(func() error {
	err := app.Start(ctx)
	if err != nil && !errors.Is(err, context.Canceled) {
		require.Fail(t, "node instantiation", "failed to start instance %d: %v", instanceIndex, err)
	}
	return nil
})
```

This would end the test (and shut down all nodes) immediately as soon as the first has some issue. (just a suggestion, maybe it makes sense to continue the test with one failing node...)

node/util_test.go
fasmat: Not necessary, but I feel like it is confusing to have both `NewApp` and `NewTestApp`, both of which are intended for testing. Maybe we can inline the later? It appears to only be used in one place and is essentially a one-liner.

node/util_test.go
piersy: That would be nice, but I don't think your suggestion would be safe, since [T.FailNow](https://pkg.go.dev/testing#T.FailNow) shouldn't be called from goroutines that are not the main go-routine.

fasmat: Then it is slightly more involved: `errgroup.WithContext` can create a sub-context that is canceled when the first go routine returns an error. If you pass that context to `app.Start` all nodes should be shut down as soon as the first returns a non-nil error.

fasmat (approved):

node/util_test.go
piersy: It would give the caller more flexibility, but also increases the burden on the caller.

Given that this is a test component I can't imagine a situation where you want to cancel the context for the network start because you already got what you needed through some other route. So I think it would boil down to callers providing a context with a timeout that they choose based on the size of the network they are creating.

On my machine the time taken for the network to start up is not that long, about 1.2 s  for a network of 3, 5s for a network of 10 and 16s for a network of 20. So in normal use I don't think that deadline will ever be reached, since this probably isn't a good fit for networks of size greater than 5 or 6, 20 caused my browsers to crash :) The deadline is there just to avoid waiting for 5 minutes if for some reason the network startup hangs.

node/util_test.go
piersy: Sounds good!

api/grpcserver/admin_service.go
piersy: @dshulyak I believe I've addressed this, could you take another look?

node/util_test.go
fasmat: The context isn't used correctly by `app.Start`. It determines not the timeout to start the node, but rather its whole lifetime. If `ctx` is canceled, all nodes shut down.

This means the way `NewTestNetwork` is implemented at the moment doesn't allow any tests that run longer than 20 seconds (including startup and the test itself), because by then all nodes will forcefully be shut down (by canceling the context), independent of if the test finished or not.

Changing it in the way I suggested it gives the nodes infinite lifetime (and they will still be shut down when the test ends by the `cancel()` call in the `t.Cleanup` function.

For the caller it makes no difference, besides that a test can be longer than 20 seconds 🙂

piersy: Hey, I misunderstood your original suggestion. Thank you for clarifying. I've removed the timeout :+1:

node/util_test.go
piersy: @fasmat done! :+1:

api/grpcserver/admin_service.go
```diff
@@ -128,3 +130,37 @@ func (a AdminService) EventsStream(req *pb.EventStreamRequest, stream pb.AdminSe
...
+			// There is no guarantee that the peers originally returned will still
+			// be connected by the time we call ConnectedPeerInfo.
+			if info == nil {
+				continue
+			}
+			connections := make([]*pb.ConnectionInfo, len(info.Connections))
+			for j, c := range info.Connections {
+				connections[j] = &pb.ConnectionInfo{
+					Address:  c.Address.String(),
+					Uptime:   c.Uptime.String(),
```
dshulyak: there is a protobuf type for duration, i think it might be nicer to use in other languages. if it will be ever needed

```
&pb.Event_PoetWaitRound{PoetWaitRound: &pb.EventPoetWaitRound{
			Current: current.Uint32(),
			Publish: publish.Uint32(),
			Wait:    durationpb.New(wait),
		}
```

dshulyak (approved):

piersy:
bors merge

PR #4846, opened 2023-08-14
[Merged by Bors] - sync: new protocol to fetch layer hash and certificate

### LINKED ISSUE #4674: sync: layer opininons protocol uses too much traffic
if i read it correctly it receives certificate from every peer, which explains unreasonably high traffic, even worth that unlike gossip - it will increase linearly with number of peers. see cyan (?) and deep purple for incoming/outgoing traffic on the screenshot below.

one solution is to switch into two step protocol:

1. request opinion

```go
type LayerOpinion struct {
	PrevAggHash types.Hash32
	Certified          bool
}
```  
and then execute step 2 if Certified is true, and it doesn't match our opinionn

2. ask certificate for a specific layer

image

countvonzero:
## Motivation
Closes #4674

## Changes
when a node poll layer opinions from peers, it gets back a response from each peer. 
```go
type LayerOpinion struct {
	PrevAggHash types.Hash32
	Certified *types.BlockID
}
```
node then go through each response and collect unique BlockIDs and fetch certificate for those blocks.

## before and after
in TestAddNodes, total amount of data a new node downloaded at layer 39 reduced from 10_647_927 bytes to 208_435 bytes

before
Screenshot from 2023-08-18 09-45-36

after
Screenshot from 2023-08-18 09-48-18

## testing
- systests TestAddNodes starts 28 nodes with new protocol and later added 2 nodes to sync with old protocol
- manually sync with mainnet with new protocol enabled

## note on deployment
- new nodes will fetch opinions with new protocol from nodes that support it. and with old protocol from nodes that don't support it.
- in the case of rollback, set syncer config `UseNewProtocol` to false
```
		Sync: syncer.Config{
                        ...
			UseNewProtocol:   false,
		},
```
all nodes will then only request/serve the old protocol


countvonzero:
bors try

fetch/wire_types.go
```diff
@@ -137,3 +137,35 @@ func (lo *LayerOpinion) MarshalLogObject(encoder log.ObjectEncoder) error {
 	}
 	return nil
 }
+
+type LayerOpinion2 struct {
+	PrevAggHash types.Hash32
+	Certified   bool
+	CertBlock   types.BlockID
```
dshulyak: i think `CertBlock *types.BlockID` is better. so if not certificate it is nil, otherwise not nil

config/mainnet.go
```diff
@@ -134,6 +134,7 @@ func MainnetConfig() Config {
 			Interval:         time.Minute,
 			EpochEndFraction: 0.8,
 			MaxStaleDuration: time.Hour,
+			UpdateLayer:      10000000,
```
countvonzero: reminder: update for release

dshulyak:
> in the case of rollback, simply set --sync-update-layer to a big number via config.

could you please add how json config changes in pr description. tbh such options should not be even added to command line as it will be harder to drop them later. but i assume this needed for tests

syncer/state_syncer.go
```diff
@@ -132,32 +139,77 @@ func (s *Syncer) needCert(ctx context.Context, lid types.LayerID) (bool, error)
...
+		if opn.CertBlock != types.EmptyBlockID {
+			s.dataFetcher.RegisterPeerHashes(opn.Peer(), []types.Hash32{opn.CertBlock.AsHash32()})
+		}
+	}
+	opinionLayer.Set(float64(lid))
+	return result, certs, nil
+}
+
+func (s *Syncer) fetchOpinions(ctx context.Context, lid types.LayerID) ([]*peerOpinion, []*types.Certificate, error) {
+	if s.ticker.CurrentLayer() >= types.LayerID(s.cfg.UpdateLayer) {
```
dshulyak: i think that upgrade path without configured layer is more convenient and robust. if there is a regression we will be able to notice it immediately

i am referring to a upgrade path, where node checks if it has any peers with new protocol and if so then uses new

syncer/state_syncer.go
```diff
@@ -132,32 +139,77 @@ func (s *Syncer) needCert(ctx context.Context, lid types.LayerID) (bool, error)
...
+	needCert, err := s.needCert(ctx, lid)
+	if err != nil {
+		return nil, nil, err
+	}
+	opinions, certs, err := s.dataFetcher.PollLayerOpinions2(ctx, lid, needCert)
+	if err != nil {
+		s.logger.WithContext(ctx).With().Warning("failed to fetch opinions", lid, log.Err(err))
+		return nil, nil, fmt.Errorf("PollLayerOpinions: %w", err)
+	}
+	var result []*peerOpinion
```
dshulyak: can be initialized with size/cap

fetch/mesh_data.go
```diff
@@ -269,3 +277,58 @@ func (f *Fetch) PeerMeshHashes(ctx context.Context, peer p2p.Peer, req *MeshHash
...
+	}
+
+	out := make(chan *types.Certificate, 1)
+	for _, peer := range peers {
+		done := make(chan error, 1)
+		okCB := func(data []byte) {
+			var peerCert types.Certificate
+			err = codec.Decode(data, &peerCert)
+			if err != nil {
+				defer close(done)
```
dshulyak: consumer reads exactly one event from done and then continues, same for output. doesn't look good with this defers

fetch/mesh_data.go
```diff
@@ -269,3 +277,58 @@ func (f *Fetch) PeerMeshHashes(ctx context.Context, peer p2p.Peer, req *MeshHash
 		return nil, ctx.Err()
 	}
 }
+
+func (f *Fetch) GetCert(ctx context.Context, lid types.LayerID, bid types.BlockID, peers []p2p.Peer) (*types.Certificate, error) {
```
dshulyak, line 287: any reason not to reuse hs/1 with hints? seems just inconsistent

fetch/mesh_data.go
```diff
@@ -269,3 +277,58 @@ func (f *Fetch) PeerMeshHashes(ctx context.Context, peer p2p.Peer, req *MeshHash
...
+		done := make(chan error, 1)
+		okCB := func(data []byte) {
+			var peerCert types.Certificate
+			err = codec.Decode(data, &peerCert)
+			if err != nil {
+				defer close(done)
+				done <- err
+				return
+			}
+			if peerCert.BlockID != bid {
```
dshulyak, line 313: this check also somewhat inconsistent now, other messages check it in handlers

syncer/data_fetch.go
```diff
@@ -383,6 +387,103 @@ func (d *DataFetch) receiveOpinions(ctx context.Context, req *opinionRequest, pe
...
+					if err != nil {
+						logger.With().Warning("failed to get cert from peers", log.Err(err))
+						continue
+					}
+					certs = append(certs, cert)
+				}
+			}
+			return req.response.opinions, certs, candidateErr
+		case <-ctx.Done():
+			logger.Warning("request timed out")
```
dshulyak: does it need to be a warning?

syncer/data_fetch.go
```diff
@@ -383,6 +387,103 @@ func (d *DataFetch) receiveOpinions(ctx context.Context, req *opinionRequest, pe
...
+					if _, ok := peerCerts[opns.CertBlock]; !ok {
+						peerCerts[opns.CertBlock] = []p2p.Peer{}
+					}
+					peerCerts[opns.CertBlock] = append(peerCerts[opns.CertBlock], opns.Peer())
+					d.fetcher.RegisterPeerHashes(opns.Peer(), []types.Hash32{opns.CertBlock.AsHash32()})
+				}
+				for bid, bidPeers := range peerCerts {
+					cert, err := d.fetcher.GetCert(ctx, lid, bid, bidPeers)
+					if err != nil {
+						logger.With().Warning("failed to get cert from peers", log.Err(err))
```
dshulyak: does it need to be a warning?

dshulyak (approved):
i would consider doing upgrade without configured layer, it seems much more convenient for testing on mainnet in this case.

i think in system like this peers are expected to not respond or disconnect randomly. printing warnings doesn't make much sense, i would replace that with prometheus counter as it is mainly useful for operations

syncer/state_syncer.go
countvonzero: used PeerStore().GetProtocols() to make upgrade path automatic.
peer will first fetch from v2 peers, and then fetch from peers that don't support v2
as this is not on critical path, i did this sequentially instead of concurrently for simpler code.

fetch/mesh_data.go
countvonzero: thanks! fixed

fetch/mesh_data.go
countvonzero: this is true for generic hash fetches. (atx/block/proposal/ballot/tx).
however, certificate doesn't go through that path. it's specifically requested, as block certificate doesn't have an ID associated with it. keeping it and added comment.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

PR #4849, opened 2023-08-15
[Merged by Bors] - fetch: requests got lost when there are no peers

countvonzero:
## Motivation
requests are not restored back to the queue when there were no peers to dispatch to.

## Changes
- restore requests when there are no peers
- push mesh to still make progress even when there are missing blocks
- increase frequency of state sync

## testing
i manually test against mainnet, artificially causing GetPeers() to return zero peers.
mesh was making very slow progress because it aborts as soon as it finds a missing block, not executing the next block even when it's available.

### DIFF
--- a/fetch/fetch.go
+++ b/fetch/fetch.go
@@ -437,6 +437,25 @@ func (f *Fetch) send(requests []RequestMessage) {
 	}
 
 	peer2batches := f.organizeRequests(requests)
+	if len(peer2batches) == 0 {
+		f.mu.Lock()
+		defer f.mu.Unlock()
+		for _, msg := range requests {
+			if req, ok := f.ongoing[msg.Hash]; ok {
+				f.logger.WithContext(req.ctx).With().Info("delaying hash request",
+					log.Stringer("hash", msg.Hash),
+					log.String("hint", string(msg.Hint)),
+				)
+				f.unprocessed[msg.Hash] = req
+			} else {
+				f.logger.With().Error("ongoing request missing",
+					log.Stringer("hash", msg.Hash),
+					log.String("hint", string(msg.Hint)),
+				)
+			}
+		}
+		return
+	}
 
 	for peer, peerBatches := range peer2batches {
 		for _, reqs := range peerBatches {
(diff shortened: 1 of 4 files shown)


mesh/mesh_test.go
```diff
@@ -404,6 +404,20 @@ func TestProcessLayer(t *testing.T) {
...
 			},
 		},
+		{
+			"missing but can make progress",
+			[]call{
+				{
+					updates: rlayers(
+						rlayer(start, rblock(idg("1"), fixture.Valid(), fixture.Data())),
+						rlayer(start+1, rblock(idg("2"), fixture.Valid(), fixture.Data())),
+						rlayer(start+2, rblock(idg("3"), fixture.Valid())),
```
dshulyak, line 414: maybe extend it with one more block with data? and one more call that both got applied

fetch/fetch.go
```diff
@@ -437,6 +437,25 @@ func (f *Fetch) send(requests []RequestMessage) {
...
+	if len(peer2batches) == 0 {
+		f.mu.Lock()
+		defer f.mu.Unlock()
+		for _, msg := range requests {
+			if req, ok := f.ongoing[msg.Hash]; ok {
+				f.logger.WithContext(req.ctx).With().Info("delaying hash request",
+					log.Stringer("hash", msg.Hash),
+					log.String("hint", string(msg.Hint)),
+				)
+				f.unprocessed[msg.Hash] = req
```
dshulyak: so GetBlocks will not return immediately with error if no peers, but instead we will do the request when peers are available?

dshulyak (approved):

fetch/fetch.go
countvonzero: yes. essentially put those requests back to the queue

countvonzero: do you prefer to fail all in that case?

dshulyak: this approach is fine for me too, hard to tell what is better

countvonzero: actually i prefer returning error right away. it's easier for debugging (and avoiding future bug)

mesh/mesh_test.go
countvonzero: done

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

PR #4851, opened 2023-08-16
[Merged by Bors] - serve malfeasance proof over grpc

countvonzero:
## Motivation
as title
depends on https://github.com/spacemeshos/api/pull/259

## Changes
- add MeshService.MalfeasanceStream and MeshService:MalfeasanceQuery
- add malfeasance proof to ActivationService.Get
- add own malfeasance proof to node event update

## output
```
{
  "proof": {
    "smesherId": {
      "id": "8JhffQsAWuCBQ8YRWrLpOj8tesNx/SKV611TDbPnXCA="
    },
    "layer": {
      "number": 8112
    },
    "kind": "MALFEASANCE_HARE",
    "debugInfo": "generate layer: 8112\nsmesher id: f0985f7d0b005ae08143c6115ab2e93a3f2d7ac371fd2295eb5d530db3e75c20\ncause: smesher published multiple hare messages in layer 8112 round 3\n1st message hash: 9c40b8789bad681d25081f85af7f9a4af87c5d70ac0637934b4e6fc935b9566c\n1st message signature: efdb895917b2303442a3b0c8f455c5cea79f8a2969af2c7f5b3bbbbfa09a77441e3d886e049f39dbf4dc0d49e78537fe3035de2d25d9d295f02c4928677ea107\n2nd message hash: d005c8ff6ece20cb87acc7ea97d5473559a3e197f9a66621505383132515a2ae\n2nd message signature: f3c014fd0f1353896222f22ee71b7fb2a10623394d4905b8ea447ac2e45350e487de04df71448cb19012b897728bb8580499b3e16a5edb546333b432a927d301\n"
  }
}
```

### DIFF
--- a/api/grpcserver/activation_service.go
+++ b/api/grpcserver/activation_service.go
@@ -2,6 +2,7 @@ package grpcserver
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/golang/protobuf/ptypes/empty"
@@ -10,7 +11,9 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/sql"
 )
 
 type activationService struct {
@@ -42,12 +45,28 @@ func (s *activationService) Get(ctx context.Context, request *pb.GetRequest) (*p
 	atxId := types.ATXID(types.BytesToHash(request.Id))
 	atx, err := s.atxProvider.GetFullAtx(atxId)
 	if err != nil || atx == nil {
-		s.logger.With().Debug("failed to get the ATX", log.Err(err), log.Stringer("id", atxId))
+		s.logger.With().Error("failed to get ATX",
+			log.Stringer("atx id", atxId),
+			log.Err(err),
+		)
 		return nil, status.Error(codes.NotFound, "id was not found")
 	}
-	return &pb.GetResponse{
+	proof, err := s.atxProvider.GetMalfeasanceProof(atx.SmesherID)
+	if err != nil && !errors.Is(err, sql.ErrNotFound) {
+		s.logger.With().Error("failed to get malfeasance proof",
+			log.Stringer("smesher", atx.SmesherID),
+			log.Stringer("id", atxId),
+			log.Err(err),
+		)
+		return nil, status.Error(codes.NotFound, "id was not found")
+	}
+	resp := &pb.GetResponse{
 		Atx: convertActivation(atx),
-	}, nil
+	}
+	if proof != nil {
+		resp.MalfeasanceProof = events.ToMalfeasancePB(atx.SmesherID, proof, false)
+	}
+	return resp, nil
 }
 
 func (s *activationService) Highest(ctx context.Context, req *empty.Empty) (*pb.HighestResponse, error) {
--- a/api/grpcserver/activation_service.go
+++ b/api/grpcserver/activation_service.go
@@ -45,7 +45,7 @@ func (s *activationService) Get(ctx context.Context, request *pb.GetRequest) (*p
 	atxId := types.ATXID(types.BytesToHash(request.Id))
 	atx, err := s.atxProvider.GetFullAtx(atxId)
 	if err != nil || atx == nil {
-		s.logger.With().Error("failed to get ATX",
+		s.logger.With().Debug("failed to get ATX",
 			log.Stringer("atx id", atxId),
 			log.Err(err),
 		)
--- a/api/grpcserver/mesh_service.go
+++ b/api/grpcserver/mesh_service.go
@@ -2,17 +2,21 @@ package grpcserver
 
 import (
 	"context"
+	"encoding/hex"
+	"errors"
 	"fmt"
 	"time"
 
 	pb "github.com/spacemeshos/api/release/go/spacemesh/v1"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/datastore"
 	"github.com/spacemeshos/go-spacemesh/events"
 	"github.com/spacemeshos/go-spacemesh/log"
+	"github.com/spacemeshos/go-spacemesh/sql"
 )
 
 // MeshService exposes mesh data such as accounts, blocks, and transactions.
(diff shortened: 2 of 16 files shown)


events/events.go
```diff
@@ -227,3 +244,66 @@ func emitUserEvent(help string, failure bool, details pb.IsEventDetails) {
...
+		DebugInfo: MalfeasanceInfo(smesher, mp),
+	}
+	if includeProof {
+		data, _ := codec.Encode(mp)
+		result.Proof = data
+	}
+	return result
+}
+
+func MalfeasanceInfo(smesher types.NodeID, mp *types.MalfeasanceProof) string {
```
countvonzero: @noamnelke this is the debug info provided so far. anything i should add?

api/grpcserver/activation_service.go
```diff
@@ -42,12 +45,28 @@ func (s *activationService) Get(ctx context.Context, request *pb.GetRequest) (*p
 	atxId := types.ATXID(types.BytesToHash(request.Id))
 	atx, err := s.atxProvider.GetFullAtx(atxId)
 	if err != nil || atx == nil {
-		s.logger.With().Debug("failed to get the ATX", log.Err(err), log.Stringer("id", atxId))
+		s.logger.With().Error("failed to get ATX",
```
fasmat: I'm not sure we should "ERROR" here. Any ATX can be requested via GRPC.

The local node could not yet be fully synced - then this warrants no error I think. The remote node could request an invalid ID, which also isn't an error on the local nodes side, etc.

malfeasance/handler_test.go
```diff
@@ -53,7 +53,7 @@ func TestHandler_HandleMalfeasanceProof_multipleATXs(t *testing.T) {
 	sigVerifier, err := signing.NewEdVerifier()
 	require.NoError(t, err)
 
-	h := malfeasance.NewHandler(datastore.NewCachedDB(db, lg), lg, "self", mcp, sigVerifier, trt)
+	h := malfeasance.NewHandler(datastore.NewCachedDB(db, lg), lg, "self", types.NodeID{}, mcp, sigVerifier, trt)
```
fasmat: NIT:

```suggestion
	h := malfeasance.NewHandler(datastore.NewCachedDB(db, lg), lg, "self", types.EmptyNodeID, mcp, sigVerifier, trt)
```

api/grpcserver/activation_service.go
```diff
@@ -42,12 +45,28 @@ func (s *activationService) Get(ctx context.Context, request *pb.GetRequest) (*p
...
+			log.Err(err),
+		)
+		return nil, status.Error(codes.NotFound, "id was not found")
+	}
+	resp := &pb.GetResponse{
 		Atx: convertActivation(atx),
-	}, nil
+	}
+	if proof != nil {
+		resp.MalfeasanceProof = events.ToMalfeasancePB(atx.SmesherID, proof, false)
```
fasmat, line 67: Having `ToMalfeasancePB` in the `events` package feels wrong; I wouldn't think of looking for this function there if I would need it, yet I am also not sure where else to put it.

api/grpcserver/mesh_service.go
```diff
@@ -610,3 +614,68 @@ func (s MeshService) EpochStream(req *pb.EpochStreamRequest, stream pb.MeshServi
 	)
 	return nil
 }
+
+func (s MeshService) MalfeasanceQuery(ctx context.Context, req *pb.MalfeasanceRequest) (*pb.MalfeasanceResponse, error) {
+	var id types.NodeID
+	if len(req.SmesherHex) == 0 {
```
fasmat: See my comment in the API PR: I think we shouldn't have two fields in the request encoding the same information.

api/grpcserver/mesh_service.go
```diff
@@ -610,3 +614,68 @@ func (s MeshService) EpochStream(req *pb.EpochStreamRequest, stream pb.MeshServi
...
+		Proof: events.ToMalfeasancePB(id, proof, req.IncludeProof),
+	}, nil
+}
+
+func (s MeshService) MalfeasanceStream(req *pb.MalfeasanceStreamRequest, stream pb.MeshService_MalfeasanceStreamServer) error {
+	sub := events.SubscribeMalfeasance()
+	if sub == nil {
+		return status.Errorf(codes.FailedPrecondition, "event reporting is not enabled")
+	}
+	eventch, fullch := consumeEvents[events.EventMalfeasance](stream.Context(), sub)
```
fasmat, line 642: NIT (and unrelated to this PR):

I don't think it makes sense to have `eventch` and `fullch` as return value in `consumeEvents` if the buffer in `eventch` runs full, the listener doesn't read from it. Why should the listener read from `fullch`?

The buffer is also huge -> 32768 events. So if `fullch` is closed the listener hasn't read events for quite some time.

api/grpcserver/mesh_service.go
```diff
@@ -610,3 +614,68 @@ func (s MeshService) EpochStream(req *pb.EpochStreamRequest, stream pb.MeshServi
...
+		return status.Error(codes.Internal, err.Error())
+	}
+
+	for {
+		select {
+		case <-stream.Context().Done():
+			return nil
+		case <-fullch:
+			return status.Errorf(codes.Canceled, "buffer is full")
+		case ev := <-eventch:
```
fasmat, line 668: Also depending on how many proofs have been stored in the DB already the first one processed might be quite old (and the buffer might be quite full already). Will this return the same malfeasance proof twice then?

api/grpcserver/mesh_service_test.go
```diff
@@ -0,0 +1,228 @@
...
+	require.NoError(t, codec.Decode(resp.Proof.Proof, &got))
+	require.Equal(t, *proof, got)
+
+	req.IncludeProof = false
+	resp, err = client.MalfeasanceQuery(context.Background(), req)
+	require.NoError(t, err)
+	require.Empty(t, resp.Proof.Proof)
+}
+
+func TestMeshService_MalfeasanceStream(t *testing.T) {
```
fasmat, line 179: Would it make sense to extend this or create a new test that persists malfeasance proofs while they are streamed?

I think it would make sense to have such a test to check for races and if every proof is only streamed once?

api/grpcserver/activation_service.go
countvonzero: you are right. i was possessed when i changed that.

malfeasance/handler_test.go
countvonzero: done thru out

api/grpcserver/activation_service.go
countvonzero: sorry. i know. but it's better for grpcserver -> events than the other way around. 
events pkg already deals with api's pb messages.

api/grpcserver/mesh_service.go
countvonzero: i see this more as a safeguard to stopgap a programming error from OOM e.g. doing something blocking when processing an event. not sure about the history here.

api/grpcserver/mesh_service.go
countvonzero: in the error handling around streaming, the receiver should already be prepared to de-dupe. it's not a problem if they get the same proof twice. it's more of a problem if the a proof is missed and not sent. that's why i started receiving events before iterating db. the code only emit the event after a proof is saved in the db. this order guarantees more that proof won't be missed.

api/grpcserver/mesh_service_test.go
countvonzero: persisting malfeasance doesn't cause event to trigger. the code there also calls `events.ReportMalfeasance(id, proof)`, like i did in this test. i don't see there is a difference btwn what you are suggesting and the test is currently doing.

api/grpcserver/mesh_service.go
fasmat: Hm... I am not sure if we are doing it correctly. I see 2 possible approaches:

- Events can be dropped / there is no guarantee that every event will be sent to every consumer. This is the "safe" approach since busy consumers won't cause the producer to hang because they are not listening
- Events cannot be dropped / every event is guaranteed to be sent to every consumer. This way a malfunctioning consumer might cause the producer to get stuck because it cannot send more events.

Both approaches can give some leeway to busy consumers by using buffered channels. Here it looks like the first approach was taken (if the consumer isn't listening eventually the producer will stop sending). Just that the consumer will also not handle the events any more that were buffered for it when it becomes ready again. Additionally his happens not immediately: the channels are read in the same select statement, so some events might still be processed before the consumer gets the message that no events will be coming any more and stops reading from `eventch`.

countvonzero: other stream APIs are implemented this way as well. if you feel strongly about changing the behavior, maybe file an issue for what you want to fix?

dshulyak: the idea is that stream subscriber should never be able to block consensus code and we shouldn't silently drop events either. so once we see that subscriber is not reading fast enough error is returned from api and channel is unregistered

api/grpcserver/mesh_service.go
```diff
@@ -610,3 +614,59 @@ func (s MeshService) EpochStream(req *pb.EpochStreamRequest, stream pb.MeshServi
 	)
 	return nil
 }
+
+func (s MeshService) MalfeasanceQuery(ctx context.Context, req *pb.MalfeasanceRequest) (*pb.MalfeasanceResponse, error) {
+	parsed, err := hex.DecodeString(req.SmesherHex)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	id := types.BytesToNodeID(parsed)
```
fasmat, line 627: I think you still need to check the length of `parsed` otherwise missing bytes will be filled with zeroes and extra bytes silently dropped.

api/grpcserver/mesh_service.go
fasmat: Yes, I understand. I think there is an issue on the consumer side with this approach where overflows of the `eventch` are not handled deterministically:

Both `eventch` and `fullch` are always read from in the same `select` statement (not just here). If `fullch` is closed the cosumer might read anything between 0 to 10 more events (more than 10 become extremely unlikely) from `eventch` before they actually stop processing events.

fasmat (approved):

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #4864, opened 2023-08-16
[Merged by Bors] - Update to go 1.20

fasmat:
## Motivation
Go 1.19 is now EOL and won't be updated any more. This PR updates our codebase to 1.20.

## Changes
- update all uses of go to 1.20: `go.mod`, `Dockerfile`s, documentation
- instead of top-level deprecated `math/rand` functions use `crypto/rand` or an instance of `*rand.Rand` instead
- use `errors.Join` instead of `github.com/hashicorp/go-multierror`
- fix `staticcheck` and `lint` complaints

## TODO

## DevOps Notes


fasmat:
bors try

fasmat:
I would like to also upgrade to 1.21 after this is merged, but will do it in 2 steps to keep the changes easier to review and test.

rand/rand.go
```diff
@@ -7,11 +7,6 @@ import (
 	"time"
 )
 
-func init() {
-	// initialize global pseudo random generator
```
poszu, line 11: Could you explain why is it removed? Is `math.rand` seeded automatically now?

Btw, do we really need this package?

poszu (approved):

rand/rand.go
fasmat: From https://golang.org/doc/go1.20#math/rand :

> The [math/rand](https://golang.org/pkg/math/rand/) package now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value, and the top-level [Seed](https://golang.org/pkg/math/rand/#Seed) function has been deprecated. Programs that need a reproducible sequence of random numbers should prefer to allocate their own random source, using rand.New(rand.NewSource(seed)).
> ...
> The top-level [Read](https://golang.org/pkg/math/rand/#Read) function has been deprecated. In almost all cases, [crypto/rand.Read](https://golang.org/pkg/crypto/rand/#Read) is more appropriate.

fasmat: I removed our own `rand` package. It was only used in a handful of places (only tests) and could easily be replaced with `math/rand` and `crypto/rand`

fasmat:
bors merge

PR #4865, opened 2023-08-17
[Merged by Bors] - Update to go 1.21

fasmat:
## Motivation
This updates our codebase to Go 1.21.

## Changes
- Update all uses of go to 1.21: `go.mod`, `Dockerfile`s, documentation
- replace custom max & min functions with builtin `max` and `mins`, use `clear` to clear slices & maps
- fix `staticcheck` and `lint` complaints

## DevOps Notes


mesh/mesh.go
```diff
@@ -291,8 +291,8 @@ func (msh *Mesh) ProcessLayer(ctx context.Context, lid types.LayerID) error {
 	results := msh.trtl.Updates()
 	pending := msh.pendingUpdates.min != 0
 	if len(results) > 0 {
-		msh.pendingUpdates.min = types.MinLayer(msh.pendingUpdates.min, results[0].Layer)
-		msh.pendingUpdates.max = types.MaxLayer(msh.pendingUpdates.max, results[len(results)-1].Layer)
+		msh.pendingUpdates.min = min(msh.pendingUpdates.min, results[0].Layer)
+		msh.pendingUpdates.max = min(msh.pendingUpdates.max, results[len(results)-1].Layer)
```
poszu: It should be `max`:
```suggestion
		msh.pendingUpdates.max = max(msh.pendingUpdates.max, results[len(results)-1].Layer)
```

hare/broker.go
```diff
@@ -395,13 +395,13 @@ func (b *Broker) Register(ctx context.Context, id types.LayerID) (chan any, *Eli
 	if len(b.outbox) >= b.limit {
 		// unregister the earliest layer to make space for the new layer
 		// cannot call unregister here because unregister blocks and this would cause a deadlock
-		min := types.LayerID(0)
+		minimum := types.LayerID(0)
 		for lid := range b.outbox {
-			min = types.MinLayer(min, lid)
+			minimum = min(minimum, lid)
 		}
```
poszu: Could it be replaced with `slices.Min`?

poszu (changes requested):

hare/broker.go
fasmat: Maybe with https://pkg.go.dev/slices#MinFunc. `types.MinLayer` returns the minimum non-zero LayerID, this might also be the cause of the failing tests.

go.mod
```diff
@@ -173,9 +173,8 @@ require (
 	github.com/prometheus/procfs v0.10.1 // indirect
 	github.com/pyroscope-io/dotnetdiag v1.2.1 // indirect
 	github.com/quic-go/qpack v0.4.0 // indirect
-	github.com/quic-go/qtls-go1-19 v0.3.3 // indirect
-	github.com/quic-go/qtls-go1-20 v0.2.3 // indirect
-	github.com/quic-go/quic-go v0.36.4 // indirect
+	github.com/quic-go/qtls-go1-20 v0.3.2 // indirect
```
fasmat: The downgrade here makes me suspicious, but `github.com/quic-go/quic-go` doesn't need `github.com/quic-go/qtls-go1-20` when compiled for 1.21, so it might not matter...

hare/broker.go
```diff
@@ -395,13 +395,13 @@ func (b *Broker) Register(ctx context.Context, id types.LayerID) (chan any, *Eli
 	if len(b.outbox) >= b.limit {
 		// unregister the earliest layer to make space for the new layer
 		// cannot call unregister here because unregister blocks and this would cause a deadlock
-		min := types.LayerID(0)
+		minimum := types.LayerID(0)
 		for lid := range b.outbox {
-			min = types.MinLayer(min, lid)
+			minimum = min(minimum, lid)
```
poszu: `types.MinLayer()` isn't equivalent to `min()`. The former returns the lower **non-zero** value.

go.mod
poszu: IDK, it happened when I updated to go-libp2p v0.30.0 🤷

common/types/layer.go
```diff
@@ -237,3 +237,15 @@ func NewLayer(layerIndex LayerID) *Layer {
 		blocks:  make([]*Block, 0, 3),
 	}
 }
+
+// MinLayer returns the lower non-zero layer.
+func MinLayer(i, j LayerID) LayerID {
```
fasmat: I intentionally removed that function because it is confusing that `MinLayer` doesn't return the minimum if it is zero and often both `i` and `j` are guaranteed non-zero; I just didn't have time yet to fix the code where it isn't.

poszu: Then I suggest opening an issue to remove `types.MinLayer()` and to refactor the code as it's not really part of bumping to a newer go version.

fasmat: I disagree; this PR also replaces `MaxLayer` with `max`. I intentionally left the PR as draft as it was neither urgent nor important and I wanted to come back to it next week 🙂

poszu: IMHO, it would be best to not mix go version bump with code refactoring at all. This PR should just be an update of go.mod, CI, Dockerfile etc. without any changes in src code, which should happen in a separate PR 🤷

fasmat: No, why update to go 1.21 if we are not using new features? Please don't take over other peoples PRs and redefine their scope 😞

fasmat: This PR is neither important nor urgent... We have a blacklog of 200 issues ready to be worked on 🙂

poszu: > No, why update to go 1.21 if we are not using new features?

Because it's useful for other reasons too, for example, it would unblock https://github.com/spacemeshos/poet/issues/363. What's the problem with refactoring the remaining parts of the code later?

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

systest/tests/common.go
```diff
@@ -13,6 +13,7 @@ import (
 	pb "github.com/spacemeshos/api/release/go/spacemesh/v1"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
+	_ "go4.org/unsafe/assume-no-moving-gc"
```
poszu: Could you explain why is it needed?

fasmat: https://github.com/spacemeshos/go-spacemesh/actions/runs/6075215589/job/16480953573

System tests fail because the version of that package we are using is not yet updated for Go 1.21, its a transitive dependency and without the forced import it cannot be updated (`go mod tidy` removes it from `go.mod` without the import).

fasmat: It still fails though, I'll look into it tomorrow again.

poszu: I cannot figure out which of our dependencies uses it. Do you know?

fasmat: `github.com/francoispqt/gojay@v1.2.13` -> `cloud.google.com/go@v0.37.0` -> `golang.org/x/build@v0.0.0-20190111050920-041ab4dc3f9d` -> `go4.org@v0.0.0-20180809161055-417644f6feb5`

`github.com/francoispqt/gojay@v1.2.13` is used by `go-libp2p` and some of its dependencies (like `ipfs` and `quic-go`).

Unfortunately it seems like the library has been abandoned around 2019. It uses some very outdated packages (like the ones listed above) and has known security issues 😞, but the author hasn't responded to any PRs or issues in a few years.

We can probably "fix" it by setting the environment variable that skips the check at the risk that if we update to 1.22 or any later version of Go and the GC behavior changes we see crashes.

fasmat: My reasoning for this is:

- In Go 1.21 the check still passes: see (https://github.com/go4org/unsafe-assume-no-moving-gc/blob/e7c30c78aeb2bf1ab602f2e789b86db8d9942388/check.go#L19-L25 and https://github.com/golang/go/blob/c19c4c566c63818dfd059b352e52c4710eecf14d/src/runtime/mgc.go#L170-L172)
- If go would resolve the version I specified in `go.mod` the check would pass, but for some reason it uses `go4.org/unsafe/assume-no-moving-gc@v0.0.0-20230204201903-c31fa085b70e` instead of `go4.org/unsafe/assume-no-moving-gc@v0.0.0-20230525183740-e7c30c78aeb2` when performing the check, which is an older version that didn't have the check for Go 1.21 yet.

fasmat: Well, I've been incorrect. The problem isn't actually any of our dependencies but rather `https://github.com/hairyhenderson/gomplate`. I'll try to see if the latest pre-release fixes the issue.

fasmat:
bors try

fasmat:
bors cancel

fasmat:
bors try

fasmat:
bors cancel

fasmat:
bors try

fasmat:
bors try

systest/tests/common.go
fasmat: Using version `v4.0.0-pre1` solves the issue 🙂

poszu: Gomplate is used to render a yaml file for a systes job. Perhaps it would be best to switch to Jinja or something else not using the unsafe voodoo assume-no-moving-gc module?

fasmat: We don't use `go4.org/unsafe/assume-no-moving-gc` in go-spacemesh, that was an incorrect assumption of mine, but `gomplate` does. The error wasn't from go-spacemesh, but from `gomplate`. Using a newer version solves the issue 🙂

tortoise/tortoise.go
```diff
@@ -415,33 +418,33 @@ func (t *turtle) countBallot(ballot *ballotInfo) error {
...
+		verified           = max(t.evicted, types.GetEffectiveGenesis())
 		nverified, changed types.LayerID
 	)
 	if t.isFull {
 		nverified, changed = t.runFull()
 		vverified, vchanged := t.runVerifying()
 		if nverified == t.processed-1 && nverified == vverified {
 			t.switchModes()
-			changed = types.MinLayer(changed, vchanged)
+			changed = min(changed, vchanged)
```
poszu, line 429: Previously if `changed` was 0, it would pick `vchanged`. Now it will pick `changed`. Is it intended?

fasmat: The problem with the previous implementation was that 0 had two meanings: 0 (as a minimum) and no minimum.

`t.runFull()` and `t.runVerifying` would return 0 for `changed` and `vchanged` if no changes were detected. Now it instead returns `math.MaxUint32`.

This removes the need for handling 0 in a special way (within `types.MinLayer`) and instead the detection for "no changed layer" is now asserting that `changed` is not `math.MaxUint32`.

fasmat: On top of that the documentation for `LayerID` fits the suggested implementation rather than the current one:

```
// LayerID is representing a layer number. Zero value is safe to use, and means 0.
// Internally it is a simple wrapper over uint32 and should be considered immutable
// the same way as any integer.
```

But the code assumes that 0 means "no layer" rather than the first layer. Also `types.MinLayer` is not transitive:

```go
types.MinLayer(a, b) <= types.MinLayer(b, c) // this can be false if b == 0
min(a, b) <= min(b, c) // this is always true
```

poszu: I know, I'm just worried that it could break some code that depends on this behaviour

fasmat:
bors try

tortoise/tortoise.go
poszu: > t.runFull() and t.runVerifying would return 0 for changed and vchanged if no changes were detected. Now it instead returns math.MaxUint32.

How about changing these functions to return a pointer, where `nil` means "not found" instead? It has the benefit of making it clear that the function returns an optional value. The downside is that it would most likely complicate the code using them :(

fasmat: I think the implementation with `math.MaxUint32` is already simpler than the current implementation?

> I know, I'm just worried that it could break some code that depends on this behaviour

The methods `runFull` and `runVerifying` only have package visibility and are only called in `verifyLayers`. Their returned values are only used in that method. For the only caller of `verifyLayers` (`onLayer`) nothing changed.

All 4 methods are well tested (100% coverage) - the other methods in that file also have near 100% coverage; unit and system tests pass. Earlier I accidentally had `max` where `min` was supposed to go and multiple tests failed.

I appreciate your concern, but also think the change is solid 🙂

poszu: fair enough

poszu (approved):

fasmat:
bors merge

PR #4866, opened 2023-08-17
[Merged by Bors] - Make grpc servers' start methods be synchronous

### LINKED ISSUE #4861: GRPC servers unnecessarily do startup asynchronously
## Description
`grpcserver.Server.Start` and `grpcserver.JSONHTTPServer.Start` return a channel upon which the caller can wait to be notified of the service having completed startup. It doesn't appear necessary and can easily be overlooked by callers.

Instead these methods should be synchronous, so that when they return the caller can safely connect to the server.
## Affected code
`api/grpcserver` package.
This issue appears in commit hash: c7a7468f3bc2ed2a159c2c81e90297deb5757efc

piersy:
## Motivation
Closes #4861 
## Changes

Updates grpc server startup code to execute synchronously, returning only after the server has started listening.

## Test Plan
Existing tests should pass

## DevOps Notes


api/grpcserver/grpcserver_test.go
```diff
@@ -491,20 +491,14 @@ func launchServer(tb testing.TB, cfg Config, services ...ServiceAPI) func() {
...
 	// start gRPC and json servers
-	grpcStarted := grpcService.Start()
-	jsonStarted := jsonService.StartService(context.Background(), services...)
+	err := grpcService.Start()
+	require.NoError(tb, err)
+	err = jsonService.StartService(context.Background(), services...)
+	require.NoError(tb, err)
 
 	timer := time.NewTimer(3 * time.Second)
 	defer timer.Stop()
```
fasmat: I think this is not needed any more 🙂

cmd/bootstrapper/generator_test.go
```diff
@@ -139,20 +139,14 @@ func launchServer(tb testing.TB, cdb *datastore.CachedDB) func() {
...
 	// start gRPC and json servers
-	grpcStarted := grpcService.Start()
-	jsonStarted := jsonService.StartService(context.Background(), s)
+	err := grpcService.Start()
+	require.NoError(tb, err)
+	err = jsonService.StartService(context.Background(), s)
+	require.NoError(tb, err)
 
 	timer := time.NewTimer(3 * time.Second)
 	defer timer.Stop()
```
fasmat: Same here, I don't think this is needed any more.

fasmat (approved):

api/grpcserver/grpc.go
```diff
@@ -44,35 +46,27 @@ func (s *Server) Start() <-chan struct{} {
...
+		}
+		return nil
+	})
+	return nil
 }
 
 // Close stops the server.
 func (s *Server) Close() error {
 	s.logger.Info("stopping the grpc server")
 	s.GrpcServer.Stop()
```
poszu, line 68: Consider a graceful stop to allow pending GRPC calls to finish:
```suggestion
	s.GrpcServer.GracefulStop()
```

piersy: I don't think just calling GracefulStop would work well with our streaming APIs, I think if you had a stream open the server would never shutdown. The approach to handling this, according to this [ticket](https://github.com/grpc/grpc-go/issues/2888), is to first call GracefulStop, wait some time and then call Stop. I can add a ticket for it, it seems to be out of scope for this PR.

piersy:
bors merge

api/grpcserver/grpc.go
poszu: Thanks, didn't know that! Do you think it's worth creating an issue to implement the proposed graceful shutdown later?

piersy: Sure will do!

piersy: https://github.com/spacemeshos/go-spacemesh/issues/4867

piersy:
bors cancel

piersy:
bors merge

piersy:
bors merge

PR #4869, opened 2023-08-17
[Merged by Bors] - Ensure that App closes started channel on startup error

### LINKED ISSUE #4868: node.App startup has tricky API
## Description
The current API for starting an App involves the method Start, which blocks till the App shuts down and another method Started which returns a channel, whereby users may wait for the App to be started.

So to start an app and wait for it to be started you could use the following code.
```go
g.Go(func() error {
	return app1.Start(ctx)
})
<-app1.Started()
```

The problem with this approach is that Start performs a lot of operations that might return an error and will result in Start returning that error without calling close on the started channel. This would result in the following code hanging at the point of the receive. It's not feasible to try and listen to the errors returned by Start since if there is no error on startup, you will end up waiting until the app is shut down.

## Affected code
`node/node.go`

This issue appears in commit hash: 6518d2b836c86729236c7a906a01867cc5234aa5

piersy:
## Motivation
Closes #4868 

## Changes
This simply moves non blocking app startup operations into a new method allowing for a deferred call to close the started channel. This ensures that started will always be closed before Start returns and thus prevents getting stuck waiting for the `App` to start. There were also a couple of services that were deferring their shutdown to the end of Start, they have now been added as fields of` node.App` and are shutdown in `stopServices` which keeps the shutdown approach consistent.
## Test Plan
Current tests should pass

## DevOps Notes


node/node.go
```diff
@@ -1285,6 +1285,36 @@ func (app *App) setupDBs(ctx context.Context, lg log.Log, dbPath string) error {
...
+		return nil
+	case err = <-app.errCh:
+		return err
+	}
+}
+
+func (app *App) startSynchronous(ctx context.Context) error {
+	// notify anyone who might be listening that the app has finished starting.
+	// this can be used by, e.g., app tests.
+	defer close(app.started)
```
poszu, line 1350: Now it will close the channel even in the case of an error during a start. Should it be that way?

piersy: Hey @poszu that is the intention, consider this code:

```go
g.Go(func() error {
	return app1.Start(ctx)
})
<-app1.Started()
```

If there's an error at startup and the started channel is not closed, the caller gets stuck, it doesn't matter that the error is returned through the error group, because the code is not able to progress to the point where it waits on the error group.

poszu: OK, now I understand what you want to do in this PR, actually. But TBH, I don't understand the purpose of the `started` channel. It now synchronizes to the point in time where `startSynchronous()` quits, but how is it useful?

fasmat: Do we even need the channel any more? It is closed when `Start` returns. So if `Start` returns `nil` the start was successful, otherwise it wasn't.

piersy: Hey @poszu & @fasmat the thing is that `Start` runs the app,  so if all goes well `Start` won't return until the app is shutting down. I believe the purpose of the started channel is to let the caller know when it is safe to call the APIs exposed by the app. An alternative would be to separate all listening from serving operations and then app startup could be done in two stages, first setting up all services to listen and a second call to start serving those services. However I think that would be a much larger change, since it would involve separating those stages for many components.

fasmat: Yeah, I momentarily forgot that `app.Start` is a blocking call. Forget my comment about the channel not being needed any more 🙂

fasmat (approved):

node/node.go
poszu: @piersy
> I believe the purpose of the started channel is to let the caller know when it is safe to call the APIs exposed by the app.

But that's not really true. The services are started asynchronously (in goroutines). For example, the GRPC services are started in a goroutine. They might not yet be started when the channel is closed.

piersy: @fasmat Hehe, easy to do, I have done the same :D

@poszu You're right, its a bit of a kludge, you don't actually know for sure that the services are going to be listening after receiving on the channel. However if the channel was removed all you could do is wait and try the API. So I think the channel use is preferable to that, or maybe I've misunderstood the original purpose of this channel, but I can't think what else it's for.

piersy: @poszu I added a ticket for the 2 stage startup. - https://github.com/spacemeshos/go-spacemesh/issues/4892

piersy:
bors merge

piersy:
bors merge

piersy:
bors try

.github/workflows/ci.yml
```diff
@@ -194,6 +194,8 @@ jobs:
         env:
           GOTESTSUM_FORMAT: standard-verbose
           GOTESTSUM_JUNITFILE: unit-tests.xml
+          # We set this to prevent segmentation faults on macOS see - https://github.com/golang/go/issues/49138
+          MallocNanoZone: 0
```
fasmat: This issue has been fixed in Go 1.17 if I understand the linked issue correctly, we are already using Go 1.20 🙂

node/node.go
```diff
@@ -167,6 +167,9 @@ func GetCommand() *cobra.Command {
 
 				// This blocks until the context is finished or until an error is produced
 				err = app.Start(ctx)
+				if err != nil {
+					app.log.With().Warning("App ended with error", log.Err(err))
```
fasmat: This error is already logged on line 194

node/node.go
```diff
@@ -1413,46 +1448,36 @@ func (app *App) Start(ctx context.Context) error {
...
-		})
-	}
-	// app blocks until it receives a signal to exit
-	// this signal may come from the node or from sig-abort (ctrl-c)
-	select {
-	case <-ctx.Done():
-		return nil
-	case err = <-app.errCh:
+	// cleanup shuts down components started here.
+	cleanup := func() error {
```
fasmat: I feel like this should be part of `app.stopServices` instead of being returned via an anonymous function.

.github/workflows/ci.yml
piersy: I can't find the run now, but I was seeing SIGSEV errors, I've removed it, lets see what happens.

node/node.go
piersy: Lets see how that looks

piersy: Yep, that looks a lot nicer thanks :D

.github/workflows/ci.yml
```diff
@@ -190,7 +190,7 @@ jobs:
       - name: Clear test cache
         run: make clear-test-cache
       - name: unit tests
-        timeout-minutes: 25
+        timeout-minutes: 35
```
fasmat, line 193: This increase shouldn't be necessary, unit tests take ~ 10 minutes on our self hosted runners and < 20 minutes on GH runners (except for Windows, but that should have a longer timeout already)

node/node.go
```diff
@@ -1221,6 +1222,17 @@ func (app *App) stopServices(ctx context.Context) {
 		app.postVerifier.Close()
 	}
 
+	if app.pprofService != nil {
+		if err := app.db.Close(); err != nil {
```
fasmat:
```suggestion
		if err := app.pprofService.Close(); err != nil {
```

fasmat (commented):
lgtm, would like to hear @poszu opinion on the change too.

.github/workflows/ci.yml
piersy: It did take 28 minutes, in this run - https://github.com/spacemeshos/go-spacemesh/actions/runs/5966242390/job/16187625547. That's why I changed it.

I see it also taking 20 minutes in this run - https://github.com/spacemeshos/go-spacemesh/actions/runs/5967856416/job/16190433807.

piersy: Hey @fasmat, in the most recent [run](https://github.com/spacemeshos/go-spacemesh/actions/runs/5968103737/job/16191204690) its back to 17 minutes, but it seems that it can fluctuate quite a lot. What do you want to do?

fasmat: OK then leave the 35 minutes for now. I just want to avoid our CI becoming slower over time 🙂

poszu (approved):

piersy:
This PR is now also suffering the same problem as #4897 see  https://github.com/spacemeshos/go-spacemesh/pull/4897#issuecomment-1693520633. My guess is that since these PRs both change the timing of shutdown events they are somehow triggering a previously unseen condition.

piersy:
bors merge

piersy:
bors merge

PR #4871, opened 2023-08-18
[Merged by Bors] - Add jitter to spread out requests to get poet proof and submit challenge

### LINKED ISSUE #4860: Add small jitter to requests to poet
## Description
Currently, all nodes hit the poet services at roughly the same moment. It creates unnecessary spike in traffic. Especially, `GET /Proof` could be distributed more evenly as the initial requests get a 404 until the proof is generated and cached in CF.

## Affected code
`nipostBuilder`, poet client

poszu:
## Motivation
Closes #4860 

## Changes
Add small positive jitter when:
- waiting for a poet round to end (to get proof),
- waiting to build a nipost challenge (to spread out challenge registrations).
Jitter is only added if the round end is in the future to avoid unnecessary wait.

The range of jitter duration is calculated as a percentage of a cycle gap duration so it also works in unit tests and system tests. 

### Jitter before getting the proof
The min and max % (0.02% and 0.04%) were chosen so that on mainnet:
- minimum jitter is roughly equivalent to the time it takes to generate a proof on bare metal poets (it happens after the round ends): 8.64s. This helps avoid situations when many nodes got a 404 response.
- maximum jitter is 2xmin - roughly 17s

### Jitter before submitting challenges
on mainnet:
- min = 0s
- max = 36s (1% of grace period)

## Test Plan
Added unit tests checking calculating wait and jitter times.

### DIFF
--- a/activation/nipost.go
+++ b/activation/nipost.go
@@ -404,9 +404,9 @@ func (nb *NIPostBuilder) getBestProof(ctx context.Context, challenge types.Hash3
 			continue
 		}
 		round := r.PoetRound.ID
-		waitTime, jitter := calcGetProofWaitTime(r.PoetRound.End.IntoTime(), nb.poetCfg.CycleGap)
+		waitTime := calcGetProofWaitTime(time.Until(r.PoetRound.End.IntoTime()), nb.poetCfg.CycleGap)
 		eg.Go(func() error {
-			logger.With().Info("waiting till poet round end", log.Duration("wait time", waitTime), log.Duration("jitter", jitter))
+			logger.With().Info("waiting till poet round end", log.Duration("wait time", waitTime))
 			select {
 			case <-ctx.Done():
 				return fmt.Errorf("waiting to query proof: %w", ctx.Err())
@@ -497,19 +497,14 @@ func constructMerkleProof(challenge types.Hash32, members []types.Member) (*type
 }
 
 func randomDurationInRange(min, max time.Duration) time.Duration {
-	return min + time.Duration(float64((max-min).Nanoseconds())*rand.Float64())
+	return min + time.Duration(rand.Int63n(int64(max-min+1)))
 }
 
-// Time to wait before querying for the proof
+// Calculate the time to wait before querying for the proof
 // We add a jitter to avoid all nodes querying for the proof at the same time.
-// There's no need to add jitter if the round has already ended.
-func calcGetProofWaitTime(roundEnd time.Time, cycleGap time.Duration) (waitTime, jitter time.Duration) {
-	waitTime = time.Until(roundEnd)
-	if waitTime > 0 {
-		minWaitTime := time.Duration(float64(cycleGap) * MinPoetGetProofJitter / 100.0)
-		maxWaitTime := time.Duration(float64(cycleGap) * MaxPoetGetProofJitter / 100.0)
-		jitter := randomDurationInRange(minWaitTime, maxWaitTime)
-		return waitTime, jitter
-	}
-	return 0, 0
+func calcGetProofWaitTime(tillRoundEnd, cycleGap time.Duration) (waitTime time.Duration) {
+	minWaitTime := time.Duration(float64(cycleGap) * MinPoetGetProofJitter / 100.0)
+	maxWaitTime := time.Duration(float64(cycleGap) * MaxPoetGetProofJitter / 100.0)
+	jitter := randomDurationInRange(minWaitTime, maxWaitTime)
+	return tillRoundEnd + jitter
 }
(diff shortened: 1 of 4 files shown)


activation/nipost.go
```diff
@@ -479,3 +495,21 @@ func constructMerkleProof(challenge types.Hash32, members []types.Member) (*type
 		Nodes:     nodesH32,
 	}, nil
 }
+
+func randomDurationInRange(min, max time.Duration) time.Duration {
+	return min + time.Duration(float64((max-min).Nanoseconds())*rand.Float64())
```
fasmat:
```suggestion
	return min + time.Duration(rand.Int63n(int64(max-min)))
```

should be equivalent but avoids conversions to and from `float64` (`time.Duration` is already an `int64` so the cast should be free).

activation/nipost.go
```diff
@@ -479,3 +495,21 @@ func constructMerkleProof(challenge types.Hash32, members []types.Member) (*type
...
+}
+
+// Time to wait before querying for the proof
+// We add a jitter to avoid all nodes querying for the proof at the same time.
+// There's no need to add jitter if the round has already ended.
+func calcGetProofWaitTime(roundEnd time.Time, cycleGap time.Duration) (waitTime, jitter time.Duration) {
+	waitTime = time.Until(roundEnd)
+	if waitTime > 0 {
```
fasmat: I think this should check for `waitTime > -jitter` so that if the node is a bit late it still jitters correctly.

activation/nipost.go
```diff
@@ -479,3 +495,21 @@ func constructMerkleProof(challenge types.Hash32, members []types.Member) (*type
...
+		minWaitTime := time.Duration(float64(cycleGap) * MinPoetGetProofJitter / 100.0)
+		maxWaitTime := time.Duration(float64(cycleGap) * MaxPoetGetProofJitter / 100.0)
+		jitter := randomDurationInRange(minWaitTime, maxWaitTime)
+		return waitTime, jitter
```
fasmat: Why not just return `waitTime + jitter`? If its because of logging purposes, we can just log the jittertime here.

activation/nipost.go
```diff
@@ -385,12 +404,9 @@ func (nb *NIPostBuilder) getBestProof(ctx context.Context, challenge types.Hash3
...
 		}
 		round := r.PoetRound.ID
-		// Time to wait before querying for the proof
-		// The additional second is an optimization to be nicer to poet
-		// and don't accidentally ask it to soon and have to retry.
-		waitTime := time.Until(r.PoetRound.End.IntoTime()) + time.Second
+		waitTime, jitter := calcGetProofWaitTime(r.PoetRound.End.IntoTime(), nb.poetCfg.CycleGap)
 		eg.Go(func() error {
-			logger.With().Info("waiting till poet round end", log.Duration("wait time", waitTime))
+			logger.With().Info("waiting till poet round end", log.Duration("wait time", waitTime), log.Duration("jitter", jitter))
```
fasmat: `jitter` isn't used, except to log it here.

activation/nipost.go
poszu: Mostly for easier testability but also to log it later.

activation/nipost.go
poszu: yes, good catch!

fasmat (approved):

activation/nipost.go
```diff
@@ -479,3 +495,16 @@ func constructMerkleProof(challenge types.Hash32, members []types.Member) (*type
...
+
+func randomDurationInRange(min, max time.Duration) time.Duration {
+	return min + time.Duration(rand.Int63n(int64(max-min+1)))
+}
+
+// Calculate the time to wait before querying for the proof
+// We add a jitter to avoid all nodes querying for the proof at the same time.
+func calcGetProofWaitTime(tillRoundEnd, cycleGap time.Duration) (waitTime time.Duration) {
+	minWaitTime := time.Duration(float64(cycleGap) * MinPoetGetProofJitter / 100.0)
+	maxWaitTime := time.Duration(float64(cycleGap) * MaxPoetGetProofJitter / 100.0)
```
dshulyak: so thats 0.04% and not 4%? i got confused

dshulyak (approved):

activation/nipost.go
poszu: Yes, that's 0.02% and 0.04%. I can change it if it's confusing.

dshulyak (approved):

poszu:
bors merge

poszu:
bors cancel

poszu:
bors merge

PR #4876, opened 2023-08-19
[Merged by Bors] - miner oracle use ref ballot for active set

### LINKED ISSUE #4837: miner/oracle: use activeset from reference ballot
if node created reference ballot with graded activeset and later was restarted it will create an activeset from first block.
such ballots will be syntactically invalid as receiver will validate them according to the reference ballot, that used graded active set.

this needs to be fixed before next epoch begins, otherwise network will very likely die 

the solution is to lookup activeset on a reference ballot  that stored locally, there is already code for it.

### LINKED ISSUE #4838: miner/oracle, sync: don't rely on the time when node was restarted for activeset selection
if all nodes were restarted before first block was created network will die. such heuristics in consensus are very risky, if something doesn't go according to the plan we will just have more issues to deal with.

to determine whether node was synced on time we can check how many atxs are with a good grade, if there is a sufficient number of them (e.g 80% fraction) then node was actually synced on time, on not just restarted

this is also highly needed before next epoch starts

countvonzero:
## Motivation
Closes #4837
Closes #4838

## Changes
- use active set from ref ballot when available
- new criteria to decide if a node should use epoch's first block to derive active set:
  90% of its epoch active set do not have acceptable grade

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -70,7 +70,8 @@ func (cfg *Config) DataDir() string {
 }
 
 type TestConfig struct {
-	SmesherKey string `mapstructure:"testing-smesher-key"`
+	SmesherKey      string `mapstructure:"testing-smesher-key"`
+	MinerGoodAtxPct int
 }
 
 // BaseConfig defines the default configuration options for spacemesh app.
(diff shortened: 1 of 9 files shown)


countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #4877, opened 2023-08-19
add metrics for optimistic filtering and block execution

countvonzero:
## Motivation
as title

## Changes
add metrics for
- optimistic filtering is enabled
- percentage agreement on state from proposals
- block executed optimistically
- block executed non-optimistically

### DIFF
--- a/blocks/utils.go
+++ b/blocks/utils.go
@@ -97,18 +97,27 @@ func getProposalMetadata(
 	}
 	majority := cfg.OptFilterThreshold * len(proposals)
 	var majorityState *meshState
+	var highest int
 	for _, ms := range meshHashes {
 		logger.With().Debug("mesh hash",
 			ms.hash,
 			log.Int("count", ms.count),
 			log.Int("threshold", cfg.OptFilterThreshold),
 			log.Int("num_proposals", len(proposals)))
-		if ms.hash != types.EmptyLayerHash && ms.count*100 >= majority {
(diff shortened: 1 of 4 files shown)


blocks/utils.go
```diff
@@ -97,18 +97,27 @@ func getProposalMetadata(
...
-			majorityState = ms
+		if ms.hash != types.EmptyLayerHash {
+			if ms.count > highest {
+				highest = ms.count
+			}
+			if ms.count*100 >= majority {
+				majorityState = ms
+				optimistic.Inc()
+				agreement.Set(float64(ms.count*100) / float64(len(proposals)))
+			}
```
dshulyak, line 115: i don't think that updating metrics in util function is an ok approach.

also storing last value is somewhat useless

countvonzero: maybe i misunderstand the metrics types.
my main goal was to see the degree of consensus in each layer.
layer 19 -> 95%
layer 20 -> 100%
layer 21 -> 92%
and see the up and down trend

dshulyak: i don't know if this will work like that, it depends on scraping timings and prometheus queuing. but i don't mind if you want to check.

i would store agreement fraction in histogram with 10 buckets (0-10, 10-20, and so on). the granularity can be even higher (e.g step can be 5%), but probably not needed

dshulyak (approved):

PR #4879, opened 2023-08-20
[Merged by Bors] - update majority calculation for optimistic filtering

### LINKED ISSUE #4878: optimistic filtering majority should be weighted
## Description
currently each proposal is weighted the same (1) for majority state calculation.
it should be weighted proportionally to the proposal's eligibility count

countvonzero:
## Motivation
Closes #4878

## Changes
this change will take effect at the first layer of epoch 3


blocks/utils.go
```diff
@@ -69,15 +69,22 @@ func getProposalMetadata(
 	if err != nil {
 		return nil, err
 	}
+	upgrade := proposals[0].Layer.Uint32() >= types.OpUpgradeLayer()
```
dshulyak: it is not obvious that proposals[0] exists

dshulyak (approved):

countvonzero:
bors merge

PR #4882, opened 2023-08-21
[Merged by Bors] - datastore: parametrize and increase cache size

dshulyak:
tried to sync node and noticed that it spends a lot of time looping over activeset in eligibility validation. i will open an issue to optimize that code path, but for now we need to increase number of activations stored in cache.

related: https://github.com/spacemeshos/go-spacemesh/issues/4883

### DIFF
--- a/datastore/store.go
+++ b/datastore/store.go
@@ -43,14 +43,44 @@ type CachedDB struct {
 	malfeasanceCache *lru.Cache[types.NodeID, *types.MalfeasanceProof]
 }
 
+type Config struct {
+	ATXSize         int `mapstructure:"atx-size"`
+	MalfeasenceSize int `mapstructure:"malfeasence-size"`
+}
+
+func DefaultConfig() Config {
+	return Config{
+		ATXSize:         10_000,
+		MalfeasenceSize: 1_000,
+	}
+}
+
+type cacheOpts struct {
+	cfg Config
+}
+
+type Opt func(*cacheOpts)
+
+func WithConfig(cfg Config) Opt {
+	return func(o *cacheOpts) {
+		o.cfg = cfg
+	}
+}
+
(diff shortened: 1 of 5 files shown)


datastore/store.go
```diff
@@ -43,14 +43,44 @@ type CachedDB struct {
...
+		opt(&o)
+	}
+	lg.With().Info("initialized datastore", log.Any("config", o.cfg))
+
+	atxHdrCache, err := lru.New[types.ATXID, *types.ActivationTxHeader](o.cfg.ATXSize)
 	if err != nil {
 		lg.Fatal("failed to create atx cache", err)
 	}
 
-	malfeasanceCache, err := lru.New[types.NodeID, *types.MalfeasanceProof](malfeasanceCacheSize)
```
piersy, line 53: I guess the `malfeasanceCacheSize` constant can be removed.

datastore/store.go
```diff
@@ -43,14 +43,44 @@ type CachedDB struct {
...
+
+func WithConfig(cfg Config) Opt {
+	return func(o *cacheOpts) {
+		o.cfg = cfg
+	}
+}
+
 // NewCachedDB create an instance of a CachedDB.
-func NewCachedDB(db *sql.Database, lg log.Log) *CachedDB {
-	atxHdrCache, err := lru.New[types.ATXID, *types.ActivationTxHeader](atxHdrCacheSize)
```
piersy, line 48: It seems that `atxHdrCacheSize` is still used but only for the `vrfNonceCache`, so it could be renamed.

piersy (approved):
Two small points, otherwise looks good!

dshulyak:
thanks, i removed both constants, and using ATXSize for vrf nonce too

dshulyak:
bors merge

PR #4893, opened 2023-08-22
[Merged by Bors] - Give NiPoSTBuilder more time to create the PoST

### LINKED ISSUE #4797: Make sure that AtxBuilder continues to build an ATX even if missed next poet round
## Description
Currently, `AtxBuilder` sets a deadline for when a NiPOST should be available at the start of the next poet round. With this approach, it might accidentally drop a valid NiPOST and not produce an ATX if POST proof generation took too long (overlapped with the next poet round start).

## Expected Behavior
It should publish an ATX even if the next poet's round was missed because this ATX is **valid**.

TODO: What's a reasonable timeout? 

## Additional Resources 
Responsible code: https://github.com/spacemeshos/go-spacemesh/blob/94532841da227903ffb0ab5aef15da9990bb7f88/activation/activation.go#L617-L622

fasmat:
## Motivation
Closes #4797

With this change ATXs can now be published until the end of their `PublishEpoch` (instead of until the start of the next PoET round). On mainnet this means a node as 4 additional days of time to generate a PoST proof and publish an ATX.

It is still required to publishing within the window between the end of one PoET round and the start of the next one to not miss any rewards. This change only allows nodes that weren't able to register for the next PoET round to still publish an ATX within the 4 days (on mainnet) between the end of the `CycleGap` and the end of the `PublishEpoch`.

## Changes
This PR adapts deadlines to the following:

- Phase 0 (submitting challenge to PoET): deadline `poetRoundStart` - unchanged
- Phase 1 (fetching proof from PoET): before the deadline was the end of the round it submitted to + a grace period of 1 hour; this effectively means there is only a 1 hour window to fetch the proof from PoET before the node gives up. Now instead it will fetch until the end of the `PublishEpoch`. This means that if the node doesn't manage to publish its ATX before next PoET round starts, it won't be able to publish one next epoch (unchanged) but will still try to publish one for the current epoch.
- Phase 2 (publishing the ATX): before deadline was `nextPoetRoundStart`, with the intention to publish the ATX and have enough time left to register for the next PoET round. This is a very short window (just the time between the end of one PoET round and the start of the next one). If the node doesn't manage to generate a PoST proof within that time it gives up on publishing an ATX and misses out on rewards it could get by instead giving up on registering on the next PoET round. This deadline was also extended to the end of the `PublishEpoch`.

## Test Plan
existing tests pass or have been adapted

## TODO

## DevOps Notes


fasmat:
bors try

activation/nipost.go
```diff
@@ -264,11 +261,18 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 	// Phase 2: Post execution.
 	var postGenDuration time.Duration = 0
 	if nb.state.NIPost.Post == nil {
+		publishDeadline := nb.layerClock.LayerToTime(challenge.TargetEpoch().FirstLayer())
+		if poetProofDeadline.Before(now) {
+			return nil, 0, fmt.Errorf("%w: deadline to publish ATX for pub epoch %d exceeded (deadline: %s, now: %s)", ErrATXChallengeExpired, challenge.PublishEpoch, publishDeadline, now)
+		}
+		publishCtx, cancel := context.WithDeadline(ctx, publishDeadline)
+		defer cancel()
```
poszu: IMHO, this should be in the upper layer, probably in `AtxBuilder::PublishActivationTx()`, wdyt?

fasmat: I think both is correct. I created a timeout in `PublishActivationTx` that sets a deadline for `createAtx` to finish before the end of `TargetEpoch` (any later makes no sense to publish, but even slightly before the end could still yield some rewards).

I also left a deadline here. From `BuildNiPoST`s perspective it wants to be finished before `nextPoetRoundEnd` so it doesn't miss more than one PoET round.

Which of the two deadlines becomes effective depends on the configuration then, the earlier of the two is the one defining when the `context` gets cancelled. In the setup we have for mainnet that would be `nextPoetRoundEnd` (4 days before the end of `TargetEpoch`), which seems sensible to me.

wdyt?

fasmat:
bors try

dshulyak:
i think it was me who set that deadline.. the idea was that you have to give up if you want to submit to the poet round. otherwise you don't guarantee success in current (because why it takes so long) and definitely will do nothing in the following, since you missed poet window. don't know if what i described is better, just sharing what was the intuition

activation/nipost.go
```diff
@@ -264,11 +265,20 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 	// Phase 2: Post execution.
 	var postGenDuration time.Duration = 0
 	if nb.state.NIPost.Post == nil {
+		now := time.Now()
+		// Deadline: end of next PoET round. We can still publish an ATX late (i.e. within target epoch) and receive rewards,
+		// but we don't want to miss more than one PoET round.
+		if nextPoetRoundEnd.Before(now) {
+			return nil, 0, fmt.Errorf("%w: deadline to publish ATX for pub epoch %d exceeded (deadline: %s, now: %s)", ErrATXChallengeExpired, challenge.PublishEpoch, nextPoetRoundEnd, now)
+		}
+		publishCtx, cancel := context.WithDeadline(ctx, nextPoetRoundEnd)
```
poszu: It's not really publishing anything here. How about `postGenCtx`?

poszu (approved):

poszu (approved):

fasmat:
bors merge

dshulyak:
>         	Error Trace:	/build/systest/tests/smeshing_test.go:163
        	            				/build/systest/tests/poets_test.go:117
        	            				/build/systest/tests/poets_test.go:35
        	Error:      	Not equal: 
        	            	expected: 12
        	            	actual  : 8


are you sure that this failure is not because of the change? if atx is not submitted properly eligibilities will be different. also to my knowledge tests are not flaky

dshulyak:
bors cancel

fasmat:
bors try

activation/nipost.go
```diff
@@ -177,35 +177,32 @@ func (nb *NIPostBuilder) UpdatePoETProvers(poetProvers []PoetProvingServiceClien
...
+	//  WE ARE HERE            PROOF BECOMES         ATX PUBLICATION
+	//                           AVAILABLE               DEADLINE
+
+	publishEpoch := challenge.PublishEpoch
+	poetRoundStart := nb.layerClock.LayerToTime((publishEpoch - 1).FirstLayer()).Add(nb.poetCfg.PhaseShift)
+	poetRoundEnd := nb.layerClock.LayerToTime(publishEpoch.FirstLayer()).Add(nb.poetCfg.PhaseShift).Add(-nb.poetCfg.CycleGap)
+
+	// we want to publish before the publish epoch ends or we won't receive rewards (due to ATX grading)
+	// GracePeriod is a safety margin to make sure we don't miss the deadline
+	publishEpochEnd := nb.layerClock.LayerToTime((publishEpoch + 1).FirstLayer()).Add(-nb.poetCfg.GracePeriod)
```
poszu: Why use a poet-specific config (`GracePeriod`) param for a deadline that is not related to poets? Perhaps it should not be in poet config at all (or there should be a separate grace period for ATX publication)

poszu (approved):

activation/nipost.go
fasmat: Yeah, I just thought I use something that we already have in the config and isn't used any more (`GracePeriod` as it was used before has been dropped).

Ideally we use 4 * `app.Config.HARE.WakeupDelta` here, since this is what is also used in ATX Grading to decide if the ATX is still good or not.

fasmat: I decided to drop this additional buffer. On mainnet it is 1min 40 seconds before the end of a 2 week epoch or 4.5 days of time between fetching the poet proof and the end of the publish epoch.

This is the most a node can waste in time to calculate an ATX that it won't receive rewards for.

fasmat:
bors merge

PR #4902, opened 2023-08-24
[Merged by Bors] - Fix handling of PoET proof fetching timeout

### LINKED ISSUE #4901: Flaky TestPoetsFailures
>         	Error Trace:	/build/systest/tests/smeshing_test.go:163
        	            				/build/systest/tests/poets_test.go:117
        	            				/build/systest/tests/poets_test.go:35
        	Error:      	Not equal: 
        	            	expected: 12
        	            	actual  : 8
        	Test:       	TestPoetsFailures
        	Messages:   	0x0e99c2958ed7365d0543f0146df5b1f172f51b2170ea59c58404afcdd5e20737
    logger.go:130: 2023-08-24T09:56:06.527Z	INFO	TestPoetsFailures	namespace was deleted	***"namespace": "test-fomj"***

https://github.com/spacemeshos/go-spacemesh/actions/runs/5961815153/job/16171771670

https://github.com/spacemeshos/go-spacemesh/actions/runs/5949146807/job/16134376481

fasmat:
## Motivation
Closes #4901

## Changes
Deadlines and Timeouts in the NiPoST builder have been updated for Phase 1 (fetching PoET Proofs):

- Deadline Before: end of publish epoch; this is too late. In system tests this causes a test to fail if a singe PoET is slow to respond
- Deadline Now: end of publish epoch - cycle gap. A node that is able to build a PoST proof within the cycle gap (which is necessary to not miss an epoch) has enough time left to publish an ATX if it receives it until then.
- Additionally a timeout for fetching PoET Proofs has been added. For this timeout I used the PoET config `grace-period`.
  - For nodes that are online at the start of the cycle gap this means the same deadline for fetching a PoET proof as before
  - For nodes that are offline when the PoET round ends this means they can still fetch a PoET proof later if the deadline has not yet been met
  - TODO: GracePeriod is not a good name for this config, it should rather be called "PoET Request Timeout" or similar. Also 1 hour is way to much, on mainnet this can easily be 1 minute or even less, in system tests it's 10 seconds at the moment.
  - Not using a timeout here can lead in some cases to the client trying to download a proof from a non-responding PoET until it is to late to publish an ATX, while a valid proof from another PoET might have been available already.
- I changed `calcGetProofWaitTime` to not calculate a duration to a deadline
  - There might be some time that passes between calculating the duration and using it (especially since this calculating and using the wait time happens on 2 different go routines). This can cause problems when the wait time and deadline are close together (e.g. tests with very short epoch durations). Using time instead of duration minimizes this problem.

## TODO

## DevOps Notes

### DIFF
--- a/activation/interface.go
+++ b/activation/interface.go
@@ -43,7 +43,7 @@ type layerClock interface {
 
 type nipostBuilder interface {
 	UpdatePoETProvers([]PoetProvingServiceClient)
-	BuildNIPost(ctx context.Context, challenge *types.NIPostChallenge) (*types.NIPost, time.Duration, error)
+	BuildNIPost(ctx context.Context, challenge *types.NIPostChallenge) (*types.NIPost, error)
 	DataDir() string
 }
 
--- a/activation/nipost.go
+++ b/activation/nipost.go
@@ -175,7 +175,7 @@ func (nb *NIPostBuilder) UpdatePoETProvers(poetProvers []PoetProvingServiceClien
 // BuildNIPost uses the given challenge to build a NIPost.
 // The process can take considerable time, because it includes waiting for the poet service to
 // publish a proof - a process that takes about an epoch.
-func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPostChallenge) (*types.NIPost, time.Duration, error) {
+func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPostChallenge) (*types.NIPost, error) {
 	logger := nb.log.WithContext(ctx)
 	// Note: to avoid missing next PoET round, we need to publish the ATX before the next PoET round starts.
 	//   We can still publish an ATX late (i.e. within publish epoch) and receive rewards, but we will miss one
@@ -214,7 +214,7 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 	nb.loadState(challengeHash)
 
 	if s := nb.postSetupProvider.Status(); s.State != PostSetupStateComplete {
-		return nil, 0, errors.New("post setup not complete")
+		return nil, errors.New("post setup not complete")
 	}
 
 	// Phase 0: Submit challenge to PoET services.
@@ -222,7 +222,7 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 		now := time.Now()
 		// Deadline: start of PoET round for publish epoch. PoET won't accept registrations after that.
 		if poetRoundStart.Before(now) {
-			return nil, 0, fmt.Errorf("%w: poet round has already started at %s (now: %s)", ErrATXChallengeExpired, poetRoundStart, now)
+			return nil, fmt.Errorf("%w: poet round has already started at %s (now: %s)", ErrATXChallengeExpired, poetRoundStart, now)
 		}
 
 		signature := nb.signer.Sign(signing.POET, challengeHash.Bytes())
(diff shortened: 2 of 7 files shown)


fasmat:
bors try

fasmat:
bors try

activation/nipost.go
```diff
@@ -237,35 +242,32 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
 	if nb.state.NIPost.Post == nil {
 		now := time.Now()
-		// Deadline: the end of the publish epoch (with a safety margin of `GracePeriod`). If we do not publish within
+		// Deadline: the end of the publish epoch. If we do not publish within
 		// the publish epoch we won't receive any rewards in the target epoch.
 		if publishEpochEnd.Before(now) {
-			return nil, 0, fmt.Errorf("%w: deadline to publish ATX for pub epoch %d exceeded (deadline: %s, now: %s)", ErrATXChallengeExpired, challenge.PublishEpoch, publishEpochEnd, now)
+			return nil, fmt.Errorf("%w: deadline to publish ATX for pub epoch %d exceeded (deadline: %s, now: %s)", ErrATXChallengeExpired, challenge.PublishEpoch, publishEpochEnd, now)
 		}
 		postCtx, cancel := context.WithDeadline(ctx, publishEpochEnd)
```
poszu, line 272: IMHO, this check shouldn't be necessary. Isn't it equivalent to checking if `ctx` is canceled already? `AtxBuilder` knows the deadline to publish an ATX and it should time out the passed ctx properly, what do you think?

activation/interface.go
```diff
@@ -43,7 +43,7 @@ type layerClock interface {
 
 type nipostBuilder interface {
 	UpdatePoETProvers([]PoetProvingServiceClient)
-	BuildNIPost(ctx context.Context, challenge *types.NIPostChallenge) (*types.NIPost, time.Duration, error)
+	BuildNIPost(ctx context.Context, challenge *types.NIPostChallenge) (*types.NIPost, error)
```
poszu, line 46: :+1:

activation/nipost_test.go
```diff
@@ -513,6 +534,10 @@ func TestNIPostBuilder_ManyPoETs_AllFinished(t *testing.T) {
 		},
 	}
 
+	poetCfg := PoetConfig{
+		GracePeriod: 1 * time.Second,
+	}
```
poszu: Why was it needed?

fasmat: In an earlier commit I was checking this value inside `BuildNiPoST` but now it's just used to build the context passed to the PoET client. I removed it from the tests where that client is mocked and it doesn't need to be set.

activation/nipost.go
fasmat: I had the following 3 things in mind when adding this section:

- this way every phase is structurally the same: check if we are in that phase, check if the deadline for that phase has been reached, try to complete the actual steps necessary to get into the next phase. If we drop this code the last step wouldn't check if the deadline for the phase has already been met (or just indirectly via the context)
- BuildNiPoST could be called in a different code path and passed a different context. The deadline however would still apply
- Only when the deadline has passed the function returns an error that wraps `ErrATXChallengeExpired` which we use as indicator to discard the current challenge, otherwise it returns an error that wraps `ctx.Err()`. This is the reason why before v1.1.2 we end up in a loop of trying to build a PoST, this check prevents this from happening.

poszu (approved):

activation/nipost.go
fasmat: As discussed offline, we decided to keep this deadline for now and possibly revisit it at a later time (e.g. when canceling `GenerateProof` becomes possible).

fasmat:
bors merge

fasmat:
bors merge

Issue #4841, opened 2023-08-14
ballots: remove GetLegacyNumEligible in the next epoch


dshulyak:
this was a fix to cover up a mistake in mainnet configuration.

to remove it:
- validation in handler for past epochs and tortoise in all epochs should use eligibility number specified in the ballot 
- ~validation in current epoch should compute expected number of ballots based on specified active set~
- ~miner should either compute based on specified active set or use active set in reference ballot https://github.com/spacemeshos/go-spacemesh/issues/4837~

PR #4923, opened 2023-08-29
[Merged by Bors] - optimize ballot eligibility validation

### LINKED ISSUE #4883: miner/oracle: optimize eligibility validation
the code below should be executed only for reference ballot. the sync is incredibly slow with that loop

image

also for current epoch we should get this data from tortoise rather then lru. lru is bad here, because if set to 10_000 and in fact there are 11_000 in active set then this cache will be threshed. tortoise on the other hand needs to store this data regardless, and it stores it in a compact form.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4883
related: https://github.com/spacemeshos/go-spacemesh/issues/4841

active set is validated only for reference ballot in current epoch, for all other ballots eligibilities validated against declared slots in the reference ballot. beside making validation more efficient, it will allow to prune active set data in future changes. reference ballots is still loading data from cached datastore, i plan to remove eventually but not in this change.

additionally i rewrote all tests for eligibility validator to get to 100% coverage and make them less dependent on the implementation.


dshulyak:
i synced node with this change couple of times. another related improvement https://github.com/spacemeshos/go-spacemesh/issues/4927

dshulyak:
bors try

dshulyak:
bors try

proposals/interface.go
```diff
@@ -25,14 +25,14 @@ type ballotDecoder interface {
...
-}
-
 type nonceFetcher interface {
 	VRFNonce(types.NodeID, types.EpochID) (types.VRFPostIndex, error)
 }
 
+type vrfVerifier interface {
+	Verify(types.NodeID, []byte, types.VrfSignature) bool
+}
+
```
countvonzero: seems a no-op

countvonzero (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors try

dshulyak:
bors merge

PR #4934, opened 2023-08-31
[Merged by Bors] - sync: require stated synced for node to be synced

### LINKED ISSUE #4931: syncer: insufficient criteria for a node to become synced
## Description
currently node will declare itself synced when it fetched all atxs/ballots from peers. 
it doesn't wait for the state to be synced with the network.
and when a node become synced, it starts to participate in hare and tortoise consensus.

we found that the condition is insufficient for the node to become synced for the following reason.
blocks were downloaded only when one of the following conditions is satisified.
A. the node received a valid block certificate
B. tortoise verified the block to be valid

when a node become synced, and the state lags, hare will finish but unable to generate blocks, because its state is different from the majority of proposals in the hare output. 
and when blocks are not locally available when hare finishes, tortoise will vote against the block in the next layer.
the ripple effect causes the node to lag even further and produce bad votes for the network.

### proposal

sync all data up to layer N and call `ProcessLayer`/`TallyVotes` up to layer N until no errors is returned from `ProcessLayer()`

countvonzero:
## Motivation
Closes #4931

## Changes
- make sure state is synced before declaring node is synced
- reduce gossip sync from 10 min to 2 min
- do not print confusing log "failed to build proposal" when node is not synced

## Test
- manually synced with mainnet

### DIFF
--- a/syncer/state_syncer.go
+++ b/syncer/state_syncer.go
@@ -35,7 +35,8 @@ func minLayer(a, b types.LayerID) types.LayerID {
 
 func (s *Syncer) stateSynced() bool {
 	current := s.ticker.CurrentLayer()
-	return current.Uint32() <= 1 || !s.mesh.ProcessedLayer().Before(current.Sub(1))
+	return current <= types.GetEffectiveGenesis() ||
+		(s.mesh.ProcessedLayer() >= current-1 && !s.stateErr.Load())
 }
 
 func (s *Syncer) processLayers(ctx context.Context) error {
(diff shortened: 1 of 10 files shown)


countvonzero:
bors try

syncer/state_syncer.go
```diff
@@ -35,7 +35,8 @@ func minLayer(a, b types.LayerID) types.LayerID {
 
 func (s *Syncer) stateSynced() bool {
 	current := s.ticker.CurrentLayer()
-	return current.Uint32() <= 1 || !s.mesh.ProcessedLayer().Before(current.Sub(1))
+	return current <= types.GetEffectiveGenesis() ||
+		(s.mesh.ProcessedLayer() >= current-1 && !s.stateErr.Load())
```
dshulyak, line 39: is there a reason why checkMeshAgreement is based on this function instead of `synced` state?

dshulyak (approved):

syncer/state_syncer.go
countvonzero: changed.

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors cancel

countvonzero:
bors merge

countvonzero:
bors merge

Issue #4927, opened 2023-08-30
fetcher: provide api to fetch objects without validating if they are available in db


dshulyak:
existing api always tries to lookup if object exists in the database, this is both inefficient and not robust.

as an example ballot validation

![image](https://github.com/spacemeshos/go-spacemesh/assets/1964576/fed48137-7497-429b-b1f0-14d0db119bd3)

in fact all this objects are available in memory in the tortoise, so there is absolutely no reason to waste resources on doing this database lookups.

another concern is that this is not robust, if tortoise doesn't have dependency when dependant object is processed - it will fail validation and it will retry hopefully in some future.

to address both of those concern component that maintain state should be able to provide information about missing objects (for example as an error), and fetcher should do request based on missing object, regardless of that is stored in the database. after fetching complete validation should be re-done. in other words it should be up to handler to check that data is available, some handlers may do it using database, but in some places we already have all necessary data in memory.

@countvonzero

dshulyak:
![image](https://github.com/spacemeshos/go-spacemesh/assets/1964576/fb4bc2c8-dede-474d-8d91-a806af579057)

PR #4935, opened 2023-08-31
[Merged by Bors] - use tortoise data for ballot eligibility validation

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/4927

this change eliminates database lookups on ballot validation path, with an exception of when lru cache is not sufficient to hold
vrf nonces and activations. in that case we will have to load them from db, that can be optimized later by fitting more in lru cache https://github.com/spacemeshos/go-spacemesh/pull/4935 or reusing tortoise dataset as well

- fetcher will be asked for ballots only if we don't have them in memory
- reference ballots for eligibility validation will be fetched from tortoise

### DIFF
--- a/proposals/eligibility_validator.go
+++ b/proposals/eligibility_validator.go
@@ -10,7 +10,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/datastore"
 	"github.com/spacemeshos/go-spacemesh/log"
-	"github.com/spacemeshos/go-spacemesh/sql/ballots"
 	"github.com/spacemeshos/go-spacemesh/system"
 )
 
@@ -30,6 +29,7 @@ type Validator struct {
 	minActiveSetWeight uint64
 	avgLayerSize       uint32
 	layersPerEpoch     uint32
+	tortoise           ballotDecoder
 	cdb                *datastore.CachedDB
 	clock              layerClock
 	beacons            system.BeaconCollector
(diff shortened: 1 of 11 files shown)


dshulyak:
image

proposals/interface.go
```diff
@@ -20,6 +20,7 @@ type eligibilityValidator interface {
 }
 
 type ballotDecoder interface {
```
dshulyak: need to rename this

proposals/eligibility_validator.go
```diff
@@ -162,27 +163,24 @@ func (v *Validator) validateReference(ballot *types.Ballot, owned *types.Activat
...
+		if ballot.EpochData == nil {
+			return nil, fmt.Errorf("%w: ref ballot %v", errMissingEpochData, ballot.ID())
 		}
+		return ballot.EpochData, nil
 	}
-	if refballot.EpochData == nil {
-		return nil, fmt.Errorf("%w: ref ballot %v", errMissingEpochData, refballot.ID())
+	refdata := v.tortoise.GetBallot(ballot.RefBallot)
+	if refdata == nil {
+		return nil, fmt.Errorf("ref ballot is missing %v", ballot.RefBallot)
```
countvonzero, line 174: in general my concern is when data is outside tortoise's window. why not query this from db in case of failure?

countvonzero (approved):

proposals/eligibility_validator.go
dshulyak: if data is outside window, we can't decode ballot votes anymore. there have to be some heuristic to reload the window at that time, like a lot of atx weight wasn't counted

this should be a prerequisite for healing for longer than window length

dshulyak:
bors merge

PR #4936, opened 2023-08-31
[Merged by Bors] - hare/oracle: reset oracle when node enters synced state

### LINKED ISSUE #4552: hare loads incorrect (tortoise) active set during sync
i noticed that during sync layers are not processed immediately, quite often they get stuck for unclear reason, sync during this time downloads more layers and forces hare to cache active set which would be incorrect later.

i think that current implementation is not robust, and should be based on event notifications, so that a particular timing of calls won't have an impact on the end result. in this particular case we should add OnAppliedBlock to hare oracle so that it can update local cache if block was applied at a later point in time

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4552

the none invasive solution to mitigate problem with loading wrong activeset during sync, until we switch to graded oracle


countvonzero (approved):

dshulyak:
bors merge

PR #4937, opened 2023-08-31
[Merged by Bors] - hare3: switch to sync validation for gossip messages

dshulyak:
currently every validation request spawns new goroutine, the main reason for that was
to enable fetching over network without blocking other threads.

however hare3 will be spawning many goroutines and it will be much more efficient to execute protocol inline.
the same is true for beacon.


dshulyak:
bors try

dshulyak:
bors try

countvonzero (approved):

dshulyak:
bors merge

PR #4944, opened 2023-09-01
[Merged by Bors] - tortoise: load ref ballots before other

dshulyak:
this can be seen as attack, where ref ballots is published, everyone is using that as base ballot, 
and then ref ballot gets cancelled. 
we still need to load it in state, otherwise it may block counting honest ballots


countvonzero (approved):

countvonzero:
bors merge

PR #4945, opened 2023-09-01
[Merged by Bors] - miner: parametrize threshold for picking active set from first block

dshulyak:
this is not a protocol parameter, no reason to hardcode, this is local heuristic for a node to decide if it wants to build block from atxs synced on time, or from first block


countvonzero (approved):

countvonzero:
bors merge

Issue #4632, opened 2023-07-04
prevent dos'ing network with equivocating data


countvonzero:
## Description
currently data object reference other data for various reasons.

- each atxs reference other atx for its
  - commitment atx. this is baked in post data (only 1 per atx chain) 
  - positioning atx (1 per new atx). this is included in the poet proof, and is used to calculate it's base tick height. 

- each ballot reference other ballot for its
  - reference ballot (miner's first ballot in the epoch) if it itself is not a reference ballot. reference ballots contain the beacon and the active set
  - base ballot for efficient encoding of votes
  
when equivocation is detected, we should stop accepting atxs / ballots from such malicious identity, unless these objects are already referenced by other objects from honest identities.

## Actual Behavior
- malicious atx/ballots are always saved
- malfeasance proof is saved and gossiped only when the first equivocating atx/ballot is received
- any malicious gossip atx/ballots after the malfeasance proof is generated/gossiped are re-gossiped. i.e. all equivocating atxs/ballots after the 1st one are saved/regossiped

## Expected Behavior
- equivocating atxs/ballots should not be saved
- equivocating atxs/ballots should not be gossiped
- malfeasance proof is saved and gossiped only when the first equivocating atx/ballot is received
- equivocating atxs/ballots should only be saved on the sync path, where atxs/ballots are fetched according to existing dependencies from honest atxs/ballot.

- less importantly, given the same tick height, prefer own atx as positioning atx.

dshulyak:
on gossip path drop any atxs from malicious identities, including second one (when proof is generated). 
on sync path request dependencies only if original object passed validation (eligibility for ballot, nipost for atx) and identity didn't equivocate. this can be accomplished by reordering validation

PR #4952, opened 2023-09-02
[Merged by Bors] - miner: do not omit own atx in active set

countvonzero:
## Motivation
syncing from devnet-405, malicious identities filter out own atx and cause other nodes to fail processing the ballot

```
2023-09-02T14:58:23.344-0700	WARN	5fbb1.sync	failed fetching new ballots	{"node_id": "5fbb17185b5850709b3c545e3be4fbea25a30d3362a0627d974495e96d8b640b", "module": "sync", "sessionId": "6123c073-483d-41ad-b514-77e47167f48f", "layer_id": 11549, "ballot_ids": ["6a70a5a8d0", "f3205c8f79", "1efadf7bbe", "dc0881473b", "57577ea37f", "1f3110b841", "c9f0910457", "f2e8dac11e", "c97b5ff911", "f824639d3b"], "errmsg": "hint: ballotDB, hash: 0x1f3110b8411255368ffc35e90c2f17ad3a898a13000000000000000000000000, err: fetch ballots: hint: ballotDB, hash: 0x926af68ad85b7db9d5a0fbd57ddefdbb32838149000000000000000000000000, err: ballot not eligible: atx 086658fac6 from ballot 926af68ad8 (refballot 926af68ad8) is not included into the active set\nhint: ballotDB, hash: 0xc9f0910457dfab80334c80e77cf6673fe408adcb000000000000000000000000, err: fetch ballots: hint: ballotDB, hash: 0x37df2b14296708054511e2627f1094f5b8528eb3000000000000000000000000, err: fetch ballots: hint: ballotDB, hash: 0x5b91d74480e89f741045aa9c4f95501b02fc3d53000000000000000000000000, err: ballot not eligible: atx e9aff052ae from ballot 5b91d74480 (refballot 5b91d74480) is not included into the active set", "name": "sync"}
```

### DIFF
--- a/miner/oracle.go
+++ b/miner/oracle.go
@@ -155,7 +155,7 @@ func (o *Oracle) activeSet(targetEpoch types.EpochID) (uint64, uint64, types.ATX
 		if err != nil {
 			return err
 		}
-		if grade != Good {
+		if grade != Good && header.NodeID != o.cfg.nodeID {
 			o.log.With().Info("atx omitted from active set",
 				header.ID,
 				log.Int("grade", int(grade)),
--- a/miner/oracle_test.go
+++ b/miner/oracle_test.go
@@ -300,57 +300,77 @@ func TestOracle_MinimalActiveSetWeight(t *testing.T) {
 }
 
 func TestOracle_ATXGrade(t *testing.T) {
-	avgLayerSize := uint32(50)
-	layersPerEpoch := uint32(10)
-	o := createTestOracle(t, avgLayerSize, layersPerEpoch, 0)
-	lid := types.LayerID(layersPerEpoch * 3)
-	epochStart := time.Now()
-	o.mClock.EXPECT().LayerToTime(lid).Return(epochStart)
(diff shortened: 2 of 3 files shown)


dshulyak:
are there any problems with that? the identity is cancelled anyway, so why would we want ballot from it?

dshulyak (approved):

countvonzero:
> are there any problems with that? the identity is cancelled anyway, so why would we want ballot from it?

the only problem is that node will keep syncing the same thing (because it's used as a reference ballot and will be recursively fetched before it's checked for maliciousness).

there is a bigger issue around this https://github.com/spacemeshos/go-spacemesh/issues/4632
eventually we don't want to save more malicious data than we have to. but for now, until we fix that issue, and also add the ability to swap out base ballot, we need to download such ballot.

in any case, this is consistent with rational miner behavior.

countvonzero:
bors merge

countvonzero:
bors merge

PR #4956, opened 2023-09-04
[Merged by Bors] - ballots: request active set if reference ballot was submitted with empty active set

### LINKED ISSUE #4672: fetcher, ballots: request active set using active set hash
every reference ballot will includes all atx reference's in the gossiped message.  
see deep purple graph, it starts at about ~150KB/s at the start of the epoch, when everybody creates references ballots, and goes lower over time.

we already have active set hash in place that should allow to download ballot without active set side car and then request sidecar only if we see it for the first time. we probably should track in memory what active set hashes we already have seen, and request new only if it is not in memory

image

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4672

starting from layer X reference ballots will be gossiped with empty active sets.
any updated node will request active set by its hash from the ballot sender, and save it in a separate table. unless  it wasn't saved before, in that case we won't download anything.

reference ballot will be stored with full active set, so that sync protocol remains compatible with nodes that didn't upgrade. for them reference ballot without active set will look syntactically invalid. but they will be able to download it later and not get stuck this way.


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

sql/migrations/0003_next.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE activesets
+(
+    id     CHAR(32) PRIMARY KEY,
```
countvonzero, line 3: how about also index by epoch id for ease of pruning later?

sql/activesets/activesets.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/codec"
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+func Add(db sql.Executor, id types.Hash32, set *types.EpochActiveSet) error {
+	buf, err := codec.Encode(set)
+	if err != nil {
```
countvonzero: mustencode?

sql/activesets/activesets.go
```diff
@@ -0,0 +1,75 @@
...
+		return fmt.Errorf("encode active set %s: %w", id, err)
+	}
+	_, err = db.Exec(`insert into activesets
+		(id, active_set)
+		values (?1, ?2);`,
```
countvonzero, line 14: on duplicate do nothing - for app level concurrency.

proposals/handler.go
```diff
@@ -431,17 +456,28 @@ func (h *Handler) checkBallotDataIntegrity(b *types.Ballot) error {
...
+			}
+			set, err := activesets.Get(h.cdb, b.EpochData.ActiveSetHash)
+			if err != nil {
+				return err
+			}
+			// NOTE(dshulyak) sidecar is stored for every ballot, so that
+			// nodes that won't update on time will be able to download it
+			// with sync
+			b.ActiveSet = set.Set
+			if len(b.ActiveSet) == 0 {
```
countvonzero: seems more coherent to check it right after call to activesets.Get()

proposals/handler.go
```diff
@@ -431,17 +456,28 @@ func (h *Handler) checkBallotDataIntegrity(b *types.Ballot) error {
...
+			}
+		} else {
+			if err := h.fetcher.GetActiveSet(ctx, b.EpochData.ActiveSetHash); err != nil {
+				return err
+			}
+			set, err := activesets.Get(h.cdb, b.EpochData.ActiveSetHash)
+			if err != nil {
+				return err
+			}
+			// NOTE(dshulyak) sidecar is stored for every ballot, so that
```
countvonzero: i don't understand this need. 
old binary and code in other component all expect to find active set in ref ballot.
but the new code inject it in every ballot (and the resulting blob).
- code becomes confusing (why chase the ref ballot in other code path when it's saved in every ballot?)
- more importantly, when nodes sync ballots, they get repeated data from the ballot blobs, which bloat up the db size unnecessarily.

i think it's sufficient to retain the original behavior - only saving it along with the ref ballot (if epochData != nil)
it's also better on the downgrade path (for some reason we had to roll back this change). the network will consistently save ref ballots with active set.

miner/proposal_builder.go
```diff
@@ -139,6 +141,12 @@ func WithMinGoodAtxPct(pct int) Opt {
 	}
 }
 
+func WithEmitEmptyActiveSet(lid types.LayerID) Opt {
```
countvonzero, line 155: should be called with the tortoise param?

btw, why is it a tortoise param?

proposals/handler.go
```diff
@@ -180,27 +179,53 @@ func (h *Handler) HandleSyncedBallot(ctx context.Context, expHash types.Hash32,
 	return nil
 }
 
-// collectHashes gathers all hashes in a proposal or ballot.
-func collectHashes(a any) []types.Hash32 {
-	p, ok := a.(types.Proposal)
-	if ok {
-		hashes := collectHashes(p.Ballot)
-		return append(hashes, types.TransactionIDsToHashes(p.TxIDs)...)
+func (h *Handler) HandleActiveSet(ctx context.Context, id types.Hash32, peer p2p.Peer, data []byte) error {
```
countvonzero, line 183: should disallow any ref ballot without empty active set before the scheduled layer.

proposals/handler.go
```diff
@@ -431,17 +456,28 @@ func (h *Handler) checkBallotDataIntegrity(b *types.Ballot) error {
...
-				return errActiveSetNotSorted
+		if len(b.ActiveSet) != 0 {
+			if err := h.handleSet(ctx, b.EpochData.ActiveSetHash, types.EpochActiveSet{
+				Epoch: b.Layer.GetEpoch(),
+				Set:   b.ActiveSet,
+			}); err != nil {
+				return err
+			}
+		} else {
+			if err := h.fetcher.GetActiveSet(ctx, b.EpochData.ActiveSetHash); err != nil {
```
countvonzero, line 471: if this ref ballot is before the scheduled layer, this is an invalid ballot.
should only call h.fetcher.GetActiveSet() after the scheduled layer.
otherwise the node will be DOSing its peers on a bad ballot.

proposals/handler.go
```diff
@@ -180,27 +179,53 @@ func (h *Handler) HandleSyncedBallot(ctx context.Context, expHash types.Hash32,
...
+	h.fetcher.RegisterPeerHashes(peer, types.ATXIDsToHashes(set.Set))
+	return h.handleSet(ctx, id, set)
+}
 
-	b, ok := a.(types.Ballot)
-	if ok {
-		hashes := []types.Hash32{b.Votes.Base.AsHash32()}
-		if b.RefBallot != types.EmptyBallotID {
-			hashes = append(hashes, b.RefBallot.AsHash32())
+func (h *Handler) handleSet(ctx context.Context, id types.Hash32, set types.EpochActiveSet) error {
```
countvonzero, line 192: add metrics for how many different active set hash we've seen from the network for this epoch. this would help us understand whether we should improve with base active set + diff

countvonzero (commented):
so after the layer takes effect, the new-version nodes will see `blob store not found activeset` from old-version nodes. but the old-version nodes will continue to operate fine. this seems to punish miners who upgraded...

proposals/handler.go
dshulyak: sorry it is bad comment, it stores only for ref ballot

sql/activesets/activesets.go
dshulyak: it will report that object exists, it is handled on the caller side

sql/migrations/0003_next.sql
dshulyak: i tend to think that you can just scan whole table and decode epoch. index will make it faster, but this table will remain small if we will prune everything from previous epochs

proposals/handler.go
dshulyak: i don't think that such code is doing anything useful

proposals/handler.go
dshulyak: why is it invalid? it for sure won't be rewarded, but otherwise such ballot is ok. or you mean that someone maliciously sends such ballot?

miner/proposal_builder.go
dshulyak: i think that this is a part of tortoise, e.g not the core protocol but ballot is a part of tortoise. it is just unfortunate separation

proposals/handler.go
dshulyak: i don't want to maintain map for that. we can check growth by looking table size and number or records in table (second we don't have, but should be added)

dshulyak:
> so after the layer takes effect, the new-version nodes will see blob store not found activeset from old-version nodes. but the old-version nodes will continue to operate fine. this seems to punish miners who upgraded...

they will ask peer that sent this data. unless someone maliciously sends ballot with empty active set then it is ok.

as for dos it makes no sense, it is very simple work (query one key). at the same time you can ask sync for atx ids, which is a range scan

proposals/handler.go
dshulyak: i think this is wrongheaded to try to solve this dos, when it is actually very simple work. and at the same time there are actually dos'able rpc's

dshulyak: i added that condition

proposals/handler.go
dshulyak: i added that condition

dshulyak:
bors try

dshulyak:
spacemeshos/go-spacemesh-dev:b22e6d9

miner/proposal_builder.go
countvonzero, line 155: i still don't see this invoked in node.go. what am i missing?

proposals/handler.go
countvonzero: yeah... there are many existing wrongs. 
here it seems a relative simple check to add.

countvonzero (approved):

miner/proposal_builder.go
dshulyak: oh sorry :)  i thought you are suggesting to rename it

dshulyak:
bors merge

PR #4957, opened 2023-09-04
api, activation: regossip stored atxs for smesher 

dshulyak:
- [ ] handler needs to allow to publish atx from self


maxpla3:
I have 11 out of 18 ATX "stuck" due to internet connection problems during the last CG.
I can see all 18 ATX in state.sql of all nodes in my LAN.
I'd be happy to test this feature.

dshulyak:
there will be binaries here https://github.com/spacemeshos/go-spacemesh/releases/tag/regatxs-v1.1.4

they will have v1.1.4 with this change. api examples:

> grpcurl -plaintext -d '{"id": "eaac7bc049bc3243dc553bdcf39b50a473b7a115d49aac998754a679d47d170"}' 127.0.0.1:9093 spacemesh.v1.SmesherService.RegossipAtxs

will lookup atx by id

> grpcurl -plaintext -d '{"epoch": 3}' 127.0.0.1:9093 spacemesh.v1.SmesherService.RegossipAtx

will lookup by epoch in which atx was published and smesher public key that configured for this node

maxpla3:
> grpcurl -plaintext -d '{"id": "eaac7bc049bc3243dc553bdcf39b50a473b7a115d49aac998754a679d47d170"}' 127.0.0.1:9093 spacemesh.v1.SmesherService.RegossipAtxs

> grpcurl -plaintext -d '{"epoch": 3}' 127.0.0.1:9093 spacemesh.v1.SmesherService.RegossipAtx

The difference between these two calls is that in the first case a node can regossip ATX of other nodes while in the second case it will regossip only its own ATX?

What is the expected behavior? Should all peers have the gossiped ATX after any of these two calls?

dshulyak:
> The difference between these two calls is that in the first case a node can regossip ATX of other nodes while in the second case it will regossip only its own ATX?

yes, in first it can be any atx. second one is just more convenient

> What is the expected behavior? Should all peers have the gossiped ATX after any of these two calls?

yes that would be expected. every will start receiving gossiped atx.

Thanos420NoScope:
What is the correct usage for this? I am running out of time and cannot figure it out. 
The examples provided do not work as intended.

```
Error invoking method "spacemesh.v1.SmesherService.RegossipAtx": service "spacemesh.v1.SmesherService" does not include a method named "RegossipAtx"
```

```
Error invoking method "spacemesh.v1.SmesherService.RegossipAtxs": error getting request data: invalid character 'i' looking for beginning of object key string
```

PR #4966, opened 2023-09-06
[Merged by Bors] - use tortoise data to enforce that referenced atxs are available when processing block

### LINKED ISSUE #4590: blocks, sync, atxs: enforce availability of the referenced atx when receiving a block
otherwise tortoise will vote on a block that maybe included unexisting atx. this will halt consensus

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4590

if block with invalid atxs crosses tortoise threshold everyone will upvote it but won't be able to apply, as such atx won't exist.
therefore we need to download any referenced atxs, and if such atx is not available then block is not accepted at all.


blocks/handler.go
```diff
@@ -72,11 +73,11 @@ func (h *Handler) HandleSyncedBlock(ctx context.Context, expHash types.Hash32, p
 	}
 
 	if b.LayerIndex <= types.GetEffectiveGenesis() {
-		return fmt.Errorf("block before effective genesis: layer %v", b.LayerIndex)
+		return fmt.Errorf("%w: block before effective genesis: layer %v", pubsub.ErrValidationReject, b.LayerIndex)
 	}
 
 	if err := ValidateRewards(b.Rewards); err != nil {
-		return fmt.Errorf("%w: %w", errInvalidRewards, err)
+		return fmt.Errorf("%w: %s", pubsub.ErrValidationReject, err.Error())
```
fasmat, line 80: There is a new API for this introduced in Go 1.20:

```suggestion
		return errors.Join(pubsub.ErrValidationReject, err)
```

fasmat (approved):

countvonzero (approved):

blocks/handler.go
dshulyak: i don't think that all errors should be chained by default. error is a part of api, so that it can be handled properly.  nobody cares about errInvalidRewards (should have been just string), thats just an information. pubsub.ErrValidationReject part of api

dshulyak:
bors merge

Issue #4857, opened 2023-08-16
malfeasance: don't accept unknown messages from remote sources signed by your own identity


dshulyak:
when node publishes a message it always saves it to the persistent store first, at least there is always should be one atx before node can participate in other protocols that don't persist messages.

if node receives different object that it doesn't know about:
- either it deleted state and needs resync
- someone runs node with the same smesher key

- [ ] don't save unknown ballots/proposals/atxs from your own identity without an acknowledgement from user

if option `accept-identity-messages` set to false and node received objects signed by your identity from a p2p id different then yours then exit the node with FATAL error and instructions how that `accept-identity-messages` needs to be set to true

- [ ] randomly delay publishing atx for a short period of time between [0, 5m]

one of the node will be able to detect equivocating atx using the method above if they won't publish exactly at the same time.

- [ ] don't submit malicious proofs for your own identity regardless if the option above was enabled

instead node should exit with FATAL error and report to user that equivocating node is detected

lrettig:
> * [ ]  randomly delay publishing atx for a short period of time between [0, 5m]
> 
> one of the node will be able to detect equivocating atx using the method above if they won't publish exactly at the same time.

@pigmej suggests that we don't actually need to do this, because the way PoST proof construction works now it's highly unlikely that two nodes (even similarly configured) would produce and gossip proofs at the same time (within ~30 sec of one another)

dshulyak:
i think 30s is too optimistic, with moderate load i would expect atx gossip to be in minutes

Issue #4856, opened 2023-08-16
malfeasance: add machine fingerprint when post data/smesher key is generated


dshulyak:
we want to prevent accidental private key reuse across several machines on copying post data. 
if there is a fingerprint from unique non-ephemeral data we can enforce that it remains the same across go-spacemesh restarts.
if it is not the same we will ask user to acknowledge that the state was indeed moved and either delete specified fingerprint (or app can help with it). this fingerprint should also enable users to run docker without additional issues.

fingerprint should contain unique fields specific to hardware:
- motherboard serial number (/sys/class/dmi/id/board_serial)
- macaddress (ip a)
- ? 

in particular on linux there is already a machine id, i assume generated from such sources as above and more. 

on startup node should compute fingerprint from hardware and read fingerprint file from disk, if it doesn't match the FATAL error should be returned to user with instructions on how to bypass it. several options:

- specify a file that has to be removed 
- provide link in the error message to instructions
- request to add option to config that will recompute file, this is not so good as people likely will add it and forger

besides that we need to make sure that:
- initialization on multiple machines is working
- node needs to have a defined behavior when files are not found: regenerate / create new ones.
- a resulting code MUST work in containerized setups.

fingerprint should be generated together with key, specifically as an upgrade path we should always generate it if not found, and after couple of weeks start failing if fingerprint is not found, but key exists.

dshulyak:
@pigmej shared a concern that if you generated data and key with postcli, then you may copy it from that place to multiple machines. we want to fail if there is no fingerprint, but for 1 week it will be slightly more risky to do this kind of stuff. beside that people that are using postcli should be mindful of doing such things anyway

lrettig:
Regarding how we calculate the fingerprint, I propose the following:

- on systems/operating systems that already have something akin to a "serial number", use this - but in most cases this seems to require superuser/admin access (e.g., [`dmidecode -s system-uuid` on Linux](https://serverfault.com/questions/733069/i-need-a-command-to-create-a-unique-fingerprint-from-unix-machine)) so I'm not sure how helpful this is; otherwise:
- use the UUID (or its equivalent on other operating systems) of the filesystem where the PoST data resides - this is also probably sufficient on its own; otherwise:
- use the OS/platform (e.g., `linux-amd64`), and
- the filesystem location of the PoST data, and
- the MAC address of the default interface and/or the UUID of the root filesystem (which doesn't require root access)

[...]

Issue #4594, opened 2023-06-26
Node should *not* delete POST files when there is more on a disk than configured


pigmej:
In the situation when there are more post files than node expects node should error and give up, and not delete the files. 

That should help not surprise users with "your data is gone because config was wrong".

In the situation when there are too less files, then node should clearly indicate that it will generate more. 

There probably should be a flag that would change node behavior:
* `--do-not-touch-pos` and then it should fail regardless of the found situation on the disk (smaller, bigger etc).
* we could have a flag that would keep current behaviour too.

schinzelh:
I just lost 1 TiB post data due to this "feature" - please fix this ASAP!

lrettig:
See also https://github.com/spacemeshos/post/issues/236

pigmej:
It will be addressed with https://github.com/spacemeshos/pm/issues/259

lrettig:
This is [still happening](https://discord.com/channels/623195163510046732/691262492390719498/1156883602752143371)

Stizerg:
I lost 4TiB today, when will you fix it? 
--do-not-touch-pos should be default behavior.
there's so many complains since testnet, more than 6 month later the issue haven't been fixed

pigmej:
Post-service implementation will have that fixed, as mentioned https://github.com/spacemeshos/go-spacemesh/issues/4594#issuecomment-1733625198

ytx1991:
I believe the issue could be resolved in less than 10 minutes; however, it seems there is a lack of willingness to address user concerns promptly.

lrettig:
> I believe the issue could be resolved in less than 10 minutes; however, it seems there is a lack of willingness to address user concerns promptly.

Our small development team is very busy focusing on network stability. This is an open source project. Feel free to make the desired change and submit it as a PR here! We're more than happy to review.

For the record, if it were a "ten minute change" we would've made it a long time ago. The change is more complex as it ties into the post service work that @pigmej mentioned above.

pigmej:
This is fully solved with post-service.

PR #4981, opened 2023-09-07
do not mark self malicious

countvonzero:
## Motivation
to ease adoption of testnet

## Changes
- statically set miner's own node id
- whenever malfeasance proof is generated, check that it's not self

### DIFF
--- a/common/types/nodeid.go
+++ b/common/types/nodeid.go
@@ -9,6 +9,18 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
+var ownNodeID NodeID
+
+func SetMinerNodeID(id NodeID) {
+	if ownNodeID == EmptyNodeID {
+		ownNodeID = id
+	}
+}
+
+func MinerNodeID() NodeID {
+	return ownNodeID
+}
+
 // BytesToNodeID is a helper to copy buffer into NodeID struct.
 func BytesToNodeID(buf []byte) (id NodeID) {
 	copy(id[:], buf)
--- a/hare3/hare.go
+++ b/hare3/hare.go
@@ -286,7 +286,7 @@ func (h *Hare) Handler(ctx context.Context, peer p2p.Peer, buf []byte) error {
 	gossip, equivocation := session.OnInput(input)
 	h.log.Debug("after on message", log.ZShortStringer("hash", input.msgHash), zap.Bool("gossip", gossip))
 	submitLatency.Observe(time.Since(start).Seconds())
-	if equivocation != nil && !malicious {
+	if equivocation != nil && !malicious && msg.Sender != types.MinerNodeID() {
 		h.log.Debug("registered equivocation",
 			zap.Uint32("lid", msg.Layer.Uint32()),
 			zap.Stringer("sender", equivocation.Messages[0].SmesherID))
--- a/hare3/hare_test.go
+++ b/hare3/hare_test.go
@@ -699,6 +699,61 @@ func TestHandler(t *testing.T) {
 	})
 }
 
+func TestHandlerOwnEquivocation(t *testing.T) {
+	t.Parallel()
+	tst := &tester{
+		TB:            t,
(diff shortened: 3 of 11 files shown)


hare3/hare_test.go
```diff
@@ -699,6 +699,61 @@ func TestHandler(t *testing.T) {
 	})
 }
 
+func TestHandlerOwnEquivocation(t *testing.T) {
```
countvonzero, line 702: sorry. i couldn't find a way to add it in TestHandler. 
i didn't know how to advance to a round that the node is eligible again

dshulyak: i really prefer not to add such tests, i can't work with them

dshulyak: why not to extend TestHandler? this is almost the same as existing test, but with exception of global state

dshulyak:
i think the solution for this problem was somewhat different https://github.com/spacemeshos/go-spacemesh/issues/4857

common/types/nodeid.go
```diff
@@ -9,6 +9,18 @@ import (
 	"github.com/spacemeshos/go-spacemesh/log"
 )
 
+var ownNodeID NodeID
```
dshulyak, line 12: i don't think that adding global public key is acceptable even as a mitigation.

beside that this code already can be removed next week

hare3/hare.go
```diff
@@ -286,7 +286,7 @@ func (h *Hare) Handler(ctx context.Context, peer p2p.Peer, buf []byte) error {
 	gossip, equivocation := session.OnInput(input)
 	h.log.Debug("after on message", log.ZShortStringer("hash", input.msgHash), zap.Bool("gossip", gossip))
 	submitLatency.Observe(time.Since(start).Seconds())
-	if equivocation != nil && !malicious {
+	if equivocation != nil && !malicious && msg.Sender != types.MinerNodeID() {
```
dshulyak, line 289: *Hare has reference to signer

dshulyak:
i don't understand what it fixes? if node sent an equivocating message thats enough, it doesn't have to generate a proof for itself. the latter is certainly should not be done, but by itself it doesn't fix much. this code also doesn't exit with fatal.

the solution in https://github.com/spacemeshos/go-spacemesh/issues/4857 prevents some configuration mistakes, but not all

hare3/hare_test.go
dshulyak: i think the way to go to introduce such test is to make initialization reusable, instead of copying it

hare3/hare_test.go
```diff
@@ -699,6 +699,61 @@ func TestHandler(t *testing.T) {
...
+	msg1.Signature = n.signer.Sign(signing.HARE, msg1.ToMetadata().ToBytes())
+
+	msg2 := &Message{}
+	msg2.Layer = layer
+	msg2.Value.Proposals = []types.ProposalID{{2}}
+	msg2.Eligibility = *elig
+	msg2.Sender = n.signer.NodeID()
+	msg2.Signature = n.signer.Sign(signing.HARE, msg2.ToMetadata().ToBytes())
+
+	types.SetMinerNodeID(n.hare.signer.NodeID())
```
dshulyak, line 741: this global state can't work with t.Parallel, but hare already has access to signer

dshulyak:
but anyway, i think this is not the approach. it doesn't solve any actual problems, introduces global state and code that needs to be removed "tomorrow".

the solutions that addresses some problems are:
- https://github.com/spacemeshos/go-spacemesh/issues/4856 to prevent copying state
- https://github.com/spacemeshos/go-spacemesh/issues/4857 to prevent syncing node with key (the simplest is to check in atx handler)
- https://github.com/spacemeshos/go-spacemesh/issues/4594
- we also discussed how to prevent accidental launches on same computer. right now it is not possible because of leveldb used in libp2p . but maybe we should move lockfile relative to datadir

PR #4986, opened 2023-09-08
[Merged by Bors] - fix regression in proposals sync

dshulyak:
on sync path activeset is downloaded together with proposals so that not updated nodes can download data.
however starting at specific layer (in systests in all layers) we only sign empty activeset, so after downloading proposals it should not pass this validation.


fasmat (approved):

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

PR #4987, opened 2023-09-08
[Merged by Bors] - testnet with 1 day settinng and 1024 labels per unit

dshulyak:
- epoch 288 layers, 1 day
- unit size 1024 labels
- beacon voting rounds from 300 to 20
- genesis is set for 2023-09-13T18:00:00Z


config/presets/testnet.go
```diff
@@ -1,73 +1,145 @@
...
+			MaxRequestRetries: 10,
+		},
+		POST: activation.PostConfig{
+			MinNumUnits:   4,
+			MaxNumUnits:   math.MaxUint32,
+			LabelsPerUnit: 1024,
+			K1:            26,
+			K2:            37,
+			K3:            37,
+			PowDifficulty: postPowDifficulty,
```
poszu: Do we want to keep mainnet difficulty for K2pow? It will probably take more time than reading the data (the difficulty scales with the number of units, not the number of labels).

If so, consider:
```suggestion
			PowDifficulty: activation.PowDifficulty(config.MainnetConfig().PowDifficulty),
```
from `github.com/spacemeshos/post/config`

config/presets/testnet.go
```diff
@@ -1,73 +1,145 @@
...
+			WeakCoinRoundDuration:    4 * time.Minute,
+			VotesLimit:               100,
+			BeaconSyncWeightUnits:    800,
+		},
+		POET: activation.PoetConfig{
+			PhaseShift:        12 * time.Hour,
+			CycleGap:          2 * time.Hour,
+			GracePeriod:       10 * time.Minute,
+			RequestRetryDelay: 10 * time.Second,
+			MaxRequestRetries: 10,
```
poszu: The total request retry time in case of 10 retries could be bigger than the grace period.

```
min_total_retry_time = (10 + 10*10) / 2 * 10 = 550s
max_total_retry_time = 2* min_total_retry_time = 1100s
```
because we set max retry to `2*RequestRetryDelay`.

see https://github.com/hashicorp/go-retryablehttp/blob/571a88bc9c3b7c64575f0e9b0f646af1510f2c76/client.go#L539-L575

```suggestion
			GracePeriod:       10 * time.Minute,
			RequestRetryDelay: 5 * time.Second,
			MaxRequestRetries: 10,
```

dshulyak: can you tell me what i should set?

config/presets/testnet.go
dshulyak: i prefer not to init mainnet config

countvonzero (approved):
my assumption is that you will deploy hare3 too?

config/presets/testnet.go
poszu: What do you mean? 🤔

dshulyak:
> my assumption is that you will deploy hare3 too?

the plan was to do it with upgrade

dshulyak:
bors merge

config/presets/testnet.go
dshulyak: thats what you suggested?

> activation.PowDifficulty(config.MainnetConfig().PowDifficulty)

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

config/presets/testnet.go
poszu: It's equivalent, just using the mainnet setting defined in `post` repo. Actually, if you want mainnet POST config, you can just use the whole `config.MainnetConfig()`.

I first asked if you really want to have the same k2pow difficulty as on mainnet. K2pow will most likely take a lot of time for smeshers that init many SU (it could take more than the 2h cycle gap).

poszu: OK, I see it's reduced already :+1:

PR #4988, opened 2023-09-08
[Merged by Bors] - 4783 poet client logging

### LINKED ISSUE #4783: Don't use a global logger in the poet client
## Description
Currently, the poet client uses a global logger.

## Affected code
https://github.com/spacemeshos/go-spacemesh/blob/develop/activation/poet.go

## Acceptance criteria

- pass `log.Log` to `HTTPPoetClient`,
- pass the logger to `retryablehttp.Client` too.

poszu:
## Motivation
Part of  #4751 
Closes #4783

## Changes
- use local logger in poet clients
- provide the logger to `retryablehttp` to gain visibility on performed requests, retries, responses etc.
- cleanup poet test harness

## Test Plan
Existing tests pass


activation/poet.go
```diff
@@ -53,6 +54,28 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
...
 }
 
+// A wrapper around zap.Logger to make it compatible with
+// retryablehttp.LeveledLogger interface.
+type retryableHttpLogger struct {
+	inner *zap.Logger
+}
+
+func (r retryableHttpLogger) Error(format string, args ...any) {
+	r.inner.Sugar().With(args...).Error(format)
```
fasmat: We should be careful not to log too much with this logger, since every call will allocate every `args` at least twice I think (once on the caller side and once within `Sugar`) even if the logs for that level are disabled 🤔

activation/poet.go
```diff
@@ -61,11 +84,22 @@ func withCustomHttpClient(client *http.Client) PoetClientOpts {
 	}
 }
 
+func WithLogger(logger *zap.Logger) PoetClientOpts {
+	return func(c *HTTPPoetClient) {
+		c.logger = logger
+		c.client.Logger = &retryableHttpLogger{inner: logger}
```
fasmat, line 90: I don't think this `Logger` is very useful for us, it only prints 2 possible messages:

- `DEBUG` performing request ...
- `ERROR` when failing to drain the request body in the case of an error (that error is not logged with `Logger`

I think we should set `RequestLogHook` instead and not use the `Logger` for the possible performance impact when logging every request on `DEBUG` which `RequestLogHook` would also do and we can use the structured logger instead of the sugared 🙂

poszu: >I don't think this Logger is very useful for us, it only prints 2 possible messages
Not exactly, check out this example retried request:
```
DEBUG	poet	performing request	{"method": "GET", "url": "http://127.0.0.1:38837/v1/proofs/0"}
INFO	poet	response received	{"url": "http://127.0.0.1:38837/v1/proofs/0", "status": 404} // <-- from ResponseLogHook
DEBUG	poet	retrying request	{"request": "GET http://127.0.0.1:38837/v1/proofs/0 (status: 404)", "timeout": "107.918576ms", "remaining": 10}
INFO	poet	response received	{"url": "http://127.0.0.1:38837/v1/proofs/0", "status": 200}
```

> I think we should set RequestLogHook instead and not use the Logger for the possible performance impact

IMHO it doesn't make **any** difference. There are literally only a handful of requests to PoET.

fasmat: Can you try with setting the `RequestLogHook` instead of the `Logger`? It should print the same (and possibly more) information but not cause any allocations if logging is disabled (while the `retryableHttpLogger` will).

activation/poet.go
poszu: Fixed by not using `.With()` to pass args, but using `Infow/Debugw` and so on - args will only be formatted if the log will be emitted.

fasmat (approved):

poszu:
Bors merge

poszu:
Failed `TestCheckpoint` which seems unrelated.

Bors merge

PR #4993, opened 2023-09-09
[Merged by Bors] - prune old proposals after block is generated

countvonzero:
## Motivation
prune mesh data

## Changes
after block is generated for layer N, prune proposals < N

### DIFF
--- a/blocks/generator.go
+++ b/blocks/generator.go
@@ -13,10 +13,10 @@ import (
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/datastore"
 	"github.com/spacemeshos/go-spacemesh/hare"
-	heligibility "github.com/spacemeshos/go-spacemesh/hare/eligibility"
+	"github.com/spacemeshos/go-spacemesh/hare/eligibility"
 	"github.com/spacemeshos/go-spacemesh/log"
 	"github.com/spacemeshos/go-spacemesh/sql/layers"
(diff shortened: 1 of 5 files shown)


blocks/generator.go
```diff
@@ -167,6 +167,12 @@ func (g *Generator) run() error {
 			if len(g.optimisticOutput) > 0 {
 				g.processOptimisticLayers(maxLayer)
 			}
+			if err := proposals.Delete(g.cdb, out.Layer); err != nil {
```
dshulyak, line 170: what about old proposals and transactions link? will it conflict with sync if someone will request proposals a bit later?

i thought that it may be better to prune old proposals & certs every 30m or so in separate goroutine, and run  autovacuum there as well but every day instead

countvonzero: i was thinking this can be shipped ASAP, as proposals is the major pain point and the db will stop growing at least.

later i will do something like you suggested and with vaccum because the timing to prune certificates and proposal_transactions are trickier.
one depends on hdist, and the other depends on mempool txs.

proposal syncing is only done during hare rounds. there it only prune up to layer N-1 at the end of hare for N.
so it's not likely that proposal sync will happen for old layers.

countvonzero: btw, mempool never cross-referenced with the proposal tables so it's ok that proposals are deleted.

dshulyak (approved):

countvonzero:
merging in case we want to make a release this week.

bors merge

PR #4996, opened 2023-09-11
[Merged by Bors] - Add new PoET config "poet-request-timeout" and use it instead of GracePeriod for timeouts

fasmat:
## Motivation
The NiPoST builder uses `GracePeriod` for two different (unrelated) things: one is a duration relative to a PoET round start. A node waits until then before it starts creating and submitting a challenge to PoET (so that it uses the highest possible ATX it sees until then).

The other use is as a request timeout for requests to PoET. For this a new config parameter "poet-request-timeout" is added to the config. This allows to set much shorter timeouts (1 hour on mainnet is much too long).

## Changes
- add new config parameter "poet-request-timeout"
- the defaults are calculated from the maximum time a request can take based on the linear jitter backoff

## Test Plan
existing tests pass

## TODO

### DIFF
--- a/activation/nipost.go
+++ b/activation/nipost.go
@@ -322,15 +322,15 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
 }
 
 // Submit the challenge to a single PoET.
-func (nb *NIPostBuilder) submitPoetChallenge(ctx context.Context, poet PoetProvingServiceClient, prefix, challenge []byte, signature types.EdSignature, nodeID types.NodeID) (*types.PoetRequest, error) {
-	poetServiceID, err := poet.PoetServiceID(ctx)
+func (nb *NIPostBuilder) submitPoetChallenge(ctx context.Context, client PoetProvingServiceClient, prefix, challenge []byte, signature types.EdSignature, nodeID types.NodeID) (*types.PoetRequest, error) {
+	poetServiceID, err := client.PoetServiceID(ctx)
 	if err != nil {
 		return nil, &PoetSvcUnstableError{msg: "failed to get PoET service ID", source: err}
 	}
 	logger := nb.log.WithContext(ctx).WithFields(log.String("poet_id", hex.EncodeToString(poetServiceID.ServiceID)))
 
 	logger.Debug("querying for poet pow parameters")
-	powParams, err := poet.PowParams(ctx)
+	powParams, err := client.PowParams(ctx)
 	if err != nil {
 		return nil, &PoetSvcUnstableError{msg: "failed to get PoW params", source: err}
 	}
@@ -345,7 +345,9 @@ func (nb *NIPostBuilder) submitPoetChallenge(ctx context.Context, poet PoetProvi
 
 	logger.Debug("submitting challenge to poet proving service")
 
-	round, err := poet.Submit(ctx, prefix, challenge, signature, nodeID, PoetPoW{
+	ctx, cancel := context.WithTimeout(ctx, nb.poetCfg.RequestTimeout)
+	defer cancel()
+	round, err := client.Submit(ctx, prefix, challenge, signature, nodeID, PoetPoW{
 		Nonce:  nonce,
 		Params: *powParams,
 	})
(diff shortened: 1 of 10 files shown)


fasmat:
bors try

fasmat:
bors try

config/mainnet.go
```diff
@@ -108,7 +108,8 @@ func MainnetConfig() Config {
 			PhaseShift:        240 * time.Hour,
 			CycleGap:          12 * time.Hour,
 			GracePeriod:       1 * time.Hour,
-			RequestRetryDelay: 10 * time.Second,
+			RequestTimeout:    550 * time.Second, // RequestRetryDelay * 2 * MaxRequestRetries*(MaxRequestRetries+1)/2
+			RequestRetryDelay: 5 * time.Second,
```
poszu: Why change the delay on mainnet?

activation/nipost.go
```diff
@@ -322,15 +322,15 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
+func (nb *NIPostBuilder) submitPoetChallenge(ctx context.Context, client PoetProvingServiceClient, prefix, challenge []byte, signature types.EdSignature, nodeID types.NodeID) (*types.PoetRequest, error) {
+	poetServiceID, err := client.PoetServiceID(ctx)
 	if err != nil {
 		return nil, &PoetSvcUnstableError{msg: "failed to get PoET service ID", source: err}
 	}
 	logger := nb.log.WithContext(ctx).WithFields(log.String("poet_id", hex.EncodeToString(poetServiceID.ServiceID)))
 
 	logger.Debug("querying for poet pow parameters")
-	powParams, err := poet.PowParams(ctx)
+	powParams, err := client.PowParams(ctx)
```
poszu: Why not set a timeout here as well? Perhaps it's best to set timeout for entire `submitPoetChallenge()` instead?

Btw, there is a timeout already set for `submitPoetChallenges()` - till the start of the round, which seems reasonable. Are timeouts per poet client necessary?

config/mainnet.go
fasmat: with the given number of max retries it can take up to 18+ Minutes to before the node gives up on fetching from PoET:

10 * 2 * (10*11)/2 = 1100 seconds or 18 Minutes 20 seconds

That felt like overkill, instead of reducing the amount of retries I opted for reducing the delay between the retries. Wdyt?

activation/nipost.go
fasmat: It makes sense to also add a timeout for `client.PowParams`, but I wouldn't add one for `submitPoetChallenges`. Reason being that the function does multiple requests to the same poet and creates a PoW. I'd rather have timeouts for the network requests and the rest be controlled by timings as before.

> Btw, there is a timeout already set for `submitPoetChallenges()` - till the start of the round, which seems reasonable. Are timeouts per poet client necessary?

Yes, the timeout for `submitPoetChallenges()` is the round start - which in my opinion is the correct timeout (abort submission when the round starts). The problem is that `submitPoetChallenges()` registers at all PoET with a Waitgroup. So if one PoET hangs, the Waitgroup blocks possibly until the round start after which the node aborts.

The idea is that one misbehaving PoET doesn't cause the node to miss their window for registration.

poszu: > So if one PoET hangs, the Waitgroup blocks possibly until the round start after which the node aborts.

It shouldn't abort

fasmat: OK, that's true but it will still hang until the ctx is canceled (at PoET round start).

poszu: Yes. 🤔 and this is bad because if other submits failed too it won't have time to retry...

config/mainnet.go
poszu: Yeah, it was somewhere between 550 and 1100s. It seems a lot but on the other hand, it also gives us _some_ time to fix poets in case of some hard failure in critical moment.

fasmat: OK I'll increase it to 1100 minutes 🙂

poszu (approved):

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

PR #4997, opened 2023-09-11
[Merged by Bors] - use first applied block in epoch for new miner active set

### LINKED ISSUE #4912: miner.ActiveSetFromEpochFirstBlock will use different block depending on synced certificate
if node wasn't online and synced certificates only for several last layers it will use different block for active set compared with 
nodes that were online for all time and have the earliest certificates.

i think that function should ignore certificates and use any applied block regardless of the reason

countvonzero:
## Motivation
Closes #4912 

## Changes
- for new miner use the first applied block for ref ballot active set


miner/util.go
```diff
@@ -10,14 +10,30 @@ import (
...
+func ActiveSetFromEpochFirstCertifiedBlock(db sql.Executor, epoch types.EpochID) ([]types.ATXID, error) {
 	bid, err := certificates.FirstInEpoch(db, epoch)
 	if err != nil {
 		return nil, err
 	}
+	return activeSetFromBlock(db, bid)
+}
+
+func ActiveSetFromEpochFirstBlock(db sql.Executor, epoch types.EpochID) ([]types.ATXID, error) {
+	result, err := ActiveSetFromEpochFirstCertifiedBlock(db, epoch)
```
dshulyak: this would be a different  block for synced node, i think it should be always first applied regardless of certificate

dshulyak (approved):

dshulyak:
i think it is this bug https://github.com/spacemeshos/go-spacemesh/issues/4912

countvonzero:
> i think it is this bug #4912

not aware of this. thanks!

countvonzero:
bors merge

countvonzero:
bors merge

Issue #4984, opened 2023-09-08
don't store activeset together with ballot object


dshulyak:
activeset is the main reason why database grows at such rate. 

- drop activeset from ballots blob. don't send it over sync either. sync will request activeset with another rpc
- across application request activeset using activesets.Get function, without an sql join to ballot table

dshulyak:
i looked into this briefly. the solution i think is to:

- run migration that decodes ballots, saves unique active sets, and updates ballot without active set but with a reference to activesets table
- vacuum && checkpoint
- all new ballots store active set only in activeset table
- allow to download active set for all blocks, not only latest (this restriction is harmful https://github.com/spacemeshos/go-spacemesh/blob/develop/proposals/handler.go#L469-L471)
- for code that needs activeset, get it by join on activeset table

countvonzero:
> for code that needs activeset, get it by join on activeset table

this requires adding activesetHash to ballots table.

> - run migration that decodes ballots, saves unique active sets, and updates ballot without active set but with a reference to activesets table
> - vacuum && checkpoint

i find it risky to do all these in one step/software upgrade, especially this is not a down-gradable release.

assuming the network has 1.1.5 now.

what do you think if we first release 
1.1.6 with
- a migration that 
  - delete propsals/certs/proposal txs
  - goes through ballots table and extract active set to the activesets table
- regular background pruning
- vacuum && checkpoint (for the proposal deleted)
- all new ballots store active set only in activeset table
- allow to download active set for all ballots 
- make sure no codepath access `ballot.ActiveSet` directly. 

at this point, there is no data loss and app only access `ballot.EpochData.ActiveSetHash`. 1.1.5 and 1.1.6 are compatible

then we release 1.2.0 with
- a migration that prune all the active sets from ballots
- vacuum && checkpoint

1.2.0  is compatible with 1.1.6 but not with 1.1.5, and is down-gradable. and we 
- get higher confidence that the data loss is ok because 1.1.6 is working
- provide a bigger incentive to upgrade 1.2.0 due to the much reduced state size.

dshulyak:
> this requires adding activesetHash to ballots table

this is one of the options. the other one is to fetch activeset with another query. would work as well

we can do it the way you suggested. the downside that 8GB will be allocated for 1-2 additional weeks. not a big deal

PR #4998, opened 2023-09-12
[Merged by Bors] - prune data and compact state

### LINKED ISSUE #3049: prune proposals from database
## Description

proposals are ephemeral. once a unified content block is generated, they serve no practical purpose.

### LINKED ISSUE #3588: certificate: prune from network after hdist layer
## Description

depends on #2921

when a new node is syncing in the network, it still needs to sync hare certificate for every layer since genesis, because the last layer in tortoise is the last synced layer. 

we want to be able to prune certificate after for all layers before current - hdist when a node is syncing.

countvonzero:
## Motivation
Closes #3049 
Closes #3588 

## Changes
- prune data consistently with tortoise hdist distance. 
  - proposals
  - certificates
  - proposal id<->tid mapping
  
  certificate is needed for consensus within hdist.
  the same distance is used for all for simplicity.

- extract active set data from ballots and save them in activesets table.
- vacuum and checkpoint database after migration 4 is complete

this PR concludes the first update described https://github.com/spacemeshos/go-spacemesh/issues/4984#issuecomment-1720169998


sql/vacuum/vacuum.go
```diff
@@ -0,0 +1,29 @@
...
+	if _, err := db.Exec("vacuum", nil, nil); err != nil {
+		return fmt.Errorf("vacuum %w", err)
+	}
+	if _, err := db.Exec("pragma wal_checkpoint(TRUNCATE)", nil, nil); err != nil {
+		return fmt.Errorf("wal checkpoint %w", err)
+	}
+	return nil
+}
+
+func FreePct(db sql.Executor) (int, error) {
```
dshulyak: could you please use full word? there is no reason to decipher this

mesh/janitor.go
```diff
@@ -0,0 +1,57 @@
...
+				)
+			}
+			logger.Info("certificates pruned", zap.Stringer("lid", oldest))
+			if err := transactions.DeleteProposalTxs(db, oldest); err != nil {
+				logger.Error("failed to delete proposal tx mapping",
+					zap.Stringer("lid", oldest),
+					zap.Error(err),
+				)
+			}
+			logger.Info("proposal tx pruned", zap.Stringer("lid", oldest))
```
dshulyak: i think this is not useful as info

mesh/janitor.go
```diff
@@ -0,0 +1,57 @@
...
+		zap.Uint32("dist", safeDist),
+		zap.Duration("interval", interval),
+	)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(interval):
+			oldest := lc.CurrentLayer() - types.LayerID(safeDist)
+			if err := proposals.Delete(db, oldest); err != nil {
```
dshulyak: one thing to consider is that database is locked for the duration of the delete. so all other writes will be delayed. for example if delete takes 1s, all latency for handling ballot writes will be increased.

we should add histogram for total duration, and maybe split writes further, by deleting layer by layer

this is very important not to screw up ballots/proposals propagation

sql/vacuum/vacuum.go
```diff
@@ -0,0 +1,29 @@
+package vacuum
+
+import (
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/sql"
+)
+
+func VacuumDB(db sql.Executor) error {
```
dshulyak: naming it Vacuum will be sufficient

cmd/root.go
```diff
@@ -78,6 +78,10 @@ func AddCommands(cmd *cobra.Command) {
 		cfg.DatabaseConnections, "configure number of active connections to enable parallel read requests")
 	cmd.PersistentFlags().BoolVar(&cfg.DatabaseLatencyMetering, "db-latency-metering",
 		cfg.DatabaseLatencyMetering, "if enabled collect latency histogram for every database query")
+	cmd.PersistentFlags().IntVar(&cfg.DatabaseCompactStatePct, "db-compact-state-pct",
+		cfg.DatabaseCompactStatePct, "configure the threshold to trigger state compaction on startup")
+	cmd.PersistentFlags().DurationVar(&cfg.DatabasePruneInterval, "db-prune-interval",
+		cfg.DatabasePruneInterval, "configure interval for database pruning")
```
dshulyak, line 82: i am not sure if we need flags for everything. for example db-compact-state-pct is arcane parameter, that likely will be never changed

node/node.go
```diff
@@ -169,6 +170,12 @@ func GetCommand() *cobra.Command {
 					return err
 				}
 
+				if app.Config.DatabaseCompactStatePct > 0 {
```
dshulyak: i realized that running vacuum will be too destructive if database size is even 20GB

we should either default to zero, or add an option that always runs compaction on startup and exit. so it is basically used as a tool

also i think there is no pressure to do it automatically, if code cleans up after itself continuously. what i think we should do is to execute migration that:
- drops proposals and txs
- nulls certificates
- update ballot to have nil active set
- run vacuum and checkpoint

once this is done, ephemeral state will not grow too much and will be reused for new data after old is cleaned up, so vacuum should be unnecessary

cmd/root.go
countvonzero: removed

sql/migrations/0004_next.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM proposals WHERE layer < 19000;
+DELETE FROM proposal_transactions WHERE layer < 19000;
+UPDATE certificates SET cert = NULL WHERE layer < 19000;
+UPDATE ballots SET ballot = prune_actives(ballot);
```
countvonzero: considering #5002 and the fact that right now the ref ballot is still saved with activeset. 
`ActiveSetFromEpochFirstBlock` also still retrieve activeset directly from ref ballot.
is there a safe range i should apply to remove the activeset from ballots?

at layer 17800, the state size changed from 8.2GB to 470MB with this migration. the migration took 1.5 min.

sql/vacuum/vacuum.go
countvonzero: deleted.

node/node.go
countvonzero: removed the automatic part.

countvonzero: please see https://github.com/spacemeshos/go-spacemesh/pull/4998#discussion_r1325486621

sql/database.go
```diff
@@ -128,13 +139,11 @@ func Open(uri string, opts ...Opt) (*Database, error) {
 		if err != nil {
 			return nil, err
 		}
-	}
-	for i := 0; i < config.connections; i++ {
-		conn := pool.Get(context.Background())
-		if err := registerFunctions(conn); err != nil {
-			return nil, err
+		if before == 3 {
+			if err := Vacuum(db); err != nil {
```
countvonzero, line 153: vaccum needs to happen outside of the db transaction, so the vaccum/checkpoint cannot be part of the migration.

sql/migrations/0004_next.sql
dshulyak: my proposal was to save active sets into activeset table while pruning. and then we either have an option to sql-join them on activesethash field (requires extending ballots table so maybe not optimal) or do another query to get activeset hash when we need it.

sql/migrations/0004_next.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM proposals WHERE layer < 19000;
+DELETE FROM proposal_transactions WHERE layer < 19000;
```
dshulyak: i think this we can delete regardless of layer

sql/migrations/0004_next.sql
dshulyak: > https://github.com/spacemeshos/go-spacemesh/issues/4984#issuecomment-1717409692

we should also stop writing active sets encoded together with ballot, after this migration is done. not necessarily within same change, but would be better to release them together

countvonzero: changed to extract active set from ballots and save the prune blob.

please consider https://github.com/spacemeshos/go-spacemesh/issues/4984#issuecomment-1720169998
i'll continue with the other code change accordingly.

mesh/janitor.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/spacemeshos/go-spacemesh/mesh/metrics"
+	"github.com/spacemeshos/go-spacemesh/sql"
+	"github.com/spacemeshos/go-spacemesh/sql/activesets"
+	"github.com/spacemeshos/go-spacemesh/sql/ballots"
+	"github.com/spacemeshos/go-spacemesh/sql/certificates"
+	"github.com/spacemeshos/go-spacemesh/sql/proposals"
+	"github.com/spacemeshos/go-spacemesh/sql/transactions"
+)
+
+func Prune(
```
dshulyak: why is it in mesh module? lets please create another one, which is called pruner (or janitor, though pruner is more specific). maybe it makes sense to be in sql/pruner

mesh/janitor.go
```diff
@@ -0,0 +1,112 @@
...
+					zap.Stringer("lid", oldest),
+					zap.Error(err),
+				)
+			}
+			metrics.PrunePropTxLatency.Observe(time.Since(t2).Seconds())
+		}
+	}
+}
+
+func ExtractActiveSet(db sql.Executor) error {
```
dshulyak: this is only used for migrations? can we put it somewhere that is not in the consensus code, maybe ballots/util/extract.go

dshulyak (approved):
please see notes about pruner/extract code

mesh/janitor.go
countvonzero: moved to pruner module.

mesh/janitor.go
countvonzero: put it in sql/ballots/util/extract.go.

countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #5002, opened 2023-09-13
[Merged by Bors] - tortoise: don't use active set to compute number of expected ballots

### LINKED ISSUE #4841: ballots: remove GetLegacyNumEligible in the next epoch

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4841

with this change active set is used only to verify eligibilities in the current epoch. 
tortoise doesn't need to load activeset into the state, which is reflected in the accepted type


countvonzero (approved):
if you want, you can also just delete the hare active set grading all together and remove legacy layer.

dshulyak:
> if you want, you can also just delete the hare active set grading all together and remove legacy layer.

probably not in this change, but will do if i will be touching that code

dshulyak:
bors merge

PR #5008, opened 2023-09-14
[Merged by Bors] - hare: use epoch first applied block for hare active set

countvonzero:
## Motivation
follow up on #4997 doing the same for hare active set


dshulyak (approved):

countvonzero:
bors merge

Issue #3006, opened 2021-12-16
tortoise: tortoise state loading will be growing in time 


dshulyak:
After persistent state was removed tortoise loads state from meshdb and atxdb when the node is started. In the initial implementation i am also re-verifying everything, as it allowed to reuse code for rerun almost as is.

However after https://github.com/spacemeshos/go-spacemesh/pull/3000 is merged rerun in full mode will be rather slow. So instead of verifying when node is booted we can just load state.

the end goal is to make restart fast, we can achieve that:
- by eliminating bottlenecks in tortoise
- by snapshotting tortoise when node is stopped gracefully
- by loading state without verifications
- by offloading state loading into background, and blocking components that rely on tortoise

dshulyak:
Note: we don't need state in full.weight for verified layers, but we need to know which ballots are good, because the definition is recursive.

dshulyak:
Also need to compute real margins for layers between verified and processed, as we will use them for voting. So it is not completely possible to avoid vote counting. Will postpone this change.

dshulyak:
We can do this after mutable ballots will be implemented. They allow us to verify goodness of the ballot without recursively decoding all previous votes, instead we can just compare if ballot encoded opinion is identical to local opinion.

countvonzero:
minor issue i encountered is that the loading is not interrupt-able with sigterm as it doesn't check context cancel.

PR #5009, opened 2023-09-14
[Merged by Bors] - tortoise: reuse decoded votes

dshulyak:

it also reduced 300MB of used memory on my machine.

ballot signs opinion hash, and provides sidecar to decode that opinion hash. if node decoded one sidecar it can store it in memory
and reuse for all following ballots with the same opinion hash.


countvonzero (approved):

dshulyak:
bors merge

PR #5010, opened 2023-09-14
[Merged by Bors] - tortoise: remove several bottlenecks during recovery

dshulyak:

- tally votes once per window, or after all ballots were loaded
- load ballots without joining identities table. that information already available in tortoise
- set tortoise log level to warn on mainnet. it is not pleasant that info is printed every time node is restarted


countvonzero (approved):

dshulyak:
bors merge

PR #5012, opened 2023-09-14
[Merged by Bors] - tortoise: use slice instead of map to store layers

dshulyak:
includes: https://github.com/spacemeshos/go-spacemesh/pull/5009

this improved recovery by 20s on my computer, from 1m to 40s. will upload benchmarks from source before merge


tortoise/tortoise.go
```diff
@@ -103,11 +101,11 @@ func (t *turtle) evict(ctx context.Context) {
 			ballotsNumber.Dec()
 			delete(t.ballotRefs, ballot.id)
 		}
-		for range t.layers[lid].blocks {
+		for range t.layer(lid).blocks {
 			blocksNumber.Dec()
 		}
-		layersNumber.Dec()
```
countvonzero, line 109: why remove?

tortoise/tortoise.go
```diff
@@ -103,11 +101,11 @@ func (t *turtle) evict(ctx context.Context) {
 			ballotsNumber.Dec()
 			delete(t.ballotRefs, ballot.id)
 		}
-		for range t.layers[lid].blocks {
+		for range t.layer(lid).blocks {
 			blocksNumber.Dec()
 		}
```
countvonzero: blocksNumber and atxsNumber can do Sub() to avoid looping?
i guess realistically there is just 1 layer each time in steady state.

tortoise/state.go
```diff
@@ -511,3 +508,26 @@ func activeSetWeight(epoch *epochInfo, aset []types.ATXID) (uint64, error) {
...
+	}
+	last := offset + lth
+	for j := types.LayerID(0); j <= i-lth; j++ {
+		s.data = append(s.data, &layerInfo{lid: last + j + 1, opinions: map[types.Hash32]votes{}})
+	}
+	return s.get(offset, index)
+}
+
+func (s *layerSlice) pop() {
+	copy(s.data[:], s.data[1:])
```
countvonzero: how about container.List? remove(front()) without the copying?

tortoise/state.go
```diff
@@ -511,3 +508,26 @@ func activeSetWeight(epoch *epochInfo, aset []types.ATXID) (uint64, error) {
...
+	i := index - offset - 1
+	lth := types.LayerID(len(s.data))
+	if i < lth {
+		return s.data[i]
+	}
+	last := offset + lth
+	for j := types.LayerID(0); j <= i-lth; j++ {
+		s.data = append(s.data, &layerInfo{lid: last + j + 1, opinions: map[types.Hash32]votes{}})
+	}
+	return s.get(offset, index)
```
countvonzero: this would be the same as `s.data[i]` as well?

countvonzero (approved):

tortoise/state.go
```diff
@@ -511,3 +508,26 @@ func activeSetWeight(epoch *epochInfo, aset []types.ATXID) (uint64, error) {
...
+}
+
+func (s *layerSlice) get(offset, index types.LayerID) *layerInfo {
+	i := index - offset - 1
+	lth := types.LayerID(len(s.data))
+	if i < lth {
+		return s.data[i]
+	}
+	last := offset + lth
+	for j := types.LayerID(0); j <= i-lth; j++ {
```
countvonzero: i find it more readable to do
```
for lid := last+1; lid <= index; lid++ {
}
```

tortoise/state.go
dshulyak: linked list doesn't support constant cost o(1) reads

tortoise/state.go
```diff
@@ -511,3 +508,26 @@ func activeSetWeight(epoch *epochInfo, aset []types.ATXID) (uint64, error) {
...
+		s.data = append(s.data, &layerInfo{lid: last + j + 1, opinions: map[types.Hash32]votes{}})
+	}
+	return s.get(offset, index)
+}
+
+func (s *layerSlice) pop() {
+	copy(s.data[:], s.data[1:])
+	s.data[len(s.data)-1] = nil
+	s.data = s.data[:len(s.data)-1]
+}
```
fasmat: Would this work to? Avoids the `copy` at the expense that the new `s.data` has equal `len` and `cap` where as in the current code `cap == len + 1`.

```suggestion
func (s *layerSlice) pop() {
	s.data = s.data[1:]
}
```

fasmat (approved):

tortoise/tortoise.go
dshulyak: changed to Set in evict method

tortoise/tortoise.go
dshulyak: thanks changed

tortoise/state.go
dshulyak: thanks, i thought there will be memory leak, but apparently golang does something smart with memory, so that capacity stays bounded

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

PR #5013, opened 2023-09-14
[Merged by Bors] - in memory index with atx data for consensus layer

### LINKED ISSUE #4164: cache: reduce amount of data stored in atxs cache and store all such headers for current and next epoch

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4164

this change introduces index with minimal data about identities which is required for consensus. the goal is to have predictable latency on the codepath that are sensitive to it, such as proposals and even more hare messages. the number of issues with that will only grow when sql db size won't fit in page cache (so querying vrf nonce and atx data will hit the disk often).

cached data per identity:
- weight
- height and base height
- vrf nonce
- malicious boolean

the data is added to the cache either on startup or when new atx is received. nonce may be loaded from previous atx if it wasn't included in this one. when block from epoch X is applied, we will evict epochs starting from X-capacity. index is configured to fit same number of epochs that fit in the tortoise window, this enables to self-healing during this period.

internally we store data indexed by epoch for simple eviction, and each epoch has an index by atx.

memory estimate for 1_000_000 identities is ~140MB


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

datastore/store.go
```diff
@@ -31,6 +32,9 @@ type CachedDB struct {
 	*sql.Database
 	logger log.Log
 
+	// cache is optional
+	cache *cache.Cache
+
 	atxHdrCache   *lru.Cache[types.ATXID, *types.ActivationTxHeader]
 	vrfNonceCache *lru.Cache[VrfNonceKey, *types.VRFPostIndex]
```
countvonzero, line 39: why did you decide to keep these lru cache (atx/vrf/malfeasance)?
at a high level i am not sure how to articulate the purpose of these cache now. i think it's better to remove them.
vrf/malfeasance are pretty much for consensus only. i think atx also, with the exception of looking up miner's initial atx.

my thought is since cachedb is already inserted everywhere, why not just add the cache interface to it so other components don't need to differentiate the two (CacheDB and cache.Cache)

cache/warmup.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+	"github.com/spacemeshos/go-spacemesh/sql"
+	"github.com/spacemeshos/go-spacemesh/sql/atxs"
+	"github.com/spacemeshos/go-spacemesh/sql/identities"
+	"github.com/spacemeshos/go-spacemesh/sql/layers"
+)
+
+func Warm(db *sql.Database, opts ...Opt) (*Cache, error) {
```
countvonzero: i find the interface of this module confusing. New/Warm/Warmup are all public.

i think just Warm() alone would be sufficient. shouldn't matter if the db has data or not. the others should be private?

cache/cache.go
```diff
@@ -0,0 +1,186 @@
...
+	"bytes"
+	"sort"
+	"sync"
+
+	"go.uber.org/atomic"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+)
+
+type ATXData struct {
```
countvonzero: i think it make sense to use this everywhere and only use atx header in the atx gossip handler

cache/cache.go
```diff
@@ -0,0 +1,186 @@
...
+	return types.EpochID(c.evicted.Load())
+}
+
+func (c *Cache) IsEvicted(epoch types.EpochID) bool {
+	return c.evicted.Load() >= epoch.Uint32()
+}
+
+// OnApplied is a notification for cache to evict epochs that are not useful
+// to keep in memory.
+func (c *Cache) OnApplied(applied types.EpochID) {
```
countvonzero: not sure this is intuitive. the term "applied" seems irrelevant in the context of cache. how about OnNewEpoch and only called when the layer is the first layer of the epoch?

cache/cache.go
```diff
@@ -0,0 +1,186 @@
...
+	if applied < c.capacity {
+		return
+	}
+	evict := applied - c.capacity
+	if c.IsEvicted(evict) {
+		return
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.evicted.Load() < applied.Uint32() {
```
countvonzero: shouldn't this compare to `evict` instead of `applied`?

cache/cache.go
```diff
@@ -0,0 +1,186 @@
...
+
+	atxsCounter.WithLabelValues(epoch.String()).Add(1)
+	if atxs == nil {
+		identitiesCounter.WithLabelValues(epoch.String()).Add(1)
+	}
+
+	atxs = append(atxs, atx)
+	// NOTE(dshulyak) doesn't make sense, as we don't guarantee that every node
+	// will see same atxs. it actually should see atmost one and malfeasence proof if node equivocated.
+	// find out if we have use case in consensus.
```
countvonzero: i don't think we need this. the identity should already be set malicious when Cache.SetMalicious() is called.
and at that point it shouldn't matter which ATX is returned for a malicious identity.

cache/cache.go
```diff
@@ -0,0 +1,186 @@
...
+	"go.uber.org/atomic"
+
+	"github.com/spacemeshos/go-spacemesh/common/types"
+)
+
+type ATXData struct {
+	Weight             uint64
+	BaseHeight, Height uint64
+	Nonce              types.VRFPostIndex
+	Malicious          bool
```
countvonzero: i think nonce and malicious is a property of identity, not atx.

if an identity is marked malicious in epoch 3, it should still be malicious in epoch 6. but with current design if nodes didn't restart btwn these time it will be forgotten.

cache/cache.go
```diff
@@ -0,0 +1,186 @@
...
+	ecache.identities[node] = atxs
+}
+
+func (c *Cache) SetMalicious(node types.NodeID) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for _, ecache := range c.epochs {
+		for _, atx := range ecache.identities[node] {
+			data := ecache.atxs[atx]
+			// copy on update instead of copying on read
```
countvonzero: don't understand. if a reference is stored in the map, why not update it directly?

cache/warmup.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		cache.Add(vatx.TargetEpoch(), vatx.SmesherID, vatx.ID(), ToATXData(vatx.ToHeader(), nonce, malicious))
+		return true
+	}); err != nil {
+		return err
+	}
+	return ierr
+}
+
+func ToATXData(atx *types.ActivationTxHeader, nonce types.VRFPostIndex, malicious bool) *ATXData {
```
countvonzero: i don't understand this interface used outside the module.
the atx handler needs to query database to see if a node is malicious, and then add the data to cache.
but at the same time the cache also holds whether an node is malicious.

i am confused about the design. a component with access to both, where does it ask for an identity's maliciousness and nonce?

cache/cache.go
dshulyak: it doesn't matter whose property logically is it. it is about how data is stored and indexed.
identity is relevant in consensus only if it has atx in this epoch.

cache/cache.go
dshulyak: yes the plan is to use it everywhere. atx header can be removed. atx handler will use database to get atx

cache/cache.go
dshulyak: it will be a race with returned values

cache/warmup.go
dshulyak: i need Warm to not spend time fixing tests that are not well written. they init data in some obscure methods and then i just load this data with Warm

cache/warmup.go
dshulyak: consensus doesn't care about malicious identities that doesn't have atx. therefore we load data when atx is received or on startup for previously received atx. you can see how it is used in EligibilityValidator

hare will query cache using GetByNode, and check malicious bit. it is 100% correct, because we really don't care about identity by itself, but only if it has an atx

datastore/store.go
dshulyak: cachedb will be removed from codebase. i didn't refactor all code to use cache. after i am done with followups slowpath will use db directly and rely on kernel page cache, fastpath will use consensus cache

cache/cache.go
dshulyak: we need to store them to support NodeHasAtx. otherwise ballots won't be accepted. and for consistency it is better to keep them ordered

cache/cache.go
dshulyak: i changed it to OnEpoch. the idea was to be explicit when it suppose to be called. there is no reason to have something generic

it is also more robust just to get every notification

cache/cache.go
countvonzero: sorry still not clear. do you mean there will be a race if you do `ecache.atxs[atx].Malicious = true` directly?

cache/cache.go
countvonzero: there was a plan to only store the first equivocating atx/ballot and only sync those when they are referenced.
i see NodeHasAtx is used to check proposal eligibility. do you think this step is needed if the ballot is malicious?

cache/warmup.go
countvonzero: right. the overall plan wasn't really communicated in the PR description.

GetByNode is not used at the moment so i couldn't see its intended purpose.

this is the usage i got form EligibilityValidator
`atx = cache.ToATXData(owned.ToHeader(), nonce, false)`
and i don't know why the malicious bit is always false.

activation/handler.go
```diff
@@ -356,6 +360,22 @@ func (h *Handler) ContextuallyValidateAtx(atx *types.VerifiedActivationTx) error
 	return err
 }
 
+func (h *Handler) cacheAtx(ctx context.Context, atx *types.ActivationTxHeader) {
+	if !h.cache.IsEvicted(atx.TargetEpoch()) {
```
countvonzero: i still think it's cleaner to encapsulate the interaction with db in the cache itself.

the whole point of cache is to reduce contention on database and improve latency and i/o. cache and db should be considered as a single component. it's up to the consensus cache to determine whether it will use a database or not.

it's nicer for other parts of the codebase to just tell cache to handle this atx without accessing db separately.

cache/cache.go
countvonzero: > identity is relevant in consensus only if it has atx in this epoch.

this is theoretically true, but in practice we already have this in hare/tortoise where we use nodeid->malicious mapping to avoid checking nodeid->"atx in this epoch" lookup.

and cachedb's IsMalicious() is also by nodeid. how do you plan to replace that?

proposals/eligibility_validator.go
```diff
@@ -74,35 +68,44 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot,
...
-	nonce, err := v.nonceFetcher.VRFNonce(ballot.SmesherID, ballot.Layer.GetEpoch())
-	if err != nil {
-		return false, fmt.Errorf("no vrf nonce for %v in epoch %v: %w", ballot.SmesherID, ballot.Layer.GetEpoch(), err)
-	}
-	var data *types.EpochData
-	if ballot.EpochData != nil && ballot.Layer.GetEpoch() == v.clock.CurrentLayer().GetEpoch() {
-		var err error
-		data, err = v.validateReference(ballot, actives, owned)
+	var atx *cache.ATXData
+	if epoch := ballot.Layer.GetEpoch(); v.cache.IsEvicted(epoch) {
```
countvonzero: i find it weird that i have to care about whether this epoch is evicted from cache or not.
this check is also not atomic. if it's not evicted now, it might be evicted in the next microsecond.
also why i think it's better to have a db inside the cache, encapsulate the logic, and serve whatever the client is asking and decide for itself whether to cache in memory or not.

proposals/eligibility_validator.go
```diff
@@ -74,35 +68,44 @@ func (v *Validator) CheckEligibility(ctx context.Context, ballot *types.Ballot,
...
+		}
+		if ballot.SmesherID != owned.SmesherID {
+			return false, fmt.Errorf("atx and ballot key mismatch: public key (%v), ATX node key (%v)", ballot.SmesherID.String(), owned.NodeID)
+		}
+		nonce, err := atxs.VRFNonce(v.db, ballot.SmesherID, epoch)
 		if err != nil {
-			return false, err
+			return false, fmt.Errorf("no vrf nonce for %v in epoch %v: %w", ballot.SmesherID, ballot.Layer.GetEpoch(), err)
+		}
+		atx = cache.ToATXData(owned.ToHeader(), nonce, false)
```
countvonzero: why is the malicious bit always false here? there should be a lookup in db for that.

countvonzero (commented):
can you document in the issue what the next-step plan is?

as is, there are 3 caches after this PR
- tortoise cache
- lru cache
- consensus cache

this PR may improve performance but make overall design around cache more confusing.

also, what is the argument against opening up tortoise cache? for example:
- a cache with a db wrapped inside
- eviction policy is based on tortoise's window
tortoise window currently covers more than 2 epochs, which should be sufficient?
i'd really love to have just 1 cache for all

cache/cache.go
dshulyak: pointer is returned in response

cache/cache.go
dshulyak: why does it matter?

cache/cache.go
dshulyak: hare GetByNode, tortoise Get

activation/handler.go
dshulyak: i can rename it from cache to consensus_data

>the whole point of cache is to reduce contention on database and improve latency and i/o. cache and db should be considered as a single component. it's up to the consensus cache to determine whether it will use a database or not.

no this is not the idea. the idea is to store small amount of data that is needed for consensus that eliminates any database access

proposals/eligibility_validator.go
dshulyak: this is very unlikely event, in such unlikely event code will have to retry. locking cache (to make it atomic with eviction and db read) so that it can read from db is mistake

proposals/eligibility_validator.go
dshulyak: i will change this code to make it clear that bit is unused

dshulyak:
i want to drop lru cache completely, i think it wasn't necessary and made problems with access hidden (https://github.com/spacemeshos/go-spacemesh/issues/5039)

slowpath will not use cache at all and will rely on disk and page cache access. fastpath will use consensus data (fastpath is generally what is sensitive to latency and requires fast propagation or severly limits sync speed), things as proposal/ballot validation, hare validation.

i wanted to reuse ATXData in tortoise. in general i am very opposed to putting any non-obvious dependencies into tortoise, especially database. for example if we do what you suggest and make db access from cache and then hide it in tortoise, then every tortoise method call may issue a disk read, at the same time every call is also mutex protected, it is just a poor way to write code 

> a cache with a db wrapped inside

this specifically will make it very hard to reason about tortoise performance and will lead to mistakes that i outlined above. 

you raised that it will be nice to have passthrough cache, but i don't know what is nice about it. it makes it non-obvious what is the worst case performance of query, with cache it is obvious that the worst is single map hash request (or btree request) so it is ok to do it in the loop, with disk you can't loop. with memory map it is also ok to do it inside mutex protected area, with db it is not ok to hold another lock while you read. coding pattern that allows developer to see what is being used (disk or memory) is nice, coding pattern that obscures it is not nice.

cache/cache.go
dshulyak: map stores a pointer, and it returns it to avoid copy on every read. during update that pointer is not protected by mutex, so if we update that memory directly race will be noticed and reported. therefore we copy it, update, and reinsert into map that is protected by mutex

cache/cache.go
dshulyak: specifically hare will use only fastpath that relies on consensus data, it won't even need to have references to db for that.

tortoise (proposal/ballots handler) will query using Get(atx)

cache/cache.go
countvonzero: thanks for the explanation. i've made mistakes like this a few times.

countvonzero:
> i want to drop lru cache completely, i think it wasn't necessary and made problems with access hidden (#5039)
> 
> slowpath will not use cache at all and will rely on disk and page cache access. fastpath will use consensus data (fastpath is generally what is sensitive to latency and requires fast propagation or severly limits sync speed), things as proposal/ballot validation, hare validation.
> 
> i wanted to reuse ATXData in tortoise. in general i am very opposed to putting any non-obvious dependencies into tortoise, especially database. for example if we do what you suggest and make db access from cache and then hide it in tortoise, then every tortoise method call may issue a disk read, at the same time every call is also mutex protected, it is just a poor way to write code
> 
> > a cache with a db wrapped inside
> 
> this specifically will make it very hard to reason about tortoise performance and will lead to mistakes that i outlined above.
> 
> you raised that it will be nice to have passthrough cache, but i don't know what is nice about it. it makes it non-obvious what is the worst case performance of query, with cache it is obvious that the worst is single map hash request (or btree request) so it is ok to do it in the loop, with disk you can't loop. with memory map it is also ok to do it inside mutex protected area, with db it is not ok to hold another lock while you read. coding pattern that allows developer to see what is being used (disk or memory) is nice, coding pattern that obscures it is not nice.

agree. my viewpoint is mostly about the clarity of interface btwn components. like i said, IsEvicted isn't atomic and is a logic that should be hidden from its users. maybe there is a compromise possible? like passing a db as you did in Warm?

countvonzero (approved):

cache/cache.go
```diff
@@ -0,0 +1,193 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	ecache, exists := c.epochs[epoch]
+	if !exists {
+		ecache = epochCache{
+			index: btree.NewG(64, func(left, right *stored) bool {
+				nodecmp := bytes.Compare(left.node[:], right.node[:])
+				if nodecmp == 0 {
+					return bytes.Compare(left.atx[:], right.atx[:]) == -1
+				}
```
poszu: Is it possible? It would mean that a single nodeID has more than 1 ATX in the epoch. Is it to handle malicious smeshers?

cache/cache.go
```diff
@@ -0,0 +1,193 @@
...
+	capacity types.EpochID
+
+	mu     sync.RWMutex
+	epochs map[types.EpochID]epochCache
+}
+
+type stored struct {
+	node types.NodeID
+	atx  types.ATXID
+	data *ATXData
```
poszu: Why is `data` a pointer?

cache/cache.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		data = s.data
+		return false // get the first one
+	})
+	return data
+}
+
+// WeightForSet computes total weight of atxs in the set and returned array with
+// atxs in the set that weren't used.
+func (c *Cache) WeightForSet(epoch types.EpochID, set []types.ATXID) (uint64, []bool) {
```
poszu: It's worth noting that the `set` must be sorted (to use `slice.BinarySearch`). Btw, is it guaranteed that the value passed in here is actually sorted?
```suggestion
// WeightForSet computes total weight of atxs in the set and returned array with
// atxs in the set that weren't used.
// The set MUST be sorted.
func (c *Cache) WeightForSet(epoch types.EpochID, set []types.ATXID) (uint64, []bool) {
```

cache/cache.go
```diff
@@ -0,0 +1,193 @@
...
+	defer c.mu.Unlock()
+	for _, ecache := range c.epochs {
+		ecache.index.AscendGreaterOrEqual(&stored{node: node}, func(s *stored) bool {
+			if node != s.node {
+				return false
+			}
+			// TODO(dshulyak) how to copy on update here?
+			update := *s.data
+			update.Malicious = true
+			s.data = &update
```
poszu: What do you mean?

poszu (commented):
TBH I also fail to understand why it is better to split cache and DB into separate things. It complicates the usage significantly where a cache miss is followed by a read from DB. 

Is it because you sometimes want to avoid fetching the data from DB on a cache-miss?

I think this could be solved with something like:
```go
type ReadOptions struct {
    FetchOnMiss bool,
}

func (c *Cache) WeightForSet(epoch types.EpochID, set []types.ATXID, opts ReadOptions)  (uint64, []bool) {
    // try to find in cache
    // if not found and opts.FetchOnMiss == true -> fetch from DB
}
```

Then, DB and cache could be kept together and the interface would be simpler for the users. Wdyt?

dshulyak:
> TBH I also fail to understand why it is better to split cache and DB into separate things. It complicates the usage significantly where a cache miss is followed by a read from DB.
> Is it because you sometimes want to avoid fetching the data from DB on a cache-miss?

i am going to rename it to data/atxs. it is not about looking up db on cache miss. it is just about data. exactly the same way tortoise stores ballot data for efficient access, the same i need to enable fast hare and tortoise message handling.

i never want to read database on "cache" miss. the only time when i do so is for epochs in the past in case if tortoise window is larger then data stored in atxs. 

> Then, DB and cache could be kept together and the interface would be simpler for the users. Wdyt?

it is not cache. it is consensus data

cache/cache.go
dshulyak: i will remove TODO. i meant that i want to copy/update/store, because returned pointer is read concurrently with modification

cache/cache.go
dshulyak: yes this is enforced in the only place that accepts active sets

cache/cache.go
dshulyak: yes this is for malicious. i want to simplify index to only order by atx. but i can't without changing hare

cache/cache.go
dshulyak: thats a leftover. i was experimenting with storing `stored` as non-pointer, i thought that it will remove pointer chasing during btree traversal and reduce cpu cache misses. but i also suspect that memory in benchmark is allocated not in the best way to highlight problems

dshulyak:
will create a separate to follow to adjust those places in a different pr

## where it will be used?
	- no atx queries in ballot/proposal handlers (implemented in this change)
	- no identity and atx queries in hare
	- no atx queries in block generator
	- atx queries during mesh hash resolution
## will allow to
	- drop atxs data from tortoise
	- drop lru cache for atxs and identities

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

atxsdata/data.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	_, exists = d.malicious[data.Node]
+	data.Malicious = exists
+	return data
+}
+
+// WeightForSet computes total weight of atxs in the set and returned array with
+// atxs in the set that weren't used.
+//
+// set is expected to be sorted.
```
poszu: It doesn't seem to be required anymore, is it?
```suggestion
```

atxsdata/warmup.go
```diff
@@ -47,16 +47,16 @@ func Warmup(db sql.Executor, cache *Cache) error {
...
 		return true
 	}); err != nil {
 		return err
 	}
 	return ierr
 }
 
-func ToATXData(atx *types.ActivationTxHeader, nonce types.VRFPostIndex, malicious bool) *ATXData {
-	return &ATXData{
+func ToATXData(atx *types.ActivationTxHeader, nonce types.VRFPostIndex, malicious bool) *ATX {
```
poszu: It's not used anymore AFAIR.

atxsdata/data.go
```diff
@@ -0,0 +1,180 @@
+package atxsdata
+
+import (
+	"sync"
+
+	"go.uber.org/atomic"
```
poszu: What's the benefit of using uber's `atomic.Uint32` compared to `sync/atomic.Uint32`?
```suggestion
	"sync"
	"sync/atomic"
```

atxsdata/data.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	ecache, exists := d.epochs[epoch]
+	if !exists {
+		ecache = epochCache{
+			index: map[types.ATXID]*ATX{},
+		}
+		d.epochs[epoch] = ecache
+	}
```
poszu: It could re-add an evicted epoch if `d.OnEpoch(epoch+d.capacity)` is called between the `d.IsEvicted(epoch)` check and the lock. Why not bring the check under the lock?
The re-added epoch would probably stay in the `d.epochs` map forever.

```suggestion
	d.mu.Lock()
	defer d.mu.Unlock()
	if d.IsEvicted(epoch) {
		return
	}
	ecache, exists := d.epochs[epoch]
	if !exists {
		ecache = epochCache{
			index: map[types.ATXID]*ATX{},
		}
		d.epochs[epoch] = ecache
	}
```

dshulyak: i wanted to be "smart" and make OnEpoch non-blocking (basically noop if it has nothing to evict) so that it doesn't interfere with read locks, but it is generally rare and won't make substantial difference.

it won't store it forever but for one more epoch, because of this

```go
for epoch := range d.epochs {
		if epoch <= evict {
			delete(d.epochs, epoch)
			atxsCounter.DeleteLabelValues(epoch.String())
		}
	}
```

atxsdata/warmup.go
dshulyak: thanks

atxsdata/data.go
dshulyak: i missed when stdlib added atomic.Uint32. i remember it only had functions to work with *uint32

atxsdata/data.go
dshulyak: btw i moved everything under the lock

poszu (approved):

dshulyak:
bors merge

PR #5020, opened 2023-09-15
[Merged by Bors] - refuse old proposals from gossip

countvonzero:
## Motivation
there are a lot of old proposals floating in the gossip network.
deleting old proposals in #4993 forced the node to process/save these old proposals repeatedly

## Changes
- refuse proposals older than current layer
- add metrics/log for time it takes to delete proposals

### DIFF
--- a/proposals/handler.go
+++ b/proposals/handler.go
@@ -262,7 +262,7 @@ func (h *Handler) handleProposal(ctx context.Context, expHash types.Hash32, peer
 		preGenesis.Inc()
 		return fmt.Errorf("proposal before effective genesis: layer %v", p.Layer)
 	}
-	if p.Layer < h.clock.CurrentLayer() {
+	if p.Layer <= h.mesh.ProcessedLayer() {
 		// old proposals have no use for the node
 		tooLate.Inc()
 		return fmt.Errorf("proposal too late: layer %v", p.Layer)
--- a/proposals/handler.go
+++ b/proposals/handler.go
@@ -262,6 +262,11 @@ func (h *Handler) handleProposal(ctx context.Context, expHash types.Hash32, peer
 		preGenesis.Inc()
 		return fmt.Errorf("proposal before effective genesis: layer %v", p.Layer)
 	}
+	if p.Layer < h.clock.CurrentLayer() {
+		// old proposals have no use for the node
+		tooLate.Inc()
+		return fmt.Errorf("proposal too late: layer %v", p.Layer)
+	}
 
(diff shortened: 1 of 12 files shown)


proposals/handler.go
```diff
@@ -262,6 +262,11 @@ func (h *Handler) handleProposal(ctx context.Context, expHash types.Hash32, peer
 		preGenesis.Inc()
 		return fmt.Errorf("proposal before effective genesis: layer %v", p.Layer)
 	}
+	if p.Layer < h.clock.CurrentLayer() {
+		// old proposals have no use for the node
```
countvonzero, line 266: in this case the enclosed ballot should be fetched via sync, if not already.

countvonzero: another option is to save the ballot only

dshulyak: what do you think about maintaining variable that is updated by block generator after it is done with this layer?

defaults to current layer on startup, but it will allow to stop accepting proposals earlier

dshulyak (approved):

dshulyak:
btw i didn't notice that all old proposals are deleted during protocol runtime https://github.com/spacemeshos/go-spacemesh/blob/develop/sql/proposals/proposals.go#L190 . the function looks like that it is limited for the layer

this was visible in write metrics. so it was somewhat risky

image

proposals/handler.go
countvonzero: it's a good idea. i chose to use `mesh.ProcessedLayer()` to be the threshold for the following reasons
- it's updated every time a block is generated, optimistically or not
- hare can terminate out of order, so block can be generated out of order as well in the non-optimistic case
- proposal handler already have mesh interface ready to use

countvonzero:
> btw i didn't notice that all old proposals are deleted during protocol runtime https://github.com/spacemeshos/go-spacemesh/blob/develop/sql/proposals/proposals.go#L190 . the function looks like that it is limited for the layer
> 
> this was visible in write metrics. so it was somewhat risky

yes. this was risky. i made the following change
- delete batch at startup (anything before the layer in state
- make the delete async
- renamed Delete to DeleteBefore
note that the last layer is kept just in case peers want to sync proposals for latest layer due to clock skew.

countvonzero:
bors try

dshulyak (approved):

countvonzero:
bors merge

countvonzero:
bors merge

countvonzero:
bors merge

PR #5022, opened 2023-09-16
Add nats and extra info to coinbaseReward

brunovale91:
## Motivation
Want to be able to index the node events in a reliable way. NATS jetstream provides durability to events, this means a view database can be indexed without having to run the full node from a clean state.

## Changes
New NATS package and NATS configuration added. Opt in.
Atx id and node id added to CoinbaseReward and reward event.

## Test Plan
It was manually tested on a mainet node and events are produced properly to NATS.

## TODO
- Tweak and add more NATS configurations. (Currently the default configuration for the stream it to keep event for one year).

### DIFF
--- a/common/types/block.go
+++ b/common/types/block.go
@@ -109,6 +109,8 @@ type AnyReward struct {
 // CoinbaseReward contains the reward information by coinbase, used as an interface to VM.
 type CoinbaseReward struct {
 	Coinbase Address
+	AtxID    ATXID
+	NodeID   NodeID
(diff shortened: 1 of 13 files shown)


dshulyak:
sorry, please open an issue for discussion. prs with new dependencies and non-trivial modifications to the client are very unlikely to be accepted without prior discussion.

wrt the problem of indexing, yes we know that API for it needs to be improved, however there is a lack of hands currently. but adding one more api (nats) is not a solution.

PR #5024, opened 2023-09-17
[Merged by Bors] - remove read/write for ballot.ActiveSet

### LINKED ISSUE #4984: don't store activeset together with ballot object

countvonzero:
## Motivation
Closes #4984

## Changes
- drop EmitEmptyActiveSet config
- remove all access to Ballot.ActiveSet and look up from db instead
- remove all writes of Ballot.ActiveSet

after this PR, miner will still accept proposal/ballot with non-empty active sets, but will save ballots without active set.

### DIFF
--- a/proposals/eligibility_validator.go
+++ b/proposals/eligibility_validator.go
@@ -10,7 +10,6 @@ import (
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/datastore"
 	"github.com/spacemeshos/go-spacemesh/log"
-	"github.com/spacemeshos/go-spacemesh/sql/activesets"
 	"github.com/spacemeshos/go-spacemesh/system"
 )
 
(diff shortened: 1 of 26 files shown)


countvonzero:
bors try

countvonzero:
bors try

countvonzero:
bors try

proposals/eligibility_validator.go
```diff
@@ -140,11 +141,15 @@ func (v *Validator) validateReference(ballot *types.Ballot, owned *types.Activat
 	if ballot.EpochData.Beacon == types.EmptyBeacon {
 		return nil, fmt.Errorf("%w: ref ballot %v", errMissingBeacon, ballot.ID())
 	}
-	if len(ballot.ActiveSet) == 0 {
+	activeSet, err := activesets.Get(v.cdb, ballot.EpochData.ActiveSetHash)
```
dshulyak: there should be a way to avoid this read. we should have loaded this data earlier, as we need to check data availability. and then we need to pass it together with the rest of the ballot for eligibility validation

dshulyak (approved):

proposals/eligibility_validator.go
countvonzero: thanks. changed to pass it from the original lookup.

countvonzero:
bors merge

countvonzero:
bors cancel

countvonzero:
bors merge

PR #5027, opened 2023-09-18
Split systests from go-spacemesh node

fasmat:
## Motivation
Separates system tests from go-spacemesh node code.

## Changes
- `systests` is now its own module that depends on `github.com/spacemeshos/go-spacemesh`
- the dependency to the node is replaced to always point at the node code of the same commit (otherwise we would need to keep `go.mod` in `systests` updated with every commit to the repository)

## Test Plan
existing tests pass

## TODO


fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

countvonzero:
what is the advantage of splitting?

fasmat:
Mostly a smaller list of dependencies for the node code and clearer separation of what are system tests and what is node code.

We could use libraries in system tests that don't support windows and mac is probably the biggest advantage (although not needed at the moment).

fasmat:
I found the PR that originally merged the two modules into one:
 https://github.com/spacemeshos/go-spacemesh/pull/3260

It looks like the reason they were merged was to be able to use go-spacemesh code in system tests (without constantly needing to update a dependency). This can also be achieved with a `replace` statement in systests `go.mod` file.

countvonzero:
using replace is suboptimal. to me this is a change of workflow without clear benefit....

countvonzero:
so my workflow is that i'll build a image for go-sm A and a image for systest A
but as i progress with development, go-sm image will keeping changing (B, C, D...) but will continue to use systest A.

fasmat:
> so my workflow is that i'll build a image for go-sm A and a image for systest A but as i progress with development, go-sm image will keeping changing (B, C, D...) but will continue to use systest A.

While I think that this should also be possible with splitting the codebase into two modules, I'll close the issue for now. There seem to be no immediate benefits to make the change.

poszu:
@countvonzero 
Why is using replace suboptimal? It doesn't require any action on developer side.

@fasmat Wouldn't go workspace help here?

fasmat:
Go workspaces would help for local development, such that the system tests are always compiled with the local version of go-spacemesh rather than the version specified in `go.mod` - they essentially auto-add `replace` between modules referenced in the workspace automagically.

`go.work` and `go.work.sum` however in my opinion do not belong into the repository, since they are only intended for local workspaces. I think that a single `replace` here is correct. This should not break the workflow @countvonzero is describing: building the system tests once and rebuilding the node code multiple times after.

`replace` guarantees that `systest` is always built with `go-spacemesh` of the same commit (same as it is now). Additionally it would prevent using `systest` code outside the `systest` module (would introduce a cyclic dependency) which I think would be a small benefit of splitting the two.

PR #5032, opened 2023-09-19
[Merged by Bors] - prune active set from ballot blob

### LINKED ISSUE #4985: compress sqlite state after pruning
we should either:
- run vacuum periodically when app is started https://www.sqlite.org/lang_vacuum.html
- enable auto_vacuum  https://www.sqlite.org/pragma.html#pragma_auto_vacuum . autovacuum doesn't work for old dbs and requires tuning to understand performance

after:
- https://github.com/spacemeshos/go-spacemesh/issues/3588
- https://github.com/spacemeshos/go-spacemesh/issues/3049
- https://github.com/spacemeshos/go-spacemesh/issues/4984

countvonzero:
## Motivation
Closes #4985

## Changes
- actually prune active set from ballot blobs in db
- use a empty migration to facilitate the pruning. 
  we only want to prune/vacuum/checkpoint once per software upgrade
  the code allow any upgrade from migration <=4 to 5 to
  - extract active set from ballots
  - prune active set from blob
  - vacuum and checkpoint

the final size of the state should be < 1GB


dshulyak (approved):
migrations need to be adjusted, as i had to add one

countvonzero:
bors merge

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ In order to enable provide following configuration:
   Ephemeral data are deleted and state compacted at the time of upgrade. In steady-state, data is pruned periodically.
 * [#5021](https://github.com/spacemeshos/go-spacemesh/pull/5021) Drop support for old certificate sync protocol.
 * [#5024](https://github.com/spacemeshos/go-spacemesh/pull/5024) Active set will be saved in state separately from ballots.
+* [#5032](https://github.com/spacemeshos/go-spacemesh/pull/5032) Ativeset data pruned from ballots.
```
dshulyak, line 37: we should put a bit more data about this. e.g expected time and amount of storage required to perform this migration (e.g. WAL is the size of database right?) . we can do that after benchmarking in cloud

countvonzero:
bors cancel

CHANGELOG.md
countvonzero: yes. WAL will be the size of the pre-vacuumed db (it's being used as a tmp storage during vacuum).
will update after testing.

countvonzero:
bors merge

PR #5038, opened 2023-09-19
[Merged by Bors] - add index for faster query to get refballot

dshulyak:
there was a regression in proposal builder that started to use codepath that was used in different domain, which was executed once per epoch. now this is executed on every proposal creation and might be too slow.


pigmej (approved):

dshulyak:
bors merge

dshulyak:
bors merge

Issue #5036, opened 2023-09-19
sync should not turn off consensus


lrettig:
## Description
I'm running two nodes (both v1.1.6), both see exactly the same layer data, but one thinks it's synced and one doesn't:

```
> grpcurl -plaintext localhost:9092 spacemesh.v1.NodeService.Status
{
  "status": {
    "connectedPeers": "44",
    "syncedLayer": {
      "number": 19386
    },
    "topLayer": {
      "number": 19395
    },
    "verifiedLayer": {
      "number": 19385
    }
  }
}
> grpcurl -plaintext localhost:9192 spacemesh.v1.NodeService.Status
{
  "status": {
    "connectedPeers": "1",
    "isSynced": true,
    "syncedLayer": {
      "number": 19386
    },
    "topLayer": {
      "number": 19395
    },
    "verifiedLayer": {
      "number": 19385
    }
  }
}
```

The one that thinks it's out of sync keeps printing:

```2023-09-19T12:17:50.296-0400    INFO    abcde.sync      node is too far behind      {"node_id": "abcde", "module": "sync", "sessionId": "06dadc41-5e8b-4f6a-8e67-7baff90bb12c", "current": "19395", "last synced": "19389", "behind threshold": 3, "name": "sync"}```

The only difference between them is that the "in sync" node is a private, local peer connected only to the other node, and the "out of sync" node is a public node with 40-50 peers. Unclear if this is related.

lrettig:
logfile for "out of sync" node:

[log.txt.gz](https://github.com/spacemeshos/go-spacemesh/files/12664552/log.txt.gz)

dshulyak:
i think this is a big problem. sync should be very conservative when it can interrupt consensus. in fact it should almost never interrupt it. current heuristic of 3 layers is no good, or something else makes it so that it doesn't work.
in latest outage it makes some nodes change sync state consistently. 

![image](https://github.com/spacemeshos/go-spacemesh/assets/1964576/904e228d-4fe2-42f3-865a-558cfaca5395)

dshulyak:
lets change to not synced only if node was offline (0 peers reported by libp2p) for 30 minutes. in all other cases sync should not interrupt consensus

lrettig:
> looks like all private nodes are in sync while public ones aren't (https://discord.com/channels/623195163510046732/1141739980306272427/1153806007785496698, also: https://discord.com/channels/623195163510046732/1141739980306272427/1153911699691282482)

[...]

PR #5040, opened 2023-09-20
[Merged by Bors] - sync: parametrize out of sync threshold and set it to 3h for mainnet

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/5036

in future we should drop it completely, and use only connectivity information to decide if node should stop participating
in consensus. there should be no risk of interrupting consensus, because of any unexpected failures in sync process.


config/mainnet.go
```diff
@@ -136,11 +136,12 @@ func MainnetConfig() Config {
...
-			EpochEndFraction: 0.8,
-			MaxStaleDuration: time.Hour,
-			Standalone:       false,
-			GossipDuration:   50 * time.Second,
+			Interval:           time.Minute,
+			EpochEndFraction:   0.8,
+			MaxStaleDuration:   time.Hour,
+			Standalone:         false,
+			GossipDuration:     50 * time.Second,
+			OutOfSyncThreshold: 36, // 3h
```
fasmat: I think the name should indicate that these are layers, maybe `LayerOutOfSyncThreshold`? Took me a bit to figure out how 36 means 3 h 😅

fasmat (approved):

dshulyak:
bors merge

PR #5059, opened 2023-09-21
[Merged by Bors] - make tortoise recovery interruptible

dshulyak:

check if context was cancelled , and exit with context error if it was cancelled


poszu (approved):
There are a bunch of loops in `RecoverLayer` too. Would it make sense to check if context is canceled there as well to allow cancelation during recovering layer, not only in between layers?

dshulyak:
> There are a bunch of loops in RecoverLayer too. Would it make sense to check if context is canceled there as well to allow cancelation during recovering layer, not only in between layers?

i think it is ok if interrupt will wait several seconds if some particular loop turns out to be slow. most of those inner loops should not take long time, except for the call to TallyVotes

dshulyak:
bors merge

PR #5061, opened 2023-09-21
[Merged by Bors] - Use dedicated PoST service for Proof generation

### LINKED ISSUE #5042: Integrate PoST draft API into node
## Description

Update node to use the new gRPC API to query for proof from the PoST service instead of using the `post` library directly for proof generation.

## Acceptance criteria

- Node exposes endpoint a PoST service can connect to 
  - For initial implementation assume only one PoST service connects to the node
- Node queries PoST service when it wants to generate a proof
- The initial API for the post service is defined in https://github.com/spacemeshos/api/issues/269
  - Completing and integrating this API is the goal of this ticket.

## Implementation hints

fasmat:
## Motivation
Part of #5042
Merge after #5091 

## Changes

- Generating poofs is now done via the GRPC API
  - Foundation laid in #5091
  - Upon connection `PostClient` is passed to services via the `postConnectionListener` interface. At the moment these are: `activation::Builder` and `activation::NIPostBuilder` who use it to generate a proof
  - Instead of starting the `PostSupervisor` when `PostServiceCmd` is set it is started when `StartSmeshing == true`
  - The connection does not require authentication (yet) - will be addressed in #5131 
  - Connection cannot handle multiple post services (yet)
- NiPoSTBuilder doesn't verify PoST proofs any more
  - PoST Service does this already before providing the proof
  - additionally when publishing the ATX it goes through the ATX handler that validates the ATX again before broadcasting it.
- Refactored tests in `activation` package to use new API
  - Integration tests that do not use mocks for generating PoST proofs have been moved to `activation/e2e` to allow the use of the `api/grpcserver` package in them
  - `e2e` tests spin up a post service using the post supervisor and query a proof from there
- Slimmed down `postSetupProvider` interface:
  - it is used by the `activation::Builder` and implemented by `activation::PostSetupManager`
  - some of its functionality has been moved into `PostClient` (`GenerateProof` -> `Proof`)
- Replaced `go-spacemesh/log` with `zap` in a few components in the `activation` package.

## Test Plan
- All existing tests involving proof generation have been migrated to use the new PoST service
- New tests added to test the connection specifically.
- Test added for custom types in config.

## TODO


Makefile-libs.Inc
```diff
@@ -55,17 +55,23 @@ POSTRS_SETUP_ZIP = libpost-$(platform)-v$(POSTRS_SETUP_REV).zip
 POSTRS_SETUP_URL_ZIP ?= https://github.com/spacemeshos/post-rs/releases/download/v$(POSTRS_SETUP_REV)/$(POSTRS_SETUP_ZIP)
 POSTRS_PROFILER_ZIP = profiler-$(platform)-v$(POSTRS_SETUP_REV).zip
 POSTRS_PROFILER_URL ?= https://github.com/spacemeshos/post-rs/releases/download/v$(POSTRS_SETUP_REV)/$(POSTRS_PROFILER_ZIP)
+POSTRS_SERVICE_ZIP = service-$(platform)-v0.5.0-alpha1.zip
+POSTRS_SERVICE_URL ?= https://github.com/spacemeshos/post-rs/releases/download/v0.5.0-alpha1/$(POSTRS_SERVICE_ZIP)
```
fasmat, line 59: TODO: update to not hardcode a version here, but use the general post-rs version

fasmat:
bors try

config/config.go
```diff
@@ -137,7 +137,7 @@ type SmeshingConfig struct {
 	Start           bool                              `mapstructure:"smeshing-start"`
 	CoinbaseAccount string                            `mapstructure:"smeshing-coinbase"`
 	Opts            activation.PostSetupOpts          `mapstructure:"smeshing-opts"`
-	ProvingOpts     activation.PostProvingOpts        `mapstructure:"smeshing-proving-opts"`
+	ProvingOpts     activation.PostProvingOpts        `mapstructure:"smeshing-proving-opts"` // TODO(mafa): remove?
```
fasmat: opinions?

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

cmd/root.go
```diff
@@ -245,6 +245,31 @@ func AddCommands(cmd *cobra.Command) {
...
+	cmd.PersistentFlags().Uint32Var(&cfg.POSTService.K2, "post-opts-k2",
+		cfg.POSTService.K2, "")
+	cmd.PersistentFlags().Uint32Var(&cfg.POSTService.K3, "post-opts-k3",
+		cfg.POSTService.K3, "")
+	cmd.PersistentFlags().UintVar(&cfg.POSTService.N, "post-opts-n",
+		cfg.POSTService.N, "")
+	cmd.PersistentFlags().UintVar(&cfg.POSTService.R, "post-opts-r",
+		cfg.POSTService.R, "")
+	cmd.PersistentFlags().UintVar(&cfg.POSTService.P, "post-opts-p",
+		cfg.POSTService.P, "")
```
fasmat: PoST service parameters have some overlap with PoST parameters and Smeshing parameters.

I wanted to keep them separate, but it makes no sense to set these to other values than POST/Smeshing options. Should we instead merge them into one block?

fasmat:
bors cancel

fasmat:
bors try

fasmat:
bors cancel

fasmat:
bors try

CHANGELOG.md
```diff
@@ -14,11 +14,14 @@ See [RELEASE](./RELEASE.md) for workflow instructions.
...
 
-> 2023-10-02T15:28:14.002+0200 WARN fd68b.sync mesh failed to process layer from sync {"node_id": "fd68b9397572556c2f329f3e5af2faf23aef85dbbbb7e38447fae2f4ef38899f", "module": "sync", "sessionId": "29422935-68d6-47d1-87a8-02293aa181f3", "layer_id": 23104, "errmsg": "requested layer 8063 is before evicted 13102", "name": "sync"}
+  > 2023-10-02T15:28:14.002+0200 WARN fd68b.sync mesh failed to process layer from sync {"node_id": "fd68b9397572556c2f329f3e5af2faf23aef85dbbbb7e38447fae2f4ef38899f", "module": "sync", "sessionId": "29422935-68d6-47d1-87a8-02293aa181f3", "layer_id": 23104, "errmsg": "requested layer 8063 is before evicted 13102", "name": "sync"}
 
 * [#5091](https://github.com/spacemeshos/go-spacemesh/pull/5091) First stage of separating PoST from the node into its own service.
+* [#5061](https://github.com/spacemeshos/go-spacemesh/pull/5061) Proof generation is now done via a dedicated service instead of the node.
+
+  Usage is unchanged for now, the service runs within the container and is started automatically when deployed via docker. When used via smapp the node starts the service automatically.
```
poszu: This is somewhat misleading. The reader could think that now the node requires Docker to run, which is not true.

api/grpcserver/post_service_test.go
```diff
@@ -99,8 +100,6 @@ func launchPostSupervisor(tb testing.TB, log *zap.Logger, cfg Config, postDir st
 }
 
 func Test_GenerateProof(t *testing.T) {
-	t.Parallel()
-
```
poszu, line 103: Why?

activation/nipost.go
```diff
@@ -83,17 +83,20 @@ func (nb *NIPostBuilder) persistState() {
 
 // NIPostBuilder holds the required state and dependencies to create Non-Interactive Proofs of Space-Time (NIPost).
 type NIPostBuilder struct {
```
poszu, line 85: The `NIPostBuilder` now contains three things that should probably be one:
- a `nipostClient` that returns some information about POST initialization (status, opts and commitment ATX). The 2 latter are required to be able to verify a POST proof
- a `postClient` to _generate_ POST proof
- a `nipostValidator` to _verify_ POST proof

Additionally, it must handle connections and disconnections of the `postClient`, which doesn't feel its concern. Ideally, it should call `.Proof()` and get a `VALID|INVALID|TRY AGAIN` response.

I think it should be possible to merge these 3 into a single interface covering what really the builder is interested in:

```go
type postClient interface {
    GenerateProof(ctx context.Context, challenge []byte) (*types.Post, *types.PostMetadata, error)
    VerifyProof(ctx context.Context, proof *types.Post, proofMetadata *types.PostMetadata) error // The client should probably be able to figure out the commitmentATX for on its own
}
```

---

IMHO the following:
```go
if s := nb.nipostClient.Status(); s.State != PostSetupStateComplete {
    return nil, errors.New("post setup not complete")
}
```

- doesn't make much sense as a check "if the user completed POST initialization" as he must had completed it already because the `nipostChallenge` argument either contains the initial POST proof or is for a subsequent ATX
- isn't the right place as a check "if the user still has the POST data" - this check should be in _generate proof_ routine, most likely on the post-service side, but could be handled on the node side for now too.

WDYT?

activation/post_supervisor.go
```diff
@@ -46,12 +47,18 @@ func MainnetPostServiceConfig() PostSupervisorConfig {
...
 
-	N string `mapstructure:"post-opts-n"`
+	PowDifficulty PowDifficulty `mapstructure:"post-opts-pow-difficulty"`
+	K1            uint32        `mapstructure:"post-opts-k1"`
+	K2            uint32        `mapstructure:"post-opts-k2"`
+	K3            uint32        `mapstructure:"post-opts-k3"`
+
+	N uint `mapstructure:"post-opts-n"`
+	R uint `mapstructure:"post-opts-r"`
+	P uint `mapstructure:"post-opts-p"`
```
poszu: Why duplicate the POST network configuration? The supervisor could reuse the POST config we already have defined.

api/grpcserver/post_service_test.go
fasmat: The tests kept failing on MacOS when executed in parallel due to timeouts. Even increasing the timeout from 5 to 10 seconds didn't help, but not executing the tests in parallel did.

cmd/root.go
poszu: I agree it makes no sense to have 2 places to configure the same thing. The post-service could just take the existing POST configuration structure

CHANGELOG.md
fasmat: That's where the second sentence comes in `When used via smapp ...`.

config/config.go
poszu: IMO this should be kept. Users need to tweak post-proof generation in the supervised post service.

activation/nipost.go
fasmat: > - a `nipostClient` that returns some information about POST initialization (status, opts and commitment ATX). The 2 latter are required to be able to verify a POST proof

As the go-doc for this interface explains: it is temporary and will be merged with the `postClient` interface in the future.

The reason for it being two interfaces (`nipostClient` and `postClient`) is purely a technical one: `nipostClient` is implemented by a different service than `postClient`; merging these now would need an adapter service facilitating calls to `grpcserver.PostClient` and `activation.PostSetupManager` just to later drop this adapter again when the full functionality is available via the `grpcserver.PostClient`.

---

> Additionally, it must handle connections and disconnections of the postClient, which doesn't feel its concern.

It doesn't really, it is just signalled when a new post service connects or an exiting connection is dropped - and handle these events accordingly. Alternatively there could be one `atxBuilder` that is initialised and started for every post service that connects, but this needs to "handle connections" as well, because it needs to be able to be started & stopped at any time when a service connects or disconnects.

It depends if we want `atxBuilder` to be a singleton that builds ATXs for every post service/identity or have a dedicated `atxBuilder` for every post service/identity? Right now the first approach is taken, similar to the implementation for `hare3`: https://github.com/spacemeshos/go-spacemesh/pull/5126/

The second approach would make the `atxBuilder` itself simpler, but could cost performance; for instance if there is one `atxBuilder` per identity every db query needs to be done once for every identity while the one `atxBuilder` approach would allow to combine queries into a single one or remove redundancy.

---

> I think it should be possible to merge these 3 into a single interface covering what really the builder is interested in:

I am not sure about this. I think we need at least 2 interfaces because proof generation and verification are different concerns in my opinion:

-  Proof generation is done by the post service, while proof verification is done by the node - so both methods would be implemented differently.
- Verification must be possible even if no post service connects and it makes no sense to verify a proof for every post service that connects, i.e. there is always exactly one provider for verification, while we can have any number of proof generators (even zero).

---

Regarding the `Status` call: I assume this was added as a sanity check in case something changes between the last call to `BuildNiPoST` and the current one: e.g. data corruption, files deleted, etc.

We can probably drop this call and replace it with a check if at least one post service is connected and ready to generate a proof (since we are already within the cycle gap when this call happens).

I'm thinking of adding a function to the gRPC API that returns the information in the `postdata_metadata.json` file (without generating a proof) with an additional status: "I am (not) ready to generate a proof" - where not being ready could mean that it is already generating a proof, or data is corrupted/missing, etc.

WDYT?

activation/post_supervisor.go
fasmat: Primarily for easier use in tests, I'll look into de-duplicating configuration where possible.

config/config.go
fasmat: Right now these settings are unused (only used in `PostSetupManager::GenerateProof` which is only used in 2 tests that can probably be deleted now). I could pass them to the post-service instead.

I see that it is possible to pass threads and nonces to the post service, but not `flags`; only `randomx-mode`. How do you think `flags` should be handled?

activation/nipost.go
poszu: > > Additionally, it must handle connections and disconnections of the postClient, which doesn't feel its concern.
> 
> It doesn't really, it is just signalled when a new post service connects or an exiting connection is dropped - and handle these events accordingly. Alternatively there could be one atxBuilder that is initialised and started for every post service that connects, but this needs to "handle connections" as well, because it needs to be able to be started & stopped at any time when a service connects or disconnects.

What I'm trying to say is that the "remoteness" of the post service is not a concern of the `nipostBuilder`. It just really wants to "generate a proof". The fact that there is something that "connects" and that there is some "grpc" call in between is not its concern at all.

>It depends if we want atxBuilder to be a singleton that builds ATXs for every post service/identity or have a dedicated atxBuilder for every post service/identity? Right now the first approach is taken, similar to the implementation for hare3: https://github.com/spacemeshos/go-spacemesh/pull/5126

It has nothing to do with it. Either there will be some "ATX builder manager" orchestrating many AtxBuilders or AtxBuilder orchestrating many IDs. It doesn't matter for this conversation. (Btw, I think error handling will be simpler if we keep 1 AtxBuilder per ID).

---

>> I think it should be possible to merge these 3 into a single interface covering what really the builder is interested in:
>
>I am not sure about this. I think we need at least 2 interfaces because proof generation and verification are different concerns in my opinion:
> - Proof generation is done by the post service, while proof verification is done by the node - so both methods would be implemented differently.
> - Verification must be possible even if no post service connects and it makes no sense to verify a proof for every post service that connects, i.e. there is always exactly one provider for verification, while we can have any number of proof generators (even zero).

The fact that one is talking with some post-service and the other is not is an implementation detail that is not a concern of the `nipostBuilder` and could change at any moment without impacting the builder. 

:bulb: Actually, we could argue that verification could be removed entirely now as the proof is already verified in the post-service. It would simplify the builder even more. Wdyt?

---

> Regarding the Status call: I assume this was added as a sanity check in case something changes between the last call to BuildNiPoST and the current one: e.g. data corruption, files deleted, etc.

Could be. As such I think it makes little sense to keep it.

> We can probably drop this call and replace it with a check if at least one post service is connected and ready to generate a proof (since we are already within the cycle gap when this call happens).

Wdym? There will be exactly 1 post-service per ID. If it's not connected, we retry until the publish epoch ends.

> I'm thinking of adding a function to the gRPC API that returns the information in the postdata_metadata.json file (without generating a proof) with an additional status: "I am (not) ready to generate a proof" - where not being ready could mean that it is already generating a proof, or data is corrupted/missing, etc.

Not part of PR - let's discuss it offline

cmd/root.go
```diff
@@ -240,11 +248,45 @@ func AddCommands(cmd *cobra.Command) {
...
-		"", "")
+	cmd.PersistentFlags().AddFlag(&pflag.Flag{
+		Name:     "smeshing-opts-provider",
+		Value:    &cfg.SMESHING.Opts.ProviderID,
+		DefValue: cfg.SMESHING.Opts.ProviderID.String(),
+	})
 	cmd.PersistentFlags().BoolVar(&cfg.SMESHING.Opts.Throttle, "smeshing-opts-throttle",
 		cfg.SMESHING.Opts.Throttle, "")
 
+	// TODO(mafa): add Scrypt-Opts?
```
fasmat: Opinions?

cmd/root.go
fasmat: De-duplicated config where possible

activation/post_supervisor.go
fasmat: Deduplicated parameters where possible.

activation/nipost.go
fasmat: To summarise the discussion with @poszu:

- We aim to remove `PostClient` as well the `Connected`/`Disconnected` methods signalling when a PoST service connects.
- Services that are dependent on `PostClient` at the moment will instead depend on `grpcserver::PostService`. `grpcserver::PostService` will expose methods like `Proof` with the intended `NodeId` passed as argument, so that it can forward the command to the correct post service.

config/config.go
fasmat: `ProvingOpts` are now passed to the post supervisor when `StartSmeshing == true`

activation/post_supervisor.go
```diff
@@ -19,39 +20,16 @@ import (
...
 // DefaultPostServiceConfig returns the default config for post service. These are intended for testing.
 func DefaultPostServiceConfig() PostSupervisorConfig {
-	cfg := PostSupervisorConfig{
-		PostServiceCmd:  "/bin/service",
-		DataDir:         config.DefaultDataDir,
-		NodeAddress:     "http://127.0.0.1:9093",
-		PowDifficulty:   config.DefaultConfig().PowDifficulty,
-		PostServiceMode: "light",
+	return PostSupervisorConfig{
+		PostServiceCmd: "/bin/service",
```
fasmat, line 24: TODO: check if this is actually a good default value - on windows it probably has be `./service.exe` or similar.

fasmat:
bors try

fasmat:
bors cancel

fasmat:
bors try

cmd/root.go
poszu: IMHO, it's not useful to set scrypt params on the CLI. It makes sense to change it only for tests where using a config is probably more convenient.

activation/post_supervisor_test.go
```diff
@@ -45,32 +67,53 @@ func Test_PostSupervisor_StartsServiceCmd(t *testing.T) {
...
-	require.NoError(t, ps.Close())
+	require.NoError(t, ps.Stop())
 
 	if runtime.GOOS != "windows" {
 		require.Error(t, process.Signal(syscall.Signal(0))) // check if process is closed
 	}
 }
 
-func Test_PostSupervisor_RestartsOnCrash(t *testing.T) {
+func Test_PostSupervisor_Restart_Possible(t *testing.T) {
```
poszu, line 77: Is it worth supporting restarting? Is it a useable feature?

activation/post_supervisor_win.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"os/exec"
+	"path/filepath"
+)
+
+// DefaultPostServiceConfig returns the default config for post service.
+func DefaultPostServiceConfig() PostSupervisorConfig {
+	return PostSupervisorConfig{
+		PostServiceCmd: "C:\\service.exe",
```
poszu: Why this location? Perhaps the default should be "not set"?

activation/post_supervisor_test.go
fasmat: `StartSmeshing` and `StopSmeshing` via gRPC start and stop the post service

activation/post_supervisor_win.go
fasmat: I updated the default configs to look for `service` and `service.exe` in the same folder as the node. This also reduced the build flags for windows and *nix to just a constant that is different for the two systems.

fasmat:
bors try

poszu (approved):
Approving as we agreed to continue refactoring in a follow-up.

fasmat:
bors merge

PR #5064, opened 2023-09-22
[Merged by Bors] - don't track latency for messages that are not gossipped futher

### LINKED ISSUE #4623: metrics: record latency only if object waasn't processed before
it slipped during initial review, we are not interested in recording time for duplicates that may appear after p2p evicts them from local cache. it messes up data

practically it means that latency should be recorded after as a last step in the processing pipeline

dshulyak:
this is mainly to prevent duplicates from skewing results. 
hare1 is not fixed in this change
latency metering is dropped for atx protocol, as we don't have good reference for it

closes: https://github.com/spacemeshos/go-spacemesh/issues/4623


activation/handler.go
```diff
@@ -545,6 +540,9 @@ func (h *Handler) handleAtx(ctx context.Context, expHash types.Hash32, peer p2p.
 	}
 	events.ReportNewActivation(vAtx)
 	h.log.WithContext(ctx).With().Info("new atx", log.Inline(vAtx), log.Int("size", len(msg)))
+
+	poetRoundEnd := h.clock.LayerToTime(atx.PublishEpoch.FirstLayer()).Add(h.poetCfg.PhaseShift - h.poetCfg.CycleGap)
+	metrics.ReportMessageLatency(pubsub.AtxProtocol, pubsub.AtxProtocol, time.Since(poetRoundEnd))
```
poszu: Calculating latency till this moment in time will include validation time, which might increase as nipost validations are queued. Is it what we want to measure?
```suggestion
	metrics.ReportMessageLatency(pubsub.AtxProtocol, pubsub.AtxProtocol, receivedTime.Since(poetRoundEnd))
```

dshulyak: actually i think it is better to remove this metric for atx, it is very hard to make sense of this metric in this case. because of the delays introduced by post verification, and because it can be gossiped at any point

dshulyak: @poszu i removed atx metric, ptal

fasmat: I think the raised point applies to the other changes as well; is it intended to measure latency + validation time?

poszu: Sorry for the late reply. Yes, the question still holds for the other metrics.

dshulyak: yes it matters when message reached critical parts of the protocol, not so much when it was received.

beside that processing time is always added to propagation latency on every network hop, so if there are 3 hops - processing time will be added 3 times. so also for consistency i think it is better to measure total latency including processing time on the final received that tracks metric

poszu (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

PR #5065, opened 2023-09-22
[Merged by Bors] - proposals: dont accept proposals from future

### LINKED ISSUE #4834: ballots, proposal: don't accept object from future
if someone's clock is off by more than a layer time, they may send objects for next layers.
this is useless from protocol perspective as users are submitting transactions with proposals.
tortoise is not supposed to count such ballots as well.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4834

no reason to gate ballots, as we request them with sync, so they won't be downloaded


fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #5071, opened 2023-09-25
list atxs from cached epochs from memory in atx sync

### LINKED ISSUE #4977: optimize atx sync codepath
it requests and downloads all activation ids that are known to the node. 

there main problem is that it scans database on every request, which should be mitigated by implementing smarter caching https://github.com/spacemeshos/go-spacemesh/issues/4164 . such caching should keep all activations for the epoch in cache, and not evict them on lru strategy.

secondary problem is amount of traffic that it adds, this is less straightforward to solve. maybe we should consider to drop atx sync and instead everyone should regossip its own atx every 30m.

ideally we should implement this before atx sync starts in next epoch.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4977
on top of: https://github.com/spacemeshos/go-spacemesh/pull/5013


dshulyak:
bors try

PR #5075, opened 2023-09-25
[Merged by Bors] - Fix flaky TestAccountMeshDataStream_comprehensive

### LINKED ISSUE #4171: Flaky `TestAccountMeshDataStream_comprehensive` test
## Description
Test `TestAccountMeshDataStream_comprehensive` is flaky:

```
--- FAIL: TestAccountMeshDataStream_comprehensive (1.00s)
    grpcserver_test.go:2138: 
        	Error Trace:	/home/runner/work/go-spacemesh/go-spacemesh/api/grpcserver/grpcserver_test.go:2138
        	Error:      	Not equal: 
        	            	expected: 0x2
        	            	actual  : 0x4
        	Test:       	TestAccountMeshDataStream_comprehensive
```

## Steps to reproduce
The problem was observed in regular CI run.

fasmat:
## Motivation
Closes #4171 

Occasionally the test will fail with an incorrect assertion of an error. The stream receive can fail either because the deadline was exceeded or due to bad timing for another (random) reason. Since it doesn't matter which return code is received, just that the stream ended I changed the assertion to also allow `code.Unkown`.

## Changes
- change assertion to allow `codes.DeadlineExceeded` and `codes.Unkown` instead of just the former.
- Additionally get rid of the global GRPC config. This will allow to run GRPC service tests in parallel (which will be used by https://github.com/spacemeshos/go-spacemesh/pull/5061)

## TODO

### DIFF
--- a/api/grpcserver/grpcserver_test.go
+++ b/api/grpcserver/grpcserver_test.go
@@ -10,7 +10,6 @@ import (
 	"log"
 	"math"
 	"math/big"
-	"math/rand"
 	"net"
 	"net/http"
 	"os"
@@ -477,14 +476,9 @@ func marshalProto(t *testing.T, msg proto.Message) string {
 func launchServer(tb testing.TB, services ...ServiceAPI) (Config, func()) {
 	cfg := DefaultTestConfig()
 
-	// run on a random port
-	port := 1024 + rand.Intn(9997)
-	cfg.PublicListener = fmt.Sprintf("127.0.0.1:%d", port)
-	cfg.PrivateListener = fmt.Sprintf("127.0.0.1:%d", port+1)
-	cfg.JSONListener = fmt.Sprintf("127.0.0.1:%d", port+2)
-
-	grpcService := New(cfg.PublicListener, logtest.New(tb).Named("grpc"))
-	jsonService := NewJSONHTTPServer(cfg.JSONListener, logtest.New(tb).WithName("grpc.JSON"))
+	// run on random ports
+	grpcService := New("127.0.0.1:0", logtest.New(tb).Named("grpc"))
+	jsonService := NewJSONHTTPServer("127.0.0.1:0", logtest.New(tb).WithName("grpc.JSON"))
 
 	// attach services
 	for _, svc := range services {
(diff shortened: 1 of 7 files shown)


api/grpcserver/grpcserver_test.go
```diff
@@ -496,13 +493,17 @@ func launchServer(tb testing.TB, cfg Config, services ...ServiceAPI) func() {
...
+
+	return cfg, func() {
+		assert.NoError(tb, jsonService.Shutdown(context.Background()))
+		assert.NoError(tb, grpcService.Close())
 	}
 }
 
-func callEndpoint(t *testing.T, endpoint, payload string) (string, int) {
+func callEndpoint(t *testing.T, cfg Config, endpoint, payload string) (string, int) {
 	url := fmt.Sprintf("http://%s/%s", cfg.JSONListener, endpoint)
```
poszu:
```suggestion
func callEndpoint(t *testing.T, address String, endpoint, payload string) (string, int) {
	url := fmt.Sprintf("http://%s/%s", address, endpoint)
```

api/grpcserver/grpcserver_test.go
```diff
@@ -477,11 +473,12 @@ func marshalProto(t *testing.T, msg proto.Message) string {
...
-var cfg = DefaultTestConfig()
+func launchServer(tb testing.TB, services ...ServiceAPI) (Config, func()) {
+	cfg := DefaultTestConfig()
 
-func launchServer(tb testing.TB, cfg Config, services ...ServiceAPI) func() {
-	grpcService := New(cfg.PublicListener, logtest.New(tb).Named("grpc"))
-	jsonService := NewJSONHTTPServer(cfg.JSONListener, logtest.New(tb).WithName("grpc.JSON"))
+	// run on random ports
+	grpcService := New("127.0.0.1:0", logtest.New(tb).Named("grpc"))
+	jsonService := NewJSONHTTPServer("127.0.0.1:0", logtest.New(tb).WithName("grpc.JSON"))
```
poszu, line 481: How about updating `DefaultTestConfig()`  to use `127.0.0.1:0` as the listener addresses?

poszu (approved):
TBH, I'd prefer to refactor the GRPC and JSON services to accept a `net.Listener` in their `Start()` methods and serve on it instead, but this works too.

api/grpcserver/grpcserver_test.go
fasmat: I'm a bit nervous as to what might break with this change. It seems that it should only affect tests in the`grpcserver` and `node` packages. I have re-run those and they passed, so it might be OK.

api/grpcserver/grpcserver_test.go
fasmat: I went for a slightly different approach that passes `url` directly.

fasmat:
bors merge

fasmat:
bors merge

PR #5079, opened 2023-09-25
[Merged by Bors] - increase cache size to 50 000 for next release

dshulyak:
there are many places that have to read atxs, if number of atxs in epoch larger then what we cache it causes cache threshing
signficant and disk reads. for example in this epoch nodes in cloud peek every 5m with 140MB/s reads.

i set it to 50 000 to account for possible growth in the next epoch.


countvonzero (approved):

dshulyak:
bors merge

countvonzero:
bors merge

Issue #5076, opened 2023-09-25
Flaky TestCache_Account_NonceTooSmall_AllPendingTXs


fasmat:
## Description
Test appears to be flaky. Failed in an unrelated change:

```
=== RUN   TestCache_Account_NonceTooSmall_AllPendingTXs
panic: test timed out after 5m0s
running tests:
	TestCache_Account_NonceTooSmall_AllPendingTXs (5m0s)

goroutine 899 [running]:
testing.(*M).startAlarm.func1()
	/Users/runner/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:2259 +0x1fc
created by time.goFunc
	/Users/runner/hostedtoolcache/go/1.21.1/x64/src/time/sleep.go:176 +0x45

goroutine 1 [chan receive, 4 minutes]:
testing.(*T).Run(0xc0001a44e0, {0x10155be1a, 0x2d}, 0x101641400)
	/Users/runner/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:1649 +0x856
testing.runTests.func1(0x0?)
	/Users/runner/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:2054 +0x85
testing.tRunner(0xc0001a44e0, 0xc0005bfb08)
	/Users/runner/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:1595 +0x239
testing.runTests(0xc0000ec000?, {0x1020982e0, 0x3f, 0x3f}, {0x1c?, 0x10007a599?, 0x1020b0980?})
	/Users/runner/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:2052 +0x897
testing.(*M).Run(0xc0000ec000)
	/Users/runner/hostedtoolcache/go/1.21.1/x64/src/testing/testing.go:1925 +0xb58
main.main()
	_testmain.go:171 +0x2be

goroutine 6 [select, 5 minutes]:
github.com/ipfs/go-log/writer.(*MirrorWriter).logRoutine(0xc0001182d0)
	/Users/runner/go/pkg/mod/github.com/ipfs/go-log@v1.0.5/writer/writer.go:71 +0x179
created by github.com/ipfs/go-log/writer.NewMirrorWriter in goroutine 1
	/Users/runner/go/pkg/mod/github.com/ipfs/go-log@v1.0.5/writer/writer.go:36 +0x17d

goroutine 7 [select]:
[...]

PR #5091, opened 2023-09-26
[Merged by Bors] - Add PoST service to node

fasmat:
## Motivation
Part of #5042

This PR adds the GRPC API for the PoST service and adds tests verifying it works correctly.

## Changes
- PoST service is downloaded with tests and builds, as well as included in the `Dockerfile`
- `PostSetupManager` now uses `zap` directly instead of the `go-spacemesh/log` wrapper
- `grpcserver.PostService` uses the PoST service API is added to gRPCs `PrivateListener` endpoint
  - temporary, will be changed to public + authentication in the future
- Any amount of PoST services can connect to the API but only the first calling `Register` will be able to successfully create a stream
  - in the future this will be extended to allow any number of PoST services to connect
  - a successful `Register` will create a `PostClient` and notify all services that are `postConnectionListener` about the service connecting (and disconnecting)
- TODO: bundled PoST service is started when the node starts 

## Test Plan
- new tests have been added for gRPC API

## TODO


activation/post.go
```diff
@@ -358,13 +358,13 @@ func (mgr *PostSetupManager) StartSession(ctx context.Context) error {
...
-		log.String("num_units", fmt.Sprintf("%d", mgr.lastOpts.NumUnits)),
-		log.String("labels_per_unit", fmt.Sprintf("%d", mgr.cfg.LabelsPerUnit)),
-		log.String("provider", fmt.Sprintf("%d", mgr.lastOpts.ProviderID)),
+	mgr.logger.Info("post setup session starting",
+		zap.Stringer("node_id", mgr.id),
+		zap.Stringer("commitment_atx", mgr.commitmentAtxId),
+		zap.String("data_dir", mgr.lastOpts.DataDir),
+		zap.String("num_units", fmt.Sprintf("%d", mgr.lastOpts.NumUnits)),
+		zap.String("labels_per_unit", fmt.Sprintf("%d", mgr.cfg.LabelsPerUnit)),
+		zap.String("provider", fmt.Sprintf("%d", mgr.lastOpts.ProviderID)),
```
poszu:
```suggestion
		zap.Uint32("num_units", mgr.lastOpts.NumUnits),
		zap.Uint64("labels_per_unit", mgr.cfg.LabelsPerUnit),
		zap.Stringer("provider", mgr.lastOpts.ProviderID),
```

activation/post.go
```diff
@@ -379,26 +379,26 @@ func (mgr *PostSetupManager) StartSession(ctx context.Context) error {
...
-		log.String("num_units", fmt.Sprintf("%d", mgr.lastOpts.NumUnits)),
-		log.String("labels_per_unit", fmt.Sprintf("%d", mgr.cfg.LabelsPerUnit)),
-		log.String("provider", fmt.Sprintf("%d", mgr.lastOpts.ProviderID)),
+	mgr.logger.Info("post setup completed",
+		zap.Stringer("node_id", mgr.id),
+		zap.Stringer("commitment_atx", mgr.commitmentAtxId),
+		zap.String("data_dir", mgr.lastOpts.DataDir),
+		zap.String("num_units", fmt.Sprintf("%d", mgr.lastOpts.NumUnits)),
+		zap.String("labels_per_unit", fmt.Sprintf("%d", mgr.cfg.LabelsPerUnit)),
+		zap.String("provider", fmt.Sprintf("%d", mgr.lastOpts.ProviderID)),
```
poszu:
```suggestion
		zap.Uint32("num_units", mgr.lastOpts.NumUnits),
		zap.Uint64("labels_per_unit", mgr.cfg.LabelsPerUnit),
		zap.Stringer("provider", mgr.lastOpts.ProviderID),
```

api/grpcserver/config.go
```diff
@@ -35,7 +36,7 @@ func DefaultConfig() Config {
 	return Config{
 		PublicServices:        []Service{Debug, GlobalState, Mesh, Transaction, Node, Activation},
 		PublicListener:        "0.0.0.0:9092",
-		PrivateServices:       []Service{Admin, Smesher},
+		PrivateServices:       []Service{Admin, Smesher, Post}, // TODO(mafa): move from private to public with authentication (probably new service category)
```
fasmat, line 39: TODO: create ticket for this

Makefile-libs.Inc
```diff
@@ -50,22 +50,28 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.4.4
+POSTRS_SETUP_REV = 0.5.0-alpha2
```
fasmat, line 53: Should we create a proper release for this or leave it an alpha for now?

api/grpcserver/post_client.go
```diff
@@ -0,0 +1,101 @@
...
+			if !ok {
+				return nil, nil, fmt.Errorf("unexpected response of type: %T", resp.GetKind())
+			}
+			switch proofResp.GenProof.GetStatus() {
+			case pb.GenProofStatus_GEN_PROOF_STATUS_OK:
+				if proofResp.GenProof.GetProof() == nil {
+					select {
+					case <-ctx.Done():
+						return nil, nil, ctx.Err()
+					case <-time.After(2 * time.Second): // TODO(mafa): make polling interval configurable
```
fasmat, line 69: Will be addressed in https://github.com/spacemeshos/go-spacemesh/pull/5061/

fasmat:
bors try

fasmat:
bors cancel

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

Makefile-libs.Inc
poszu: IMHO, a regular release and then add the next features in patch versions.

activation/post_service.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		ps.eg.Go(ps.captureCmdOutput(pipe))
+		if err := ps.cmd.Start(); err != nil {
+			pipe.Close()
+			stop()
+			ps.logger.Error("restart post service", zap.Error(err))
+			return nil
+		}
+		ps.stop = stop
+		ps.eg.Go(ps.monitorCmd(opts))
```
poszu: The `ps.cmdMtx` is still held here, so the spawned `monitorCmd()` might just return after failed `TryLock()` _sometimes_.

activation/post_service.go
```diff
@@ -0,0 +1,166 @@
...
+				ps.logger.Warn("read from post service pipe", zap.Error(err))
+				return nil
+			}
+
+			ps.logger.Debug(string(buf[:n]))
+		}
+	}
+}
+
+func (ps *PostService) monitorCmd(opts PostServiceConfig) func() error {
```
poszu: I think that the code would be simpler and could avoid sharing memory (and mutexes) if this method was:
- just running a flat loop instead of recursion
- keeping the `cmd` local. AFAIR, you need `cmd` on the `PostService` struct only for testing. Perhaps this can be tested in other way? Or you could just set the `pid` in an `PostService::atomic.Value`?

Consider something like this:
```go
// PostService manages a local post service.
type PostService struct {
	logger *zap.Logger

	stop context.CancelFunc // stops the running command.

	eg errgroup.Group // eg manages post service supervisor
}

// NewPostService returns a new post service.
func NewPostService(logger *zap.Logger, opts PostServiceConfig) (*PostService, error) {
	ctx, stop := context.WithCancel(context.Background())

	ps := &PostService{
		logger: logger,
		stop:   stop,
	}
	ps.eg.Go(func() error { return ps.run(ctx, opts) })

	return ps, nil
}

func (ps *PostService) Close() error {
	ps.stop()
	ps.eg.Wait()
	return nil
}

// captureCmdOutput returns a function that reads from the given pipe and logs the output.
// it returns when the pipe is closed.
func (ps *PostService) captureCmdOutput(pipe io.ReadCloser) func() error {
	return func() error {
		for {
			buf := make([]byte, 1024)
			n, err := pipe.Read(buf)
			switch err {
			case nil:
			case io.EOF:
				return nil
			default:
				ps.logger.Warn("read from post service pipe", zap.Error(err))
				return nil
			}

			ps.logger.Debug(string(buf[:n]))
		}
	}
}

func (ps *PostService) run(ctx context.Context, opts PostServiceConfig) error {
	for {
		cmd := exec.CommandContext(
			ctx,
			opts.PostServiceCmd,
			"--dir", opts.DataDir,
			"--address", opts.NodeAddress,
			"--pow-difficulty", opts.PowDifficulty.String(),
			"--randomx-mode", opts.PostServiceMode,
		)
		cmd.Dir = filepath.Dir(opts.PostServiceCmd)
		pipe, err := cmd.StderrPipe()
		if err != nil {
			return fmt.Errorf("setup stderr pipe for post service: %w", err)
		}

		if cmd.Start(); err != nil {
			pipe.Close()
			return fmt.Errorf("start post service: %w", err)
		}
		events.EmitPostServiceStarted()

		var eg errgroup.Group
		eg.Go(ps.captureCmdOutput(pipe))
		err = cmd.Wait()
		ps.logger.Warn("post service exited", zap.Error(err))
		events.EmitPostServiceStopped()
		if err := ctx.Err(); err != nil {
                        if err := eg.Wait(); err != nil {
                                ps.logger.Warn("output reading goroutine failed", zap.Error(err))
                        }
			return ctx.Err()
		}
                if err := eg.Wait(); err != nil {
                        ps.logger.Warn("output reading goroutine failed", zap.Error(err))
                }
	}
}
```

fasmat: I ended up doing something quite similar to what you suggested; I even found a way to make it testable 🙂, although with the need for a new dependency.

While `os.FindProcess` and related functionality in the stdlib can do what I need on Linux and Mac, it fails to e.g. check for a running process on windows. Doing this with only the stdlib would be a pain, so I decided to use `github.com/shirou/gopsutil` for this.

activation/post_service.go
```diff
@@ -0,0 +1,148 @@
...
+	PowDifficulty   PowDifficulty `mapstructure:"post-opts-pow-difficulty"`
+	PostServiceMode string        `mapstructure:"post-opts-post-service-mode"`
+}
+
+// PostService manages a local post service.
+type PostService struct {
+	logger *zap.Logger
+
+	pidMtx sync.RWMutex
+	pid    int // pid of the running post service.
```
poszu: It's just a single value that needs to be set atomically. How about:
```suggestion
	pid atomic.Int64 // pid of the running post service.
```

activation/post_service.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// captureCmdOutput returns a function that reads from the given pipe and logs the output.
+// it returns when the pipe is closed.
+func (ps *PostService) captureCmdOutput(pipe io.ReadCloser) func() error {
+	return func() error {
+		for {
+			buf := make([]byte, 1024)
+			n, err := pipe.Read(buf)
+			switch err {
```
poszu: Would it make sense to reuse the `buf`?
```suggestion
		buf := make([]byte, 1024)
		for {
			n, err := pipe.Read(buf)
			switch err {
```

activation/post_service.go
```diff
@@ -0,0 +1,148 @@
...
+			events.EmitPostServiceStopped()
+			if err := eg.Wait(); err != nil {
+				ps.logger.Warn("output reading goroutine failed", zap.Error(err))
+			}
+			return ctx.Err()
+		}
+		ps.logger.Warn("post service exited", zap.Error(err))
+		if err := eg.Wait(); err != nil {
+			ps.logger.Warn("output reading goroutine failed", zap.Error(err))
+		}
```
poszu: I just noticed that it never returns an error. How about
```suggestion
		_ = eg.Wait()
```

activation/post_service.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+}
+
+func (ps *PostService) Pid() int {
+	ps.pidMtx.RLock()
+	defer ps.pidMtx.RUnlock()
+	return ps.pid
```
poszu: With the atomic it becomes
```suggestion
	return int(ps.pid.Load())
```

activation/post_service.go
```diff
@@ -0,0 +1,148 @@
...
+
+		var eg errgroup.Group
+		eg.Go(ps.captureCmdOutput(pipe))
+		if cmd.Start(); err != nil {
+			pipe.Close()
+			return fmt.Errorf("start post service: %w", err)
+		}
+		ps.pidMtx.Lock()
+		ps.pid = cmd.Process.Pid
+		ps.pidMtx.Unlock()
```
poszu: With the atomic:
```suggestion
		ps.pid.Store(int64(cmd.Process.Pid))
```

activation/post_service_test.go
```diff
@@ -0,0 +1,161 @@
...
+	require.NoError(tb, mgr.StartSession(context.Background()))
+	require.NoError(tb, eg.Wait())
+	require.Equal(tb, PostSetupStateComplete, mgr.Status().State)
+}
+
+func Test_PostService_StartsServiceCmd(t *testing.T) {
+	log := zaptest.NewLogger(t)
+
+	postDir := t.TempDir()
+	initPost(t, log.Named("post init"), postDir)
```
poszu: You don't really need to init POST for this test (and the other one either).

api/grpcserver/post_client.go
```diff
@@ -0,0 +1,101 @@
...
+
+func newPostClient(service *PostService) *postClient {
+	return &postClient{
+		service: service,
+		closed:  make(chan struct{}),
+	}
+}
+
+func (pc *postClient) Proof(ctx context.Context, challenge []byte) (*types.Post, *types.PostMetadata, error) {
+	ch := make(chan *pb.ServiceResponse, 1)
```
poszu: I prefer to name variables after they use, rather than by their type. How about
```suggestion
	response := make(chan *pb.ServiceResponse, 1)
```

api/grpcserver/post_client.go
```diff
@@ -0,0 +1,101 @@
...
+			},
+		},
+		ch: ch,
+	}
+
+	select {
+	case <-pc.closed:
+		return nil, nil, fmt.Errorf("post client closed")
+	default:
+	}
```
poszu: Isn't it redundant? It doesn't really guard for not calling `pc.service.Send()` on a closed `pc`, because it could be closed right after this check. Thus `pc.service` should be able to handle this case anyway. It makes it more difficult to test this corner case.

api/grpcserver/post_client.go
```diff
@@ -0,0 +1,101 @@
...
+				if proofResp.GenProof.GetProof() == nil {
+					select {
+					case <-ctx.Done():
+						return nil, nil, ctx.Err()
+					case <-time.After(2 * time.Second): // TODO(mafa): make polling interval configurable
+					}
+				}
+
+				proof := proofResp.GenProof.GetProof()
+				meta := proofResp.GenProof.GetMetadata()
```
poszu: There is a helper for this: `resp.GetGenProof()`:
```suggestion
			genProof := resp.GetGenProof()
			if genProof == nil {
				return nil, nil, fmt.Errorf("unexpected response of type: %T", resp.GetKind())
			}
			switch genProof.GetStatus() {
			case pb.GenProofStatus_GEN_PROOF_STATUS_OK:
				if genProof.GetProof() == nil {
					select {
					case <-ctx.Done():
						return nil, nil, ctx.Err()
					case <-time.After(2 * time.Second): // TODO(mafa): make polling interval configurable
					}
				}

				proof := genProof.GetProof()
				meta := genProof.GetMetadata()
```

api/grpcserver/post_service.go
```diff
@@ -0,0 +1,135 @@
...
+// The bidirectional stream established between the node and the PoST node can be used
+// to send challenges and receive proofs.
+type PostService struct {
+	log       *zap.Logger
+	callbacks []postConnectionListener
+
+	// connections is a map of IDs -> channel of commands for every connected node.
+	// the node calls functions on this service which build API requests for every
+	// connected node and send them on the channel. The PoST node receives the requests
+	// and sends responses back on the response channel of the command.
```
poszu: This comment seems outdated. There is only 1 channel for postCommands

api/grpcserver/post_service.go
```diff
@@ -0,0 +1,135 @@
...
+// requests to the PoST node and receive responses.
+func (s *PostService) Register(stream pb.PostService_RegisterServer) error {
+	ch := make(chan postCommand)
+	s.setConnection(ch)
+	defer s.dropConnection()
+
+	for {
+		select {
+		case <-stream.Context().Done():
+			return stream.Context().Err()
```
poszu: The producer of commands (who pushes into `ch`) might deadlock (stuck in send to channel) if the post service disconnects because the channel is unbuffered.

api/grpcserver/post_service.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return s.con != nil
+}
+
+func (s *PostService) dropConnection() error {
+	s.conMtx.Lock()
+	defer s.conMtx.Unlock()
+
+	if s.con == nil {
+		return fmt.Errorf("connection not established")
```
poszu: Hitting this case means a serious bug. As the error code is not checked anyway, it will slip unnoticed, I'd recommend just panicking here.

api/grpcserver/post_service_test.go
```diff
@@ -0,0 +1,262 @@
...
+	initPost(t, log.Named("post1"), postDir1)
+	t.Cleanup(launchPostService(t, log.Named("post1"), cfg, postDir1))
+
+	postDir2 := t.TempDir()
+	initPost(t, log.Named("post2"), postDir2)
+	t.Cleanup(launchPostService(t, log.Named("post2"), cfg, postDir2))
+
+	postDir3 := t.TempDir()
+	initPost(t, log.Named("post3"), postDir3)
+	t.Cleanup(launchPostService(t, log.Named("post3"), cfg, postDir3))
```
poszu: Why would the other fail to connect?

api/grpcserver/post_service_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+			tb.Log(string(buf[:n]))
+		}
+	})
+
+	return func() {
+		cmd.Process.Kill()
+		assert.NoError(tb, eg.Wait())
+	}
+}
```
poszu: Wouldn't it be better to reuse the post-service supervisor that is already implemented?

api/grpcserver/interface.go
```diff
@@ -86,3 +86,8 @@ type meshAPI interface {
 type oracle interface {
 	ActiveSet(context.Context, types.EpochID) ([]types.ATXID, error)
 }
+
+type postConnectionListener interface {
+	Connected(client activation.PostClient)
+	Disconnected(client activation.PostClient)
+}
```
poszu, line 93: Note: the system is more interested in post-services _registering_ with an ID than just _connecting_. (I know that it doesn't send the ID _yet_ - TBD).

What's the purpose of `Disconnected()`?

activation/post_service.go
fasmat: Yeah, it is also only used in a single test and shouldn't be used anywhere else. I'll get rid of the `Pid()` method as well.

activation/post_service.go
fasmat: I instead opted for accessing the field directly in a test because it shouldn't be used outside the test anyway.

activation/post_service.go
fasmat: I ended up using a `bufio.Reader` instead of manually processing the output in chunks.

This also solved a cosmetic issue I had with the logs where a lot of empty lines where printed.

activation/post_service_test.go
fasmat: when I wrote the test I needed it (or the service would exit). I guess it can be removed now 🙂

api/grpcserver/post_service_test.go
fasmat: Fail to connect is probably incorrectly worded, what I meant was all but the first ervice will not be able to open a stream with the node by calling `Register`.

I'll update the comment to make it clearer.

api/grpcserver/post_service_test.go
fasmat: I wrote the test before I had the supervisor - updated the test to use it instead 👍

activation/post_supervisor_test.go
```diff
@@ -0,0 +1,92 @@
...
+		PowDifficulty:   DefaultPostConfig().PowDifficulty,
+		PostServiceMode: "light",
+	}
+
+	ps, err := NewPostSupervisor(log.Named("supervisor"), opts)
+	require.NoError(t, err)
+	require.NotNil(t, ps)
+	t.Cleanup(func() { assert.NoError(t, ps.Close()) })
+
+	time.Sleep(500 * time.Millisecond) // to ensure cmd actually started
```
poszu: How about using `Eventually()` with a bigger timeout instead? Otherwise, it will be flaky.

fasmat: What would the condition for `Eventually` be?

fasmat: Never mind, I think I found an assertion that works 🙂

api/grpcserver/interface.go
fasmat: I used a pub-sub like pattern. Any component of the node that is interested in post services connecting or disconnecting might use this interface to be informed about a service opening a stream to the node (or dropping the stream).

Since a component is informed when a service opened a stream ("Connected") it is also informed when this service closes the stream (the object passed into the service will be the same object in both cases / `==` will be true) and can free resources when this happens.

api/grpcserver/post_service.go
fasmat: It just means that `dropConnection()` was called twice. I would even go the other way and `return nil` in that case since this check is only there to avoid calling `close` on a `nil` channel and thereby preventing a panic. It's kinda like calling `Close` on an `io.Closer` twice. It might return an error but I would not expect code to panic just because I closed something that was already closed?

api/grpcserver/post_service.go
fasmat: Both the sending and the closing of the `ch` channel are protected by the same mutex:

```go
func (s *PostService) dropConnection() error {
	s.conMtx.Lock()
	defer s.conMtx.Unlock()

	if s.con == nil {
		return fmt.Errorf("connection not established")
	}

	close(s.con)
	s.con = nil
	s.client.Close()

	for _, cb := range s.callbacks {
		cb.Disconnected(s.client)
	}

	s.client = nil
	return nil
}
```

```go
func (s *PostService) Send(ctx context.Context, cmd postCommand) error {
	s.conMtx.Lock()
	defer s.conMtx.Unlock()

	if s.con == nil {
		return fmt.Errorf("connection not established")
	}

	select {
	case <-ctx.Done():
		return ctx.Err()
	case s.con <- cmd:
		return nil
	}
}
```

I don't think a deadlock as you describe is possible in this code. The reason `Send` is mutually exclusive is to avoid two parts of the node sending a command to the post service before the response is received.

api/grpcserver/post_client.go
fasmat: I was primarily concerned with giving a concrete error when trying to send a request when the client has already been closed.

If `Proof` is called when `pc.closed` has already been closed it will return the error `post client closed`. Should the close happen right between this check and the following `pc.service.Send` it will return `failed to send request: connection not established`.

Without this check the client would send to a (possibly) different Post service as soon as connection is re-established. I know this can theoretically still happen, but for this the current connection has to be dropped and a new connection established between here and `pc.service.Send` which I don't think is enough time to happen?

api/grpcserver/post_service.go
fasmat: I removed `PostService::Send` in favour of passing `s.con` directly to the `PostClient`. Every time it tries to send on this channel it will also check `pc.closed` in case it has been closed already (an no one is receiving on `s.con` anymore). This should avoid deadlocks.

api/grpcserver/post_service.go
fasmat: I changed it to just return `nil` since I don't think calling `dropConnection` when there is no connection is worth returning an error.

poszu (approved):

fasmat:
bors merge

fasmat:
flaky test: https://github.com/spacemeshos/go-spacemesh/issues/5076

bors try

fasmat:
bors merge

PR #5092, opened 2023-09-26
[Merged by Bors] - Reenable SA1019 lint

poszu:
## Motivation
Reenable disabled lint

## Changes
- fixed code
- reenabled staticcheck SA1019

### DIFF
--- a/node/node_test.go
+++ b/node/node_test.go
@@ -1,6 +1,5 @@
 package node
 
-//lint:file-ignore SA1019 hide deprecated protobuf version error
 import (
 	"bytes"
 	"context"
@@ -12,13 +11,10 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
-	"strings"
 	"sync"
 	"testing"
 	"time"
 
-	"github.com/golang/protobuf/jsonpb"
-	"github.com/golang/protobuf/proto"
 	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
 	pb "github.com/spacemeshos/api/release/go/spacemesh/v1"
 	"github.com/spacemeshos/post/initialization"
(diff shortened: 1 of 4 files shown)


node/node_test.go
```diff
@@ -256,23 +254,22 @@ func TestSpacemeshApp_Cmd(t *testing.T) {
...
-	return buf.String()
+func marshalProto(t *testing.T, msg proto.Message) []byte {
+	buf, err := protojson.Marshal(msg)
+	require.NoError(t, err)
+	return buf
 }
 
-func callEndpoint(t *testing.T, endpoint, payload, address string) (string, int) {
+func callEndpoint(t *testing.T, endpoint string, payload []byte, address string) ([]byte, int) {
 	url := fmt.Sprintf("http://%s/%s", address, endpoint)
```
fasmat: The parameter order is odd to me: `endpoint`, `payload`, `address`. Maybe this could be changed to just `url`, `payload` similar to how it is in the `grpcserver_test.go` file?

node/node_test.go
```diff
@@ -300,11 +297,13 @@ func TestSpacemeshApp_GrpcService(t *testing.T) {
 	r.NoError(err)
 	r.Empty(app.Config.API.PublicServices)
 
-	_, err = grpc.Dial(
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
fasmat, line 299: The reduction in timeout is intentional? maybe the longer timeout is necessary on slower builders?

fasmat (approved):

node/node_test.go
poszu: Yes... this test checks if the GRPC service is not running when not enabled. It will basically wait for the whole timeout duration. I think 1 second is more than enough.

poszu:
bors merge

poszu:
Unrelated flaky UT on arm64 Mac.

Bors merge

PR #5097, opened 2023-09-27
[Merged by Bors] - atxs: regossip atxs in publish epoch

### LINKED ISSUE #5068: regossip atx before target epoch
amount of atx traffic at the start of this epoch indicates that atxs are not sufficiently propagated. it doesn't create any immediate issues with protocol, but not as efficient as it could be.

image

to mitigate this problem we should regossip our own atx every 30-60m after it was built before target epoch.

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/5068

there is an indication that not all atxs are propagated before target epoch starts. it results in efficiency in some data structures and in future will lead to lost rewards.

if node was offline when built an atx or experienced some other problems, the only chance it has to broadcast an atx is when peers asks for it (atx sync). that approach is in general not scalable and naive about concurrent requests.  

as atx are not large (~1KB) we can regossip them periodically. if peer already stores this atx it will not gossip it futher

### DIFF
--- a/activation/activation.go
+++ b/activation/activation.go
@@ -59,9 +59,10 @@ const (
 
 // Config defines configuration for Builder.
 type Config struct {
-	CoinbaseAccount types.Address
-	GoldenATXID     types.ATXID
-	LayersPerEpoch  uint32
+	CoinbaseAccount  types.Address
+	GoldenATXID      types.ATXID
+	LayersPerEpoch   uint32
+	RegossipInterval time.Duration
 }
 
 // Builder struct is the struct that orchestrates the creation of activation transactions
@@ -79,6 +80,7 @@ type Builder struct {
 	coinbaseAccount   types.Address
 	goldenATXID       types.ATXID
 	layersPerEpoch    uint32
+	regossipInterval  time.Duration
 	cdb               *datastore.CachedDB
 	atxHandler        atxHandler
 	publisher         pubsub.Publisher
@@ -165,6 +167,7 @@ func NewBuilder(
 		coinbaseAccount:       conf.CoinbaseAccount,
 		goldenATXID:           conf.GoldenATXID,
 		layersPerEpoch:        conf.LayersPerEpoch,
+		regossipInterval:      conf.RegossipInterval,
 		cdb:                   cdb,
 		atxHandler:            hdlr,
 		publisher:             publisher,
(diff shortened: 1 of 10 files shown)


activation/activation.go
```diff
@@ -165,6 +167,7 @@ func NewBuilder(
 		coinbaseAccount:       conf.CoinbaseAccount,
 		goldenATXID:           conf.GoldenATXID,
 		layersPerEpoch:        conf.LayersPerEpoch,
+		regossipInterval:      conf.RegossipInterval,
```
poszu, line 170: Would it make sense to set a minimum interval to avoid spamming when bad config is applied? Like:
```suggestion
		regossipInterval:      max(time.Second * 30, conf.RegossipInterval),
```

activation/handler.go
```diff
@@ -493,6 +496,9 @@ func (h *Handler) HandleGossipAtx(ctx context.Context, peer p2p.Peer, msg []byte
 			log.Err(err),
 		)
 	}
+	if errors.Is(err, errKnownAtx) && peer == h.local {
+		return nil
+	}
```
fasmat, line 501: What happens on remote nodes if they receive the re-gossiped ATXs? Will this cause all nodes to keep logging warnings / errors about known ATXs?

activation/activation_test.go
```diff
@@ -1271,6 +1271,34 @@ func TestWaitPositioningAtx(t *testing.T) {
...
+	t.Run("success", func(t *testing.T) {
+		h := newTestBuilder(t)
+		atx := newActivationTx(t,
+			h.signer, 0, types.EmptyATXID, types.EmptyATXID, nil,
+			layer.GetEpoch(), 0, 1, types.Address{}, 1, &types.NIPost{})
+		require.NoError(t, atxs.Add(h.cdb.Database, atx))
+		blob, err := atxs.GetBlob(h.cdb.Database, atx.ID().Bytes())
+		require.NoError(t, err)
+		h.mclock.EXPECT().CurrentLayer().Return(layer)
+		h.mpub.EXPECT().Publish(gomock.Any(), gomock.Any(), blob)
```
poszu: NIT:
```suggestion
		h.mpub.EXPECT().Publish(gomock.Any(), pubsub.AtxProtocol, blob)
```

poszu (approved):

activation/handler.go
dshulyak: the expected behavior that they will drop them without gossiping further.
my hope was that we don't log them every time, i will check that it is the case.

activation/activation.go
dshulyak: libp2p also maintains a time-based cache with sent/received messages. so spamming Publish won't spam neighbours. i will check how long it keeps that in memory

dshulyak: > 	// TimeCacheDuration specifies how long a message ID will be remembered as seen.
	// Use WithSeenMessagesTTL to configure this per pubsub instance, instead of overriding the global default.
	TimeCacheDuration = 120 * time.Second

activation/handler.go
dshulyak: code doesn't spam, so neighbor will simply drop atx if it exists already

dshulyak:
bors merge

dshulyak:
bors merge

PR #4714, opened 2023-07-17
hardcode remaining parameters

dshulyak:

PR #5102, opened 2023-09-27
[Merged by Bors] - Add poet112 and bootnodes to the config

pigmej:
Some time ago, we decided to NOT include these ( https://github.com/spacemeshos/go-spacemesh/pull/4714 ), but I think we're stable enough on the bootnodes and poet servers currently, so we can commit to it.


dshulyak (approved):

pigmej:
bors merge

poszu (approved):

PR #5103, opened 2023-09-27
[Merged by Bors] - Pass proof deadline when registering in poet

poszu:
## Motivation
Avoid slipping into the next poet round when submitting a challenge late. It's a prerequisite to support phased poets. It will let the nodes avoid wrongly registering in poets with earlier phases (their round has already started). 

Depends on https://github.com/spacemeshos/poet/pull/410

## Changes
Pass the required time (deadline) by which the poet proof must be ready. If poet decides that the current open round ends after that time - it will reject challenge registration.

## Test Plan
- existing tests should pass
- added poet client unit test to see if the error is properly detected and handled
- the functionality itself is tested in poet tests


poszu:
bors try

activation/nipost.go
```diff
@@ -372,28 +375,45 @@ func (nb *NIPostBuilder) submitPoetChallenge(ctx context.Context, client PoetPro
...
+	for result := range poetRequestsChannel {
+		if result.err == nil {
+			poetRequests = append(poetRequests, *result.request)
+			allTooLate = false
+		} else {
+			nb.log.With().Warning("failed to submit challenge to poet", log.Err(result.err))
+			if !errors.Is(result.err, ErrSubmitTooLate) {
+				allTooLate = false
+			}
+		}
```
fasmat: NIT: less nesting is usually easier to read 🙂 

```suggestion
		if result.err == nil {
			poetRequests = append(poetRequests, *result.request)
			allTooLate = false
			continue
		}
		
		nb.log.With().Warning("failed to submit challenge to poet", log.Err(result.err))
		if !errors.Is(result.err, ErrSubmitTooLate) {
			allTooLate = false
		}
```

activation/poet.go
```diff
@@ -257,13 +262,15 @@ func (c *HTTPPoetClient) req(ctx context.Context, method, path string, reqBody,
...
+	case res.StatusCode == http.StatusOK:
+	case res.StatusCode == http.StatusNotFound:
 		return fmt.Errorf("%w: response status code: %s, body: %s", ErrNotFound, res.Status, string(data))
-	case http.StatusServiceUnavailable:
+	case res.StatusCode == http.StatusServiceUnavailable:
 		return fmt.Errorf("%w: response status code: %s, body: %s", ErrUnavailable, res.Status, string(data))
-	case http.StatusBadRequest:
+	case res.StatusCode == http.StatusBadRequest && strings.Contains(string(data), registration.ErrTooLateToRegister.Error()):
+		return fmt.Errorf("%w: response status code: %s, body: %s", ErrSubmitTooLate, res.Status, string(data))
+	case res.StatusCode == http.StatusBadRequest:
```
fasmat: Instead of splitting these two into their own cases, I think I would leave `switch res.StatusCode` and just have a `if` in the `case http.StatusBadRequest`.

activation/poet.go
```diff
@@ -257,13 +262,15 @@ func (c *HTTPPoetClient) req(ctx context.Context, method, path string, reqBody,
...
+	switch {
+	case res.StatusCode == http.StatusOK:
+	case res.StatusCode == http.StatusNotFound:
 		return fmt.Errorf("%w: response status code: %s, body: %s", ErrNotFound, res.Status, string(data))
-	case http.StatusServiceUnavailable:
+	case res.StatusCode == http.StatusServiceUnavailable:
 		return fmt.Errorf("%w: response status code: %s, body: %s", ErrUnavailable, res.Status, string(data))
-	case http.StatusBadRequest:
+	case res.StatusCode == http.StatusBadRequest && strings.Contains(string(data), registration.ErrTooLateToRegister.Error()):
+		return fmt.Errorf("%w: response status code: %s, body: %s", ErrSubmitTooLate, res.Status, string(data))
```
fasmat: I'm not a huge fan of asserting strings in the response. Arguably `ErrSubmitTooLate` doesn't need to be its own error since it is an invalid request independent of why. `ErrSubmitTooLate` is also only used in a test where asserting that the returned error is `ErrInvalidRequest` in the case of a deadline that cannot be met would in my opinion be sufficient as well.

The full error text still contains the info that it was because of the deadline for additional information in logging.

poszu: Me neither, but GRPC doesn't really give another way (at least I'm not aware of it) to signal different error reasons.

I need to distinguish between just an invalid request (for example bad pow params, signature etc. that are bugs) and a plausible scenario of submitting too late. This latter is handled in the code by expiring the ATX.

fasmat: What I mean is all of these (bad pow params, signature and deadline) are invalid requests in the case of the first two the node just tries again 5 minutes later (probably receiving the same response) but with the later it expires the challenge.

Shouldn't it expire the challenge either way?

poszu: Updated as discussed. Now every 400 is treated the same way as a client error.

activation/poet_test.go
```diff
@@ -62,14 +59,61 @@ func TestHTTPPoet(t *testing.T) {
 	signature := signer.Sign(signing.POET, ch.Bytes())
 	prefix := bytes.Join([][]byte{signer.Prefix(), {byte(signing.POET)}}, nil)
 
-	poetRound, err := client.Submit(context.Background(), prefix, ch.Bytes(), signature, signer.NodeID(), PoetPoW{
+	poetRound, err := client.Submit(context.Background(), time.Time{}, prefix, ch.Bytes(), signature, signer.NodeID(), PoetPoW{
```
fasmat, line 62: I'm surprised this works. Is the `HTTPPoetTestHarness` not configured correctly or an older version that is not yet aware about deadlines?

Or does the zero value of `time.Time` mean no deadline?

poszu: The zero value means no deadline: https://github.com/spacemeshos/poet/blob/bc84a36c1c4f9e7b16bbb230bccdefa7d1c94a71/registration/registration.go#L306

fasmat (approved):

poszu:
bors merge

fasmat (approved):

poszu:
bors merge

PR #5109, opened 2023-09-28
[Merged by Bors] - recover only N latest layers required for correct ballots decoding 

### LINKED ISSUE #3006: tortoise: tortoise state loading will be growing in time

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/3006

tortoise recovers state to decode votes (that are encoded as base ballot + delta) and verify that signed opinion is available locally. when we are accepting votes from network we need to verify those two things. you can see how this recursiveness goes back to genesis. current code reloads ballots from genesis as it is the most fail safe approach. but this have a downside that overtime we will have to load unbounded amount of data, which is not practical. 

in this change tortoise loads ballots that are useful only for next ballots that we expect to receive from the network. the algorithm looks up latest applied layer, subtracts number of layers which are set as tortoise window (10 000), and finds the first layer of that epoch.


dshulyak:
bors try

tortoise/recover.go
```diff
@@ -29,6 +29,23 @@ func Recover(ctx context.Context, db *datastore.CachedDB, latest types.LayerID,
...
 
+	applied, err := layers.GetLastApplied(db)
+	if err != nil {
+		return nil, fmt.Errorf("get last applied: %w", err)
+	}
+	start := types.GetEffectiveGenesis() + 1
+	if applied > types.LayerID(trtl.cfg.WindowSize) {
+		window := applied - types.LayerID(trtl.cfg.WindowSize)
+		window = window.GetEpoch().FirstLayer() // windback to the start of the epoch to load ref ballots
+		prev, err1 := layers.GetAggregatedHash(db, window-1)
```
poszu: Could it underflow (when `window.GetEpoch()` returns epoch 0)?

tortoise/recover.go
```diff
@@ -51,42 +68,62 @@ func Recover(ctx context.Context, db *datastore.CachedDB, latest types.LayerID,
...
 	}
+	if last == 0 {
+		last = current
+	} else {
+		last = min(last, current)
+	}
+	trtl.TallyVotes(ctx, last)
+	// find topmost layer that was already applied and reset pending
+	// so that result for that layer is not returned
+	for prev := last - 1; prev >= start; prev-- {
```
poszu: Is it possible that it will run with `last == current == 0` (on a fresh network?)? If so, it would underflow and enter a very long loop.

tortoise/state.go
```diff
@@ -466,9 +470,10 @@ func decodeVotes(evicted, blid types.LayerID, base *ballotInfo, exceptions types
 			return votes{}, 0, fmt.Errorf("votes on layer %d conflict with abstain", lid)
 		}
 	}
-	if from <= evicted {
-		return votes{}, 0, fmt.Errorf("votes for a block in the layer (%d) outside the window (evicted %d)", from, evicted)
-	}
+	// FIXME(dshulyak) this needs to be ignored when recovering from disk
+	// if from <= evicted {
+	// 	return votes{}, 0, fmt.Errorf("votes for a block in the layer (%d) outside the window (evicted %d)", from, evicted)
+	// }
```
poszu, line 476: What about this? Will it be fixed in this PR or a follow up?

tortoise/recover.go
dshulyak: thanks

tortoise/state.go
dshulyak: this is a leftover, i moved it to a layer that calls decodeVotes. turns out i didn't push my latest branch :/

dshulyak: not sure what is going on. github shows this code, but i don't have it

poszu (approved):

dshulyak:
bors merge

PR #5118, opened 2023-10-02
[Merged by Bors] - find topmost applied layer and prune allready applied results

dshulyak:
in 1.2.0 i noticed one regression:
- if after restart same applied layer considered by tortoise undecided
- and mesh fails to apply layers for **any** reason, such as missing block locallyy

node will loop with 

> 2023-10-02T15:28:14.002+0200	WARN	fd68b.sync	mesh failed to process layer from sync	{"node_id": "fd68b9397572556c2f329f3e5af2faf23aef85dbbbb7e38447fae2f4ef38899f", "module": "sync", "sessionId": "29422935-68d6-47d1-87a8-02293aa181f3", "layer_id": 23104, "errmsg": "requested layer 8063 is before evicted 13102", "name": "sync"}

this change is a hotfix for that problem. after tallying votes we will scan all layers before that and find topmost layer that was applied before and considered valid by tortoise.


poszu (approved):
How about adding a test covering it to avoid regressions?

dshulyak:
bors merge

dshulyak:
bors merge

PR #5119, opened 2023-10-02
[Merged by Bors] - sync: hash resolution use atx cache

### LINKED ISSUE #5078: hash resolution implementation works ineffeciently with atx data
image

image

countvonzero:
## Motivation
Closes #5078

## Changes
use atx cache during hash resolution

### DIFF
--- a/syncer/state_syncer.go
+++ b/syncer/state_syncer.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"go.uber.org/zap/zapcore"
-	"golang.org/x/exp/maps"
 
 	"github.com/spacemeshos/go-spacemesh/common/types"
 	"github.com/spacemeshos/go-spacemesh/fetch"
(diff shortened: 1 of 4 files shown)


fasmat (approved):

syncer/state_syncer.go
```diff
@@ -284,35 +283,24 @@ func (s *Syncer) ensureMeshAgreement(
...
-			if _, ok := missing[id]; ok {
-				continue
-			}
-			hdr, _ := s.cdb.GetAtxHeader(id)
-			if hdr == nil {
-				missing[id] = struct{}{}
-				continue
-			}
-		}
+		missing := s.asCache.GetMissingActiveSet(diffLayer.GetEpoch(), ed.AtxIDs)
```
dshulyak, line 286: do you know how frequently this can be executed? there will be exclusive lock on this codepath

countvonzero: - in steady state, a node resolves hash with the same peer on the same diff hash once every hour (MaxStaleDuration)
- for each sync run (once every minute), it resolves hash diff with N peers, once per peer. N is the number of peers that has different mesh hash.

so the worst case scenario is N times per minute where N capped by number of peers.

dshulyak (approved):

countvonzero:
bors merge

PR #5126, opened 2023-10-04
[Merged by Bors] - hare3: support multiple keys

### LINKED ISSUE #5085: hare3: support multiple smeshers
the end goal is to support multiple identities in hare3 protocol

#### storing multiple smeshers and registration

`type Hare` needs to store collection of smeshers, this collection is extended with a Register(Smesher) call.

```go
type Protocol interface {
    Register(*EdSigner)
}
```

Register can't fail, so if atx doesn't exist yet it simply registers smesher in the internal collection.

in order to prevent new participants joining mid session when hare session starts, session needs to copy pointers to registered smeshers at that time and store them in different collection. 

#### eligibility and emitting messages

computing ecvrf is ~200us. ed25519 sig is ~100us.

lets assume that we target 10_000 identities. that is about 3s of on cpu time for every message. we need to parallelize that to support this amount. the efficient approach would be to create N worker goroutines managed by hare. each worker will accept batch of vrf's to compute / messages to sign and output result. this is also good change that can be submitted after initial implementation to simplify review.

we also should disable vrf / sig verification which is run when nodes Publish message. to implement check that message is emitted by our own host.ID() and do the necessary skip.

to vote on the same set of proposals, `proposals` method should be modified to not vote on the proposal if atx that belongs to any smesher is lower than the proposals base height. 

#### otherrefactoring

reference to vrf signer needs to be moved out of the legacy oracle https://github.com/spacemeshos/go-spacemesh/blob/develop/hare/eligibility/oracle.go#L75

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/5085

this change refactors original hare oracle to accept key that will be used to determine eligibility. and extends hare3 with a collection of keys that can be used to participate from the same physical node.

limitation is that ecvrf and ed signatures are computed sequentially and may cause messages to be delayed, if number of nodes are larger then 100. larger numbers are useful only if we expect it to be used with smallest possible atxs, which is by itself a problem.


hare3/hare.go
```diff
@@ -323,15 +329,32 @@ func (h *Hare) onLayer(layer types.LayerID) {
...
+	var (
+		signers []*signing.EdSigner
+		proto   = newProtocol(h.config.Committee/2 + 1)
+	)
 	h.mu.Lock()
+	// signer can't join mid session
+	signers = make([]*signing.EdSigner, 0, len(h.signers))
+	for _, signer := range h.signers {
+		signers = append(signers, signer)
+	}
```
poszu: consider using `Values` from `exp/maps`:
```suggestion
	signers = maps.Values(h.signers)
```

hare3/hare.go
```diff
@@ -354,56 +377,67 @@ func (h *Hare) onLayer(layer types.LayerID) {
...
+		current := session.proto.IterRound
+
+		start := time.Now()
+		for i := range session.signers {
+			if current.IsMessageRound() {
+				session.vrfs[i] = h.oracle.active(session.signers[i], session.beacon, session.lid, current)
+			} else {
+				session.vrfs[i] = nil
+			}
+			active = active || session.vrfs[i] != nil
```
poszu, line 418: Previously `active` was expressed as `vrf != nil`. And previously `vrf` was set fresh every for loop iteration. For example, it could theoretically be `!=nil` in iter 0 and `nil` in the next one. Now, the `active` remains `true` once it's set. Is it on purpose?

hare3/hare_test.go
```diff
@@ -275,6 +300,21 @@ func (cl *lockstepCluster) addNode(n *node) {
...
+	for i, signer := range cl.signers {
+		cl.nodes[i%len(cl.nodes)].register(signer.signer)
+	}
+}
+
+func (cl *lockstepCluster) addSigner(n int) *lockstepCluster {
+	last := len(cl.signers)
+	for i := last; i < last+n; i++ {
+		cl.signers = append(cl.signers, (&node{t: cl.t, i: i}).
+			withSigner().withAtx(cl.units.min, cl.units.max))
```
poszu: It's hard to track the brackets, how about:
```suggestion
		n := (&node{t: cl.t, i: i}).withSigner().withAtx(cl.units.min, cl.units.max)
		cl.signers = append(cl.signers, n)
```

poszu (commented):
overall LGTM, left a few comments

hare3/hare.go
dshulyak: i realized that i can simplify code a bit and drop `active` completely. code can always generate a message and decide later whether to publish it or not. but i will reset active to false for now, so that it is consistent with previous behavior

poszu (approved):

hare/eligibility/oracle.go
```diff
@@ -344,11 +338,16 @@ func (o *Oracle) CalcEligibility(
...
 func (o *Oracle) Proof(ctx context.Context, layer types.LayerID, round uint32) (types.VrfSignature, error) {
-	msg, err := o.buildVRFMessage(ctx, layer, round)
+	beacon, err := o.beacons.GetBeacon(layer.GetEpoch())
 	if err != nil {
-		return types.EmptyVrfSignature, err
+		return types.EmptyVrfSignature, fmt.Errorf("get beacon: %w", err)
 	}
-	return o.vrfSigner.Sign(msg), nil
+	return o.GenVRF(ctx, o.vrfSigner, beacon, layer, round), nil
+}
```
fasmat, line 346: `buildVRFMessage` and `GenVRF` do essentially the same. Are both needed? This could also be written as:

```go
	msg, err := o.buildVRFMessage(ctx, layer, round)
	if err != nil {
		return types.EmptyVrfSignature, err
	}
	return msg, nil
```

hare3/hare.go
```diff
@@ -195,13 +196,19 @@ type Hare struct {
...
 	sync      system.SyncStateProvider
 	patrol    *layerpatrol.LayerPatrol
 	tracer    Tracer
 }
 
+func (h *Hare) Register(signer *signing.EdSigner) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.log.Info("register signing key", zap.Stringer("node", signer.NodeID()))
+	h.signers[string(signer.PublicKey().Bytes())] = signer
```
fasmat: NIT: I think `NodeID` is more expressive here.

```suggestion
	h.signers[string(signer.NodeID().Bytes())] = signer
```

hare3/hare.go
```diff
@@ -323,15 +330,24 @@ func (h *Hare) onLayer(layer types.LayerID) {
...
 	h.patrol.SetHareInCharge(layer)
-	proto := newProtocol(h.config.Committee/2 + 1)
+
 	h.mu.Lock()
-	h.sessions[layer] = proto
+	// signer can't join mid session
+	s := &session{
+		lid:     layer,
+		beacon:  beacon,
+		signers: maps.Values(h.signers),
```
fasmat, line 339: I think this needs to be protected by a mutex in case `map.Values` and `hare.Register` are called concurrently.

EDIT: nvm I just saw it is already 🙂

hare/eligibility/oracle.go
dshulyak: it is used in code path for hare1, don't want to spend time refactoring that

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

Issue #5113, opened 2023-09-30
refactor code to use one of the ballot indexes (either by public key or atxid) 


dshulyak:
orange and red are this two indexes. both are roughly ~50MB, while ballots table size is ~370MB. thats ~13% overhead per index. 

![image](https://github.com/spacemeshos/go-spacemesh/assets/1964576/0ec73f1d-e7da-43ff-bec7-dfb168225a61)

Issue #5106, opened 2023-09-27
proposal building took too long


lrettig:
## Description

I've seen 5-6 reports of this issue from community members: a smesher that is eligible to produce a proposal fails to do so with this error:

`ERROR   46c52.proposalBuilder   proposal building took too long`

I've confirmed that they're running go-spacemesh v1.1.8 and that the state database is stored on SSD, so there's no immediate reason this should be slow. Also confirmed with users that their system clock is correct.

Some of these nodes successfully create one proposal then fail to create the next one, e.g.:

```
2023-09-26T09:55:00.424+0200    INFO    46c52.proposalBuilder   proposal created        {"node_id": "46c52d6764d315a58d93e904779a166d3cff5a984f72206e64018a08604c0cdd", "module": "proposalBuilder", "event": true, "sessionId": "afbdee96-59c9-4394-a931-56fbc21192f0", "layer_id": 21311, "proposal_id": "3fbbf89f79", "num txs": 0, "name": "proposalBuilder"}
2023-09-26T20:28:17.385+0200    INFO    46c52.proposalBuilder   proposal created        {"node_id": "46c52d6764d315a58d93e904779a166d3cff5a984f72206e64018a08604c0cdd", "module": "proposalBuilder", "event": true, "sessionId": "68596208-bcf4-4f80-b006-f9b861847d85", "layer_id": 21437, "proposal_id": "7172dc5fcd", "num txs": 0, "name": "proposalBuilder"}
2023-09-26T20:28:17.385+0200    ERROR   46c52.proposalBuilder   proposal building took too long         {"node_id": "46c52d6764d315a58d93e904779a166d3cff5a984f72206e64018a08604c0cdd", "module": "proposalBuilder", "sessionId": "68596208-bcf4-4f80-b006-f9b861847d85", "layer_id": 21437, "epoch_id": 5, "elapsed": "3m17.018414207s", "name": "proposalBuilder"}
```

Sometimes the proposal takes ~12-15 sec to create, sometimes (as in the above log) it's several minutes.

## Steps to reproduce
Unknown

## Actual Behavior
Proposal building sometimes succeeds, sometimes fails

## Expected Behavior
Proposal building should not take so long

## Environment
Please complete the following information:
 - OS: win11
 - Node Version: v1.1.8

## Additional Resources 
Log attached

[output2.txt.gz](https://github.com/spacemeshos/go-spacemesh/files/12742957/output2.txt.gz)

[...]

Issue #5087, opened 2023-09-26
tortoise: support multiple smeshers


dshulyak:
consensus state is isolated in the `tortoise` module. 
all changes related to multiple smeshers support should be in in `miner` module.

for consistent interface for registration please see https://github.com/spacemeshos/go-spacemesh/issues/5085

#### computing active set and eligibilities

activeset should be prepared only once per epoch, regardless the number of registered smeshers. 
need to be careful not to create multiple parallel readers where each one will try to prepare its own activeset.

data structure should reflect that we store only one copy of activeset. one way to implement it would be to check on startup if any of the registered smeshers already created activeset, if so then we cache that activeset. otherwise code prepares activeset before going into parallel part.

the other non-parallel part is computing tortoise.EncodeVotes only once per layer, regardless of the registered smeshers.

the parallel part consists of preparing eligibility cache per smesher, selecting txs for proposals and signing transactions.

#### oracle refactoring

oracle should not store any state. no signers, no eligibilities.
all shared state should be stored on miner instance, the structure should reflect what we compute in parallel and what not.

#### skipping vrf/sig validation on self-publish

unnecessary work, without this refactoring will require parallelization as well

PR #5130, opened 2023-10-05
[Merged by Bors] - miner: refactor code for multiple smesher and better observability

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/5113 

it drops RefBallot function, and will allow to drop that index in a followup

related: https://github.com/spacemeshos/go-spacemesh/issues/5106 

it eliminates repetitive disk reads and makes potentially expensive calls more transparent. added latency of execution to the logs

related: https://github.com/spacemeshos/go-spacemesh/issues/5087 

refactoring to draw a line between per-smesher data that needs to be loaded once per epoch, and calls to external components. tortoise/mesh hash are reusable for every smesher, get txs is not reusable. it makes adding support for multiple smeshers significantly simpler


dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors cancel

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

miner/proposal_builder.go
```diff
@@ -2,62 +2,133 @@
...
+		encoder.AddInt("eligible", len(s.eligibilities.proofs))
+		encoder.AddArray("eligible by layer", log.ArrayMarshalerFunc(func(encoder log.ArrayEncoder) error {
+			// Sort the layer map to log the layer data in order
+			keys := make([]types.LayerID, 0, len(s.eligibilities.proofs))
+			for k := range s.eligibilities.proofs {
+				keys = append(keys, k)
+			}
+			sort.Slice(keys, func(i, j int) bool {
+				return keys[i].Before(keys[j])
+			})
```
poszu: How about:
```suggestion
			keys := maps.Keys(s.eligibilities.proofs)
			slices.Sort(keys)
```
[slices.Sort](https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230905200255-921286631fa9/slices#Sort) sorts in ascending order.

miner/proposal_builder.go
```diff
@@ -374,137 +335,370 @@ func (pb *ProposalBuilder) decideMeshHash(ctx context.Context, current types.Lay
...
+		pb.session.beacon = beacon
+	}
+	if pb.session.prev == 0 {
+		prev, err := ballots.LastInEpoch(pb.cdb, pb.session.atx, pb.session.epoch)
+		if err != nil && !errors.Is(err, sql.ErrNotFound) {
+			return err
+		}
+		if err == nil {
+			pb.session.prev = prev.Layer
+		}
```
poszu: NIT: This is more readable IMO:
```suggestion
		switch {
		case err == nil:
			pb.session.prev = prev.Layer
		case errors.Is(err, sql.ErrNotFound):
		default:
			return err
		}
```

miner/proposal_builder.go
```diff
@@ -374,137 +335,370 @@ func (pb *ProposalBuilder) decideMeshHash(ctx context.Context, current types.Lay
...
+			pb.session.active.weight = weight
+			pb.session.eligibilities.slots = ballot.EpochData.EligibilityCount
 		}
-		return err
 	}
-
-	if beacon, err = pb.beaconProvider.GetBeacon(epoch); err != nil {
-		return errNoBeacon
+	if pb.session.eligibilities.proofs == nil {
+		pb.session.eligibilities.proofs = calcEligibilityProofs(pb.signer.MustVRFSigner(), pb.session.epoch, pb.session.beacon, pb.session.nonce, pb.session.eligibilities.slots, pb.cfg.layersPerEpoch)
```
poszu: Could you please reformat the code to avoid 200+ column long lines?

There is a tool which I use in the poet repo that does it automatically and can be setup in CI: https://github.com/segmentio/golines

miner/proposal_builder.go
```diff
@@ -374,137 +335,370 @@ func (pb *ProposalBuilder) decideMeshHash(ctx context.Context, current types.Lay
...
+		return fmt.Errorf("layer %d was already built", lid)
 	}
-	proofs := epochEligibility.Proofs[layerID]
+	pb.session.prev = lid
+
+	proofs := pb.session.eligibilities.proofs[lid]
 	if len(proofs) == 0 {
-		pb.logger.WithContext(ctx).With().Debug("not eligible for proposal in layer", layerID)
+		pb.logger.WithContext(ctx).With().Debug("not eligible for proposal in layer",
+			log.Uint32("lid", lid.Uint32()), log.Uint32("epoch", lid.GetEpoch().Uint32()))
```
poszu:
```suggestion
		pb.logger.With().Debug("not eligible for proposal in layer",
			log.Context(ctx), log.Uint32("lid", lid.Uint32()), log.Uint32("epoch", lid.GetEpoch().Uint32()))
```

miner/proposal_builder.go
```diff
@@ -374,137 +335,370 @@ func (pb *ProposalBuilder) decideMeshHash(ctx context.Context, current types.Lay
...
+			return 0, nil, err
+		}
+		totalWeight += atx.GetWeight()
+	}
+	if !ownIncluded {
+		// miner is not included in the active set derived from the epoch's first block
+		set = append(set, ownAtx)
+		totalWeight += ownWeight
 	}
+	return totalWeight, set, nil
```
poszu, line 605: It could be simpler if `ActiveSetFromEpochFirstBlock` returned a set (a `map[types.ATXID]struct{}`) - you could just insert `ownAtx` w/o checking if its already there.

miner/proposal_builder.go
```diff
@@ -374,137 +335,370 @@ func (pb *ProposalBuilder) decideMeshHash(ctx context.Context, current types.Lay
...
+	} else {
+		logger.With().Info("active set selected for proposal using grades",
+			log.Int("num atx", len(set)),
+			log.Int("num omitted", numOmitted),
+			log.Int("min atx good pct", goodAtxPercent),
+		)
+	}
+	sort.Slice(set, func(i, j int) bool {
+		return bytes.Compare(set[i].Bytes(), set[j].Bytes()) < 0
 	})
```
poszu, line 673: Why does it need to be sorted?

proposals/util.go
```diff
@@ -13,6 +11,14 @@ var (
 	GetNumEligibleSlots = util.GetNumEligibleSlots
 )
 
+func MustGeetNumEligibleSlots(weight, minWeight, totalWeight uint64, committeeSize, layersPerEpoch uint32) uint32 {
```
poszu: Typo
```suggestion
func MustGetNumEligibleSlots(weight, minWeight, totalWeight uint64, committeeSize, layersPerEpoch uint32) uint32 {
```

miner/proposal_builder.go
```diff
@@ -144,175 +195,85 @@ func WithNetworkDelay(delay time.Duration) Opt {
...
-		signer:         signer,
-		clock:          clock,
-		cdb:            cdb,
-		publisher:      publisher,
-		tortoise:       trtl,
-		beaconProvider: beaconProvider,
-		syncer:         syncer,
-		conState:       conState,
+		logger:    log.NewNop(),
+		ctx:       ctx,
```
poszu: Why keep the context in a struct instead of accepting a context from the caller in `Start()` method? It would stop when the passed context is canceled (app shutdown).

The `Stop()` could become
```go
func (pb *ProposalBuilder) WaitToStop() {
	_ = pb.eg.Wait()
}
```

---

I prefer to make Start/Run methods blocking and offload putting it onto a goroutine on the caller but this requires more refactoring of the existing codebase. Something like:
```go
var eg errgroup.Group
eg.Go(func() error { return proposalBuilder.Start(ctx) }
// start other "services"
eg.Go(func() error { return other.Start(ctx) }
...
eg.Wait() 
```

- `eg.Wait()` blocks waiting for all "services" to complete, there is no need to call Stop(), Close() etc.
- no need to store `ctx`, `cancel` and `eg` in service structs. `eg` is a local variable

WDYT?

miner/proposal_builder.go
dshulyak: i am changing this in a followup, as i will not forcefully include my own atx. wanted to keep behavior the same as before in this pr

miner/proposal_builder.go
dshulyak: activeset later hashed and referenced as in ballot/proposal . we want to keep number of unique activesets low, as otherwise they bloat network and state

miner/proposal_builder.go
dshulyak: i didn't change as it was old code and didn't seem very important but in general i like two approaches:
- Run(ctx) the way you described
- Start() / Stop() and ctx initialized locally , so that lifecycle is managed by this object without external dependencies

Start/Stop has an "advantage" that it can be aware of specific order for Stop. will refactor to use Run(ctx)

miner/proposal_builder.go
dshulyak: it annoys me too but i forget sometimes. will try to automate this

dshulyak:
@poszu thanks, updated

poszu (approved):

dshulyak:
bors merge

miner/proposal_builder.go
fasmat: A bit late to the discussion: I tend to favor the second approach for services and the first one for tasks / requests that are expected to return before the application ends. The reason is the following:

```go
svc := NewService(...)
err := svc.Start() // or already implicit with NewService
err := svc.Stop() // Stop -> Start can be called again; otherwise Close() -> NewService has to be called again

// errs can be easily handled
```

is much easier to use than

```go
svc := NewService(...)
var eg errgroup.Group
eg.Go(func() {
    return eg.Run(ctx) // or don't return the error? since if we do the errgroup cannot be used to start new routines any more
})
err := eg.Wait() // how to handle error(s) here? the err returned can be from any go routine in the errgroup

// wait blocks so this call has to be done again from a different go routine
cancel() // stops all services, if individual services need to be stopped each one needs its own context
```

I agree the service should not have a `context.Context` as a field. In the previous version of the code with `Start` a simple `chan struct{}` that is closed on `proposalBuilder.Close()` would have sufficed for the purpose it was used.

dshulyak: > I agree the service should not have a context.Context as a field

the advantage of storing context in Start/Stop approach is that context can be passed down the stack. so if some method makes a network request or generally stoppable we can pass context as is. with a struct channel there will be a need for additional context

fasmat: From the official documentation (https://pkg.go.dev/context):

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

For the previous design I would have probably done something like the following:

```go
func (pb *proposalBuilder) Start() error {
    if pb.stop != nil {
        return errors.New("service already started")
    }

    ctx, cancel := context.WithCancel(context.Background)
    pb.stop = cancel
    pb.eg.Go(func() {
       return pb.longRunningTask1(ctx, ...)
    })
    pb.eg.Go(func() {
       return pb.longRunningTask2(ctx, ...)
    })
   return nil
}

func (pb *proposalBuilder) Close() error {
    pb.stop()
    return pb.eg.Wait()
}
```

This way only a reference to `context.CancelFunc` is held.

poszu: storing a context in struct leads to code that is difficult to reason about like:
```go
func (f *Foo) doSomething(ctx context.Context) error {
	// do something...
	select {
	case <-ctx.Done():
	case <-f.ctx.Done():
		return f.ctx.Err()
	}
	return nil
}
```

PR #5135, opened 2023-10-09
[Merged by Bors] - miner: multi signers

### LINKED ISSUE #5087: tortoise: support multiple smeshers

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/5087

data is separated into shared data (beacon and active set) and signer specific data. both beacon and activeset are used from shared data, until smeshers generated a reference ballot. once any smesher generated a ballot it will be using data recorded in the reference ballot. 

build method now loops over all signers (copied at the start of the layer). there are parts that have to be run once for all signers and parts that makes sense to run in parallel.

serial parts:
- loading share data (beacon and active set)
- deciding on mesh hash
- tally votes & encode votes tortoise calls

parallel parts:
- loading data (it can be also run serially, but it was convenient to run it in parallel)
- computing eligibilities (this is done once per node startup)
- selecting txs
- publishing proposal. this is the most important to avoid blocking serially in Publish while it runs validation

worker pool (errgroup) is limited by number of cores as there is no network requests during parallel work.


miner/proposal_builder.go
```diff
@@ -68,23 +71,37 @@ type ProposalBuilder struct {
 	tortoise  votesEncoder
 	syncer    system.SyncStateProvider
 
+	mu      sync.Mutex
+	signers map[string]*signerSession
```
poszu: Why not use `types.NodeID` as the key? It's much easier/more readable to use later and expresses the intent of mapping clearly.
```suggestion
	signers map[types.NodeID]*signerSession
```

miner/proposal_builder.go
```diff
@@ -244,7 +293,7 @@ func (pb *ProposalBuilder) Run(ctx context.Context) error {
 			if current <= types.GetEffectiveGenesis() || !pb.syncer.IsSynced(sctx) {
 				continue
 			}
-			if err := pb.build(ctx, current); err != nil {
+			if err := pb.build(sctx, current); err != nil {
```
poszu: Perhaps instead consider shadowing `ctx` by changing: `sctx := log.WithNewSessionID(ctx)` :point_right: `ctx := log.WithNewSessionID(ctx)`? Then this kind of bug wouldn't be possible.

miner/proposal_builder.go
```diff
@@ -317,141 +366,174 @@ func (pb *ProposalBuilder) decideMeshHash(ctx context.Context, current types.Lay
...
+	var eg errgroup.Group
+	eg.SetLimit(pb.cfg.workersLimit)
+	for _, ss := range signers {
+		ss := ss
+		ss.latency.start = start
+		eg.Go(func() error {
+			if err := pb.initSignerData(ctx, ss, lid); err != nil {
+				if errors.Is(err, errAtxNotAvailable) {
+					ss.log.With().Info("smesher doesn't have atx that targets this epoch",
+						log.Context(ctx), log.Uint32("epoch", ss.session.epoch.Uint32()),
```
poszu: types.EpochID implements `Field()` already:
```suggestion
						log.Context(ctx), ss.session.epoch,
```

miner/proposal_builder.go
```diff
@@ -317,141 +366,174 @@ func (pb *ProposalBuilder) decideMeshHash(ctx context.Context, current types.Lay
...
+	for _, ss := range signers {
+		if n := len(ss.session.eligibilities.proofs[lid]); n == 0 {
+			ss.log.With().Debug("not eligible for proposal in layer",
+				log.Context(ctx),
+				log.Uint32("lid", lid.Uint32()), log.Uint32("epoch", lid.GetEpoch().Uint32()))
+			continue
+		} else {
+			ss.log.With().Debug("eligible for proposals in layer",
+				log.Context(ctx),
+				log.Uint32("lid", lid.Uint32()), log.Int("num proposals", n),
```
poszu: types.LayerID implements `Field()` already (also in other places):
```suggestion
				lid, log.Int("num proposals", n),
```

miner/proposal_builder.go
poszu: Is it required to persist the session here? Wouldn't it suffice to persist `signing.EdSigner` in here and create a fresh session for each signer in the `build()` method?

poszu (approved):

miner/proposal_builder.go
dshulyak: those fields should be deprecated. they can't be passed as is to zap. and because of that every log method needs to copy whole collection and "unpack" fields

miner/proposal_builder.go
dshulyak: it will query database for various data repeatedly on every layer otherwise. it is better not to, if we can store a bit of data in memory

miner/proposal_builder.go
dshulyak: i understood what you meant, changed

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #5145, opened 2023-10-11
[Merged by Bors] - Refactor gRPC PostService to be injected into services

fasmat:
## Motivation
Follow up on https://github.com/spacemeshos/go-spacemesh/pull/5061#discussion_r1350788801

This changes the `activation.Builder` and `activation.NIPostBuilder` to use `grpcservice.PostService` directly instead of the later signalling new and dropped connections to them.

## Changes
- Inverse dependency between the three components as discussed in previous PR
- During startup the `grpcservice.PostService` is now always initialized along the other node services, but only registered in the `grpcserver` when it is configured for at least one of the available listeners.

## Test Plan
n/a

## TODO


activation/interface.go
```diff
@@ -78,6 +78,33 @@ type SmeshingProvider interface {
 	UpdatePoETServers(ctx context.Context, endpoints []string) error
 }
 
+// poetProvingServiceClient provides a gateway to a trust-less public proving service, which may serve many PoET
+// proving clients, and thus enormously reduce the cost-per-proof for PoET since each additional proof adds only
+// a small number of hash evaluations to the total cost.
+type poetProvingServiceClient interface {
```
poszu: As you are already changing it. Why not just:
```suggestion
// poetClient provides a gateway to a trust-less public proving service, which may serve many PoET
// proving clients, and thus enormously reduce the cost-per-proof for PoET since each additional proof adds only
// a small number of hash evaluations to the total cost.
type poetClient interface {
```

By the way, the description in the comment is weird and not helpful in my opinion.

node/node.go
```diff
@@ -843,9 +844,12 @@ func (app *App) initServices(ctx context.Context) error {
 		app.log.Panic("failed to create post setup manager: %v", err)
 	}
 
+	app.grpcPostService = grpcserver.NewPostService(app.log.Zap())
```
poszu: How about using a dedicated post logger
```suggestion
	app.grpcPostService = grpcserver.NewPostService(app.addLogger(PostLogger, lg).Zap())
```

poszu (approved):

fasmat:
bors merge

PR #5148, opened 2023-10-12
[Merged by Bors] - Beacon cleanups

poszu:
Some clean-ups in beacon code:
- removed redundant:
  -  nodeID (can be derived from edSigner)
  -  and metricsRegistry (unused
- use `codec.MustEncode` where an error lead to shutdown anyway
- avoid costly eager marshaling for log purposes

:bulb: It's recommended to review commit-by-commit as changes are separated.

### DIFF
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -835,12 +835,7 @@ func (pd *ProtocolDriver) sendProposal(ctx context.Context, epoch types.EpochID,
 	}
 
 	logger.With().Debug("own proposal passes threshold", log.Inline(proposal))
-	serialized, err := codec.Encode(&m)
-	if err != nil {
-		logger.With().Fatal("failed to encode beacon proposal", log.Err(err))
-	}
-
-	pd.sendToGossip(ctx, pubsub.BeaconProposalProtocol, serialized)
+	pd.sendToGossip(ctx, pubsub.BeaconProposalProtocol, codec.MustEncode(&m))
 	logger.With().Info("beacon proposal sent", log.Inline(proposal))
 }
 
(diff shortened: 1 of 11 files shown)


beacon/beacon.go
```diff
@@ -586,7 +574,8 @@ func (pd *ProtocolDriver) initEpochStateIfNotPresent(logger log.Log, epoch types
 				log.Bool("malicious", malicious),
 				log.Stringer("smesher", header.NodeID))
 		}
-		if header.NodeID == pd.nodeID {
+		if header.NodeID == pd.edSigner.NodeID() {
+			// TODO(poszu): support many IDs
```
poszu:
```suggestion
```

beacon/beacon.go
```diff
@@ -237,7 +225,7 @@ func (pd *ProtocolDriver) Start(ctx context.Context) {
 			pd.logger.Info("beacon protocol disabled")
 			return
 		}
-		pd.logger.With().Info("starting beacon protocol", log.String("config", fmt.Sprintf("%+v", pd.config)))
+		pd.logger.With().Info("starting beacon protocol", log.Any("config", pd.config))
```
fasmat, line 226: I think using `log.Any` is expensive, because zap needs to go over the object with reflection to generate a string from it. I'd prefer if `pd.config` would implement `zapcore.ObjectMarshaler` by adding a `MarshalLogObject` method.

beacon/beacon.go
```diff
@@ -776,17 +764,19 @@ func (pd *ProtocolDriver) runProtocol(ctx context.Context, epoch types.EpochID,
...
 	logger.With().Info("calculated beacon",
 		beacon,
 		log.Int("num_hashes", len(allProposals)),
-		log.String("proposals", strings.Join(allHexes, ", ")),
+		log.Array("proposals", zapcore.ArrayMarshalerFunc(func(enc zapcore.ArrayEncoder) error {
+			for _, h := range allProposals {
+				enc.AppendString(hex.EncodeToString(h[:]))
+			}
+			return nil
+		})),
```
fasmat: Adding a small method to `proposalList` would allow to log `allProposals` directly:

```go
func (hl proposalList) MarshalLogArray(enc zapcore.ArrayEncoder) error {
	for _, proposal := range hl {
		enc.AppendString(hex.EncodeToString(proposal[:]))
	}
	return nil
}
```

```suggestion
		log.Array("proposals", allProposals),
	),
```

beacon/proposal_set.go
```diff
@@ -1,5 +1,11 @@
...
 
+import (
+	"encoding/hex"
+
+	"go.uber.org/zap/zapcore"
+)
+
 type proposalSet map[Proposal]struct{}
 
 func (vs proposalSet) list() proposalList {
```
fasmat: This method can be written as:

```go
func (vs proposalSet) list() proposalList {
	return maps.Keys(vs)
}
```

beacon/beacon.go
poszu: Generally yes, but this is a `Start()` method that is executed once per application life, so IMHO it doesn't make sense to bloat code with special implementation just for this use-case.

fasmat (approved):

poszu:
Bors merge

PR #5150, opened 2023-10-13
drop ballots_by_layer_by_pubkey to reduce db size

### LINKED ISSUE #5113: refactor code to use one of the ballot indexes (either by public key or atxid)

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/5113

there are two similar indexes, each of them has ~13% overhead compared to ballots data


sql/ballots/ballots.go
```diff
@@ -169,15 +174,19 @@ func IDsInLayer(db sql.Executor, lid types.LayerID) (rst []types.BallotID, err e
 	return rst, err
 }
 
-// LayerBallotByNodeID returns any ballot by the specified NodeID in a given layer.
-func LayerBallotByNodeID(db sql.Executor, lid types.LayerID, nodeID types.NodeID) (*types.Ballot, error) {
+// LayerBallotByAtxID returns any ballot by the specified NodeID in a given layer.
```
poszu:
```suggestion
// LayerBallotByAtxID returns any ballot by the specified ATX ID in a given layer.
```

sql/ballots/ballots_test.go
```diff
@@ -123,36 +138,48 @@ func TestLatest(t *testing.T) {
...
 	for _, ballot := range ballots {
 		require.NoError(t, Add(db, &ballot))
 	}
 
-	prev, err := LayerBallotByNodeID(db, lid, nodeID1)
+	prev, err := LayerBallotByAtxID(db, lid, ballots[0].AtxID)
 	require.ErrorIs(t, err, sql.ErrNotFound)
 	require.Nil(t, prev)
-	prev, err = LayerBallotByNodeID(db, lid, nodeID2)
+	prev, err = LayerBallotByAtxID(db, lid, ballots[1].AtxID)
```
poszu: Why is the variable named "prev" instead of "ballot"?

poszu (approved):

sql/ballots/ballots_test.go
dshulyak: like previous ballot, but yes ballot is clearer

dshulyak:
bors merge

dshulyak:
it somehow changed order in some queries with mainnet data, will need to look more into this

PR #5151, opened 2023-10-13
[Merged by Bors] - sync: enable rate limiting for servers

### LINKED ISSUE #4603: sync: limit resources spent on serving data to peers
## Description
currently it's trivial to dos an honest node by repeatedly asking the node to serve data via the sync protocol.

we need to 
- keep track of the resources spent on a peer
- rate-limit once a peer reached a configured max value

### LINKED ISSUE #4977: optimize atx sync codepath

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/4977
closes: https://github.com/spacemeshos/go-spacemesh/issues/4603

this change introduces two configuration parameter for every server:
- requests per interval pace, for example 10 req/s, this caps the maximum bandwidth that every server can use
- queue size, it is set to serve requests within expected latency. every other request is dropped immediately so that client can retry with different node. currently the timeout is set to 10s, so the queue should be roughly 10 times larger then rps

it doesn't provide global limit for bandwidth, but we have limit for the number of peers. and honest peer doesn't run many concurrent queries. so what we really want to handle is peers with intentionally malicious behavior, but thats not a pressing issue 

example configuration:

```json
"fetch": {
        "servers": {
            "ax/1": {"queue": 10, "requests": 1, "interval": "1s"},
            "ld/1": {"queue": 1000, "requests": 100, "interval": "1s"},
            "hs/1": {"queue": 2000, "requests": 200, "interval": "1s"},
            "mh/1": {"queue": 1000, "requests": 100, "interval": "1s"},
            "ml/1": {"queue": 100, "requests": 10, "interval": "1s"},
            "lp/2": {"queue": 10000, "requests": 1000, "interval": "1s"}
        }
    }
```

https://github.com/spacemeshos/go-spacemesh/blob/3cf02146bf27f53c001bffcacffbda05933c27c4/fetch/fetch.go#L130-L144

metrics are per server:

https://github.com/spacemeshos/go-spacemesh/blob/3cf02146bf27f53c001bffcacffbda05933c27c4/p2p/server/metrics.go#L15-L52

have to be enabled for all servers with

```json
"fetch": {
        "servers-metrics": true
    }
```


dshulyak:
bors try

fetch/fetch_test.go
```diff
@@ -373,7 +377,12 @@ func TestFetch_PeerDroppedWhenMessageResultsInValidationReject(t *testing.T) {
 		}
 		return result, nil
 	}
-	server.New(badPeerHost, hashProtocol, badPeerHandler)
+	badsrv := server.New(badPeerHost, hashProtocol, badPeerHandler)
+	var eg errgroup.Group
+	eg.Go(func() error {
+		badsrv.Run(ctx)
+		return nil
+	})
```
fasmat, line 385: This `errgroup.Group` isn't awaited?

p2p/server/server.go
```diff
@@ -73,36 +99,91 @@ type Host interface {
...
+			}
+			stream.Close()
+		}
+	})
+
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.SetLimit(s.queueSize)
+	for {
+		select {
+		case <-ctx.Done():
```
fasmat, line 166: This `ctx` is not the `ctx` passed to `Run` but the context created from `errgroup.WithContext` which should only be used by go routines managed by `errgroup`.

p2p/server/server.go
```diff
@@ -73,36 +99,91 @@ type Host interface {
...
+			eg.Wait()
+			return nil
+		case req := <-queue:
+			if err := limit.Wait(ctx); err != nil {
```
fasmat, line 170: This `ctx` is not the `ctx` passed to `Run` but the context created from `errgroup.WithContext` which should only be used by go routines managed by `errgroup`.

p2p/server/server.go
```diff
@@ -147,26 +228,51 @@ func (s *Server) streamHandler(stream network.Stream) {
...
 }
 
+func (s *Server) trackClientLatency(start time.Time, err error) {
+	if s.metrics != nil {
+		if err != nil {
+			s.metrics.clientLatencyFailure.Observe(time.Since(start).Seconds())
+		} else {
+			s.metrics.clientLatency.Observe(time.Since(start).Seconds())
+		}
+	}
```
fasmat: NIT:

```suggestion
	switch {
	case s.metrics == nil:
		return
	case err != nil:
		s.metrics.clientLatencyFailure.Observe(time.Since(start).Seconds())
	default:
		s.metrics.clientLatency.Observe(time.Since(start).Seconds())
	}
```

p2p/server/server.go
```diff
@@ -147,26 +228,51 @@ func (s *Server) streamHandler(stream network.Stream) {
...
+	failure func(error),
+) error {
+	start := time.Now()
 	if len(req) > s.requestLimit {
 		return fmt.Errorf("request length (%d) is longer than limit %d", len(req), s.requestLimit)
 	}
 	if s.h.Network().Connectedness(pid) != network.Connected {
 		return fmt.Errorf("%w: %s", ErrNotConnected, pid)
 	}
 	go func() {
```
fasmat, line 247: Should this also be part of an `errgroup.Group` and be awaited?

p2p/server/server.go
```diff
@@ -147,26 +228,51 @@ func (s *Server) streamHandler(stream network.Stream) {
...
-		start := time.Now()
+		var err error
 		defer func() {
 			s.logger.WithContext(ctx).With().Debug("request execution time",
 				log.String("protocol", s.protocol),
 				log.Duration("duration", time.Since(start)),
+				log.Err(err),
 			)
+			s.trackClientLatency(start, err)
 		}()
```
fasmat: With an errgroup this could be just what comes after the `eg.Wait()`:

```go
eg.Go(func() {
	// most of the code in `go func() { ... }()`
})
err := eg.Wait()
s.logger.WithContext(ctx).With().Debug("request execution time",
	log.String("protocol", s.protocol),
	log.Duration("duration", time.Since(start)),
	log.Err(err),
)
s.trackClientLatency(start, err)
```

p2p/server/server_test.go
```diff
@@ -32,12 +34,23 @@ func TestServer(t *testing.T) {
...
+	eg.Go(func() error {
+		return srv1.Run(ctx)
+	})
+	eg.Go(func() error {
+		return srv2.Run(ctx)
+	})
+	t.Cleanup(func() {
+		cancel()
+		eg.Wait()
+	})
```
fasmat: I don't see how a blocking `Run` method is more convenient to use than something like this:

```suggestion
	require.NoError(t, srv1.Start())
	t.Cleanup(t, func { assert.NoError(t, srv1.Stop()) })
	require.NoError(t, srv2.Start())
	t.Cleanup(t, func { assert.NoError(t, srv2.Stop()) })
```

p2p/server/server.go
```diff
@@ -147,26 +228,51 @@ func (s *Server) streamHandler(stream network.Stream) {
...
 // Request sends a binary request to the peer. Request is executed in the background, one of the callbacks
 // is guaranteed to be called on success/error.
-func (s *Server) Request(ctx context.Context, pid peer.ID, req []byte, resp func([]byte), failure func(error)) error {
+func (s *Server) Request(
+	ctx context.Context,
+	pid peer.ID,
+	req []byte,
+	resp func([]byte),
+	failure func(error),
+) error {
```
fasmat, line 239: I just noticed `Request` isn't blocking at the moment. That seems incorrect to me. I would assume that I can discard the `ctx` after a function that receives it returns. Imagine the following code:

```go
func myFunc(ctx context.Context, ...) error {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	return srv.Request(ctx, ...)
}
```

This will aways fail immediately with no error returned and not after 5 seconds with `context.DeadlineExceeded` as how I would expect.

p2p/server/server.go
fasmat: Alternatively we could even get rid of the `go func()` completely and just make `Request` blocking:

```go
func (s *Server) Request(ctx context.Context, pid peer.ID, req []byte) (resp []byte, err error)
```

p2p/server/server.go
```diff
@@ -188,18 +294,21 @@ func (s *Server) Request(ctx context.Context, pid peer.ID, req []byte, resp func
 			failure(err)
 			return
 		}
-		if err := wr.Flush(); err != nil {
+		err = wr.Flush()
+		if err != nil {
```
fasmat: I don't understand this and the change below? The behaviour is still the same as before?

p2p/server/server.go
dshulyak: i need to recall why it did that, maybe for no reason.

but what is wrong with using errgroup context?

p2p/server/server.go
dshulyak: i am not sure how switch makes it better

p2p/server/server.go
dshulyak: maybe but it is not related to the change

p2p/server/server.go
dshulyak: making request blocking is not a part of this change. note that p2p/server should exist at all, and requests should be made over libp2p streams

p2p/server/server_test.go
dshulyak: it is convenient to integrate in the App. additionally if any of the Run fails with error, all other runs should fail as well, it is not clear how to achieve the same with Start/Stop

consider several components that are started with Run(ctx) error, if any of them fails app should terminate all and exit with fatal. with Run interface this is very clean

p2p/server/server.go
dshulyak: this is not a part of this change. this is legacy code that in generally should be written differently to make good use of libp2p streams

fetch/fetch_test.go
dshulyak: added await

p2p/server/server.go
dshulyak: actually i don't see why i need that shared context, maybe i removed something after i wrote that

p2p/server/server.go
dshulyak: the change is to make `err` visible in defer. the behavior is the same

fasmat: I feel like the code would be easier to follow if instead of the `defer` the body of the `go func` would be its own method:

```go
go func() {
	err := s.streamData(...)
	s.logger.WithContext(ctx).With().Debug("request execution time",
		log.String("protocol", s.protocol),
		log.Duration("duration", time.Since(start)),
		log.Err(err),
	)
	s.trackClientLatency(start, err)
}()
```

p2p/server/server.go
fasmat: Is there an issue tracking the need to change this component? Holding on to parameters after a function returns can lead to problems and data races which are to debug later.

dshulyak: there is 1st point in this issue https://github.com/spacemeshos/pm/issues/272

p2p/server/server.go
dshulyak: yes you are right, i will change

fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors try

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

PR #5153, opened 2023-10-13
[Merged by Bors] - reduce info log spam in mesh, ballots, atxs modules

dshulyak:
change log levels to warn for:
- mesh
- proposals listener
- atx listener

we should keep it as info on some nodes in cloud to see minimally useful info


poszu (approved):

dshulyak:
bors merge

PR #5154, opened 2023-10-13
[Merged by Bors] - Add mTLS grpcserver

### LINKED ISSUE #5131: gRPC: add mTLS authentication to post service
## Description

The post service is able to use mTLS authentication and the node should expose a dedicated gRPC endpoint that requires authentication for the post service to connect to.

## Acceptance criteria

- Additionally to `PublicListener` and `PrivateListener` a new endpoint `AuthenticatedListener` is added
- `AuthenticatedListener` expects clients to connect with `mTLS` Authentication
  - only clients with known certificates are allowed to connect
- The nodes key and certificate as well as certificates for post services that can connect to the node are read from the nodes local configuration
  - Preferably from a directory on disk instead of from the database or configuration, but make the directory with the certificates configurable
- An API call is added to the `PrivateListener` that can be used to trigger the node to reload known certificates from disk
- In standalone mode the supervisor for the post service generates the keys & certificates for both the node and the post service.
  - It triggers the node to load the certificates from disk after generating them and starts the post service with its generated key & certificate

fasmat:
## Motivation
Closes #5131

do not merge before https://github.com/spacemeshos/api/pull/268 and https://github.com/spacemeshos/post/pull/245

## Changes
- setup for gRPC servers has been moved from node startup into the `grpcserver` package
- `NewPublic`, `NewPrivate` and `NewTLS` create servers for the given purposes based on the configuration passed to them
- replaced more instances of `go-spacemesh/log` with `zap`

## Test Plan
- existing tests pass

## TODO


CHANGELOG.md
```diff
@@ -18,10 +18,44 @@ See [RELEASE](./RELEASE.md) for workflow instructions.
...
   > 2023-10-02T15:28:14.002+0200 WARN fd68b.sync mesh failed to process layer from sync {"node_id": "fd68b9397572556c2f329f3e5af2faf23aef85dbbbb7e38447fae2f4ef38899f", "module": "sync", "sessionId": "29422935-68d6-47d1-87a8-02293aa181f3", "layer_id": 23104, "errmsg": "requested layer 8063 is before evicted 13102", "name": "sync"}
 
-* [#5091](https://github.com/spacemeshos/go-spacemesh/pull/5091) First stage of separating PoST from the node into its own service.
+* [#5091](https://github.com/spacemeshos/go-spacemesh/pull/5091) Separating PoST from the node into its own service.
 * [#5061](https://github.com/spacemeshos/go-spacemesh/pull/5061) Proof generation is now done via a dedicated service instead of the node.
-
-  Operating a node doesn't require any changes at the moment. The service will be automatically started by the node if needed and will be stopped when the node is stopped.
+* [#5154](https://github.com/spacemeshos/go-spacemesh/pull/5154) Enable TLS connections between node and PoST service.
+
+  PoST proofs are now done via a dedicated process / service that the node communicates with via gRPC. Smapp users can continue to smash as they are used to. The node will
```
poszu:
```suggestion
  PoST proofs are now done via a dedicated process / service that the node communicates with via gRPC. Smapp users can continue to smesh as they used to. The node will
```

CHANGELOG.md
```diff
@@ -18,10 +18,44 @@ See [RELEASE](./RELEASE.md) for workflow instructions.
...
+    "post-opts-node-address": "http://domain:port",      // defaults to 127.0.0.1:9093 - the same default value as for "grpc-private-listener"
+    
+    // the following settings are optional but allow to expose the post service publicly and connect via TLS if needed
+    "post-opts-tls-ca-cert": "/path/to/ca.pem",  // CA certificate that signed the node's and the PoST service's certificates
+    "post-opts-tls-cert": "/path/to/cert.pem",   // certificate for the PoST service
+    "post-opts-tls-key": "/path/to/key.pem",     // private key for the PoST service
+  }
+  ```
+
+  By default the node listens for the PoST service on `grpc-private-listener` (defaults to 127.0.0.1:9093). This endpoint is not require authentication and
```
poszu:
```suggestion
  By default the node listens for the PoST service on `grpc-private-listener` (defaults to 127.0.0.1:9093). This endpoint does not require authentication and
```

CHANGELOG.md
```diff
@@ -18,10 +18,44 @@ See [RELEASE](./RELEASE.md) for workflow instructions.
...
+  To setup TLS-secured public connections the API config has been extended with the following options:
+
+  ```json
+  "api": {
+    "grpc-tls-services": [],               // list of services that should be exposed via TLS
+    "grpc-tls-listener": "0.0.0.0:9094",   // address to listen on for TLS connections
+    "gprc-tls-ca-cert": "/path/to/ca.pem", // CA certificate that signed the node's and the PoST service's certificates
+    "grpc-tls-cert": "/path/to/cert.pem",  // certificate for the node
+    "grpc-tls-key": "/path/to/key.pem",    // private key for the node
+  }
```
poszu: IMHO, this should be placed in the README. The changelog should only inform about the new feature and point to the readme for configuration guide.

activation/post_supervisor.go
```diff
@@ -47,8 +47,11 @@ func DefaultTestPostServiceConfig() PostSupervisorConfig {
 
 type PostSupervisorConfig struct {
 	PostServiceCmd string `mapstructure:"post-opts-post-service"`
+	NodeAddress    string `mapstructure:"post-opts-node-address"`
 
-	NodeAddress string `mapstructure:"post-opts-node-address"`
+	CACert string `mapstructure:"post-opts-ca-cert"`
+	Cert   string `mapstructure:"post-opts-cert"`
+	Key    string `mapstructure:"post-opts-key"`
```
poszu, line 54: The post-service also accepts `domain` for which the certificate is valid for with `--domain`

api/grpcserver/http_server.go
```diff
@@ -90,19 +89,21 @@ func (s *JSONHTTPServer) StartService(
 			err = pb.RegisterTransactionServiceHandlerServer(ctx, mux, typed)
 		case *DebugService:
 			err = pb.RegisterDebugServiceHandlerServer(ctx, mux, typed)
+		default:
+			s.logger.Error("unsupported service", zap.Stringer("service", svc))
```
poszu: This could also be a `s.logger.Panic` to catch it early and not let is slip as this indicates a logic error.

api/grpcserver/http_server.go
```diff
@@ -90,19 +89,21 @@ func (s *JSONHTTPServer) StartService(
...
 	}
 
-	s.logger.With().Info("starting grpc gateway server", log.String("address", s.listener))
+	s.logger.Info("starting grpc gateway server", zap.String("address", s.listener))
 
 	lis, err := net.Listen("tcp", s.listener)
 	if err != nil {
-		s.logger.Error("error listening: %v", err)
+		s.logger.Error("start listen server", zap.Error(err))
 		return err
```
poszu: How about:
```suggestion
		return fmt.Errorf("listening on %s: %w", s.listener, err)
```

and let the caller do the logging. Similarly in other places. Reasoning: IMHO, the error should be logged in the place where it is handled, otherwise there is a high chance the same error will be logged multiple times. Bubble-ups should only add useful context for identifying the issue.

CHANGELOG.md
fasmat: Expanded description and moved to README.md

api/grpcserver/http_server.go
fasmat: I'm not sure about this. Before trying to expose a service that is not a `case` in this `switch` would just silently be ignored. Now instead panicing seems a drastic change 🤔

api/grpcserver/http_server.go
fasmat: While I do agree the logging message here is the only thing that happens when the JSON server fails to start, the caller (`node.go`) ignores any error that is returned from this method.

I changed to instead log in `node.go`, again not adding a panic, since going from completely ignoring a failing configuration to the node failing to start could break setups that might already be in place.

api/grpcserver/http_server.go
poszu: It was a mistake in my opinion. Hitting `default` case is clearly a programming mistake and a bug to fix. The earlier it crashes, the less time one will spend debugging.

api/grpcserver/http_server.go
poszu: >going from completely ignoring a failing configuration to the node failing to start could break setups that might already be in place.

These setups are already broken but don't know it :D.

activation/post_supervisor.go
fasmat: Version `v0.5.0-alpha2` does not yet have this CLI parameter, that's why I didn't include it and upgrading to `v0.5.0-alpha3` isn't possible yet because it fails to compile in with `go-spacemeshos/post`.

We can update this section when `v0.5.0` of `post-rs` is released (together with an update of `go-spacemeshos/post`).

api/grpcserver/http_server.go
fasmat: Since the default configuration would lead to a panic on startup if the `JSONListener` is set I updated the `Service` interface to include a register method for the `JSONServer` and updated all existing services to implement this interface.

Now this error cannot occur any more and the `switch` doesn't need to be updated every time a new service is added (just the service needs to implement the interface).

api/grpcserver/http_server.go
fasmat: With my latest commit all services have been updated to be usable with the JSONServer and a failing start now results in the node panicing and shutting down.

poszu (approved):

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

PR #5155, opened 2023-10-13
[Merged by Bors] - Simplify NewEdVerifier() to not return useless error

poszu:
## Motivation
`signing.NewEdVierifier()` is infallible, it never really returns an error and its definition could be simplfied.

## Changes
Changed `signing.NewEdVierifier()` to not return an error.


signing/verifier.go
```diff
@@ -47,3 +43,8 @@ func (es *EdVerifier) Verify(d Domain, nodeID types.NodeID, m []byte, sig types.
 	msg = append(msg, m...)
 	return ed25519.Verify(nodeID[:], msg, sig[:])
 }
+
+// Shortcut to verify with the default EdVerifier.
+func Verify(d Domain, nodeID types.NodeID, m []byte, sig types.EdSignature) bool {
+	return NewEdVerifier().Verify(d, nodeID, m, sig)
+}
```
fasmat: I think this will have limited usability. Outside tests all signatures (should) be signed with a prefix and this function doesn't allow to provide one.

I would even lean towards not providing this shortcut:

```go
signing.Verify(d, nodeID, m, sig)
```

is only minimally shorter than:

```go
signing.NewEdVerifier().Verify(d, nodeID, m, sig)
```

but having it might lead to accidentally using it somewhere in a service.

poszu: `signing.NewEdVerifier().Verify()` could also be mistakenly used :shrug: . If a verifier w/o a prefix is not valid, the prefix should be required, not optional.

fasmat (approved):

poszu:
bors merge

dshulyak (approved):

poszu:
bors merge

poszu:
Flaky UT (#5156).

bors merge

PR #5158, opened 2023-10-15
[Merged by Bors] - Support multiple signers in beacon

### LINKED ISSUE #5088: beacon: support multiple smeshers
please see https://github.com/spacemeshos/go-spacemesh/issues/5085 for:
- consistent interface
- parallelization considerations
- validation on self-publishing optimization

one another bottleneck that we need to be careful about is not to emit all messages at once. basically [jitter](https://github.com/spacemeshos/go-spacemesh/issues/4989) needs to be considered for every identity separately. one working option to implement it would be to use time.AfterFunc

poszu:
## Motivation
Part of https://github.com/spacemeshos/pm/issues/261
Closes #5088

## Changes
- beacon protocol driver supports registering new keys to participate in the protocol
- when a new epoch starts, the protocol selects active signers that will participate
- state is changed to contain the set of active protocol participants. A participant session consists of:
  * signer
  * VRF nonce
- weakcoin supports multiple participants. Each one publishes the same proposal.

## Test Plan
Existing unit-tests were extended to run beacon and weakcoin with multiple signers per node.

### DIFF
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ALTree/bigfloat"
 	"github.com/spacemeshos/fixed"
+	"go.uber.org/zap/zapcore"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/spacemeshos/go-spacemesh/beacon/metrics"
@@ -153,6 +154,12 @@ type participant struct {
 
 func (s *participant) Id() log.Field {
 	return log.ShortStringer("id", s.signer.NodeID())
+}
+
+func (s participant) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	enc.AddString("id", s.signer.NodeID().ShortString())
+	enc.AddUint64("nonce", uint64(s.nonce))
+	return nil
 }
 
 // ProtocolDriver is the driver for the beacon protocol.
@@ -602,6 +609,17 @@ func (pd *ProtocolDriver) initEpochStateIfNotPresent(logger log.Log, epoch types
 		}
 	}
 
+	logger.With().Info(
+		"selected active signers",
+		log.Int("count", len(active)),
+		log.Array("signers", zapcore.ArrayMarshalerFunc(func(enc zapcore.ArrayEncoder) error {
+			for _, p := range active {
+				enc.AppendObject(p)
+			}
+			return nil
+		})),
+	)
+
 	checker := createProposalChecker(logger, pd.config, w1, w1+w2)
 	pd.states[epoch] = newState(logger, pd.config, active, epochWeight, miners, checker)
 	return pd.states[epoch], nil
@@ -884,6 +902,7 @@ func (pd *ProtocolDriver) runConsensusPhase(ctx context.Context, epoch types.Epo
 	case <-ctx.Done():
 		return allVotes{}, fmt.Errorf("context done: %w", ctx.Err())
 	}
+	ownVotes, _ = pd.calcVotesBeforeWeakCoin(logger, st)
 
 	// Subsequent rounds
 	for round := types.FirstRound + 1; round < pd.config.RoundsNumber; round++ {
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -768,7 +768,7 @@ func (pd *ProtocolDriver) runProtocol(ctx context.Context, epoch types.EpochID, 
 		logger.With().Warning("proposal phase failed", log.Err(err))
 		return
 	}
-	lastRoundOwnVotes, err := pd.runConsensusPhase(ctx, epoch, st.active)
+	lastRoundOwnVotes, err := pd.runConsensusPhase(ctx, epoch, st)
 	if err != nil {
 		logger.With().Warning("consensus phase failed", log.Err(err))
 		return
@@ -791,7 +791,7 @@ func (pd *ProtocolDriver) runProtocol(ctx context.Context, epoch types.EpochID, 
 }
 
 func calcBeacon(logger log.Log, set proposalSet) types.Beacon {
-	allProposals := set.sort()
+	allProposals := set.sorted()
 
 	// Beacon should appear to have the same entropy as the initial proposals, hence cropping it
 	// to the same size as the proposal
@@ -825,11 +825,9 @@ func (pd *ProtocolDriver) runProposalPhase(ctx context.Context, epoch types.Epoc
 		return pd.ctx.Err()
 	}
 
-	if err := pd.markProposalPhaseFinished(epoch, time.Now()); err != nil {
-		return err
-	}
-
-	logger.Info("beacon proposal phase finished")
+	finished := time.Now()
+	pd.markProposalPhaseFinished(st, finished)
+	logger.With().Info("proposal phase finished", log.Time("finished_at", finished))
 	return nil
 }
 
@@ -863,7 +861,7 @@ func (pd *ProtocolDriver) sendProposal(ctx context.Context, epoch types.EpochID,
 }
 
 // runConsensusPhase runs K voting rounds and returns result from last weak coin round.
-func (pd *ProtocolDriver) runConsensusPhase(ctx context.Context, epoch types.EpochID, active map[types.NodeID]signerSession) (allVotes, error) {
+func (pd *ProtocolDriver) runConsensusPhase(ctx context.Context, epoch types.EpochID, st *state) (allVotes, error) {
 	logger := pd.logger.WithContext(ctx).WithFields(epoch)
 	logger.Info("starting consensus phase")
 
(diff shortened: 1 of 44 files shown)


beacon/beacon.go
```diff
@@ -135,22 +120,60 @@ func New(
...
-		pd.nonceFetcher = defaultFetcher{cdb: cdb}
-	}
 
-	if pd.weakCoin == nil {
-		pd.weakCoin = weakcoin.New(pd.publisher, edSigner.VRFSigner(), vrfVerifier, pd.nonceFetcher, pd,
-			pd.msgTimes,
-			weakcoin.WithLog(pd.logger.WithName("weakCoin")),
-			weakcoin.WithMaxRound(pd.config.RoundsNumber),
-		)
+	if pd.createWeakCoin == nil {
```
dshulyak: it doesn't make sense to have several weak coin instances. there should be on instance that holds state, and routine that goes over all signers and creates proposal for each one

in current implementation i think the simplest would be to split StartRound into two methods, one generic for all signers and another one accepts signer and publishes proposal

beacon/beacon.go
```diff
@@ -848,21 +879,42 @@ func (pd *ProtocolDriver) runConsensusPhase(ctx context.Context, epoch types.Epo
...
+
+	// First round
+	round := types.FirstRound
+	pd.setRoundInProgress(round)
+	if msg, err := pd.genFirstRoundMsgBody(epoch); err != nil {
+		logger.With().Error("failed to get firstround message", log.Err(err), round)
+	} else {
+		for _, session := range active {
+			session := session
+			pd.eg.Go(func() error {
```
dshulyak: could you remove pb.eg.Go in sendToGossip?

beacon/handlers.go
```diff
@@ -38,12 +40,21 @@ var (
...
 	}
+	var eg errgroup.Group
+	pd.mu.RLock()
+	for _, signer := range pd.signers {
+		coin := signer.coin
+		eg.Go(func() error {
+			return coin.HandleProposal(ctx, peer, msg)
+		})
+	}
+	pd.mu.Unlock()
```
dshulyak: should RUnlock? but i think this can be refactored to avoid creating multiple weak coin instances

beacon/handlers.go
```diff
@@ -229,17 +240,17 @@ func (pd *ProtocolDriver) HandleFirstVotes(ctx context.Context, peer p2p.Peer, m
 	currentEpoch := pd.currentEpoch()
 	if m.EpochID != currentEpoch {
 		logger.With().Debug("first votes from different epoch",
-			log.Uint32("current_epoch", uint32(currentEpoch)),
-			log.Uint32("message_epoch", uint32(m.EpochID)))
+			log.FieldNamed("current_epoch", currentEpoch),
+			log.FieldNamed("message_epoch", m.EpochID))
```
dshulyak: FieldNamed is deprecated as it relies on copying whole array and then unpacking original fields

beacon/handlers.go
```diff
@@ -229,17 +240,17 @@ func (pd *ProtocolDriver) HandleFirstVotes(ctx context.Context, peer p2p.Peer, m
...
 	}
 
 	// don't accept more first vote after the round ends
 	currentRound := pd.currentRound()
 	if currentRound > types.FirstRound {
 		logger.With().Debug("first votes too late",
-			log.Uint32("current_round", uint32(currentRound)),
-			log.Uint32("message_round", uint32(types.FirstRound)))
+			log.FieldNamed("current_round", currentRound),
+			log.FieldNamed("message_round", types.FirstRound))
```
dshulyak: FieldNamed is deprecated

beacon/beacon.go
```diff
@@ -135,22 +120,60 @@ func New(
...
+
+func (s *participant) Id() signerID {
+	return signerID(s.signer.NodeID())
+}
+
+type signerID types.NodeID
+
+func (id signerID) Field() log.Field {
+	return log.ShortStringer("id", types.NodeID(id))
+}
```
fasmat: I'm not sure about introducing a new type here just to change the logging behavour of `NodeID` only for `beacon`. We should consider `(NodeID) Field()` to return a `log.ShortStringer` and have that behaviour everywhere.

beacon/beacon.go
```diff
@@ -848,21 +879,42 @@ func (pd *ProtocolDriver) runConsensusPhase(ctx context.Context, epoch types.Epo
...
 		err       error
 	)
-	for round := types.FirstRound; round < pd.config.RoundsNumber; round++ {
+
+	// First round
+	round := types.FirstRound
+	pd.setRoundInProgress(round)
+	if msg, err := pd.genFirstRoundMsgBody(epoch); err != nil {
+		logger.With().Error("failed to get firstround message", log.Err(err), round)
+	} else {
```
fasmat: The only reason `pd.genFirstRoundMsgBody` might fail is if the identity isn't active in the given epoch. But `active` already has been checked for that so can this method return an error?

Alternatively `pd.genFirstRoundMsgBody` could just be part of `sendFirstRoundVote` to simplify the logic.

beacon/handlers.go
poszu: I don't understand - what array do you have in mind and where is it unpacked? Is the problem that `func (e EpochID) Field()` is called eagerly inside of `FieldNamed()`?

Anyway - if it's deprecated let's mark it so and create an issue to delete it.

beacon/beacon.go
poszu: What's wrong with having a private type here? Btw, it changed the key as well (NodeID logs "node_id", not "id").

fasmat: The problem isn't the private type, it's that `beacon` now logs `NodeID` differently from all other components, which makes it harder to read them imo.

beacon/beacon.go
poszu: It's generally a problem with creating too many "small helper methods" that it is hard to follow in what conditions a method is called (for example, what is already asserted). I looked over code again. It turns out `genFirstRoundMsgBody()` is called in 1 place and the `state` for the epoch has already been picked and asserted that it exist. Hence it makes no sense to pick it up again from `pd.states[epoch]` in the first place.

I will refactor the code to pass the once picked state down the stack.

>Alternatively pd.genFirstRoundMsgBody could just be part of sendFirstRoundVote to simplify the logic.

It cannot because:
- this work is common for all signers so it makes no sense to repeat it
- it mutates shared state by sorting incoming proposals
 
That's why I extracted it from `sendFirstRoundVote`.

fasmat: > this work is common for all signers so it makes no sense to repeat it

I initially assumed the method just creates a message and returns -> that doesn't need to be its own method. The problem here is that the method has an additional side effect (sorting proposals) that isn't not obvious from its name or on a first glance when looking at the method.

> it mutates shared state by sorting incoming proposals

That's fair, but then sorting should be done once in a method that's named "sortIncomingProposals" or similar and the instantiation of the Message object can be done in `sendFirstRoundVote`. At least I think that would make it clearer what's happening 🙂

beacon/beacon.go
poszu: IMHO, it doesn't make sense to log messages per-signer with a "node_id" field, it has nothing to do with "node" anymore (it's not a node ID in this context, there are many IDs governed by a single node).

Perhaps instead, the `NodeID` type should be dropped entirely in favor of a `SignerID` or something similar? But that's a bigger refactor.

beacon/beacon.go
poszu: I removed the method entirely. ~Btw, I discovered that it had a bug - it took a shared Rlock and mutated the data :(~

beacon/handlers.go
poszu: Good catch, thanks! The sad part is that it means this code is never executed in tests :(

beacon/handlers.go
dshulyak: i meant this code in spacemesh logger

```go
func (fl FieldLogger) Info(msg string, fields ...LoggableField) {
	fl.l.Info(msg, unpack(append(fields, String("name", fl.name)))...)
}
```

beacon/beacon.go
poszu: You are right, I refactored the code, but I didn't split the `StartRound` method as it didn't seem necessary. It now takes a slice of participants (vrf signer + vrf nonce).

beacon/handlers.go
poszu: OK, but this happens anyway, doesn't it?

poszu:
bors try

beacon/weakcoin/weak_coin.go
```diff
@@ -193,7 +192,7 @@ func (wc *WeakCoin) FinishEpoch(ctx context.Context, epoch types.EpochID) {
 	if epoch != wc.epoch {
 		logger.With().Fatal("attempted to finish beacon weak coin for the wrong epoch",
 			epoch,
-			log.Stringer("weak_coin_epoch", wc.epoch),
+			log.FieldNamed("weak_coin_epoch", wc.epoch),
```
dshulyak: can you Uint32(), my idea was that we remove FieldNamed over time and then it reduces amount of refactoring

dshulyak (approved):

beacon/handlers.go
fasmat: Would it be much effort to migrate `ProtocolDriver` from using `go-spacemesh/log` to using `zap`? I also don't fully understand why loggers are passed around via context, when `ProtocolDriver` already has an instance of a logger?

poszu:
bors try

beacon/handlers.go
poszu: Probably not much effort. I will try to migrate this/next week.

poszu:
bors merge

poszu:
Failure in unrelated UT

bors merge

PR #5161, opened 2023-10-16
[Merged by Bors] - fetch: parametrize success rate threshold

dshulyak:
this threshold used to determine if peers are seen as equally responsive within certain threshold.
parametrizing it to run some experiments


fasmat (approved):

dshulyak:
bors merge

PR #5162, opened 2023-10-16
[Merged by Bors] - Specify gcloud version in release and systest jobs

fasmat:
## Motivation
Not specifying a version leads the jobs to download the latest version on every build. Since a new version of gcloud is released on average once per week and those installations take up ~550 MB on our runners and are not removed automatically this eventually fills up the disks of the runners.

## Changes
- Specify a gcloud version in `systests` and `release` jobs. This way only that version is downloaded and re-used on every subsequent run.
-  We can manually update gcloud to a newer version if needed.

## Test Plan
n/a

## TODO


dshulyak (approved):

fasmat:
bors merge

Issue #4668, opened 2023-07-07
Flaky test: TestCalcEligibilityWithSpaceUnit


fasmat:
## Description
Failed at least once in a PR without code changes

```
=== Failed
=== FAIL: hare/eligibility TestCalcEligibilityWithSpaceUnit/large_network (1.68s)
    oracle_test.go:285: diff=87 (10.875% of committeeSize)
    oracle_test.go:286: 
        	Error Trace:	/home/runner/work/go-spacemesh/go-spacemesh/hare/eligibility/oracle_test.go:286
        	Error:      	"87" is not less than "80"
        	Test:       	TestCalcEligibilityWithSpaceUnit/large_network
    --- FAIL: TestCalcEligibilityWithSpaceUnit/large_network (1.68s)

=== FAIL: hare/eligibility TestCalcEligibilityWithSpaceUnit (1.88s)
```

https://github.com/spacemeshos/go-spacemesh/actions/runs/5485685554/jobs/9994806988

fasmat:
Has been fixed in #5496

PR #5163, opened 2023-10-16
[Merged by Bors] - Remove confusing clock warning

lrettig:
Don't warn about skipping layers right after fresh start. This causes constant confusion among community members.

### DIFF
--- a/timesync/clock.go
+++ b/timesync/clock.go
@@ -136,14 +136,12 @@ func (t *NodeClock) tick() {
 		)
 		d := t.lastTicked.Difference(layer)
 		tickDistance.Observe(float64(-d))
-	case layer.Difference(t.lastTicked) > 1:
-		// don't warn right after fresh startup
-		if t.lastTicked != 0 {
-			t.log.With().Warning("clock skipped layers",
-				log.Stringer("layer", layer),
-				log.Stringer("last_ticked_layer", t.lastTicked),
-			)
-		}
+	// don't warn right after fresh startup
+	case layer.Difference(t.lastTicked) > 1 && t.lastTicked > 0:
+		t.log.With().Warning("clock skipped layers",
+			log.Stringer("layer", layer),
+			log.Stringer("last_ticked_layer", t.lastTicked),
+		)
 		d := layer.Difference(t.lastTicked)
 		tickDistance.Observe(float64(d))
 	case layer == t.lastTicked:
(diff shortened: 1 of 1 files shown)


timesync/clock.go
```diff
@@ -137,10 +137,13 @@ func (t *NodeClock) tick() {
...
-			log.Stringer("layer", layer),
-			log.Stringer("last_ticked_layer", t.lastTicked),
-		)
+		// don't warn right after fresh startup
+		if t.lastTicked != 0 {
+			t.log.With().Warning("clock skipped layers",
+				log.Stringer("layer", layer),
+				log.Stringer("last_ticked_layer", t.lastTicked),
+			)
+		}
```
fasmat: You can combine the check into the case:

```suggestion
	case layer.Difference(t.lastTicked) > 1 && t.lastTicked > 0:
		t.log.With().Warning("clock skipped layers",
			log.Stringer("layer", layer),
			log.Stringer("last_ticked_layer", t.lastTicked),
		)
```

lrettig: @fasmat that changes the behavior of the code. read the following context.

fasmat: I'm just suggesting that instead of

```go
if layer.Difference(t.lastTick) > 1 {
    if t.lastTick != 0 {
        t.og.With.Warning("...")
    }

    // some more code
}
```

to write:

```go
if layer.Difference(t.lastTick) > 1 && if t.lastTick != 0 {
    t.log.With().Warning("...")
    // some more code
}
```

which should behave the same?

fasmat: Regarding the code that comes afterwards; I think it should also not be executed if `t.lastTick` is 0 if that's where the confusion comes from 🙂 - sorry for not writing a complete code example. I updated my previous comment to clarify.

lrettig: why wouldn't you want to execute `tickDistance.Observe()`? I'm trying to change a log here, not any other behavior.

lrettig: i see, that's only used in metrics... didn't check that

fasmat (approved):

lrettig:
bors merge

lrettig:
bors merge

fasmat:
The merge failed twice because of the node not being able to connect to `0.0.0.0` to submit a challenge to PoET in `TestAdminEvents`. https://github.com/spacemeshos/go-spacemesh/pull/5165 should fix that issue.

fasmat:
bors merge

fasmat:
flaky test: https://github.com/spacemeshos/go-spacemesh/issues/4668

bors merge

PR #5164, opened 2023-10-17
[Merged by Bors] - set cache to 100_000 and disable libp2p resource manager

dshulyak:
- raising atx cache size to 100_000 to prevent threshing of the lru cache
- disable libp2p resource manager. we are not relying on it

### DIFF
--- a/datastore/store.go
+++ b/datastore/store.go
@@ -46,7 +46,7 @@ type Config struct {
 
 func DefaultConfig() Config {
 	return Config{
-		ATXSize:         50_000,
+		ATXSize:         100_000,
 		MalfeasenceSize: 1_000,
 	}
 }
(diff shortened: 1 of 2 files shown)


datastore/store.go
```diff
@@ -237,7 +237,10 @@ func (db *CachedDB) GetEpochWeight(epoch types.EpochID) (uint64, []types.ATXID,
 }
 
 // IterateEpochATXHeaders iterates over ActivationTxs that target an epoch.
-func (db *CachedDB) IterateEpochATXHeaders(epoch types.EpochID, iter func(*types.ActivationTxHeader) error) error {
+func (db *CachedDB) IterateEpochATXHeaders(
+	epoch types.EpochID,
+	iter func(*types.ActivationTxHeader) error,
+) error {
```
fasmat, line 243: Did you change your formatting settings? I don't mind lines that are 120 (or even 160) characters long. Limiting ourselves to 80 characters seems archaic to me.

fasmat (approved):

datastore/store.go
dshulyak: i installed golines per @poszu request :) 

i think default is 100, i overlooked that. @poszu do you prefer 200 or 160?

dshulyak: will merge it as is (because i cherrypicked it to v1.2.2) and update config for next patches

dshulyak:
bors merge

datastore/store.go
poszu: @dshulyak I prefer up to 120. Longer lines don't fit in side-by-side reviews. 80 is in fact too short.

PR #5169, opened 2023-10-18
[Merged by Bors] - implement pruner for activesets

### LINKED ISSUE #5114: prune or re-encode activesets from previous epochs
stored activesets are ~530MB, while total db size is 1.4GB. 

they contain a lot of duplicate data, so there is an opportunity to dedup them, either on gossip or on database level:
- delta encoding. reference to original activeset plus new/removed ids
- store single collection of ids in the epoch ordered by receive time, every new activeset stores bitfield with 1 for atxs that it has in its own activeset

the simplest would be to prune it, Iddo shared that we may want to wait with that:
> About pruning, the only reason that it might not be ok is if we have protocol rule that measure the median observed total weight. What Tal described in the call about not rewarding if the active set weight is too small (but above the hardcoded minimal weight) was about ephemeral data, but we also have ongoing discussion regarding how to prevent DoS attack (regarding if your own weight is too low in the far future after genesis, and this issue of observed weight is relevant in this context too). So we should be careful before deciding to prune active sets (deduplication is benign and we should do it asap).

it is caused by unreliable propagation of atxs. expected to get improved by https://github.com/spacemeshos/go-spacemesh/pull/5097 . so need to recheck it in next epoch

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/5114

pruning activesets will significantly reduce space (as of now ~1.5GB / 2.9GB).


prune/prune.go
```diff
@@ -8,63 +8,87 @@ import (
...
+}
+
+type Pruner struct {
+	logger         *zap.Logger
+	db             *sql.Database
+	safeDist       uint32
+	activesetEpoch types.EpochID
+}
+
+func Run(ctx context.Context, p *Pruner, clock *timesync.NodeClock, interval time.Duration) {
```
fasmat, line 46: Could this be a method of `Pruner`?

fasmat (approved):

prune/prune.go
dshulyak: it can be, i thought that it is better to remove dependencies from Pruner that can be passed only to Run

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

prune/prune.go
fasmat: Run takes a pruner as an argument, so someone calling it has to instantiate it anyway?

```go
pruner := prune.New(...)
prune.Run(ctx, pruner, ...)
// vs
pruner := prune.New(...)
pruner.Run(ctx, ...)
```

dshulyak:
bors merge

PR #5171, opened 2023-10-18
[Merged by Bors] - setup min activeset weight with data from latest epoch

### LINKED ISSUE #5137: Smesher with many small proposals/rewards
Right now we let smeshers choose their own active sets and correspondingly their eligibility slots for proposals each epoch, which can result in a smesher submitting many proposals that each receive only a tiny reward. This appears to be happening with one smesher: 0x318cceb3df7174d42b582ffb4ccfd6ef6fbce4def8aee342c33e264ff78909f8

Example reward: https://explorer.spacemesh.io/rewards/6524347792abdc943072d99c

Coinbase: sm1qqqqqq8q9hxmhdqnktagjx7r65cf3vjfk79knvqnzy7yg

CC @dshulyak

dshulyak:
closes: https://github.com/spacemeshos/go-spacemesh/issues/5137

- added a tool that can be used to read db and compute total weight of atxs in the latest epoch
- activeset weight configured as  a sorted array of tuples (epoch, weight). components select latest tuple that match epoch based on timing of usage


cmd/activeset/activeset.go
```diff
@@ -0,0 +1,43 @@
...
+	flag.Usage = func() {
+		fmt.Println("activeset <publish epoch> <db path>")
+		flag.PrintDefaults()
+	}
+	flag.Parse()
+
+	publish, err := strconv.Atoi(flag.Arg(0))
+	must(err, "publish epoch %v is not a valid integer: %s", flag.Arg(0), err)
+	dbpath := flag.Arg(1)
+	db, err := sql.Open("file:" + dbpath)
```
poszu: It would make sense to verify if the arg was provided (`if dbpath != ""`) and print the usage message.

config/mainnet.go
```diff
@@ -95,9 +96,12 @@ func MainnetConfig() Config {
...
 			BadBeaconVoteDelayLayers: 4032,
-			// 1000 - is assumed minimal number of units
-			// 5000 - half of the expected poet ticks
-			MinimalActiveSetWeight: 1000 * 5000,
+			MinimalActiveSetWeight: []types.EpochMinimalActiveWeight{
+				{Weight: 1_000_000},
+				// generated using ./cmd/activeset for publish epoch 6
+				// it will be used starting from epoch 8, because we will only release it in 7th
+				{Epoch: 8, Weight: 7_879_129_244},
+			},
```
poszu, line 105: Is this list expected to be small? If it's going to have an entry per every epoch, then perhaps using something like a [btree](https://pkg.go.dev/github.com/google/btree#BTree) would be beneficial (for faster lookup in `Select()`).

miner/minweight/minweight_test.go
```diff
@@ -0,0 +1,32 @@
...
+		require.EqualValues(t, 0, Select(0, nil))
+	})
+	t.Run("sorted", func(t *testing.T) {
+		require.EqualValues(t, 10, Select(5, []types.EpochMinimalActiveWeight{
+			{Epoch: 0, Weight: 1},
+			{Epoch: 4, Weight: 5},
+			{Epoch: 5, Weight: 10},
+			{Epoch: 7, Weight: 11},
+		}))
+	})
```
poszu, line 22: How about
```suggestion
	t.Run("sorted", func(t *testing.T) {
		require.EqualValues(t, 10, Select(5, []types.EpochMinimalActiveWeight{
			{Epoch: 0, Weight: 1},
			{Epoch: 4, Weight: 5},
			{Epoch: 5, Weight: 10},
			{Epoch: 7, Weight: 11},
		}))
	})
	t.Run("in-between epochs", func(t *testing.T) {
		require.EqualValues(t, 10, Select(6, []types.EpochMinimalActiveWeight{
			{Epoch: 0, Weight: 1},
			{Epoch: 4, Weight: 5},
			{Epoch: 5, Weight: 10},
			{Epoch: 7, Weight: 11},
		}))
	})
```

poszu (approved):

config/mainnet.go
dshulyak: it is expected to be small. min activeset value is only needed to compute eligibilities in the current epoch. for previous epochs we store a counter. there are two values so that we can update in the middle of the epoch, likely it will be at most 2 in any release

miner/minweight/minweight_test.go
dshulyak: thanks added a few more test cases

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #5165, opened 2023-10-17
[Merged by Bors] - Try possible fix for flaky TestAdminEvents on Windows

fasmat:
## Motivation
Instead of trying to access the local PoET via `0.0.0.0:10010` access it via `127.0.0.1:10010`. The first address seems to cause problems on some Windows systems

## Changes
Updated standalone config to use 127.0.0.1 for local addresses instead of 0.0.0.0

## Test Plan
n/a

## TODO

- [x] Explain motivation or link existing issue(s)
- [x] Test changes and document test plan
- [x] Update documentation as needed
- [x] Update [changelog](../CHANGELOG.md) as needed

PR #5172, opened 2023-10-18
[Merged by Bors] - Let system decide which port to use for PoET gRPC in standalone mode

fasmat:
## Motivation
Follow up to https://github.com/spacemeshos/go-spacemesh/pull/5165

Additionally to Windows not liking `0.0.0.0` as connection address, the Windows GH runners sometimes disallow to listen on port 50002 (default for PoET gRPC). It seems to be an issue with some of the windows runners, since no other test uses a standalone PoET (so nothing should be listening there already) and it only appears to happen on Windows. Assigning a random (free) port as listen address seems generally a good idea (for parallelisation, less likely to fail, etc.)

## Changes
- changed `RawRPCListener` to use a random port instead of the default one if node is launched in standalone mode

## Test Plan
n/a

## TODO

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -1159,6 +1159,7 @@ func (app *App) launchStandalone(ctx context.Context) error {
 		return err
 	}
 	cfg.RawRESTListener = parsed.Host
+	cfg.RawRPCListener = parsed.Hostname() + ":0"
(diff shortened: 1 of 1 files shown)


dshulyak (approved):

fasmat:
bots merge

fasmat:
bors merge

PR #5173, opened 2023-10-18
tally command line tool to sanity check vm

dshulyak:
compile it and run with any address:

> ./tally -db state.sql -acc sm1qqqqqqpy8svgfxfhh2w42ujhrynsxgwsrrzgplss9t0lv
address=sm1qqqqqqpy8svgfxfhh2w42ujhrynsxgwsrrzgplss9t0lv
layer=27658
earned=1549272967127595
rewards=0
spent=1075931911672603
tally=473341055454992
account=473341055454992

### DIFF
--- a/genvm/cmd/tally/main.go
+++ b/genvm/cmd/tally/main.go
@@ -36,6 +36,17 @@ func main() {
 	filter := transactions.ResultsFilter{Address: &addr}
 	must(transactions.IterateResults(db, filter,
 		func(rst *types.TransactionWithResult) bool {
+			status := "PASS"
+			if rst.Status != types.TransactionSuccess {
+				status = "FAIL"
+			}
+			fmt.Printf(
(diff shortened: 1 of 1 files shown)


brunovale91:
@dshulyak do ineffective transaction have a status different from types.TransactionSuccess?

dshulyak:
all effective transactions should have a result. transaction ineffective if it is included in the block and block was applied, but has no result (e.g field in db is null)

> select distinct id, tx, header, result 
		from transactions
		left join transactions_results_addresses on id=tid
		where result is not null

brunovale91:
> all effective transactions should have a result. transaction ineffective if it is included in the block and block was applied, but has no result (e.g field in db is null)
> 
> > select distinct id, tx, header, result
> > from transactions
> > left join transactions_results_addresses on id=tid
> > where result is not null

Thanks for the explanation.

fasmat:
Closing as this seems abandoned. @dshulyak feel free to re-open if needed 🙂

PR #5180, opened 2023-10-19
[Merged by Bors] - Remove ctx field from Generator and Certifier

poszu:
## Motivation
It's considered an anti-pattern to store a `Context` in a struct. It leads to code that is difficult to reason about.

## Changes
Removed `ctx` field from `Generator` and `Certifier` structs. Instead, their `Start()` methods take a context. The `Stop()` is handled by cancelling the context derived from the context passed to `Start()`.

## Test Plan
Existing tests should pass

### DIFF
--- a/blocks/certifier.go
+++ b/blocks/certifier.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"golang.org/x/sync/errgroup"
 
@@ -50,13 +51,6 @@ func defaultCertConfig() CertConfig {
 // CertifierOpt for configuring Certifier.
 type CertifierOpt func(*Certifier)
 
-// WithCertContext modifies parent context for Certifier.
-func WithCertContext(ctx context.Context) CertifierOpt {
-	return func(c *Certifier) {
-		c.ctx = ctx
-	}
-}
-
 // WithCertConfig defines cfg for Certifier.
 func WithCertConfig(cfg CertConfig) CertifierOpt {
 	return func(c *Certifier) {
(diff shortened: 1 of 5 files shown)


blocks/certifier.go
```diff
@@ -137,44 +131,39 @@ func NewCertifier(
...
 	})
 }
 
 // Stop stops the outstanding goroutines.
 func (c *Certifier) Stop() {
-	c.cancel()
+	c.stopped.Store(true)
+	if c.stop == nil {
+		return // not started
+	}
```
fasmat, line 152: Otherwise calling `Stop` and then `Start` creates a service that cannot be stopped.

```suggestion
	if c.stop == nil {
		return // not started
	}
	c.stopped.Store(true)
```

poszu: Yes, it's intended. Both the Generator and Certifier are single-use. They cannot be reused after being stopped. There's even a unit test that would fail after this change.

fasmat: Would it then make sense to call the `Stop` method `Close`? `Stop` to me implies the ability to be able to restart it while `Close` means the object cannot be used afterwards any more.

fasmat (approved):

blocks/certifier.go
poszu: Perhaps :shrug:. It would be good to sit together and rethink all the services and plan to make them work and look similar.

poszu:
bors merge

poszu:
bors merge

poszu:
The `TestSmeshing` failed, seems due to infra - retrying

bors merge

PR #5181, opened 2023-10-19
[Merged by Bors] - Support multiple signers in certifier

### LINKED ISSUE #5086: cert: support multiple smeshers
all considerations here are exactly the same as https://github.com/spacemeshos/go-spacemesh/issues/5085 , but can be implemented in parallel

poszu:
Closes #5086 

## Changes
- `Certifier` keeps a set of signers
- removed signer from the hare Oracle. It now takes vrf signer as an argument to `Proof()` method
- removed logging from `CertifyIfEligible` in favor of returning context-rich errors, which are logged up the stack

## Test Plan
- existing tests for `CertifyIfElligible` were extended to use many (3) signers
- each signer must publish a block certification message

### DIFF
--- a/blocks/certifier.go
+++ b/blocks/certifier.go
@@ -220,7 +220,7 @@ func (c *Certifier) RegisterForCert(ctx context.Context, lid types.LayerID, bid 
 
 // CertifyIfEligible signs the hare output, along with its role proof as a certifier, and gossip the CertifyMessage
 // if the node is eligible to be a certifier.
-func (c *Certifier) CertifyIfEligible(ctx context.Context, logger log.Log, lid types.LayerID, bid types.BlockID) error {
+func (c *Certifier) CertifyIfEligible(ctx context.Context, lid types.LayerID, bid types.BlockID) error {
 	if _, err := c.beacon.GetBeacon(lid.GetEpoch()); err != nil {
 		return errBeaconNotAvailable
 	}
@@ -234,7 +234,7 @@ func (c *Certifier) CertifyIfEligible(ctx context.Context, logger log.Log, lid t
 	for _, s := range c.signers {
 		s := s
 		eg.Go(func() error {
-			if err := c.certifySingleSigner(ctx, logger, s, lid, bid); err != nil {
+			if err := c.certifySingleSigner(ctx, s, lid, bid); err != nil {
 				errsMu.Lock()
 				errs = errors.Join(errs, fmt.Errorf("certifying block %v/%v by %s: %w", lid, bid, s.NodeID().ShortString(), err))
 				errsMu.Unlock()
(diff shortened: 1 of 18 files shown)


poszu:
bors try

poszu:
:bulb: The [Reformat blocks/certifier_test.go](https://github.com/spacemeshos/go-spacemesh/pull/5181/commits/e19304d1e2a75824e9b4b927179793dd0c35ac31) commit contains only formatting changes, so it's best to review it separately.

poszu:
bors try

poszu:
bors try

blocks/certifier.go
```diff
@@ -203,27 +214,50 @@ func (c *Certifier) RegisterForCert(ctx context.Context, lid types.LayerID, bid
...
+}
+
+func (c *Certifier) certifySingleSigner(ctx context.Context, s *signing.EdSigner, lid types.LayerID, bid types.BlockID) error {
+	// check if the signer is eligible to certify the hare output
+	proof, err := c.oracle.Proof(ctx, s.VRFSigner(), lid, eligibility.CertifyRound)
 	if err != nil {
-		logger.With().Error("failed to get eligibility proof to certify", log.Err(err))
-		return err
+		return fmt.Errorf("getting eligibility proof: %w", err)
 	}
```
dshulyak, line 239: you can use GenVRF here, and avoid useless error because you already got beacon before calling this method

blocks/certifier.go
```diff
@@ -203,27 +214,50 @@ func (c *Certifier) RegisterForCert(ctx context.Context, lid types.LayerID, bid
...
+	info := c.createIfNeeded(lid, bid)
+	info.registered = true
+	return c.tryGenCert(ctx, logger, lid, bid, info)
 }
 
 // CertifyIfEligible signs the hare output, along with its role proof as a certifier, and gossip the CertifyMessage
 // if the node is eligible to be a certifier.
-func (c *Certifier) CertifyIfEligible(ctx context.Context, logger log.Log, lid types.LayerID, bid types.BlockID) error {
+func (c *Certifier) CertifyIfEligible(ctx context.Context, lid types.LayerID, bid types.BlockID) error {
 	if _, err := c.beacon.GetBeacon(lid.GetEpoch()); err != nil {
```
dshulyak: this beacon can be reused below instead (Proof vs GenVRF)

blocks/certifier.go
```diff
@@ -203,27 +214,50 @@ func (c *Certifier) RegisterForCert(ctx context.Context, lid types.LayerID, bid
...
+	var (
+		errsMu sync.Mutex
+		errs   error
+		eg     errgroup.Group
+	)
+	c.mu.Lock()
+	for _, s := range c.signers {
+		s := s
+		eg.Go(func() error {
+			if err := c.certifySingleSigner(ctx, s, lid, bid); err != nil {
```
dshulyak: how about inlining certifySingleSigner here?

dshulyak (approved):

blocks/certifier.go
```diff
@@ -203,27 +214,50 @@ func (c *Certifier) RegisterForCert(ctx context.Context, lid types.LayerID, bid
...
+
+	var (
+		errsMu sync.Mutex
+		errs   error
+		eg     errgroup.Group
+	)
+	c.mu.Lock()
+	for _, s := range c.signers {
+		s := s
+		eg.Go(func() error {
```
dshulyak: btw parallelization here is not an obvious choice.. there are two expensive operations (ed25519 and ecvrf) and both are measured in us, so maybe single key is 400us. parallelization is maybe useful for > 1000 signers, but then no need to create 1000 goroutines and it should be limited by number of cores

dshulyak (approved):

blocks/certifier.go
poszu: I extracted it to simplify error handling.

dshulyak (approved):

poszu:
bors merge

Issue #5149, opened 2023-10-12
activation: support multiple smeshers


fasmat:
###  Description

To fully utilise the separation of PoST from go-spacemesh a node needs to be able to handle multiple post services simultaneously.

### Acceptance criteria

- Allow multiple PoST services to connect via the grpc interface.
- Code in the activation package needs to be updated to handle multiple identities correctly.
  - The activation package builds an ATX for every identity that is set up for the node and has a post service connected
  - Only one post service per identity can connect to a node
- Connections from PoST services over public interfaces have to be authenticated with mTLS
  - Assume that both sides have each others certificate

### Implementation hints

- The general design of the setup is described here: https://github.com/spacemeshos/pm/issues/259
- Consider storing the nipost builder state, - challenge, and initial post in the DB instead of on disk (`post.bin`, `nipost_challenge.bin` and `nipost_builder_state.bin`).
  - There will be at least one initial post and at least one nipost challenge for every identity managed by the node
  - Also consider splitting the builder state into one state for each identiy & poet combination, so they can be tracked individually.
  - [x] https://github.com/spacemeshos/go-spacemesh/issues/5206

dshulyak:
ideally we also should skip post validation for self-published atxs

fasmat:
> ideally we also should skip post validation for self-published atxs

validation in the NiPost builder was already removed in https://github.com/spacemeshos/go-spacemesh/pull/5061. The activation package trusts the post service to deliver a valid proof and creates an ATX with it. That ATX however will still be validated (including proof validation) before/as part of broadcasting it, I think.

fasmat:
Removing the dedicated validation of the NIPostBuilder lead to some (rare) cases where publishing of an ATX might result in weird behaviour:

- if the `activation.Builder` successfully publishes an ATX but fails to persist that it did, it might publish a second ATX causing the node to create a malfeasance proof against itself
  - this should be changed to instead fail the ATX verification but without creating the malfeasance proof against oneself
- there were cases reported where the broadcast of the ATX failed (due to a falling validation by the activation handler). The `activation.Builder` however seems to still consider it a successful Publication and continues with the next ATX.
  - this won't lead to any adverse effects for the node (besides missing rewards - but the ATX would have been invalid anyway).
[...]

PR #5186, opened 2023-10-20
[Merged by Bors] - Add Info request to post client

fasmat:
## Motivation
Part of #5149 

This extends the `PostClient` to be able to fetch metadata about the PoST from the service

## Changes
- Add method to fetch metadata from PoST service

## Test Plan
- Tests were added for new functionality with and without TLS

## TODO

### DIFF
--- a/api/grpcserver/post_client.go
+++ b/api/grpcserver/post_client.go
@@ -29,81 +29,131 @@ func newPostClient(con chan<- postCommand) *postClient {
 	}
 }
 
+func (pc *postClient) Info(ctx context.Context) (*types.PostInfo, error) {
+	req := &pb.NodeRequest{
+		Kind: &pb.NodeRequest_Metadata{
+			Metadata: &pb.MetadataRequest{},
+		},
+	}
+	resp, err := pc.send(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	metadataResp := resp.GetMetadata()
+	if metadataResp == nil {
+		return nil, fmt.Errorf("unexpected response of type: %T", resp.GetKind())
+	}
+	meta := metadataResp.GetMeta()
+	if meta == nil {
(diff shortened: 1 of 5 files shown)


api/grpcserver/post_client.go
```diff
@@ -29,82 +29,132 @@ func newPostClient(con chan<- postCommand) *postClient {
...
-			}
+		case <-time.After(2 * time.Second):
+			// TODO(mafa): make polling interval configurable
+			continue
 		}
 	}
+
+	proof := proofResp.GetProof()
+	metaData := proofResp.GetMetadata()
+	if metaData == nil {
```
poszu: This looks weird. How about:
```suggestion
	metadata := proofResp.GetMetadata()
	if metadata == nil {
```

api/grpcserver/post_service_test.go
```diff
@@ -233,6 +233,86 @@ func Test_Cancel_GenerateProof(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		meta, err = client.Info(context.Background())
+		return err != nil
+	}, 5*time.Second, 100*time.Millisecond)
+
+	require.ErrorContains(t, err, "post client closed")
+	require.Nil(t, meta)
+}
+
+func Test_Metadata_TLS(t *testing.T) {
```
poszu: Is this test case really needed? TLS is orthogonal to the grpc method functionality, isn't? I mean that it's sufficient to test if TLS works in isolation and then test all methods just in "plain mode".

poszu (approved):

api/grpcserver/post_service_test.go
fasmat: It doesn't hurt to have both, the test only takes ~ 3 seconds to execute. But sure I removed it.

fasmat:
bors merge

fasmat:
bors merge

PR #5187, opened 2023-10-20
[Merged by Bors] - Move initialization to supervisor

fasmat:
## Motivation
Part of #5149  merge after #5186 & #5189

This moves the initialization code into the PoST supervisor. The ATX builder should not concern itself with initialization:
- A remote PoST service that connects is expected to already be fully initialized 
- In supervised mode the supervisor takes care of initialization

This allows to simplify the ATX builder, which has partially already been done in this PR and will continue in future PRs.

## Changes
- Removed dependency on `postSetupManager` from `atxBuilder`
- Simplified `atxBuilder`
  - Initial Proofs are not verified any more (PoST service is trusted)
  - Information about the PoST is fetched from the PoST service rather than accessed via the `postSetupManager`
- Simplified `postSetupManager`
  - Since meta information is now fetched via the client everything besides `StartSession`, `StopSession` and `Reset` has been removed from the interface
  - `postSetupManager` isn't used by the gRPC `SmeshingService` any more, instead it's now used by `PostSupervisor` to init a supervised node before starting it

For next PR: 
- change `atxBuilder` from `Start`/`Stop` to `Run`
  - a post service connecting triggers the builder loop (one per connected service)
  - simplify `atxBuilder` further and consider merging `nipostBuilder` into it.

## Test Plan
- added new tests to PoST supervisor to cover new functionality
- updated existing tests

## TODO


fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

activation/activation.go
```diff
@@ -334,71 +292,50 @@ func (b *Builder) generateInitialPost(ctx context.Context) error {
...
 	startTime := time.Now()
-	var err error
-	events.EmitPostStart(shared.ZeroChallenge)
-	post, metadata, err := b.proof(ctx, shared.ZeroChallenge)
+	post, _, err := b.proof(ctx, shared.ZeroChallenge)
 	if err != nil {
 		events.EmitPostFailure()
 		return fmt.Errorf("post execution: %w", err)
 	}
 	events.EmitPostComplete(shared.ZeroChallenge)
```
poszu: It's weird that `EmitPostStart` is called in `b.proof()` but `EmitPostComplete` / `Failure` is called here. How about moving these 2 events into `b.proof()` as well to keep it symmetrical?

activation/activation.go
```diff
@@ -676,21 +617,27 @@ func (b *Builder) createAtx(ctx context.Context, challenge *types.NIPostChalleng
...
 	}
 
+	client, err := b.postService.Client(b.nodeID)
+	if err != nil {
+		return nil, fmt.Errorf("get post client: %w", err)
+	}
+	info, err := client.Info(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("get post client info: %w", err)
+	}
```
poszu, line 642: At this point in time, the (remote, not supervised) post-service could have already been shut down. `Info` should probably be queried once, when the post-service becomes available and cached for the time of building ATX.

activation/post_supervisor_test.go
```diff
@@ -18,27 +23,86 @@ func Test_PostSupervisor_ErrorOnMissingBinary(t *testing.T) {
...
+	mgr := NewMockpostSetupProvider(ctrl)
+	mgr.EXPECT().PrepareInitializer(postOpts).Return(nil)
+	mgr.EXPECT().StartSession(gomock.Any()).Return(errors.New("failed start session"))
+
+	sync := NewMocksyncer(ctrl)
+	sync.EXPECT().RegisterForATXSynced().DoAndReturn(func() <-chan struct{} {
+		ch := make(chan struct{})
+		close(ch)
+		return ch
+	})
```
poszu: This pattern repeats over and over. How about a little helper to increase readability:
```go
func closedChannel() <-chan struct{} {
	ch := make(chan struct{})
	close(ch)
	return ch
}
```
```suggestion
	sync.EXPECT().RegisterForATXSynced().Return(closedChannel())
```

activation/post_supervisor_test.go
```diff
@@ -131,5 +233,87 @@ func Test_PostSupervisor_RestartsOnCrash(t *testing.T) {
...
+
+	for _, p := range providers {
+		if p.ID == initialization.CPUProviderID() {
+			return
+		}
+	}
+	require.Fail(t, "no CPU provider found")
+}
+
+func TestPostSetupManager_Benchmark(t *testing.T) {
```
poszu:
```suggestion
func Test_PostSupervisor_Benchmark(t *testing.T) {
```

activation/activation.go
poszu: Or you could just save the metadata that comes in the response with the proof.

fasmat: I extended the `postClient.Proof` method to return a `types.PostInfo` object instead of a `types.PostMetadata` object. This removes the need to fetch info from the post service again after already having received the proof in (most) cases.

All cases will be covered in future PRs where I re-write ATX builder to work with multiple identities and not store data on disk any more.

activation/post_supervisor_test.go
fasmat: I like to repeat myself in tests over factoring out code into helper/utility functions. Test code in my opinion should be dumb and obvious why it fails when it fails. 

Consider a future change where some tests break with `closedChannel()` and some don't. Often this leads to `closedChannel` becoming more complex instead of a second function being added for the changed behaviour in the failing tests. Over multiple iterations this leads to test code that is a) hard to change/extend without breaking existing tests and b) increased complexity where it's not obvious any more what is happening in a test.

This here is trivial enough, that I don't mind adding a helper function if you insist, but I'd prefer not to 🙂

poszu: In general, I agree with the statement that test helpers often become too big and do too much, which results in difficult future refactoring or adding new tests. But on the other hand, this helper would be super simple. What could be broken there realistically? The benefit is that the test code is not bloated with unnecessary code. `closedChannel()` represents the intention nicely.

common/types/post.go
```diff
@@ -6,6 +6,7 @@ type PostInfo struct {
 	CommitmentATX ATXID
 	Nonce         *VRFPostIndex
 
+	Challenge     []byte
```
poszu: `Challenge` is part of the proof, not POST metadata, but this structure is returned from `Info()` which only contains the post metadata.

fasmat:
bors try

poszu (approved):

fasmat:
bors merge

fasmat:
bors merge

PR #5188, opened 2023-10-22
[Merged by Bors] - keep activesets for one more epoch

dshulyak:
activesets in epoch X are used during epoch X and at the start of epoch X + 1. therefore we should not prune them
immediately at the start of epoch X.


poszu (approved):

dshulyak:
bors merge

PR #5190, opened 2023-10-23
[Merged by Bors] - Reformat to 120 columns

poszu:
## Motivation
Keep consistent code format

## Changes
- install golines in ./bin directory automatically by the makefile,
- run golines in dry mode as part of `lint` target. Fail if any lines would be changed,
- run golines in write mode as part of `lint-fix` target.


dshulyak (approved):

Makefile
```diff
@@ -42,6 +42,14 @@ FUZZTIME ?= "10s"
 all: install build
 .PHONY: all
 
+GOLINES_VERSION ?= v0.11.0
+GOLINES ?= ./bin/golines-$(GOLINES_VERSION)
+
+$(GOLINES):
+	rm -f $(dir $@)golines-*
+	GOBIN=$(abspath $(dir $@)) go install github.com/segmentio/golines@$(GOLINES_VERSION)
+	mv $(dir $@)golines $@
```
fasmat: Why treat `golines` differently than the other tools and not add it to `make install`?

Makefile
```diff
@@ -117,13 +125,21 @@ test-fmt:
 	git diff --exit-code || (git --no-pager diff && git checkout . && exit 1)
 .PHONY: test-fmt
 
-lint: get-libs
+lint: get-libs $(GOLINES)
 	./bin/golangci-lint run --config .golangci.yml
+	@output="$$($(GOLINES) -m 120 --dry-run --base-formatter=gofmt --ignore-generated .)"; \
```
fasmat: It would be cool if we could use https://github.com/mvdan/gofumpt here instead of `gofmt` since `golangci-lint` also verifies that the code has been formatted with this.

Makefile
poszu: IMHO, none of the tools should be installed globally by default. Others should be, in my opinion, refactored to be installed in the bin directory as well. I didn't want to do it as part of this PR though to keep it focused on 1 thing.

Makefile
poszu: I will see if that's possible. Out of curiosity: why is gofumpt better?

fasmat: it is stricter than `gofmt`. In a sense that code that passes `gofumpt` will also pass `gofmt` but not necessarily the other way round.

Makefile
```diff
@@ -117,13 +125,21 @@ test-fmt:
...
+	fi
+
 .PHONY: lint
 
-# Auto-fixes golangci-lint issues where possible.
-lint-fix: get-libs
+# Auto-fixes golangci-lint and golines issues where possible.
+lint-fix: get-libs $(GOLINES)
 	./bin/golangci-lint run --config .golangci.yml --fix
+	$(GOLINES) -m 120 -w --base-formatter=gofmt --ignore-generated .
```
fasmat: If you want to keep this you have to execute it before `golangci-lint`. `golangci-lint` will detect lines longer than 120 characters, not fix them and then exit with 1 causing golines not to be executed.

fasmat (approved):

fasmat:
bors merge

fasmat:
bors merge

PR #5191, opened 2023-10-23
[Merged by Bors] - Remove dependency on go.uber.org/atomic

poszu:
## Motivation
Removing dependency that is not really required.

## Changes
-  removed `go.uber.org/atomic`
- changed usages to `sync/atomic`

Regarding `syncedTargetTime atomic.Time` -> `time.Time`, I looked over the code, and I believe this variable is accessed from a single thread and doesn't need to be atomic. But please double-check :pray:.

It's used in the following way:

`App::startServices` -> `Syncer::Start/s.syncOnce.Do` -> spawned goroutine (**only once**) -> loop -> `Syncer::synchronize` -> `Syncer::setStateAfterSync` -> read & write


poszu:
bors try

fasmat (approved):

fasmat:
I also don't see `syncedTargetTime` being accessed concurrently, maybe this was the case some time ago but was refactored since then? `time.Time` should be OK as its type.

poszu:
bors merge

poszu:
bors merge

poszu:
bors cancel

poszu:
bors merge

poszu:
bors merge

PR #5193, opened 2023-10-24
GraphQL proof of concept (WIP)

lrettig:
## Motivation
https://github.com/spacemeshos/pm/issues/269

## Changes
- pulls in graphQL API from go-ethereum
- removes irrelevant/unnecessary schemas
- updates graphQL implementation to use spacemesh schema and types
- adds an API "backend" interface (for now used only by graphql but could be used by rpc in future)
- minimally refactors some existing grpc API code around these changes

## Test Plan
Imported code includes lots of tests. These still need to be updated.

## TODO

- [ ] Update [changelog](../CHANGELOG.md) as needed


lrettig:
For reference: this was quite old. We were previously experimenting with a graphQL-based API but decided to take a different path with the [v2 api](https://github.com/spacemeshos/api/tree/master/spacemesh/v2alpha1).

Issue #5192, opened 2023-10-24
followup for atx index


dshulyak:
will create a separate to follow to adjust those places in a different pr

## where it will be used?

- [x] no atx queries in ballot/proposal handlers (implemented in this change)
- [x] no identity and atx queries in hare
- [x] no atx queries in block generator https://github.com/spacemeshos/go-spacemesh/pull/5194
- [ ] atx queries during mesh hash resolution

## will allow to

- [x] drop atxs data from tortoise
- [ ] drop lru cache for atxs and identities

dshulyak:
in block generator we need access to coinbases, each one is 24 bytes. as of now they are not stored in atxsdata. several options:
- get them from database, block building is not on latency sensitive path, so if it will take 100-200ms more to build a block thats  should not a problem
- add coinbases to atxsdata structure, it will use 20MB per 1_000_000 atxs, checked in TestMemory

PR #5194, opened 2023-10-25
[Merged by Bors] - use atx index in block generator

dshulyak:
related: https://github.com/spacemeshos/go-spacemesh/issues/5192


blocks/utils.go
```diff
@@ -238,40 +253,33 @@ func rewardInfoAndHeight(logger log.Log, cdb *datastore.CachedDB, cfg Config, pr
...
+			return 0, nil, fmt.Errorf("%w: multiple proposals with the same ATX atx %v proposal %v", errDuplicateATX, p.AtxID, p.ID())
 		}
 		events.ReportProposal(events.ProposalIncluded, p)
 	}
-	// make sure we output coinbase in a stable order.
+	atxids := maps.Keys(weights)
+	// keys in order so that everyone generates same block
 	sort.Slice(atxids, func(i, j int) bool {
 		return bytes.Compare(atxids[i].Bytes(), atxids[j].Bytes()) < 0
 	})
```
poszu, line 276: How about instead of sorting it here, sort the rewards by ID in the place that creates a block?

poszu (approved):

blocks/utils.go
dshulyak: it makes sense. i am afraid to break something in this code, won't do that with this change

dshulyak:
bors merge

dshulyak:
bors merge

PR #5195, opened 2023-10-25
[Merged by Bors] - reduce tortoise window to 1 epoch and enforce min capacity for atxs data

dshulyak:
the practical impact here is that if your node ended up in a fork due to a bug (though it depends which one), 
malicious attack, or partition, you can rejoin original network without manual intervention.

if it happened within window from your current layer, we can do it fast because of keeping data in memory. 
and if happened before that then there is a slower approach which is not implemented.

there is no practical difference between 10_000 and 4032 so we can just tune it down and reduce memory usage


fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #5197, opened 2023-10-25
[Merged by Bors] - hare3: simplify code by removing active boolean

dshulyak:
even passive node will generate message but will not send it.
this is also better for debug, as anyone can see what is going on without setting up smesher.
and i will use it in a follow up to extend that message with additional stats, such as voting tallies


fasmat (approved):

poszu (approved):

dshulyak:
bors merge

dshulyak:
bors merge

PR #5198, opened 2023-10-25
Add ProcessExitGroup to ensure children are killed on exit

fasmat:
## Motivation
If the node exits unexpectedly or is killed with `SIGKILL` instead of `SIGTERM` it will not kill post services that it spawned.

This PR fixes this by adding a job object that kills processes assigned to it on close. For information on how this works see the official Microsoft documentation on job objects: https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-assignprocesstojobobject and https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-jobobject_basic_limit_information

## Changes
- On windows `ProcessExitGroup` adds the new process to a job object (itself) and ensures that they are closed when itself is closed
- On other systems the ProcessExitGroup only ensures that new processes are in the same process group as the main process.

For windows this ensures that if the main process crashes / is killed all commands started through the `ProcessExitGroup` will be killed as well. On other systems this is only true for crashes. Killing the node with `kill -9` will not cause the child processes to exit automatically. For this they have to listen to signals about their parent and close themselves.

## Test Plan
- n/a

## TODO

### DIFF
--- a/activation/post_supervisor_nix.go
+++ b/activation/post_supervisor_nix.go
@@ -2,4 +2,34 @@
 
 package activation
 
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+// DefaultPostServiceName is the default name of the post service executable.
 const DefaultPostServiceName = "service"
+
+// ProcessExitGroup holds references to processes that should be closed when the main process exits.
+type ProcessExitGroup struct{}
+
+// NewProcessExitGroup returns a new ProcessExitGroup.
+func NewProcessExitGroup() (ProcessExitGroup, error) {
+	return ProcessExitGroup{}, nil
+}
+
+// Dispose closes the ProcessExitGroup.
+func (g ProcessExitGroup) Dispose() error {
+	return nil
+}
+
+// StartCommand starts the given command and adds it to the ProcessExitGroup.
+func (g ProcessExitGroup) StartCommand(cmd *exec.Cmd) error {
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pgid: os.Getpid()}
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("start post service: %w", err)
+	}
+	return nil
+}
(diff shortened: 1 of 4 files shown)


activation/post_supervisor_nix.go
```diff
@@ -2,4 +2,33 @@
...
+}
+
+// StartCommand starts the given command and adds it to the ProcessExitGroup.
+func (g ProcessExitGroup) StartCommand(cmd *exec.Cmd) error {
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pgid: os.Getpid()}
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	return nil
+}
```
poszu: How about making it all private?

poszu (commented):
Would it make sense to add a unit test that would:
1. spawn process A that spawns process B
2. kill process A
3. observe that process B died

fasmat:
> Would it make sense to add a unit test that would:
> 
> 1. spawn process A that spawns process B
> 2. kill process A
> 3. observe that process B died

wouldn't work on linux (because the code actually doesn't work on non-windows systems)

activation/post_supervisor_nix.go
fasmat: wdym? These methods need to be called from outside the object?

poszu: But it's used in the same package, isn't it?

poszu:
> > Would it make sense to add a unit test that would:
> > 
> > 1. spawn process A that spawns process B
> > 2. kill process A
> > 3. observe that process B died
> 
> wouldn't work on linux (because the code actually doesn't work on non-windows systems)

I think it's fine to test windows-specific code only on windows :) Just skip/don't compile the test on a different OS.

I realize that the test would be much longer than the implementation, but on the other hand, the code for it is cryptic and prone to regressions :(

poszu:
This shouldn't be required with https://github.com/spacemeshos/post-rs/pull/141.

fasmat:
Superseded by https://github.com/spacemeshos/go-spacemesh/pull/5201

PR #5183, opened 2023-10-19
add smesher in coinbase

hairtail:
## Motivation

To fix incorrect reward smesher issue in explorer, this PR adds reward smeshers to rpc response.

## TODO

- [ ] Explain motivation or link existing issue(s)
- [ ] Test changes and document test plan
- [ ] Update documentation as needed
- [ ] Update [changelog](../CHANGELOG.md) as needed

PR #5199, opened 2023-10-25
[Merged by Bors] - vm, api: add smesherID to rewards

### LINKED ISSUE #4529: Send SmesherId in grpc reward object.
Currently, go-spacemesh sends only coinbase with the reward, which is not enough to identify which smesher got their reward. 
The ask is to send smesher id along with the coinbase, so we will be able to identify multiple smeshers which are working on one coinbase.

GlobalStateStream reward: https://github.com/spacemeshos/go-spacemesh/blob/develop/api/grpcserver/globalstate_service.go#L409

This is important to fix problem in explorer, because currently we cannot map coinbase with smesher id which is confusing for end users.
Explorer is using GlobalStateStream to get the rewards.

https://github.com/spacemeshos/explorer-frontend/issues/189

### LINKED ISSUE #4863: Tracking rewards
## Description
Rewards are generated in response to work done by a smesher, the system then sends the rewards to the coinbase (account) specified by that smesher.

Although the smesher id that was responsible for generating a reward is available at the time a reward is generated we do not currently store that in our database.

The current schema is:
```
    coinbase     CHAR(24),
    layer        INT NOT NULL,
    total_reward UNSIGNED LONG INT,
    layer_reward UNSIGNED LONG INT,
    PRIMARY KEY (coinbase, layer)
```

Smeshers are free to select their coinbase and also free to change their coinbase at each epoch. This means that there can be many smeshers sharing the same coinbase and also that over time a single smesher can be associated with multiple coinbase addresses.

Currently it seems there are users that are running multiple smeshers with a single coinbase and they are not able to discover how rewards were allocated between their smeshers, only the total rewards that all smeshers earned.

It seems like we would want to be able to retrieve rewards by smesher id, coinbase or both.

## Affected code
The storage and retrieval of rewards ocurrs at `sql/rewards/rewards.go`

This issue appears in commit hash: c7a7468f3bc2ed2a159c2c81e90297deb5757efc

## Approach
@countvonzero  suggested factoring out the purely functional parts of the VM's execution so that we could re run all blocks through there and use the results to populate a new rewards table without the possibility of accidentally mutating existing database state. Then at some later point we could switch the main code from the old table to the new table. Performing this work without mutating existing tables also provides the option to rollback if we encounter any unexpected issues.

lrettig:
## Motivation

See #4863

Based on #4850

Closes #4529
Closes #4850 
Closes #4863
Closes #4964
Closes #5183

## Changes
- Refactors `rewards` table. Adds smesherID column, migrates data. Does not update old data (smesherID will be NULL for old data, unless resynced).
- Updates database queries.
- Returns smesherID in all API queries


genvm/vm_test.go
```diff
@@ -431,6 +431,9 @@ func (t *tester) rewards(all ...reward) []types.CoinbaseReward {
 		rat := new(big.Rat).SetFloat64(rew.share)
 		rst = append(rst, types.CoinbaseReward{
 			Coinbase: t.accounts[rew.address].getAddress(),
+			// there can be many rewards per coinbase in the same layer,
+			// but each must have a distinct smesherID.
+			SmesherID: types.RandomNodeID(),
```
dshulyak: types.NodeID{byte(i)} . those types.RandomX is a mistake and should not be used at all

sql/migrations/state/0007_next.sql
```diff
@@ -1,3 +1,18 @@
 ALTER TABLE activesets ADD epoch INT DEFAULT 0 NOT NULL;
 CREATE INDEX activesets_by_epoch ON activesets (epoch asc);
 UPDATE activesets SET epoch = 7 WHERE epoch = 0;
+DROP INDEX rewards_by_layer;
```
dshulyak: i think 0007 was released. we should tag it as released (or atleast in the release candidate branch) and write new migration code to incremented version (0008)

sql/rewards/rewards.go
```diff
@@ -10,16 +10,13 @@ import (
 // Add reward to the database.
 func Add(db sql.Executor, reward *types.Reward) error {
 	if _, err := db.Exec(`
-		insert into rewards (coinbase, layer, total_reward, layer_reward) values (?1, ?2, ?3, ?4)
-		on conflict(coinbase, layer)
-			do update set
-				total_reward=add_uint64(total_reward, ?3),
-				layer_reward=add_uint64(layer_reward, ?4);`,
```
dshulyak, line 17: i think this was the only use case for add_uint64. could you please drop sql/functions.go

genvm/vm.go
```diff
@@ -261,12 +261,13 @@ func (v *VM) Apply(
...
 		return true
 	})
-	for _, reward := range rewardsResult {
+	for i, reward := range rewardsResult {
 		events.ReportRewardReceived(events.Reward{
 			Layer:       reward.Layer,
 			Total:       reward.TotalReward,
 			LayerReward: reward.LayerReward,
 			Coinbase:    reward.Coinbase,
+			SmesherID:   blockRewards[i].SmesherID,
```
dshulyak: SmesherID: reward.SmesherID

sql/rewards/rewards.go
```diff
@@ -37,17 +34,41 @@ func Revert(db sql.Executor, revertTo types.LayerID) error {
...
+			stmt.BindBytes(1, smesherID[:])
+		}, func(stmt *sql.Statement) bool {
+			coinbase := types.Address{}
+			stmt.ColumnBytes(0, coinbase[:])
+			reward := &types.Reward{
+				SmesherID:   smesherID,
+				Coinbase:    coinbase,
+				Layer:       types.LayerID(uint32(stmt.ColumnInt64(1))),
+				TotalReward: uint64(stmt.ColumnInt64(2)),
+				LayerReward: uint64(stmt.ColumnInt64(3)),
```
dshulyak: this two functions can be written without copying repetition

genvm/vm_test.go
fasmat: `types.BytesToNodeID(bytes)` is a helper that can also be used to convert `[]byte` to `NodeID`.

Why do you not like `types.RandomX`? In tests they are quite useful in my opinion.

sql/migrations/state/0007_next.sql
fasmat: Yes, this is my bad. I will update the `RELEASE.md` on in a separate PR to include information that this needs to be done when releasing a new version.

fasmat: PR: https://github.com/spacemeshos/go-spacemesh/pull/5250/

cmd/bootstrapper/generator_test.go
```diff
@@ -46,11 +46,17 @@ var bitcoinResponse2 string
...
-func (m *MeshAPIMock) GetRewards(types.Address) ([]*types.Reward, error) { panic("not implemented") }
-func (m *MeshAPIMock) GetLayer(types.LayerID) (*types.Layer, error)      { panic("not implemented") }
+func (m *MeshAPIMock) LatestLayer() types.LayerID        { panic("not implemented") }
+func (m *MeshAPIMock) LatestLayerInState() types.LayerID { panic("not implemented") }
+func (m *MeshAPIMock) ProcessedLayer() types.LayerID     { panic("not implemented") }
+func (m *MeshAPIMock) GetRewardsByCoinbase(types.Address) ([]*types.Reward, error) {
+	panic("not implemented")
+}
+
+func (m *MeshAPIMock) GetRewardsBySmesherId(types.NodeID) ([]*types.Reward, error) {
```
fasmat: `MeshAPIMock` is only used in a single place and can be replaced with `grpcserver.NewMockmeshAPI(gomock.NewController(tb))` there.

genvm/vm_test.go
dshulyak: they are not deterministic, i prefer to use randomness generated from constant seed or use deterministic values that, if that is simple enough.

sql/rewards/rewards.go
lrettig: you mean using generics? (the types are different.) it's going to be messy. or maybe I don't understand.

lrettig: I think I see what you mean, let me try

lrettig: this is the best I can do, still not sure it's better than what was there before, have a look

fasmat (approved):

dshulyak (approved):

lrettig:
@fasmat @dshulyak please take a look at the latest commit: https://github.com/spacemeshos/go-spacemesh/pull/5199/commits/1595ea9092224146556c0bfaf1813a74abf438d2

I had to remove the `WITHOUT ROWID` from the new `rewards` table definition to allow the migration of the old data, since it obviously doesn't have `pubkey` data (and this column is part of the primary key). This is a constraint on sqlite. Do we have concerns that this might impact performance?

sql/database.go
```diff
@@ -180,11 +180,7 @@ func Open(uri string, opts ...Opt) (*Database, error) {
 		}
 	}
 	for i := 0; i < config.connections; i++ {
-		conn := pool.Get(context.Background())
-		if err := registerFunctions(conn); err != nil {
-			return nil, err
-		}
-		pool.Put(conn)
+		pool.Put(pool.Get(context.Background()))
 	}
```
fasmat: I think this loop only existed for `registerFunctions` and can be removed completely.

fasmat:
> @fasmat @dshulyak please take a look at the latest commit: [1595ea9](https://github.com/spacemeshos/go-spacemesh/commit/1595ea9092224146556c0bfaf1813a74abf438d2)
> 
> I had to remove the `WITHOUT ROWID` from the new `rewards` table definition to allow the migration of the old data, since it obviously doesn't have `pubkey` data (and this column is part of the primary key). This is a constraint on sqlite. Do we have concerns that this might impact performance?

I think this will add 8 bytes per row that are persisted in the DB but not really needed even at ~ 1 mio entries we have for that table that's probably negligible since the PR also adds 32 bytes for the smesher ID. I don't think queries are slower with a Row ID, so I'd be OK with the change.

dshulyak:
what it does is that you have to query 2 b-trees instead of one. so if you have small unique key it is always better to have one b-tree. i can't tell without measuring, but generally single tree will be much better for range queries that hit that exact primary index (for example all rewards from smesher X, or all rewards inbetween from smesher X inbeetween layers Y and Z)

lrettig:
The only other option here is to force a resync from scratch, or to rerun all transactions through the VM to reconstruct the missing data. Even if we merge this PR as-is, those options are still on the table for the future - that would be my preference.

dshulyak:
yes i don't mind looking into any perf issues later

lrettig:
bors merge

PR #5207, opened 2023-10-27
[Merged by Bors] - Move NiPoST state into local state db

### LINKED ISSUE #5206: Add second database for local state

fasmat:
## Motivation
Part of #5206 

This PR adds a new database for local state and migrates state files from `activation.Builder` and from disk to this DB.

## Changes
- open a second database connection during node startup.
  - the original DB `state.sql` and code using it is left unchanged
  - new db `node_state.sql` is located in the same directory and is intended to be used for ephemeral data
- add initial migration creating nipost state table
- add code to `activation.Builder` that migrates existing state files from disk into db
- update `activation.Builder` to use state from DB instead of reading from disk

## Test Plan
- existing tests pass
- existing tests have been updated where needed
- new tests were added to test functionality of new database and migration code

## TODO


activation/activation.go
```diff
@@ -653,7 +669,9 @@ func (b *Builder) createAtx(ctx context.Context, challenge *types.NIPostChalleng
 	var nonce *types.VRFPostIndex
 	var nodeID *types.NodeID
 	if challenge.PrevATXID == types.EmptyATXID {
-		nodeID = &b.nodeID
+		nodeID = new(types.NodeID)
+		*nodeID = b.signer.NodeID()
+		// TODO(mafa): put this into the nipost state so we don't have to query it again from the post service
```
fasmat, line 674: TODO: extend nipost state with info about VRF

activation/activation_test.go
```diff
@@ -394,33 +398,49 @@ func TestBuilder_StopSmeshing_Delete(t *testing.T) {
 		AnyTimes()
 
 	// Create state files
+	// TODO(mafa): fully migrate to DB
```
fasmat, line 401: TODO: migrate nipost builder state as well

activation/activation.go
```diff
@@ -254,53 +255,97 @@ func (b *Builder) StopSmeshing(deleteFiles bool) error {
...
+	post, err := loadPost(b.nipostBuilder.DataDir())
+	switch {
+	case os.IsNotExist(err):
+		return nil // no post file, nothing to do
+	case err != nil:
+		return fmt.Errorf("loading post: %w", err)
+	default:
+	}
+
+	// TODO(mafa): for migration instead use initialization.LoadMetadata?
```
fasmat, line 294: Seems more reliable to me, opinions?

activation/activation_test.go
```diff
@@ -1301,3 +1387,55 @@ func TestWaitingToBuildNipostChallengeWithJitter(t *testing.T) {
...
+	}, nil)
+	require.NoError(t, tab.MovePostToDb())
+
+	challenge, err := nipost.Challenge(tab.localDb, tab.sig.NodeID())
+	require.NoError(t, err)
+	require.NotNil(t, challenge)
+	require.Equal(t, uint64(0), challenge.Sequence)
+	require.Equal(t, types.EmptyATXID, challenge.PrevATXID)
+	require.Equal(t, post, challenge.InitialPost)
+	require.NotNil(t, challenge.CommitmentATX) // TODO(mafa): also add VRFNonce
```
fasmat, line 1415: related to other TODO

checkpoint/recovery.go
```diff
@@ -331,7 +332,7 @@ func collectOwnAtxDeps(
 	}
 
 	// check for if miner is building any atx
-	nipostCh, _ := activation.LoadNipostChallenge(cfg.PostDataDir)
+	nipostCh, _ := activation.LoadNipostChallenge(cfg.PostDataDir) // TODO(mafa): fetch from DB instead
```
fasmat, line 335: TODO: update checkpoint to read from local state db instead of reading files from disk

checkpoint/recovery_test.go
```diff
@@ -537,7 +537,7 @@ func TestRecover_OwnAtxNotInCheckpoint_Preserve_IncludePending(t *testing.T) {
 	// write pending nipost challenge to simulate a pending atx still waiting for poet proof.
 	prevAtx := vatxs[len(vatxs)-2]
 	posAtx := vatxs[len(vatxs)-1]
-	require.NoError(t, activation.SaveNipostChallenge(cfg.PostDataDir, &types.NIPostChallenge{
+	require.NoError(t, activation.SaveNipostChallenge(cfg.PostDataDir, &types.NIPostChallenge{ // TODO(mafa): safe to DB instead
```
fasmat, line 540: same as above

node/node.go
```diff
@@ -1613,6 +1620,15 @@ func (app *App) setupDBs(ctx context.Context, lg log.Log) error {
 		datastore.WithConfig(app.Config.Cache),
 		datastore.WithConsensusCache(data),
 	)
+	localDb, err := sql.Open("file:"+filepath.Join(dbPath, localDbFile),
+		sql.WithMigrations(sql.LocalMigrations),
+		sql.WithConnections(app.Config.DatabaseConnections),
+		// TODO(mafa): add data migration to DB here
```
fasmat, line 1626: Not sure if having a `sql.WithV5Migration` here would be better than just having the to `MoveToDB` methods on the `activation.Builder`?

fasmat:
bors try

activation/activation.go
```diff
@@ -254,53 +255,97 @@ func (b *Builder) StopSmeshing(deleteFiles bool) error {
...
 			return err
 		}
-
 		return nil
 	default:
 		return fmt.Errorf("failed to stop smeshing: %w", err)
 	}
 }
 
+func (b *Builder) MovePostToDb() error {
```
fasmat, line 284: call these in `node.go` and add info about data migration to `CHANGELOG.md`

checkpoint/recovery.go
```diff
@@ -315,25 +319,22 @@ func collectOwnAtxDeps(
...
-	if ref == types.EmptyATXID && nipostCh == nil {
-		return nil, nil, nil
+	nipostCh, _ := nipost.Challenge(localDb, cfg.NodeID)
+	if ref == types.EmptyATXID {
+		if nipostCh == nil {
+			return nil, nil, nil
+		}
+		if nipostCh.CommitmentATX != nil {
+			ref = *nipostCh.CommitmentATX
+			// TODO(mafa): if nil query post service for commitment atx (in case it is currently initializing)
```
fasmat, line 336: For a local post service this works, but we cannot guarantee that any remote post service is online while we are checkpointing.

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

fasmat:
bors try

sql/migrations.go
```diff
@@ -34,9 +34,17 @@ func version(db Executor) (int, error) {
...
+}
+
+func LocalMigrations(db Executor) error {
+	return embeddedMigrations(db, "local")
+}
+
+func embeddedMigrations(db Executor, dbname string) error {
 	var migrations []migration
-	fs.WalkDir(embedded, "migrations", func(path string, d fs.DirEntry, err error) error {
+	fs.WalkDir(embedded, fmt.Sprintf("migrations/%s", dbname), func(path string, d fs.DirEntry, err error) error {
```
poszu, line 47:
```suggestion
	fs.WalkDir(embedded, filepath.Join("migrations", dbname), func(path string, d fs.DirEntry, err error) error {
```

node/node.go
```diff
@@ -988,6 +989,12 @@ func (app *App) initServices(ctx context.Context) error {
 		activation.WithPoetRetryInterval(app.Config.HARE.WakeupDelta),
 		activation.WithValidator(app.validator),
 	)
+	if err := atxBuilder.MovePostToDb(); err != nil {
+		app.log.Panic("failed to move initial post state to db: %v", err)
+	}
+	if err := atxBuilder.MoveNipostChallengeToDb(); err != nil {
+		app.log.Panic("failed to move nipost challenge to db: %v", err)
+	}
```
poszu: Would it make sense to make it a single call `atxBuilder.Migrate()`, and make these two private and call them inside? This way if we add more migrations to atxbuilder, the node code doesn't need to be updated.

datastore/nipost/nipost.go
```diff
@@ -0,0 +1,125 @@
...
+		stmt.BindBytes(5, ch.PositioningATX.Bytes())
+		if ch.CommitmentATX != nil {
+			stmt.BindBytes(6, ch.CommitmentATX.Bytes())
+		} else {
+			stmt.BindNull(6)
+		}
+		if ch.InitialPost != nil {
+			stmt.BindInt64(7, int64(ch.InitialPost.Nonce))
+			stmt.BindBytes(8, ch.InitialPost.Indices)
+			stmt.BindInt64(9, int64(ch.InitialPost.Pow))
```
poszu: Is this conversion sound? The k2pow nonce can be out of the int64 positive range.

datastore/nipost/nipost.go
```diff
@@ -0,0 +1,125 @@
...
+		} else {
+			stmt.BindInt64(7, -1)
+			stmt.BindNull(8)
+			stmt.BindInt64(9, -1)
```
poszu: Is it possible to assign NULL instead of -1?

datastore/nipost/nipost.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
```
poszu: It's quite "complicated"/long, how about extracting this part into something like `func nipostChallengeEncoder(nodeID types.NodeID, ch *types.NIPostChallenge) func(stmt *sql.Statement)` and reuse in `Add` and `Update` methods?

datastore/nipost/nipost.go
```diff
@@ -0,0 +1,125 @@
...
+		ch.InitialPost = &types.Post{}
+		if n := stmt.ColumnLen(6); n > 0 {
+			ch.InitialPost.Nonce = uint32(stmt.ColumnInt64(5))
+			ch.InitialPost.Indices = make([]byte, n)
+			stmt.ColumnBytes(6, ch.InitialPost.Indices)
+			ch.InitialPost.Pow = uint64(stmt.ColumnInt64(7))
+			return true
+		}
+
+		ch.InitialPost = nil
```
poszu: How about
```suggestion
		if n := stmt.ColumnLen(6); n > 0 {
			ch.InitialPost = &types.Post{
				Nonce: uint32(stmt.ColumnInt64(5))
				Indices: make([]byte, n)
				Pow: uint64(stmt.ColumnInt64(7))
			}
			stmt.ColumnBytes(6, ch.InitialPost.Indices)
			return true
		}
```

checkpoint/recovery.go
```diff
@@ -107,7 +107,15 @@ func Recover(
 		return nil, fmt.Errorf("open old database: %w", err)
 	}
 	defer db.Close()
-	preserve, err := RecoverWithDb(ctx, logger, db, fs, cfg)
+	localDBFile, err := sql.Open("file:"+filepath.Join(cfg.DataDir, cfg.LocalDbFile),
+		sql.WithMigrations(sql.LocalMigrations),
```
poszu: Why is the `DbFile` not opened with migrations, but the local one is?

activation/activation.go
```diff
@@ -381,24 +424,21 @@ func (b *Builder) buildNIPostChallenge(ctx context.Context) (*types.NIPostChalle
...
-	if err != nil {
-		if !errors.Is(err, sql.ErrNotFound) {
-			return nil, err
-		}
-	} else if prev.PublishEpoch == current+1 {
-		current += 1
+	prev, err := b.cdb.GetLastAtx(b.signer.NodeID())
+	switch {
+	case err == nil:
+		current = max(current, prev.PublishEpoch)
```
poszu, line 431: TBH, I don't understand why the current epoch is overwritten here. What if the previous ATX was published a year ago etc.? The `current++` part below won't fix this.

activation/activation.go
```diff
@@ -240,53 +239,92 @@ func (b *Builder) StopSmeshing(deleteFiles bool) error {
...
+	}
+	commitmentAtxId := types.BytesToATXID(meta.CommitmentAtxId)
+
+	ch := &types.NIPostChallenge{
+		InitialPost:   post,
+		CommitmentATX: &commitmentAtxId,
+	}
+	if err := nipost.AddChallenge(b.localDB, b.signer.NodeID(), ch); err != nil {
+		return fmt.Errorf("adding post to db: %w", err)
+	}
```
poszu: I'm not sure about keeping the initial POST as a part of `NipostChallenge`. I think it's unreliable because it is easy to accidentally wipe out the initial POST proof. I think this can already happen on a few paths when the nipostBuilder returns the `ErrATXChallengeExpired` error.

sql/migrations.go
fasmat: I used `filepath.Join` here before, which broke the windows build: it adds `\` on windows as path separator, which doesn't work with embedded, so I changed it to `fmt.Sprintf`

node/node.go
fasmat: I created a new Migrate method to migrate both states at once and made the change transactional so that when it fails it will be automatically rolled back.

datastore/nipost/nipost.go
fasmat: sqlite doesn't have unsigned integers. `BindInt64` calls `sqlite3_bind_int64` in the background and there is no `sqlite3_bind_uint64`. The cast from `uint64` to `int64` might overflow but that only means that values between 2^63 and 2^64-1 will be cast to negative numbers and then back to positive numbers when reading them again. We are doing this with all our `uint64`s we store in the database. There are only 2 other options:

- store the integer as string
- store the integer as bytes

both of which require conversion from and to string/bytes 🤷

datastore/nipost/nipost.go
fasmat: I can bind `NULL` to a nullable value, but `stmt.ColumnInt64` will then return `0` for that value if used and I thought `-1` was a better "not available" value.

Since the other two values are assumed to be `NULL` anyway when `Indices` is `NULL`, it doesn't hurt to use `stmt.BindNull` 🙂

datastore/nipost/nipost.go
poszu: I understand, but is the conversion uint64 -> int64 -> uint64 here lossless? If yes, then fine.

datastore/nipost/nipost.go
fasmat: I am not convinced to extract the encoding function from `Add` and `Update`. The are only the same by coincidence and do not depend on `nodeID` and `types.NIPostChallenge` but rather on the specific SQL query of `Add` and `Update`.

If the encoding is extracted and shared between both methods changing one of the two SQL queries might break the encoding function for it or the other 😕

sql/migrations.go
poszu: Funny, both `embed` and `filepath` are in stdlib and they can't agree on directory separators :see_no_evil: .

datastore/nipost/nipost.go
poszu: I don't think it's a coincidence. The whole point of extracting is to have them equal to avoid bugs :shrug:. It can always be unsplit if there is actually a need for different implementations.

datastore/nipost/nipost.go
fasmat: I did a quick test and saw no issues with it, a quick google search also gave recommendations to just treat the unsigned integer as signed.

datastore/nipost/nipost.go
fasmat: The sql query could state the columns in any order. The reason the code is the same is because I assigned the same columns to the same indices, but I didn't have to.

checkpoint/recovery.go
fasmat: `sql.Open` defaults to `sql.WithMigrations(sql.StateMigrations)` if no other set of migrations is specified. This was the case before the change as well. I can make it explicit, such that the state database also needs to explicitly define which set of migrations it uses, but that would mean changing a lot of tests (`sql.InMemory` just calls `sql.Open` with a special path)

activation/activation.go
fasmat: If the previous ATX was published a year ago it's publish epoch will be 26 less then `current` -> `max(current, current-26)` is `current`.

Before the code was:

```go
if prev.PublishEpoch == current+1 {
	current += 1
}
```

there seems to have been a need for this check. It also makes intuitively sense; if there was a previous ATX we want to set `current` to at least its `PublishEpoch` so we don't accidentally try to publish two ATXs in the same epoch. But the check itself was weird: why only check if publish epoch was `current+1`?

Regarding the `current++` check I added below: I added this to avoid the situation where if one misses the registration window for the poet round the node would complain about it every 5 minutes from PoET round start (Monday morning) until the end of the epoch (Friday morning). It also prevents clearing the challenge in the case this happens with the initial ATX.

poszu: Thanks, you are right about the atx year ago, my bad.

checkpoint/recovery.go
poszu: :unamused: 

OK, thx for explanation

activation/activation.go
fasmat: In my opinion the problem here is that we return `ErrATXChallengeExpired` when we want to clear the challenge instead of just doing that with `nipost.RemoveChallenge` immediately. It feels like a hidden goto.

I think we shouldn't handle the initial PoST any different from all the other posts we do. We could even go as far as never deleting a challenge but just updating it when it becomes stale. In any case where `ErrATXChallengeExpired` is returned the only things that need to be updated in the challenge are the `PositioningATX` and the `PublishEpoch`, all other fields (must) stay the same.

datastore/nipost/nipost.go
fasmat: I extracted the encoding function for now, but I might split it again in the next PR.

activation/activation.go
fasmat: I separated the initial post from the challenge in the local DB. The initial post is only cleared on successful publication, the challenge on successful publication and when it expires.

fasmat:
bors try

fasmat:
bors try

activation/activation.go
```diff
@@ -67,32 +68,27 @@ type Config struct {
...
 	eg errgroup.Group
 
 	signer           *signing.EdSigner
 	accountLock      sync.RWMutex
-	nodeID           types.NodeID
 	coinbaseAccount  types.Address
 	goldenATXID      types.ATXID
 	regossipInterval time.Duration
 	cdb              *datastore.CachedDB
+	localDB          *datastore.LocalDB
```
dshulyak: please use sql instance directly. datastore module will be removed

datastore/state.go
```diff
@@ -0,0 +1,11 @@
+package datastore
+
+import "github.com/spacemeshos/go-spacemesh/sql"
+
+type LocalDB struct {
+	*sql.Database
+}
```
dshulyak: lets store this code in some place other than datastore. maybe localsql or sql/local

fasmat: I moved the code to the `localsql` package and made it a bit easier to use.

activation/activation.go
fasmat: I moved the code to the `localsql` package and made it easier to use.

activation/activation.go
```diff
@@ -240,53 +239,110 @@ func (b *Builder) StopSmeshing(deleteFiles bool) error {
...
+}
+
 // SmesherID returns the ID of the smesher that created this activation.
 func (b *Builder) SmesherID() types.NodeID {
-	return b.nodeID
+	return b.signer.NodeID()
 }
 
-func (b *Builder) generateInitialPost(ctx context.Context) error {
+func (b *Builder) initialPost(ctx context.Context) error {
```
poszu: Why change the name? IMHO, it's now hard to tell what it does.

activation/activation.go
```diff
@@ -472,74 +527,55 @@ func (b *Builder) Coinbase() types.Address {
...
-		b.pendingATX, err = b.createAtx(ctx, challenge)
-		if err != nil {
-			return fmt.Errorf("create ATX: %w", err)
+	for {
+		size, err := b.broadcast(ctx, atx)
+		if err == nil {
+			b.log.Info("atx published", zap.Inline(atx), zap.Int("size", size))
+			break
+		}
+		if errors.Is(err, context.DeadlineExceeded) {
```
poszu: What if `err == context.Canceled`?

poszu (approved):

activation/activation.go
fasmat: This is actually a hack to make one test easier. If the context expires it will be caught by the `select` statement after this check.

Looking again at it I think I can adapt the test to not require this hack any more.

fasmat: I updated the test so that the special handling of `context.DeadlineExceeded` isn't necessary any more.

activation/activation.go
fasmat: Renamed to `buildInitialPost`

fasmat:
bors merge

fasmat:
bors merge

fasmat:
bors merge

PR #5216, opened 2023-11-02
[Merged by Bors] - hare3: count all early message as on time

dshulyak:
this change fixes a bug that would decrease grade of significantly earlier message, 
such that was created by earlier by one more network delay.


hare3/types.go
```diff
@@ -40,7 +40,7 @@ type IterRound struct {
 
 // Delay returns number of network delays since specified iterround.
 func (ir IterRound) Delay(since IterRound) int {
```
poszu: How about instead avoid type casting (return uint32) and just check if `since` > `ir`:
```go
func (ir IterRound) Delay(since IterRound) uint32 {
	if ir.Absolute() > since.Absolute() {
		return ir.Absolute() - since.Absolute()
	}
	return 0
}
```

poszu (approved):

dshulyak:
bors merge

dshulyak:
bors merge

fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors cancel

dshulyak:
bors merge

PR #5217, opened 2023-11-02
[Merged by Bors] - tortoise: collect details on the number of specified layers

dshulyak:
this will collect:
- opinion on latest layer
- block that should be applied on latest layer
- number of blocks by layer 

this collector is expensive and should be enabled only for debug. to enable it set `tortoise-collect-details: 200` in `tortoise` section. number of layer that are metered can't be larger then the window size.


tortoise/metrics.go
```diff
@@ -155,11 +155,46 @@ var (
...
 
-// LayerDistanceToBaseBallot checks how far back a node needs to find a good ballot.
-var layerDistanceToBaseBallot = metrics.NewHistogramWithBuckets(
-	"layer_distance_to_base_ballot",
-	namespace,
-	"How far back a node needs to find a good ballot",
-	[]string{},
-	prometheus.ExponentialBuckets(1, 2, 16),
-).WithLabelValues()
+var _ prometheus.Collector = (*collector)(nil)
```
fasmat: This isn't needed; if `collector` wouldn't implement `prometheus.Collector` line 161 would fail to compile with an error message that it doesn't 🙂

fasmat (approved):

dshulyak:
bors merge

dshulyak:
bors merge

dshulyak:
bors merge

PR #5219, opened 2023-11-02
[Merged by Bors] - activation: update NIPostBuilder to use state DB instead of files

### LINKED ISSUE #5206: Add second database for local state

fasmat:
## Motivation
Closes #5206

## Changes
- Migrations for `localsql` were extended
  - All migrations are now in code to ensure that data is moved from disk into DB when migrating, failing to do so will result in the DB migration not completing
  - All Go Migrations now have a SQL migration that does the same DB changes, but doesn't migrate data from disk into the DB. This is helpful for when no data needs to be migrated (e.g. in tests). Additionally the Go Migrations can just be deleted in the future when most people are running v1.3.0 or newer versions.
  - Go migrations have tests to ensure data is migrated properly and SQL migrations match the Go Migrations (produce the same Schema after migrating)
- NIPostBuilder now stores its state inside the local database instead of on disk
  - Added new `nipost` localsql functions that allow to store and fetch the data
- Deflake some tests in the `activation/e2e` package by avoiding the spawned PoET from having slightly different timings than the component under test.

## Test Plan
- existing tests have been updated to work with new code
- added tests for new code

## TODO


activation/activation_test.go
```diff
@@ -400,8 +400,9 @@ func TestBuilder_StopSmeshing_Delete(t *testing.T) {
 
 	// Create state files
 	// TODO(mafa): fully migrate to DB
```
fasmat, line 402: TODO

activation/activation_test.go
```diff
@@ -1377,13 +1379,14 @@ func TestBuilder_MovePostToDb(t *testing.T) {
...
 	require.NotNil(t, challenge.CommitmentATX) // TODO(mafa): also add VRFNonce
 	require.Equal(t, commitmentAtx, *challenge.CommitmentATX)
-	require.NoFileExists(t, filepath.Join(tab.nipostBuilder.DataDir(), postFilename))
+	require.NoFileExists(t, filepath.Join(dataDir, postFilename))
 
-	require.NoError(t, tab.MovePostToDb()) // should not fail if post is already in db
+	require.NoError(t, tab.MovePostToDb(dataDir)) // should not fail if post is already in db
 }
 
 func TestBuilder_MoveNipostChallengeToDb(t *testing.T) {
```
fasmat, line 1387: TODO: add migration code + tests for nipost builder state

activation/nipost.go
```diff
@@ -211,21 +174,28 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
-		if err != nil {
+
+		if err := nb.submitPoetChallenges(submitCtx, poetProofDeadline, challenge.Hash().Bytes()); err != nil {
 			return nil, fmt.Errorf("submitting to poets: %w", err)
 		}
-		if len(poetRequests) == 0 {
-			return nil, &PoetSvcUnstableError{msg: "failed to submit challenge to any PoET", source: submitCtx.Err()}
+		count, err := nipost.PoetRegistrationCount(nb.localDB, nb.signer.NodeID())
+		if err != nil {
+			// TODO(mafa): panic instead?
```
fasmat, line 183: Opinions? Alternatively I could revert `nb.submitPoetChallenges` to return the list of registrations

activation/nipost.go
```diff
@@ -211,21 +174,28 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
+			return nil, &PoetSvcUnstableError{msg: "failed to submit challenge to any PoET"}
 		}
+	}
 
-		nb.state.Challenge = challengeHash
-		nb.state.PoetRequests = poetRequests
-		nb.persistState()
+	poetProofRef, err := nipost.PoetProofRef(nb.localDB, nb.signer.NodeID())
+	if err != nil {
+		// TODO(mafa): panic instead?
```
fasmat, line 193: same as above; before if the state could not be read from disk it would just assume that there were not registrations yet.

activation/nipost.go
```diff
@@ -240,20 +210,27 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
 			return nil, &PoetSvcUnstableError{msg: "getBestProof failed", source: err}
 		}
 		if poetProofRef == types.EmptyPoetProofRef {
 			return nil, &PoetSvcUnstableError{source: ErrPoetProofNotReceived}
 		}
-		nb.state.PoetProofRef = poetProofRef
-		nb.state.NIPost.Membership = *membership
-		nb.persistState()
+		if err := nipost.UpdatePoetProofRef(nb.localDB, nb.signer.NodeID(), poetProofRef); err != nil {
+			// TODO: panic instead?
```
fasmat, line 221: `nb.persistState()` did also only log on warning that it couldn't persist but would continue.

activation/nipost.go
```diff
@@ -240,20 +210,27 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
+			nb.log.Warn("cannot set poet proof ref", zap.Error(err))
+		}
+		if err := nipost.UpdateNIPostMembership(nb.localDB, nb.signer.NodeID(), membership); err != nil {
+			// TODO: panic instead?
```
fasmat, line 225: nb.persistState() did also only log on warning that it couldn't persist but would continue.

activation/nipost.go
```diff
@@ -240,20 +210,27 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
+			nb.log.Warn("cannot set nipost membership", zap.Error(err))
+		}
 	}
 
+	var state *types.NIPostBuilderState
```
fasmat, line 230: TODO: fetch from DB if available

activation/nipost.go
```diff
@@ -269,31 +246,36 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
 
 	nb.log.Info("finished nipost construction")
-	return nb.state.NIPost, nil
+	return &types.NIPost{
+		Post: state.NIPost.Post,
+		// Membership: membership,
+		PostMetadata: &types.PostMetadata{
+			Challenge: poetProofRef[:],
+			// LabelsPerUnit: postInfo.LabelsPerUnit,
+		},
```
fasmat, line 277: TODO: re-add missing data

activation/nipost.go
```diff
@@ -503,21 +421,30 @@ func (nb *NIPostBuilder) getBestProof(
...
 
 	var eg errgroup.Group
-	for _, r := range nb.state.PoetRequests {
-		logger := nb.log.WithContext(ctx).
-			WithFields(log.String("poet_id", hex.EncodeToString(r.PoetServiceID.ServiceID)), log.String("round", r.PoetRound.ID))
-		client := nb.getPoetClient(ctx, r.PoetServiceID)
+	for _, r := range registrations {
+		logger := nb.log.With(
+			zap.String("poet_address", r.Address),
+			// zap.String("round", r.PoetRound.ID),
```
fasmat, line 435: TODO: Add PoetRound ID to local state

activation/nipost.go
```diff
@@ -503,21 +421,30 @@ func (nb *NIPostBuilder) getBestProof(
...
+		client := nb.getPoetClient(ctx, r.Address)
 		if client == nil {
-			logger.Warning("poet client not found")
+			logger.Warn("poet client not found")
 			continue
 		}
-		round := r.PoetRound.ID
-		waitDeadline := proofDeadline(r.PoetRound.End.IntoTime(), nb.poetCfg.CycleGap)
+		// round := r.PoetRound.ID
+		round := "0" // TODO(mafa): fix this
```
fasmat, line 444: TODO: fetch PoetRound ID from local state

common/types/nipost.go
```diff
@@ -1,8 +1,11 @@
 package types
 
+// TODO(mafa): remove these types after all is fully migrated to local DB
```
fasmat, line 3: TODO: consider moving them from here to `nipost_state` with the other deprecated functions

sql/migrations/local/0001_initial.sql
```diff
@@ -1,12 +1,28 @@
...
     pos_atx       CHAR(32) NOT NULL,
     commit_atx    CHAR(32),
     post_nonce    UNSIGNED INT,
     post_indices  VARCHAR,
-    post_pow      UNSIGNED LONG INT
-) WITHOUT ROWID;
\ No newline at end of file
+    post_pow      UNSIGNED LONG INT,
+
+    vrf_nonce     UNSIGNED LONG,
```
fasmat, line 14: TODO: use this field

sql/migrations/local/0001_initial.sql
```diff
@@ -1,12 +1,28 @@
...
+
+    poet_proof_ref    CHAR(32),
+    nipost_membership VARCHAR
```
fasmat, line 17: TODO: use this field

activation/nipost_test.go
```diff
@@ -41,7 +41,38 @@ func defaultLayerClockMock(ctrl *gomock.Controller) *MocklayerClock {
...
+		mclock,
+	)
+	require.NoError(t, err)
+
+	// TODO(mafa): add state
+
+	err = nb.ResetState()
+	require.NoError(t, err)
+
+	// TODO(mafa): check state is empty
```
fasmat, line 72: TODO: complete test

activation/nipost.go
```diff
@@ -41,37 +42,17 @@ const (
...
-	}
-	if state.Challenge == challenge {
-		nb.state = state
-	} else {
-		nb.log.Info("discarding stale nipost state")
-		nb.state = &types.NIPostBuilderState{Challenge: challenge, NIPost: &types.NIPost{}}
-	}
-}
-
-func (nb *NIPostBuilder) persistState() {
```
fasmat, line 58: TODO: add migration code from old state to new

activation/nipost.go
```diff
@@ -212,21 +187,24 @@ func (nb *NIPostBuilder) BuildNIPost(ctx context.Context, challenge *types.NIPos
...
-		poetRequests, err := nb.submitPoetChallenges(submitCtx, poetProofDeadline, challengeHash.Bytes())
-		if err != nil {
+		if err := nb.submitPoetChallenges(submitCtx, poetProofDeadline, challenge.Hash().Bytes()); err != nil {
 			return nil, fmt.Errorf("submitting to poets: %w", err)
 		}
-		if len(poetRequests) == 0 {
+		count, err := nipost.PoetRegistrationCount(nb.localDB, nb.signer.NodeID())
+		if err != nil {
+			nb.log.Panic("failed to get poet registration count", zap.Error(err))
+		}
```
poszu: It seems that the code always reacts on error with a panic. How about inlining the err check + panic into `nipost.PoetRegistrationCount()`, perhaps naming it `MustPoet....()`?

common/types/nipost.go
```diff
@@ -1,27 +1,7 @@
 package types
 
-//go:generate scalegen -types NIPostBuilderState,PoetRequest,PoetServiceID
+//go:generate scalegen
 
 type PoetServiceID struct {
 	ServiceID []byte `scale:"max=32"` // public key of the PoET service
 }
```
poszu: I think you could remove `PoetServiceID` as well (along with the related getter in `poetClient interface`.

sql/localsql/0003_migration.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if len(state.PoetRequests) == 0 {
+		return discardBuilderState(dataDir) // Phase 0: Submit PoET challenge to PoET services
+	}
+	// Phase 0 completed.
+	for _, req := range state.PoetRequests {
+		address, err := m.getAddress(req.PoetServiceID)
+		if err != nil {
+			return fmt.Errorf("get address for poet service id %x: %w", req.PoetServiceID.ServiceID, err)
+		}
```
poszu, line 118: I'm not sure about this. All it takes is one poet client down or unresponsive and it will fail migration. What if we roll this out and then one of the poets crashes completely?

IMHO, to make this migration reliable we should first change the way poets are configured. Currently the `BaseConfig` takes just a slice of poet addresses:
```go
PoETServers []string `mapstructure:"poet-server"`
```
How about we change it to take an address and a poet's public key?
```go
type PoetServer struct {
	Address	string		`mapstructure:"address"`
	Pubkey	Base64Enc	`mapstructure:"pubkey"`
}

type BaseConfig struct {
...
	PoetServers []PoetServer `mapstructure:"poets"`
}
```
and we stop querying the poet service for its pubkey?

activation/nipost.go
fasmat: An error here means that there is some issue that the program cannot easily recover from (SQL query cannot be parsed, DB connection dropped, etc.) and it makes sense to exit the program. I'm not a big fan of moving the panic into the function. `PoetRegistrationCount` is only called in two places and to safe this check would require passing the logger to the DB (which we do not do anywhere else).

common/types/nipost.go
fasmat: At the moment I still need the getter for the migration from state-files to DB. (see also my comment there)

sql/localsql/0003_migration.go
fasmat: A failing migration should revert all changes that happened during the DB migration. On that note I probably need to not delete the migrated `*.bin` files but rather rename them (`*.bak`) and only delete them after the DB migration successfully completes (or leave them there in case we want a way to be able to downgrade if something goes wrong).

> What if we roll this out and then one of the poets crashes completely?

Ideally users can revert back to 1.2.6 and continue without issues.

EDIT: I also don't think the whole network will update at once. We don't have to force this update. Even if we would it will probably take at least an hour for most nodes to update and we are only querying `/info` which I thought was cached?

I agree we shouldn't fail if any request fails, but maybe only if ALL fail (the same way we do in NIPostBuilder). This however has the drawback that some PoETs might be dropped during migration that will then be online again during the cyclegap 😕.

> IMHO, to make this migration reliable we should first change the way poets are configured.

This will probably break things. Changes to the config are not trivial in the setup we are currently using, since the config is overwritten by Smapp from the config here: https://configs.spacemesh.network/config.mainnet.json So any change we make must work with both old and new nodes, i.e. we need to have poets twice in the config (old and new layout).

I think maybe we should just hard code a map of the official PoETs (Key -> URL) and provide a way for users to provide such a mapping for their custom PoETs if needed; but not via the config but a new file that is only read during migration and can be discarded afterwards.

activation/nipost.go
poszu: Right, hence the panic - I'm totally fine with it. I'm only proposing to move the panic inside the function. You don't need to pass the logger to do that - just call `panic(...)`. IMHO it makes more sense to handle bugs like a broken SQL query inside of the function rather than leaking this to the caller (who has no idea why it could fail).

common/types/nipost.go
poszu: Right, but you moved the other ones into the migration files. Why keep this one here?

sql/localsql/0003_migration.go
poszu: By "poet crashes completely" I meant a situation when one of the poets dies and doesn't boot anymore, and needs maintenance. In such a situation, all nodes would fail the migration until we fix the poet (if we even can fix it - it could take a week or so). In effect, the upgrade would be blocked until we fix the poet, which IMHO is a no-go.

---

>>   IMHO, to make this migration reliable we should first change the way poets are configured.
>
> This will probably break things. Changes to the config are not trivial in the setup we are currently using, since the config is overwritten by Smapp from the config here: https://configs.spacemesh.network/config.mainnet.json So any change we make must work with both old and new nodes, i.e. we need to have poets twice in the config (old and new layout).

Are you sure? Can't Smapp use config **A** for node version **a** and config **B** for node version **b**? Even if so, I think having poets temporarily configured twice (two separate fields) in the ` https://configs.spacemesh.network/config.mainnet.json` is not a big deal. It doesn't mean that the config structure of the new node version needs both. It will only contain the new field and the older one should be omitted during parsing.

We should communicate upgrade steps properly in the release nodes with an instruction on how to migrate the config in case somebody overwrites the list of poet servers (the default config would contain the list of SM poets with their pubkeys).

I can create a PR to change poet configuration as described.

common/types/nipost.go
fasmat: I thought that it would cause cyclic dependencies which doesn't seem to be the case so I moved it. 🙂 

With `v1.4.0` I would like to get rid of all the old code handling state files (`nipost_state*.go` in the `localsql` package) and also remove the data migration from disk to DB. New nodes don't need it and old nodes can upgrade `v1.2.x` -> `v1.3.x` -> `v1.4.x` to keep their state. We can then also remove the `PoetServiceID` method on `HTTPPoetClient`.

Wdyt?

activation/nipost.go
fasmat: Interestingly, we seem to not have used `select count(*)` before in any other query, so there is no precedence to compare against.

I'm a bit hesitant to make `PoetRegistrationCount` directly panic on an error, because most if not all DB queries currently handle errors from `db.Exec` by just wrapping and returning them to the caller. The only way `db.Exec` however can fail is by either a) the connection to the DB was closed or b) the query is syntactically wrong. In both cases it makes no sense to continue and the node should just panic in my opinion; not just here.

I think the better way is to change `db.Exec` to not return an error and panic instead - to be consistent across all DB queries and if its OK for you I would do that in a dedicated PR.

sql/localsql/0003_migration.go
fasmat: I see three options:

- fail the migration if a PoET isn't available while the migration happens. This will revert all changes and users can try again or users can downgrade to v1.2.6 and try upgrading later. As you said that's probably not the best solution since worst case scenario people might need to wait a full epoch to be able to upgrade (then the PoET wouldn't be in the new `nipost_builder_state.bin` file any more if it is still down)
- not fail the migration if at least one PoET is available when the migration happens. This would match the behaviour that the `NIPostBuilder` uses to fetch PoET proofs and would be my preferred solution. A PoET that is down (error or maintenance) is unlikely to have the highest tick count anyway, so losing that registration might not even matter.
- change the config to include both the URL and the public key of a PoET and force users to update their config accordingly. I would like to avoid that change, people might use the wrong public keys for the nodes (copy & paste errors, typos, swap keys accidentally) which can lead to multiple problems that might not surface until the cyclegap.

> Can't Smapp use config A for node version a and config B for node version b?

Afaik at the moment it cannot; also a lot of users don't use Smapp.

> We should communicate upgrade steps properly in the release nodes with an instruction on how to migrate the config in case somebody overwrites the list of poet servers (the default config would contain the list of SM poets with their pubkeys).

That could work if you think that the automatic "at least one PoET" migration is too risky.

Wdyt?

activation/nipost.go
fasmat: I created the following issue to keep track of the change: https://github.com/spacemeshos/go-spacemesh/issues/5298

dshulyak: it is not the exhaustive list. it can also exist with https://www.sqlite.org/rescode.html#busy depending on how it is used

fasmat: Changed to instead return the error (so it will be logged and a bit later retried).

sql/localsql/0003_migration.go
poszu: I finished and opened PR allowing to configure poets' public keys: https://github.com/spacemeshos/go-spacemesh/pull/5293.

Regarding:
>>Can't Smapp use config A for node version a and config B for node version b?
>
>Afaik at the moment it cannot; also a lot of users don't use Smapp.

@nj confirmed that it can serve different configs based on node version.

fasmat: @poszu Updated the PR after #5293 was merged. It now uses the public keys from the config instead of querying them.

fasmat:
bors try

fasmat:
bors try

CHANGELOG.md
```diff
@@ -150,6 +150,11 @@ for more information on how to configure the node to work with the PoST service.
   query rewards by smesherID. Additionally, it does not re-index old data. Rewards will contain smesherID going forward,
   but to refresh data for all rewards, a node will have to delete its database and resync from genesis.
 
+* [#5219](https://github.com/spacemeshos/go-spacemesh/pull/5219) Migrate data from `nipost_builder_state.bin` to `node_state.sql`.
+
+  The node will automatically migrate the data from disk and store it in the database. The migration will take place at the
+  first startup after the upgrade.
+
```
poszu: This probably needs to be relocated to either v1.3.1 or _unreleased_.

activation/e2e/nipost_test.go
```diff
@@ -177,22 +179,24 @@ func TestNIPostBuilderWithClients(t *testing.T) {
 		return err == nil
 	}, 10*time.Second, 100*time.Millisecond, "timed out waiting for connection")
 
+	db := localsql.InMemory()
+	challenge := types.NIPostChallenge{
+		PublishEpoch: postGenesisEpoch + 2,
+	}
+	require.NoError(t, nipost.AddChallenge(db, sig.NodeID(), &challenge))
```
poszu: Why does the challenge need to be added?

activation/poet.go
```diff
@@ -39,7 +40,7 @@ type PoetPoW struct {
 // HTTPPoetClient implements PoetProvingServiceClient interface.
 type HTTPPoetClient struct {
 	baseURL       *url.URL
-	poetServiceID types.PoetServiceID
+	poetServiceID localsql.PoetServiceID
```
poszu: How about not depending on `localsql` here and just use `[]byte`?

sql/localsql/0003_migration_interfaces.go
```diff
@@ -0,0 +1,12 @@
+package localsql
+
+import (
+	"context"
+)
+
+//go:generate mockgen -typed -package=localsql -destination=./0003_mocks.go -source=./0003_migration_interfaces.go
+
+type PoetClient interface {
```
poszu, line 9: I think there's no need to export it.
```suggestion
type poetClient interface {
```

sql/localsql/nipost_state.go
```diff
@@ -136,6 +137,30 @@ func discardNipostChallenge(dir string) error {
 	return nil
 }
 
+func saveBuilderState(dir string, state *NIPostBuilderState) error {
+	if err := save(filepath.Join(dir, builderFilename), state); err != nil {
+		return fmt.Errorf("saving builder state: %w", err)
+	}
+	return nil
+}
```
poszu: The `saveBuilderState` and `discardBuilderState` are only used in tests. How about moving them to _test.go?

poszu (approved):

sql/localsql/0003_migration_interfaces.go
fasmat: I tried to have it private, but I need it in `node.go` (around line 1620) when I instantiate the poet clients and pass them to the migration that needs it. Since the `node` package already imports `localsql` this made sense to me but wdyt?

sql/localsql/nipost_state.go
fasmat: `saveBuilderState` is only used by the test and I can move it there, but `discard-` and `loadBuilderState` are needed by the migration.

activation/e2e/nipost_test.go
fasmat: seems to be a leftover

fasmat:
bors try

sql/localsql/0003_migration_interfaces.go
poszu: I don't understand why the interface needs to be exported. The poet HTTP client from the `activation` pkg satisfies it regardless (of exported/private) so it should work. Why not?

fasmat: I tried it, but the call in `node.go` takes a slice of interfaces and Go doesn't support covariance on slices. I would need to make `New0003Migration` take `[]*activation.HTTPClient` if we don't export the interface, but that would complicate testing and requires some refactoring (to avoid cyclic dependencies).

fasmat:
bors merge

